PR #2054, opened 2019-12-13
support/render/problem: make problem package in an isolated instance without globals

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Make the `problem` package usable without needing to use a single global state.

Move all package functionality into a new type that has the same interface as the package and can be instantiated with `New` and used and tested in isolation. Keep the package functions, but have them call a single default instance.

The interface of this package remains identical and unchanged, except for the removal of the exported global variable `ServiceHost`.

### Why

I'd like to extract the error rendering logic out of this package and separate it from the logic that is specific to rendering problem errors. It's easier to layer packages if they aren't using globals, and so the first step is to make this package work optionally without a global registry.

This pattern of exporting both package-level functionality that operates on a global instance and a type that can be instantiated for isolated use is very common in Go applications and in the Go stdlib. It allows for the use of functionality without global state and register patterns that can make an application difficult to understand over time.

We already use this package with global state in a few applications, and this PR doesn't change that, but it gives us the option to use it without global state in new applications, and in an application I am writing, and makes it possible for me to extract the general error rendering logic.

### Known limitations

N/A

cc @howardtw

### DIFF
--- a/services/horizon/internal/assertions_test.go
+++ b/services/horizon/internal/assertions_test.go
@@ -50,7 +50,7 @@ func (a *Assertions) Problem(body *bytes.Buffer, expected problem.P) bool {
 		return false
 	}
 
-	actual.Type = strings.TrimPrefix(actual.Type, problem.ServiceHost)
+	actual.Type = strings.TrimPrefix(actual.Type, problem.Default.ServiceHost())
 	if expected.Type != "" && a.Equal(expected.Type, actual.Type, "problem type didn't match") {
 		return false
 	}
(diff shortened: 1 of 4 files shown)


leighmcculloch:
If this change is accepted, I'd like to open a PR with the change below that extracts all of the error mapping logic out into a new package `support/render/errors` so that the `problem` package only contains the specifics of how to render a `P` problem error, but we can re-use all the other logic to render errors of any format. Here's the diff so you can see where this is going:
https://github.com/leighmcculloch/stellar--go/compare/makeproblemusablewithoutglobals...separateerrorsfromproblems

howardtw:
👍

Issue #2046, opened 2019-12-12
services/horizon: Data race during expingest shutdown


bartekn:
### What version are you using?

Horizon 0.24.0 RC

### What did you do?

1. Run Horizon with race detector (`-race`).
2. Send `SIGTERM` signal.

### What did you expect to see?

No data races.

### What did you see instead?

First occurrence:
```
INFO[2019-12-11T15:29:52.344+01:00] Reading new ledger                            ledger=27212116 pid=1997 service=expingest
INFO[2019-12-11T15:29:52.345+01:00] Processing ledger                             ledger=27212116 pid=1997 service=expingest type=ledger_pipeline updating_database=true
INFO[2019-12-11T15:29:55.556+01:00] Processed ledger                              ledger=27212116 pid=1997 service=expingest type=ledger_pipeline
INFO[2019-12-11T15:29:55.557+01:00] Finished processing ledger                    duration=3.212663749 ledger=27212116 pid=1997 service=expingest transactions=54
^CINFO[2019-12-11T15:29:56.283+01:00] received signal, gracefully stopping          pid=1997
INFO[2019-12-11T15:29:56.283+01:00] Shutting down ingestion system...             pid=1997 service=expingest
INFO[2019-12-11T15:29:57.267+01:00] finished background ticker                    pid=1997
INFO[2019-12-11T15:29:57.352+01:00] Reading new ledger                            ledger=27212117 pid=1997 service=expingest
INFO[2019-12-11T15:29:57.354+01:00] Processing ledger                             ledger=27212117 pid=1997 service=expingest type=ledger_pipeline updating_database=true
INFO[2019-12-11T15:29:57.355+01:00] Processing ledger shutdown                    duration=0.002914144 ledger=27212117 pid=1997 service=expingest
INFO[2019-12-11T15:29:57.356+01:00] Session shut down                             pid=1997 service=expingest
INFO[2019-12-11T15:29:57.358+01:00] stopped                                       pid=1997
WARNING: DATA RACE
Read at 0x00c0003996f8 by goroutine 87:
  github.com/stellar/go/support/db.(*Session).ExecRaw()
      /Users/bartek/gocode/src/github.com/stellar/go/support/db/session.go:362 +0xe1
  github.com/stellar/go/support/db.(*Session).Exec()
      /Users/bartek/gocode/src/github.com/stellar/go/support/db/session.go:170 +0x100
  github.com/stellar/go/services/horizon/internal/db2/history.(*Q).UpdateAccount()
[...]

PR #2056, opened 2019-12-13
Merge horizon v0.24.1 into master

tamirms:
Update master with commits from v0.24.1


bartekn:
LGTM, but let's merge it when Horizon 0.24.1 is released. It's possible that we'll have more commits (ex. I'm planning to fix #2046).

PR #2057, opened 2019-12-13
support/render/errors: extract error rendering logic from problems

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Extract the general error rendering logic from the `support/render/problem` package into `support/render/errors`.

### Why

The `problem` package contains fantastic logic for rendering errors to JSON error responses, but it bundles it with specific logic for rendering [RFC7807] JSON problem responses. I'm writing an application that returns error responses in the format required by our SEPs, specifically [SEP-10]. I'd like to make use of the really good pattern for rendering errors safely, but I can't use the problem JSON format.

[RFC7807]: https://tools.ietf.org/html/rfc7807
[SEP-10]: https://github.com/stellar/stellar-protocol/blob/dd3a45b5065acc3f6e6ab155b6186935304955bf/ecosystem/sep-0010.md

### Known limitations

There is some casting of errors to `P` and `*P` types and some type testing that doesn't seem to be very well tested. Some of the changes in this PR remove that code because I don't think it is required. I've added a test to try and ensure there is no regression here, but it's not clear what situation the original author was attempting to solve for.


support/render/errors/errors.go
```diff
@@ -0,0 +1,144 @@
...
+	"github.com/stellar/go/support/log"
+)
+
+// Renderable is a value that can be rendered as a HTTP JSON response.
+type E interface {
+	error
+	// StatusCode returns the HTTP status code that will be used in the HTTP
+	// response when this error is rendered.
+	StatusCode() int
+}
```
leighmcculloch: I'm a little concerned that this interface might be too easy to satisfy, and that any object that implements this interface will be rendered out of a server even if it maybe contains sensitive information.

I could add an additional method like `IsE() bool`. It's also a little gross to what is otherwise a simple interface.

Another option is when creating the `Errors` instance I could specify a type, and only values that assign to that type will be rendered out as they are. That's not very intuitive though.

Thoughts @tamirms?

leighmcculloch: I ended up doing this 29b49e2.

tamirms: maybe the `E` interface is unnecessary. Perhaps you can leave `Errors` so that it acts on `P` instances. Then you can create an interface which defines how `P` instances are supposed to be rendered using a `http.ResponseWriter`.

In summary, `Errors` accepts `P` instances and to construct an `Errors` instance you need to provide something which satisfies the `ProblemRenderer` interface:

```golang
type ProblemRenderer interface{
      Render(ctx context.Context, w http.ResponseWriter, p P)
}
```

leighmcculloch: That would mostly work too, but it would fix in place the data that an application can use for its errors, and the fields in a `P` and the `problem` package are specific to rendering problems in a specific format. I think the data and what is rendered go hand in hand.

leighmcculloch: I took a stab at what we DM'd about, adding the `Render` function to the `E`/`P` types, but that doesn't work with `P` because it needs access to the `Problem` type to get the `ServiceHost`.

leighmcculloch:
I think my plans for this were too grand. I think the mapping of errors to problems is cool, but unnecessary for small to moderate applications and I guess there's a reason that only our largest application, Horizon, is using the `problem` package's mapping features meaningfully.

For example, in the new app I'm building this is all the code I really need:

Somewhere in my app:
```go
var ServerError = errorResponse{
    Status:   http.StatusInternalServerError,
    ErrorStr: "An error occurred while processing this request.",
}       
var NotFound = errorResponse{
    Status:   http.StatusNotFound,
    ErrorStr: "The resource at the url requested was not found.",
}   
var BadRequest = errorResponse{
    Status:   http.StatusBadRequest,
    ErrorStr: "The request you sent was invalid in some way.",
}       
var Unauthorized = errorResponse{
    Status:   http.StatusUnauthorized,
    ErrorStr: "The request you sent could not be authenticated.",
}   
        
type errorResponse struct {
    Status   int    `json:"-"`
    ErrorStr string `json:"error"`
}
    
func (e errorResponse) Render(w http.ResponseWriter) {
    w.WriteHeader(e.Status)
    httpjson.Render(w, e, httpjson.JSON)
}
```

In a handler:
```go
err := validateRequest(...)
if err != nil {
    httpjson.Render(w, BadRequest, httpjson.JSON)
    return
}

..., err := doDatabaseThing(...)
if err != nil {
    log.Ctx(ctx).WithStack(err).Error(err)
    httpjson.Render(w, ServerError, httpjson.JSON)
    return
}
```

leighmcculloch:
With this PR closed the previous change #2054 was unnecessary, but given it changes the package to follow a common Go idiom I think I'll keep it.

Issue #2063, opened 2019-12-16
services/ticker: TestInsertOrUpdateAsset random failure


bartekn:
`TestInsertOrUpdateAsset` failed when testing a commit that was a markdown file edit: https://circleci.com/gh/stellar/go/12298

leighmcculloch:
This test used to flake more frequently and I thought I fixed it by replacing use of `Truncate` with `Round`. That was in this commit: 3e2307012937199681852493b1b679a93689dbf5. The fix wasn't perfect, Postgres and Go round time differently:
>The fix in this change isn't perfect according to discussion at lib/pq#227 (comment), where a commenter states that Postgres' rounding function rounds to nearest and *in the event of a tie rounds to even*, which is different to Go's `Round` function which rounds to nearest and *in the event of a tie rounds away from zero*.

I'll see if I can find a better fix.

leighmcculloch:
I've opened issue #2064 to add test logs to give us 👀 into what time inputs can be used to reproduce the issue. I won't attempt to fix the issue right now as the time cost to fix it without knowing what inputs it is failing on will be very high. Next time it fails with these test logs report here and I'll fix it.

_Hindsight: I wish I had added these test logs when I added the last fix._

leighmcculloch:
To my knowledge these tests haven't flaked since this issue was opened, or since logs were added in #2064. The logs are there to help us fix the issue if or when it happens again. I'll close this issue now. It can be reopened or a new issue created if this test flakes again.

bartekn:
Run into this again: https://app.circleci.com/pipelines/github/stellar/go/8281/workflows/262b90cf-5920-4a0d-9f1a-fbac765c442b/jobs/44938

bartekn:
Another instance: https://app.circleci.com/pipelines/github/stellar/go/9166/workflows/f31d23c8-2110-4bbc-a73b-3c39b2966e38/jobs/52124

Issue #1733, opened 2019-09-09
Intermittent failing test TestInsertOrUpdateAsset in services/ticker/internal/tickerdb


leighmcculloch:
The test `TestInsertOrUpdateAsset` in in the `services/ticker/internal/tickerdb` package has intermittently failed on CI. It appears to pass locally when executed a few hundred times.

Given that the failure is on time, it's possible `time.Now()` is playing a role and should be mocked within.

https://circleci.com/gh/stellar/go/7988

```
--- FAIL: TestInsertOrUpdateAsset (0.43s)
	Error Trace:	queries_asset_test.go:76
	Error:		Not equal: 
        	expected: time.Time{wall:0x4f27ac0, ext:63703668822, loc:(*time.Location)(0x12d22e0)}
        	received: time.Time{wall:0x501bd00, ext:63703668822, loc:(*time.Location)(0x12d22e0)}
        	
        	Diff:
        	--- Expected
        	+++ Actual
        	@@ -1,2 +1,2 @@
        	-(time.Time) 2019-09-09 23:33:42.083 +0000 UTC
        	+(time.Time) 2019-09-09 23:33:42.084 +0000 UTC

	Error Trace:	queries_asset_test.go:81
	Error:		Not equal: 
        	expected: time.Time{wall:0x4f27ac0, ext:63703668822, loc:(*time.Location)(0x12d22e0)}
        	received: time.Time{wall:0x501bd00, ext:63703668822, loc:(*time.Location)(0x12d22e0)}
        	
        	Diff:
        	--- Expected
        	+++ Actual
        	@@ -1,2 +1,2 @@
        	-(time.Time) 2019-09-09 23:33:42.083 +0000 UTC
        	+(time.Time) 2019-09-09 23:33:42.084 +0000 UTC

FAIL
```

leighmcculloch:
This is reproducible with certain `time.Now()`s. I managed to find a time that it reproduces with (see patch below). With this, and other specific times the time returned by postgres will be one millisecond ahead of what the code is expecting.

I believe this is related to lib/pq#227. This may not be fixable though since the `pq` passes time through to postgres and it gets truncated there.

[...]

PR #2064, opened 2019-12-16
services/ticker: add test logs to identify inputs that make test flaky

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add logs to the `TestInsertOrUpdateAsset` that print out the exact times that are being used.

### Why

The test has been reported as flaky a couple times, first in #1733 then in #2063. There was an imperfect fix put in place in 3e23070 but we've seen another test failure. Previously when I debugged this I wrote some test cases to try long series of inputs to find an input that caused a failure. This time I think it will be a better use of our time if we improve the test logs in this test so that the next time it fails we know exactly what inputs are being used for time and then we can address the continued flakyness.

For #2063 

### Known limitations

This doesn't fix the issue reported, just gives us more visibility into a set of inputs we can reproduce with so that when we spend time fixing it we can be confident if we have fixed it or not.

### DIFF
--- a/services/ticker/internal/tickerdb/queries_asset_test.go
+++ b/services/ticker/internal/tickerdb/queries_asset_test.go
@@ -53,6 +53,7 @@ func TestInsertOrUpdateAsset(t *testing.T) {
 
 	// Creating first asset:
 	firstTime := time.Now()
+	t.Log("firstTime:", firstTime)
 	a := Asset{
 		Code:          code,
 		IssuerAccount: issuerAccount,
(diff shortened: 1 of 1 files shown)


leighmcculloch:
`WithinDuration` would probably be a good way to go 👍. Either way, I'd like a set of inputs so that I know I have actually fixed the issue since it is flaky. I can use that function once it happens one more time.

Also lots of people have this issue interacting between Go and Postgres and assuming any cost to do so is low I'd like to see how a fix for this could be contributed back in a way that helps everyone.

PR #2065, opened 2019-12-16
txnbuild: remove redundant ok from internal verify functions

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Remove the `ok` bool value that the `verifyTxSignature` function returns alongside its `error` value.

### Why

The `bool` is redundant.

The `verifyTxSignature` function returns both a `bool` and an `error` but both are communicating the same result. The bool is only ever true when there is no error, and only ever false when there is an error. I'm refactoring some of this code and plan to use this function in a new function and it makes that reuse more difficult if this function returns bool and error.

In my new function I won't be passing on the bool to the caller, and so the question remains in my new function should I handle the bool and error separately so that if bool ever returns false without an error I generate some type of new error. I'd like to not have to solve this problem since it isn't important and instead remove the redundant bool.

### Known limitations

N/A

### DIFF
--- a/txnbuild/transaction.go
+++ b/txnbuild/transaction.go
@@ -477,32 +477,38 @@ func VerifyChallengeTx(challengeTx, serverAccountID, network string) (bool, erro
 	}
 
 	// verify signature from operation source
-	ok, err = verifyTxSignature(tx, op.SourceAccount.GetAccountID())
+	err = verifyTxSignature(tx, op.SourceAccount.GetAccountID())
 	if err != nil {
-		return ok, err
+		return false, err
 	}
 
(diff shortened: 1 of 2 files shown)


tamirms:
should the signature of `VerifyChallengeTx` change as well to remove the bool return value? It seems that the function never returns `false, nil`. `false` is only ever returned if there is an error. Having said that, `VerifyChallengeTx` is an exported function so changing the signature would be a breaking change.

leighmcculloch:
@ire-and-curses Yup, this change is internal only, no breaking changes.

PR #2071, opened 2019-12-17
txnbuild: update challenge tx helpers for SEP-10 v1.3.0

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add new challenge transactions helpers to `txnbuild` to support verifying a challenge transaction that has multiple signatures. Add three new functions:

- `ReadChallengeTx`: Read the details like the transaction and the client account ID out of the transaction.
- `VerifyChallengeTxSigners`: Verify that the signers of the transaction are a subset of a list of signers given to the function.
- `VerifyChallengeTxThreshold`: Verify that the signers of the transaction are a subset of a list of signers that meet a required threshold.

Deprecate `VerifyChallengeTx`.

### Why

The SEP-10 change that is open in stellar/stellar-protocol#489 clarifies how an implementer should verify that signers of the transaction are signers on the account and that accounts may have multiple signers.

An implementer needs to read out of the transaction the client account before verifying the transaction. For this reason we need two functions, one to read out details we need before verification and one to perform verification.

The read call also validates the server signature because no challenge transaction would ever be valid to read if it wasn't signed by the server.

### Known limitations

These helper functions take a decent amount of complexity away from an implementer, but they still leave a decent amount up to the implementer. An implementer would still need to make a call to Horizon, and do any weights to threshold comparison. I think it might be out-of-scope of the `txnbuild` package to offer that additional logic and SEP-10 technically leaves the exact details of what happens there up to the implementer so it would also be difficult to implementer in a library function.


txnbuild/transaction.go
```diff
@@ -413,77 +414,142 @@ func (tx *Transaction) SignWithKeyString(keys ...string) error {
...
 	if tx.SourceAccount == nil {
-		return false, errors.New("transaction requires a source account")
+		return tx, clientAccountID, errors.New("transaction requires a source account")
 	}
 	if tx.SourceAccount.GetAccountID() != serverAccountID {
-		return false, errors.New("transaction source account is not equal to server's account")
+		return tx, clientAccountID, errors.New("transaction source account is not equal to server's account")
 	}
 
 	//verify sequence number
```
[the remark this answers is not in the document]
leighmcculloch: 1bd551f5

txnbuild/transaction.go
```diff
@@ -413,77 +414,142 @@ func (tx *Transaction) SignWithKeyString(keys ...string) error {
...
+// verified if it is signed by the server account, and all other signatures
+// match a signer that has been provided as an argument. Additional signers can
+// be provided that do not have a signature, but all signatures must be matched
+// to a signer for verification to succeed. If verification succeeds a list of
+// signers that were found is returned, excluding the server account ID.
+//
+// Errors will be raised if:
+//  - The transaction is invalid according to ReadChallengeTx.
+//  - No client signatures are found on the transaction.
+//  - One or more signatures in the transaction are not identifiable as the server account or one of the signers provided in the arguments.
```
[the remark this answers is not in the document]
leighmcculloch: 1bd551f5

txnbuild/transaction.go
```diff
@@ -413,77 +414,142 @@ func (tx *Transaction) SignWithKeyString(keys ...string) error {
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Remove the server signer from the signers list if it is present. It's
+	// important when verifying signers of a challenge transaction that we only
+	// verify and return client signers. If an account has the server as a
+	// signer the server should not play a part in the authentication of the
+	// client.
+	clientSigners := make([]string, 0, len(signers)-1)
```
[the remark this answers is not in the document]
leighmcculloch: You're right. This is a carry over when I had written this code a different way. There is no functional impact because the slice is expanded to length required, but we shouldn't need to do that. 1bd551f5

leighmcculloch:
@msfeldstein provided some feedback to me directly that we need to be able to offer this as a similar level of drop-in as the prior functions that support SEP-10. In light of this I think I'm going to move the functions I'm adding out of `txnbuild` and into a new package, maybe called `webauth` or `challenge` or `txnchallenge`, and then explore how to offer verification using a configurable strategy, while providing some default strategies for popular use cases like anchors.

txnbuild/transaction.go
```diff
@@ -413,77 +414,186 @@ func (tx *Transaction) SignWithKeyString(keys ...string) error {
...
+// provided as an argument, and those signatures meet a threshold on the
+// account.
+//
+// Errors will be raised if:
+//  - The transaction is invalid according to ReadChallengeTx.
+//  - No client signatures are found on the transaction.
+//  - One or more signatures in the transaction are not identifiable as the
+//    server account or one of the signers provided in the arguments.
+//  - The signatures are all valid but do not meet the threshold.
+func VerifyChallengeTxThreshold(challengeTx, serverAccountID, network string, threshold Threshold, signers []Signer) (signersFound []Signer, err error) {
```
leighmcculloch: TODO:
- [ ] Add more tests for this function

leighmcculloch:
@msfeldstein @ire-and-curses Could you take a look at this? I've added a function `VerifyChallengeTxThreshold` that makes the process simpler. For an example of what this will look like in an application, see this PR that adds SEP-10 to the Go SEP-10 server: #1685

txnbuild/set_options.go
```diff
@@ -1,6 +1,7 @@
 package txnbuild
 
 import (
+	"github.com/stellar/go/clients/horizon"
```
ire-and-curses: As we discussed, I think this is a problem. `txnbuild` is a more basic library that is responsible for transactions. It doesn't know about `horizonclient`.

I do see the rationale as the similar but different nature of these signer structs is annoying.

One option would be to consider a third package that could hold helpers like this. Another reasonable compromise might be to put these helpers in horizonclient, as it already depends on txnbuild.

We should think again about a path forward to rationalise these structs.

txnbuild/transaction.go
```diff
@@ -413,77 +414,192 @@ func (tx *Transaction) SignWithKeyString(keys ...string) error {
 	return tx.Sign(signers...)
 }
 
-// VerifyChallengeTx is a factory method that verifies a SEP 10 challenge transaction,
-// for use in web authentication. It can be used by a server to verify that the challenge
-// has been signed by the client.
-// More details on SEP 10: https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0010.md
-func VerifyChallengeTx(challengeTx, serverAccountID, network string) (bool, error) {
```
ire-and-curses: This was a public method, so removing this is a breaking change. I suspect it is not widely used so I think I'm ok with it, but we should make it very clear in the release notes.

txnbuild/set_options.go
leighmcculloch: 👍 Thanks for pointing this out, I forgot we discussed this last year. I'll remove this dependency.

leighmcculloch: f20703f3

leighmcculloch: @ire-and-curses and I chatted about this offline and he said we need to keep horizon protocol out of the dependency list too. We agree the best way forward is to move these functions out to a new package `webauth`, but this is going to require sometime because it will require rewriting the existing verification code entirely.

The rewrite is necessary because the code is dependent on txnbuild, and the only way to move it to a new package and maintain backwards compatibility by having the function here call the new function is to remove that dependency back on txnbuild. This is required because otherwise a circular dependency would occur.

leighmcculloch: @ire-and-curses In b7bec9f and 731c0d4 I removed the horizon code and replaced it with no dependency on Horizon. I've changed the interface of the functions to accept a map of signers to weights and exported that same native data type in the horizon protocol package. We can pass data from one to the other without having an explicit dependency, and I think this is the best path forward. The data type I'm using is what we use in the XDR package for passing around a set of signers and weights (see [this](https://github.com/stellar/go/blob/2a7c16b0b50a953bbf8b14b11ea9b9e76c06fa43/xdr/account_entry.go#L3-L14)) Could you re-:eyes: and 👍 if you're happy?

txnbuild/transaction.go
```diff
@@ -413,77 +414,188 @@ func (tx *Transaction) SignWithKeyString(keys ...string) error {
...
+	err = verifyTxSignature(tx, serverAccountID)
+	if err != nil {
+		return tx, clientAccountID, err
+	}
+
+	return tx, clientAccountID, nil
+}
+
+// SignerSummary is a map of signers to their weights.
+type SignerSummary map[string]int32
```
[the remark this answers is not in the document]
leighmcculloch: I don't feel strongly and I'm not sure what the best place is in this case. Given it's isolated use I think here is okay. Given that we expect `Account` types to implement it, it might make sense to put it alongside the `Account` type or in its own file `signer_summary.go`.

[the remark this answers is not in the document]
leighmcculloch: I moved it. 0a9dbc88

leighmcculloch:
@ire-and-curses Sorry, I fixed one bug since you last 👀d that I don't think I could merge and fix separately and is worth a second set of eyes. See 99529e7.

leighmcculloch:
I need to add test cases for tests 6 and 7 listed here:
https://github.com/stellar/transfer-server-validator/pull/9#pullrequestreview-348298642

txnbuild/transaction.go
```diff
@@ -546,34 +546,63 @@ func VerifyChallengeTxSigners(challengeTx, serverAccountID, network string, sign
...
 		}
+		clientSignersSet[signer] = struct{}{}
+	}
+	clientSigners := make([]string, 0, len(clientSignersSet))
+	for signer := range clientSignersSet {
 		clientSigners = append(clientSigners, signer)
 	}
 
-	signersFound, err := verifyTxSignatures(tx, clientSigners...)
+	// Verify that all the transaction's signers (server and client) in one
```
ire-and-curses:
```suggestion
	// Verify all the transaction's signers (server and client) in one
```

leighmcculloch:
Sorry @ire-and-curses I'd pushed the restructure yesterday, but I hadn't finished cleaning up some things that showed up during it. I just pushed those changes. I think this is it unless you have any concerns.

txnbuild/transaction.go
```diff
@@ -413,77 +414,222 @@ func (tx *Transaction) SignWithKeyString(keys ...string) error {
...
+	for _, signer := range allSignersFound {
+		if signer == serverKP.Address() {
+			serverSignerFound = true
+			continue
+		}
+		signersFound = append(signersFound, signer)
+	}
+
+	// Confirm we matched a signature to the server signer.
+	if !serverSignerFound {
```
overcat: We have verified the server's signature in `ReadChallengeTx`, is it necessary to verify it again here?

PR #2072, opened 2019-12-17
exp/services/webauth: add SEP-10 v1.3.0 implementation

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Update the SEP-10 implementation to support the changes in stellar/stellar-protocol#489.

### Why

The SEP-10 reference implementation should be kept up-to-date with SEP-10 as it changes, and making these changes now ahead of the protocol change ensures we are validating what the change will require of other implementors.

### Known limitations

N/A


exp/services/webauth/internal/commands/serve.go
```diff
@@ -0,0 +1,96 @@
...
+			ConfigKey:   &opts.HorizonURL,
+			FlagDefault: horizonclient.DefaultTestNetClient.HorizonURL,
+			Required:    true,
+		},
+		{
+			Name:        "network-passphrase",
+			Usage:       "Network passphrase of the Stellar network transactions should be signed for",
+			OptType:     types.String,
+			ConfigKey:   &opts.NetworkPassphrase,
+			FlagDefault: network.PublicNetworkPassphrase,
```
ire-and-curses: Should this default to the test network passphrase, since the Horizon server is testnet by default?

exp/services/webauth/internal/commands/serve.go
```diff
@@ -0,0 +1,96 @@
...
+		{
+			Name:      "signing-key",
+			Usage:     "Stellar signing key used for signing transactions",
+			OptType:   types.String,
+			ConfigKey: &opts.SigningKey,
+			Required:  true,
+		},
+		{
+			Name:           "challenge-expires-in",
+			Usage:          "The time period that after which the challenge transaction expires",
```
ire-and-curses:
```suggestion
			Usage:          "The time period after which the challenge transaction expires",
```

exp/services/webauth/internal/jwtkey/jwtkey.go
```diff
@@ -0,0 +1,79 @@
+// Package jwtkey provide utility functions for generating, serializing and
```
ire-and-curses:
```suggestion
// Package jwtkey provides utility functions for generating, serializing and
```

exp/services/webauth/internal/commands/serve.go
leighmcculloch: It seems like some of our other apps like Ticker default to the public net. I don't feel strongly, but would be good to be consistent with our apps built on Horizon.

exp/services/webauth/internal/jwtkey/jwtkey.go
leighmcculloch: fa8e4659

exp/services/webauth/internal/commands/serve.go
leighmcculloch: fa8e4659

exp/services/webauth/internal/commands/serve.go
leighmcculloch: Oops, I see what you mean. I'm setting testnet in the URL and publicnet as the passphrase. Thanks! Fixed in 4f60806e6ef629fba75cb6791bc7812c83ffd52e.

exp/services/webauth/internal/serve/token.go
```diff
@@ -43,32 +43,22 @@ func (h tokenHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 		serverError.Render(w)
 		return
 	}
-	clientAccountID := tx.Operations[0].(*txnbuild.ManageData).SourceAccount.GetAccountID()
+	requiredThreshold := txnbuild.Threshold(clientAccount.Thresholds.HighThreshold)
+	clientSignerSummary := clientAccount.SignerSummary()
 
-	clientAccount, err := h.HorizonClient.AccountDetail(horizonclient.AccountRequest{AccountID: clientAccountID})
+	_, err = txnbuild.VerifyChallengeTxThreshold(req.Transaction, h.SigningAddress.Address(), h.NetworkPassphrase, requiredThreshold, clientSignerSummary)
 	if err != nil {
```
leighmcculloch: TODO: 
- [ ] Add handling for account doesn't exist.

leighmcculloch: I decided to do this in another PR.

exp/services/webauth/internal/serve/token.go
```diff
@@ -43,32 +43,22 @@ func (h tokenHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	_, err = txnbuild.VerifyChallengeTx(req.Transaction, h.SigningAddress.Address(), h.NetworkPassphrase)
+	_, clientAccountID, err := txnbuild.ReadChallengeTx(req.Transaction, h.SigningAddress.Address(), h.NetworkPassphrase)
 	if err != nil {
-		unauthorized.Render(w)
+		badRequest.Render(w)
```
howardtw: Sorry I come late and this might not be in the scope of this PR: it seems a bit odd to me that we are not looking at the content of `err` and do `badRequest.Render` (same for `serverError.Render`). It seems more natural to me that we return the corresponding error message based on what's in `err`.

exp/services/webauth/internal/serve/token_test.go
```diff
@@ -314,13 +314,35 @@ func TestToken_jsonInputValidMultipleSigners(t *testing.T) {
...
+		if _, ok := token.Method.(*jwt.SigningMethodECDSA); !ok {
+			return nil, fmt.Errorf("unexpected signing method: %v", token.Header["alg"])
+		}
+		return &jwtPrivateKey.PublicKey, nil
+	})
+	require.NoError(t, err)
+
+	claims := token.Claims.(jwt.MapClaims)
+	assert.Equal(t, serverKey.Address(), claims["iss"])
+	assert.Equal(t, account.Address(), claims["sub"])
```
howardtw: What's the difference between this line and the next line? 🤔

exp/services/webauth/internal/serve/token.go
leighmcculloch: That's an option, but in Go our ability to accurately align the many errors we get from a service with bad request, server error or unauthorized is pretty small. So attempting to do that is error prone and also a lot of effort to maintain since changes to code will change errors, and we need to keep a mapping of errors elsewhere disconnected.

So what I'm doing here is for each operation asking the question, "what are we doing here?" If the answer is reading or validating a request, then if it fails treat it like a bad request. If the answer is verifying a token or signature, then if it fails treat it like unauthorized. If something fails that shouldn't fail, then use server error.

exp/services/webauth/internal/serve/token_test.go
leighmcculloch: Nothing it would seem. Must be a duplicate line.

exp/services/webauth/internal/serve/token.go
howardtw: Not sure if you remember, but I've been wanting to implement something to substitute the root error for one you want, in this case `errors.Sub(err, BadRequest)` for example. This way, we can also preserve the error stack for better debugging, and it solves the problem you mentioned above.

leighmcculloch: >errors.Sub(err, BadRequest)

I'd love to hear more about this, I don't remember the details if we've already discussed it, sorry. Could you point me at something that would help me understand the idea?

>we can also preserve the error stack for better debugging

We can definitely add logging of errors to this app if needed. This function is a HTTP handler so there's nothing to return an error to, but I could add logging of an error. That would require identifying what errors are worth logging though. I don't think we'd want to log at this point on every bad request.

howardtw: I don't remember where we talked about it before, but let me rephrase it:
We usually return/display the desired error message based on the root error. As long as we have the root error mapped (in this case BadRequest), we can always substitute any root error for the one we want. With it, it will act like what you described "what are we doing here? If the answer is reading or validating a request, then if it fails treat it like a bad request.". In other words, `errors.Sub` will be essentially a function that manipulates the error stack. It will swap the root error out with something else and save the rest of the error stack.

PR #2074, opened 2019-12-18
exp/services/webauth: add SEP-10 v1.2.0 implementation

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add SEP-10 web authentication implementation based on SEP-10 v1.2.0 that requires the master key have a high threshold for authentication to succeed.

### Why

We need a standalone server implementation of SEP-10 for the mobile-wallet and this provides a server supporting the absolute basics of the existing SEP-10 protocol.

The SEP-10 protocol doesn't define what threshold a server should require a signing master key to have on an account, but for the sake of demonstration and our use case it requires the high threshold. It could be configurable but isn't at the moment.

This implementation has been written with the proposal in mind that we are making to SEP-10 (stellar/stellar-protocol#489) also, and already sets up the test cases with where we expect multi-sig to go but has those tests set with expectations that are appropriate given the limitations of SEP-10 today.

### Known limitations

This application is not polished which is why it is being added to the `exp` package and why this is a draft PR.


exp/services/webauth/README.md
```diff
@@ -0,0 +1,44 @@
+# webauth
+
+This is a SEP-10 Web Authentication implementation based on SEP-10 v1.2.0.
```
ire-and-curses: Link to SEP here would be good

ire-and-curses: Could also consider a sentence motivating the value, maybe something lifted from the SEP summary

exp/services/webauth/internal/serve/token.go
```diff
@@ -0,0 +1,93 @@
...
+		unauthorized.Render(w)
+		return
+	}
+
+	tx, err := txnbuild.TransactionFromXDR(req.Transaction)
+	if err != nil {
+		serverError.Render(w)
+		return
+	}
+	clientAccountID := tx.Operations[0].(*txnbuild.ManageData).SourceAccount.GetAccountID()
```
ire-and-curses: 1) Isn't it a concern that these are chained and could fail? What if the operation isn't a `ManageData`?
2) Do you want to get the source account from the operation, or from the TX itself? I don't remember the SEP but the TX is more usual.

exp/services/webauth/README.md
leighmcculloch: fa66f333

exp/services/webauth/internal/serve/token.go
leighmcculloch: The source of the transaction is the server key and the source of the manage data operation is the client key. Unfortunately the Go SDK provides no helpers to get at the account ID and in this implementation I'm avoiding changing the SDK.

In the proposed change #2071 I'm taking the opportunity to include a separate function that reads valuable information out like the client's account ID.

leighmcculloch: >Isn't it a concern that these are chained and could fail?

The Go SDK does a few checks as part of the verification call that occurs a few lines up. I could copy and paste those checks here but it'd be redundant.

ire-and-curses: cool, makes sense

exp/services/webauth/internal/serve/challenge_test.go
```diff
@@ -0,0 +1,96 @@
...
+	supportlog "github.com/stellar/go/support/log"
+	"github.com/stellar/go/xdr"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestChallenge(t *testing.T) {
+	serverKey, err := keypair.Random()
+	require.NoError(t, err)
+	account, err := keypair.Random()
```
howardtw: I think we missed checking this error?

exp/services/webauth/internal/serve/challenge_test.go
```diff
@@ -0,0 +1,95 @@
...
+	h.ServeHTTP(w, r)
+	resp := w.Result()
+
+	require.Equal(t, 200, resp.StatusCode)
+	assert.Equal(t, "application/json; charset=utf-8", resp.Header.Get("Content-Type"))
+
+	res := struct {
+		Transaction       string `json:"transaction"`
+		NetworkPassphrase string `json:"network_passphrase"`
+	}{}
```
howardtw: What's the reason that we are not doing `var res challengeResponse` here?

exp/services/webauth/internal/serve/challenge_test.go
```diff
@@ -0,0 +1,95 @@
...
+		SigningKey:         serverKey,
+		ChallengeExpiresIn: time.Minute,
+	}
+
+	r := httptest.NewRequest("GET", "/?account="+account.Address(), nil)
+	w := httptest.NewRecorder()
+	h.ServeHTTP(w, r)
+	resp := w.Result()
+
+	require.Equal(t, 200, resp.StatusCode)
```
howardtw: nit: we can use `http.StatusOK` instead of 200

exp/services/webauth/internal/serve/challenge_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+func TestChallengeNoAccount(t *testing.T) {
+	h := challengeHandler{}
+
+	r := httptest.NewRequest("GET", "/", nil)
+	w := httptest.NewRecorder()
+	h.ServeHTTP(w, r)
+	resp := w.Result()
+
+	require.Equal(t, 400, resp.StatusCode)
```
howardtw: nit: we can use `http.StatusBadRequest` instead of 400

exp/services/webauth/internal/serve/challenge_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+func TestChallengeInvalidAccount(t *testing.T) {
+	h := challengeHandler{}
+
+	r := httptest.NewRequest("GET", "/?account=GREATACCOUNT", nil)
+	w := httptest.NewRecorder()
+	h.ServeHTTP(w, r)
+	resp := w.Result()
+
+	require.Equal(t, 400, resp.StatusCode)
```
howardtw: nit: same as above

exp/services/webauth/internal/serve/errors_test.go
```diff
@@ -0,0 +1,32 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestErrorResponseRender(t *testing.T) {
+	w := httptest.NewRecorder()
+	serverError.Render(w)
+	resp := w.Result()
+	assert.Equal(t, 500, resp.StatusCode)
```
howardtw: nit: we can use `http.StatusInternalServerError`

exp/services/webauth/internal/serve/errors_test.go
```diff
@@ -0,0 +1,32 @@
...
+	assert.JSONEq(t, `{"error":"An error occurred while processing this request."}`, string(body))
+}
+
+func TestErrorHandler(t *testing.T) {
+	r := httptest.NewRequest("GET", "/404", nil)
+	w := httptest.NewRecorder()
+	handler := errorHandler{Error: notFound}
+	handler.ServeHTTP(w, r)
+	resp := w.Result()
+	assert.Equal(t, 404, resp.StatusCode)
```
howardtw: nit: we can use `http.StatusNotFound`

exp/services/webauth/internal/serve/serve.go
```diff
@@ -0,0 +1,81 @@
...
+	JWTPrivateKey      string
+	JWTExpiresIn       time.Duration
+}
+
+func Serve(opts Options) {
+	handler := handler(opts)
+	addr := fmt.Sprintf(":%d", opts.Port)
+	supporthttp.Run(supporthttp.Config{
+		ListenAddr: addr,
+		Handler:    handler,
```
howardtw: nit: can we inline this guy? `Hander: handler(opts)`

exp/services/webauth/internal/serve/token.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+func (h tokenHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+
+	req := tokenRequest{}
+
+	err := httpdecode.Decode(r, &req)
+	if err != nil {
+		badRequest.Render(w)
```
howardtw: I think we need a `return` after this line

exp/services/webauth/internal/serve/token_test.go
```diff
@@ -0,0 +1,427 @@
+package serve
```
howardtw: nit: likewise, it will be more readable if we can swap all the status codes of the responses with `http` status constants in the assertions.

exp/services/webauth/internal/serve/token_test.go
```diff
@@ -0,0 +1,427 @@
...
+	w := httptest.NewRecorder()
+	h.ServeHTTP(w, r)
+	resp := w.Result()
+
+	require.Equal(t, 200, resp.StatusCode)
+	assert.Equal(t, "application/json; charset=utf-8", resp.Header.Get("Content-Type"))
+
+	res := struct {
+		Token string `json:"token"`
+	}{}
```
howardtw: Likewise, maybe we can use `var res tokenResponse`?

exp/support/jwtkey/jwtkey.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/stellar/go/support/errors"
+)
+
+// GenerateKey is a convenience function for generating an ECDSA key for use as
+// a JWT key. It uses the P256 curve. To use other curves use the crypto/ecdsa
+// package directly.
+func GenerateKey() (*ecdsa.PrivateKey, error) {
+	k, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+	if err != nil {
+		return nil, err
```
howardtw: Do you think it'd be better to wrap this error as well?

exp/services/webauth/internal/commands/genjwtkey.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+func (c *GenJWTKeyCommand) Run() {
+	k, err := jwtkey.GenerateKey()
+	if err != nil {
+		c.Logger.Fatal(err)
+	}
+
+	if public, err := jwtkey.PublicKeyToString(&k.PublicKey); err == nil {
+		c.Logger.Print("Public:", public)
```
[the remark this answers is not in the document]
leighmcculloch: I was just changing the output of the string, it isn't important and I don't think it should block merging this.

exp/services/webauth/internal/serve/challenge_test.go
leighmcculloch: I'm writing API tests, it's ideal if they don't use code they're testing, they should test all the code behind the endpoint and it's inexpensive to use a type here.

exp/support/jwtkey/jwtkey.go
leighmcculloch: a1a8106e

exp/services/webauth/internal/serve/token.go
leighmcculloch: 6b7e851a

exp/services/webauth/internal/serve/serve.go
leighmcculloch: I ended up changing this code so that all the fatal calls happen in the top-level func instead of burying them in other funcs, so I don't think this is relevant now. I think that makes this code easier to see where the failures can occur. 8a115199

exp/services/webauth/internal/serve/challenge_test.go
leighmcculloch: 79c0fee3

exp/services/webauth/internal/serve/challenge_test.go
leighmcculloch: 79c0fee3

exp/services/webauth/internal/serve/challenge_test.go
leighmcculloch: 79c0fee3

exp/services/webauth/internal/serve/errors_test.go
leighmcculloch: 6dc1db5c

exp/services/webauth/internal/serve/errors_test.go
leighmcculloch: 6dc1db5c

exp/services/webauth/internal/serve/token_test.go
leighmcculloch: See https://github.com/stellar/go/pull/2074#discussion_r360429034.

exp/services/webauth/internal/serve/token_test.go
leighmcculloch: 79c0fee

exp/services/webauth/internal/commands/serve.go
```diff
@@ -70,11 +69,11 @@ func (c *ServeCommand) Command() *cobra.Command {
 		},
 		{
 			Name:           "jwt-expires-in",
-			Usage:          "The time period that after which the JWT expires",
+			Usage:          "The time period in seconds after which the JWT expires",
 			OptType:        types.Int,
 			CustomSetValue: config.SetDuration,
 			ConfigKey:      &opts.JWTExpiresIn,
-			FlagDefault:    int(300 * time.Minute),
+			FlagDefault:    300,
```
ire-and-curses: This default used to be a lot longer (300 minutes). Was it intended to change it to 5 minutes?

leighmcculloch: Ah thanks. Yeah I want it to be 300 seconds to keep it shorter by default. Shorter default is more secure and makes less assumptions and then a user can expand it if their use case makes it appropriate.

howardtw:
LGTM as well.

PR #2075, opened 2019-12-19
support/http/httpdecode: move http decoding to its own package

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Move the http decoding logic we use for decoding JSON messages to its own package.

### Why

This logic is at the moment buried as an not-exported function in the `support/render/httpjson` package. I think it belongs in its own package because:
- I'd like to use this function directly to decode some messages that do not use the handler that is exposed by the httpjson package.
- The `support/render/httpjson` is in the `render` package that implies it is primarily for rendering responses. To break this function out to call separately just for decoding feels a little out of place in render.
- I'd like to add functionality in another decode function to do both decoding of form urlencoded requests as well as JSON. @howardtw mentioned it's a little odd for that logic to live in httpjson, and so I plan to add a decode function alongside this one to do form urlencoded decoding, and then a third function that does both and forks based on the content type.

I've changed the inputs to this function to be a `http.Request` instead of a `io.Reader` because the other functions I'll add will use a request to fork based on the content-type and I think it is easier to understand if they are all consistent. I don't feel strongly about this last part though.

### Known limitations

N/A

### DIFF
--- a/support/http/httpdecode/httpdecode.go
+++ b/support/http/httpdecode/httpdecode.go
@@ -1,0 +1,18 @@
+package httpdecode
+
+import (
+	"encoding/json"
+	"net/http"
+)
+
+// Decode decodes JSON request from r into v.
+func DecodeJSON(r *http.Request, v interface{}) error {
+	dec := json.NewDecoder(r.Body)
+	dec.UseNumber()
+	err := dec.Decode(v)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
--- a/support/http/httpdecode/httpdecode.go
+++ b/support/http/httpdecode/httpdecode.go
@@ -5,7 +5,7 @@ import (
 	"net/http"
 )
 
-// Decode decodes JSON request from r into v.
+// DecodeJSON decodes JSON request from r into v.
 func DecodeJSON(r *http.Request, v interface{}) error {
 	dec := json.NewDecoder(r.Body)
 	dec.UseNumber()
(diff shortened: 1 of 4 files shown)


support/http/httpdecode/httpdecode.go
```diff
@@ -0,0 +1,18 @@
...
+import (
+	"encoding/json"
+	"net/http"
+)
+
+// DecodeJSON decodes JSON request from r into v.
+func DecodeJSON(r *http.Request, v interface{}) error {
+	dec := json.NewDecoder(r.Body)
+	dec.UseNumber()
+	err := dec.Decode(v)
```
howardtw: nit: how about just `return dec.Decode(v)`?

support/render/httpjson/handler.go
```diff
@@ -82,9 +83,9 @@ func (h *handler) executeFunc(ctx context.Context, req *http.Request) (interface
 	if h.inType != nil {
 		if h.readFromBody {
 			inPtr := reflect.New(h.inType)
-			err := read(req.Body, inPtr.Interface())
+			err := httpdecode.DecodeJSON(req, inPtr.Interface())
 			if err != nil {
-				return nil, err
+				return nil, ErrBadRequest
```
howardtw: Not in this PR, but I've been wanting to implement something to keep the error stack and substitute the root error so we can do something like `errors.Sub(ErrBadRequest, err)`

support/http/httpdecode/httpdecode.go
leighmcculloch: 5d7451d4

PR #2081, opened 2019-12-23
services/horizon/internal/expingest: Effects processor

### LINKED ISSUE #1981: Implement effects DB processors
- While working on this please make sure to use [`BatchInsertBuilder`](https://godoc.org/github.com/stellar/go/support/db#BatchInsertBuilder). Without it processing will be too slow for large ledger (we experience this in the old ingestion).
- Should be easy to do with [`LedgerTransaction.GetChanges`](https://godoc.org/github.com/stellar/go/exp/ingest/io#LedgerTransaction.GetChanges).
- The old ingestion system has this huge [`switch`](https://github.com/stellar/go/blob/master/services/horizon/internal/ingest/session.go#L151) checking operation type and creating effects based on this. Would be great to use pipeline to pass only transactions with operations a given processor is interested. So something like:
```
-> FilterTransactionsProcessor{OperationType: CreateAccount}
  -> CreateAccountEffectsProcessor{}
-> FilterTransactionsProcessor{OperationType: Payment}
  -> PaymentEffectsProcessor{}
```
OR, filter operations needed by a given effects group:
```
-> FilterTransactionsProcessor{OperationType: []{CreateAccount, Payment, PathPayment, ...}
  -> AccountEffectsProcessor{}
-> FilterTransactionsProcessor{OperationType: []{SetOptions}
  -> SignerEffectsProcessor{}
```
both ways will make the code easier to understand and test.

abuiles:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Implement effects DB processors - Fixes #1981 

### Why

This is part of the work to move ingestion to the new system. This will allow us to start processing operations events  and compare results from out implementation with the old ingestion system.


services/horizon/internal/db2/history/effect_batch_insert_builder.go
```diff
@@ -0,0 +1,54 @@
...
+	accountID int64,
+	operationID int64,
+	order uint32,
+	effectType EffectType,
+	details []byte,
+) error {
+	return i.builder.Row(map[string]interface{}{
+		"history_account_id":   accountID,
+		"history_operation_id": operationID,
+		"\"order\"":            order,
```
tamirms: why is order quoted?

services/horizon/internal/expingest/processors/effects_processor.go
```diff
@@ -0,0 +1,548 @@
...
+	order     uint32
+}
+
+func (e *effectsWrapper) add(address string, effectType history.EffectType, details map[string]interface{}) {
+	e.order++
+	e.effects = append(e.effects, map[string]interface{}{
+		"address":     address,
+		"operationID": e.operation.ID(),
+		"effectType":  effectType,
+		"order":       e.order,
```
tamirms: it seems like order is always `len(e.effects)+1`

services/horizon/internal/expingest/processors/effects_processor.go
```diff
@@ -0,0 +1,548 @@
...
+		"details":     details,
+	})
+}
+
+// Effects returns the operation effects
+func (operation *transactionOperationWrapper) Effects() (effects []map[string]interface{}, err error) {
```
tamirms: should this function be exported?

services/horizon/internal/db2/history/effect_batch_insert_builder.go
abuiles: because it is a reserved word, the query fails if we try to send it unquoted.

services/horizon/internal/expingest/processors/effects_processor.go
abuiles: good point! Also TIL that the len of a slice is cached.

services/horizon/internal/expingest/processors/effects_processor.go
abuiles: Good point, there is no reason for this to be exported.

services/horizon/internal/expingest/processors/effects_processor.go
```diff
@@ -0,0 +1,763 @@
...
+		var transaction io.LedgerTransaction
+		transaction, err = r.Read()
+		if err != nil {
+			if err == stdio.EOF {
+				break
+			} else {
+				return err
+			}
+		}
+
```
bartekn: I think we should a global processor if here:
```go
if !transaction.Successful() {
```
we have it in some methods but in general we will never ingest effects for failed txs so it's better to put it here instead of adding this in many functions (easier to forget).

services/horizon/internal/expingest/processors/effects_processor.go
```diff
@@ -0,0 +1,763 @@
...
+		e, err2 := operationsEffects(transaction, sequence)
+
+		if err2 != nil {
```
bartekn: Nit: Just curious: any reason for naming this `err2`?

services/horizon/internal/expingest/processors/effects_processor.go
```diff
@@ -0,0 +1,763 @@
...
+
+	op := operation.operation.Body.MustChangeTrustOp()
+	details := map[string]interface{}{"limit": amount.String(op.Limit)}
+
+	effect := history.EffectType(0)
+	assetDetails(details, op.Line, "")
+
+	changes := operation.transaction.GetOperationChanges(operation.index)
+
+	// TODO/ASK: since we know this is trustline change, how many changes can we find?
```
bartekn: Other than trust line itself, account subentry count can change when creating/removing a trust line. It's possible other side effect will be added to the protocol in the future (ex. removing offers over the limit). One thing I think it's worth changing is either `break` after first trust line found or call `effects.add` every time when trust line is found.

services/horizon/internal/expingest/processors/effects_processor_test.go
```diff
@@ -0,0 +1,940 @@
...
+}
+
+func (s *EffectsProcessorTestSuiteLedger) mockSuccessfulEffectBatchAdds() {
+	s.mockBatchInsertBuilder.On(
+		"Add",
+		s.addressToID[s.sortedAddresses[2]],
+		toid.New(int32(s.sequence), 1, 1).ToInt64(),
+		uint32(1),
+		history.EffectSequenceBumped,
+		[]byte{0x7b, 0x22, 0x6e, 0x65, 0x77, 0x5f, 0x73, 0x65, 0x71, 0x22, 0x3a, 0x33, 0x30, 0x30, 0x30, 0x30, 0x30, 0x30, 0x30, 0x30, 0x30, 0x30, 0x30, 0x7d},
```
bartekn: It's impossible to read to confirm it's mocked correctly. What about changing it to `[]byte("{...}")`?

services/horizon/internal/expingest/processors/effects_processor_test.go
```diff
@@ -0,0 +1,940 @@
...
+	s.mockQ.On("NewEffectBatchInsertBuilder", maxBatchSize).
+		Return(s.mockBatchInsertBuilder).Once()
+
+	s.mockBatchInsertBuilder.On(
+		"Add",
+		s.addressToID[s.sortedAddresses[2]],
+		toid.New(int32(s.sequence), 1, 1).ToInt64(),
+		uint32(1),
+		history.EffectSequenceBumped,
+		[]byte{0x7b, 0x22, 0x6e, 0x65, 0x77, 0x5f, 0x73, 0x65, 0x71, 0x22, 0x3a, 0x33, 0x30, 0x30, 0x30, 0x30, 0x30, 0x30, 0x30, 0x30, 0x30, 0x30, 0x30, 0x7d},
```
bartekn: Change to `[]byte("string")`.

exp/ingest/io/ledger_transaction_test.go
```diff
@@ -87,6 +87,26 @@ func TestFeeAndMetaChangesSeparate(t *testing.T) {
...
+		Result: xdr.TransactionResultPair{
+			Result: xdr.TransactionResult{
+				Result: xdr.TransactionResultResult{
+					Code: xdr.TransactionResultCodeTxFailed,
+				},
+			},
+		},
+	}
+
+	feeChanges := tx.GetOperationChanges(0)
```
tamirms: `operationChanges` instead of `feeChanges`

services/horizon/internal/db2/history/effect.go
```diff
@@ -216,7 +216,104 @@ func (q *EffectsQ) orderBookFilter(a xdr.Asset, prefix string) {
...
+		).
+		Where("heff.history_operation_id >= ? AND heff.history_operation_id <= ? ", from, to).
+		OrderBy(
+			"heff.history_operation_id asc, heff.order asc",
+		)
+
+	err := q.Select(&effects, sql)
+
+	if err != nil {
+		return effects, errors.Errorf(
```
tamirms: errors.Wrap

services/horizon/internal/expingest/processors/effects_processor_test.go
```diff
@@ -0,0 +1,940 @@
...
+			resultXDR:     "AAAAAAAAAGT/////AAAAAQAAAAAAAAAD////+QAAAAA=",
+			metaXDR:       "AAAAAQAAAAIAAAADABDLGAAAAAAAAAAA8Kr+J6EPYBImiuqVPIS3S6yfZYbfIXjD3EGSB3iK27oAAAB2ucIg2AAMpboAADNFAAAA4wAAAAAAAAAAAAAAAAEAAAAAAAAAAAAAAQAABHT9ws4fAAAAAAAAAAAAAAAAAAAAAAAAAAEAEMsYAAAAAAAAAADwqv4noQ9gEiaK6pU8hLdLrJ9lht8heMPcQZIHeIrbugAAAHa5wiDYAAylugAAM0YAAADjAAAAAAAAAAAAAAAAAQAAAAAAAAAAAAABAAAEdP3Czh8AAAAAAAAAAAAAAAAAAAAAAAAAAA==",
+			feeChangesXDR: "AAAAAgAAAAMAEMsCAAAAAAAAAADwqv4noQ9gEiaK6pU8hLdLrJ9lht8heMPcQZIHeIrbugAAAHa5wiE8AAylugAAM0UAAADjAAAAAAAAAAAAAAAAAQAAAAAAAAAAAAABAAAEdP3Czh8AAAAAAAAAAAAAAAAAAAAAAAAAAQAQyxgAAAAAAAAAAPCq/iehD2ASJorqlTyEt0usn2WG3yF4w9xBkgd4itu6AAAAdrnCINgADKW6AAAzRQAAAOMAAAAAAAAAAAAAAAABAAAAAAAAAAAAAAEAAAR0/cLOHwAAAAAAAAAAAAAAAAAAAAA=",
+			hash:          "24206737a02f7f855c46e367418e38c223f897792c76bbfb948e1b0dbd695f8b",
+			index:         0,
+			sequence:      58,
+			expected:      []effect{},
+		},
+	}
```
bartekn: Effects tests look great but it's impossible to check if the assertions are OK without checking base64 values in XDR viewer. Would be cool to have XDR structs here. Can be done in a separate PR sometime in a future.

services/horizon/internal/db2/history/effect.go
```diff
@@ -216,7 +216,104 @@ func (q *EffectsQ) orderBookFilter(a xdr.Asset, prefix string) {
...
+// the given ledger matches the same effects as in history_effects
+func (q *Q) CheckExpOperationEffects(seq int32) (bool, error) {
+	expEffects, err := q.findOperationEffects(
+		"exp_history_effects",
+		"exp_history_accounts",
+		seq,
+	)
+
+	if err != nil {
+		return false, errors.Errorf(
```
tamirms: errors.Wrap

services/horizon/internal/db2/history/effect.go
```diff
@@ -216,7 +216,104 @@ func (q *EffectsQ) orderBookFilter(a xdr.Asset, prefix string) {
...
+	}
+
+	effects, err := q.findOperationEffects(
+		"history_effects",
+		"history_accounts",
+		seq,
+	)
+
+	if err != nil {
+		return false, errors.Errorf(
```
tamirms: errors.Wrap

services/horizon/internal/db2/history/effect_batch_insert_builder_test.go
```diff
@@ -0,0 +1,57 @@
...
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/stellar/go/services/horizon/internal/test"
+	"github.com/stellar/go/services/horizon/internal/toid"
+)
+
+// TODO: move to history/effects.go
```
tamirms: TODO

services/horizon/internal/expingest/processors/effects_processor.go
```diff
@@ -0,0 +1,763 @@
...
+type EffectProcessor struct {
+	EffectsQ history.QEffects
+}
+
+func (p *EffectProcessor) loadAccountIDs(accountSet map[string]int64) error {
+	addresses := make([]string, 0, len(accountSet))
+	for address := range accountSet {
+		addresses = append(addresses, address)
+	}
+	sort.Strings(addresses)
```
tamirms: the sort used here and in the participants processor is used so that there is a deterministic order we can test against in the mocks. but it turns out we can ignore order in the mocks (see https://github.com/stellar/go/pull/2092/files ). let's get rid of the sort and setup the mocks to compare against addresses without a specified order

services/horizon/internal/expingest/processors/effects_processor.go
```diff
@@ -0,0 +1,763 @@
...
+		accountSet[address] = id
+	}
+
+	return nil
+}
+
+func operationsEffects(transaction io.LedgerTransaction, sequence uint32) ([]effect, error) {
+	effects := []effect{}
+
+	if !transaction.Successful() {
```
tamirms: since there is a filter in the parent processor which ensures only successful transactions are processed, it means that the EffectProcessor should never receive unsuccessful transactions

tamirms: actually, please ignore this comment. it is possible for the processor to receive failed transactions

services/horizon/internal/db2/history/effect.go
```diff
@@ -216,7 +216,104 @@ func (q *EffectsQ) orderBookFilter(a xdr.Asset, prefix string) {
...
+			"could not load history_effects for ledger: %v",
+			seq,
+		)
+	}
+
+	if len(expEffects) == 0 || len(effects) == 0 {
```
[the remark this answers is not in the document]
abuiles: @bartekn we want to compare only if we have data in both old and new ingestion system, otherwise skip it. This is because we might have delays between the new and old system. Now if there is data in both we move forward with the other checks which is len and data equality

exp/ingest/io/ledger_transaction.go
```diff
@@ -189,6 +189,32 @@ func (t *LedgerTransaction) GetChanges() []Change {
 	return changes
 }
 
+// GetOperationChanges returns a developer friendly representation of LedgerEntryChanges.
+// It contains only operation changes.
+func (t *LedgerTransaction) GetOperationChanges(operationIndex uint32) []Change {
```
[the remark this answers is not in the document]
abuiles: Do you mean calling `GetOperationChanges` from `GetChanges`?   If that's what you mean, I'd prefer to keep `GetChanges` as it is and leave this one as an accessor to the changes for a single operation. 

I thought we could avoid the call to `t.Meta.V` in `GetChanges` by using this, but after close inspection I don't think that would be the case, we'll need to to call `t.Meta.V` in both scenarios:  In `GetChanges` to decide if we need to include `getChangesFromLedgerEntryChanges(v1Meta.TxChanges)`  and in `GetOperationChanges` to decide if we need `t.Meta.MustOperations()` or `v1Meta.Operations`

exp/ingest/io/ledger_transaction.go
```diff
@@ -189,6 +189,37 @@ func (t *LedgerTransaction) GetChanges() []Change {
...
+	var changes []Change
+
+	// If the transaction was not successful there are no changes
+	if !t.Successful() {
+		return []Change{}
+	}
```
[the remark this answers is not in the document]
abuiles: @bartekn addressed in 3a04386

services/horizon/internal/db2/history/effect.go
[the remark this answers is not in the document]
abuiles: Added a comment to explain this

services/horizon/internal/expingest/processors/effects_processor.go
abuiles: Good point, done. 

I keep the check in `func (operation *transactionOperationWrapper) effects()` since `transactionOperationWrapper` can be used by other processors.

services/horizon/internal/expingest/processors/effects_processor_test.go
abuiles: makes sense! will do

services/horizon/internal/expingest/processors/effects_processor_test.go
abuiles: 👍 sounds good! I decided to go this route since we had known results for each of the transactions used here.  I ca do that update once we ship this.

services/horizon/internal/expingest/processors/effects_processor.go
abuiles: addressed, let's break.

services/horizon/internal/expingest/processors/effects_processor.go
abuiles: fixed

abuiles:
@tamirms @bartekn I addressed all your comments, thanks!

services/horizon/internal/expingest/processors/effects_processor.go
```diff
@@ -0,0 +1,759 @@
...
+		}
+
+		if err = p.insertDBOperationsEffects(effects, accountSet); err != nil {
+			return err
+		}
+	}
+
+	// use an older lookup sequence because the experimental ingestion system and the
+	// legacy ingestion system might not be in sync
+	checkSequence := int32(sequence - 10)
```
tamirms: we need to only perform this check if `sequence > 10 `

services/horizon/internal/expingest/processors/effects_processor.go
```diff
@@ -0,0 +1,759 @@
...
+		if err != nil {
+			if err == stdio.EOF {
+				break
+			} else {
+				return err
+			}
+		}
+
+		if transaction.Successful() {
+			e, err2 := operationsEffects(transaction, sequence)
```
tamirms:
```golang
var effectsForTx []effect
effectsForTx, err = operationsEffects(transaction, sequence)
if err != nil {
	return err
}
effects = append(effects, effectsForTx...)
```

services/horizon/internal/expingest/processors/effects_processor.go
abuiles: cool! I was just reviewing the verify-range and notice we'll need this, updating it.

exp/ingest/io/ledger_transaction.go
```diff
@@ -189,6 +189,45 @@ func (t *LedgerTransaction) GetChanges() []Change {
...
+		// the operation failed it might not contain operations in meta.
+		if len(ops) == 0 || int(operationIndex) >= len(ops) {
+			return []Change{}
+		}
+
+		operationMeta := ops[operationIndex]
+		changes = getChangesFromLedgerEntryChanges(
+			operationMeta.Changes,
+		)
+	case 1:
```
abuiles: update coming for v2

abuiles: fixed  via cfe5f8d631439636f49ada01839856dfbefe88a0

services/horizon/internal/expingest/processors/effects_processor.go
abuiles: fixed

exp/ingest/io/ledger_transaction.go
```diff
@@ -212,36 +212,31 @@ func (t *LedgerTransaction) GetOperationChanges(operationIndex uint32) []Change
...
-		// the operation failed it might not contain operations in meta.
-		if len(ops) == 0 || int(operationIndex) >= len(ops) {
-			return []Change{}
-		}
-
-		operationMeta := ops[operationIndex]
-		changes = getChangesFromLedgerEntryChanges(
-			operationMeta.Changes,
-		)
+		changes = operationChanges(ops, operationIndex)
```
bartekn: Please remove V0 support as in: #2101.

abuiles: @bartekn done -- I rebase off your branch bartekn:remove-support-txmeta-v0 -- once this one is approved and #2101  merged, we'll need a last rebase and we'll be good to go.

PR #2086, opened 2020-01-07
services/horizon: verify-range command

### LINKED ISSUE #1848: Create a script that ingests transactions from the first ledger
Since we have all ledger entries in Horizon ingestion pipeline (after #1819 is merged) we can run the pipeline starting at ledger 1. This will allow us to find corner cases and bugs that appeared due to processing many transactions from the last couple years. We can then use the script to check pipeline if it changes in the future.

### LINKED ISSUE #2087: Update ingestion sessions to ingest state at ledger 1
stellar-core empty state for a new network contains a single master account. It's not part of the history archives. We need to update `exp/ingest` session to be able to ingest state at ledger 1.

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Adds `horizon expingest verify-range` command and `exp/ingest.RangeSession`. It also adds `GenesisLedgerStateReader` to that streams master account in a given network to allow ingesting ledger entry changes from ledger 1.

Close #1848.
Close #2087.

### Why

In order to ensure that new ingestion system works correctly we plan to reingest pubnet history from the first ledger. In order to achieve this we add `horizon expingest verify-range` command that accepts required two params: `--from` and `--to` that describe the ledger range. When started it:
* It ingests ledger entries of `from` ledger (must be a checkpoint ledger).
* Ingests ledgers `[from+1, to]`.
* If `--verify-state` is set, it runs state verifier on `to` ledger  (must be a checkpoint ledger).

When merged we can start multiple workers ingesting full history in parallel like:
```
horizon expingest verify-range --from 9983 --to 16383 --verify-state
horizon expingest verify-range --from 16383 --to 22783 --verify-state
...
```

Internally, it's using a new `RangeSession` from `exp/ingest` package that process a given range of ledger. `RangeSession` will be also used in `horizon reingest` commands.

It's also helpful for finding performance issues as it's easy to reingest a given range with a debug server enabled.

```
runs ingestion pipeline between X and Y sequence number (inclusive)

Usage:
  horizon expingest verify-range [flags]

Flags:
      --debug-server-port uint32   [optional] opens a net/http/pprof server at given port
      --from uint32                first ledger of the range to ingest
      --to uint32                  last ledger of the range to ingest
      --verify-state               [optional] verifies state at the last ledger of the range when true
```

### Known limitations

`LiveSession` and `RangeSession` have a shared code that can be extracted to a shared function. Should be done in a separate PR to not decrease readability of this PR.

### DIFF
--- a/exp/ingest/io/genesis_ledger_state_reader.go
+++ b/exp/ingest/io/genesis_ledger_state_reader.go
@@ -1,0 +1,64 @@
+package io
+
+import (
+	"io"
+	"sync"
+
+	"github.com/stellar/go/amount"
+	"github.com/stellar/go/keypair"
+	"github.com/stellar/go/xdr"
+)
+
+// GenesisLedgerStateReader is a streaming ledger entries for genesis ledger
+// (ledgerseq = 1) for of the network with the given passphrase.
+type GenesisLedgerStateReader struct {
+	NetworkPassphrase string
+
+	mutex sync.Mutex
+	done  bool
+}
+
+// Ensure GenesisLedgerStateReader implements StateReader
+var _ StateReader = &GenesisLedgerStateReader{}
+
+// GetSequence returns the sequence of the ledger.
+func (r *GenesisLedgerStateReader) GetSequence() uint32 {
+	return 1
+}
+
+// Read returns a new ledger entry change on each call, returning io.EOF when the stream ends.
+func (r *GenesisLedgerStateReader) Read() (xdr.LedgerEntryChange, error) {
+	r.mutex.Lock()
+	defer r.mutex.Unlock()
+
+	if r.done {
+		return xdr.LedgerEntryChange{}, io.EOF
+	}
+
+	masterKeyPair := keypair.Master(r.NetworkPassphrase)
+
+	masterAccountEntry := xdr.LedgerEntry{
+		LastModifiedLedgerSeq: 1,
+		Data: xdr.LedgerEntryData{
+			Type: xdr.LedgerEntryTypeAccount,
+			Account: &xdr.AccountEntry{
+				AccountId: xdr.MustAddress(masterKeyPair.Address()),
+				// 100B
+				Balance:    amount.MustParse("100000000000"),
+				SeqNum:     0,
+				Thresholds: xdr.Thresholds{1, 0, 0, 0},
+			},
+		},
+	}
+
+	r.done = true
+	return xdr.LedgerEntryChange{
+		Type:  xdr.LedgerEntryChangeTypeLedgerEntryState,
+		State: &masterAccountEntry,
+	}, nil
+}
+
+// Close should be called when reading is finished.
+func (r *GenesisLedgerStateReader) Close() error {
+	return nil
+}
--- a/services/horizon/internal/expingest/main.go
+++ b/services/horizon/internal/expingest/main.go
@@ -99,17 +99,25 @@ const (
 	loadOffersIntoMemoryState systemState = "loadOffersIntoMemory"
 	buildStateAndResumeState  systemState = "buildStateAndResume"
 	resumeState               systemState = "resume"
+	verifyRangeState          systemState = "verifyRange"
 	shutdownState             systemState = "shutdown"
 )
 
 type state struct {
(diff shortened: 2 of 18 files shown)


bartekn:
Added `GenesisLedgerStateReader` in https://github.com/stellar/go/pull/2086/commits/4e73f2181cf633fcb1f51274f5c22e75f6104860 to fix #2087.

services/horizon/internal/expingest/main.go
```diff
@@ -223,6 +250,23 @@ func NewSystem(config Config) (*System, error) {
 //   * If instances is a NOT leader, it runs ledger pipeline without updating a
 //     a database so order book graph is updated but database is not overwritten.
 func (s *System) Run() {
+	s.state = state{systemState: initState}
+	s.run()
+}
+
+// VerifyRange runs the ingestion pipeline on the range of ledgers. When
+// verifyState is true it verifies the state when ingestion is complete.
+func (s *System) VerifyRange(fromLedger, toLedger uint32, verifyState bool) error {
```
tamirms: I don't think `VerifyRange` is the best name because it implies that there will always be some verification process. my suggestion is to rename the function as `RunOnRange()`

tamirms:
Currently, `historyQ.RemoveExpIngestHistory(ledgerSeq)` is called before executing a state ingestion pipeline to ensure that we don't encounter uniqueness constraint errors when inserting rows in the historical ingestion tables. See https://github.com/stellar/go/pull/2055 for more context.

I'm not sure we want to maintain this behavior if we're going to introduce range sessions. If the horizon db contains ingestion data up to ledger `N` and we run `horizon expingest verify-range` using a range of `[N-1024, N-256]` that will have a side effect of removing ingestion data in the horizon db for ledgers beyond `N-256`.

Before executing the range request, should we first check if there will be data loss as a side effect? We could add a `force` flag to skip the check.

bartekn:
Answering both comments here. `RangeSession` will be used in three places:
* `verify-range` - this PR. The reason it's called _verify_ is that it's primary purpose is to verify ingestion pipeline on a given range. It ingests both history and state and there's an option to start a debug server and run state verifier on the last ledger. This command will be used in https://github.com/stellar/go/issues/1848.
* `reingest-range` - this will work exactly like the current `horizon db reingest range` command of the previous ingest package: it will ingest history data only (without state) and will be used to ingest old ledger (or full history) if Horizon admin needs it. https://github.com/stellar/go/issues/2089
* Upgrading from the previous ingestion system - it's possible that the last ledger ingested by the old system is not a check point ledger. In such case we need to catchup history ingestion to a checkpoint ledger and then ingest state and continue. This and `RemoveExpIngestHistory` issue you mentioned will be solved in: https://github.com/stellar/go/issues/2088.

tamirms:
@bartekn once #2088 is implemented, what are the potential side effects of running `verify-range`?

IMO, the name `verify-range` gives the impression that there should be no side effects including data loss when running the command. My expectation would be that `verify-range` performs some a set of stateless checks.

bartekn:
`verify-range` should be only run on an empty DB and there's a simple check to ensure this: https://github.com/stellar/go/pull/2086/files#diff-3b66fe1a6ca34a1d7747791e3054a199R506. It's a tool for developers and shouldn't be used by users. I can update the command description to make it more clear. Or maybe change the name to `verify-ingestion`?

tamirms:
@bartekn sorry, I overlooked the empty DB check when I was reviewing the PR. I also see that "requires clean DB" is included in the command description. that makes the semantics of the command clearer to me.  👍

exp/ingest/io/genesis_ledger_state_reader.go
```diff
@@ -0,0 +1,64 @@
...
+	"io"
+	"sync"
+
+	"github.com/stellar/go/amount"
+	"github.com/stellar/go/keypair"
+	"github.com/stellar/go/xdr"
+)
+
+// GenesisLedgerStateReader is a streaming ledger entries for genesis ledger
+// (ledgerseq = 1) for of the network with the given passphrase.
```
abuiles: typo

PR #2094, opened 2020-01-09
services/horizon/internal/db2: Add exp_history_trades table

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add a migration for the `exp_history_trades` table. Also add functions to insert and query the table. These functions will be used by the trades ingestion processor.

### Why

The trades processor in the experimental ingestion system will populate a new history trades table (following the same pattern we've used ledgers, transactions, and operations). The trades processor will use `TradeBatchInsertBuilder` to insert rows in the `exp_history_trades` table. The processor will use `CheckExpTrades()` to compare rows in `exp_history_trades` against rows in `history_trades` (populated by the legacy ingestion system).

### Known limitations

[N/A]

### DIFF
--- a/services/horizon/internal/db2/history/trade_batch_insert_builder.go
+++ b/services/horizon/internal/db2/history/trade_batch_insert_builder.go
@@ -108,7 +108,7 @@ func (i *tradeBatchInsertBuilder) Add(entries ...InsertTrade) error {
 			"price_d":              entry.SellPrice.D,
 		})
 		if err != nil {
-			return errors.Wrap(err, "failed to exec sql")
+			return errors.Wrap(err, "failed to add trade")
 		}
 	}
 
--- a/services/horizon/internal/db2/history/trade_batch_insert_builder.go
+++ b/services/horizon/internal/db2/history/trade_batch_insert_builder.go
@@ -1,0 +1,116 @@
+package history
+
+import (
+	"time"
+
+	"github.com/stellar/go/support/db"
+	"github.com/stellar/go/support/errors"
+	"github.com/stellar/go/xdr"
+)
+
+// InsertTrade represents the arguments to TradeBatchInsertBuilder.Add() which is used to insert
+// rows into the exp_history_trades table
+type InsertTrade struct {
+	HistoryOperationID int64
+	Order              int32
+	LedgerCloseTime    time.Time
+	BuyOfferExists     bool
+	BuyOfferID         int64
+	SellerAccountID    int64
+	BuyerAccountID     int64
+	SoldAssetID        int64
+	BoughtAssetID      int64
+	Trade              xdr.ClaimOfferAtom
+	SellPrice          xdr.Price
+}
+
+// TradeBatchInsertBuilder is used to insert trades into the
+// exp_history_trades table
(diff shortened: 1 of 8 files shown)


services/horizon/internal/db2/history/trade_batch_insert_builder.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+}
+
+// Exec flushes all outstanding trades to the database
+func (i *tradeBatchInsertBuilder) Exec() error {
+	return i.builder.Exec()
+}
+
+// Add adds a new trade to the batch
+func (i *tradeBatchInsertBuilder) Add(entries ...InsertTrade) error {
```
abuiles: What about moving the ingestion logic below to the the process to keep this `Add` as simple as possible?

services/horizon/internal/db2/history/trade_batch_insert_builder.go
```diff
@@ -0,0 +1,116 @@
...
+			"counter_offer_id":     counterOfferID,
+			"counter_account_id":   counterAccountID,
+			"counter_asset_id":     counterAssetID,
+			"counter_amount":       counterAmount,
+			"base_is_seller":       orderPreserved,
+			"price_n":              entry.SellPrice.N,
+			"price_d":              entry.SellPrice.D,
+		})
+		if err != nil {
+			return errors.Wrap(err, "failed to exec sql")
```
abuiles: > "failed to exec sql"
Should this be failed to add row?  since we are not executing yet

services/horizon/internal/db2/history/trade_test.go
```diff
@@ -69,3 +72,348 @@ func TestTradeQueries(t *testing.T) {
 	tt.Assert.Equal(xdr.Int64(2000000000), trades[0].CounterAmount)
 	tt.Assert.Equal(false, trades[0].BaseIsSeller)
 }
+
+type tradeRow struct {
```
abuiles: @tamirms would is make sense to add this as an exported struct to history? And future replacement for the current `history.Trade`

services/horizon/internal/db2/history/trade_batch_insert_builder.go
tamirms: this code mimics `InsertTrade()` which is defined in the same package and it uses the same helper functions (`EncodeOfferId()` and `getCanonicalAssetOrder()`) as `InsertTrade()`. If we were to move logic from this function out of this package it would require we modify `InsertTrade()` or have some code duplication because we'd need to copy the helper functions from the history package. These are the downsides to moving the logic out of the package.

I don't really see any upsides to moving the logic out of the package. If we make `Add()` simpler we do that by moving the complexity elsewhere. But the overall complexity remains the same. Currently, the complexity is isolated to one function and that is good. I don't see how moving the logic out would improve encapsulation.

services/horizon/internal/db2/history/trade_test.go
tamirms: this struct is actually unnecessary so I have removed it

PR #2114, opened 2020-01-14
Sort signers before creating event.

abuiles:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Save effect details order by address when ingesting set_option operations.

### Why

We got some failures when running `CheckExpOperationEffects`, however, after analyzing the data, we found out that  we have the desired result but the order was swapped:

```
horizon_prod=# select * from history_effects where history_operation_id=119137419218964481 ORDER BY "order";
select * from history_effects where history_operation_id=119137419218964481 ORDER BY "order";
 history_account_id | history_operation_id | order | type |                                         details
--------------------+----------------------+-------+------+-----------------------------------------------------------------------------------------
             192294 |   119137419218964481 |     1 |    4 | {"high_threshold": 0}
             192294 |   119137419218964481 |     2 |   12 | {"weight": 1, "public_key": "GDWTQTPLJRSPHEZQTXD7XBZM4URTR52Q7QYPQWWI35KXBT67Q4A7J3UQ"}
             192294 |   119137419218964481 |     3 |   11 | {"public_key": "GC7BWB2ME4LII3TVWTHUIT7KGJXU4D5M6JUNLQ57WA7JERDNSAEXLOAN"}

```
vs 

```
select * from exp_history_effects where history_operation_id=119137419218964481 ORDER BY "order";
 history_account_id | history_operation_id | order | type |                                         details
--------------------+----------------------+-------+------+-----------------------------------------------------------------------------------------
             198270 |   119137419218964481 |     1 |    4 | {"high_threshold": 0}
             198270 |   119137419218964481 |     2 |   11 | {"public_key": "GC7BWB2ME4LII3TVWTHUIT7KGJXU4D5M6JUNLQ57WA7JERDNSAEXLOAN"}
             198270 |   119137419218964481 |     3 |   12 | {"weight": 1, "public_key": "GDWTQTPLJRSPHEZQTXD7XBZM4URTR52Q7QYPQWWI35KXBT67Q4A7J3UQ"}
```

The reason for this is that doing a `for` in the result from `SignerSummary` can yield a different order each time because we are iterating over a map.

To fix this mismatch,  we are going to sort the addresses before saving them, allowing us to store the information in the same order in both the old and new ingestion system.

### Known limitations

N/A

### DIFF
--- a/services/horizon/internal/ingest/session.go
+++ b/services/horizon/internal/ingest/session.go
@@ -505,13 +505,13 @@ func (is *Session) ingestSignerEffects(effects *EffectIngestion, op xdr.SetOptio
 		return
 	}
 
-	sortedSigners := []string{}
+	beforeSortedSigners := []string{}
 	for signer := range before {
-		sortedSigners = append(sortedSigners, signer)
+		beforeSortedSigners = append(beforeSortedSigners, signer)
 	}
-	sort.Strings(sortedSigners)
+	sort.Strings(beforeSortedSigners)
 
-	for _, addy := range sortedSigners {
+	for _, addy := range beforeSortedSigners {
 		weight, ok := after[addy]
 		if !ok {
 			effects.Add(source, history.EffectSignerRemoved, map[string]interface{}{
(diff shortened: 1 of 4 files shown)


services/horizon/internal/ingest/session.go
```diff
@@ -517,8 +524,16 @@ func (is *Session) ingestSignerEffects(effects *EffectIngestion, op xdr.SetOptio
 			"weight":     weight,
 		})
 	}
+
+	sortedSigners = []string{}
```
tamirms: in services/horizon/internal/expingest/processors/effects_processor.go you create a new slice to keep sorted signers from the `after` map. 

I think you should make this code consistent with the effects processor.

So, either the effects processor can use a single `sortedSigners` slice which is cleared out before populating it with keys from the `after` map, or, this function should use two separate slices to store the keys from the `before` and `after` maps.

abuiles: you are right! I forgot to update this one!

Issue #2102, opened 2020-01-09
services/horizon: Use new ingestion data in /accounts/{id}


bartekn:
Currently we still serve account data from the core DB and just compare the data to the new ingestion system. We should switch to the new ingestion data but keep the compare code there. It should emit `ERROR` if a difference is found.

tamirms:
fixed in https://github.com/stellar/go/pull/2115

PR #2115, opened 2020-01-14
services/horizon/internal/actions: Serve /accounts/{id} using expingest data

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Fixes https://github.com/stellar/go/issues/2102

> Currently we still serve account data from the core DB and just compare the data to the new ingestion system. We should switch to the new ingestion data but keep the compare code there. It should emit ERROR if a difference is found.

### Known limitations

[N/A]

### DIFF
--- a/services/horizon/internal/actions/account.go
+++ b/services/horizon/internal/actions/account.go
@@ -2,8 +2,6 @@ package actions
 
 import (
 	"context"
-	"encoding/json"
-	"fmt"
 	"net/http"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


services/horizon/internal/actions/account.go
```diff
@@ -60,89 +57,91 @@ func AccountInfo(ctx context.Context, cq *core.Q, hq *history.Q, addr string, en
...
 
+		otherResource, err = accountFromExperimentalIngestion(ctx, hq, addr)
+		if err != nil {
+			err = errors.Wrap(err, "Could not fetch account from expingest tables")
+		}
 	}
 
-	return nil
+	if err == nil {
+		err = accountResourcesEqual(*resource, *otherResource)
```
bartekn: @tamirms I wonder if we should return `err` in `AccountInfo` if new system is running so the user doesn't see incorrect data. We didn't before because we were serving data from core.

PR #2116, opened 2020-01-14
services/horizon: Update ingestion state machine to be able to upgrade from old ingestion system

### LINKED ISSUE #2088: Update ingestion state machine to be able to upgrade from old ingestion system
When user upgrades from the old ingestion system to the new one that also needs to ingest state in addition to history data the coordination is needed to properly catchup historical data. The state machine needs to be updated to handle several upgrade cases. See #1969. The `init` state function can look like:
```go
func init() (State, err) {
	// Here or as function params:
	ingestVersion := s.historyQ.GetExpIngestVersion()
	lastIngestedLedger := s.historyQ.GetLastLedgerExpIngest()
	lastHistoryLedger := s.historyQ.GetLastHistoryLedger()
	lastCheckpoint := historyArchive.GetLastCheckpoint()

	if lastHistoryLedger == 0 {
		// No ledgers in history_ledgers so start from the latest checkpoint and
		// continue using ledgerbackend.
		return BuildStateAndResume, nil
	}

	// Now: lastHistoryLedger != 0 => there are ledgers in history_ledgers

	if ingestVersion != CurrentVersion {
		// Ingestion version upgrade.
		switch {
		case lastHistoryLedger < lastCheckpoint-1:
			// Catchup to lastCheckpoint-1 and then BuildStateAndResume
			return Catchup, nil
		case lastHistoryLedger == lastCheckpoint-1:
			// We can BuildStateAndResume without catching up
			return BuildStateAndResume, nil
		case lastHistoryLedger > lastCheckpoint-1:
			// Wait for the next checkpoint that is ahead of lastHistoryLedger
			return WaitForCheckpoint, nil
		}
	}

	// Now:
	// - ingestVersion == CurrentVersion => no upgrade
	// - lastHistoryLedger != 0 => there are ledgers in history_ledgers
	return LoadStateInMemory, nil
}
```

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This commit adds ingest state machine states to allow upgrading from the old ingestion system.

Close #2088.

### Why

When `expingest` system becomes a default ingestion system in Horizon it's possible that the old system ingested ledgers into `history_*` tables. It's also possible that experimental ingestion system was running in the past but was later turned off. We need to support upgrading in both scenarios.

Two new states have been added to the state machine:
* `catchupHistoryState` - runs `RangeSession` to ingest history data up to a given checkpoint or ledger. It's important to not update last ingested ledger in the DB during catchup.
* `waitForCheckpointState` - waits for 10 seconds for stellar-core to create a new HAS/checkpoint.

`initState` has been updated to correctly transition to the new states. The algorithm is as follows:
* If we need to reingest state (`ingestVersion != CurrentVersion || lastIngestedLedger == 0`):
  * If there are ledgers in `history_ledgers`:
    * If the latest ledger in history is greater than the latest checkpoint => `waitForCheckpointState`.
    * If the latest ledger in history is less than the latest checkpoint => `catchupHistoryState` up to the latest checkpoint.
    * If the latest ledger in history is equal to the latest checkpoint => `buildStateAndResumeIngestionState` but build the state from the specific checkpoint (the one loaded in this state to prevent situation when new checkpoint is created in a meantime).
  * Else (no are ledgers in `history_ledgers`): `buildStateAndResumeIngestionState` using the latest checkpoint.
* Else: we resume ingestion (`ingestVersion == CurrentVersion && lastIngestedLedger != 0`):
  * If the latest ledger in history is greater than the last expingest ledger => we need to catchup state but instead of complicating the pipeline to ingest meta only we just reset last expingest ledger so the state is rebuilt. Then we run `init` state.
  * If the latest ledger in history is less than the last expingest ledger => `catchupHistoryState` up to the latest expingest ledger.
  * If the latest ledger in history is equal to the last expingest ledger => we follow the old code, we ingest offers into memory, then resume ingestion.

To support `catchupHistoryState` pipeline hooks have been updated to not ingest state and not update exphistory latest ledger and version if processing in this state.


services/horizon/internal/expingest/main.go
```diff
@@ -367,6 +384,11 @@ func (s *System) init() (state, error) {
 		return state{systemState: initState}, errors.Wrap(err, "Error getting exp ingest version")
 	}
 
+	lastHistoryLedger, err := s.historyQ.GetLatestLedger()
```
tamirms: is it possible that `GetLatestLedger` returns a no rows error?

abuiles: I think it will always return a value because it's using `COALESCE` in the query `SELECT COALESCE(MAX(sequence), 0) FROM history_ledgers`

services/horizon/internal/expingest/main.go
```diff
@@ -376,35 +398,82 @@ func (s *System) init() (state, error) {
...
+			return state{systemState: initState}, errors.Wrap(err, "Error updating last ingested ledger")
+		}
+		err = s.historyQ.Commit()
 		if err != nil {
-			return state{systemState: initState}, errors.Wrap(err, "Error clearing ingest tables")
+			return state{systemState: initState}, errors.Wrap(err, "Error updating last ingested ledger")
 		}
+		return state{systemState: initState}, nil
+	case lastHistoryLedger < lastIngestedLedger:
+		// Expingest was running at some point the past but was turned off.
```
abuiles: isn't the "old ingest" aka "ingestion" which was turned off in this scenario (not expingest)?

services/horizon/internal/expingest/main.go
```diff
@@ -494,6 +585,26 @@ func (s *System) resume() (state, error) {
...
+		s.disableStateVerification = s.config.DisableStateVerification
+	}()
+
+	err := s.rangeSession.Run()
+	return state{systemState: initState}, err
+}
+
+func (s *System) waitForCheckpoint() (state, error) {
+	log.Info("Waiting for the next checkpoint...")
+	time.Sleep(10 * time.Second)
```
abuiles: @bartekn when doing state verification we wait 40 seconds -- is there a specific reason (different than it will wait less) for using one vs the other?

bartekn: In state verification we know that the checkpoint ledger was closed, here we are waiting for it. It's possible that we need to wait entire 5 minutes. I choose 10 seconds to update the user but if `waitForCheckpoint` moves to `init` and the checkpoint is still not there it will go to `waitForCheckpoint` again.

services/horizon/internal/expingest/main.go
bartekn: To reach this line all of the conditions must be true:
* `lastIngestedLedger != 0`: Expingest was running (`exp_latest_ledger` was updated)
* `ingestVersion == CurrentVersion`:  Expingest was running the current version
* `lastHistoryLedger < lastIngestedLedger`: `history_` tables are behind `exp_latest_ledger`. This can actually only happen if `INGEST=false` and `EXPINGEST=true` in the past. Weird, but can happen.

In such case we need to synchronise history tables with state tables. To do it we catchup history.

services/horizon/internal/expingest/processors/ledgers_processor.go
```diff
@@ -0,0 +1,94 @@
...
+		// Do not overwrite the previous error
+		if err == nil {
+			err = closeErr
+		}
+	}()
+	defer w.Close()
+	r.IgnoreUpgradeChanges()
+
+	// Exit early if not ingesting into a DB
+	if v := ctx.Value(IngestUpdateDatabase); v == nil {
```
abuiles: would this still work if we put `false` in `IngestUpdateDatabase`?

services/horizon/internal/expingest/processors/database_processor.go
```diff
@@ -191,6 +191,13 @@ func (p *DatabaseProcessor) ProcessLedger(ctx context.Context, store *pipeline.S
 	}()
 	defer w.Close()
 
+	// Exit early if not ingesting state (history catchup). The filtering in parent
+	// processor should do it, unfortunately it won't work in case of meta upgrades.
+	// Should be fixed after ingest refactoring.
+	if v := ctx.Value(IngestUpdateState); v == nil {
```
abuiles: same question as below

services/horizon/internal/expingest/main.go
tamirms: I think it's worth adding this test case:

```golang
func TestGetLatestLedgerEmptyDB(t *testing.T) {
	tt := test.Start(t)
	defer tt.Finish()
	q := &Q{tt.HorizonSession()}

	value, err := q.GetLatestLedger()
	tt.Assert.NoError(err)
	tt.Assert.Equal(uint32(0), value)
}
```

services/horizon/internal/expingest/main.go
```diff
@@ -69,16 +70,19 @@ type Config struct {
...
 	GetLastLedgerExpIngest() (uint32, error)
 	GetExpIngestVersion() (int, error)
 	UpdateLastLedgerExpIngest(uint32) error
 	UpdateExpStateInvalid(bool) error
+	UpdateExpIngestVersion(int) error
 	GetExpStateInvalid() (bool, error)
 	GetAllOffers() ([]history.Offer, error)
+	GetLatestLedger() (uint32, error)
 	TruncateExpingestStateTables() error
-	RemoveExpIngestHistory(uint32) (history.ExpIngestRemovalSummary, error)
```
abuiles: I forgot it, what was the reason we were doing this and not doing it any more?

abuiles: nvm - we don't want to do this anymore since we won't have both systems running.

services/horizon/internal/expingest/pipelines.go
```diff
@@ -223,6 +217,11 @@ func preProcessingHook(
 		}
 	}
 
+	// Update context so pipeline skip state processors when doing history catchup.
```
tamirms: I think we also need to make sure `system.setStateReady()` is only called if `system.state.systemState != catchupHistoryState`

services/horizon/internal/expingest/main.go
```diff
@@ -376,35 +398,82 @@ func (s *System) init() (state, error) {
...
+			// * Build state from the last checkpoint if the latest history ledger
+			//   is equal to the latest checkpoint ledger.
+			lastCheckpoint, err := s.historyAdapter.GetLatestLedgerSequence()
+			if err != nil {
+				return state{systemState: initState}, errors.Wrap(err, "Error getting last checkpoint")
+			}
+
+			switch {
+			case lastHistoryLedger > lastCheckpoint:
+				return state{systemState: waitForCheckpointState}, nil
```
tamirms: would it make sense to release the transaction either her or in the beginning of `waitForCheckpoint()`?

services/horizon/internal/expingest/processors/ledgers_processor.go
bartekn: Good point, yes it would! Will add an extra check for `true` here.

services/horizon/internal/expingest/main.go
bartekn: We could do it but it won't change anything. The other ingesting instances would also transition to `waitForCheckpoint` state (if checkpoint is not there yet).

services/horizon/internal/expingest/pipelines.go
bartekn: It's going to be removed in #2123 so leaving as is.

Issue #2111, opened 2020-01-13
exp/ingest/io: Ignore operation meta if tx result is txInternalError


bartekn:
Due to a tx meta generation bug in stellar-core (https://github.com/stellar/stellar-core/pull/2383) it's possible that tx meta is incorrect for transactions with `txInternalError` result code. For such transactions, operations meta should be ignored but we should still apply `TxChanges` in v1 and `txChangesBefore` in v2 ([confirmed](https://stellarfoundation.slack.com/archives/C0367MGFL/p1578704980059000) by @MonsieurNicolas).

abuiles:
We need to fix it here too https://github.com/stellar/go/blob/release-horizon-v0.25.0/exp/ingest/io/ledger_transaction.go#L204

abuiles:
Fixed in #2117

PR #2117, opened 2020-01-15
exp/ingest/io: Ignore operation and txChangesAfter meta if tx result is txInternalError.

abuiles:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Ignore operation and txChangesAfter meta if tx result is txInternalError.

### Why

Due to a tx meta generation bug in stellar-core (https://github.com/stellar/stellar-core/pull/2383) it's possible that tx meta is incorrect for transactions with `txInternalError` result code. For such transactions, operations meta should be ignored but we should still apply `TxChanges` in v1 and `txChangesBefore` in v2. See #2111  for more info.

### Known limitations

[TODO or N/A]

### DIFF
--- a/exp/ingest/io/ledger_transaction.go
+++ b/exp/ingest/io/ledger_transaction.go
@@ -151,6 +151,11 @@ func (t *LedgerTransaction) Successful() bool {
 	return t.Result.Result.Result.Code == xdr.TransactionResultCodeTxSuccess
 }
 
+// TxInternalError returns true if the transaction result is TxInternalError
+func (t *LedgerTransaction) TxInternalError() bool {
+	return t.Result.Result.Result.Code == xdr.TransactionResultCodeTxInternalError
+}
+
 // GetFeeChanges returns a developer friendly representation of LedgerEntryChanges
 // connected to fees.
 func (t *LedgerTransaction) GetFeeChanges() []Change {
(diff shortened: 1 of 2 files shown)


exp/ingest/io/ledger_transaction.go
```diff
@@ -151,15 +151,21 @@ func (t *LedgerTransaction) Successful() bool {
 	return t.Result.Result.Result.Code == xdr.TransactionResultCodeTxSuccess
 }
 
+// TxInternalError returns true if the transaction result is TxInternalError
+func (t *LedgerTransaction) TxInternalError() bool {
```
abuiles: maybe this should be private since we are the only consumers at the moment, but I wanted to be consistent with the `Successful` method above.

exp/ingest/io/ledger_transaction.go
```diff
@@ -202,7 +220,12 @@ func (t *LedgerTransaction) GetChanges() ([]Change, error) {
 // GetOperationChanges returns a developer friendly representation of LedgerEntryChanges.
 // It contains only operation changes.
 func (t *LedgerTransaction) GetOperationChanges(operationIndex uint32) ([]Change, error) {
-	var changes []Change
+	changes := []Change{}
+
+	// Ignore operations meta if txInternalError https://github.com/stellar/go/issues/2111
+	if t.Meta.V > 0 && t.TxInternalError() {
```
tamirms: should we restrict the check to v1 and v2 or will we need to apply the check on future versions of meta?

abuiles: oh good point! Let's restrict it to v1 and v2 for now.

abuiles: fixed in 1ff614e

exp/ingest/io/ledger_transaction.go
bartekn: Nit: I'd make it private or make a public getter function like:
```go
func (t *LedgerTransaction) TxResultCode() TransactionResultCode  {
   return t.Result.Result.Result.Code
}
```
for easier access to the tx result code. Then we can use it inside `Successful`.

abuiles: @bartekn sounds good! I added `TxResultCode` and made `TxInternalError` private

Issue #2109, opened 2020-01-13
services/horizon/expingest: Ingest history data into `history_` tables


bartekn:
In #2108 we removed `ingest` package so all the new processors should now ingest data into `history_` tables into `exp_history_` tables. We should also create a DB migration removing all the experimental tables. We should also remove all the check functions in processors that compare new system objects to the old objects.

tamirms:
fixed in https://github.com/stellar/go/pull/2118

PR #2118, opened 2020-01-15
services/horizon/internal/db2/history: Remove exp_history_* tables

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Fixes https://github.com/stellar/go/issues/2109

* Add migration to remove exp_history_* tables
* Remove CheckExp functions in the processors
* Modify db ingestion functions to insert into history_ * tables instead of exp_history_* tables

### Why

We will be eliminating the legacy ingestion system in the 0.25.0 release. We should get rid of the exp_history_* tables because we will no longer be running multiple ingestion systems in parallel. 

### Known limitations

[N/A]


tamirms:
> LGTM! I wonder if we can remove `asset_stats` table since it will be no longer used (we're using `exp_asset_stats` now). Even if someone downgrades later it will be in a broken state.

yeah, I think we should remove that table. I can do it in another PR

PR #2120, opened 2020-01-15
services/horizon/internal/actions: Remove GetCreateAssetID usage from actions.

### LINKED ISSUE #2061: Remove GetCreateAssetID usage from actions
There are several places outside `ingest` package where `history.Q.GetCreateAssetID` method is used. It's possible that when an asset does not exist there may be attempts to create it more than once (ex. multiple requests at the same time). In such situation the following error will be logged:
```
get failed: pq: duplicate key value violates unique constraint "history_assets_asset_code_asset_type_asset_issuer_key"
```
Instead, `history.Q.GetAssetID` method should be used. In general, Horizon shouldn't need write access to a DB outside `ingest` package.

abuiles:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

 Remove `GetCreateAssetID` usage from actions. Fix #2061

### Why
`GetCreateAssetID` writes data to the DB when an asset is not found, but Horizon shouldn't need write access to a DB outside `ingest` package, this also causes some errors as described in #2061. 

This PR updates the action to use `GetAssetID`  and returns `404` if the user is trying to fetch data for an asset which doesn't exist.

### Known limitations

[TODO or N/A]

### DIFF
--- a/support/render/problem/problem.go
+++ b/support/render/problem/problem.go
@@ -188,9 +188,14 @@ func (ps *Problem) renderProblem(ctx context.Context, w http.ResponseWriter, p P
 // MakeInvalidFieldProblem is a helper function to make a BadRequest with extras
 func MakeInvalidFieldProblem(name string, reason error) *P {
 	br := BadRequest
-	br.Extras = map[string]interface{}{
+	AddInvalidField(&br, name, reason)
+	return &br
+}
+
+// AddInvalidField adds invalid field extras to a given P
+func AddInvalidField(p *P, name string, reason error) {
+	p.Extras = map[string]interface{}{
 		"invalid_field": name,
 		"reason":        reason.Error(),
 	}
-	return &br
 }
--- a/support/render/problem/problem.go
+++ b/support/render/problem/problem.go
@@ -192,7 +192,7 @@ func MakeInvalidFieldProblem(name string, reason error) *P {
 	return &br
 }
 
-// SetInvalidField adds invalid field extras to a given P
+// SetInvalidField sets the invalid_field key in extras with the given reason.
 func SetInvalidField(p *P, name string, reason error) {
 	p.Extras = map[string]interface{}{
 		"invalid_field": name,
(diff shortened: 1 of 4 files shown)


support/render/problem/problem.go
```diff
@@ -188,9 +188,14 @@ func (ps *Problem) renderProblem(ctx context.Context, w http.ResponseWriter, p P
 // MakeInvalidFieldProblem is a helper function to make a BadRequest with extras
 func MakeInvalidFieldProblem(name string, reason error) *P {
 	br := BadRequest
-	br.Extras = map[string]interface{}{
+	AddInvalidField(&br, name, reason)
+	return &br
+}
+
+// AddInvalidField adds invalid field extras to a given P
+func AddInvalidField(p *P, name string, reason error) {
```
tamirms: I prefer the name `SetInvalidField()`

abuiles: @tamirms good point! `Add` makes you think you can keep adding into it

support/render/problem/problem.go
```diff
@@ -188,9 +188,14 @@ func (ps *Problem) renderProblem(ctx context.Context, w http.ResponseWriter, p P
...
 func MakeInvalidFieldProblem(name string, reason error) *P {
 	br := BadRequest
-	br.Extras = map[string]interface{}{
+	SetInvalidField(&br, name, reason)
+	return &br
+}
+
+// SetInvalidField sets the invalid_field key in extras with the given reason.
+func SetInvalidField(p *P, name string, reason error) {
+	p.Extras = map[string]interface{}{
```
bartekn: Using pointer to `P` is dangerous here. For example, in trade action you call it like:
```go
notFound := problem.NotFound
problem.SetInvalidField(
	&notFound,
	"base_asset",
	errors.New("not found"),
)
```
what means that you'll add `Extras` to the global `problem.NotFound`. So all of the other usages of `problem.NotFound` will have this value set. Instead, you should copy the object.

abuiles: fixed via 3a91efb

PR #2138, opened 2020-01-21
services/horizon/expingest/processors: Refactor Horizon processors

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

### Why

### Known limitations

[TODO or N/A]

### DIFF
--- a/services/horizon/internal/expingest/processors/account_data_processor.go
+++ b/services/horizon/internal/expingest/processors/account_data_processor.go
@@ -1,0 +1,131 @@
+package processors
+
+import (
+	ingesterrors "github.com/stellar/go/exp/ingest/errors"
+	"github.com/stellar/go/exp/ingest/io"
+	"github.com/stellar/go/services/horizon/internal/db2/history"
+	"github.com/stellar/go/support/errors"
+	"github.com/stellar/go/xdr"
+)
+
+type ChangesMode int
+
+const (
+	InitChangesMode ChangesMode = iota
+	LedgerChangesMode
+)
+
+type AccountDataProcessor struct {
+	DataQ history.QData
+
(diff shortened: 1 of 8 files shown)


services/horizon/internal/expingest/processors/account_data_processor.go
```diff
@@ -0,0 +1,136 @@
...
+	cache *io.LedgerEntryChangeCache
+	batch history.AccountDataBatchInsertBuilder
+}
+
+func (p *AccountDataProcessor) Init(header xdr.LedgerHeader, mode ChangesMode) error {
+	p.mode = mode
+	p.batch = nil
+	p.cache = nil
+	switch p.mode {
+	case InitChangesMode:
```
abuiles: while this makes sense, maybe it would be easier to follow and understand if we keep each mode as a different processor, that way it's easier to reason about this code  (you don't have two thing about mode, batch and cache)

bartekn: I agree! Please check the code with the latest updates. `ProcessChange` not works exactly the same when processing history and tx meta. However, a couple notes:
* We need to flush changes in `LedgerEntryChangesCache` when it becomes too big (especially when ingesting history buckets) but we have to do it anyway.
* We need to change `LedgerEntryChangeType` streamed by `StateReader` to `LedgerEntryChangeTypeLedgerEntryCreated` (we should do it anyway). This way a single processor can handle both cases -> less code duplication.
* We actually need to switch to batch insert. This removes the ability to check correctness using `rowsAffected` but we have `StateVerifier` so I think it's fine.

PR #2139, opened 2020-01-21
services/horizon: Remove EnableExperimentalIngestion and EnableAssetStats flags

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This commit removes `EnableExperimentalIngestion` and `EnableAssetStats` flags.

### Why

New ingestion becomes the default system so feature flag is not needed. In the new system asset stats ingestion does not cause performance issues so it can't be disabled.

### Known limitations

Some actions' tests require DB migrations to work. Test scenarios will be rebuilt in another PR for readability.
Some structs are not longer used. Will be removed in another PR for readability.


services/horizon/internal/actions_account_test.go
```diff
@@ -1,81 +1,81 @@
 package horizon
```
tamirms: I think we can delete this file since there are tests covering this endpoint in services/horizon/internal/actions/account_test.go

services/horizon/internal/actions_assets_test.go
```diff
@@ -1,303 +1,285 @@
 package horizon
```
tamirms: I think we can delete this file since there are tests covering this endpoint in services/horizon/internal/actions/asset_test.go

services/horizon/internal/actions_offer_test.go
```diff
@@ -1,94 +1,94 @@
 package horizon
 
-import (
```
tamirms: I think we can delete this file since there are tests covering this endpoint in services/horizon/internal/actions/offer_test.go

services/horizon/internal/actions_order_book_test.go
```diff
@@ -1,120 +1,120 @@
 package horizon
 
-import (
```
tamirms: I think we can delete this file since there are tests covering this endpoint in services/horizon/internal/actions/orderbook_test.go

services/horizon/internal/actions_path_test.go
```diff
@@ -1,612 +1,612 @@
 package horizon
```
tamirms: I don't think we need to comment out the entire file, we can just delete the tests which covered the db path finding implementation

tamirms: @bartekn let me know if you'd like me to take this task and I can just push a commit to this PR

bartekn:
@tamirms I removed `actions_assets_test.go`, `actions_offer_test.go` and `actions_order_book_test.go` files as the files you mentioned have similar or better coverage. I was able to revert all tests in `actions_path_test.go` they were failing because of expingest middleware that is not tested in them. I left `actions_account_test.go` because it seems there are some tests not covered in the other file.

PR #2154, opened 2020-01-23
support/render/health: add package for simple health check endpoint

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Add a package that contains simple HTTP handler and response types for a simple health endpoint for our services. Uses the IETF draft _Health Check Response Format for HTTP APIs_ for the format of the response.

```json
{
  "status": "pass"
}
```

https://tools.ietf.org/id/draft-inadarei-api-health-check-01.html

### Why
We don't currently surface simple health endpoints on many of the services we have in this repository. I'd like to start surfacing a very simple `/health` endpoint for two new services I'm adding to the repository and I'd like to follow some consistent pattern with both that will work with any other service we build.

It's worth noting that while we don't surface explicit health endpoints, we already surface metrics endpoints in some apps specifically for consumption by Promethesus. Metrics endpoints often surface a lot of health data, but this endpoint is not attempting to replicate that endpoint or surface any data we might normally surface as a metric.

Looking around at what most web services implement in this area, there is not much consistency. The things that are consistent is that a health endpoint is typically named something like `/health` or `/healthz`, and when it responds with status `200 OK` then all is good. Everything beyond that is without standard. Most services reply with content, most JSON (e.g. etcd), some plain text (e.g. Promothesus). Most return a simple 👍 or 👎 indicator in that content, and very few will provide some detail beyond that.

It seems simple and intuitive to follow the crowd on the path and use `/health` for our endpoints as it contains nothing that needs explaining and is unlikely to overlap with a path we provide for some other purpose.

The consumer of the endpoint will be deployment environments like Kubernetes, or load balances like AWS ELBs or Google Cloud Load Balancers. The most important thing to those systems is that the endpoint replies with a status code 200 when all is good, and 4xx-5xx when the service isn't healthy, but some HTTP load balancers may also check the response content for text.

It seems worthwhile to follow the crowd on the format too and use JSON since it is human and easily parsed by many tools. If a deployer of any of our reference implementations wish to parse the contents it will be easy to do so with a variety of tools. I'm not using Promothesus exposition format here because any data that we'd want to get into Promothesus should be on a separate `/metrics` endpoint which is its standard. If at any point we added additional data to the health endpoint that we or someone else wanted to import into Promethesus we could either duplicate that data into `/metrics` or offer a promethesus version of the health endpoint.

For the exact structure of the JSON we can really choose anything we want. Using a standard for this format removes the need for us to make decisions if we need to expand the response and means we don't need to be as creative.

I've selected the IETF draft because it looks like it meets the simple requirements noted above, and it already matches some existing web application frameworks that provide health endpoints (e.g. Node Terminus, Java's SpringBoot). The smallest message it defines for a health response is `{"status":"pass"}` which is basically what we need.

Today the services we have will likely return a static `200` `{"status":"pass"}` because if the HTTP server is up, the service is ready, and in the future we might have services that return `500` `{"status":"fail"}` if the service cannot access critical infrastructure like AWS services or a database.

Because of the lightweight nature of this standard, we can abandon this standard at anytime if we discover it constrains us in ways we do not like or is not serving us.

### Known limitations

N/A


exp/services/webauth/internal/serve/health.go
```diff
@@ -0,0 +1,20 @@
...
+import (
+	"net/http"
+
+	"github.com/stellar/go/support/render/httpjson"
+)
+
+type healthHandler struct{}
+
+type healthResponse struct {
+	OK bool `json:"ok"`
```
howardtw: We use [this](https://github.com/stellar/go/blob/master/support/render/httpjson/handler.go#L20) response as our default response at most of the endpoints that return a non-empty response body. The `/health` endpoint of pasapeso actually returns an empty body. It would be great if it's consistent.

exp/services/webauth/internal/serve/health_test.go
```diff
@@ -0,0 +1,28 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestHealth(t *testing.T) {
+	h := healthHandler{}
+
+	r := httptest.NewRequest("POST", "/", nil)
```
howardtw: Should this be a `GET` request?

exp/services/webauth/internal/serve/health.go
leighmcculloch: What cases do we use the default response `{"message":"ok"}` in? I don't think we need to necessarily use that here.

I think we should avoid an empty body. In many deployments it may be difficult to differentiate where exactly the 200 is coming from, if there is some other system in between that is returning the 200 and no content. The use of some meaningful content is also easier to view in a browser.

All of the services I checked in this repository do not have a health endpoint. Horizon's root path is commonly used for health checks. So we don't have a ton of precedence to follow for a response format.

I agree it would be useful if we had a consistent format. If we were to pick one, there's a IETF network working group draft that's relatively new from a couple years ago.

https://tools.ietf.org/id/draft-inadarei-api-health-check-01.html

While the response they define provides the option of using many fields, we could start with the most basic implementation that communicates least information. It'd look like the below. We can expand on this whenever we feel like we need to.

```json
{
  "status": "pass"
}
```

exp/services/webauth/internal/serve/health_test.go
leighmcculloch: Oops, yup it should. Thanks. b1b676da

exp/services/webauth/internal/serve/health.go
leighmcculloch: I've added a starting point here, what do you think? 9cb27b3b

leighmcculloch: We could take it a step further and put this basic health handler in another package. I expect I'll be implementing it 3-4 times, may as well put it in one place, and if we add more fields it will be easier to discover for new apps. I've done that in c9a97ebb.

howardtw: > What cases do we use the default response {"message":"ok"} in?

If we have a function that doesn't return anything, this would be the default response. However, as far as I can tell, every function that we've implemented and is wrapped by this package has at least one returned argument, which is an error. So no, we are probably not using the default response at this moment.

You're right. Maybe they are orthogonal, and the format suggested by ietf seems good. I like the idea of handling `/health` endpoint with a handler or other means. Separating it from HTTPJSON is great.

howardtw:
One request, can we do it in separate PRs? I think it would be ideal if we can handle heathjson in a standalone PR.

support/render/health/example_test.go
```diff
@@ -0,0 +1,67 @@
+package health_test
```
howardtw: What are we testing here? Any reason that we don't want to use testify in this file?

leighmcculloch: It's an example test file to show someone how to use the package, and not intended to test the package. I'm not using testify because we wouldn't use it in application code and so don't think it belongs in an example. There is also no testing.T in an example test to give to testify.


On the line you commented on I'm using the _test suffix so that I have to import the health package. I'm not sure if that's what you were concerned about, but example files often do this so that the example code looks like code another package will write with functions in this package referenced via an import.

leighmcculloch: I added the example fine when I removed the webauth changes since after removing the webauth changes the was no longer any example of how the new package should be used.

howardtw: Ok cool. I was just wondering the motivation of having this file. I have no problem with the `_test suffix`. It's a common pattern. Thanks!

PR #2152, opened 2020-01-23
services/horizon: Cleanup routes in webserver init code

bartekn:
<details>
  <summary>PR Checklist</summary>
  
### PR Structure

* [ ] This PR has reasonably narrow scope (if not, break it down into smaller PRs).
* [ ] This PR avoids mixing refactoring changes with feature changes (split into two PRs
  otherwise).
* [ ] This PR's title starts with name of package that is most changed in the PR, ex.
  `services/friendbot`, or `all` or `doc` if the changes are broad or impact many
  packages.

### Thoroughness

* [ ] This PR adds tests for the most critical parts of the new functionality or fixes.
* [ ] I've updated any docs ([developer docs](https://www.stellar.org/developers/reference/), `.md`
  files, etc... affected by this change). Take a look in the `docs` folder for a given service,
  like [this one](https://github.com/stellar/go/tree/master/services/horizon/internal/docs).

### Release planning

* [ ] I've updated the relevant CHANGELOG ([here](services/horizon/CHANGELOG.md) for Horizon) if
  needed with deprecations, added features, breaking changes, and DB schema changes.
* [ ] I've decided if this PR requires a new major/minor version according to
  [semver](https://semver.org/), or if it's mainly a patch change. The PR is targeted at the next
  release branch if it's not a patch change.
</details>

### What

This commit cleans the code responsible for Horizon webserver init.

### Why

* Moved all routes using the `StateMiddleware` (previously `ExperimentalIngestionMiddleware`) to avoid code duplication.
* Start routes using order book graph only if the instance is ingesting.

PR #2155, opened 2020-01-23
services/horizon: Fix empty route param bug

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

Fixes a bug introduced in #2152. Paths like: `/accounts//payments` return all payments instead of 404. This can lead to dangerous bugs when account ID is omited in request by accident.

### DIFF
--- a/services/horizon/internal/web.go
+++ b/services/horizon/internal/web.go
@@ -200,12 +200,13 @@ func (w *web) mustInstallActions(
 	})
 
 	// account actions - /accounts/{account_id} has been created above so we
-	// need to use absolute routes here.
-	r.Get("/accounts/{account_id}/transactions", w.streamIndexActionHandler(w.getTransactionPage, w.streamTransactions))
-	r.Get("/accounts/{account_id}/operations", OperationIndexAction{}.Handle)
-	r.Get("/accounts/{account_id}/payments", OperationIndexAction{OnlyPayments: true}.Handle)
-	r.Get("/accounts/{account_id}/effects", EffectIndexAction{}.Handle)
-	r.Get("/accounts/{account_id}/trades", TradeIndexAction{}.Handle)
+	// need to use absolute routes here. Make sure we use regexp check here for
+	// emptiness. Without it, requesting `/accounts//payments` return all payments!
+	r.Get("/accounts/{account_id:\\w+}/transactions", w.streamIndexActionHandler(w.getTransactionPage, w.streamTransactions))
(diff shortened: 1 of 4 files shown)


services/horizon/internal/web.go
```diff
@@ -200,12 +200,13 @@ func (w *web) mustInstallActions(
...
-	r.Get("/accounts/{account_id}/payments", OperationIndexAction{OnlyPayments: true}.Handle)
-	r.Get("/accounts/{account_id}/effects", EffectIndexAction{}.Handle)
-	r.Get("/accounts/{account_id}/trades", TradeIndexAction{}.Handle)
+	// need to use absolute routes here. Make sure we use regexp check here for
+	// emptiness. Without it, requesting `/accounts//payments` return all payments!
+	r.Get("/accounts/{account_id:\\w+}/transactions", w.streamIndexActionHandler(w.getTransactionPage, w.streamTransactions))
+	r.Get("/accounts/{account_id:\\w+}/operations", OperationIndexAction{}.Handle)
+	r.Get("/accounts/{account_id:\\w+}/payments", OperationIndexAction{OnlyPayments: true}.Handle)
+	r.Get("/accounts/{account_id:\\w+}/effects", EffectIndexAction{}.Handle)
+	r.Get("/accounts/{account_id:\\w+}/trades", TradeIndexAction{}.Handle)
```
leighmcculloch: It's sad there's not a way to specify this without firing up a regular expression, but I don't see any other way to do it. I wonder if we should try to come up with something and contribute it back to go-chi/chi, since regular expressions are expensive.

bartekn: I agree. The other thing is that it worked correctly before https://github.com/stellar/go/pull/2152 so when the code looked like:
```go
	r.Route("/accounts", func(r chi.Router) {	
		r.Route("/{account_id}", func(r chi.Router) {	
			r.Get("/", handler)
		})
	})
```
so it's confusing.

When it comes to contributing back, I think it's a good idea. There are many options: the required params can use different brackets (`[]`?) or be prepended with a special character (`{!account_id}`?).

leighmcculloch: I've opened an issue about this here: go-chi/chi#492.

pkieltyka:
this is a not a bug of chi, you should always be validating your inputs.

leighmcculloch:
@pkieltyka 👍 That's what we discovered. The bug was in our code because we made the wrong assumption about how empty path parameters work and are using the regular expression to indicate the path parameter is required to have a value.

PR #2164, opened 2020-01-27
Processor Interfaces for new ingestion design 

tamirms:
exp/ingest/processors.go has all the interfaces for the new ingestion design.'
All processors in the `services/horizon/internal/expingest/processors` package should implement `ChangeProcessor` or `LedgerTransactionProcessor`

### DIFF
--- a/exp/ingest/processors.go
+++ b/exp/ingest/processors.go
@@ -1,0 +1,88 @@
+package ingest
+
+import (
+	"github.com/stellar/go/exp/ingest/io"
+	"github.com/stellar/go/xdr"
+)
+
+type ChangeProcessor interface {
+	Init(sequence uint32) error
+	ProcessChange(change io.Change) error
+	Commit() error
+}
+
+type LedgerTransactionProcessor interface {
+	Init(header xdr.LedgerHeader) error
+	ProcessTransaction(transaction io.LedgerTransaction) error
+	Commit() error
+}
+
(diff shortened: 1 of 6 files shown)


tamirms:
fixes https://github.com/stellar/go/issues/2165

exp/ingest/processors.go
```diff
@@ -0,0 +1,88 @@
+package ingest
+
+import (
+	"github.com/stellar/go/exp/ingest/io"
+	"github.com/stellar/go/xdr"
+)
+
+type ChangeProcessor interface {
+	Init(sequence uint32) error
```
bartekn: I wonder if we should change it to:
```suggestion
	Init(header xdr.LedgerHeader) error
```
With different `Init` definition than in `LedgerTransactionProcessor` it won't be possible to create a processor that processes both changes and transactions (you cannot redeclare method in Golang). But what's more important `LedgerHeader` contains information that can be useful for processing in the future (ex. protocol version, close time, etc.).

tamirms: how do you get a `xdr.LedgerHeader` from the history archive snapshot?

bartekn: It's there in `ledgers` directory (see: https://github.com/stellar/stellar-core/blob/master/docs/history.md). For now, we can get it from stellar-core DB (because we get it anyway to compare bucket list hash).

Issue #2168, opened 2020-01-27
Implement Change Reader on top of DatabaseBackend


tamirms:
We currently use [`exp/ingest/io/DBLedgerReader`](https://github.com/stellar/go/blob/master/exp/ingest/ledgerbackend/database_backend.go) to extract ledger information from stellar core. 

Our experience building out the ingestion system has showed us that there are essentially two types of data we need to extract from Stellar Core when ingesting new ledgers:

* `[]io.Change` - changes from operations, fee meta, upgrades, etc 
* `[]io.LedgerTransaction` - for each transaction in the ledger

Although obtaining a list of `io,LedgerTransaction`s is straight forward using a `DBLedgerReader`, getting a list of `io.Change`s and applying them to processors in the correct order is not easy. To address this use case we should create a new reader built on top of `DatabaseBackend`.

The `ChangeReader` is bound to a specific ledger sequence and implements the following function which can be used to iterate through all changes (in the correct order from the ingestion perspective) within a ledger:

`Read() (io.Change, error)`

tamirms:
fixed in https://github.com/stellar/go/pull/2178

PR #2171, opened 2020-01-27
exp/ingest/io: Add ChangeReader implementation

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

You can ignore commits before https://github.com/stellar/go/commit/83feb925453c108cf0bf1502bc4807784340c8b2 because those are part of this PR https://github.com/stellar/go/pull/2164

* Reduce scope of `LedgerReader` interface so that it's only responsible for streaming `io.LedgerTransactions`
* Add `ChangeReader` which is implemented using a `DBLedgerReader` to stream `io.Changes` in the correct order

### Why

Fixes https://github.com/stellar/go/issues/2168

> Our experience building out the ingestion system has showed us that there are essentially two types of data we need to extract from Stellar Core when ingesting new ledgers:

> * `[]io.Change` - changes from operations, fee meta, upgrades, etc 
> * `[]io.LedgerTransaction` - for each transaction in the ledger

> Although obtaining a list of `io,LedgerTransaction`s is straight forward using a `DBLedgerReader`, getting a list of `io.Change`s and applying them to processors in the correct order is not easy. To address this use case we should create a new reader built on top of `DatabaseBackend`.

### Known limitations

* `DBLedgerReader` and `changeReader` are no longer thread-safe. We do not intend to share `LeddgerReaders` or `ChangeReaders` with multiple goroutines so I think it's ok to remove thread-safety.
* The PR lacks tests. If the design looks good, I will add tests in a subsequent commit.

### DIFF
--- a/exp/ingest/io/change_reader.go
+++ b/exp/ingest/io/change_reader.go
@@ -68,10 +68,7 @@ func (r *changeReader) getNextFeeChange() (Change, error) {
 			}
 		}
 
-		var changes []Change
-		for _, change := range transaction.GetFeeChanges() {
-			changes = append(changes, change)
-		}
+		changes := transaction.GetFeeChanges()
 		if len(changes) >= 1 {
 			r.pending = append(r.pending, changes[1:]...)
 			return changes[0], nil
--- a/exp/ingest/io/change_reader.go
+++ b/exp/ingest/io/change_reader.go
@@ -25,7 +25,7 @@ type changeReader struct {
 	pendingIndex           int
 }
 
-// Ensure DBLedgerReader implements LedgerReader
+// Ensure changeReader implements ChangeReader
 var _ ChangeReader = (*changeReader)(nil)
 
 // NewChangeReader constructs a new ChangeReader instance bound to the given eldger
--- a/exp/ingest/io/change_reader.go
+++ b/exp/ingest/io/change_reader.go
@@ -1,0 +1,149 @@
+package io
+
+import (
+	"io"
+
+	"github.com/stellar/go/exp/ingest/ledgerbackend"
+	"github.com/stellar/go/xdr"
+)
+
+// ChangeReader provides convenient, streaming access to changes within a ledger.
+type ChangeReader interface {
+	GetSequence() uint32
+	GetHeader() xdr.LedgerHeaderHistoryEntry
+	// Read should return the next `Change` in the leader. If there are no more
+	// changes left it should return an `io.EOF` error.
+	Read() (Change, error)
+}
+
+type changeReader struct {
+	dbReader               DBLedgerReader
+	streamedFeeChanges     bool
+	streamedMetaChanges    bool
+	streamedUpgradeChanges bool
+	pending                []Change
+	pendingIndex           int
+}
+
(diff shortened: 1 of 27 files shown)


exp/ingest/io/change_reader.go
```diff
@@ -0,0 +1,146 @@
...
+	"github.com/stellar/go/xdr"
+)
+
+// ChangeReader provides convenient, streaming access to changes within a ledger.
+type ChangeReader interface {
+	GetSequence() uint32
+	GetHeader() xdr.LedgerHeaderHistoryEntry
+	// Read should return the next `Change` in the leader. If there are no more
+	// changes left it should return an `io.EOF` error.
+	Read() (Change, error)
```
bartekn: We need to add `Close()` to be able to close DB connection or others in other implementations.

exp/ingest/io/ledger_reader.go
```diff
@@ -2,26 +2,31 @@ package io
...
 	"github.com/stellar/go/xdr"
 )
 
+// LedgerReader provides convenient, streaming access to the transactions within a ledger.
+type LedgerReader interface {
+	GetSequence() uint32
+	GetHeader() xdr.LedgerHeaderHistoryEntry
+	// Read should return the next transaction. If there are no more
+	// transactions it should return `io.EOF` error.
+	Read() (LedgerTransaction, error)
```
bartekn: `Close()` method - see other comment.

exp/ingest/io/ledger_reader.go
```diff
@@ -55,52 +60,26 @@ func (dblrc *DBLedgerReader) GetHeader() xdr.LedgerHeaderHistoryEntry {
 // Read returns the next transaction in the ledger, ordered by tx number, each time it is called. When there
 // are no more transactions to return, an EOF error is returned.
 func (dblrc *DBLedgerReader) Read() (LedgerTransaction, error) {
-	// Protect all accesses to dblrc.readIdx
-	dblrc.readMutex.Lock()
-	defer dblrc.readMutex.Unlock()
-
```
bartekn: Since thread-safety can be guaranteed at a very low cost, I'd leave it here. Otherwise, please update the comment that the methods here are not thread-safe.

exp/ingest/io/change_reader.go
tamirms: when you create a ChangeReader or a LedgerReader  you pass in a DB connection. I don't think it makes sense for the object to close a database connection which is not owned by the object.

ChangeReaders and LedgerReaders essentially do queries operations on the stellar core db. It would be weird to have a function which takes a database connection returns the query result and then closes the database connection as a side effect.

bartekn: Maybe `Close()` is not essential in the current implementation of the interface but it doesn't mean that it won't be in the future and changing the interface later can be costly. I imagine that once we have streaming API from stellar-core we'd need to close the streaming connection.

Anyway, I guess we can omit it now but we should discuss it again when stellar-core API is there or we decide to release `ingest` package for outside use (whichever is first).

PR #2177, opened 2020-01-28
services/horizon/expingest/processors: Refactor history processors

abuiles:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

### Why

### Known limitations

[TODO or N/A]


services/horizon/internal/expingest/processors/participants_processor.go
```diff
@@ -17,6 +12,17 @@ import (
 // from different sources (transactions, operations, etc)
 type ParticipantsProcessor struct {
 	ParticipantsQ history.QParticipants
+
+	ledger         xdr.LedgerHeaderHistoryEntry
```
tamirms: I think you can just store the ledger sequence because the rest of  the `xdr.LedgerHeaderHistoryEntry` variables are not used. I think this change can be applied to a few of the other TransactionProcessors as well

abuiles: sounds good!

abuiles: @tamirms fixed https://github.com/abuiles/go/commit/721ec82d84fd235d34bbad158d1d673263dac9ba

services/horizon/internal/expingest/processors/effects_processor.go
```diff
@@ -22,6 +18,16 @@ import (
 // EffectProcessor process effects
 type EffectProcessor struct {
 	EffectsQ history.QEffects
+
+	sequence uint32
+	effects  []effect
+}
+
+func NewEffectProcessor(effectsQ history.QEffects, sequence uint32) *EffectProcessor {
```
bartekn: I wonder if we should make `EffectProcessor` (and others) private to avoid creating it using composite literal. In such case `sequence` will be invalid (0).

tamirms: I think it would be good to make all the fields in the processor structs private

bartekn: Yes, this would solve the problem too.

abuiles: @tamirms @bartekn making all fields in the struct private makes sense, I'd update that. Yesterday I was actually thinking why they were even public.

PR #2197, opened 2020-01-29
clients/horizonclient: add IsNotFoundError

### LINKED ISSUE #2194: Feature Request: Simple detection of accounts not existing in Horizon

It is remarkably verbose and difficult to detect if an account exists in Horizon. The code to do so looks like this:
https://github.com/stellar/go/blob/8ff0848c31a940ab76f1a9177af0d9dd405ffa9b/txnbuild/transaction_challenge_example_test.go#L81-L92

I'd like to see a helper that can be used to inspect errors returned by the Horizon client that provides immediate and definitive feedback on whether a specific type of error has occurred.

The helper could be named something like `IsNotFoundError`.

e.g.
```go
horizonClientAccount, err := horizonClient.AccountDetail(horizonclient.AccountRequest{AccountID: txClientAccountID})
if horizonclient.IsNotFoundError(err) {
	// Account doesn't exist
} else if err != nil {
	// Some other error
}
// Code continues
```

1. We could also make changes to the error object that is returned by the Horizon Client so that we can do a quick comparison or use the new `errors.Is` logic, but that would be a breaking change to the SDK since folks could already be dependent on the error structure having the form and values it has today.

2. We could do something very similar to my proposal but add the helper to the horizonclient.Error object. This would seemingly be the right thing to do from an OOP point-of-view, but it would still require the user to unwrap the error. A helper function like above can check if the err is nil, unwrap it, and skip it if it's not a horizonclient.Error.

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add `horizonclient.IsNotFoundError` that returns true if an `error` is a Horizon `not_found` problem error.

### Why

It is remarkably verbose and difficult to detect if an account exists in Horizon. The code to do so must unwrap the horizonclient.Error, know to do so as a pointer and not as the pointer-less type, and then check that the problem's type is a long URI. In application code this is very verbose and easy to get wrong.

I'm writing code that tests if an account is present in the `exp/services/webauth` service and attempting to simplifying it.

Closes #2194 

### Known limitations

N/A

### DIFF
--- a/clients/horizonclient/error_helpers.go
+++ b/clients/horizonclient/error_helpers.go
@@ -1,0 +1,21 @@
+package horizonclient
+
+// IsNotFoundError returns true if the error is a horizonclient.Error with
+// a not_found problem indicating that the resource is not found on
+// Horizon.
+func IsNotFoundError(err error) bool {
+	var hErr *Error
+
+	switch err := err.(type) {
+	case *Error:
+		hErr = err
+	case Error:
+		hErr = &err
+	}
+
+	if hErr == nil {
+		return false
+	}
+
+	return hErr.Problem.Type == "https://stellar.org/horizon-errors/not_found"
+}
--- a/clients/horizonclient/CHANGELOG.md
+++ b/clients/horizonclient/CHANGELOG.md
@@ -3,6 +3,10 @@
 All notable changes to this project will be documented in this
 file.  This project adheres to [Semantic Versioning](http://semver.org/).
 
+## Unreleased
+
+- Add `IsNotFoundError`
+
 ## [v2.0.0](https://github.com/stellar/go/releases/tag/horizonclient-v2.0.0) - 2020-01-13
 
 - Add custom `UnmarshalJSON()` implementations to Horizon protocol structs so `int64` fields can be parsed as JSON numbers or JSON strings
--- a/clients/horizonclient/error_helpers_test.go
+++ b/clients/horizonclient/error_helpers_test.go
@@ -1,0 +1,84 @@
+package horizonclient
+
+import (
+	"testing"
+
+	"github.com/stellar/go/support/errors"
+	"github.com/stellar/go/support/render/problem"
+	"github.com/stretchr/testify/assert"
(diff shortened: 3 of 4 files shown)


clients/horizonclient/error_helpers.go
```diff
@@ -0,0 +1,21 @@
...
+		hErr = err
+	case Error:
+		hErr = &err
+	}
+
+	if hErr == nil {
+		return false
+	}
+
+	return hErr.Problem.Type == "https://stellar.org/horizon-errors/not_found"
```
abuiles: @leighmcculloch Is there a particular reason for using `Type` rather than `Status`? Using `Status` could help us prevent possible issues if we decide to change the URL or something like that. A `404` will always be a `404`, unless, it can't be found.

leighmcculloch: According to the Horizon documentation the important thing is that the error is a `not_found` error, not that the status code is 404.

>When Horizon can’t find whatever data you are requesting, it will return a `not_found` error. This is similar to a HTTP 404 Error error response.

Ref: https://www.stellar.org/developers/horizon/reference/errors/not-found.html.

Also, according to [RFC7807] the type should uniquely identify the problem, so I don't think it's something we'd ever change, or if we did it would be a breaking API change.

>"type" (string) - A URI reference [RFC3986] that identifies the problem type.

Ref: https://tools.ietf.org/html/draft-ietf-appsawg-http-problem-00#section-3.1.

leighmcculloch: Another reason: I think ideally we use the same pattern for any error helpers like this, and others will need to use the type. If the API ever returns a 404 that doesn't fit the horizon error, then it isn't really a horizon error and something else more significant has happened. This helps to protect a client that points to the wrong URL and is getting 404s from thinking resources don't exist on Horizon.

abuiles: Makes sense! thanks for the explanation

PR #2198, opened 2020-01-30
exp/services/webauth: add support for verifying accounts that do not exist

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add support for verifying accounts that do not exist behind an application option. When enabled the webauth application will allow a client to authenticate with an account that has not been created using the accounts master key.

### Why

[SEP-10 v1.3.0] describes how to optionally authenticate an account that has not yet been created on the Stellar network using the master key. It'd be good if our reference implementation of SEP-10 captured this so that implementers have a reference of how to implement this. Anchors implementing SEP-10 need to support authentication of accounts that are not yet created.

### Known limitations

N/A

### DIFF
--- a/exp/services/webauth/internal/serve/token.go
+++ b/exp/services/webauth/internal/serve/token.go
@@ -16,12 +16,13 @@ import (
 )
 
 type tokenHandler struct {
-	Logger            *supportlog.Entry
-	HorizonClient     horizonclient.ClientInterface
-	NetworkPassphrase string
-	SigningAddress    *keypair.FromAddress
-	JWTPrivateKey     *ecdsa.PrivateKey
-	JWTExpiresIn      time.Duration
+	Logger                      *supportlog.Entry
+	HorizonClient               horizonclient.ClientInterface
+	NetworkPassphrase           string
+	SigningAddress              *keypair.FromAddress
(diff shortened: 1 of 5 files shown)


exp/services/webauth/internal/serve/token.go
```diff
@@ -49,18 +50,36 @@ func (h tokenHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	var clientAccountExists bool
```
accordeiro: If we tested first if we `AllowAccountsThatDoNotExist`, we could prevent doing some Horizon calls (when `AllowAccountsThatDoNotExist = true`) which could speed things up. Does that make sense? I might be missing something.

leighmcculloch: We'd still need to call Horizon because if the account exists we want to use the signers that Horizon gives us. The does not exist is just a fallback if it turns out the account doesn't exist. It's optional because some SEP-10 servers won't need this, but anchors will if they are implementing SEP-24 and choose to support accounts that do not exist yet.

PR #2202, opened 2020-01-30
services/horizon/expingest: Revert buildState removal, add tests

### LINKED ISSUE #2183: Reintroduce buildState which transitions to resume on success or back to init on failure
The redesigned ingestion has compressed the `init` and `buildState` into one. It will be easier to test and understand if kept the `buildState` node in the state machine. However, we should make sure that no transactions are shared between any two nodes in the state machine.

It's probably necessary that `buildState` checks the following invariants:

`(ingestVersion != CurrentVersion || lastIngestedLedger == 0) && (lastHistoryLedger == 0 || lastHistoryLedger == lastCheckpoint)`

If those conditions fail, then it should return to `init`. On success, it should progress to `resume`

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This commit is reverting removal of `buildState` state and adds initial tests (starting with `init` state tests).

Close #2183.

### Why

With separate state that build state from history archives it will be easier to test and understand the ingestion state machine. 

### Known limitations

I had to update `dbQ` interface and `mockQ` to be able to use it in tests. As you can see it's much larger now and looks awkward (because some methods were duplicated in two or more interfaces). It will also make testing state machine harder: instead of checking if the correct set of processors is executed we'd need to mock all the calls of the processors.

To solve the problems above I propose to move all methods in `processors.go` (ex. `runAllProcessorsOnLedger`) to a new struct that can be mocked in tests. Then while testing state machine we only test if the correct function of the new struct was called and we can revert `dbQ` interface to it's previous, simple form.

### DIFF
--- a/services/horizon/internal/db2/history/effect.go
+++ b/services/horizon/internal/db2/history/effect.go
@@ -218,8 +218,8 @@ func (q *EffectsQ) orderBookFilter(a xdr.Asset, prefix string) {
 
 // QEffects defines history_effects related queries.
 type QEffects interface {
+	QCreateAccountsHistory
 	NewEffectBatchInsertBuilder(maxBatchSize int) EffectBatchInsertBuilder
-	CreateAccounts(addresses []string) (map[string]int64, error)
 }
 
 var selectEffect = sq.Select("heff.*, hacc.address").
--- a/services/horizon/internal/expingest/main.go
+++ b/services/horizon/internal/expingest/main.go
@@ -492,7 +492,7 @@ func (s *System) buildState() (state, error) {
 	// Double check if we should proceed with state ingestion. It's possible that
 	// another ingesting instance will be redirected to this state from `init`
 	// but it's first to complete the task.
-	if !(ingestVersion != CurrentVersion || lastIngestedLedger == 0) {
+	if ingestVersion == CurrentVersion && lastIngestedLedger > 0 {
 		log.Info("Another instance completed `buildState`. Skipping...")
 		return state{systemState: initState}, nil
 	}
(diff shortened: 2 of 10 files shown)


services/horizon/internal/db2/history/effect.go
```diff
@@ -218,8 +218,8 @@ func (q *EffectsQ) orderBookFilter(a xdr.Asset, prefix string) {
 
 // QEffects defines history_effects related queries.
 type QEffects interface {
+	QCreateAccountsHistory
```
abuiles: 👍

services/horizon/internal/expingest/main.go
```diff
@@ -489,7 +466,102 @@ func (s *System) loadOffersIntoMemory() error {
...
+
+	ingestVersion, err := s.historyQ.GetExpIngestVersion()
+	if err != nil {
+		return state{systemState: initState}, errors.Wrap(err, "Error getting exp ingest version")
+	}
+
+	// Double check if we should proceed with state ingestion. It's possible that
+	// another ingesting instance will be redirected to this state from `init`
+	// but it's first to complete the task.
+	if !(ingestVersion != CurrentVersion || lastIngestedLedger == 0) {
```
tamirms: I think it's easier to understand if the condition is:

`ingestVersion == CurrentVersion && lastIngestedLedger > 0`

bartekn: Fixed in https://github.com/stellar/go/pull/2202/commits/2874b8bbb01d4eb72454da178a39aaab9118f1ed.

PR #2203, opened 2020-01-30
services/horizon/expingest: State machine tests

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This commit is adding tests to each state of ingestion state machine.

### Why

We need to add tests to increase test coverage of Horizon's ingestion system. While working on the tests several issues has been found:
* Off-by-one bug in `resume` state (see `TestIngestOrderbookOnlyWhenLastLedgerExpEqualsCurrent` test for more info).
* Transition to invalid state in `init` when last ingested ledger > 0 but latest history ledger = 0 (see `TestResumeStateBehindHistory0` for more details).
* `graph.Apply` was called with invalid value in `loadOffersIntoMemory`.
* Added the code to update cursor in stellar-core (and take the cursor name from config).
* Removed waiting between `resume` transitions when catching up (see `noSleep` state variable).
* Simplified returning error in `run()` method in `shutdown` state.
* Extracted `ProcessorsRunner` to simplify testing.


.vscode/launch.json
```diff
@@ -10,7 +10,19 @@
             "port": 2345,
             "host": "127.0.0.1",
             "program": "${workspaceRoot}/services/horizon/main.go",
-            "env": {},
+            "env": {
```
tamirms: I think this was mistakenly included in the PR

services/horizon/internal/expingest/processors_runner.go
```diff
@@ -18,63 +22,36 @@ const (
...
-type groupTransactionProcessors []horizonTransactionProcessor
-
-func (g groupTransactionProcessors) ProcessTransaction(tx io.LedgerTransaction) error {
-	for _, p := range g {
-		if err := p.ProcessTransaction(tx); err != nil {
-			return err
-		}
-	}
-	return nil
+type ProcessorsRunnerInterface interface {
```
tamirms: I think ProcessorRunnerInterface is a better name

services/horizon/internal/expingest/processors_runner.go
```diff
@@ -18,63 +22,36 @@ const (
...
-func (g groupTransactionProcessors) Commit() error {
-	for _, p := range g {
-		if err := p.Commit(); err != nil {
-			return err
-		}
-	}
-	return nil
+var _ ProcessorsRunnerInterface = (*ProcessorsRunner)(nil)
+
+type ProcessorsRunner struct {
```
tamirms: I prefer ProcessorRunner

services/horizon/internal/expingest/processors_runner.go
```diff
@@ -178,7 +153,7 @@ func (s *System) runHistoryArchiveIngestion(
 			s.historyArchive,
 			&io.MemoryTempSet{},
 			checkpointLedger,
-			s.maxStreamRetries,
+			s.config.MaxStreamRetries,
```
tamirms: maybe include `maxStreamRetries` in ProcessorsRunner so we can remove the dependency on config?

tamirms: nevermind, I see config is used in other places

.vscode/launch.json
abuiles: fixed

services/horizon/internal/expingest/main.go
```diff
@@ -899,6 +988,26 @@ func (s *System) resetStateVerificationErrors() {
 	s.stateVerificationErrors = 0
 }
 
+func (s *System) updateCursor(ledgerSequence uint32) error {
```
tamirms: what does this do?

bartekn: It updates the cursor in stellar-core with the latest ingested ledger sequence so stellar-core knows it's safe to remove the ledgers from it's DB. Without it, it's possible that core removes the ledger before it's processed by Horizon. See: https://www.stellar.org/developers/stellar-core/software/commands.html#http-commands

tamirms:
overall, looks good! I have two comments:

are there other places where we want to skip sleep? The transitions I have in mind are going from init to build, init to resume, init to ingestHistoryRange, build to resume, ingestHistoryRange to init (when there are no errors). It seems like we want to skip sleep if we transition states with no errors.

because it is possible to go back to init from resume, I think we should be clearing the orderbook graph (e.g. removing all offers) at the beginning of `loadOffersIntoMemory()` and `buildState()`. Otherwise there will be errors when applying orderbook graph batches.

bartekn:
> because it is possible to go back to init from resume, I think we should be clearing the orderbook graph (e.g. removing all offers) at the beginning of loadOffersIntoMemory() and buildState(). Otherwise there will be errors when applying orderbook graph batches.

Good catch! I added `Clear()` method that removes all offers from the graph and call it in `loadOffersIntoMemory`.

> are there other places where we want to skip sleep? The transitions I have in mind are going from init to build, init to resume, init to ingestHistoryRange, build to resume, ingestHistoryRange to init (when there are no errors). It seems like we want to skip sleep if we transition states with no errors.

I think we can do it but let's do it in a separate PR. My only worry is that in the future we'll introduce transition between two states without errors and we'll be flooded with the logs.

PR #2212, opened 2020-01-31
services/horizon/expingest: Fix ProcessorRunner init code

bartekn:
This commit fixes initialization code of `ProcessorRunner`. It also adds initial test to checj `NewSystem`.

### DIFF
--- a/services/horizon/internal/expingest/main_test.go
+++ b/services/horizon/internal/expingest/main_test.go
@@ -97,6 +97,7 @@ func TestNewSystem(t *testing.T) {
 
 	assert.Equal(t, config, system.runner.(*ProcessorRunner).config)
 	assert.Equal(t, config.OrderBookGraph, system.runner.(*ProcessorRunner).graph)
+	assert.Equal(t, system.ctx, system.runner.(*ProcessorRunner).ctx)
 }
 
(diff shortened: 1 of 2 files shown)


services/horizon/internal/expingest/main_test.go
```diff
@@ -69,6 +70,35 @@ func TestCheckVerifyStateVersion(t *testing.T) {
...
+	system, err := NewSystem(config)
+	assert.NoError(t, err)
+
+	assert.Equal(t, config, system.config)
+	assert.Equal(t, config.OrderBookGraph, system.graph)
+	assert.Equal(t, config.DisableStateVerification, system.disableStateVerification)
+	assert.Equal(t, config.MaxStreamRetries, system.maxStreamRetries)
+
+	assert.Equal(t, config, system.runner.(*ProcessorRunner).config)
+	assert.Equal(t, config.OrderBookGraph, system.runner.(*ProcessorRunner).graph)
```
tamirms: I think it's also worth checking the context is the same as well

bartekn: Added! We should spend more time on testing this function and `ProcessorRunner`.

Issue #2204, opened 2020-01-30
Possible failure scenarios in distributed ingestion during rolling deployments


tamirms:
In a rolling deployment, if you are currently running horizon at version N and you want to upgrade to version N+1, there is a period where both versions of horizon will be running (and sharing the same database). The workflow for rolling deployments is that you keep the old version running, then you spin up an instance with the new version. Once that new instance is healthy, you shut down the old instance. You continue doing this until there are no more instances running the old version.

If the horizon build which is getting deployed does **not** bump the ingestion version, everything should be ok.

In this scenario, the existing horizon instances cycle through the [resume state](https://github.com/stellar/go/blob/74923441559d8c6a02199535ab98c395b316e272/services/horizon/internal/expingest/main.go#L560). The new horizon instance starts at the [init state](https://github.com/stellar/go/blob/74923441559d8c6a02199535ab98c395b316e272/services/horizon/internal/expingest/main.go#L316). From the init state, the new horizon instance will progress to the resume state because `ingestVersion == CurrentVersion` and `lastHistoryLedger == lastIngestedLedger`. This is fine because only one node is able to run ingestion while the others are blocked on the select for update lock. Furthermore, both the old and the new horizon instances are on the same ingestion version.

Let's consider the case where the horizon build which is getting deployed bumps the ingestion version.

In this case, the new horizon instance will transition to [buildState](https://github.com/stellar/go/blob/74923441559d8c6a02199535ab98c395b316e272/services/horizon/internal/expingest/main.go#L469) from init. Once the new horizon instance acquires the select for update lock, it will truncate some ingestion tables and start ingesting state from the beginning using the history archive.

It is possible that the old horizon instances will continue running in the resume state after they acquire the select for update lock from the new horizon instances. If this happens, it may result in weird behavior because both the old and the new ingestion versions will be running in an alternating fashion.

For example, the new horizon instances can ingest ledger N and N+1 and the old horizon instances can ingest ledger N+2. Thus the database will contain a mix of both versions.

Another scenario which can happen is that the new horizon instances and old horizon instances will run buildState repeatedly. In this case, the database will not contain a mix of both versions. However, buildState is intensive in terms of performance (for both the horizon node and the postgres db). So, it won't be ideal to have all the horizon instances thrashing in buildState.

tamirms:
fixed in https://github.com/stellar/go/pull/2220

PR #2213, opened 2020-01-31
services/horizon/internal/expingest: Add missing invariant assertions in the beginning of resume and init

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Fixes https://github.com/stellar/go/issues/2204

* Add missing invariant assertions in the beginning of resume and init

### Why

https://github.com/stellar/go/issues/2204 provides some details on scenarios where we can have multiple versions of ingestion running at the same time during a rolling deployment.

If we have checks against the ingestion version found in the db during the init and resume state we can detect these cases and shutdown.

### Known limitations

[N/A]

### DIFF
--- a/services/horizon/internal/expingest/main.go
+++ b/services/horizon/internal/expingest/main.go
@@ -354,7 +354,7 @@ func (s *System) init() (state, error) {
 	if ingestVersion > CurrentVersion {
 		return state{systemState: shutdownState},
 			errors.Errorf(
-				"ingestion version in db %v is greater than current version %v",
+				"ingestion version in db %d is greater than current version %d",
 				ingestVersion,
 				CurrentVersion,
 			)
--- a/services/horizon/internal/expingest/main.go
+++ b/services/horizon/internal/expingest/main.go
@@ -351,6 +351,15 @@ func (s *System) init() (state, error) {
 		return state{systemState: initState}, errors.Wrap(err, getExpIngestVersionErrMsg)
 	}
 
+	if ingestVersion > CurrentVersion {
+		return state{systemState: shutdownState},
+			errors.Errorf(
+				"ingestion version in db %v is greater than current version %v",
+				ingestVersion,
+				CurrentVersion,
+			)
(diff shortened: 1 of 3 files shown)


services/horizon/internal/expingest/main.go
```diff
@@ -623,6 +632,35 @@ func (s *System) resume() (state, error) {
...
+	}
+
+	if lastHistoryLedger != 0 && lastHistoryLedger != lastIngestedLedger {
+		return state{systemState: initState},
+			errors.Errorf(
+				"last history ledger %v does not match last ingested ledger %v",
+				lastHistoryLedger,
+				lastIngestedLedger,
+			)
+	}
```
bartekn: Can you explain why is it necessary to check?

tamirms: the init state ensures that either `lastHistoryLedger == 0` or `lastHistoryLedger == lastIngestedLedger` before the system can transition to `resume`. So this check makes that assumption explicit. If the condition does not hold it means something is wrong.

tamirms:
> We probably need to lock access to order book graph if version mismatch is detected.

if the version does not match the ingestion system will stop running the resume state and it will go back to init. if the version is less than the current version it will rebuild state otherwise it will shut down. So, in both cases ingestion will stop adding offers to the orderbook graph until the db version is consistent with the current version

bartekn:
Sorry, my message wasn't clear. I was thinking about API access to `/order_book` and `/paths`.

tamirms:
> Sorry, my message wasn't clear. I was thinking about API access to /order_book and /paths.

I think the ingestion status check in the StateMiddleware will ensure that requests to the orderbook will fail: 

https://github.com/stellar/go/blob/release-horizon-v0.25.0/services/horizon/internal/middleware.go#L274-L278

tamirms:
closed in favor of https://github.com/stellar/go/pull/2220

Issue #1969, opened 2019-11-21
services/horizon/expingest: Rewrite System to a state machine


bartekn:
`expingest.System` code is still easy to understand but after adding history ingestion code it will become much harder to follow and test.

I think we should rewrite `System` to a state machine. This will make it much easier to understand, test and will remove a need for `retry` functions. Each state will have it's own `System` method that will perform actions and change the state based on the action code and input params. The initial design can be found below.

### States and transitions

State can be one of the following:
* `Init` - Initial state that determines the next state based on the input values, like DB state, ingestion version, if current node is a leader or not, etc.
* `LoadStateInMemory` - Loads all the data that needs to be available in memory, like offers graph.
* `WaitForCheckpoint` - Waits for the next checkpoint in history archive. This supports one specific case when the last ledger in history table is ahead of history archive latest.
* `Catchup` - Runs `RangeSession` (https://github.com/stellar/go/pull/1890) to ingest history data up to the latest checkpoint in history archive.
* `BuildStateAndResume` - Clears state tables and runs `LiveSession` starting on a given ledger sequence, then continues ingesting using ledger backend.
* `Resume` - Resumes ingestion from the last ingested ledger.
* `Shutdown` - Does nothing, breaks main loop.

The main function that runs the machine will look like this:
```go
func run() {
	for {
		var err error
		var nextState State

		switch currentState {
		case Init:
			nextState, err := init()
		case LoadState:
			nextState, err := loadState()
		// ...
		case Shutdown:
			log.Info("Ingestion shut down")
			return
		default:
			panic("Unknown state!")
		}

		if err != nil {
			log.Error(err)
			time.Sleep(time.Second)
		}

		currentState = nextState
	}
}
```
#### `Init`

Here's a first draft of `Init` transition function:
```go
func init() (State, err) {
	// Here or as function params:
	ingestVersion := s.historyQ.GetExpIngestVersion()
	lastIngestedLedger := s.historyQ.GetLastLedgerExpIngest()
	lastHistoryLedger := s.historyQ.GetLastHistoryLedger()
	lastCheckpoint := historyArchive.GetLastCheckpoint()

	if lastHistoryLedger == 0 {
		// No ledgers in history_ledgers so start from the latest checkpoint and
		// continue using ledgerbackend.
		return BuildStateAndResume, nil
	}

	// Now: lastHistoryLedger != 0 => there are ledgers in history_ledgers

	if ingestVersion != CurrentVersion {
		// Ingestion version upgrade.
[...]

PR #2220, opened 2020-02-04
services/horizon/internal/expingest: Refactor ingestion system to use explicit FSM

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

The ingestion system uses a finite state machine to model its behaviors (see https://github.com/stellar/go/issues/1969 ). In the current implementation the state machine is represented by the following struct:

```golang
type systemState string

const (
	initState               systemState = "init"
	ingestHistoryRangeState systemState = "ingestHistoryRange"
	waitForCheckpointState  systemState = "waitForCheckpoint"
	buildStateState         systemState = "buildState"
	resumeState             systemState = "resume"
	verifyRangeState        systemState = "verifyRange"
	shutdownState           systemState = "shutdown"
)

type state struct {
	systemState                       systemState
	latestSuccessfullyProcessedLedger uint32

	checkpointLedger uint32

	rangeFromLedger   uint32
	rangeToLedger     uint32
	rangeVerifyState  bool
	rangeClearHistory bool

	shutdownWhenDone bool

	// noSleep informs state machine to not sleep between state transitions
	noSleep bool
}
```
This PR replaces this representation with following interface:

```golang
type stateMachineNode interface {
	run(*System) (transition, error)
}

type transition struct {
	node          stateMachineNode
	sleepDuration time.Duration
}
```

Every node in the state machine is represented by a struct which implements the `stateMachineNode` interface.

Note this PR is rebased on top of https://github.com/stellar/go/pull/2213 . Please ignore the commits belonging to https://github.com/stellar/go/pull/2213  when reviewing this PR.

Also, since this PR moves existing functions to a new file, it is easier to review this PR by reviewing each commit individually instead of looking at the overall diff.

### Why

There are two advantages to this approach:

* All the state machine structs are immutable. The structs do not need to mutate their own fields (none of the structs have functions with pointer receivers) or fields in `System` .

* The parameters for each node in the state machine are isolated. The previous implementation kept the parameters for all states in one `state` struct. In the old implementation it is possible for a state function to mistakenly use parameters that are not relevant to the sate.

### Known limitations

Having each state be represented by a separate struct adds some boilerplate. If there isn't consensus that this refactor is an improvement over the old version I will close the PR.

### DIFF
--- a/services/horizon/internal/expingest/build_state_test.go
+++ b/services/horizon/internal/expingest/build_state_test.go
@@ -36,7 +36,6 @@ func (s *BuildStateTestSuite) SetupTest() {
 	s.lastLedger = 0
 	s.system = &System{
 		ctx:               context.Background(),
-		state:             state{systemState: buildStateState, checkpointLedger: s.checkpointLedger},
 		historyQ:          s.historyQ,
 		historyAdapter:    s.historyAdapter,
 		graph:             s.graph,
@@ -44,8 +43,6 @@ func (s *BuildStateTestSuite) SetupTest() {
 		stellarCoreClient: s.stellarCoreClient,
 	}
 
-	s.Assert().Equal(buildStateState, s.system.state.systemState)
-	s.historyQ.On("GetTx").Return(nil).Once()
 	s.historyQ.On("Begin").Return(nil).Once()
 	s.historyQ.On("Rollback").Return(nil).Once()
 	s.graph.On("Discard").Once()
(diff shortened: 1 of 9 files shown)


services/horizon/internal/expingest/build_state_test.go
```diff
@@ -257,11 +250,11 @@ func (s *BuildStateTestSuite) TestUpdateCommitReturnsError() {
 		Return(errors.New("my error")).
 		Once()
 	s.graph.On("Clear").Return(nil).Once()
-	nextState, err := s.system.runCurrentState()
+	next, err := buildState{checkpointLedger: s.checkpointLedger}.run(s.system)
 
 	s.Assert().Error(err)
 	s.Assert().EqualError(err, "Error committing db transaction: my error")
-	s.Assert().Equal(initState, nextState.systemState)
+	s.Assert().Equal(startState{}, next.node)
```
bartekn: Can you also add tests for `sleepDuration` in all tests? Should be easy and will help ensure that we sleep when error happens.

services/horizon/internal/expingest/fsm.go
```diff
@@ -0,0 +1,701 @@
...
+	if err != nil {
+		return start(), errors.Wrap(err, getLastIngestedErrMsg)
+	}
+
+	ingestVersion, err := s.historyQ.GetExpIngestVersion()
+	if err != nil {
+		return start(), errors.Wrap(err, getExpIngestVersionErrMsg)
+	}
+
+	if ingestVersion > CurrentVersion {
```
bartekn: I think the code from #2213 is included here. Maybe let's close it and we can continue discussion here?

services/horizon/internal/expingest/fsm.go
```diff
@@ -0,0 +1,701 @@
...
+		return terminate(), errors.Errorf(
+			"ingestion version in db %v is greater than current version %v",
+			ingestVersion,
+			CurrentVersion,
+		)
```
bartekn: I wonder if we should ensure this is not logged with `ERROR` level. This is expected when doing the rolling deployment.

services/horizon/internal/expingest/fsm.go
```diff
@@ -0,0 +1,701 @@
...
+	// If successful, continue from the next ledger
+	return resume(b.checkpointLedger), nil
+}
+
+type resumeState struct {
+	latestSuccessfullyProcessedLedger uint32
+}
+
+func (r resumeState) String() string {
+	return fmt.Sprintf("resume(%d)", r.latestSuccessfullyProcessedLedger)
```
bartekn:
```suggestion
	return fmt.Sprintf("resume(latestSuccessfullyProcessedLedger=%d)", r.latestSuccessfullyProcessedLedger)
```

services/horizon/internal/expingest/verify.go
```diff
@@ -84,20 +87,17 @@ func (s *System) verifyState(graphOffers map[xdr.Int64]xdr.OfferEntry) error {
...
-	historyLatestSequence, err := s.historyAdapter.GetLatestLedgerSequence()
-	if err != nil {
-		return errors.Wrap(err, "Error getting the latest ledger sequence")
-	}
 
-	if s.state.systemState == verifyRangeState {
-		if ledgerSequence != s.state.rangeToLedger {
-			localLog.Info("Current ledger is not the last ledger in the range. Cancelling...")
-			return nil
+	if verifyAgainstLatestCheckpoint {
```
bartekn: This can be removed. The first version of verify-range was doing checking this in post-processing hook and now we call `System.verifyRange` directly.

services/horizon/internal/expingest/fsm.go
```diff
@@ -0,0 +1,700 @@
...
+		// lastHistoryLedger. Otherwise, some other horizon node has
+		// already completed the ingest history range operation and
+		// we should go back to the init state
+		if lastHistoryLedger != h.fromLedger-1 {
+			return next, nil
+		}
+	}
+
+	for cur := h.fromLedger; cur <= h.toLedger; cur++ {
+		log.WithField("sequence", cur).Info("Processing ledger")
```
bartekn: While we're at it.
```suggestion
		log.WithFields(logpkg.F{
			"sequence": cur,
			"state":    false,
			"ledger":   true,
			"graph":    false,
			"commit":   false,
		}).Info("Processing ledger")
```

services/horizon/internal/expingest/fsm.go
```diff
@@ -0,0 +1,700 @@
...
+		return stop(), err
+	}
+
+	log.WithFields(logpkg.F{
+		"ledger":   v.fromLedger,
+		"duration": time.Since(startTime).Seconds(),
+	}).Info("Processed state")
+
+	for sequence := v.fromLedger + 1; sequence <= v.toLedger; sequence++ {
+		log.WithField("sequence", sequence).Info("Processing ledger")
```
bartekn:
```suggestion
		log.WithFields(logpkg.F{
			"sequence": sequence,
			"state":    true,
			"ledger":   true,
			"graph":    true,
			"commit":   true,
		}).Info("Processed ledger")
```

PR #2221, opened 2020-02-04
exp/ingest: Use errors.Wrap to propagate errors

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

There are several places in exp/ingest/adapters/history_archive_adapter.go and exp/ingest/io/single_ledger_state_reader.go where we forget to use `errors.Wrap()` to propagate errors.

It's important to wrap errors because the ingestion system uses `errors.Cause()` to check if an error was caused by a shutdown. This check is used to determine whether an error should be logged.

### DIFF
--- a/exp/ingest/io/single_ledger_state_reader.go
+++ b/exp/ingest/io/single_ledger_state_reader.go
@@ -3,7 +3,6 @@ package io
 import (
 	"context"
 	"encoding/base64"
-	"fmt"
 	"io"
 	"sync"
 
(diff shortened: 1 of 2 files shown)


exp/ingest/io/single_ledger_state_reader.go
```diff
@@ -80,7 +79,7 @@ func MakeSingleLedgerStateReader(
 ) (*SingleLedgerStateReader, error) {
 	has, err := archive.GetCheckpointHAS(sequence)
 	if err != nil {
-		return nil, fmt.Errorf("unable to get checkpoint HAS at ledger sequence %d: %s", sequence, err)
+		return nil, errors.Wrapf(err, "unable to get checkpoint HAS at ledger sequence %d", sequence)
```
bartekn: TIL! 👍  I've been using `errors.Wrap(err, fmt.Sprintf(...))`.

PR #2227, opened 2020-02-05
tools/horizon-cmp: Add history mode

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This commit adds `horizon-cmp history` command with the following params:
```
compares history endpoints for a given range of ledgers

Usage:
  horizon-cmp history [flags]

Flags:
      --count uint32   number of last ledgers to check (if from/to not set)
      --from uint32    start of the range
      --to uint32      end of the range

Global Flags:
  -b, --base string   URL of the base/old version Horizon server
  -t, --test string   URL of the test/new version Horizon server
```

### Why

The existing options don't allow easy testing of history endpoints only.

### Known limitations

Currently all paths are fetched twice (to build a list of paths and then to compare responses). Changing this requires a larger refactoring and should be done in a separate PR.


tools/horizon-cmp/history.go
```diff
@@ -0,0 +1,147 @@
...
+			status = "diff"
+		}
+
+		log = log.WithFields(slog.F{
+			"status_code": a.StatusCode,
+			"size_base":   a.Size(),
+			"size_test":   b.Size(),
+		})
+
+		if status == "diff" {
```
tamirms: it seems that `status` is not used for anything so you can inline `a.Equal(b)` into the if statement

Issue #2014, opened 2019-12-05
Horizon v1.0.0 Compatibility 


abuiles:
The upcoming Horizon release is coming, and there are multiple breaking changes plus new features 🎉🎉🎉!

The following are the list of changes required to support this new release:

- [x] ➕Update `/fee_stats` response.
    <details>

     - ✂ Remove the following fields:

       ```json
       min_accepted_fee
       mode_accepted_fee
       p10_accepted_fee
       p20_accepted_fee
       p30_accepted_fee
       p40_accepted_fee
       p50_accepted_fee
       p60_accepted_fee
       p70_accepted_fee
       p80_accepted_fee
       p90_accepted_fee
       p95_accepted_fee
       p99_accepted_fee
       ```

     - ➕Add support for `max_fee` and `fee_charged` fields. Each field contains a JSON object that looks like this:

       ```json
       {
         "last_ledger": "22606298",
         "last_ledger_base_fee": "100",
         "ledger_capacity_usage": "0.97",
         "fee_charged": {
           "max": "100",
           "min": "100",
           "mode": "100",
           "p10": "100",
           "p20": "100",
           "p30": "100",
           "p40": "100",
           "p50": "100",
           "p60": "100",
           "p70": "100",
           "p80": "100",
           "p90": "100",
           "p95": "100",
           "p99": "100"
         },
         "max_fee": {
           "max": "100000",
           "min": "100",
           "mode": "100",
           "p10": "100",
           "p20": "100",
           "p30": "100",
           "p40": "100",
           "p50": "100",
           "p60": "100",
           "p70": "100",
           "p80": "100",
           "p90": "15000",
           "p95": "100000",
           "p99": "100000"
         }
        }
        ```

     See https://github.com/stellar/go/pull/2140 for more info.
    </details>

- [x] ✂ Remove deprecated `fee_paid` field from Transaction response (replace it with the `fee_charged` and `max_fee` fields, see [#1372](https://github.com/stellar/go/issues/1372)).

- [x] 🚨Multiple fields changed from `int64` to `string`. If your SDK has types, they need to be updated. If possible, we recommend implementing backward compatibility support, as we did [here](https://github.com/stellar/go/blob/master/protocols/horizon/main.go#L258).
    <details>

[...]

PR #2229, opened 2020-02-05
clients/horizonclient: Add support for Accounts endpoint.

abuiles:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Extend the horizonclient to support accounts filtering. It allows
account retrival by signer or with a trustline to an asset.

```go
	client := horizonclient.DefaultPublicNetClient
        accountID := "GCLWGQPMKXQSPF776IU33AH4PZNOOWNAWGGKVTBQMIC5IMKUNP3E6NVU"
	accountsRequest := horizonclient.AccountsRequest{Signer: accountID}

	account, err := client.Accounts(accountsRequest)
	if err != nil {
		fmt.Println(err)
		return
	}

	fmt.Print(account)
```

### Why

This is part of the work in #2014, which is required to have full compatibility with Horizon 1.0.

### Known limitations

[TODO or N/A]

### DIFF
--- a/clients/horizonclient/client.go
+++ b/clients/horizonclient/client.go
@@ -226,6 +226,17 @@ func (c *Client) HorizonTimeOut() time.Duration {
 	return c.horizonTimeOut
 }
 
+// Accounts returns accounts who have a given signer or
+// have a trustline to an asset.
+// See https://www.stellar.org/developers/horizon/reference/endpoints/accounts.html
+func (c *Client) Accounts(request AccountsRequest) (accounts hProtocol.AccountsPage, err error) {
+	err = c.sendRequest(request, &accounts)
+	if err != nil {
+		return
+	}
+	return
+}
+
 // AccountDetail returns information for a single account.
 // See https://www.stellar.org/developers/horizon/reference/endpoints/accounts-single.html
 func (c *Client) AccountDetail(request AccountRequest) (account hProtocol.Account, err error) {
--- a/clients/horizonclient/client.go
+++ b/clients/horizonclient/client.go
@@ -231,9 +231,6 @@ func (c *Client) HorizonTimeOut() time.Duration {
 // See https://www.stellar.org/developers/horizon/reference/endpoints/accounts.html
 func (c *Client) Accounts(request AccountsRequest) (accounts hProtocol.AccountsPage, err error) {
 	err = c.sendRequest(request, &accounts)
-	if err != nil {
-		return
-	}
 	return
 }
 
--- a/clients/horizonclient/main.go
+++ b/clients/horizonclient/main.go
@@ -141,6 +141,7 @@ type Client struct {
 
 // ClientInterface contains methods implemented by the horizon client
 type ClientInterface interface {
+	Accounts(request AccountsRequest) (hProtocol.AccountsPage, error)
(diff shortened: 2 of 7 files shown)


clients/horizonclient/client.go
```diff
@@ -226,6 +226,17 @@ func (c *Client) HorizonTimeOut() time.Duration {
 	return c.horizonTimeOut
 }
 
+// Accounts returns accounts who have a given signer or
+// have a trustline to an asset.
+// See https://www.stellar.org/developers/horizon/reference/endpoints/accounts.html
+func (c *Client) Accounts(request AccountsRequest) (accounts hProtocol.AccountsPage, err error) {
+	err = c.sendRequest(request, &accounts)
+	if err != nil {
```
ire-and-curses: Seems like this `if` is redundant - we return `err` either way

clients/horizonclient/main.go
```diff
@@ -212,7 +213,18 @@ type HorizonRequest interface {
 	BuildURL() (string, error)
 }
 
-// AccountRequest struct contains data for making requests to the accounts endpoint of a horizon server.
+// AccountsRequest struct contains data for making requests to the accounts endpoint of a horizon server.
+// "Signer" or "Asset" fields should be set when retrieving AccountsData.
```
ire-and-curses:
```suggestion
// Either "Signer" or "Asset" fields should be set when retrieving AccountsData.
```

clients/horizonclient/main.go
```diff
@@ -212,7 +213,18 @@ type HorizonRequest interface {
...
+// At the moment, you can't use both filters are the same time.
+type AccountsRequest struct {
+	Signer string
+	Asset  string
+	Order  Order
+	Cursor string
+	Limit  uint
+}
+
+// AccountRequest struct contains data for making requests to the show account endpoint of a horizon server.
```
ire-and-curses: It doesn't seem great that we have two structs with very similar names `AccountsRequest` vs `AccountRequest`. Is there something we can do to name them more clearly?

ire-and-curses: (I don't really like the plural name AccountsRequest`)

clients/horizonclient/client.go
abuiles: good catch!

clients/horizonclient/main.go
abuiles: @ire-and-curses I agree with you, initially I thought about following a similar approach for `EffectRequest`, but it seems to me that we are trying to do too much with those structs.

That said though, we could be consistent and try to follow a similar approach. Basically extend the `AccountRequest` struct to also take `ForSigner` and `ForAsset`. When called through `client.Accounts` they will be the only fields we look at + Page params, ignoring the rest. 

I personally don't  like that interface, but it is consistent with how the other `*Request` like structs behave.

leighmcculloch: I think this is okay given that we can't make it more consistent with other functions without it being backwards compatible, and the alternatives of smooshing the two types together into a single shared types would be worse and could result in bad human errors. The compiler will make sure the caller uses the right type so whilst there might be some confusion with similar named types any errors arising from that should be prevented.

Issue #2231, opened 2020-02-06
Running `db reingest range` blocks main ingestion process


bartekn:
### What version are you using?

Horizon 1.0.0-beta

### What did you do?

Run `horizon db reingest range x y`.

### What did you expect to see?

Horizon reingests selected range, main ingestion proceeds.

### What did you see instead?

Main ingestion blocked during reingestion process.

### Comment

When user runs `db reingest range` the state machine of the ingestion process executes `historyRangeState` that always acquires the ingestion lock on ledger sequence value. We should update the state data to:
```go
type historyRangeState struct {
	fromLedger       uint32
	toLedger         uint32
	reingesting      bool
}
```
If `reingesting` is `true`, it should:
* not acquire the lock,
* clear history range,
* go to `stopState` when done.

Additionally, we should probably have an extra check to ensure user does not reingest the latest range - this can cause some issues in the main ingestion process.

tamirms:
fixed in https://github.com/stellar/go/pull/2242

PR #2242, opened 2020-02-07
services/horizon/internal/expingest: Allow horizon db reingest range to run concurrently with Horizon

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

* Separate `horizon db reingest range` code from `historyRangeState`
* Add a force flag to the `horizon db reingest range` command

Fixes https://github.com/stellar/go/issues/2231

### Why

Currently when you run `horizon db reingest range` horizon ingestion is effectively  paused until the command completes. This happens because all entry points into the ingestion system begin by acquiring a select for update lock. This lock prevents multiple actors from ingesting the same data at the same time. 

This PR modifies the behavior of `horizon db reingest range` so that it can run without the select for update lock. The reingest command is allowed to run ingestion without the lock on the condition that the last ingested ledger by horizon is either 0 or ahead of the range we want to reingest. We don't want the reingest range to overlap with horizon because in that case the reingest command and horizon will attempt to insert the same data and it will cause one of the two components to fail.

Note: there is a force flag which allows you to run the reingest command on a range which overlaps with horizon. When the force flag is applied, the reingest command will block horizon ingestion by acquiring the select for update lock.

 
### Known limitations

[N/A]

### DIFF
--- a/services/horizon/internal/expingest/fsm.go
+++ b/services/horizon/internal/expingest/fsm.go
@@ -9,6 +9,13 @@ import (
 	logpkg "github.com/stellar/go/support/log"
 )
 
+var (
+	defaultSleep = time.Second
+	// ErrReingestRangeConflict indicates that the reingest range overlaps with
+	// horizon's most recently ingested ledger
+	ErrReingestRangeConflict = errors.New("reingest range overlaps with horizon ingestion")
+)
+
 type stateMachineNode interface {
 	run(*System) (transition, error)
 	String() string
@@ -18,8 +25,6 @@ type transition struct {
 	node          stateMachineNode
 	sleepDuration time.Duration
 }
-
-var defaultSleep = time.Second
 
 func stop() transition {
 	return transition{node: stopState{}, sleepDuration: 0}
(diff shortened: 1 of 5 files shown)


services/horizon/internal/expingest/fsm.go
```diff
@@ -467,84 +472,61 @@ func (r resumeState) run(s *System) (transition, error) {
...
 
-		err = s.historyQ.DeleteRangeAll(start, end)
-		if err != nil {
-			return next, errors.Wrap(err, "error in DeleteRangeAll")
-		}
+	// We should be ingesting the ledger which occurs after
+	// lastHistoryLedger. Otherwise, some other horizon node has
+	// already completed the ingest history range operation and
+	// we should go back to the init state
+	if lastHistoryLedger != h.fromLedger-1 {
```
abuiles: @tamirms is the comment here correct? it seems like we are not checking that the ledger occurs after lastHistoryLedger but rather that is it different

tamirms: We should be ingesting the ledger which occurs *right after* lastHistoryLedger, or, in other words, we expect `fromLedger` to be equal to `lastHistoryLedger` + 1. does that make sense?

abuiles: sorry, I got confused by `start()`, I forgot we change that to return the start state -- all good :)

services/horizon/internal/expingest/fsm.go
```diff
@@ -568,12 +549,75 @@ func (h historyRangeState) run(s *System) (transition, error) {
...
+	if err := s.historyQ.Begin(); err != nil {
+		return stop(), errors.Wrap(err, "Error starting a transaction")
+	}
+	defer s.historyQ.Rollback()
+
+	if h.force {
+		// acquire distributed lock so no one else can perform ingestion operations.
+		if _, err := s.historyQ.GetLastLedgerExpIngest(); err != nil {
+			return stop(), errors.Wrap(err, getLastIngestedErrMsg)
+		}
```
bartekn: It makes sense but we shouldn't allow reingesting any ledger sequence >= last ingested ledger. We don't update last ingested ledger in `ingestHistoryRange` (correct) so when main ingestion is unblocked it will result in unique constraint violation type of errors because it will attempt to ingest history again.

tamirms: @bartekn how can that happen?

in history range you have the following code which protects against the unique constraint error you're describing:

```golang
	if validateStartLedger {
		lastHistoryLedger, err := s.historyQ.GetLatestLedger()
		if err != nil {
			return next, errors.Wrap(err, "could not get latest history ledger")
		}

		// We should be ingesting the ledger which occurs after
		// lastHistoryLedger. Otherwise, some other horizon node has
		// already completed the ingest history range operation and
		// we should go back to the init state
		if lastHistoryLedger != h.fromLedger-1 {
			return next, nil
		}
	}
```

In the resume state there's a similar check:

```golang
	lastHistoryLedger, err := s.historyQ.GetLatestLedger()
	if err != nil {
		return retryResume(r), errors.Wrap(err, "could not get latest history ledger")
	}

	if lastHistoryLedger != 0 && lastHistoryLedger != lastIngestedLedger {
		log.WithFields(logpkg.F{
			"lastHistoryLedger":  lastHistoryLedger,
			"lastIngestedLedger": lastIngestedLedger,
		}).Info(
			"last history ledger does not match last ingested ledger, " +
				"going back to start state",
		)
		return start(), nil
	}
```

bartekn: Oh, I forgot about these new checks. It seems that in case of running reingest on a range that ends after the latest ledger it will always rebuild state so state endpoint may not be available for some time but it's probably fine 👍.

PR #2244, opened 2020-02-07
support/http/httpdecode: add decoding of querys

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Add decoding of queries from requests into structs to the `httpdecode` package in a new `DecodeQuery` function and into the existing `Decode` function.

### Why
We use httpdecode to decode requests into structs and this works really well for JSON and form bodys, but we have the odd endpoint that also has request parameters provided via the query string. In Horizon internals we have some code that works similar to this and pulls out the query parameters and upgrades them into the struct alongside other body parameters.

This change pulls some of that concept into a place we can use in more apps and maybe if it works well we can take this code back into Horizon via using `httpdecode.Decode` in the `support/render/httpjson` package.

This PR is a step towards that but that benefit won't be realized yet.

### Known limitations

N/A

### DIFF
--- a/support/http/httpdecode/httpdecode.go
+++ b/support/http/httpdecode/httpdecode.go
@@ -61,6 +61,7 @@ func DecodeForm(r *http.Request, v interface{}) error {
 }
 
 // Decode decodes form URL encoded requests and JSON requests from r into v.
+// Also decodes query parameters.
 //
 // The requests Content-Type header informs if the request should be decoded
 // using a form URL encoded decoder or using a JSON decoder.
@@ -74,8 +75,8 @@ func DecodeForm(r *http.Request, v interface{}) error {
 // An error is returned if the Content-Type cannot be parsed by a mime
 // media-type parser.
 //
-// See DecodeForm and DecodeJSON for details about the types of errors that may
-// occur.
+// See DecodeQuery, DecodeForm and DecodeJSON for details about the types of
+// errors that may occur.
 func Decode(r *http.Request, v interface{}) error {
 	err := DecodeQuery(r, v)
 	if err != nil {
--- a/support/http/httpdecode/httpdecode.go
+++ b/support/http/httpdecode/httpdecode.go
@@ -9,6 +9,14 @@ import (
 	"github.com/stellar/go/support/errors"
 )
 
+// DecodeQuery decodes the query string from r into v.
+func DecodeQuery(r *http.Request, v interface{}) error {
+	dec := schema.NewDecoder()
+	dec.SetAliasTag("query")
+	dec.IgnoreUnknownKeys(true)
+	return dec.Decode(v, r.URL.Query())
+}
+
 // DecodeJSON decodes JSON request from r into v.
 func DecodeJSON(r *http.Request, v interface{}) error {
 	dec := json.NewDecoder(r.Body)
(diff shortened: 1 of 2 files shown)


support/http/httpdecode/httpdecode.go
```diff
@@ -69,6 +77,10 @@ func DecodeForm(r *http.Request, v interface{}) error {
 // See DecodeForm and DecodeJSON for details about the types of errors that may
 // occur.
```
howardtw: We might want add `DecodeQuery` here as well. In addition, maybe it's worth adding a comment saying `DecodeJSON` and `DecodeForm` will take precedence over `DecodeQuery`.

leighmcculloch: Good call, d80ffc850a093f0fa3d16502fe0ddfda2c54599e.

leighmcculloch:
>I'm not sure whether it's good to take values from both URL and the request body at the same time. It might be more clear if we do one or another at once.

@howardtw I agree in API design having parameters in one place is ideal, but these functions aren't dictating API design, they're just providing some building tools. I assume the part you have an issue with is that `DecodeQuery` is included in the `Decode` function. The motivation for including `DecodeQuery` inside `Decode` is to step towards us being able to use this logic in `support/render/httpjson` where an application provides a request type, like a struct with tags, to the `httpjson` framework and for the framework can unpack query parameters or body parameters into the type as specified by the tags on the type. So likely our structs will only have query tags, or only have json tags, and unlikely have both, but this tool isn't dictating that, it's just providing us the ability to do either.

Can you elaborate if you see a specific issue?

howardtw:
Even though embedding `DecodeQuery` inside `Decode` gives the client more flexibility on how it can unpack the incoming params, it also the door for getting unwanted param values. In [one](https://github.com/stellar/go/pull/2244/files#diff-f910299d678c64e65002773a137adb09R304) of the test cases, we do have a field having two tags. I agree having multiple tags in one field  is unlikely to happen in production, but in this case should we just not allow this to happen? IMO `Decode` function can still be useful if we make it do one thing at once. Also, what is the behavior if the client provides a struct that doesn't have any of the three tags (`json`, `query`, `form`) in its fields?

leighmcculloch:
>having multiple tags in one field  is unlikely to happen in production, but in this case should we just not allow this to happen?

I still think this is up to the designer of the API. There are APIs out there in the world that use both, even maybe we will want to in the future. If it involves writing more code or making this code more complex to prevent a user from doing it, I think we should stick with less code and they're unlikely to do it anyway. How do you see us designing this and integrating this with `support/render/httpjson` in a way that prevents using one and not the other?

>what is the behavior if the client provides a struct that doesn't have any of the three tags (`json`, `query`, `form`) in its fields?

The body will be parsed for JSON as is the existing behavior of `Decode`. No fields will be populated based on query params, or form fields. Those integrations will only parse into struct fields that explicitly include their respective tags.

JSON is opt-out. Everything else is opt-in. Separate to this PR I'll add some tests and comments that make this clear.

howardtw:
> There are APIs out there in the world that use both, even maybe we will want to in the future.

Interesting. It's good to know.

> How do you see us designing this and integrating this with support/render/httpjson in a way that prevents using one and not the other?

Hmm 🤔 if the goal is to simplify this [block](https://github.com/stellar/go/blob/master/support/render/httpjson/handler.go#L84-L93) of code, I can see that we use `Decode` without a if statement. Indeed, it doesn't make sense to check the request method in `httpjson`.

> Separate to this PR I'll add some tests and comments that make this clear.

Sounds good to me.

PR #2245, opened 2020-02-07
support/http/httpdecode: add decoding of path params

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Add decoding of path parameters from requests into structs to the `httpdecode` package in a new `DecodePath` function and into the existing `Decode` function.

### Why
We use httpdecode to decode requests into structs and this works really well for JSON and form bodys, but we have the odd endpoint that also has request parameters provided via the path string. An example of this is where an endpoint is for an account. In some situations it might be convenient to use middleware to pick out those path parameters earlier on and provide it in the context. In other situations, which this PR addresses, it can keep our code more succinct if we include those path parameters into the request object that we're decoding the entire request into.

The hope is we can take this code into the `support/render/httpjson` package by having that package use `httpdecode.Decode` so that it is possible to capture all the request input in a single object for the handler.

This PR is a step towards that but that benefit won't be realized yet.

### Notes
This change includes an update to the version of chi that is in use. Version 4 of chi allows for retrieving the context safely even if it is not present.

### TODO
There are some outstanding things before this PR is ready for review:

- [x] This PR is intended to be merged after #2244 and contains the changes from that PR. After it is merged I need to merge master into this PR to remove them from the diff.
- [x] Check for any breaking changes in chi v4.0.3 that are meaningful to all our uses of chi.
- [x] Review the diff of chi v3.1.5 to v4.0.3.

### Known limitations

N/A

### DIFF
--- a/support/http/httpdecode/httpdecode.go
+++ b/support/http/httpdecode/httpdecode.go
@@ -2,7 +2,6 @@ package httpdecode
 
 import (
 	"encoding/json"
-	"fmt"
 	"mime"
 	"net/http"
 
--- a/support/http/httpdecode/httpdecode.go
+++ b/support/http/httpdecode/httpdecode.go
@@ -27,7 +27,6 @@ func DecodePath(r *http.Request, v interface{}) error {
 		v := params.Values[i]
 		paramMap[k] = append(paramMap[k], v)
 	}
-	fmt.Println(params)
 	dec := schema.NewDecoder()
 	dec.SetAliasTag("path")
 	dec.IgnoreUnknownKeys(true)
--- a/support/http/httpdecode/httpdecode.go
+++ b/support/http/httpdecode/httpdecode.go
@@ -2,12 +2,37 @@ package httpdecode
 
 import (
 	"encoding/json"
+	"fmt"
 	"mime"
 	"net/http"
 
+	"github.com/go-chi/chi"
 	"github.com/gorilla/schema"
 	"github.com/stellar/go/support/errors"
 )
+
+// DecodePath decodes parameters from the path in a request used with the
+// github.com/go-chi/chi muxing module.
+func DecodePath(r *http.Request, v interface{}) error {
+	rctx := chi.RouteContext(r.Context())
+	if rctx == nil {
+		return nil
+	}
+	params := rctx.URLParams
+	paramMap := map[string][]string{}
+	for i, k := range params.Keys {
+		if i >= len(params.Values) {
+			break
+		}
+		v := params.Values[i]
+		paramMap[k] = append(paramMap[k], v)
+	}
+	fmt.Println(params)
+	dec := schema.NewDecoder()
+	dec.SetAliasTag("path")
+	dec.IgnoreUnknownKeys(true)
+	return dec.Decode(v, paramMap)
+}
 
 // DecodeQuery decodes the query string from r into v.
 func DecodeQuery(r *http.Request, v interface{}) error {
@@ -77,7 +102,11 @@ func DecodeForm(r *http.Request, v interface{}) error {
 // See DecodeForm and DecodeJSON for details about the types of errors that may
 // occur.
 func Decode(r *http.Request, v interface{}) error {
-	err := DecodeQuery(r, v)
+	err := DecodePath(r, v)
+	if err != nil {
+		return errors.Wrap(err, "path params could not be parsed")
+	}
+	err = DecodeQuery(r, v)
 	if err != nil {
 		return errors.Wrap(err, "query could not be parsed")
 	}
--- a/support/http/httpdecode/httpdecode.go
+++ b/support/http/httpdecode/httpdecode.go
@@ -9,6 +9,14 @@ import (
 	"github.com/stellar/go/support/errors"
 )
 
+// DecodeQuery decodes the query string from r into v.
+func DecodeQuery(r *http.Request, v interface{}) error {
+	dec := schema.NewDecoder()
+	dec.SetAliasTag("query")
+	dec.IgnoreUnknownKeys(true)
+	return dec.Decode(v, r.URL.Query())
+}
+
 // DecodeJSON decodes JSON request from r into v.
 func DecodeJSON(r *http.Request, v interface{}) error {
 	dec := json.NewDecoder(r.Body)
@@ -69,6 +77,10 @@ func DecodeForm(r *http.Request, v interface{}) error {
 // See DecodeForm and DecodeJSON for details about the types of errors that may
 // occur.
 func Decode(r *http.Request, v interface{}) error {
+	err := DecodeQuery(r, v)
+	if err != nil {
+		return errors.Wrap(err, "query could not be parsed")
+	}
 	contentType := r.Header.Get("Content-Type")
 	if contentType != "" {
 		mediaType, _, err := mime.ParseMediaType(contentType)
--- a/support/http/httpdecode/httpdecode.go
+++ b/support/http/httpdecode/httpdecode.go
@@ -84,7 +84,7 @@ func DecodeForm(r *http.Request, v interface{}) error {
 }
 
 // Decode decodes form URL encoded requests and JSON requests from r into v.
-// Also decodes query parameters.
+// Also decodes path (chi only) and query parameters.
 //
 // The requests Content-Type header informs if the request should be decoded
 // using a form URL encoded decoder or using a JSON decoder.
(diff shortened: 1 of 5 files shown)


leighmcculloch:
The reason I'm upgrading chi is because I need this fix: go-chi/chi@221acf29d02b708ec13791ff562888a730bfe6a7. If need be I can work around this but it'd be better long term to handle the upgrade and get this and other fixes in the dependency.

I reviewed the upgrade of chi from [v3.1.5 to v4.0.3](https://github.com/go-chi/chi/compare/v3.1.5...v4.0.0) and the main thing I see which could be meaningful is that chi no longer supports `http.CloseNotifier` which was deprecated in an earlier version of Go ([GoDoc](https://golang.org/pkg/net/http/#CloseNotifier), golang/go@00ebfcde7f3b6e63447e2a3962975e0a8c3729dd).

In Horizon we use the `http.CloseNotifier` interface and check for it in a couple places, so this could be risky to change:

1. https://github.com/stellar/go/blob/2e0313b01b0e317ec73937ca1ce0b9f3dc0d3f2d/support/http/mutil/writer_proxy.go#L32-L45
2. https://github.com/stellar/go/blob/2e0313b01b0e317ec73937ca1ce0b9f3dc0d3f2d/services/horizon/internal/httpx/request.go#L19-L32

@bartekn Do you have a lot of context for why we're relying on `http.CloseNotifier` and thoughts about this upgrade?

bartekn:
@leighmcculloch I checked Horizon code. It looks like the usage in `writer_proxy.go` is there only to be able to use it in `httpx/request.go` (doesn't actually call `CloseNotify` internally other than in a wrapper function).

When it comes to `httpx/request.go` it should be easy to switch to `Request.Context` as recommended in Godoc. I think it's worth fixing this in this PR. Do you want to do it? If not, I can try to fix it tomorrow.

leighmcculloch:
@bartekn Thanks. Leave it with me. I'll add a test that captures the existing behavior provided by httpx's `RequestWithContext` relating to cancelation and see if I can keep that behavior consistent by removing the CloseNotifier code.

leighmcculloch:
I opened #2256 to remove the CloseNotifier code identified above.

support/http/httpdecode/httpdecode.go
```diff
@@ -2,13 +2,38 @@ package httpdecode
...
+	rctx := chi.RouteContext(r.Context())
+	if rctx == nil {
+		return nil
+	}
+	params := rctx.URLParams
+	paramMap := map[string][]string{}
+	for i, k := range params.Keys {
+		if i >= len(params.Values) {
+			break
+		}
```
howardtw: Theoretically the lengths of `params.Keys` and `params.Values` should always equal, right? (If `chi` only uses this [function](https://github.com/go-chi/chi/blob/master/context.go#L128) to add). Also, is it possible that a key would have multiple values so length of `params.Values` is greater than `params.Keys`?

nit: maybe we can just do `i == len(params.Values)` since we are never gonna hit `i > len` in this loop?

support/http/httpdecode/httpdecode.go
```diff
@@ -2,13 +2,38 @@ package httpdecode
...
+		v := params.Values[i]
+		paramMap[k] = append(paramMap[k], v)
+	}
+	fmt.Println(params)
```
abuiles: @leighmcculloch was this for debugging?

support/http/httpdecode/httpdecode.go
```diff
@@ -78,7 +103,11 @@ func DecodeForm(r *http.Request, v interface{}) error {
 // See DecodeQuery, DecodeForm and DecodeJSON for details about the types of
 // errors that may occur.
 func Decode(r *http.Request, v interface{}) error {
-	err := DecodeQuery(r, v)
+	err := DecodePath(r, v)
+	if err != nil {
+		return errors.Wrap(err, "path params could not be parsed")
+	}
+	err = DecodeQuery(r, v)
```
howardtw: Can these two work together? Is there any application that uses both at a single URL? I can't find it on [ietf](https://tools.ietf.org/html/rfc3986#section-3.3), and I see we have no test case for it.

support/http/httpdecode/httpdecode.go
```diff
@@ -78,7 +103,11 @@ func DecodeForm(r *http.Request, v interface{}) error {
 // See DecodeQuery, DecodeForm and DecodeJSON for details about the types of
```
howardtw: Maybe we want to add `DecodePath` here as well.

support/http/httpdecode/httpdecode.go
```diff
@@ -2,13 +2,38 @@ package httpdecode
...
 
+	"github.com/go-chi/chi"
 	"github.com/gorilla/schema"
 	"github.com/stellar/go/support/errors"
 )
 
+// DecodePath decodes parameters from the path in a request used with the
+// github.com/go-chi/chi muxing module.
+func DecodePath(r *http.Request, v interface{}) error {
+	rctx := chi.RouteContext(r.Context())
```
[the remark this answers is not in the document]
leighmcculloch: I don't think we need to avoid `chi` here and the effort for us to support URL parameter parsing ourselves when chi provides that complex logic wouldn't be worth our time.

The `support` package also already has dependencies on `chi`. If a service uses `support/http.NewAPIMux` you get a mux that is setup with all the standard middleware setup and ready to go for an API server. As far as I can tell every service in this repo except for `keystore` and `horizon` uses this, and horizon uses chi directly instead for more custom setup.

support/http/httpdecode/httpdecode.go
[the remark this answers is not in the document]
leighmcculloch: Oops! Thanks! df618e4fbda4e6c45f75cfd564a89b9058b18b6d

leighmcculloch: Oops! Thanks! df618e4fbda4e6c45f75cfd564a89b9058b18b6d

support/http/httpdecode/httpdecode.go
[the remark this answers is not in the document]
howardtw: > I don't think we need to avoid chi here

I'm not saying we need to avoid chi, but I feel like we should make `DecodePath` work with services that don't use chi in the future too.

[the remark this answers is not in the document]
leighmcculloch: Ah interesting. That could be a lot of work though, URL parameter matching with all the features chi supports is rather complex. It might be lower effort and also have other benefits to use chi in all our services?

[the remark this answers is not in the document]
howardtw: Making all our services use chi sounds a bit too much to me. IMO it should be optional. I would say if you find yourself spend a lot of time dealing with complex incoming requests, then using chi makes sense. However, for small and simple services, using chi might be overkill. Needless to say, we would have to manage the impact of the package upgrade on all services that use it once in a while.

 Anyways, I think what we have right now is good. 👍 (But perhaps it is more clear if we rename this function to `DecodePathChi` or something similar.)

support/http/httpdecode/httpdecode.go
leighmcculloch: Yup, best I can tell they should always be equal. I really wish they used a map internally, but I suspect they don't because order of the parameters might be important.

I use `>=` on conditions like this to make this line of code less dependent on how the for loop is written. I'm saying this for loop should only run while `i` is less than the max number of values rather than trying to detect a specific condition at a specific iteration of the loop. Maybe the structure of the values changes one day and as you say there are multiple values per key. Someone refactors the line above to increment more than one, suddenly we are outside the bounds and the for loop goes on forever. It's a bit far fetched but years ago in C I witnessed something to this affect and it became a habit of defensive coding.

support/http/httpdecode/httpdecode.go
leighmcculloch: 💯 Thanks. 500bf0e7

support/http/httpdecode/httpdecode.go
leighmcculloch: Good call, yup they can work together, I added two tests in bdeb2c5b. I think one example where this might be the case is pagination queries since they will often exist on both GET and POST endpoints but it can be helpful to collect the pagination parameters one way, always via a query string. I don't know if we do this today, but either way this code supports it if we were to.

support/http/httpdecode/httpdecode.go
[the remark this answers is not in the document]
leighmcculloch: I agree, we can always be flexible, and we should just be intentional about it, and for any service chi provides building blocks we need like request ID tagging, request logging, and it ensures we do it the same way every time.

>But perhaps it is more clear if we rename this function to `DecodePathChi`

Let's rename it if we need to support multiple types of path formatting.

PR #2246, opened 2020-02-08
keypair: add SignBase64 that returns signatures strings

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

keypair: add SignBase64 that returns signatures strings

### What
Add `SignBase64` to the `keypair.KP`, `keypair.Full`, and
`keypair.FromAddress` types that performs the same operation as `Sign`
but returns the signature as a base64 encoded string.

### Why
In the ecosystem we exchange signatures as base64 strings. In the
Stellar Labratory we display signatures as base64 strings. In the
JavaScript SDK we have functions that add signatures to transactions,
and those accept base64 encoded strings.

I'm writing code that will generate signatures and hand them to other
systems, but right now I don't see any helpers we have for doing that in
a consistent way. This seems like the best place to add that.

### Known limitations

N/A

### DIFF
--- a/keypair/full.go
+++ b/keypair/full.go
@@ -2,6 +2,7 @@ package keypair
 
 import (
 	"bytes"
+	"encoding/base64"
 
 	"github.com/stellar/go/strkey"
 	"github.com/stellar/go/xdr"
@@ -45,6 +46,16 @@ func (kp *Full) Verify(input []byte, sig []byte) error {
 func (kp *Full) Sign(input []byte) ([]byte, error) {
 	_, priv := kp.keys()
 	return xdr.Signature(ed25519.Sign(priv, input)[:]), nil
+}
+
+// SignBase64 signs the input data and returns a base64 encoded string, the
+// common format signatures are exchanged.
+func (kp *Full) SignBase64(input []byte) (string, error) {
+	sig, err := kp.Sign(input)
+	if err != nil {
+		return "", err
+	}
+	return base64.StdEncoding.EncodeToString(sig), nil
 }
 
 func (kp *Full) SignDecorated(input []byte) (xdr.DecoratedSignature, error) {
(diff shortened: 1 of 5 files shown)


keypair/full_test.go
```diff
@@ -41,6 +41,24 @@ var _ = Describe("keypair.Full", func() {
...
+		func(c SignCase) {
+			sig, err := subject.SignBase64([]byte(c.Message))
+
+			Expect(sig).To(Equal(c.Signature))
+			Expect(err).To(BeNil())
+		},
+
+		Entry("hello", SignCase{
+			"hello",
+			"LnXMINUZERyqqt3fRku2UNLq8KXRjXRWk6FhAPKkk3vB3/qLCx9honaZbX7o3rLQ3Z7lEFVgd7At7BZ5LpFcCg==",
```
howardtw: How do we know this is the corresponding base64 message that we are expecting?

leighmcculloch: If you run this command using the JavaScript SDK, that is also the result:
```javascript
var seed = "SDHOAMBNLGCE2MV5ZKIVZAQD3VCLGP53P3OBSBI6UN5L5XZI5TKHFQL4";
var kp = StellarSdk.Keypair.fromSecret(seed);
var sig = kp.sign("hello").toString('base64')
```

howardtw: I understand that. I was trying to think of a better way to know this is the result we want just by looking at this test. I guess it could be difficult given that what we are testing is associated with cryptography.

leighmcculloch: Unfortunately I'm not sure there is. 😞

keypair/full.go
```diff
@@ -47,6 +48,16 @@ func (kp *Full) Sign(input []byte) ([]byte, error) {
 	return xdr.Signature(ed25519.Sign(priv, input)[:]), nil
 }
 
+// SignBase64 signs the input data and returns a base64 encoded string, the
+// common format signatures are exchanged.
```
ire-and-curses:
```suggestion
// common format in which signatures are exchanged.
```

PR #2252, opened 2020-02-10
services/horizon/internal: Support filtering offers by `buying` or `selling`.

### LINKED ISSUE #2225: Use canonical asset representation for filtering offers by selling or buying
Instead of using `selling_asset_type`, `selling_asset_code`, and `selling_asset_issuer`, let's have a single query param call `selling` which takes `AssetCode:AssetIssuer` or `native`.

The same should be done for `buying`.

abuiles:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Extend the offer end-point to support filtering using the canonical representation of an asset in the `selling` and `buying` query parameter.

Fix #2225.

### Why

Using `native` for Lumens or `AssetCode:AssetIssuer` for issued, brings a better developer experience than having to specify three different
query parameters to represent an asset.

### Known limitations

### DIFF
--- a/services/horizon/internal/actions/query_params.go
+++ b/services/horizon/internal/actions/query_params.go
@@ -66,23 +66,23 @@ func (q SellingBuyingAssetQueryParams) Selling() *xdr.Asset {
 
 			return &asset
 		}
-	} else {
-		if len(q.SellingAssetType) == 0 {
-			return nil
-		}
-
-		selling, err := xdr.BuildAsset(
-			q.SellingAssetType,
-			q.SellingAssetIssuer,
-			q.SellingAssetCode,
-		)
-
-		if err != nil {
-			panic(err)
-		}
-
-		return &selling
 	}
+
+	if len(q.SellingAssetType) == 0 {
+		return nil
+	}
+
+	selling, err := xdr.BuildAsset(
+		q.SellingAssetType,
+		q.SellingAssetIssuer,
+		q.SellingAssetCode,
+	)
+
+	if err != nil {
+		panic(err)
+	}
+
+	return &selling
 }
 
 // Buying returns an *xdr.Asset representing the buying side of the offer.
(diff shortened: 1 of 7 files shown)


services/horizon/internal/actions/query_params.go
```diff
@@ -30,38 +55,64 @@ func (q SellingBuyingAssetQueryParams) Validate() error {
...
+			asset := xdr.MustNewCreditAsset(parts[0], parts[1])
 
-	selling, err := xdr.BuildAsset(
-		q.SellingAssetType,
-		q.SellingAssetIssuer,
-		q.SellingAssetCode,
-	)
+			return &asset
+		}
+	} else {
```
tamirms: nit: you don't need the else clause because once you enter the `len(q.SellingAsset) > 0` if statement you always return.

tamirms: in other words you could rewrite this as:
```golang
	if len(q.SellingAsset) > 0 {
		switch q.SellingAsset {
		case "native":
			asset := xdr.MustNewNativeAsset()
			return &asset
		default:
			parts := strings.Split(q.SellingAsset, ":")
			asset := xdr.MustNewCreditAsset(parts[0], parts[1])

			return &asset
		}
	}

	if len(q.SellingAssetType) == 0 {
		return nil
	}
       ...
```

tamirms: I think less indentation makes the code easier to read

abuiles: makes sense!

services/horizon/internal/actions/query_params.go
```diff
@@ -30,6 +55,19 @@ func (q SellingBuyingAssetQueryParams) Validate() error {
...
 // Selling returns an xdr.Asset representing the selling side of the offer.
 func (q SellingBuyingAssetQueryParams) Selling() *xdr.Asset {
+	if len(q.SellingAsset) > 0 {
+		switch q.SellingAsset {
+		case "native":
+			asset := xdr.MustNewNativeAsset()
+			return &asset
+		default:
+			parts := strings.Split(q.SellingAsset, ":")
+			asset := xdr.MustNewCreditAsset(parts[0], parts[1])
```
bartekn: It looks like `SellingAsset` is not validated in `Validate` so it's possible that `MustNewCreditAsset` will `panic` and will generate error log. We should check if params are correct before passing them to `MustNewCreditAsset`.

services/horizon/internal/actions/query_params.go
```diff
@@ -49,6 +87,19 @@ func (q SellingBuyingAssetQueryParams) Selling() *xdr.Asset {
...
 // Buying returns an *xdr.Asset representing the buying side of the offer.
 func (q SellingBuyingAssetQueryParams) Buying() *xdr.Asset {
+	if len(q.BuyingAsset) > 0 {
+		switch q.BuyingAsset {
+		case "native":
+			asset := xdr.MustNewNativeAsset()
+			return &asset
+		default:
+			parts := strings.Split(q.BuyingAsset, ":")
+			asset := xdr.MustNewCreditAsset(parts[0], parts[1])
```
bartekn: Same issue as in `SellingAsset` above.

services/horizon/internal/actions/query_params.go
abuiles: They are validated when we parse the query params via tag validation -- https://github.com/stellar/go/pull/2252/files#diff-608cc9ca294d287778194f4cf676baccR25. I"ll add a test anyways to capture future regressions

PR #2256, opened 2020-02-10
services/horizon/internal/httpx: remove obsolete CloseNotifier code

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Remove old code that uses a http.ResponseWriter's CloseNotifier channel to cause the request's context to be canceled on the server side.

### Why

I want to upgrade us to chi v4 which no longer supports CloseNotifier and I want to remove any use of it in our code to ensure we aren't dependent on chi providing ResponseWriter's that support it. It turns out the removed code is no longer necessary and can be just deleted without any new functionality. In Go 1.8  the context was changed to automatically cancel if the client closes its connection and so we do not need to also listen to the CloseNotifier channel or replace it with anything else.

As a bonus this also reduces the number of Go routines required to handle each request in Horizon by one.

### References

Go 1.8 Release Notes:

>The Context returned by Request.Context is canceled if the underlying net.Conn closes. For instance, if the user closes their browser in the middle of a slow request, the Handler can now detect that the user is gone. This complements the existing CloseNotifier support.

https://golang.org/doc/go1.8#net_http

The CloseNotifier interface has been deprecated since Go 1.8:

https://golang.org/pkg/net/http/#CloseNotifier

The muxer/router we use, `chi`, removed any functionality relating to CloseNotifier in its major 4.0.0 release:

https://github.com/go-chi/chi/issues/347

### Testing

I wrote the included test and confirmed that it fails on Go 1.7 when the code in the httpx package was deleted and passes when it is present. I then ran this test with and without the deleted code in Go 1.8 and Go 1.13 and in both cases it passes, which confirms that this would have been fine to remove once the oldest version of Go we supported was 1.8.

### Known limitations

N/A


services/horizon/internal/middleware.go
```diff
@@ -50,8 +50,7 @@ func contextMiddleware(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		ctx := r.Context()
 		ctx = hchi.WithChiRequestID(ctx)
-		ctx, cancel := httpx.RequestContext(ctx, w, r)
-		defer cancel()
```
leighmcculloch: Technically this PR also changes that every context within Horizon used to be canceled by this line when the middleware stack was unwinding. That will no longer happen, but I don't think that needs to happen. I think the reason this was added was probably to ensure that the cancel function didn't leak in memory because the Go docs say that any cancel function not call will leak. Also to prevent the go routine from leaking. @bartekn Let me know if you have any deeper suspicions about this.

services/horizon/internal/httpx/request_test.go
```diff
@@ -0,0 +1,74 @@
...
+	case <-finishedInServer:
+		if canceledInServer {
+			t.Log("Test: Success! Server confirmed the context canceled! 🎉")
+		} else {
+			t.Error("Test: Fail! Server finished without the context being canceled")
+		}
+	case <-time.After(10 * time.Second):
+		t.Error("Test: Fail! 10s timeout reached without acknowledgement server finished")
+	}
+}
```
leighmcculloch: I don't think there's value in keeping this test around. I'll leave it here for 👀 during PR review and I'll remove it before merging. Please comment here if you'd rather it kept.

services/horizon/internal/middleware.go
bartekn: I think it was there to make sure context is cancelled when `<-ctx.Done()`  in `RequestContext`.

leighmcculloch:
I removed the test I wrote in 2b3d072 because I don't think it is valuable long term for us to be testing the behavior of the Go stdlib.

PR #2251, opened 2020-02-10
exp/ingest/io: Add StatsChangeProcessor

bartekn:
<details>
  <summary>PR Checklist</summary>
  
### PR Structure

* [x] This PR has reasonably narrow scope (if not, break it down into smaller PRs).
* [x] This PR avoids mixing refactoring changes with feature changes (split into two PRs
  otherwise).
* [x] This PR's title starts with name of package that is most changed in the PR, ex.
  `services/friendbot`, or `all` or `doc` if the changes are broad or impact many
  packages.

### Thoroughness

* [ ] This PR adds tests for the most critical parts of the new functionality or fixes.
* [ ] I've updated any docs ([developer docs](https://www.stellar.org/developers/reference/), `.md`
  files, etc... affected by this change). Take a look in the `docs` folder for a given service,
  like [this one](https://github.com/stellar/go/tree/master/services/horizon/internal/docs).

### Release planning

* [ ] I've updated the relevant CHANGELOG ([here](services/horizon/CHANGELOG.md) for Horizon) if
  needed with deprecations, added features, breaking changes, and DB schema changes.
* [ ] I've decided if this PR requires a new major/minor version according to
  [semver](https://semver.org/), or if it's mainly a patch change. The PR is targeted at the next
  release branch if it's not a patch change.
</details>

### What

This commit adds `exp/ingest/io.StatsChangeProcessor` struct responsible for reporting meta changes stats and adds it to Horizon processors.

#2208

### Why

Meta changes stats are essential to understanding possible performance issues in ingestion process (if/as they happen).

### Known limitations

Similar `LedgerTransactionProcessor` will be added in a separate PR.

PR #2260, opened 2020-02-11
services/horizon: Add new ingestion system metrics to /metrics

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This commit adds a few ingestion metrics to `/metrics` endpoint. More will be added when #2251 is merged.

### Why

Ingestion metrics will help us understand production performance of the system and create alerts.

### DIFF
--- a/clients/horizonclient/main_test.go
+++ b/clients/horizonclient/main_test.go
@@ -272,8 +272,9 @@ func TestMetrics(t *testing.T) {
 		assert.Equal(t, metrics.GoRoutines.Value, 1893)
 		assert.Equal(t, metrics.HistoryElderLedger.Value, 1)
 		assert.Equal(t, metrics.HistoryLatestLedger.Value, 22826153)
-		assert.Equal(t, metrics.IngesterClearLedger.Median, float64(0))
-		assert.Equal(t, metrics.IngesterIngestLedger.Percent99_9, 185115016.58600014)
+		assert.Equal(t, metrics.IngestLedgerGraphOnlyIngestion.Min, float64(10))
+		assert.Equal(t, metrics.IngestLedgerIngestion.Median, float64(0.0706217925))
+		assert.Equal(t, metrics.IngestStateVerify.Percent99_9, float64(230.123456))
 		assert.Equal(t, metrics.LoggingDebug.Count, 0)
 		assert.Equal(t, metrics.LoggingError.Rate15m, float64(0))
 		assert.Equal(t, metrics.LoggingInfo.MeanRate, 227.30356525388274)
@@ -993,7 +994,7 @@ var metricsResponse = `{
   "history.open_connections": {
     "value": 27
 	},
-	 "ingester.clear_ledger": {
+  "ingest.ledger_graph_only_ingestion": {
     "15m.rate": 0,
     "1m.rate": 0,
     "5m.rate": 0,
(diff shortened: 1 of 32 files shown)


services/horizon/internal/actions_metrics.go
```diff
@@ -72,15 +74,15 @@ func (action *MetricsAction) LoadSnapshot() {
...
-			values["99.9%"] = ps[4]
+			values["min"] = time.Duration(t.Min()).Seconds()
+			values["max"] = time.Duration(t.Max()).Seconds()
+			values["mean"] = time.Duration(t.Mean()).Seconds()
+			values["stddev"] = time.Duration(t.StdDev()).Seconds()
+			values["median"] = ps[0] / float64(time.Second)
+			values["75%"] = ps[1] / float64(time.Second)
+			values["95%"] = ps[2] / float64(time.Second)
+			values["99%"] = ps[3] / float64(time.Second)
+			values["99.9%"] = ps[4] / float64(time.Second)
```
bartekn: These values were previously presented as nanoseconds. I changed them to seconds for readability.

services/horizon/internal/expingest/main.go
```diff
@@ -81,6 +82,20 @@ type stellarCoreClient interface {
 }
 
 type System struct {
+	Metrics struct {
+		// LedgerIngestionTimer exposes timing metrics about the rate and
+		// duration of ledger ingestion (including updating DB and graph).
+		LedgerIngestionTimer metrics.Timer
+
+		// LedgerGraphOnlyIngestionTimer exposes timing metrics about the rate and
+		// duration of ingestion into graph only.
```
ire-and-curses:
```suggestion
		// duration of ingestion into the in-memory graph.
```

services/horizon/internal/expingest/main.go
```diff
@@ -81,6 +82,20 @@ type stellarCoreClient interface {
...
+		LedgerGraphOnlyIngestionTimer metrics.Timer
```
ire-and-curses: Consider renaming this to something like `LedgerInMemoryIngestionTimer`. "Graph only" was confusing to me until I read the code and understood how this was being used (for a few minutes I wondered if these were data points intended for graphing!). Also if we ever change the data structure the name won't be part of the API so we preserve the abstraction.

clients/horizonclient/main_test.go
```diff
@@ -993,7 +994,7 @@ var metricsResponse = `{
   "history.open_connections": {
     "value": 27
 	},
-	 "ingester.clear_ledger": {
+  "ingest.ledger_graph_only_ingestion": {
```
ire-and-curses: Would there be any performance downside to including a descriptive line in the JSON explaining what these metrics are (basically copying the comments from the metric declarations)? It would make it self-documenting but would be part of every call. Interested in your thoughts @abuiles @bartekn

abuiles: I wouldn't think doing that could impact performance

services/horizon/internal/expingest/main.go
bartekn: Good point, changed!

clients/horizonclient/main_test.go
bartekn: I think we shouldn't do it:
1. I think there's to add annotations to JSON other than a new field like `description` but it seems to be anti-pattern. There is a way to add comments in prometheus format but see the next point.
2. In general this endpoint will be read by tools like Prometheus and you won't find explanations there too.

I think we should just update the docs. Going to do it in another PR.

PR #2261, opened 2020-02-11
services/horizon: Move /metrics to internal server

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This commit moves `/metrics` endpoint to the new internal server running on `INTERNAL_PORT`.

### Why

After reviewing metrics exposed in `/metrics` it seems that none of the values should be exposed publicly. It will also allow us to expose more internal metrics.

### DIFF
--- a/services/horizon/cmd/root.go
+++ b/services/horizon/cmd/root.go
@@ -147,7 +147,7 @@ var configOpts = support.ConfigOptions{
 		ConfigKey:   &config.InternalPort,
 		OptType:     types.Uint,
 		FlagDefault: uint(0),
-		Usage:       "WARNING: this should not be accesible from the Internet, tcp port to listen on for internal http requests, 0 (default) disables the internal server",
+		Usage:       "WARNING: this should not be accessible from the Internet and does not use TLS, tcp port to listen on for internal http requests, 0 (default) disables the internal server",
 	},
 	&support.ConfigOption{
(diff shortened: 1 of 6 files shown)


services/horizon/cmd/root.go
```diff
@@ -142,6 +142,13 @@ var configOpts = support.ConfigOptions{
 		FlagDefault: uint(8000),
 		Usage:       "tcp port to listen on for http requests",
 	},
+	&support.ConfigOption{
+		Name:        "internal-port",
+		ConfigKey:   &config.InternalPort,
+		OptType:     types.Uint,
+		FlagDefault: uint(0),
+		Usage:       "WARNING: this should not be accesible from the Internet, tcp port to listen on for internal http requests, 0 (default) disables the internal server",
```
leighmcculloch: Should we add a note here that the port doesn't use TLS, so even if TLS is configured with the other parameters this port will not use it?

bartekn: @leighmcculloch good point! Does the following change look good?
```suggestion
		Usage:       "WARNING: this should not be accessible from the Internet and does not use TLS, tcp port to listen on for internal http requests, 0 (default) disables the internal server",
```

leighmcculloch: 👍

PR #2274, opened 2020-02-14
exp/services/recoverysigner: add account registration endpoint

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add account registration endpoint.

### Why

Accounts will need to register with the recovery signer to be able to recover their account.

### Merging

### Known limitations

N/A


exp/services/recoverysigner/internal/serve/account_post.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	if req.Address != claims.Address {
+		unauthorized.Render(w)
+		return
+	}
+
+	acc := account.Account{
+		Address: req.Address,
+		Type:    req.Type, // TODO: Validate input type.
```
marcelosalloum: Should we open an issue for this `TODO`?

leighmcculloch: Good idea, once I get these three PRs in I'll create some issues for outstanding TODOs. Thanks.

howardtw:
👀

exp/services/recoverysigner/internal/serve/account_post.go
```diff
@@ -0,0 +1,87 @@
...
+type accountPostRequestIdentities struct {
+	Address     string `json:"account" form:"account"`
+	PhoneNumber string `json:"phone_number" form:"phone_number"`
+	Email       string `json:"email" form:"email"`
+}
+
+func (h accountPostHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+
+	claims, _ := auth.FromContext(ctx)
```
howardtw: Something to consider: I still think it's worth including `claim` in this function, like `auth.ClaimsFromContext`, which is more self-explanatory. The example you pointed to might just mean to demonstrate how to use the context

exp/services/recoverysigner/internal/serve/account_post.go
```diff
@@ -0,0 +1,87 @@
...
+	if claims.Address == "" {
+		unauthorized.Render(w)
+		return
+	}
+
+	req := accountPostRequest{}
+	err := httpdecode.Decode(r, &req)
+	if err != nil || req.Address == "" {
```
howardtw: Should we also validate whether `req.Address` is a good address by calling `keypair.ParseAddress`?

exp/services/recoverysigner/internal/serve/account_post.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/stellar/go/exp/services/recoverysigner/internal/account"
+	"github.com/stellar/go/exp/services/recoverysigner/internal/serve/auth"
+	"github.com/stellar/go/keypair"
+	"github.com/stellar/go/support/http/httpdecode"
+	supportlog "github.com/stellar/go/support/log"
+	"github.com/stellar/go/support/render/httpjson"
+)
+
+type accountPostHandler struct {
+	Logger         *supportlog.Entry
```
howardtw: Do we plan on making this logger configurable? In other words, do we plan on having different loggers in recovery signer? Not sure why we want to make a logger part of the handler.

exp/services/recoverysigner/internal/serve/account_post.go
```diff
@@ -0,0 +1,87 @@
...
+	} else if err != nil {
+		h.Logger.Error(err)
+		serverError.Render(w)
+		return
+	}
+
+	resp := accountResponse{
+		Address:  req.Address,
+		Type:     req.Type,
+		Identity: "account",
```
howardtw: Maybe we want to make `account` a constant? I lost track what this field will be used for.

exp/services/recoverysigner/internal/serve/account_post.go
leighmcculloch: It's a dependency of the handler. All the handlers dependencies are defined on the handler so it is crystal clear what dependencies a handler has.

howardtw: Can you elaborate why we want to make our handlers in the recovery signer depend on a logger?

leighmcculloch: Its easier to write tests that interact with dependencies if we inject dependencies. If we use global versions of things it becomes harder to test. I'm injecting all the other dependencies and it is simple to also inject the logger. It also makes it really clear what the dependencies of a handler are. Do you have a concern with putting it here? Alternative?

exp/services/recoverysigner/internal/serve/account_post.go
leighmcculloch: Another approach might be to rename `Claims` to `Auth`, which makes it a less tied to specifically the concepts of JWTs, and I think easier to understand. You retrieve the Auth details from the context, and 💥 you know if you are authenticated or not. I'll follow up with this.

exp/services/recoverysigner/internal/serve/account_post.go
leighmcculloch: We do that in the code below, which causes an unauthorized response to the user.

If authentication data is invalid, 401 is a much more appropriate specific response than 400 which is broadly indicating that something is wrong with the request. A developer is more likely to see 400 and assume something is incorrect with the response body or the URL, rather than the authorization header.

https://github.com/stellar/go/blob/ae63c683a6fba2bf6d198543e777e3202c1811f7/exp/services/recoverysigner/internal/serve/auth/sep10.go#L60-L63

It's also worth noting that for an address to not be a parseable keypair our SEP-10 server would need to have created a JWT with an invalid keypair so this should also be an impossible scenario, but even though it is something we're protecting against already.

exp/services/recoverysigner/internal/serve/account_post.go
howardtw: > If we use global versions of things it becomes harder to test.

As long as make the output of our logger configurable, we can always point the output to a local variable and test against its value.

> It also makes it really clear what the dependencies of a handler are without reading its code.

I feel logger should be a system-wide thing, so I wouldn't think it's necessary to explicitly include it to make it more clear.

> Do you have a concern with putting it here? Alternative?

It's not a thing that would block a PR, but it's more natural to me to make logger a global thing with a configurable output.

exp/services/recoverysigner/internal/serve/account_post.go
leighmcculloch: #2291

exp/services/recoverysigner/internal/serve/account_post.go
howardtw: > A developer is more likely to see 400 and assume something is incorrect with the response body or the URL, rather than the authorization 

You mean request body right? I thought in this case, `req.Address` would be coming from the request body as opposed to the header tho?

leighmcculloch: Thanks. I meant request body. You're right, we could choose to return bad request on an invalid address in the URL. It's still highly unlikely that we'd ever store an account because for us to store the account a user has to have proven ownership of it via a SEP-10 server and be signed by that trusted JWT. But I agree it'd be nice to not have that dependence and to be clearer here what constitutes a valid address especially given the term address is so easily misunderstood. I think types will help here rather than explicitly doing the validation here. I've opened two PRs that make that happen: #2294 (`keypair`) and #2295 (`exp/services/recoverysigner`).

PR #2275, opened 2020-02-14
exp/services/recoverysigner: add account list endpoint

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add account list endpoint that returns a list of accounts the authenticate user has access to.

### Why

A user will need to be able to ask a recovery signer server what accounts are stored that it is allowed to recover. A user will use this endpoint to discover what accounts it has.

### Merging

### Known limitations

N/A


exp/services/recoverysigner/internal/serve/account_list.go
```diff
@@ -0,0 +1,134 @@
...
+				identity = "other"
+			}
+			accResp := accountResponse{
+				Address:  acc.Address,
+				Type:     acc.Type,
+				Identity: identity,
+				Signer:   h.SigningAddress.Address(),
+			}
+			resp.Accounts = append(resp.Accounts, accResp)
+		}
```
marcelosalloum: Can we avoid this code repetition? We basically have the same code in three chunks: 
* [58-78](https://github.com/stellar/go/pull/2275/files#diff-119d97a59fe7685cf481265102ad9c5fR58-R78)
* [84-104](https://github.com/stellar/go/pull/2275/files#diff-119d97a59fe7685cf481265102ad9c5fR84-R104)
* [110-130](https://github.com/stellar/go/pull/2275/files#diff-119d97a59fe7685cf481265102ad9c5fR110-R130)

Could we extract them to a separate method that uses the `reflect` package? Something like:

```go
reflect.ValueOf(acc.OwnerIdentities).FieldByName("Email")
```

howardtw:
👀

exp/services/recoverysigner/internal/serve/account_list.go
leighmcculloch: There might be a way to avoid the repetition here. I don't want to use `reflect` because then we will lose type safety and the code will be more brittle. Repetitive code > `reflect` in Go most of the time. I'll have a think about this and I might open a follow up PR addressing this if you're okay with that.

howardtw: Another idea is to include `identity` in `accout.Account` directly (@leighmcculloch I think you've mentioned before) so that we don't have to do the above loops. Given that the information will be overlapped (can be derived from) with `OwnerIdentities` or `OtherIdentities`, we could also implement a series of methods on Account like `Account.AddressIdentity(address)`, `Account.EmailIdentity(email)`, and `Account.PhoneNumberIdentity(phoneNumber)`.

leighmcculloch: Identity is a function of both the authenticated user and the account, so that's why I didn't place it on the account, and what makes it a little more difficult. It's value changes depending on which user is authenticated. I'd like some more time to think about it and revisit it in a follow-up.

exp/services/recoverysigner/internal/serve/account_list.go
```diff
@@ -0,0 +1,134 @@
...
+		if err != nil {
+			h.Logger.Error(err)
+			serverError.Render(w)
+			return
+		}
+		for _, acc := range accs {
+			identity := ""
+			switch {
+			case claims.Address == acc.OwnerIdentities.Address:
+				identity = "owner"
```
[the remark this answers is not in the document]
leighmcculloch: 👍 I'll make an issue and follow up with doing this in this sprint. Thanks.

PR #2276, opened 2020-02-14
exp/services/recoverysigner: add sign endpoint

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add account list endpoint that returns a list of accounts the authenticate user has access to.

### Why

A user will need to be able to ask a recovery signer server what accounts are stored that it is allowed to recover. A user will use this endpoint to discover what accounts it has.

### Merging

### Known limitations

N/A


howardtw:
👀 in 5 mins. Taking a short break

exp/services/recoverysigner/internal/serve/account_sign.go
```diff
@@ -0,0 +1,127 @@
...
+type accountSignHandler struct {
+	Logger            *supportlog.Entry
+	SigningKey        *keypair.Full
+	NetworkPassphrase string
+	AccountStore      account.Store
+}
+
+type accountSignRequest struct {
+	Address     string `path:"address"`
+	Transaction string `json:"transaction" form:"transaction"`
```
howardtw: I really like what we do in `httpdecode.Decode` so that we can make what we expect from a request extra clear by just looking at the tags.

exp/services/recoverysigner/internal/serve/account_sign.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+type accountSignResponse struct {
+	PublicKey         string `json:"public_key"`
```
howardtw: I just realized the difference in the format of the json tags. Looks we are adopting the snake case here in recovery signer while VegaPay is using the camel case. Do you see it can become a problem for the clients? Did we decide to use the snake case because of some SEPs?

exp/services/recoverysigner/internal/serve/account_sign.go
```diff
@@ -0,0 +1,127 @@
...
+	phoneNumberAuthenticated := false
+	if claims.PhoneNumber != "" {
+		phoneNumberAuthenticated = claims.PhoneNumber == acc.OwnerIdentities.PhoneNumber ||
+			claims.PhoneNumber == acc.OtherIdentities.PhoneNumber
+	}
+	emailAuthenticated := false
+	if claims.Email != "" {
+		emailAuthenticated = claims.Email == acc.OwnerIdentities.Email ||
+			claims.Email == acc.OtherIdentities.Email
+	}
```
howardtw: It's probably worth making it a helper function with three returned values `func verifyClaimsAccess(Claims) (bool, bool, bool)`. I'm trying to think of ways to make this `ServeHTTP` func shorter with helper functions given that it's obviously doing several steps to resolves different tasks. However, it seems a bit tricky to do so because of the way we do render and return.

exp/services/recoverysigner/internal/serve/account_sign_test.go
```diff
@@ -0,0 +1,800 @@
...
+	require.NoError(t, err)
+
+	wantBody := `{
+	"error": "The resource at the url requested was not found."
+}`
+	assert.JSONEq(t, wantBody, string(body))
+}
+
+// Test that when the account exists but the authenticated client does not have
+// perission to access it returns not found.
```
howardtw: perission -> permission

exp/services/recoverysigner/internal/serve/account_sign_test.go
```diff
@@ -0,0 +1,800 @@
...
```
howardtw: perission -> permission

exp/services/recoverysigner/internal/serve/account_sign_test.go
```diff
@@ -0,0 +1,800 @@
...
```
howardtw: perission -> permission

exp/services/recoverysigner/internal/serve/account_sign.go
leighmcculloch: It's because of our SEPs. I reviewed our existing things when I was deciding which to use. It's unfortunate that we've used camel case in VegaPay, but where this project lives I think it is more important that our SEPs are consistent.

exp/services/recoverysigner/internal/serve/account_sign_test.go
leighmcculloch: Thanks. 52730ad3

exp/services/recoverysigner/internal/serve/account_sign_test.go
leighmcculloch: Thanks. 52730ad3

exp/services/recoverysigner/internal/serve/account_sign_test.go
leighmcculloch: Thanks. 52730ad3

exp/services/recoverysigner/internal/serve/account_sign.go
leighmcculloch: It's also worth noting that the slightly more complex than normal authentication is actually part of the core logic that this service provides rather than being peripheral, so in that sense we can expect to see more of that code sitting center stage. But I agree, we can probably simplify this. I'll revisit it. I like the idea of using a function, that might work. I'll give it a go.

exp/services/recoverysigner/internal/serve/account_sign.go
howardtw: That makes sense. FWIW the reason that we chose to do camel case is that the technology the frontend team adopts uses camel case convention.

marcelosalloum:
> Nothing it too concerning except this [comment](https://github.com/stellar/go/pull/2276#discussion_r381575264) regarding the json format we use. (This, however, applies to all the endpoints.).

Do you think we should ping the frontend team about the `snake_case` vs `CamelCase`?

leighmcculloch:
@marcelosalloum Good call, I'll talk to them about it today.

Issue #2272, opened 2020-02-12
Friendbot fund is 200 but account GET is 404


tyvdh:
### What version are you using?
`"stellar-sdk": "^4.0.0"`

### What did you do?
Funded and then immediately attempted to GET an account.

http://tyler.link/z79z3k

You can replicate this issue here:
https://hack-stellar.now.sh/

Just click "Generate Keypair" and then "Fund Account".

Note it won't happen consistently, you may have to repeat the above sequence several times before you get your first 404.

### What did you expect to see?
A successful 200 response for the account GET request

### What did you see instead?
A 404. If the funding was successful you shouldn't get a 404 no matter how soon you make subsequent requests.

leighmcculloch:
After some period of time does Horizon return 200 OK for the account? How long does it take?

Horizon ingests data from the network and there's no guarantee that it ingests the last ledger so fast that there is no period of time between the account existing and Horizon knowing it exists. Ideally this period of time is small but it is eventually consistent, not strongly consistent.

tyvdh:
> After some period of time does Horizon return 200 OK for the account? How long does it take?
Yes. Quick. Maybe ~1 second though I haven’t tested that specifically. My complaint is less that core and horizon aren’t in sync and more that friendbot should wait to respond until we’re sure a GET on that account will be 200. However I’m thinking now this is perhaps unnecessary as we’ll need to be building for these time gaps in every horizon POST, PATCH type call. No need to make an exception here. 

I am curious though if we track the time gap and have any guarantee or even just average, max times that those gaps can be/have been. ~1 second is very different than ~30 seconds. Curious what affects that time and how susceptible it is to variation. Hard to build for inconsistency especially when a 404 could mean either doesn’t exist or just doesn’t exist yet... but it might... in a second or 3.

bartekn:
Thanks for the report. Yes, this seems important and I suspect the data source for `/accounts/{id}` is to blame. Previously this was read from the Core DB, now we use ingested data in horizon DB. This is happening because in `txsub` we check both sources: Core and Horizon DBs to check if tx was added to a ledger. It's possible that the latest ledger is still being ingested by Horizon thus 404. Unfortunately, we can't just remove Core DB check because some Horizon instances don't ingest failed transactions. I can see two fixed, long term and short term:
1. Long term fix is to always ingest all transactions (including failed). This may increase storage needs but failed transactions are essential for some components like fee stats or txsub.
[...]

PR #2277, opened 2020-02-14
services/horizon/txsub: Skip checking Core DB if Horizon is ingesting failed txs

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Skip checking Stellar-Core DB for transaction results if Horizon is ingesting failed transactions.

### Why

Due to a delay between Stellar-Core closing a ledger and Horizon ingesting it, it's possible that results of transaction submitted to Stellar-Core via Horizon may not be immediately visible. This is happening because `txsub` package checks two sources when checking a transaction result: Stellar-Core and Horizon DB.

The extreme case is https://github.com/stellar/go/issues/2272 where results of transaction creating an account are not visible: requesting account details in Horizon returns `404 Not Found`:

```
     Horizon DB                  Core DB                  User
         |                          |                      |
         |                          |                      |
         |                          | <------- Submit create_account op
         |                          |                      |
         |                  Insert transaction             |
         |                          |                      |
         |                     Tx found -----------------> |
         |                          |                      |
         |                          |                      |
         | <--------------------------------------- Get account info
         |                          |                      |
         |                          |                      |
Account NOT found ---------------------------------------> |
         |                          |                      |
   Insert account                   |                      |
```

### Known limitations

This is just a temporary solution. The long term fix needs to be properly tested before a release.

### DIFF
--- a/services/horizon/internal/init.go
+++ b/services/horizon/internal/init.go
@@ -198,6 +198,37 @@ func dialRedis(redisURL *url.URL) func() (redis.Conn, error) {
 }
 
 func initSubmissionSystem(app *App) {
+	// Due to a delay between Stellar-Core closing a ledger and Horizon
+	// ingesting it, it's possible that results of transaction submitted to
+	// Stellar-Core via Horizon may not be immediately visible. This is
+	// happening because `txsub` package checks two sources when checking a
+	// transaction result: Stellar-Core and Horizon DB.
+	//
+	// The extreme case is https://github.com/stellar/go/issues/2272 where
+	// results of transaction creating an account are not visible: requesting
+	// account details in Horizon returns `404 Not Found`:
+	//
+	// ```
+	// 	 Horizon DB                  Core DB                  User
+	// 		 |                          |                      |
+	// 		 |                          |                      |
+	// 		 |                          | <------- Submit create_account op
+	// 		 |                          |                      |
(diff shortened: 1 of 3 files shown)


services/horizon/internal/init.go
```diff
@@ -198,7 +198,10 @@ func dialRedis(redisURL *url.URL) func() (redis.Conn, error) {
 }
 
 func initSubmissionSystem(app *App) {
-	cq := &core.Q{Session: app.CoreSession(context.Background())}
+	var cq *core.Q
+	if !app.config.IngestFailedTransactions {
```
ire-and-curses: Can we add a comment to explain this? Even a link back to this PR will be helpful since this is temporary

services/horizon/internal/txsub/results/db/main.go
```diff
@@ -15,7 +15,7 @@ import (
 )
 
 // DB provides transactio submission results by querying the
```
ire-and-curses: saw a typo in the previous line ;)
```suggestion
// DB provides transaction submission results by querying the
```

services/horizon/internal/init.go
bartekn: Good point! Added this PR message as a the comment for full context.

PR #2283, opened 2020-02-18
all: add ci builds for Go 1.14rc1

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add a CI build that runs tests with Go 1.14rc1.

### Why

To prepare for us supporting Go 1.14 when it is released. Adding our builds so that we're testing and running checks with 1.14 will give us an opportunity to 👀 any to the stdlib and see how they interact with our code, possibly even report them to the Go project, and make moving to 1.14 a non-event when it is released.

### Known limitations

The builds added in this PR will not be required builds and so they can fail and we can still merge PRs. They're intended to be informational, and something I'll use to track whether we're ready or not.


leighmcculloch:
It is possible for us to add builds for Go 1.14rc1 on Debian Stretch only because we requested those Docker images to be published in docker-library/golang#318. This might be the last time a build using Stretch is granted.

leighmcculloch:
Yeah I've still got this marked as a draft PR because it looks like checks are broken. I think it is something to do with we have an old version of golang.org/x/tools in our go.mod file, but that is only part of the picture. I need to dig into it more. I'll fix that and include that in this PR before I merge it.

leighmcculloch:
Superseded by #2325.

PR #2286, opened 2020-02-19
tools/horizon-cmp: Retry on Latest-Ledger header mismatch

### LINKED ISSUE #2017: tools/horizon-cmp: Use Latest-Ledger header
In #1791 we added `Latest-Ledger` header. We should check it in `horizon-cmp` and repeat requests later if `Latest-Ledger` is different in base and test servers. This will decrease the number of false positives.

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This commit queues a path to retry later when `Latest-Ledger` header is not equal.

### Why

If `Latest-Ledger` is not equal horizon-cmp can report the given path even though it's fine if responses are not equal.

Close #2017.

### DIFF
--- a/tools/horizon-cmp/main.go
+++ b/tools/horizon-cmp/main.go
@@ -183,6 +183,16 @@ func run(cmd *cobra.Command) {
 
 			requestWg.Wait()
 
+			// Retry when LatestLedger not equal but only if not empty because
+			// older Horizon versions don't send this header.
+			if a.LatestLedger != "" && b.LatestLedger != "" &&
+				a.LatestLedger != b.LatestLedger {
+				visitedPaths[pl.ID()] = false
+				paths <- pl
+				log.Warnf("LatestLedger does not match, retry queued: %s", pl.Path)
+				return
+			}
+
 			var status string
 			if a.Equal(b) {
 				status = "ok"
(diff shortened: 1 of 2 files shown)


tools/horizon-cmp/main.go
```diff
@@ -183,6 +183,16 @@ func run(cmd *cobra.Command) {
 
 			requestWg.Wait()
 
+			// Retry when LatestLedger not equal but only if not empty because
+			// older Horizon versions don't send this header.
+			if a.LatestLedger != "" && b.LatestLedger != "" &&
+				a.LatestLedger != b.LatestLedger {
+				visitedPaths[pl.ID()] = false
```
tamirms: is it safe to update `visitedPaths` from multiple goroutines?

tamirms:
I wonder if closing the paths channel is safe in `addPathsFromResponse()`
https://github.com/stellar/go/blob/3392e7b5a46f89bfef2376b47bfafd06eaacb538/tools/horizon-cmp/main.go#L345-L347

It seems like it could be possible that paths is empty but there could be multiple go routines that are still comparing horizon responses. The other go routines might attempt to add to the paths channel after it has been closed.

tools/horizon-cmp/main.go
bartekn: Good point, fixed in https://github.com/stellar/go/pull/2286/commits/ea2af4eb91c1a67063f52da9c31f2d03248a4ba0.

bartekn:
> It seems like it could be possible that paths is empty but there could be multiple go routines that are still comparing horizon responses. The other go routines might attempt to add to the paths channel after it has been closed.

You're right! Going to create an issue for this.

PR #2290, opened 2020-02-19
Backfill fee stats data to make the  client compatible with horizon 1.X and 0.X.

abuiles:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Make the client compatible with fee stats data from Horizon 1.X or 0.X.

### Why

There are a couple of breaking changes in the Horizon response which we are handling in horizonclient 2.0 in a backward compatibility manner, for example, being able to read a `number` or a `string` in some of the response fields. It allows a smooth transition between the new and old versions of Horizon. 

This PR follows the same approach for fee stats, Horizon is dropping all the `_accepted_fee` fields in `fee_stats` but instead of adding this breaking change in the next release of the client, we'll make it backward compatible, so people can still use `p99_accepted_fields` in their code and have time to upgrade up to horizonclient 3.0.

### Known limitations

[TODO or N/A]

### DIFF
--- a/clients/horizonclient/main_test.go
+++ b/clients/horizonclient/main_test.go
@@ -418,6 +418,36 @@ func TestFeeStats(t *testing.T) {
 		assert.Equal(t, fees.P90AcceptedFee, 4000)
 		assert.Equal(t, fees.P95AcceptedFee, 5000)
 		assert.Equal(t, fees.P99AcceptedFee, 8000)
+
+		assert.Equal(t, fees.MaxFee.Min, int64(130))
+		assert.Equal(t, fees.MaxFee.Max, int64(8000))
+		assert.Equal(t, fees.MaxFee.Mode, int64(250))
+		assert.Equal(t, fees.MaxFee.P10, int64(150))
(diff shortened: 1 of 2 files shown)


clients/horizonclient/main_test.go
```diff
@@ -1363,19 +1363,37 @@ var feesResponse = `{
   "last_ledger": "22606298",
   "last_ledger_base_fee": "100",
   "ledger_capacity_usage": "0.97",
-  "min_accepted_fee": "130",
```
ire-and-curses: Shouldn't these fields be kept, not removed? These are the ones being backfilled right, so we'd expect them in the response, wouldn't we?

abuiles: @ire-and-curses no, they are being removed from the response and then we backfill from the values in `max_fee` -- but the backfill doesn't happen in the response but inside the `FeeStat` struct which is what we do here
https://github.com/stellar/go/pull/2290/files#diff-a12e7e90b7de9b52aa97a8dca1ea276cR324

PR #2291, opened 2020-02-19
exp/services/recoverysigner: rename auth.Claims to auth.Auth

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Rename auth.Claims to auth.Auth.

### Why
The term claims is inherited from JWTs, but the code using the Claims
type is not concerned with JWTs, its concerned with do we have an
authenticated user and who are they. I renamed the type to more
generally refer to that concept. It also helps keep the number of
concepts that the package exposes to a minimum and makes it clear that
the package operates around this central Auth type and going from
external forms of authentication to that type. As a bonus it also means
functions like auth.NewContext and auth.FromContext are clearer to
understand. All the functions in this package operate around getting to
an Auth.

### Merging

This change is based on #2276 and should be merged to master after it.

### Known limitations

N/A


marcelosalloum:
👀 👍

PR #2292, opened 2020-02-19
exp/services/recoverysigner: setup firebase auth client only once

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Setup Firebase auth.Client only once at the beginning of service start
instead of on every request.

### Why
@howardtw pointed out that the verification function where the
auth.Client instantiation occurred had to bury any failure to setup the
auth.Client because it doesn't expose errors. This doesn't seem ideal
since a failure to setup the client is not a failure in authentication.
So if there was an issue there we wouldn't know about it, users would
just be failing to authenticate silently.

It also looks like the Firebase auth.Client when instantiated goes and
discovers what keys it can use for verifying JWTs. We don't need to do
this on every request, and in fact on every request this is unnecessary.
There's always a chance the keys Google publishes will change, but
getting them on app start is still probably sufficient.

These two reasons combined seem like a good reason to setup the auth
client once at service start.
### Known limitations

N/A


marcelosalloum:
👀 👏

PR #2294, opened 2020-02-19
keypair: add unmarshal and marshal functions to FromAddress

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Add UnmarshalText and MarshalText functions to keypair.FromAddress.

### Why
So that it is easy to use the keypair.FromAddress type in requests.
When using the encoding/json, encoding/xml, and other encoding
implementations it can be useful to be able to define a field on a type
as a keypair.FromAddress so that parsing of the data fails in the event
of an invalid address instead of needing to do this parsing and
validation separately.

By implementing the *Text functions we can support any encoding
implementation.

I didn't make this functions on the KP interface because it is
unnecessary for the interface to adhere to encoding.TextMarshaler, and I
didn't add the functions to keypair.Full because it is not a common
practice to pass private keys around so this would be of little value to
that type.

See #2295 for an example of how this will be used.

### Known limitations

N/A


keypair/from_address.go
```diff
@@ -57,3 +59,18 @@ func (kp *FromAddress) SignDecorated(input []byte) (xdr.DecoratedSignature, erro
 func (kp *FromAddress) publicKey() ed25519.PublicKey {
 	return ed25519.PublicKey(strkey.MustDecode(strkey.VersionByteAccountID, kp.address))
 }
+
+var _ = encoding.TextMarshaler(&FromAddress{})
```
howardtw: This is just to test `*FromAddress` implements `TextMarshaler` right? If so, do we need one for `TextUnmarshaler`? It would make more sense to have both or nothing.

leighmcculloch: Oops, good point. I missed that. Added in #2302.

leighmcculloch: And yup, this is to check that the `FromAddress` type implements the interfaces, and to make that relationship explicit within the keypair package. Without it, we could change the keypair package, build it, and then only find out we had broken the contract when building a package that imports keypair.

PR #2295, opened 2020-02-20
exp/services/recoverysigner: validate account address in urls

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Validate account addresses that are in URLs that are accepted into the
recoverysigner service.

### Why
@howardtw pointed out that it'd be good to validate the address in the
request URLs.

The account addresses in the requests are not explicitly validated. It
is highly unlikely an invalid address could end up in these fields
because a SEP-10 server has to sign a JWT indicating the address is a
valid account for it ever to be accepted and registered. However, it's
worthwhile to validate rather than rely entirely on that other system on
that fact.

### Merging
This PR is dependent on #2294 and should be merged to `master` after it.

### Known limitations

N/A


exp/services/recoverysigner/internal/serve/account_post_test.go
```diff
@@ -161,6 +202,6 @@ func TestAccountPost_notAuthenticatedForAccount(t *testing.T) {
 }`
 	assert.JSONEq(t, wantBody, string(body))
 
-	_, err = s.Get("GDIXCQJ2W2N6TAS6AYW4LW2EBV7XNRUCLNHQB37FARDEWBQXRWP47Q6N")
+	_, err = s.Get("GDUKTYDY3RDNTNOUFJ2GPL5PIZTMTRD5P2CT274SYH67Q5J3NYI7XKYB")
```
leighmcculloch: This test was using an invalid address and so I needed to update it to use a valid address as part of this change since we no longer allow invalid addresses.

howardtw:
👀

exp/services/recoverysigner/internal/serve/account_post.go
```diff
@@ -43,18 +43,18 @@ func (h accountPostHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	req := accountPostRequest{}
 	err := httpdecode.Decode(r, &req)
-	if err != nil || req.Address == "" {
+	if err != nil || req.Address == nil {
```
howardtw: I'm actually wondering whether `req.Address == nil` will ever get hit given that `ParseAddress` returns an error for all invalid addresses including an empty address.

leighmcculloch: In practice it won't get hit, but theoretically if the `{address}` parameter was removed from the URL path then we would never attempt to deserialize a value for address with keypair and the field would remain `nil`, so I think the check is still good.

PR #2299, opened 2020-02-20
services/horizon/expingest: Ingest into memory only flag

### LINKED ISSUE #2250: Ingestion nodes cannot be separated from request serving horizon nodes
We would like support the following deployment topology for horizon:

There are two pools of horizon instances. The first pool consists of horizon nodes which solely focus on ingestion. The nodes in this pool do not serve API requests. The other pool consists of horizon nodes which do not participate in ingestion and only serve API requests.

We discussed this issue previously in https://github.com/stellar/go/issues/1529 and https://github.com/stellar/go/pull/1519#issuecomment-513864010 . We came to the conclusion that the simplest solution was to make all horizon instances participate in ingestion AND serve API requests.

Now that we have several months of experience running horizon in this scheme (where all nodes serve API requests and ingest) we've observed that ingestion can be quite demanding in terms of CPU and memory. We would like to separate the ingestion operations from request serving operations to reduce resource contention between the two.

The difficulty in separating ingestion nodes from API nodes is that the `/order_book` and `/paths` endpoint rely on an in memory orderbook which is populated by ingestion. As a workaround we have chosen to still maintain the two pools of horizon instances but, the ingestion nodes also serve `/order_book` and `/paths` requests and the API nodes serve all other requests.

Ideally, we would like a solution that does not require special routing to handle the in memory orderbook case.

Here are some ideas:

For the `/order_book` endpoint we actually don't need to use the in memory graph. We have all the offers recorded in the horizon database:

https://github.com/stellar/go/blob/release-horizon-v0.25.0/services/horizon/internal/db2/schema/migrations/19_offers.sql

It should be possible to obtain all the data we need to fulfill the `/order_book` requests with sql queries on the `offers` table. However, I don't know how quickly those queries would run and we may need to add some additional indexes to improve performance.

---

For the path finding endpoints, I think we still need to maintain an in memory graph. But, I think it should be possible to decouple maintaining the in memory graph from ingestion. Here's the idea which was originally proposed in https://github.com/stellar/go/pull/1519#issuecomment-513864010 :

When the API horizon instances start up, they build an in memory order book graph by reading all rows from the offers table. The API horizon instances will periodically poll the offers table for new updates and apply them to the in memory order book graph. We should be able to find updates from the offers table using the `last_modified_ledger` column.

However, when an offer is removed from the order book, the corresponding row in the offers table is deleted. When polling for updates we would not be able to observe changes where offers are removed from the orderbook. To fix this issue we could add a deleted column on the offers table which acts as a tombstone. Instead of deleting a row, we would set the deleted flag to true. To limit the size of the offers table, we can periodically delete tombstone rows from the offers table which are older than 10,000 ledgers (or some other large cutoff).

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This commit adds a config flag that makes ingestion system to ingest into memory structures only.

Close #2250.

### Why

Currently it's not possible to move ingestion system entirely to backend server. Because of this a spike in requests to frontend servers may increase CPU usage and slow down ingestion. To solve this, it's now possible to set `INGEST_IN_MEMORY_ONLY` config flag that would make the instance wait until other ingesting instance updates a DB and then it will just catchup ingesting into memory structures only.

### Known limitations

This requires more tests before releasing so probably shouldn't be included in 1.0.0.


bartekn:
I realized that we should also check the condition in `historyRangeState` state (and possibly just in `start`). Basically, when `INGEST_IN_MEMORY_ONLY` is set, the instance should be able to operate with read-only DB connection.

ire-and-curses:
Nice! With this change, do I understand correctly that it would be possible to have front-end instances that serve e.g. the orderbook graph from memory, without having them be ingesters writing to the DB? That would imply that we could then split a cluster into true front end (serving requests) and backend (pure ingestion to DB), correct?

bartekn:
@ire-and-curses exactly!

services/horizon/internal/expingest/fsm.go
```diff
@@ -172,6 +172,12 @@ func (startState) run(s *System) (transition, error) {
 
 	switch {
 	case lastHistoryLedger > lastIngestedLedger:
+		// If ingesting into memory wait until other ingesting instance ingests state.
+		if s.config.IngestInMemoryOnly {
+			log.Info("Waiting for other ingesting instance to ingest into DB...")
+			return start(), nil
```
tamirms: I think it's worth adding a test for this case

tamirms: nevermind, I see the test already exists

tamirms:
@bartekn I like the idea you mentioned of using a read only connection. Along the same lines, I think you can set the ingestion transaction to be read only by calling `BeginTx(&sql.TxOptions{ReadOnly: true})`. Would it make sense to have a function called `beginIngestionTx()` which creates a transaction that is read only iff `config.IngestInMemoryOnly` is true? Then all the state machine nodes can call `beginIngestionTx()` to create their transactions

tamirms:
@bartekn it turns out that postgres does not allow select for update queries on read only transactions. if we want to restrict ingestion to use read only connections when `config.IngestInMemoryOnly` is true, we can either use a postgres advisory lock to replace the select for update  or, we can avoid locking entirely when `config.IngestInMemoryOnly` is true.

I've confirmed that postgres advisory locks do work on read only transactions. The advantage of this approach is that it maintains the existing locking behavior.

I think that it is probably safe for us to avoid locking when `config.IngestInMemoryOnly` is true but I'm not 100% sure because it's hard to analyze all the possible scenarios. The advantage of avoiding locking is that, if we have many read only horizon nodes it is possible for those read only nodes to delay ingestion if they repeatedly acquire the lock before the horizon ingestion nodes. But, I'm not sure how much of an issue this would be in practice.

bartekn:
@tamirms thanks for checking this! I prefixed the flag with `EXP_` in https://github.com/stellar/go/pull/2299/commits/951c73063940dde6c2b52ba273d29a4c62b7015e so let's merge it without using read-only transaction. I think you are right about not locking in `IngestInMemoryOnly` mode - let's confirm it while working on the next version.

PR #2301, opened 2020-02-20
services/horizon: Fix docker-compose Postgres' auth

2opremio:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Fix the authentication of the Horizon container against the Postgres Horizon container at
`services/horizon/docker/docker-compose.yml`.

### Why/How

By not pinning the full Postgres version, authentication stopped working since now it requires either setting the Postgres password or an explicit  `POSTGRES_HOST_AUTH_METHOD=trust`:

```
Error: Database is uninitialized and superuser password is not specified.
       You must specify POSTGRES_PASSWORD for the superuser. Use
       "-e POSTGRES_PASSWORD=password" to set it in "docker run".

       You may also use POSTGRES_HOST_AUTH_METHOD=trust to allow all connections
       without a password. This is *not* recommended. See PostgreSQL
       documentation about "trust":
       https://www.postgresql.org/docs/current/auth-trust.html
```

Making this breaking change on a patch release is arguably a mistake on Postgres' end (see https://github.com/docker-library/postgres/issues/681#issuecomment-586511283 ) but ... it is what it is.

This PR sets `POSTGRES_HOST_AUTH_METHOD=trust` and also pins the patch number of Postgres to make it future-proof.

### Known limitations

N/A

### DIFF
--- a/services/horizon/docker/docker-compose.yml
+++ b/services/horizon/docker/docker-compose.yml
@@ -1,7 +1,7 @@
 version: '3'
 services:
   core-postgres:
-    image: postgres:9.6-alpine
+    image: postgres:9.6.17-alpine
     restart: on-failure
     environment:
       - POSTGRES_PASSWORD=mysecretpassword
@@ -30,9 +30,10 @@ services:
     network_mode: '${NETWORK_MODE:-bridge}'
 
   horizon-postgres:
-    image: postgres:9.6-alpine
+    image: postgres:9.6.17-alpine
     restart: on-failure
     environment:
+      - POSTGRES_HOST_AUTH_METHOD=trust
       - POSTGRES_DB=horizon
     ports:
       - "5432:5432"


2opremio:
@cla-bot check

leighmcculloch:
Hi @2opremio! Thanks for bringing this problem to our attention and opening a fix! Please bare with me a moment while I update our clabot-config.

---

In terms of how we address this breaking change in Postgres, I think we should set `POSTGRES_HOST_AUTH_METHOD=trust` in the `docker-compose.yml`. While pinning to the previous version works, I think if that will only serve to pin us into the past. It is a simple enough change to make to add the environment variable.

2opremio:
> In terms of how we address this breaking change in Postgres, I think we should set `POSTGRES_HOST_AUTH_METHOD=trust` in the `docker-compose.yml`

Yes, that's what the PR does.

2opremio:
> While pinning to the previous version works, I think if that will only serve to pin us into the past

It pins the current version, not the past one. Do you mean you prefer to have the patch number unpinned?

leighmcculloch:
@cla-bot check

services/horizon/docker/start.sh
```diff
@@ -16,4 +16,4 @@ linux-*)
     ;;
 esac
 
-docker-compose up --build -d
\ No newline at end of file
+docker-compose up --build
```
leighmcculloch: This change seems unrelated to the main purpose of the PR. Could you revert this line? Thanks!

services/horizon/docker/docker-compose.yml
```diff
@@ -1,7 +1,7 @@
 version: '3'
 services:
   core-postgres:
-    image: postgres:9.6-alpine
+    image: postgres:9.6.17-alpine
```
leighmcculloch: I think we should leave this change out and continue to use the latest 9.6 patch version. It's valuable for us to continue to use the latest security patches and to not need to update this manually on each patch release. Normally a patch release won't include breaking changes and this seems like an exception, so the pain of a new version breaking use is unlikely to be frequent.

services/horizon/docker/docker-compose.yml
```diff
@@ -30,9 +30,10 @@ services:
     network_mode: '${NETWORK_MODE:-bridge}'
 
   horizon-postgres:
-    image: postgres:9.6-alpine
+    image: postgres:9.6.17-alpine
     restart: on-failure
     environment:
+      - POSTGRES_HOST_AUTH_METHOD=trust
       - POSTGRES_DB=horizon
```
leighmcculloch: @tamirms Do you think we could add a POSTGRES_PASSWORD instead, which would be more consistent with `core-postges`? It might encourage better practices when folks move from using this to deployment.

services/horizon/docker/docker-compose.yml
2opremio: >  It's valuable for us to continue to use the latest security patches and to not need to update this manually on each patch release

As far as I understand this is only used for local development, so I am not sure security is more important than stability but, happy to make the change if you want.

services/horizon/docker/docker-compose.yml
2opremio: Note that it didn't have a password before, so this PR doesn't worsen the situation.

services/horizon/docker/docker-compose.yml
leighmcculloch: Agreed, but nobody wants their developer environment exploitable.

services/horizon/docker/docker-compose.yml
2opremio: Also note that adding a password now may break existing dev environments.

However, happy to make the change if that's what's decided.

leighmcculloch: That's a good point. I know that myself and others sometimes copy configurations from one place to another and sometimes they can make it all the way into a production environment. The environment variable is also easily misunderstood to be something that might not matter to have in a production environment. If it's not safe to add `POSTGRES_PASSWORD` now then lets run with it this, but I think if we can specify a password and configure the password in the Horizon default development config to match, that it is worthwhile.

tamirms: The integration tests for horizon create a connection to the horizon postgres db. The integration tests assume that the horizon db is not configured with a password:

https://github.com/stellar/go/blob/master/support/db/dbtest/postgres.go

If we add a password to the horizon postgres containers we would need to add the same password to the integration tests setup. We would also need to update the developing guide https://github.com/stellar/go/blob/master/services/horizon/internal/docs/developing.md#set-up-horizons-database to ensure that developers who are running postgres locally configure their instance with the same password.

I don't think it's worth the effort to implement all those changes in order to have the docker image use a password.

I don't think we need to make the docker compose configuration resemble the configuration that you would use in a production environment. Instead, it would be better to make explicit recommendations (e.g. make sure you set a password for your database) in the admin guide: https://github.com/stellar/go/blob/master/services/horizon/internal/docs/admin.md
We could also provide sample kubernetes manifests or a helm chart for those who would like to run horizon in production.

leighmcculloch: 👍 Makes a lot of sense.

services/horizon/docker/docker-compose.yml
leighmcculloch: It's worth noting that this is only pinning to a version of Postgres, not to a version of the Docker image. The version number is the version of Postgres, and the Docker images can change independent of that with any change to the official Dockerfiles.

Specifying the patch version number wouldn't have prevented this breakage from occurring because the breaking change was a change to the Docker image, and not Postgres. See this PR for more information that is adding a warning to their docs about this as a result of this break: docker-library/postgres#689.

2opremio: Your point doesn't hold, because tags with the patch number are not rebuilt, and thus can be considered immutable for all they are worth
.

leighmcculloch: Oh nice, TIL. Thanks!

PR #2304, opened 2020-02-21
clients/horizonclient: make HorizonTimeOut actually 60s

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Set the HorizonTimeOut time.Duration constant to 60 seconds.

Also remove some unnecessary casts to time.Duration because the values are already time.Durations.

### Why

It is currently set to 60 nanoseconds, but the comments claim it should be 60 seconds.

### Known limitations

N/A

### DIFF
--- a/clients/horizonclient/CHANGELOG.md
+++ b/clients/horizonclient/CHANGELOG.md
@@ -6,6 +6,7 @@ file.  This project adheres to [Semantic Versioning](http://semver.org/).
 ## Unreleased
 
 - Add `IsNotFoundError`
+- Change HorizonTimeOut to 60 seconds, which was set to 60 nanoseconds contrary to its documentation.
 
 ## [v2.0.0](https://github.com/stellar/go/releases/tag/horizonclient-v2.0.0) - 2020-01-13
 
--- a/clients/horizonclient/main.go
+++ b/clients/horizonclient/main.go
@@ -89,22 +89,22 @@ var (
 	HorizonTimeOut = 60 * time.Second
 
 	// MinuteResolution represents 1 minute used as `resolution` parameter in trade aggregation
-	MinuteResolution = 1 * time.Minute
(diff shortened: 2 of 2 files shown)


leighmcculloch:
So this PR isn't sufficient at fixing this confusion. It looks like the value does actually result in a timeout of 60s because of this: https://github.com/stellar/go/blob/b9639d223a4c2181dacc1ee311885c06c13890bd/clients/horizonclient/client.go#L60

This isn't good because the constant uses time.Durarion and so is expressing itself as 60 nanoseconds which we then multiply to be seconds. 

I won't merge this, I'll revisit this tomorrow.

leighmcculloch:
The reason I caught this was because I used the value in a http.Client as well, and 60s caused an immediate timeout in the client.

abuiles:
@leighmcculloch thanks!  I'll cherry-pick this commits into the horizonclient release 2.1

leighmcculloch:
Fixing this is not so straight forward. In fact I don't see anyway of fixing this in a non-breaking way. To summarize:

- `HorizonTimeOut` is a `time.Duration` which means whatever value it is given is in nanoseconds. It is given the value `60`, so it is `60` nanoseconds.

- In `Client` `sendRequestURL` we multiple the value so that it becomes seconds.
   https://github.com/stellar/go/blob/b9639d223a4c2181dacc1ee311885c06c13890bd/clients/horizonclient/client.go#L42-L70 

- So in actual fact the value `60` is in seconds, but the type we're using is communicating it is in nanoseconds to a reader.

I think the safest thing to do here is to change the value from a `time.Duration` to an `int`. It's a breaking change because the type changes, but it's a bug, not a feature deviation. If it impacts someone, it's very possible it _should_ impact them  because it might be a bug for them too.

Thoughts?

PR #2305, opened 2020-02-21
support/db/dbtest: Remove postgresql-client-9.6 dependency for running tests

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

The integration tests for horizon use `os/exec` to invoke `psql`, `createdb`, and `dropdb` to initialize postgres. Instead of invoking those executables we should use github.com/jmoiron/sqlx to initialize postgres.

### Why

This allows us to remove the postgresql-client-9.6 dependency which was only necessary for running Horizon tests (it wasn't required for running the Horizon service). Removing this dependency makes it easier for new Horizon developers because now they are able to run Horizon tests without having postgres installed locally (postgres can be running in a docker container instead).

### Known limitations

[N/A]


.circleci/config.yml
```diff
@@ -195,6 +157,7 @@ jobs:
           PGHOST: localhost
           PGPORT: 5432
           PGUSER: circleci
+          PGDATABASE: circle_test
```
tamirms: the database for the circleci/postgres:9.6.5-alpine-ram image is set to `circle_test`:

https://hub.docker.com/layers/circleci/postgres/9.6.5-alpine-ram/images/sha256-a09e3b13ae4ee0e44480704a2e8297ab7b119110c0cc3a17b60f61d9c15e1bc1?context=explore

services/horizon/internal/simplepath/order_book_test.go
```diff
@@ -56,28 +56,6 @@ func TestOrderBook(t *testing.T) {
 	})
 }
 
-func TestOrderBook_BadCost(t *testing.T) {
-	tt := test.Start(t).Scenario("bad_cost")
```
tamirms: the `bad_cost` scenario is different from the other scenarios. The scenario populates  its data using `COPY FROM` statements:

```
COPY offers (sellerid, offerid, sellingassettype, sellingassetcode, sellingissuer, buyingassettype, buyingassetcode, buyingissuer, amount, pricen, priced, price, flags, lastmodified) FROM stdin;
...
```

These statements fail to execute when using github.com/jmoiron/sqlx. I opted to remove this test entirely because I could not figure out what behavior the test was checking (I looked at the PR which introduced the test https://github.com/stellar-deprecated/horizon/pull/241/files but that wasn't so helpful). Also, in the new ingestion system we no longer rely on stellar core for populating /order_book responses. I think it doesn't make sense to preserve this test given that this legacy orderbook code will likely not be used again.

services/horizon/internal/test/scenarios/account_merge-core.sql
```diff
@@ -1,5 +1,3 @@
-running recipe
```
tamirms: I had to remove this prologue from all the scenarios generated by scc. "running recipe" and "recipe finished" are not valid sql statements and these strings caused failures when loading scenarios.

tamirms: addressed the underlying issue here: https://github.com/stellar/stellar_core_commander/pull/139

.circleci/config.yml
```diff
@@ -43,32 +43,6 @@ commands:
               ! egrep -irn -A 1 --include=*.go "Action.+needed.+in.+release:.+$CIRCLE_TAG" ./
             fi
 
-  install_stellar_core_debian_package:
```
tamirms: note that this job isn't used anymore. I think the dump ledger state job used this job previously but we have since moved the dump ledger state job out of circleci to kubernetes

support/db/dbtest/db.go
```diff
@@ -1,11 +1,47 @@
...
+	"testing"
+
 	"github.com/jmoiron/sqlx"
+	"github.com/lib/pq"
 	"github.com/stellar/go/support/db/sqlutils"
+	"github.com/stellar/go/support/errors"
 	"github.com/stretchr/testify/require"
 )
 
+// DB represents an ephemeral database that can be starts blank and can be used
```
bartekn:
```suggestion
// DB represents an ephemeral database that starts blank and can be used
```

.circleci/config.yml
```diff
@@ -120,18 +94,6 @@ commands:
...
-          command: |
-            sudo apt-get update
-            (
-              sudo apt-get install postgresql-client-9.6 || (
-                echo "deb http://apt.postgresql.org/pub/repos/apt/ $(env -i bash -c '. /etc/os-release; echo $VERSION_CODENAME')-pgdg main" | sudo tee /etc/apt/sources.list.d/pgdg.list &&
-                wget --quiet -O - https://www.postgresql.org/media/keys/ACCC4CF8.asc | sudo apt-key add - &&
-                sudo apt-get update &&
-                sudo apt-get install postgresql-client-9.6
-              )
-            )
```
leighmcculloch: I think this is left over when we had some of our builds run on Debian Buster and we're going to need to re-add this when we upgrade to Debian Buster when Stretch moves out of support in a few months. Would you mind leaving this in, or does this need to be yanked?

services/horizon/internal/simplepath/order_book_test.go
leighmcculloch: Is it possible to use the stdlib sql package instead of sqlx, does that work? What does SQLX give us in tests given we're just executing statements?

support/db/dbtest/db.go
```diff
@@ -46,3 +82,38 @@ func (db *DB) Open() *sqlx.DB {
...
+
+	pgUser := os.Getenv("PGUSER")
+	if len(pgUser) == 0 {
+		pgUser = "postgres"
+	}
+
+	// create the db
+	execStatement(t, pgUser, "CREATE DATABASE "+pq.QuoteIdentifier(result.dbName))
+
+	result.DSN = fmt.Sprintf("postgres://%s@localhost/%s?sslmode=disable", pgUser, result.dbName)
```
leighmcculloch: ❗️ Can we use the env vars that Postgres uses like `PGHOST`, `PGPORT`, etc here like you are above with `PGUSER`? The existing code which uses `createdb` picks those up, and those of us using docker development containers have been able to configure the tests to either hit a instance of postgres running inside the same container on a socket, or another container running postgres. I suspect this line is going to require a specific setup rather than allow for the flexibility we have today.

leighmcculloch: I realize now that we had localhost hardcoded before and somehow my setup works :shrug:‍ . Please ignore my comment. If I see issues after this merges I’ll take a look.

leighmcculloch: I realize now that we had localhost hardcoded before and somehow my setup works :shrug:‍ . Please ignore my comment. If I see issues after this merges I’ll take a look.

.circleci/config.yml
tamirms: @leighmcculloch this step installs postgresql-client-9.6 but given that we no longer need postgresql-client-9.6 I don't see why we still need to execute this step. Could you explain why this specific is still necessary?

leighmcculloch: Ohhhhhh! This is awesome!

tamirms:
> Is it possible to use the stdlib sql package instead of sqlx, does that work? What does SQLX give us in tests given we're just executing statements?

@leighmcculloch I think it should be possible to use the stdlib sql package. However, our tests are already coupled to sqlx as you can see below:

https://github.com/stellar/go/blob/master/support/db/dbtest/db.go#L43

https://github.com/stellar/go/blob/master/services/horizon/internal/test/main.go#L34

I kept using sqlx for the sake of consistency

tamirms:
> Is it possible to use the stdlib sql package instead of sqlx, does that work? What does SQLX give us in tests given we're just executing statements?

@leighmcculloch I think it should be possible to use the stdlib sql package. However, our tests are already coupled to sqlx as you can see below:

https://github.com/stellar/go/blob/master/support/db/dbtest/db.go#L43

https://github.com/stellar/go/blob/master/services/horizon/internal/test/main.go#L34

I kept using sqlx for the sake of consistency

Issue #2293, opened 2020-02-19
Create horizonclient release 3.0 once Horizon 1.0 has been merged into master


abuiles:
We should cherry-pick the following commits into it:

https://github.com/stellar/go/commit/ef52dffcb6600523ef410d59790000df03b49866 
https://github.com/stellar/go/commit/b973a23b767241b6d2a7fb3196c245e7ec96a004
https://github.com/stellar/go/commit/ababa11f23459aa6e7e7e8dee46778d49b9ee829

ire-and-curses:
For context: this is because we are releasing Horizon 2.1 with backwards compatibility, which means it depends on the older version of Horizon's protocols file. We're maintaining backwards compatibility because it would be unexpected for a user of the client to discover that their code broke when they upgraded on a minor release.

Once Horizon 1.0 is merged back to master, it's possible to create the new (breaking) 3.0 release branch for horizonclient, where we finally discontinue the pre-1.0 support. At that point, we need to pick up these commits from the divergent 2.1 branch to maintain functionality.

We'll plan to release this major release to sync with the upcoming stellar-core protocol 13, since that will likely force additional breaking changes to the SDK.

abuiles:
Done! 

The branch is here https://github.com/stellar/go/tree/release-horizonclient-v3.0.0 and it has all the relevant commits applied to it.

PR #2310, opened 2020-02-21
clients/horizonclient: fix inconsistent units of time in HorizonTimeOut 

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Remove HorizonTimeOut and its related functions and fields that operate
on a value of seconds and replace with HorizonTimeout that operates on a
value of nanoseconds.

### Why
The HorizonTimeOut identifier is defined as a time.Duration, and by
using that type we are stating that any value it holds is in
nanoseconds. Right now the code uses the type incorrectly to hold
seconds and this is confusing and misleading.

There is no good way to fix this because to change the type or the units
it holds is a breaking change. If we simply change the type someone
might see compiler errors. If we change the units someone might
unexpectedly continue to pass in seconds when they should now be using
nanoseconds. Alternatively, we can break this in a way that will cause
compiler errors and we can correct the units at the same time, ensuring
that the problem is fixed, fixed well, and any surprise is maybe
warranted because anyone using this field may also have a bug in their
own system.

The least breaking approach is definitely the first option above and to
just change this type from a `time.Duration` to a `uint` and to have it
continue to be seconds. But, given the likely rare use of this field,
and that someone else who is using it could easily have misunderstood
its meaning, it might be better to just break it and consider this like
a bug rather than a feature deviation.

By removing the function we ensure that anyone using it reassess their
use of it. It's easy to rename this identifier because the word 'Timeout'
is actually one word so we can call the new so we can use the one word
name 'Timeout' instead of 'TimeOut'.

### Known limitations

N/A

### DIFF
--- a/clients/horizonclient/CHANGELOG.md
+++ b/clients/horizonclient/CHANGELOG.md
@@ -5,6 +5,10 @@ file.  This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+- Fixed a bug where HorizonTimeOut has misleading units of time by:
+  - Removed HorizonTimeOut (seconds)
+  - Added HorizonTimeout (nanoseconds)
+
 ## [v2.1.0](https://github.com/stellar/go/releases/tag/horizonclient-v2.1.0) - 2020-02-24
 
 ### Add
--- a/clients/horizonclient/client.go
+++ b/clients/horizonclient/client.go
@@ -55,10 +55,10 @@ func (c *Client) sendRequestURL(requestURL string, method string, a interface{})
 	}
 	c.setClientAppHeaders(req)
 	c.setDefaultClient()
-	if c.horizonTimeOut == 0 {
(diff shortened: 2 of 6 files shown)


leighmcculloch:
We talked outside this PR and this change is going to be destined for horizonclient v3, and will be rebased onto the v3 branch once it is created, #2293.

leighmcculloch:
@abuiles Thanks for updating the base branch. I have resolved conflicts. If you folks are 👍 with this going into v3 let me know and I'll merge it, or please feel free to squash & merge it yourself to the branch.

abuiles:
👍

leighmcculloch:
@abuiles Could the release branch be forwarded to the current master to get the Go 1.14 changes?

abuiles:
@leighmcculloch sure thing!

PR #2311, opened 2020-02-21
clients/horizonclient: make HorizonTimeOut a uint

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Change the type of HorizonTimeOut and it's related functions and fields
from time.Duration to uint.

### Why
The value stored in the field is a count of seconds. The time.Duration
type explicitly stores a count of nanoseconds and so to use this type
with a different unit is confusing and misleading. I myself have used
the value thinking it was truly a time.Duration and had a unit of
nanoseconds. Other's may have made the same mistake.

While this is a breaking change I think it is fine to make in a patch
release as this is truly a bug. Any downstream breakages it causes is
probably warranted so that importers check that they are using the value
correctly in their own code.

Since the default clients come configured with a default timeout it is
also less likely people are using this field, so any impact from a
breaking change should be minimal.

### Known limitations

N/A

### DIFF
--- a/clients/horizonclient/CHANGELOG.md
+++ b/clients/horizonclient/CHANGELOG.md
@@ -6,6 +6,7 @@ file.  This project adheres to [Semantic Versioning](http://semver.org/).
 ## Unreleased
 
 - Add `IsNotFoundError`
+- Change `HorizonTimeOut` type from `time.Duration` to `uint`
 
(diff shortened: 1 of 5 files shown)


leighmcculloch:
@ire-and-curses has said he'd like to defer fixing this issue till v3, so in that case I'm going to stick with the better but more breaking solution #2310.

PR #2313, opened 2020-02-21
exp/services/recoverysigner: fix the horizon timeout

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Set the Horizon timeout to 1 minute for both the HTTP client and the horizonclient SDK.

### Why

While it looks like the same timeout is being used by the HTTP client and the horizonclient SDK it is not. The horizonclient SDK uses units of seconds and the HTTP client uses nanoseconds and the value being passed in to both is 60s. This causes the HTTP client to terminate most outgoing requests before they complete.

### Known limitations

This doesn't address the confusion that is this horizonclient timeout value which is being addressed in #2310.

### DIFF
--- a/exp/services/recoverysigner/internal/serve/serve.go
+++ b/exp/services/recoverysigner/internal/serve/serve.go
@@ -4,6 +4,7 @@ import (
 	"crypto/ecdsa"
 	"fmt"
 	"net/http"
+	"time"
 
 	firebaseauth "firebase.google.com/go/auth"
 	"github.com/go-chi/chi"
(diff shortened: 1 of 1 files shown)


exp/services/recoverysigner/internal/serve/serve.go
```diff
@@ -74,13 +75,15 @@ func getHandlerDeps(opts Options) (handlerDeps, error) {
 	}
 	opts.Logger.Info("SEP-10 JWT Public key: ", sep10JWTPublicKey)
 
+	horizonTimeout := 1 * time.Minute
+	httpClient := &http.Client{
+		Timeout: horizonTimeout,
+	}
```
marcelosalloum: Cool, defining the `httpClient` as a variable like this will make things clearer

leighmcculloch: 💯

PR #2314, opened 2020-02-22
support/http: allow services to specify the logger used by the logging middleware

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Allow services to specify the logger used by the logging middleware.

### Why

`LoggingMiddleware` gets the logger from the context using `"support/log".Ctx(...)`. If none is specified in the context the `"support/log".DefaultLogger` is used. Because the `LoggingMiddleware` is placed into the `chi.Mux` in the `NewMux` and `NewAPIMux` functions a service using these functions has no opportunity to set a logger on the context because a middleware cannot be inserted earlier. This means by using `NewMux` and `NewAPIMux` you are stuck with the default logger.

### Why Behind the Why

The reason I'm wanting this is because `webauth` and `recoverysigner` are similar to `ticker` in that they explicitly define all their dependencies and pass them down to the code using them as explicitly as possible. Because of this both of those applications define their own logger. Because we have some code like the LoggingMiddleware that magically pulls the logger out of the context setting it in the context is the only way to fully use a custom logger.

### Alternative Implementations

If you look at the commits on this PR I attempted to implement this one way first using functional options. It seems pretty cool 😎 but more complicated that is necessary so I dropped it.

### Known limitations

N/A

### DIFF
--- a/support/http/mux.go
+++ b/support/http/mux.go
@@ -4,15 +4,21 @@ import (
 	"github.com/go-chi/chi"
 	"github.com/go-chi/chi/middleware"
 	"github.com/rs/cors"
+	"github.com/stellar/go/support/log"
 )
 
 // NewMux returns a new server mux configured with the common defaults used across all
 // stellar services.
-func NewMux() *chi.Mux {
+func NewMux(options ...MuxOption) *chi.Mux {
+	muxOptions := NewMuxOptions(options...)
+
 	mux := chi.NewMux()
 
 	mux.Use(middleware.RequestID)
 	mux.Use(middleware.Recoverer)
+	if muxOptions.Logger != nil {
+		mux.Use(SetLoggerMiddleware(muxOptions.Logger))
+	}
 	mux.Use(LoggingMiddleware)
 
 	return mux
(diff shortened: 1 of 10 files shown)


support/http/mux.go
```diff
@@ -4,24 +4,28 @@ import (
...
 
 // NewMux returns a new server mux configured with the common defaults used across all
 // stellar services.
-func NewMux() *chi.Mux {
+func NewMux(l *log.Entry) *chi.Mux {
 	mux := chi.NewMux()
 
 	mux.Use(middleware.RequestID)
 	mux.Use(middleware.Recoverer)
+	if l != nil {
```
accordeiro: I'm a bit torn on this specific behavior. It feels like if I'm passing `nil` as a logger, my expectation would be for logs to not show up (as opposed to meaning "use the default logger").

If we're aiming to have this behavior more predictable and less magic, I'd favor us making the logger param required. Does that make sense?

leighmcculloch: Makes sense. c789584

PR #2316, opened 2020-02-22
services/horizon: expose date header

### LINKED ISSUE #2315: horizon: expose date header to CORS clients
Horizon clients might be interested in obtaining an accurate timestamp. See https://github.com/stellar/dashboard/issues/88. 

Horizon's HTTP responses do contain a `Date` header. However, this is not a CORS approved header and horizon does not expose it using [`Access-Control-Expose-Headers`](
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Expose-Headers
). 

I suggest fixing this by setting `Access-Control-Expose-Headers: Date` on all Horizon HTTP responses.

tomerweller:
closes #2315
~**do not merge yet**~

### DIFF
--- a/services/horizon/internal/web.go
+++ b/services/horizon/internal/web.go
@@ -118,6 +118,7 @@ func (w *web) mustInstallMiddlewares(app *App, connTimeout time.Duration) {
 	c := cors.New(cors.Options{
 		AllowedOrigins: []string{"*"},
 		AllowedHeaders: []string{"*"},
+		ExposedHeaders: []string{"Date"},
 	})
 	r.Use(c.Handler)
 


tomerweller:
I've tested and rebased

tomerweller:
The JS sdk already [uses horizon server time](https://github.com/stellar/js-stellar-sdk/blob/5607d31bd528bb46968f3d7e6ef79a7ea7c6f0f0/src/horizon_axios_client.ts#L40) but this functionality doesn't work inside of the browser due to `Date` not being exposed to CORS clients.

I definitely don't think that horizon should be used as a time server but if a client is already communicating with a horizon server I don't find it offensive to use that value as a sanity test.

PR #2319, opened 2020-02-24
services/ticker add stripped down ticker Dockerfile

jacekn:
* rename ticker Dockerfile to Dockerfile-dev. The new name is more appropriate because this is all-in-one image
* add stripped down Dockerfile that ships the ticker binary and minimal dependencies only
* add ticker Makefile


services/ticker/docker/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM golang:1.13-stretch as build
+
+ADD . /src/ticker
+WORKDIR /src/ticker
+RUN go build -o /bin/ticker ./services/ticker
+
+
+FROM ubuntu:16.04
+
+RUN apt-get update && apt-get install -y --no-install-recommends ca-certificates
```
accordeiro: I see that we're not running the `dependencies` shell script anymore. This should be fine since this is a stripped down version, but I'm wondering if `libpq-dev` is indeed optional, or if not installing it prevents the ticker from being able to use postgresql. Have you tested this connecting to an external psql instance @jacekn?

leighmcculloch: I haven't tried running Ticker locally but I don't think libpq-dev should be required. Ticker uses `github.com/lib/pq` which is a pure-Go implementation of a PG client. Looking at the other things installed I can't figure out what would require `libpq-dev`, so I could be wrong here.

accordeiro: That makes sense to me. Let's roll without it, worst case scenario we know which dependency might be missing later :)

jacekn:
> LGTM. Would the plan be to replace all those other elements of initializing Ticker into Kubernetes jobs if we needed to run them again outside of a dev context?

Yes exactly. It's still work in progress at this point but the idea is to use k8s cronjobs and k8s deployment for serving files. All those objects will use the same docker image (so same ticker binary)

services/ticker/docker/Dockerfile
jacekn: @accordeiro good point!

I confirmed that the image built using proposed Dockerfile can succesfuly run DB migrations so we should be good.

PR #2322, opened 2020-02-25
[WIP] exp/ingest Captive-stellar-core sketch

graydon:
This is an initial sketch of support for an ingestion backend driven by a captive stellar-core subprocess replaying ledgers in memory. It's going to need a bunch of iteration before it's viable, but this PR should serve as a place to examine & discuss the approach, fix bugs, etc.


ire-and-curses:
Moving discussion here from email:

Bartek said

> Hey Graydon!
> 
> I've run some tests with your branch. The good part is that I was able to easily replace existing DatabaseBackend with captiveStellarCore. I started with db reingest range command because it's the simplest command we can try to test new reader.
> 
> Unfortunately, I wasn't able to make it work. The following condition in the code always evaluates to true for any possible ledger range and returns an error:
> 
> if seq != c.nextLedger {
> 
> I always get "ledger not found" error but the actual error variable e is shadowed inside the loop so when you break higher level is still nil.
> 
> I think it would be great if you could rebase your branch on current master and create a draft PR so commenting will be much easier and I can also update the code so you can test it on your side as well. The other suggestions below:
> 
> I started it first with 12.2.0 without the new functionality and instead of returning an error the code just waited (I printed the log and got: "Unrecognised token: —replay-in-memory").
> It should be possible to pass ledger range. I noticed that currently it tries to download 100k of ledgers even if I request a small range like 63-127.

ire-and-curses:
Graydon said

>  There's no error message "ledger not found" in the file; what error
> are you actually getting?
> 
> It requires 12.3.0 for --replay-in-memory, and nothing will work
> without that. Sorry, should have specified that.
> 
> You should probably start by trying to get the unit test working. This
> should work (though it does take a minute or two):
> 
> $ go test ./exp/ingest/ledgerbackend -run TestCaptiveCore
> ok  github.com/stellar/go/exp/ingest/ledgerbackend 69.032
> 
> Does it not work for you with stellar-core 12.3.0 in your PATH?

exp/ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -0,0 +1,354 @@
...
+
+	var e error
+	for {
+		var xlcm xdr.LedgerCloseMeta
+		_, e = unmarshalFramed(c.metaPipe, &xlcm)
+		if e != nil {
+			e = errors.Wrap(e, "unmarshalling framed LedgerCloseMeta")
+			break
+		}
+		seq, e := peekLedgerSequence(&xlcm)
```
bartekn: `e` is shadowed here. So if there's an error anywhere below and you `break` the higher level `e` will be `nil`.

exp/ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -0,0 +1,354 @@
...
+		if e != nil {
+			break
+		}
+		if seq != c.nextLedger {
+			// We got something unexpected; close and reset
+			e = errors.Errorf("unexpected ledger %d", seq)
```
bartekn: Not sure why but when I run `horizon db reingest range` using this backend I always go to this code path. See my next commit to reproduce.

bartekn:
> It requires 12.3.0 for --replay-in-memory, and nothing will work
without that. Sorry, should have specified that.

@graydon I understand, what I was trying to say is that we should probably return an error in such case. Currently the code just hangs.

exp/ingest/ledgerbackend/captive_core_backend_posix.go
```diff
@@ -0,0 +1,64 @@
...
+	e = c.cmd.Start()
+	if e != nil {
+		return e
+	}
+
+	// Launch a goroutine to reap immediately on exit (I think this is right,
+	// as we do not want zombies and we might abruptly forget / kill / close
+	// the process, but I'm not certain).
+	go func() {
+		c.cmd.Process.Wait()
```
tamirms: I'm wondering if we should call `c.cmd.Wait()` instead of `c.cmd.Process.Wait()`. 

https://golang.org/pkg/os/exec/#Cmd.Wait

exp/ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -0,0 +1,354 @@
...
+
+// TODO: switch from history URLs to history archive interface provided from support package, to permit mocking
+
+// In this (crude, initial) sketch, we replay ledgers in blocks of 100,000
+const ledgersPerProcess = 100000
+const ledgersPerCheckpoint = 64
+const numLogLinesToKeep = 1024
+
+// Circular buffer of recent log lines from captive stellar core
+type recentLogLines struct {
```
tamirms: how will we use logs from stellar core? if it's only used for debugging purposes would it make sense to just print the lines immediately to horizon's stdout when a verbose flag is enabled?

exp/ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -0,0 +1,354 @@
...
+}
+
+// We assume that we'll be called repeatedly asking for ledgers in ascending
+// order, so when asked for ledger 23 we start a subprocess doing catchup
+// "100023/100000", which should replay 23, 24, 25, ... 100023. The wrinkle in
+// this is that core will actually replay from the _checkpoint before_
+// the implicit start ledger, so we might need to skip a few ledgers until
+// we hit the one requested (this routine does so transparently if needed).
+func (c *captiveStellarCore) GetLedger(sequence uint32) (bool, LedgerCloseMeta, error) {
+	if c.nextLedger != sequence {
```
tamirms: I understand that if `sequence < c.nextLedger` we want to close and restart stellar core. 

But, if  `sequence == c.nextLedger + n` where `n` is small, it seems that it would be better to keep the existing stellar core process alive and discard all the incoming ledgers from the pipe until we arrive at our desired sequence. I'm assuming that overhead of starting up a new instance of stellar core (since we need to download the nearest checkpoint ledger again) is greater than seeking ahead until we obtain our desired ledger.

exp/ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -0,0 +1,354 @@
...
+			return false, LedgerCloseMeta{}, errors.Errorf("sequence %d greater than max available %d",
+				sequence, maxLedger)
+		}
+		lastLedger := sequence + ledgersPerProcess
+		if lastLedger > maxLedger {
+			lastLedger = maxLedger
+		}
+		cmd := exec.Command("stellar-core",
+			"--conf", c.getConfFileName(),
+			"catchup", fmt.Sprintf("%d/%d", lastLedger, lastLedger - sequence),
```
tamirms: if `sequence` is equal to `maxLedger` then the catchup argument will be `n/0`. is it ok that the `LEDGER-COUNT` is 0?

exp/ingest/ledgerbackend/captive_core_backend.go
graydon: fixed

exp/ingest/ledgerbackend/captive_core_backend.go
graydon: fixed (got subsequent commit to work reliably!)

exp/ingest/ledgerbackend/captive_core_backend_posix.go
graydon: sure, fixed.

exp/ingest/ledgerbackend/captive_core_backend.go
graydon: fixed (plumbed logs directly into horizon, removed this part)

exp/ingest/ledgerbackend/captive_core_backend.go
graydon: fixed (refactored loop to close only when out of range, then reopen only when closed).

exp/ingest/ledgerbackend/captive_core_backend.go
graydon: fixed the arithmetic here.

graydon:
Pushed updates with several fixes (comments above plus some edge cases I discovered myself). Still waiting on some changes to land on core's side but this is able to ingest many different ranges (including 63 127 but also ones that actually have content!) on my workstation.

ire-and-curses:
This is great!

I see the builds are failing, error is `Go file go.mod is dirty, update the file with 'go mod tidy' locally.`

ire-and-curses:
Related, Graydon's stellar-core range fix: https://github.com/stellar/stellar-core/pull/2465

exp/ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -0,0 +1,429 @@
...
+
+	// Next, if we're closed, open.
+	if c.IsClosed() {
+		if e := c.openOfflineReplaySubprocess(sequence, sequence+ledgersPerProcess); e != nil {
+			return false, LedgerCloseMeta{}, errors.Wrap(e, "opening subprocess")
+		}
+	}
+
+	// Check that we're where we expect to be: in range ...
+	if !c.LedgerWithinCheckpoints(sequence, 1) {
```
tamirms, line 251: it seems like this check should be inside the `if c.IsClosed() {` statement above because we already check `c.LedgerWithinCheckpoints(sequence, numCheckpointsLeeway)` at the top of the function

exp/ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -0,0 +1,429 @@
...
+		lcm.TransactionMeta = append(lcm.TransactionMeta, trm.TxApplyProcessing)
+		lcm.TransactionFeeChanges = append(lcm.TransactionFeeChanges, trm.FeeProcessing)
+	}
+	for _, urm := range v0.UpgradesProcessing {
+		lcm.UpgradesMeta = append(lcm.UpgradesMeta, urm.Changes)
+	}
+	return nil
+}
+
+func (c *captiveStellarCore) openOfflineReplaySubprocess(nextLedger, lastLedger uint32) error {
```
tamirms, line 174: `lastLedger` is a bit confusing as a variable name because it gives the impression that `lastLedger` precedes `nextLedger`. 

I think it would be better to rename `nextLedger, lastLedger uint32` to `fromLedger, toLedger uint32`

services/horizon/internal/expingest/fsm.go
```diff
@@ -627,6 +627,23 @@ func (h reingestHistoryRangeState) run(s *System) (transition, error) {
 		return stop(), errors.Errorf("invalid range: [%d, %d]", h.fromLedger, h.toLedger)
 	}
 
+	log.WithFields(logpkg.F{
+		"from": h.fromLedger,
+		"to":   h.toLedger,
+	}).Info("Preparing ledger backend to retrieve range")
+	startTime := time.Now()
+
+	err := s.ledgerBackend.PrepareRange(h.fromLedger, h.toLedger)
```
tamirms, line 636: what if the ledger range is huge, e.g. we're reingesting the entire history? will  `PrepareRange()` take a really long time to complete in that case?

exp/ingest/ledgerbackend/captive_core_backend.go
bartekn: Moved the comments to a clean up issue: https://github.com/stellar/go/issues/2551#issuecomment-631743617

exp/ingest/ledgerbackend/captive_core_backend.go
bartekn: Moved the comments to a clean up issue: https://github.com/stellar/go/issues/2551#issuecomment-631743617

services/horizon/internal/expingest/fsm.go
bartekn: Moved the comments to a clean up issue: https://github.com/stellar/go/issues/2551#issuecomment-631743617

PR #2323, opened 2020-02-26
exp/services/recoverysigner: remove type, add identities

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Remove the `type` parameter stored alongside registered accounts and replace it with exposing information about which identities are set on an account to clients.

Remove the `type` request and response parameters. Add the `identities` response parameter, and replace `owner_identities` and `other_identities` with `identities.owner` and `identities.other`.

### Why
The `type` existed purely for clients to get some sense of what purpose an account was registered, but it's really meta data about how a client is using the service and isn't necessarily a core concept for the
service.

The value was also synonymous with which identities are configured for an account. A `type` of `personal` will always have only an `owner` identity, and a `type` of `share` was always going to have an `other` identity alone or in addition to `owner`.

By removing the concept of `type` and instead just giving a client visibility into which identities are set on the account we reduce the number of concepts a client has to learn and know about which makes the API simpler, even if the change appears to make the API slightly more complex.

The reason I replaced `owner_identities` and `other_identities` is while thinking about the right way to expose the identities to the client it occurred to me the simplest method for a JavaScript client would be to have the identities be keys of an object, and so I structured the response that way, and it made sense to also modify the request to match.

### Known limitations

This is a breaking change to the API, but there are no consumers and the API is experimental.

### DIFF
--- a/exp/services/recoverysigner/internal/account/account.go
+++ b/exp/services/recoverysigner/internal/account/account.go
@@ -2,7 +2,6 @@ package account
 
 type Account struct {
 	Address         string
-	Type            string
 	OwnerIdentities Identities
 	OtherIdentities Identities
 }
@@ -11,4 +10,10 @@ type Identities struct {
 	Address     string
 	PhoneNumber string
 	Email       string
+}
+
+// Present indicates if the Identities contains at least one identity. Returns
+// false if it is an empty/zero value.
+func (i Identities) Present() bool {
+	return i != Identities{}
 }
--- a/exp/services/recoverysigner/internal/serve/account_list.go
+++ b/exp/services/recoverysigner/internal/serve/account_list.go
@@ -45,8 +45,15 @@ func (h accountListHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 			return
 		} else {
 			accResp := accountResponse{
-				Address:  acc.Address,
-				Type:     acc.Type,
+				Address: acc.Address,
+				Identities: accountResponseIdentities{
+					Owner: accountResponseIdentity{
+						Present: acc.OwnerIdentities.Present(),
+					},
+					Other: accountResponseIdentity{
+						Present: acc.OtherIdentities.Present(),
+					},
+				},
 				Identity: "account",
 				Signer:   h.SigningAddress.Address(),
 			}
@@ -69,8 +76,15 @@ func (h accountListHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 				identity = "other"
 			}
 			accResp := accountResponse{
-				Address:  acc.Address,
-				Type:     acc.Type,
+				Address: acc.Address,
+				Identities: accountResponseIdentities{
+					Owner: accountResponseIdentity{
+						Present: acc.OwnerIdentities.Present(),
+					},
+					Other: accountResponseIdentity{
+						Present: acc.OtherIdentities.Present(),
+					},
+				},
 				Identity: identity,
 				Signer:   h.SigningAddress.Address(),
 			}
@@ -95,8 +109,15 @@ func (h accountListHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 				identity = "other"
 			}
 			accResp := accountResponse{
-				Address:  acc.Address,
-				Type:     acc.Type,
+				Address: acc.Address,
+				Identities: accountResponseIdentities{
+					Owner: accountResponseIdentity{
+						Present: acc.OwnerIdentities.Present(),
+					},
+					Other: accountResponseIdentity{
+						Present: acc.OtherIdentities.Present(),
+					},
+				},
 				Identity: identity,
 				Signer:   h.SigningAddress.Address(),
 			}
@@ -121,8 +142,15 @@ func (h accountListHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 				identity = "other"
 			}
 			accResp := accountResponse{
-				Address:  acc.Address,
-				Type:     acc.Type,
+				Address: acc.Address,
+				Identities: accountResponseIdentities{
+					Owner: accountResponseIdentity{
+						Present: acc.OwnerIdentities.Present(),
+					},
+					Other: accountResponseIdentity{
+						Present: acc.OtherIdentities.Present(),
+					},
+				},
 				Identity: identity,
 				Signer:   h.SigningAddress.Address(),
 			}
--- a/exp/services/recoverysigner/internal/serve/account_response.go
+++ b/exp/services/recoverysigner/internal/serve/account_response.go
@@ -1,8 +1,18 @@
 package serve
 
 type accountResponse struct {
-	Address  string `json:"address"`
-	Type     string `json:"type"`
-	Identity string `json:"identity"`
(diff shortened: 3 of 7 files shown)


exp/services/recoverysigner/internal/serve/account_list.go
```diff
@@ -45,8 +45,15 @@ func (h accountListHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-				Type:     acc.Type,
+				Address: acc.Address,
+				Identities: accountResponseIdentities{
+					Owner: accountResponseIdentity{
+						Present: acc.OwnerIdentities.Present(),
+					},
+					Other: accountResponseIdentity{
+						Present: acc.OtherIdentities.Present(),
+					},
+				},
```
leighmcculloch: You'll notice this code block is duplicate a bit. I could reduce some code here, but I'm going to expend more effort holistically looking at reducing the duplicity here. Follow up: #2324.

howardtw:
👀

exp/services/recoverysigner/internal/serve/account_response.go
```diff
@@ -1,8 +1,18 @@
...
-	Signer   string `json:"signer"`
+	Address    string                    `json:"address"`
+	Type       string                    `json:"-"`
+	Identities accountResponseIdentities `json:"identities"`
+	Identity   string                    `json:"identity"`
+	Signer     string                    `json:"signer"`
+}
+
+type accountResponseIdentities struct {
+	Owner accountResponseIdentity `json:"owner"`
```
howardtw: I wonder if we can make these two fields `omitempty` instead of using the `Present` bool in another struct. If there's one field presented in the `accountResponseIdentities`, then the client should be able to tell whether it's `owner` or `other`.

exp/services/recoverysigner/internal/serve/account_response.go
```diff
@@ -1,8 +1,18 @@
...
 
 type accountResponse struct {
-	Address  string `json:"address"`
-	Type     string `json:"type"`
-	Identity string `json:"identity"`
-	Signer   string `json:"signer"`
+	Address    string                    `json:"address"`
+	Type       string                    `json:"-"`
+	Identities accountResponseIdentities `json:"identities"`
+	Identity   string                    `json:"identity"`
```
howardtw: What is this field used for?

exp/services/recoverysigner/internal/serve/account_response.go
```diff
@@ -1,8 +1,18 @@
 package serve
 
 type accountResponse struct {
-	Address  string `json:"address"`
-	Type     string `json:"type"`
-	Identity string `json:"identity"`
-	Signer   string `json:"signer"`
+	Address    string                    `json:"address"`
+	Type       string                    `json:"-"`
```
howardtw: Are we able to remove this field?

exp/services/recoverysigner/internal/serve/account_response.go
leighmcculloch: I like that too, but our frontend team have mentioned before that it's much easier from a clients perspective is fields are always present and have a clear signal that something isn't available.

howardtw: Got it. If that's the case, would making it `IsOwner bool` and `IsOther bool` be sufficient?

exp/services/recoverysigner/internal/account/account.go
```diff
@@ -12,3 +11,9 @@ type Identities struct {
 	PhoneNumber string
 	Email       string
 }
+
+// Present indicates if the Identities contains at least one identity. Returns
+// false if it is an empty/zero value.
+func (i Identities) Present() bool {
```
marcelosalloum: Nit: I personally like leading bool methods with `Is`, just to make things clearer:
```suggestion
func (i Identities) IsPresent() bool {
```

exp/services/recoverysigner/internal/serve/account_response.go
leighmcculloch: It tells the caller what their identity is. If the owner is authenticated it will say owner. If the other is authenticated it'll say other. I've started writing a spec that I'll open later today.

exp/services/recoverysigner/internal/serve/account_response.go
leighmcculloch: It is another way to solve the same goal, but it doesn't communicate the same intent. The account is not an owner, or an other, it has an owner identity, or an other identity. If we ever want to expose more details about the owner or other the object let's us expand by adding more fields alongside present. 

Do you have any specific concerns?

howardtw: I usually avoid having nested objects that are too deep. However, if the goal is to incorporate more fields in the future, it makes sense. I still feel like the cleanest way is to make it the way I suggested in the first [place](https://github.com/stellar/go/pull/2323#discussion_r384677261) as `Present` feels redundant. Let me confirm with the frontend team again.

leighmcculloch: I like the omitempty approach too. If I do that, what do we do when the thing is present but has no fields? It will be like this: `"owner": {}` which seems a bit odd and I think is harder to test for in JavaScript? Being able to do `owner.present` for `true`/`false` explicitly seems really simple.

exp/services/recoverysigner/internal/serve/account_response.go
howardtw: I think the client can tell by looking at `Identities` tho? And by looking at the test [cases](https://github.com/stellar/go/pull/2323/files#diff-3d3a2f8f712a84966724ba79351cfd15R62-R66), it seems possible to have both? That test case also shows `account`.

exp/services/recoverysigner/internal/serve/account_response.go
howardtw: > It will be like this: "owner": {} which seems a bit odd and I think is harder to test for in JavaScript.

Maybe we can make the backend never do that?

> Being able to do owner.present for true/false explicitly seems really simple.

Sure. It's not a blocker. Approving it.

exp/services/recoverysigner/internal/serve/account_response.go
leighmcculloch: 👍 822f8bc6

exp/services/recoverysigner/internal/account/account.go
leighmcculloch: I've liked this pattern too in other languages in the past, but in Go code it's more common to omit the `Is`. Grabbing a random example the [`Response`](https://godoc.org/net/http#Response) type has `Close` and `Uncompressed`, but there are loads of examples of this in the stdlib and other packages.

exp/services/recoverysigner/internal/serve/account_response.go
leighmcculloch: It will be `account` if you are authenticated as the account and not one of the identities.

The client cannot look at `Identities` because we don't expose the identities to the user. This is because we'd be exposing potentially sensitive information to another party.

PR #2330, opened 2020-02-26
horizonclient: add GetError func for retrieving a horizon error

howardtw:
### What

This PR adds a helper function `GetError` to check whether if an error can be interpreted as a horizon-related error. If it can't, a nil error is returned.

### Why

People who use `horizonclient` to submit a transaction often need to check the error returned and look into what the problem is when something went wrong. However, there are two types of errors that can be returned by this package, one is Error, and the other is `*Error`. This function function makes it convenient to always get back a `*Error` even if the input is `Error`.

### Known limitations

N/A

### DIFF
--- a/clients/horizonclient/error_helpers.go
+++ b/clients/horizonclient/error_helpers.go
@@ -19,3 +19,19 @@ func IsNotFoundError(err error) bool {
 
 	return hErr.Problem.Type == "https://stellar.org/horizon-errors/not_found"
 }
+
+// GetError returns an error that can be interpreted as a horizon-specific
+// error. If err cannot be interpreted as a horizon-specific error, a nil error
+// is returned. The caller should still check whether err is nil.
+func GetError(err error) *Error {
+	var hErr *Error
+
+	switch e := err.(type) {
+	case *Error:
+		hErr = e
+	case Error:
+		hErr = &e
+	}
+
+	return hErr
+}
--- a/clients/horizonclient/error_helpers_test.go
+++ b/clients/horizonclient/error_helpers_test.go
@@ -82,3 +82,72 @@ func TestIsNotFoundError(t *testing.T) {
 		})
 	}
 }
+
+func TestGetError(t *testing.T) {
+	testCases := []struct {
+		desc    string
+		err     error
+		wantErr error
+	}{
+		{
+			desc:    "nil error",
+			err:     nil,
+			wantErr: nil,
+		},
+		{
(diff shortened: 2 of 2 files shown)


clients/horizonclient/error_helpers.go
```diff
@@ -19,3 +19,19 @@ func IsNotFoundError(err error) bool {
...
+// is returned. The caller should still check whether err is nil.
+func GetError(err error) *Error {
+	var hErr *Error
+
+	switch e := err.(type) {
+	case *Error:
+		hErr = e
+	case Error:
+		hErr = &e
+	}
```
leighmcculloch: There's one other use case that thereotically could happen, but is unlikely to happen. If the err is one of these two types, but is a `nil` `*horizonclient.Error`, then `err` will not be `nil`, but the return value of this function will be `nil`. It could be confusing for the caller to get `nil` back from this, but then `err != nil` to be true, but the err is technically a `*horizonclient.Error`. I don't think we can do anything about this.

clients/horizonclient/error_helpers_test.go
```diff
@@ -82,3 +82,72 @@ func TestIsNotFoundError(t *testing.T) {
...
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.desc, func(t *testing.T) {
+			gotErr := GetError(tc.err)
+			if tc.wantErr == nil {
+				assert.Nil(t, gotErr)
+			} else {
+				assert.Equal(t, tc.wantErr, gotErr)
+			}
```
leighmcculloch: You can test if gotErr matches wantErr in the case that wantErr is nil, so I don't think if you need the if/else block.
```suggestion
			assert.Equal(t, tc.wantErr, gotErr)
```

If you're having trouble with comparing some error types, you're better off making `wantErr` a `string` of the error output, and then use `assert.EqualError`.

clients/horizonclient/error_helpers.go
howardtw: When err is `(*Error)(nil)`, we get back `(*Error)(nil)` from this function, which is still nil, meaning that the_returned_err == nil is still true. However, when we use `assert.Equal(t, nil, returnedErr)`, the result is false given the types are different. That's why I wrote the test this way.

clients/horizonclient/error_helpers_test.go
howardtw: https://github.com/stellar/go/pull/2330/files#r384829801

PR #2337, opened 2020-02-28
exp/services/recoverysigner: add form url-encoded tests

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Add tests that test that form url-encoded requests are supported.

### Why
The application is intended to support both form url-encoded and JSON request bodies, and actually does support it because our `support/http/httpdecode` package supports it, but no tests exist exercising the former.

### Known limitations

[TODO or N/A]

### DIFF
--- a/exp/services/recoverysigner/internal/serve/account_post_test.go
+++ b/exp/services/recoverysigner/internal/serve/account_post_test.go
@@ -5,6 +5,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
+	"net/url"
 	"strings"
 	"testing"
 
@@ -17,7 +18,7 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestAccountPost_new(t *testing.T) {
+func TestAccountPost_newContentTypeJSON(t *testing.T) {
 	s := account.NewMemoryStore()
 	h := accountPostHandler{
 		Logger:         supportlog.DefaultLogger,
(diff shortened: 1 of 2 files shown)


exp/services/recoverysigner/internal/serve/account_post_test.go
```diff
@@ -86,6 +87,70 @@ func TestAccountPost_new(t *testing.T) {
...
+	r.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+
+	w := httptest.NewRecorder()
+	m := chi.NewMux()
+	m.Post("/{address}", h.ServeHTTP)
+	m.ServeHTTP(w, r)
+	resp := w.Result()
+
+	require.Equal(t, http.StatusOK, resp.StatusCode)
+	assert.Equal(t, "application/json; charset=utf-8", resp.Header.Get("Content-Type"))
```
marcelosalloum: Just out of curiosity, what's the difference between `assign.Equal` and `require.Equal`? Would the code stop executing as soon as a `require` is not met, while the `assign` will fail the test but continue executing the code?

marcelosalloum:
LGTM

exp/services/recoverysigner/internal/serve/account_post_test.go
leighmcculloch: The functions in the `assert` and `require` make the same assertions, but what they do with the assertions if they fail changes exactly as you described. They align to the `t.Error` and `t.Fatal` functions respectively. To read more about them see https://godoc.org/github.com/stretchr/testify, and to read more about the `t` functions checkout https://godoc.org/testing.

exp/services/recoverysigner/internal/serve/account_post_test.go
```diff
@@ -86,6 +87,70 @@ func TestAccountPost_new(t *testing.T) {
...
+	ctx = auth.NewContext(ctx, auth.Auth{Address: "GDIXCQJ2W2N6TAS6AYW4LW2EBV7XNRUCLNHQB37FARDEWBQXRWP47Q6N"})
+	reqValues := url.Values{}
+	reqValues.Set("identities.owner.account", "GBF3XFXGBGNQDN3HOSZ7NVRF6TJ2JOD5U6ELIWJOOEI6T5WKMQT2YSXQ")
+	reqValues.Set("identities.owner.phone_number", "+10000000000")
+	reqValues.Set("identities.owner.email", "user1@example.com")
+	reqValues.Set("identities.other.account", "GB5VOTKJ3IPGIYQBJ6GVJMUVEAIYGXZUJE4WYLPBJSHOTKLZTETBYOBI")
+	reqValues.Set("identities.other.phone_number", "+20000000000")
+	reqValues.Set("identities.other.email", "user2@example.com")
+	req := reqValues.Encode()
+	t.Log("Request Body:", req)
```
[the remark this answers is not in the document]
leighmcculloch: Yup it's intended. We don't do a great job of this but you can use the testing `Log` function to make the test describe itself during the success / fail cases. Can be useful especially in the case that test setup data has been generated and isn't hardcoded, because somebody can run a test in verbose mode and see it succeed and all of the inputs. That's essentially what I did here. It's completely fine to test good testing logs.

exp/services/recoverysigner/internal/serve/account_post_test.go
marcelosalloum: Thanks!

Issue #2331, opened 2020-02-26
Inconsistent ledger view in `/accounts/{id}` when streaming 


bartekn:
I noticed `Some extra balances` error in `/accounts/{id}` stream. At first, I thought that it was the inconsistency issue propagated from stellar-core (we build Account object using a few DB queries without `REPEATABLE READ` transaction). But then I realized that we don't use `REPEATABLE READ` transaction in Horizon DB while streaming.

We should update the streaming code to start a `REPEATABLE READ` transaction (if not in a transaction already) before calling an action function and then rollback (only if not in a previously started transaction).

Additionally, we may do the same thing for a function getting data from core inside the action however we should first check if it affects core DB in any way.

bartekn:
> Additionally, we may do the same thing for a function getting data from core inside the action however we should first check if it affects core DB in any way.

We've seen another instance today for non-streaming request and the reason is getting data from core in multiple requests. So we should probably add start a new `REPEATABLE READ` transaction when getting data from Stellar-Core.

PR #2344, opened 2020-02-28
services/horizon/internal/actions: Wrap SSE handlers in repeatable read tx

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

* Execute SSE handlers in repeatable read transactions
* Obtain account info from stellar core db using repeatable read transaction 

### Why

Close https://github.com/stellar/go/issues/2331

> I noticed Some extra balances error in /accounts/{id} stream. At first, I thought that it was the inconsistency issue propagated from stellar-core (we build Account object using a few DB queries without REPEATABLE READ transaction). But then I realized that we don't use REPEATABLE READ transaction in Horizon DB while streaming.

> We should update the streaming code to start a REPEATABLE READ transaction (if not in a transaction already) before calling an action function and then rollback (only if not in a previously started transaction).

...
> We've seen another instance today for non-streaming request and the reason is getting data from core in multiple requests. So we should probably add start a new REPEATABLE READ transaction when getting data from Stellar-Core.

### Known limitations

[N/A]

### DIFF
--- a/services/horizon/internal/handler.go
+++ b/services/horizon/internal/handler.go
@@ -4,12 +4,14 @@ import (
 	"bytes"
 	"context"
 	"crypto/sha256"
+	"database/sql"
 	"encoding/json"
 	"net/http"
 	"strings"
 	"time"
 
 	"github.com/stellar/go/services/horizon/internal/actions"
+	horizonContext "github.com/stellar/go/services/horizon/internal/context"
 	"github.com/stellar/go/services/horizon/internal/db2"
 	"github.com/stellar/go/services/horizon/internal/hchi"
 	"github.com/stellar/go/services/horizon/internal/ledger"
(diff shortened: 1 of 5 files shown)


services/horizon/internal/handler.go
```diff
@@ -413,6 +416,31 @@ func (handler streamableObjectActionHandler) ServeHTTP(
...
+	r *http.Request,
+	generateEvents sse.GenerateEventsFunc,
+) sse.GenerateEventsFunc {
+	var session db.SessionInterface
+	if val := r.Context().Value(&horizonContext.SessionContextKey); val != nil {
+		session = val.(db.SessionInterface)
+	}
+
+	return func() ([]sse.Event, error) {
+		if session != nil {
```
abuiles: @tamirms under which scenarios would `session` be nil?

tamirms: all horizon handlers should have a non nil session. This check is only included because some of the handlers in services/horizon/internal/stream_handler_test.go do not include a session in the request context

PR #2348, opened 2020-03-02
services/horizon: Add HISTORY_ARCHIVE_URLS to docker-compose

2opremio:
Followup from https://github.com/stellar/go/pull/2306

Running `docker/start.sh` led to error `--history-archive-urls must be set when --ingest is set`

### DIFF
--- a/services/horizon/docker/docker-compose.yml
+++ b/services/horizon/docker/docker-compose.yml
@@ -57,6 +57,9 @@ services:
       - STELLAR_CORE_DATABASE_URL=postgres://postgres:mysecretpassword@host.docker.internal:5641/stellar?sslmode=disable
       # to connect to the public stellar network
       # - NETWORK_PASSPHRASE=Public Global Stellar Network ; September 2015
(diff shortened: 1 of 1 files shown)


2opremio:
Should we add a minimal test for the `start.sh` script in CI?

ire-and-curses:
See https://github.com/stellar/docker-stellar-core-horizon/issues/160 - fix is in review

Adding the test sounds like a good idea.

ire-and-curses:
Actually this is a different docker isn't it @tamirms?

tamirms:
@ire-and-curses the same fix needs to be applied to both the quickstart image (which runs horizon, stellar core, and their databases all in one image) and the docker compose workflow.

@2opremio provided a fix for the docker compose workflow. The docker compose workflow exists for development because it can run horizon from source. The quickstart image does not run horizon from source; the quickstart image is published for every horizon release.

abuiles:
@tamirms I think we are already handling this in the quick start image.

Issue #2352, opened 2020-03-02
support/historyarchive - Fails instead of retrying on http 503 response


nyetwurk:
### What version are you using?
master

### What did you do?
use history archive on the http endpoint of a google storage bucket

### What did you expect to see?
Retry when a GET results in a 503

### What did you see instead?
historyarchive fails.

bartekn:
I think that retry code should be in the caller. Which part of the history archive you are trying to download?

nyetwurk:
I looked into doing it in the callers, but there are a LOT of callers.

Also, it doesn't matter which part of the archive you're trying to retrieve, 503 can happen on literally any http request.

nyetwurk:
Also, the `resp.StatusCode` never makes it to the caller, so they can't check specifically for 503, which needs to be handled differently from all other errors, which are generally outright unretryable failures.

leighmcculloch:
I think the retry code should be in the caller as well, or we should define the retry logic in the HTTP Client, or in a separate type that provides that capability on-top of the type. If the retry logic is added in between the caller and the http archive code it keeps the archive code simple.

There are some open source packages that could help do this that could be explored, for example https://github.com/sethgrid/pester.

jacekn:
On top of handling 503s it would be good to support rate limiting gracefully (http 429). Currently archivist panics when it gets 429.
```
panic: Unkown status code=429
```

PR #2353, opened 2020-03-02
Issue #2352 - WIP: historyarchive: Retry on http 503 response

nyetwurk:
POC fix for #2352

WIP: Retry forever on 503
1) should give up eventually
2) should do a backoff

Please don't merge, this is just an example of how this could be fixed.

### DIFF
--- a/support/historyarchive/http_archive.go
+++ b/support/historyarchive/http_archive.go
@@ -38,12 +38,20 @@ func (b *HttpArchiveBackend) GetFile(pth string) (io.ReadCloser, error) {
 		return nil, err
 	}
 	req = req.WithContext(b.ctx)
-	resp, err := b.client.Do(req)
-	if err != nil {
-		if resp != nil && resp.Body != nil {
-			resp.Body.Close()
+	var resp *http.Response
+	for {
+		resp, err := b.client.Do(req)
+		if err != nil {
+			if resp != nil && resp.Body != nil {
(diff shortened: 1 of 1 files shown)


support/historyarchive/http_archive.go
```diff
@@ -38,12 +38,20 @@ func (b *HttpArchiveBackend) GetFile(pth string) (io.ReadCloser, error) {
...
+				resp.Body.Close()
+			}
+			return nil, err
 		}
-		return nil, err
+		if resp.StatusCode != 503 {
+			break
+		}
+		// FIXME: exponential backoff
+		// sleep(exponential_backoff)
```
nyetwurk: 1) should sleep with exponential pseudo-random backoff
2) should eventually give up

nyetwurk: Also, this might be better done in `.Do()` itself, since there are other things that can cause a 503

2opremio:
Do you still want to pursue this or shall we close it?

nyetwurk:
I was hoping someone would pick up where I left off, it still affects us.

leighmcculloch:
I think the problem can be solved but we probably should take another route. @bartekn and I commented over on the issue with some thoughts. Let's discuss there and see what we can find. See https://github.com/stellar/go/issues/2352#issuecomment-700960427.

Issue #2249, opened 2020-02-10
Create cron kubernetes job that runs `horizon-cmp history`


bartekn:
Create a cron (every 5/10 minutes) kubernetes job that runs #2227:
```
horizon-cmp -b [history.stellar.org] -t [horizon-stg-pubnet] history
```
By default it checks last 120 ledgers (around 10 mins) so ranges should overlap to not miss any ledges. If non-zero exit code is returned it should send a Slack notification.

abuiles:
@bartekn  I was discussing this with @tamirms and he suggested a different approach -- instead of doing this as a cron job, we could create a service that runs horizon-cmp and then exposes the results as prometheus metrics.

I think that approach is more robust since it might also leave us with some shared infra that we can use for other tasks of this kind and we can built on top of the alerting infrastructure which prometheus has already. 

Bartek - what do you think about it?

abuiles:
@bartekn maybe it would be worth taking a step back and figuring what's the problem we want to solve and how each tool fits into this problem. 

@tamirms also has some ideas around this, maybe we can start a document for brainstorming and then we can start talking about how to solve the problem. cc @ire-and-curses

ire-and-curses:
Yeah I agree with this approach. A document with what we're trying to achieve that we can discuss would be ideal. I can see several possible goals and they may need different approaches. For example, some goals might be:

* moving us in the direction of continuous deployment
* reducing our manual testing burden by auto-running some tests
* seeking out regressions
* giving us better alerting at a level more fine than "Horizon is down"

These may need multiple solutions, and some are more important than others.

PR #2356, opened 2020-03-04
POC: health check scheduler

abuiles:
## Proof of concept for https://github.com/stellar/go/issues/2249#issuecomment-590472791

This is a draft showing a proof of concepts for the comment on the link above, this is not mean to be production ready. 

The idea with this is to explore at a very high level what do we want to achieve, cons, pros and try to do a full e2e testing with out k8s cluster.

For now the commands are fixed in the code, but for a final version the server should take a config file listing the commands to run and the frequency. The server will run each command and then report via prometheus metrics whether the command succeeded or not (eventually we can add more data, like how long did it take a command to run, etc). If there is a failure, we'll receive a slack alert and then we can go to kibana and use the logs to identify what was the failure.

The server exposes two routes for now.

1. Index Route `/` -- shows info about each job, like frequency, next and previous run.

```
[
  {
    "Schedule": {
      "Delay": 20000000000
    },
    "Next": "2020-03-03T20:29:22-05:00",
    "Prev": "2020-03-03T20:29:02.001903-05:00",
    "Job": {
      "name": "HorizonCmp"
    }
  },
  {
    "Schedule": {
      "Delay": 30000000000
    },
    "Next": "2020-03-03T20:29:42-05:00",
    "Prev": "2020-03-03T20:29:12.000779-05:00",
    "Job": {
      "name": "HorizonCmpMain"
    }
  }
]
```

2. Metrics route `/metrics` - exposes metrics in prometheus format

```
health_horizon_cmp_testnet_failures 1

health_horizon_cmp_pubnet_failures 0
```

## Next steps
- [ ] Setup k8s deployment
- [ ] Present to ops team and brainstorm
- [ ] Present to other Horizon team member and brainstorm


services/internal/health/main.go
```diff
@@ -0,0 +1,192 @@
...
+}
+
+func (handler MetricsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	handler.app.metrics.Each(func(name string, i interface{}) {
+		// Replace `.` with `_` to follow Prometheus metric name convention.
+		name = strings.ReplaceAll(name, ".", "_")
+
+		switch metric := i.(type) {
+		case metrics.Gauge:
+			fmt.Fprintf(w, "health_%s %d\n", name, metric.Value())
```
jacekn: it would be good to have `horizon_` prefix on all metrics. That way they will be easy to find in prometheus

services/internal/health/main.go
```diff
@@ -0,0 +1,192 @@
...
+	a.cron.AddJob("@every 20s", &horizonCmp)
+	a.metrics.Register("horizon_cmp_testnet.failures", horizonCmp.metric)
+
+	horizonCmpMain := Cmd{
+		Name:   "HorizonCmpMain",
+		metric: metrics.NewGauge(),
+		base:   "https://horizon-blue.stellar.org",
+	}
+	a.cron.AddJob("@every 30s", &horizonCmpMain)
+	a.metrics.Register("horizon_cmp_pubnet.failures", horizonCmpMain.metric)
```
jacekn: Dots are not allowed in prometheus metric names

abuiles: @jacekn thanks :) - we actually replace `.` for `_` on the metrics end-point :)

abuiles:
We won't move forward in this direction for now.

PR #2362, opened 2020-03-06
services/ticker: update *.json files atomically

### LINKED ISSUE #2361: ticker: updates of markets.json are not atomic
We noticed that occasionally GET requests for markets.json return empty result. The problem always corrects itself without intervention. It also always happens 2-3s after past a minute which makes me thing it's related with the following cronjob:

```
* * * * * /opt/stellar/bin/ticker generate market-data --db-url=xxx -o /opt/stellar/www/markets.json
```

I looked at the code and the file is truncated using `os.Create()` and then written again:
https://github.com/stellar/go/blob/master/services/ticker/internal/utils/main.go#L21

If there is a temporary storate slowness we can serve empty or incomplete file for a brief moment. 

It would be good to make the update operation atomic (simple way could be to write markets.json.new and then move to markets.json but there may be a better way in go).

accordeiro:
This PR fixes #2361 – see original issue for further context.

### DIFF
--- a/services/ticker/internal/utils/main.go
+++ b/services/ticker/internal/utils/main.go
@@ -16,9 +16,12 @@ func PanicIfError(e error) {
 	}
 }
 
-// WriteJSONToFile wrtites a json []byte dump to <filename>
+// WriteJSONToFile atomically writes a json []byte dump to <filename>
+// It ensures atomicity by first creating a tmp file (filename.tmp), writing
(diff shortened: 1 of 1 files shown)


accordeiro:
Yeah, I couldn't think of a good way to test this either. I will merge this so we can push the fix to prod, and will put some thought into how to test this afterwards. Thanks for the review! :)

Issue #2339, opened 2020-02-28
exp/services/recoverysigner: add data persistence with postgres


leighmcculloch:
Add data persistence with Postgres. Currently the service only stores information in memory.

PR #2374, opened 2020-03-10
exp/services/recoverysigner: add db migration using postgres

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add a `db migrations [up|down] [count]` command to recoverysigner that migrates the database up or down.

### Why

This is a first step to adding data persistence to recoverysigner and will allow us to make changes to the database schema and apply them incrementally. The structure of how the migrations work and are organized is based off what I see our other applications doing.

This is starting out with an empty schema because it's a well defined isolated chunk of functionality that is only focused on the mechanism for migrations and not the schema itself or the code that uses the database. I'll open subsequent PRs for those other things.

I've included functionality for migrating both up and down because it seems like we have that functionality in most of our applications, except Ticker. I prefer not to ship down migrations because I think in practice they are not useful and sometimes dangerous because they're often not tested well, they can destroy data, and production databases should always move forward in non-destructive ways. Down migrations can be useful in testing and also might be useful to others since this is open source and I assume the reason we've included them with most of our apps is due to a need. I'd like to put a guard in place that makes it a little harder to use the down migration so it isn't run accidentally in production but that isn't included yet. If you have any thoughts please let me know.

For #2339 

### Known limitations

N/A


exp/services/recoverysigner/internal/db/dbmigrate/dbmigrate.go
```diff
@@ -0,0 +1,35 @@
+package dbmigrate
+
+import (
+	"github.com/jmoiron/sqlx"
+	migrate "github.com/rubenv/sql-migrate"
+)
+
+//go:generate go-bindata -nometadata -ignore .+\.(go|swp)$ -pkg dbmigrate -o dbmigrate_generated.go ./migrations
```
marcelosalloum: Is this comment somehow responsible for generating the file `dbmigrate_generated.go`?

exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200309000001-initial-2.sql
```diff
@@ -0,0 +1,7 @@
+-- This migration file is intentionally empty and is a first starting point for
+-- our migrations before we yet have a schema.
+
+-- +migrate Up
+
+-- +migrate Down
+
```
marcelosalloum: Do we need two "first starting point"? I ask this because we have two empty migration files: `20200309000000-initial-1.sql` and `20200309000001-initial-2.sql`.

exp/services/recoverysigner/internal/db/dbmigrate/dbmigrate_test.go
```diff
@@ -0,0 +1,239 @@
...
+
+func TestPlanMigration_upApplyAll(t *testing.T) {
+	db := dbtest.Postgres(t)
+	session, err := dbpkg.Open(db.DSN)
+	require.NoError(t, err)
+
+	migrations, err := PlanMigration(session, migrate.Up, 0)
+	require.NoError(t, err)
+	require.Greater(t, len(migrations), 1)
+	assert.Equal(t, "20200309000000-initial-1.sql", migrations[0])
```
marcelosalloum: I think you could also test for `migrations[1]`, like in lines [R92-R95](https://github.com/stellar/go/pull/2374/files#diff-1e33b298fa035a5984c3a09eac9f0ed8R92-R95):
```suggestion
	assert.Equal(t, "20200309000000-initial-1.sql", migrations[0])
	assert.Equal(t, ""20200309000001-initial-2.sql"", migrations[1])
```

marcelosalloum:
👀

exp/services/recoverysigner/internal/db/dbmigrate/dbmigrate.go
leighmcculloch: Yup! If you run `go generate ./...` all the `go:generate` comments get executed. The go-bindata tool generates that file. If you run it with the `./gogenerate.sh` script it will install go-bindata for you.

exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200309000001-initial-2.sql
leighmcculloch: I couldn't find a way to test that migrations up and down when passing a count work unless there are multiple migrations. Originally I was going to include the schema but that would have expanded the scope of this PR too far. I think migrations are very important to get right and to prevent them breaking and by having the two we can test it more thoroughly so the weirdness of having them is out balanced by that. 

Let me know if you can think of some better ways to do this.

exp/services/recoverysigner/internal/db/dbmigrate/dbmigrate_test.go
leighmcculloch: Good point. I did it slightly differently but the same result, see 1504a71.

exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200309000001-initial-2.sql
marcelosalloum: Good approach! Thanks for explaining

exp/services/recoverysigner/cmd/db.go
```diff
@@ -0,0 +1,109 @@
...
+		},
+		Run: func(cmd *cobra.Command, args []string) {
+			cmd.Help()
+		},
+	}
+	configOpts.Init(cmd)
+
+	migrateCmd := &cobra.Command{
+		Use:   "migrate [up|down] [count]",
+		Short: "Run migrations on the database",
```
howardtw: We might want to document that when `count` is zero, it means "no limit"

leighmcculloch: Good point, that isn't obvious. I think I'll change this so that `0` is an invalid value instead. #2380

PR #2380, opened 2020-03-11
exp/services/recoverysigner: disallow db migrate with a zero count

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Disallow the recoverysigner's `db migrate [up|down] [count]` command with a zero (`0`) count.

### Why

If you pass zero as the count value to the command the command will run all possible migrations up or down. This isn't intuitive even though it's the behavior of the migration library we're using. We don't really need to expose this in our API since the count is optional and this is the behavior if no count is provided.

Thanks for the feedback @howardtw on https://github.com/stellar/go/pull/2374#discussion_r391170652.

### Known limitations

N/A

### DIFF
--- a/exp/services/recoverysigner/cmd/db.go
+++ b/exp/services/recoverysigner/cmd/db.go
@@ -85,6 +85,10 @@ func (c *DBCommand) Migrate(cmd *cobra.Command, args []string) {
 			c.Logger.Errorf("Invalid migration count, must be a number or not provided.")
 			return
 		}
+		if count < 1 {
+			c.Logger.Errorf("Invalid migration count, must be a number greater than zero.")
+			return
+		}
 	}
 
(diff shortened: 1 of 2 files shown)


howardtw:
How does one set up the database without knowing the total number of migrations from scratch?

leighmcculloch:
@howardtw I may not completely understand your question, but the user shouldn't need to know how many migrations there are to setup Postgres.

If you're referring to how does someone determine what number to enter for `count`, see the description:
>We don't really need to expose this in our API since the count is optional

howardtw:
Ah my bad. So `migrate up` with a default value of count being `0` will still execute all migrations. But we disallow a user to manually enter `0`. Is that right?

leighmcculloch:
Exactly.

PR #2381, opened 2020-03-11
exp/services/recoverysigner: make the db migrate invalid count error simpler

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Change the error outputted when an invalid value is provided to the
recoverysigner's `db migrate [up|down] [count]` command as hte count
value from:
```
Invalid migration count, must be a number or not provided.
```
To:
```
Invalid migration count, must be a number.
```

### Why
The error message makes a suggestion to the user about something they
alternatively could do, remove the number, but this might not be what
they need to do and makes assumptions about what a user is doing. For
example, if the user saw this message and removed the invalid value and
were running a down migration it would be destructive. Separate to the
issue that it is too easy to migrate down all the way, this message goes
beyond just telling the user what is wrong and it's unnecessary for the
tool to go further than that.

I noticed this message while I was working on #2380.

### Known limitations

N/A

### DIFF
--- a/exp/services/recoverysigner/cmd/db.go
+++ b/exp/services/recoverysigner/cmd/db.go
@@ -82,7 +82,7 @@ func (c *DBCommand) Migrate(cmd *cobra.Command, args []string) {
 	if len(args) >= 2 {
 		count, err = strconv.Atoi(args[1])
 		if err != nil {
-			c.Logger.Errorf("Invalid migration count, must be a number or not provided.")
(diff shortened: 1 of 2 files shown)


leighmcculloch:
>Do we also want to check for a "positive" number?

@howardtw I'm not following what you mean. #2380 that you already reviewed checks for a positive number.

howardtw:
Oh yeah, they are in different PRs tho. My brain were scoped to this PR only.

PR #2382, opened 2020-03-11
txbuild/data: Add NewManageDataMemoRequired to txbuild.

### LINKED ISSUE #2378: Add `buildMemoRequiredTx` to txnbuild
This will make it easy for consumers to setup `memoRequired` in their accounts

A helper function which creates a valid transaction with the MANAGE_DATA operation that adds a data entry with the name `config.memo_required`

Refs #2329

abuiles:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add helper function to set `config.memo_required` to `1` or `0`.

Fix #2378.

### Why

This will help as a reference and helper for SEP29.

### Known limitations

### DIFF
--- a/txnbuild/data/main.go
+++ b/txnbuild/data/main.go
@@ -1,0 +1,15 @@
+package data
+
+import "github.com/stellar/go/txnbuild"
+
+// NewManageDataMemoRequired returns a valid txnbuild.ManageData operation setting memo required to 1 or 0.
+func NewManageDataMemoRequired(b bool) *txnbuild.ManageData {
+	value := "0"
+	if b {
+		value = "1"
+	}
+	return &txnbuild.ManageData{
+		Name:  "config.memo_required",
+		Value: []byte(value),
+	}
+}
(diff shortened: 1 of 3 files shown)


txnbuild/data/main.go
```diff
@@ -0,0 +1,15 @@
+package data
```
leighmcculloch: This package name might be problematic for importers because it is easily shadowed. Should we call this `txndata` or `txnbuilddata`, or is that too long do you think?

txnbuild/example_test.go
```diff
@@ -199,6 +200,29 @@ func ExampleManageData() {
...
+	client := horizonclient.DefaultTestNetClient
+	ar := horizonclient.AccountRequest{AccountID: kp.Address()}
+	sourceAccount, err := client.AccountDetail(ar)
+	check(err)
+
+	op := data.NewManageDataMemoRequired(true)
+
+	tx := Transaction{
+		SourceAccount: &sourceAccount,
+		Operations:    []Operation{&op},
```
leighmcculloch: Do you need the `&` here? I asked because the return value of `NewManageDataMemoRequired` is an `*txnbuild.ManageData` already, so you're getting a pointer to a pointer.
```suggestion
		Operations:    []Operation{op},
```

PR #2383, opened 2020-03-12
exp/services/recoverysigner: update API to reflect SEP-30 v0.2.0

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Update API and related code to reflect the changes proposed in SEP-30 v0.2.0 in stellar/stellar-protocol#566.

### Why

To align the reference implementation with the SEP change being made to allow for easier expansion in the future of both additional roles and multiple values for each identity type (email, phone, etc).

I'm updating this code before finalizing the protocol change because I wanted to validate the protocol change in code. I'll merge both together.


exp/services/recoverysigner/internal/serve/account_post.go
```diff
@@ -57,18 +56,22 @@ func (h accountPostHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		Address:    req.Address.Address(),
+		Identities: []account.Identity{},
+	}
+	for _, i := range req.Identities {
+		accIdentity := account.Identity{
+			Role: i.Role,
+		}
+		for _, m := range i.AuthMethods {
+			accIdentity.AuthMethods = append(accIdentity.AuthMethods, account.AuthMethod{
+				Type:  account.AuthMethodType(m.Type), // TODO: Validate AuthMethodType
```
accordeiro: Do we need to be opinionated about the AuthMethod types? Or are we always simply going to compare the stored `method` with the one provided in the JWT?

accordeiro: After reading the rest of the code, it seems like we are explicit all over the codebase about the methods we support. In that case, shouldn't we add this validation now?

leighmcculloch: >are we always simply going to compare the stored method with the one provided in the JWT?

This is an interesting idea, but I could see it being easy to make a mistake leading to an exploit. I'd rather be explicit. Any implementer can add new auth methods and support them too.

We can add validation now. I'll add it before merging.

exp/services/recoverysigner/internal/account/account.go
```diff
@@ -1,19 +1,39 @@
...
+
+type Identity struct {
+	Role        string
+	AuthMethods []AuthMethod
+}
+
+type AuthMethodType string
+
+const (
+	AuthMethodTypeAddress     AuthMethodType = "stellar_address"
```
howardtw: I find this name a little confusing while I'm reviewing the code as I'm not sure whether it's a stellar address or an email address.

leighmcculloch: Inside the application wherever `address` is referenced it means stellar address. I think everything will get a bit wordy if we add the word `stellar` prefixed too. If you look a couple lines down email is right there as another option, so I doubt too many people will make this mistake?

leighmcculloch:
>I think it's time to add a SPEC for the service.

I don't plan to write a `SPEC.md` file for this service. SEP-30 is the specification. If anything isn't clear it should be added there, not here. This is just a reference implementation of SEP-30. Can you open issues or PRs for anything in SEP-30 that isn't clear? Anything you already mentioned on PRs I've got noted down to add, so no need to open anything that's already been mentioned.

PR #2386, opened 2020-03-12
exp/services/recoverysigner: add migrations for accounts, identities, auth_methods

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add migrations creating the tables to store accounts, identities and authentication methods (auth_methods).

```
+--------+ 1  * +----------+ 1  * +------------+   
|accounts|------|identities|------|auth_methods|   
+--------+      +----------+      +------------+  
```

The migrations align with the changes being made in stellar/go#2383 and stellar/stellar-protocol#566.

### Why

The recoverysigner service currently stores its data in memory. We need it to store its data persisted in Postgres. These migrations add the schema that the data will be stored in.

For #2339 

### Known limitations

This change doesn't include the use of the tables yet and I'm opening it early to give folks a chance to provide feedback about how the data is structured before I write a significant amount of code around that.


exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200311000000-create-accounts.sql
```diff
@@ -0,0 +1,15 @@
+-- +migrate Up
+
+CREATE TABLE accounts (
+  id TEXT NOT NULL PRIMARY KEY,
+  created_at TIMESTAMP NOT NULL,
```
howardtw: Do you think it would be better if we use TIMESTAMP WITH TIMEZONE?

exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200311000001-create-identities.sql
```diff
@@ -0,0 +1,16 @@
+-- +migrate Up
+
+CREATE TABLE identities (
+  id TEXT NOT NULL PRIMARY KEY,
+  created_at TIMESTAMP NOT NULL,
+  deleted_at TIMESTAMP,
+
+  account_id TEXT NOT NULL REFERENCES accounts(id) ON DELETE CASCADE,
```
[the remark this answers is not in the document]
howardtw: I'm not sure whether we need to hard-delete a row from this table while preserving the row in accounts. Using the `ON DELETE CASCADE` poses a certain limit on ourselves. Maybe it's not worth having it given that we use soft-deletion.

exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200311000001-create-identities.sql
```diff
@@ -0,0 +1,16 @@
+-- +migrate Up
+
+CREATE TABLE identities (
+  id TEXT NOT NULL PRIMARY KEY,
```
howardtw: What does the `id` of an identity stand for? Is it the id derived from the jwt token? And is `account_id` essentially a stellar address?

exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200311000000-create-accounts.sql
leighmcculloch: What value do you see using the timezone aware timestamp?

exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200311000001-create-identities.sql
leighmcculloch: It's just the unique identifier for the row. Since an account and identities can be deleted and re-added it is helpful to have a unique identifier for entities. In logs we can reference that ID instead of the address or other fields of the identity and we'll know exactly which instance of an account or an identity we're referring to.

The `account_id` is the foreign key to the accounts table. The names are unfortunate, but the schema uses the references keyword to make that relationship clearer. From what I can see of our other applications we use the term `id` for the row primary key.

exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200311000000-create-accounts.sql
howardtw: TIMESTAMP WITH TIMEZONE is safer given the reason that this service will be used by different parties and possibly in different timezone. If one client were to access recovery signers hosted in different timezone, the time can be confusing or impossible to interpret without any timezone information. Furthermore, whenever a party decides to change a timezone of its server, `TIMESTAMP` cannot be changed automatically since it does not have any information regarding the timezone. In short, I think starting with `TIMESTAMP WITH TIMEZONE` will make our lives easier.

leighmcculloch: Timestamp with timezone stores the timestamp in UTC on the database, and is subject to conversion back and forth with the current tzdata on the machine. If an application is always storing UTC this is unnecessary.

Id argue that all servers should be using UTC. There might be edge cases where people don't want to use UTC, but they're exceptions, and to not use UTC from the outset can cause much larger problems long term. You're right we don't have control over how somebody configures their database server, but we can write the code to always accept, store and read time in UTC and to not rely on the database's knowledge about time or tzdata by using timezone without timezone.

howardtw: I don't want to set any expectation on which timezone the servers should be using. As you mentioned, `timestamp with timezone` stores the timestamp in UTC, which is what we want.

> we can write the code to always accept, store and read time in UTC and to not rely on the database's knowledge about time or tzdata by using timezone without timezone.

I feel like this is something we should rely on the database's knowledge without having to write our own code. I personally don't see any disadvantage of using the timestamp with timezone type, and having the timezone information that works anywhere is more flexible.

leighmcculloch: I don't think it will matter how the servers are configured, it matters how we write the application. If our application always stores UTC times, then the times stored in the DB are always UTC. This is the simplest approach to working with times across systems.

Relying on the DB to do timezone conversion is adding another moving part to the system. We'd be relying on that system having consistent tzdata with the application server. More moving parts = more potential problems.

exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200311000001-create-identities.sql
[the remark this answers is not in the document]
leighmcculloch: In any system there will be the need to hard delete to meet some privacy requirements, if for example somebody asks to have all their information removed from a service, but for everyday operation we don't need to delete.

What limit do you see this posing?

exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200311000000-create-accounts.sql
leighmcculloch: TIL [PostgreSQL Wiki recommends not using `without` to store UTC times](https://wiki.postgresql.org/wiki/Don't_Do_This#Don.27t_use_timestamp_.28without_time_zone.29_to_store_UTC_times), but for different reasons. The reasons it recommends using `with` when storing only UTC is so that you can write queries using relative time ranges when the timezone isn't configured to match. I admit I've never used relative time ranges outside of a psql console, so this has never affected backend code for me, but I can see that being useful.

If we use `with` I still think the application should only store UTC time, so we are not using the database for time conversion.

leighmcculloch: 792a774807972b5160c9080c6e199c3ca87a2315

exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200311000000-create-accounts.sql
```diff
@@ -0,0 +1,15 @@
+-- +migrate Up
+
+CREATE TABLE accounts (
+  id UUID NOT NULL PRIMARY KEY,
+  created_at TIMESTAMP WITH TIME ZONE NOT NULL,
```
howardtw:
```suggestion
  created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
```

Since the column is for `created_at`, I usually have the db do `Now()`, which returns a timestamp with timezone, and I haven't had any problem relying on db. If we were to use `timestamp`, we would essentially cut the timezone off when we store it (if we want to use the database function and if the conversion works.)

exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200311000002-create-auth-methods.sql
```diff
@@ -0,0 +1,28 @@
+-- +migrate Up
+
+CREATE TYPE auth_method_type AS ENUM ('stellar_address', 'phone_number', 'email');
```
marcelosalloum: 👍

exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200311000000-create-accounts.sql
marcelosalloum: I also like using `DEFAULT NOW()` for the `created_at` field. If you agree with that @leighmcculloch, please remember to add it to the other tables as well:
* [identities table](https://github.com/stellar/go/pull/2386/files#diff-ac186b3068915e04924859e4ff6372b7R10).
* [auth_methods table](https://github.com/stellar/go/pull/2386/files#diff-2e653a05c60233a279a2972786c41c1fR14).

leighmcculloch: I've used setting a default value before too, why do you guys like using it?

It definitely works, but it makes writing tests difficult.
- Instead of writing simple tests that check that a set of rows exist in the database, we have to write tests that check each value independently and do things like check that the value is within a certain duration. In some languages and testing frameworks (e.g. Ruby with Rspec) this is really easy, in Go not so much.
- Mocking time also becomes more difficult.

In the application code it's really easy to set the `created_at` with a value which is what I'm doing in the code at the moment. Do you see an issue with that?

marcelosalloum: Hum, I was not thinking from that perspective... I agree with not including ~`DEFAULT NOW()`~.

exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200311000001-create-identities.sql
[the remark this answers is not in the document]
leighmcculloch: I ended up dropping the `ON DELETE CASCADE` and not specifying the on delete behavior. The default behavior is equivalent to `ON DELETE RESTRICT` which I think is safer as it will prevent us from accidentally deleting a record that is referenced by another table.

exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200311000000-create-accounts.sql
howardtw: > we have to write tests that check each value independently and do things like check that the value is within a certain duration. In some languages and testing frameworks (e.g. Ruby with Rspec) this is really easy, in Go not so much.

How is it more difficult with Go than with Rspec? Can you elaborate? Note that we can also write a query to check whether the timestamp is within a certain range if you don't like the Go `time` package. In addition, I don't understand how it is going to make a difference in testing whether we generate a timestamp in the function or in the database. After all, we need to test if the function stores the data correctly, and the client shouldn't be concerned about whether the timestamp is generated before or during the db call, right?

I don't feel strongly about this, but atm I still prefer to have `DEFAULT NOW()` rather than writing our own logic in the application level.

accordeiro: One interesting issue with setting this in the application layer is that, if we have 2+ server instances behind a load balancer, there's a chance that they will have slightly skewed clocks, which could lead to a false history of the order in which rows were created (or even you could see an entry with a higher incremental `pk` but lower `created_at` than another). 

Allowing the database to be the source of time truth in this case mitigates the issue, but I'm not sure if this is a big issue.

leighmcculloch: I don't think that's a big issue. You can get that inconsistency between PKs and timestamps anyway because `NOW()` is the timestamp of the start of the database transaction, not the time of the insert statement, and the PK is incremented when the insert statement is executed inside the transaction not at the start of the transaction. Once an app gets to scale of multiple write db servers there's also chance with inconsistency.

This app logic does other important things with time (e.g. validate JWTs) so I'd hope we weren't allowing too great a time skew.

leighmcculloch:
Thanks all for the great discussion on the schema! This PR is now superseded by #2403 where you can see the schema in use.

Issue #1941, opened 2019-11-15
Add state reingestion integration tests


bartekn:
Add state reingestion integration (with DB) tests to ensure the process works correctly (including truncating tables). See: https://github.com/stellar/go/pull/1940.

PR #2387, opened 2020-03-13
services/horizon/internal/expingest: Add ingestion integration tests which exercise horizon db

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Close https://github.com/stellar/go/issues/1941

Add integration tests for horizon ingestion which run against postgres instead of db mocks. The test cases cover two scenarios:

* Ingesting state from the history archives into a clean db
* Ingesting state from the history archives into a db which is already populated with data from a previous ingestion run.

### Why

When implementing experimental ingestion we ran into an error during a deployment. The error occurred because we added a new ingestion table but we forgot to truncate it when rebuilding state (see https://github.com/stellar/go/pull/1940 for more details).

This class of error is difficult to catch with unit tests so that is why we created these integration tests which run against a postgres database.

### Known limitations

The integration tests ingest from a fake history archive snapshot. The fake snapshot is created by sampling `xdr.LedgerEntryChanges` from a real history archive snapshot. The reason why we need to do this is because running ingestion on a full snapshot takes at least 15 minutes. If we were to have such a long running test it would likely be very flakey.


ire-and-curses:
Some extra background: We talked about this a bit before settling on this implementation. Other approaches included running the full test (slow and potentially flakey as mentioned) or running the test only on branch merge. We liked this solution because it's fast and can be run all the time. If a new type were ever to be added to the ledger XDR then to be representative this snapshot should be updated. But we expect this to be a rare occurance.

PR #2394, opened 2020-03-17
services/horizon: Validate transaction hash IDs as 64 lowecase hex chars

### LINKED ISSUE #2392: Transaction details endpoint should validate hash parameter
The [endpoint](https://www.stellar.org/developers/horizon/reference/endpoints/transactions-single.html) which processes `/transactions/{hash}` requests does not validate the hash parameter. 

If you use a hash which is not a valid utf8 sequence this will result in horizon responding with a 500 status. Here is an example of this error which was logged recently:

```
time="2020-03-12T17:30:39.411Z" level=error msg="loading transaction record: get failed: pq: invalid byte sequence for encoding \"UTF8\": 0x00" pid=5777 req=horizon-testnet-green-002c/xuy67jVTC6-44865331 stack="[main.go:43 transaction.go:118 action.go:217 value.go:460 value.go:321 handler.go:100 handler.go:69 handler.go:222 server.go:2007 mux.go:418 server.go:2007 mux.go:69 mux.go:295 server.go:2007 mux.go:418 server.go:2007 mux.go:69 mux.go:295 server.go:2007 mux.go:418 server.go:2007 http.go:73 server.go:2007 cors.go:190 server.go:2007 compress.go:66 server.go:2007 middleware.go:220 server.go:2007 middleware.go:231 timer.go:227 middleware.go:230]"
```

2opremio:
Fixes #2392 

<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Validate transaction ID URL parameters as 64 lowercase hexadecimal characters

### Why

Injecting 0x00 was causing internal errors. Also, it's more intuitive to receive a bad request error than a not-found error when there is a problem in the format.

### Known limitations

Hardcoding the length may not be a good idea if we change the hash format/length.

This requires a minor release, since we change the HTTP error returned for badly-formatted hashes.


services/horizon/internal/actions_transaction_test.go
```diff
@@ -31,7 +31,7 @@ func TestTransactionActions_Show(t *testing.T) {
 
 	// missing tx
 	w = ht.Get("/transactions/not_real")
-	ht.Assert.Equal(404, w.Code)
+	ht.Assert.Equal(400, w.Code)
```
leighmcculloch: The `missing tx` comment suggests the intent behind these expectations was to check that an unknown transaction returned a 404. I think we should probably change the input `not_real` to a valid tx hash that is for a tx that doesn't exist on the ledger so that we're still testing that case.

We probably need a new test case to test that passing in an invalid value results in 400.

services/horizon/internal/handler.go
```diff
@@ -252,6 +252,11 @@ func getAccountID(r *http.Request, key string, required bool) (string, error) {
 // getShowActionQueryParams gets the available query params for all non-indexable endpoints.
 func getShowActionQueryParams(r *http.Request, requireAccountID bool) (*showActionQueryParams, error) {
 	txHash, err := hchi.GetStringFromURL(r, "tx_id")
+	if err == nil {
+		if _, err = hex.DecodeString(txHash); err != nil || len(txHash) != 64 || strings.ToLower(txHash) != txHash {
```
leighmcculloch: It wasn't obvious to me why we'd also want to check that the case of the hash was lower case, and that we do that because Horizon looks up the transaction hashes lookup is case sensitive favoring lower case. Could we add a test for that to knowledge share why that is and to test that logic? I don't see a test for a valid transaction hash that is upper case.

services/horizon/internal/actions_transaction_test.go
```diff
@@ -31,7 +31,7 @@ func TestTransactionActions_Show(t *testing.T) {
...
 }
```
leighmcculloch: Could we also add a test for the reported issue that is being fixed?

tamirms:
@2opremio the `tx_id` parameter is used in the following endpoints:

`GET /transactions/{tx_id}`
`GET /transactions/{tx_id}/effects`
`GET /transactions/{tx_id}/operations`
`GET /transactions/{tx_id}/payments`

https://github.com/stellar/go/blob/master/services/horizon/internal/web.go#L234-L238

The fix you have implemented only covers the `GET /transactions/{tx_id}` request.

There are two patterns for implementing HTTP handlers in Horizon. The legacy pattern resembles: https://github.com/stellar/go/blob/master/services/horizon/internal/actions_effects.go
The new pattern, which we eventually want to adopt for all endpoints, is to define all http handlers in the actions package: https://github.com/stellar/go/blob/master/services/horizon/internal/actions

Unfortunately, all the endpoints which have a `tx_id` parameter are implemented in the legacy style. Eventually, we should refactor those endpoints and move them to the actions package. But this is outside the scope of your PR.

Our strategy for migrating endpoints to the actions package has been to implement functions which parse and validate request parameters in https://github.com/stellar/go/blob/master/services/horizon/internal/actions/helpers.go . Those functions can be used by the legacy handlers which are defined outside the actions package. But, eventually we will move the handlers to the actions package so having the validators live in https://github.com/stellar/go/blob/master/services/horizon/internal/actions/helpers.go makes the migration easier.

My suggestion is to implement a function to extract the transaction hash from a request in helpers.go . Here is a similar function that you can use as a model https://github.com/stellar/go/blob/master/services/horizon/internal/actions/helpers.go#L379-L396

Then you can call the function in each of the handlers for the endpoints I listed above.

2opremio:
@tamirms Thanks a lot for the comprehensive explanation. I think everything is addressed. PTAL

services/horizon/internal/actions_transaction_test.go
```diff
@@ -30,8 +30,16 @@ func TestTransactionActions_Show(t *testing.T) {
 	}
 
 	// missing tx
-	w = ht.Get("/transactions/not_real")
```
tamirms: could you add similar test cases for the remaining endpoints?

```
GET /transactions/{tx_id}/effects
GET /transactions/{tx_id}/operations
GET /transactions/{tx_id}/payments
```
https://github.com/stellar/go/blob/ddf5898ed9e3dd37594c5dd6527a132b0b635189/services/horizon/internal/actions_effects_test.go
https://github.com/stellar/go/blob/ddf5898ed9e3dd37594c5dd6527a132b0b635189/services/horizon/internal/actions_operation_test.go
https://github.com/stellar/go/blob/ddf5898ed9e3dd37594c5dd6527a132b0b635189/services/horizon/internal/actions_payment_test.go

2opremio: Sure thing!

services/horizon/internal/actions/helpers.go
```diff
@@ -376,6 +378,35 @@ func (base *Base) GetAddress(name string, opts ...Opt) (result string) {
...
+			return "", problem.MakeInvalidFieldProblem(name, errors.New("invalid hash format"))
+		}
+	}
+
+	return value, nil
+}
+
+// GetTransactionID retireves  a transaction identifier by attempting to decode an hex-encoded,
+// 64-digit lowercase string at the provided name.
+func (base *Base) GetTransactionID(name string) string {
```
abuiles: btw, since we are trying to get rid of the `base` helpers -- let's not include this one and use the one defined above instead.

PR #2396, opened 2020-03-17
clients/horizonclient: Add SEP0029 memo required check.

### LINKED ISSUE #2379: Implement SEP29 check in `horizonclient.SubmitTransaction`

Implements SEP29 check in horizonclient

Modify transaction submission to check for payment operations, and if present, call Horizon to check the account for the presence of config.memo_required.

When a payment sender submits a transaction containing a PAYMENT, PATH_PAYMENT_STRICT_SEND, PATH_PAYMENT_STRICT_RECEIVE, or MERGE_ACCOUNT operation they:

1. Load details about the destination account.
2. Check if the destination account has the config.memo_required data entry set. If so: Verify that the transaction memo field is set.
3. If memo is not set, show a notification and prevent the transaction from submission.

For transactions containing more than one operation, perform the same checks for each operation

This check will be done by default, however users can skip the check by passing it as an argument to `horizonclient.SubmitTransaction`

Refs #2329

abuiles:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add [SEP0029](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0029.md) memo required check in `client.SubmitTransaction`.

Fixes #2379

### Why

We want to include out of the box support for SEP0029.

### Known limitations

[TODO or N/A]


clients/horizonclient/client.go
```diff
@@ -38,6 +38,62 @@ func (c *Client) sendRequest(hr HorizonRequest, resp interface{}) (err error) {
...
+			continue
+		}
+		destinations[destination] = true
+
+		request := AccountRequest{
+			AccountID: destination,
+			DataKey:   "config.memo_required",
+		}
+
+		data, err := c.AccountData(request)
```
abuiles: We decided to do this sequentially for now without any caching. Once we release it and have more data from the community, we can decide what to do to make it faster.  Like memoizing the responses or adding batch loading in horizon for the `/accounts` end-point.

clients/horizonclient/client.go
```diff
@@ -38,6 +38,62 @@ func (c *Client) sendRequest(hr HorizonRequest, resp interface{}) (err error) {
...
+		if err != nil {
+			horizonError, ok := err.(*Error)
+
+			if !ok || horizonError.Response.StatusCode != 404 {
+				return err
+			}
+
+			continue
+		}
```
leighmcculloch: Does the fact this ignores the 404 response code mean that if an account doesn't exist yet because it is being created inside this transaction that we'll still allow the transaction to be submitted?

clients/horizonclient/client.go
```diff
@@ -38,6 +38,62 @@ func (c *Client) sendRequest(hr HorizonRequest, resp interface{}) (err error) {
...
+
+		if data.Value == "MQ==" {
```
leighmcculloch: Could we add a comment here indicating that `MQ==` is `"1"` base64 encoded? I have context for what this means and why we're checking for this value so when I look at this it makes sense, but I think somebody stumbling onto this without context is unlikely to.

clients/horizonclient/main.go
```diff
@@ -139,6 +139,14 @@ type Client struct {
 	clock *clock.Clock
 }
 
+// SubmitTxOpt optional parameters for SubmitTransaction
+type SubmitTxOpt int
+
+const (
+	// SkipMemoRequiredCheck skips SEP0029 memo required check
+	SkipMemoRequiredCheck SubmitTxOpt = 99
```
leighmcculloch: What significance does the value `99` have. We could use `iota` if this is going to be a list of integer constants and the value is not intended for use outside the SDK.

clients/horizonclient/main.go
```diff
@@ -139,6 +139,14 @@ type Client struct {
 	clock *clock.Clock
 }
 
+// SubmitTxOpt optional parameters for SubmitTransaction
+type SubmitTxOpt int
```
leighmcculloch: We only get one shot at adding options to `SubmitTransaction` in a non-breaking way, but making this type an integer is limiting the types of options we can add to being boolean on/off features. Is keeping this flexible important to us? 

If keeping it more flexible is valuable to us we can make the option type a function instead of an integer.

```go
type SubmitTxOpts struct {
	SkipMemoRequiredCheck boolean
}

type SubmitTxOpt func(*SubmitTxOpts)

func SubmitTxOptSkipMemoRequiredCheck(opts *SubmitTxOpts) {
	opts.SkipMemoRequiredCheck = true
}
```

Example use:

```go
func (c *Client) SubmitTransaction(transaction txnbuild.Transaction, opts ...SubmitTxOpt) (txSuccess hProtocol.TransactionSuccess, err error) {
	options := SubmitTxOpts{}
	for _, opt := range opts {
		opt(&options)
	}
...
```

This is a little overkill for adding a single option, but it'll be a breaking change if we want to go beyond boolean options in the future.

clients/horizonclient/CHANGELOG.md
```diff
@@ -5,6 +5,24 @@ file.  This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+### Add 
+
+- Add [SEP0029](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0029.md) memo required check in `client.SubmitTransaction`.
+    If any of the operations included in `client.SubmitTransaction` is of type
+    `payment`, `pathPaymentStrictReceive`, `pathPaymentStrictSend`, or
+    `mergeAccount`, then the SDK will load the destination account and check if
```
leighmcculloch: >load the destination account

If I was reading this without context I wouldn't expect this to mean that this is a call out to Horizon, but I think it's important we call that out. Could we clarify that, maybe with the change below?
```suggestion
    `mergeAccount`, then the SDK will load the destination account from Horizon and check if
```

clients/horizonclient/client.go
```diff
@@ -38,6 +38,62 @@ func (c *Client) sendRequest(hr HorizonRequest, resp interface{}) (err error) {
...
+			return errors.New(fmt.Sprintf("MemoRequired:Operation[%d](%s) - Destination: %s requires a memo in the transaction", i, operationFormatted, destination))
```
leighmcculloch: If an application would like to do something specific when a memo isn't set but one is required, how do they detect this specific error and fork their logic on it?

clients/horizonclient/main.go
tamirms: I agree. We should be using functional options here. This idiom is pretty common in the Go community, see https://commandcenter.blogspot.com/2014/01/self-referential-functions-and-design.html and https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

clients/horizonclient/client.go
tamirms: or better yet, store "MQ==" in a `const` variable and then reference it in the if statement:

```golang
// AccountRequiresMemo is the base64 encoding of "1".
// SEP 29 uses this value is to define transaction memo requirements for incoming payments.
const AccountRequiresMemo = "MQ=="
...

if data.Value == AccountRequiresMemo {
...
}
```

clients/horizonclient/client.go
abuiles: @leighmcculloch If the account doesn't exist but it is being created inside the current transaction, I'd say it is okay to let this transaction be submitted.  We can have a 404 for two reasons, we are loading the data from 

`/accounts/{account_id}/data/config.memo_required`

It will 404 if the data key is not set or if the account doesn't exist. 

I'm trying to think more about the scenario  you described, but it seems quite complex/unlikely to happen, probably beyond the scope of SEP0029.

Let's say you are performing the following operations inside the same transaction:

```
1. creating an account
2. setting the memo required value
3. sending a payment or merging another account into the account you created n 1.
```

This will add more complexity into the SEP29 check, now, not only do we have to look for payments or merge accounts, we'll have to look for create account operations, check if any of the payments is going to that account and then check if there is a memo required operation added. And then you could keep going, what if memo required is added after the payment op? 

Maybe the SEP should be explicit about this? I don't think it makes sense to create an account, set memo required ,and then send a payment to it inside the same tx.

clients/horizonclient/CHANGELOG.md
```diff
@@ -5,6 +5,24 @@ file.  This project adheres to [Semantic Versioning](http://semver.org/).
...
+    `config.memo_required` is set to `1` as defined in SEP0029.
+
+    This behavior is the default. You can pass skip the check providing the
```
ire-and-curses:
```suggestion
    This behavior is the default. For performance reasons, you may choose to skip the check by providing the
```

clients/horizonclient/CHANGELOG.md
```diff
@@ -5,6 +5,24 @@ file.  This project adheres to [Semantic Versioning](http://semver.org/).
...
+    optional parameter  `horizonclient.SkipMemoRequiredCheck` to
+    `client.SubmitTransaction`:
+
+	```
+		client.SubmitTransaction(tx, horizonclient.SkipMemoRequiredCheck)
+	```
+
+	The check is also skipped if the transaction includes a memo.
```
ire-and-curses:
```suggestion
	Additionally, the check will be skipped automatically if the transaction includes a memo.
```

clients/horizonclient/client.go
```diff
@@ -38,6 +38,62 @@ func (c *Client) sendRequest(hr HorizonRequest, resp interface{}) (err error) {
 	return c.sendRequestURL(c.HorizonURL+endpoint, "get", resp)
 }
 
+func (c *Client) checkMemoRequired(transaction txnbuild.Transaction) error {
```
ire-and-curses: Docstring would be great here

clients/horizonclient/client.go
```diff
@@ -38,6 +38,62 @@ func (c *Client) sendRequest(hr HorizonRequest, resp interface{}) (err error) {
...
+		destinations[destination] = true
+
+		request := AccountRequest{
+			AccountID: destination,
+			DataKey:   "config.memo_required",
+		}
+
+		data, err := c.AccountData(request)
+		if err != nil {
+			horizonError, ok := err.(*Error)
```
ire-and-curses: I believe we added a helper method recently that can abstract this casting a bit, might be worth considering here.

clients/horizonclient/main.go
abuiles: My reasoning for assigning 99 was that we might remove this once muxed accounts are rolled out, so I wanted to reserved iota .... , for long term options. However, this might be irrelevant based on your comment here https://github.com/stellar/go/pull/2396#discussion_r394047347 

I'll update to functional options :)

clients/horizonclient/client.go
ire-and-curses: yeah I also agree it seems unlikely that this would happen. If one account per exchange with the field set, then that call is rare. And why would you send a payment in the same action?

clients/horizonclient/client.go
```diff
@@ -38,6 +38,62 @@ func (c *Client) sendRequest(hr HorizonRequest, resp interface{}) (err error) {
...
+
+			if !ok || horizonError.Response.StatusCode != 404 {
+				return err
+			}
+
+			continue
+		}
+
+		if data.Value == AccountRequiresMemo {
+			return errors.New(fmt.Sprintf("MemoRequired:Operation[%d](%s) - Destination: %s requires a memo in the transaction", i, operationFormatted, destination))
```
ire-and-curses: lowercase error string

clients/horizonclient/client.go
```diff
@@ -413,8 +469,21 @@ func (c *Client) SubmitTransactionXDR(transactionXdr string) (txSuccess hProtoco
...
-	err error) {
+func (c *Client) SubmitTransaction(transaction txnbuild.Transaction, opts ...SubmitTxOpt) (txSuccess hProtocol.TransactionSuccess, err error) {
+	skipMemoRequired := false
+	for _, opt := range opts {
+		if opt == SkipMemoRequiredCheck {
+			skipMemoRequired = true
+		}
+	}
+
+	if !skipMemoRequired && transaction.Memo == nil {
```
ire-and-curses: why are we checking for a nil memo here? Shouldn't we always do the `checkMemoRequired` if `skipMemoRequired` is false? Can you add an explanatory comment?

clients/horizonclient/main.go
```diff
@@ -61,6 +61,9 @@ const (
 	AssetType12 AssetType = "credit_alphanum12"
 	// AssetTypeNative represents the asset type for Stellar Lumens (XLM)
 	AssetTypeNative AssetType = "native"
+	// AccountRequiresMemo is the base64 encoding of "1".
+	// SEP 29 uses this value is to define transaction memo requirements for incoming payments.
```
ire-and-curses:
```suggestion
	// SEP 29 uses this value to define transaction memo requirements for incoming payments.
```

clients/horizonclient/main.go
ire-and-curses: downside of this is that we lose type information on the option, is that a concern? I don't have a strong opinion either way

clients/horizonclient/client.go
leighmcculloch: I think it would be very valid to send a payment in the same operation, but I agree we don't need to block in this case, I was just wanting to understand all the reasons we would handle the 404 in this way. I don't think anything more is needed here, resolving this comment thread.

clients/horizonclient/main.go
leighmcculloch: Yeah I'd say it is a concern, it's a tradeoff. I don't think we should use this pattern when we don't need to, they're complex, but they are a reasonably common pattern and they fit our requirements of adding options in a non-breaking way.

Alternative: Add a new function that accepts the `SubmitTxOpts` struct. `SubmitTransactionWithOptions`. It's much simpler.

I could go either way.

clients/horizonclient/client.go
abuiles: @ire-and-curses  

> Shouldn't we always do the checkMemoRequired if skipMemoRequired is false?

If there is a memo in the transaction, it doesn't matter if any of the destination accounts require a memo or not, since it will be there anyways.

Now the question is -- shall we add extra validation for the value contained in the memo - checking that at least it is present?

ire-and-curses: oh I see. makes sense :+1:

clients/horizonclient/client.go
abuiles: @ire-and-curses do you remember which is the helper?

ire-and-curses: #2330

clients/horizonclient/main.go
abuiles: @leighmcculloch I was leaning about your second proposal, but after thinking about it, I think the initial idea of functional options makes  sense and it will let us roll this out and start applying the check in an automatic opt-in kind of way. 

If we were to add a new function then we wouldn't have the auto opt-in behavior.

leighmcculloch: @abuiles Add the new function, and have the old function call it with an empty SubmitTxOpts. You should have both functions operate the exact same way, where SubmitTransaction just operates with the default options. The default being that skip memo required check is false/off.

clients/horizonclient/client.go
```diff
@@ -411,10 +466,32 @@ func (c *Client) SubmitTransactionXDR(transactionXdr string) (txSuccess hProtoco
...
+//
+// This function will always check if the destination account requires a memo in
+// the transaction (as defined in https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0029.md)
+// if you want to skip this check, use `SubmitTransactionWithOptions{tx, SubmitTxOpts{SkipMemoRequiredCheck: true})`
+//
 // See https://www.stellar.org/developers/horizon/reference/endpoints/transactions-create.html
-func (c *Client) SubmitTransaction(transaction txnbuild.Transaction) (txSuccess hProtocol.TransactionSuccess,
-	err error) {
+func (c *Client) SubmitTransaction(transaction txnbuild.Transaction) (txSuccess hProtocol.TransactionSuccess, err error) {
+	return c.SubmitTransactionWithOptions(transaction, SubmitTxOpts{SkipMemoRequiredCheck: false})
```
leighmcculloch: Specifying the default value of fields in SubmitTxOpts is unnecessary, and if we add more options I imagine we won't keep coming back to this line to add them. Probably unnecessary now?
```suggestion
	return c.SubmitTransactionWithOptions(transaction, SubmitTxOpts{})
```

clients/horizonclient/main.go
```diff
@@ -61,6 +61,9 @@ const (
 	AssetType12 AssetType = "credit_alphanum12"
 	// AssetTypeNative represents the asset type for Stellar Lumens (XLM)
 	AssetTypeNative AssetType = "native"
+	// AccountRequiresMemo is the base64 encoding of "1".
+	// SEP 29 uses this value to define transaction memo requirements for incoming payments.
+	AccountRequiresMemo = "MQ=="
```
leighmcculloch: Because this is exported it will be surfaced in the SDK's API and documentation. The constant is only intended for use internally though, so this adds noise to the API/docs. Could we leave it not exported, or are we really wanting this to be exported and used by others?

clients/horizonclient/client.go
abuiles: cool! Used here https://github.com/abuiles/go/commit/642689190df55c870dcd2e6b780a22729d31b3e4

clients/horizonclient/client.go
abuiles: You are right, I did it on purpose to be explicit about it. I was thinking more from the perspective of someone who is new to go and is not aware defaults. But you are right, we can update it.

clients/horizonclient/main.go
abuiles: fixed.

Issue #2388, opened 2020-03-13
exp/services/recoverysigner: validate request parameters on POST/PUT endpoints


leighmcculloch:
The recoverysigner has a couple endpoints that accept fields for storing alongside a registered account but does very little validation on the contents of those fields.

Validation should be implemented for the following endpoints:

- `POST /accounts/<address-id>`
- `PUT /accounts/<address-id>` (this endpoint isn't implemented yet, but will be in #2340)

Some ideas for what type of validation should be added:

- Fields that are not listed as optional should always be provided in the request with a non-empty value. e.g. `identities[].role`, `identities[].auth_methods[].value`.
- Fields that contain specific types of data should have their format validated. e.g. `identities[].auth_methods[].value` should have its contents validated as either a valid address, phone number, or email based on the adjacent `type` field.
- Arrays that we'd normally expect to contain at least one item should validate for that. e.g. `identities` and `identities[].auth_methods`.
- Limit the number of identities that can be added to something reasonable.
- Limit the number of auth methods that can be added to something reasonable.

leighmcculloch:
#2390 adds validations of identities, role, auth methods and auth method type. Still need to add validation of auth method values to ensure the stellar address, phone number, or email address provided in the field are valid for each.

mollykarcher:
Closing all recoverysigner issues

Issue #2358, opened 2020-03-04
exp/services/recoverysigner: improve logs and alerts


leighmcculloch:
Logs to add:
- [ ] Any action the results in a db operation (insert, update, delete).
- [ ] Each successful or unsuccessful authentication (the unsuccessful case might be something to remove if we see a lot of them)
- [ ] Each successful transaction signed including the transaction hash in the log

Alerts to add:
- [ ] Any error resulting in a 500 being returned to the client.

leighmcculloch:
#2567 resolves this issue, and I've opened #2568 for a specific improvement.

PR #2403, opened 2020-03-21
exp/services/recoverysigner: use postgres to persist account data

### LINKED ISSUE #2339: exp/services/recoverysigner: add data persistence with postgres

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add an implementation of `account.Store` that uses Postgres to persist accounts, identities, and their auth methods in three tables.

### Why

To persist accounts so they are not just stored in memory.

Note that hard deletes are used, not soft deletes. We don't really have a use case for soft deletes since it's not a feature of the application to offer to a user the ability to undelete. Instead of soft deletes I plan to add an audit table that tracks operations on the main tables. This keeps the main tables focused on data the application will actually use. This will also store the history of an account in an easier to use format for any administrative view that might be placed on the data. To see what the audit table will look like, take a look at c50d7608077982cedfc646767df03b8c75120dac.

Close #2339 

### Known limitations

The schema assumes a single database master using incremental bigints as primary keys for tables. 

Memory store is still the default store to ensure backwards compatibility with any existing deployment. We probably won't maintain that backwards compatibility indefinitely since this is an experimental app but it is easy not to break the app today.


exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200311000000-create-accounts.sql
```diff
@@ -0,0 +1,15 @@
+-- +migrate Up
+
+CREATE TABLE accounts (
+  id BIGSERIAL NOT NULL PRIMARY KEY,
+
+  created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
```
leighmcculloch: Given that I plan to add an audit table alongside each table (see c50d7608077982cedfc646767df03b8c75120dac which is a work in progress) I'm no longer convinced we need a `created_at` column each table. If we plan to maybe surface it in the API back to the client then I think we should keep it, but if we don't I think we should drop it.

exp/services/recoverysigner/internal/account/db_store_delete.go
```diff
@@ -0,0 +1,33 @@
...
+	}
+
+	err = tx.Commit()
+	if err != nil {
+		return err
+	}
+
+	return nil
+
+	return nil
```
marcelosalloum: Remove duplicated return statement:
```suggestion
	return nil
```

exp/services/recoverysigner/internal/account/db_store_get.go
```diff
@@ -0,0 +1,90 @@
...
+	for rows.Next() {
+		var r struct {
+			AccountID       int64   `db:"account_id"`
+			AccountAddress  string  `db:"account_address"`
+			IdentityID      *int64  `db:"identity_id"`
+			IdentityRole    *string `db:"identity_role"`
+			AuthMethodType  *string `db:"auth_method_type"`
+			AuthMethodValue *string `db:"auth_method_value"`
+		}
+		err = rows.StructScan(&r)
```
marcelosalloum: #TIL

exp/services/recoverysigner/internal/account/db_store_get.go
```diff
@@ -0,0 +1,90 @@
...
+
+			if r.AuthMethodType != nil && r.AuthMethodValue != nil {
+				m := AuthMethod{
+					Type:  AuthMethodType(*r.AuthMethodType),
+					Value: *r.AuthMethodValue,
+				}
+				i.AuthMethods = append(i.AuthMethods, m)
+			}
+
+			a.Identities[identityIndex] = i
```
marcelosalloum: Quick question: is there any limit to the amount of identities an account can have?

howardtw:
👀

exp/services/recoverysigner/internal/account/db_store_get.go
```diff
@@ -0,0 +1,90 @@
...
+func (s *DBStore) getAccounts(where string, args ...interface{}) ([]Account, error) {
+	query := `SELECT
+			accounts.id AS account_id,
+			accounts.address AS account_address,
+			identities.id AS identity_id,
+			identities.role AS identity_role,
+			auth_methods."type" AS auth_method_type,
+			auth_methods.value AS auth_method_value
+		FROM accounts
+		LEFT JOIN identities ON identities.account_id = accounts.id
```
howardtw: Do we use LEFT JOIN on purpose? Maybe it should be `INNER JOIN` as it can be confusing to return a NULL identity or a NULL auth_method? I see we have the type be pointers below and have all the associated logic , so I assume you know this is expected. This is more of a question.

exp/services/recoverysigner/internal/account/db_store_get.go
```diff
@@ -0,0 +1,90 @@
...
+		if !ok {
+			a := Account{Address: r.AccountAddress}
+			accounts = append(accounts, a)
+			accountIndex = len(accounts) - 1
+			accountIndexByAccountID[r.AccountID] = accountIndex
+		}
+		a := accounts[accountIndex]
+
+		if r.IdentityID != nil || r.IdentityRole != nil {
+			identityIndex, ok := identityIndexByIdentityID[*r.IdentityID]
```
howardtw: Should we check for `r.IdentityID != nil` before dereferencing it? We only check for `r.IdentityID != nil || r.IdentityRole != nil`, meaning that we can get a case where `r.IdentityID == nil && r.IdentityRole != nil`?

exp/services/recoverysigner/internal/account/db_store_delete.go
leighmcculloch: Oops! Thanks. 23617fb8

exp/services/recoverysigner/internal/account/db_store_delete.go
```diff
@@ -0,0 +1,33 @@
+package account
+
+func (s *DBStore) Delete(address string) error {
+	tx, err := s.DB.Beginx()
```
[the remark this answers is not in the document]
leighmcculloch: Nice catch, thanks, we do not need this. Removed in f94e1ec8.

exp/services/recoverysigner/internal/account/db_store_delete.go
```diff
@@ -0,0 +1,33 @@
...
+	if err != nil {
+		return err
+	}
+	rowsAffected, err := result.RowsAffected()
+	if err != nil {
+		return err
+	}
+	if rowsAffected == 0 {
+		return ErrNotFound
+	}
```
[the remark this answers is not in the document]
leighmcculloch: I think it makes little difference here if we return the error or not. If a user tries to delete an account a second time they'll get a 404 before they hit this line of code because they can no longer authenticate for that account. Having this function return not found just makes it more consistent with some other methods like the `Get`. Let me know if I'm missing something.

exp/services/recoverysigner/internal/account/db_store_find_with_identity.go
```diff
@@ -0,0 +1,37 @@
+package account
+
+import "github.com/lib/pq"
+
+func (s *DBStore) FindWithIdentityAuthMethod(t AuthMethodType, value string) ([]Account, error) {
+	query := `SELECT account_id
+		FROM auth_methods
+		WHERE "type" = $1
+		AND value = $2`
```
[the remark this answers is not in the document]
leighmcculloch: The schema no longer has a deleted_at because this PR does hard deletes not soft deletes. See the PR description. I'll add auditing as a separate follow up. An example is here for how the auditing will work: c50d760.

exp/services/recoverysigner/internal/account/db_store_get.go
```diff
@@ -0,0 +1,90 @@
...
+			accounts.address AS account_address,
+			identities.id AS identity_id,
+			identities.role AS identity_role,
+			auth_methods."type" AS auth_method_type,
+			auth_methods.value AS auth_method_value
+		FROM accounts
+		LEFT JOIN identities ON identities.account_id = accounts.id
+		LEFT JOIN auth_methods ON auth_methods.identity_id = identities.id
+		WHERE ` + where + `
+		ORDER BY accounts.id, identities.id, auth_methods.id`
```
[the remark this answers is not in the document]
leighmcculloch: See https://github.com/stellar/go/pull/2403/files/964f5a103d0ff77cba17fb7f4a56d708ac1ee31f#r396615828.

exp/services/recoverysigner/internal/account/db_store_get.go
leighmcculloch: This is a feature of [`github.com/jmoiron/sqlx`], not of the [`database/sql`] package. It makes some things easier.

[`github.com/jmoiron/sqlx`]: https://github.com/jmoiron/sqlx
[`database/sql`]: https://golang.org/pkg/database/sql/

exp/services/recoverysigner/internal/account/db_store_get.go
leighmcculloch: I've made this code more defensive in 670fc2fe. Technically the two fields will only ever be `nil` together due to the `LEFT JOIN` in the query, so what you're describing won't happen, but the code the way it was written was fraught with traps. In the commit I've moved dereferencing to be right next to checks so that the danger code is close together and easily spotted, and I've changed the `||` to `&&` to ensure we never end up in this block unless all are dereferencable.

exp/services/recoverysigner/internal/account/db_store_get.go
leighmcculloch: There is no limit enforced by the code yet, but we should add a limit as part of #2388. I've updated that issue to include adding a limit.

exp/services/recoverysigner/internal/account/db_store_add.go
```diff
@@ -0,0 +1,56 @@
...
+			INSERT INTO identities (account_id, role)
+			VALUES ($1, $2)
+			RETURNING id
+		`, accountID, i.Role)
+		if err != nil {
+			return err
+		}
+
+		for _, m := range i.AuthMethods {
+			authMethodID := int64(0)
```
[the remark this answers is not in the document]
leighmcculloch: Thanks, we're not, removed in 8b1124f5. I was planning to add logging for each of these inserts, but decide to defer that to #2358. I added a mention of adding this specific type of log to that issue.

exp/services/recoverysigner/internal/account/db_store_get.go
leighmcculloch: Yup, the LEFT JOIN is on purpose. Its so that the store returns an account regardless of whether it has any identities, and an identity is returned regardless of whether it has any auth methods. We could probably just disallow those situations with validations though. I'll think about that some more independent of this PR and when doing the validations if we decide to make it so that every account needs one identity and every identity needs one auth method we can come back to this function and simplify the query.

exp/services/recoverysigner/internal/account/db_store_find_with_identity.go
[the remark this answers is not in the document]
howardtw: What's the advantage of having this audit system work along the side vs using a soft deletion? I wasn't aware that we are doing this in the previous draft PR. I can see that we are able to catch the `update` with the audit system.

exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200311000000-create-accounts.sql
howardtw: My thought is that we should keep this until we know that the audit system works as expected. On top of that, I think having the timestamp in the API can potentially make it more clear when clients register their accounts.

exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200311000002-create-auth-methods.sql
```diff
@@ -0,0 +1,27 @@
...
+);
+
+CREATE TABLE auth_methods (
+  account_id BIGINT NOT NULL REFERENCES accounts (id) ON DELETE CASCADE,
+  identity_id BIGINT NOT NULL REFERENCES identities (id) ON DELETE CASCADE,
+  id BIGSERIAL NOT NULL PRIMARY KEY,
+
+  created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
+
+  "type" auth_method_type NOT NULL,
```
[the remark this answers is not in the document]
leighmcculloch: We could, we could also use the term `kind` instead which is very common replacement of `type`. I don't feel strongly going either way. I doubt we'll need to type `"type"` so often as to cause us problems, but I'm happy to rename `type` to `kind` everywhere. Does that work for you?

exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200311000000-create-accounts.sql
```diff
@@ -0,0 +1,15 @@
...
+
+  address TEXT NOT NULL
```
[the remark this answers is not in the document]
leighmcculloch: The size concern is one reason, but the bigger concern is that primary keys that are provided from outside our system makes it difficult to scale a database beyond a single master db in the future. If we define the primary keys, we control how to uniquely identify an account, or what DB an account is recorded in. We also make it difficult to distinguish between accounts that have been added a second time in the audit table and in logs. I prefer to always use a separate column for the primary key for these reasons. Let me know if you've got any concerns.

exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200311000000-create-accounts.sql
leighmcculloch: I'll keep it for now, but I'll have no code interact with it, which is the reason I changed this to be `DEFAULT NOW()`. We can drop the column if we'd like to after the fact.

leighmcculloch: Actually, now I'm wondering if we're going to have created_at, then I guess we should also have updated_at to record when the row gets updated. But again, we won't actually use these two fields for anything so maybe I should leave both out. 🤔

exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200311000002-create-auth-methods.sql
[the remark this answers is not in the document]
accordeiro: Yep, I like replacing it with `kind`. I feel like the trickiest part is that whenever you're debugging something / running something in the db directly, it's easy to forget about the `"` and all you'll get back is a not very intuitive syntax / parse error.

exp/services/recoverysigner/internal/account/db_store_add.go
```diff
@@ -0,0 +1,56 @@
+package account
+
+import "github.com/lib/pq"
+
+func (s *DBStore) Add(a Account) error {
+	tx, err := s.DB.Beginx()
```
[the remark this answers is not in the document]
leighmcculloch: I considered that but I think this would be a pretty complex `WITH` clause to write because there are three levels of many-to-many relationships that would involve unnecesting. While using a WITH might be more efficient, I don't think it is worth us optimizing early if the query becomes significantly more complex.

If you've got a simple way to write a CTE for this, could you open a PR against this PR with it?

exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200311000002-create-auth-methods.sql
```diff
@@ -0,0 +1,27 @@
+-- +migrate Up
+
+CREATE TYPE auth_method_type AS ENUM (
+  'stellar_address',
+  'phone_number',
+  'email'
+);
+
+CREATE TABLE auth_methods (
+  account_id BIGINT NOT NULL REFERENCES accounts (id) ON DELETE CASCADE,
```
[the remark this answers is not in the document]
leighmcculloch: It is redundant, we also don't need to use it in our queries as is evidenced [here](https://github.com/stellar/go/blob/8b1124f5a4d9ee96a8cf1c511c2ea3a656ff01df/exp/services/recoverysigner/internal/account/db_store_get.go#L29) but it makes crafting queries manually at a console a lot easier since you don't have to join through multiple tables, and in multi-tenant systems it's not uncommon to tag all data with the tenant to make it harder to accidentally misinterpret one tenants data for the wrong tenant. The queries we write that insert into these tables will be few, the queries we write to read out of the tables will be many if we include all the times we manually run our own queries when looking at data. I think we should keep it for these reasons.

Let me know if you think that concern of possibility of inconsistency trumps this.

exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200311000000-create-accounts.sql
leighmcculloch: I've included `updated_at` in dcf272c9. We can drop these two columns if we decide they are invaluable. If we're going to require one, I think we should have the other too.

exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200311000000-create-accounts.sql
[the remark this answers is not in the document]
accordeiro: Nope, those are very valid points. Thanks! :)

exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200311000002-create-auth-methods.sql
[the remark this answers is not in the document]
accordeiro: That makes sense. I think the goal is to always write to the db programatically, so it shouldn't be a big deal, as long as we remember to fill both fields – and I think making it not null already helps make sure we do this. 

Worst case scenario the source of truth should be the `account_id` in the `identities` table.

exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200311000002-create-auth-methods.sql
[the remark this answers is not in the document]
leighmcculloch: Actually your idea to continue to use the existing concepts in the name was better. Using `kind` is introduces a new concept that we then have to map to `type` in the API, and new concepts is a cognitive burden. I renamed it to `type_` in b296a0ea.

leighmcculloch:
Since this PR was approved I've made one final change to use IDENTITY instead of SERIAL for primary keys. They are preferred in Postgres v10+. Postgres' [don't do this page](https://wiki.postgresql.org/wiki/Don%27t_Do_This#Don.27t_use_serial) even mentions that we shouldn't use SERIAL unless we really need to. IDENTITY also works much better with audit tables (where we mirror tables off each other), so there's motivation for us to use it in our specific situation.

As part of supporting running tests with a newer version of Postgres I need to make larger changes to the CircleCI test runs, so I'm doing that in #2413, and once that PR merges the diff of this PR will shrink a bit again because I've temporarily merged those changes into this PR to get the tests to pass.

Issue #2402, opened 2020-03-20
Horizon: Add all endpoints to root '_links' response


JakeUrban:
### What problem does your feature solve?
The Ruby SDK uses Horizon's `hal+json` content via [Hyperclient](https://github.com/codegram/hyperclient), an API client for auto-discovering endpoints. It reads horizon's root endpoint response and constructs methods for each endpoint listed under `_links`.

However, many endpoints are simply missing, making it difficult to access these endpoints via the Ruby SDK.

### What would you like to see?
All endpoints listed in Horizon's [API Reference ](https://www.stellar.org/developers/reference/) included in the response to `GET https://horizon.stellar.org` under `_links`.

### What alternatives are there?
The Ruby SDK doesn't have any alternative other than building explicitly defined requests for each endpoint, which would be a ton of work and not worth the effort compared to this fix.

ire-and-curses:
Can you enumerate which endpoints have this problem?

JakeUrban:
It would be easier to list what is already present. If you click [this](https://horizon.stellar.org/) you'll see we have properties like `account_transactions` under `_links`. The Ruby SDK takes these and automatically builds methods:
```
client = Stellar::Client.default
client.horizon.account_transactions(account_id:'<account_id')
```
But for endpoints like `account/{account_id}/offers`, there is no such entry under `_links` and therefore the Ruby SDK's client doesn't have a natural way to hit it.

tomerweller:
Endpoints linked the root resource:
- [ ] Account Details
- [ ] Accounts
- [ ] All Assets
- [ ] Data for Account
- [ ] All Effects
- [ ] Effects for Account
- [ ] Effects for Ledger
- [ ] Effects for Operation
- [ ] Effects for Transaction
- [ ] Fee Stats
- [ ] All Ledgers
- [ ] Ledger Details
- [ ] Metrics
- [ ] Offer Details
- [ ] Offers for Account
- [ ] Offers
- [ ] All Operations
- [ ] Operations for Account
- [ ] Operations for Ledger
- [ ] Operations for Transaction
- [ ] Operation Details
- [ ] Orderbook Details
- [ ] Strict Receive Payment Paths
- [ ] Strict Send Payment Paths
- [ ] Find Payment Paths
- [ ] All Payments
- [ ] Payments for Account
- [ ] Payments for Ledger
- [ ] Payments for Transaction
- [ ] Trade Aggregations
- [ ] Trades for Account
- [ ] Trades for Offer
[...]

PR #2407, opened 2020-03-24
Add missing top-level HAL links (except for /POST transactions)

2opremio:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

Addresses part of #2402 

This change adds the missing top-level HAL links to `GET /`, with the exception of `POST /transactions` which clashes with `GET /transactions` and isn't properly covered by HAL.


protocols/horizon/main.go
```diff
@@ -298,13 +298,22 @@ type Root struct {
...
 		Offers              *hal.Link `json:"offers,omitempty"`
+		Operation           hal.Link  `json:"operation"`
+		Operations          hal.Link  `json:"operations"`
 		OrderBook           hal.Link  `json:"order_book"`
+		Payments            hal.Link  `json:"payments"`
 		Self                hal.Link  `json:"self"`
 		StrictReceivePaths  *hal.Link `json:"strict_receive_paths"`
 		StrictSendPaths     *hal.Link `json:"strict_send_paths"`
+		TradeAggregations   hal.Link  `json:"trade_aggregations"`
+		Trades              hal.Link  `json:"trades"`
```
leighmcculloch: It's interesting that no test needed to change when adding these, which suggests that we don't have any tests for this structure. Given this PR it sounds like the structure of this response is important, so we probably should have tests for it.

ire-and-curses: We discussed this. Our thought is adding the tests would either be duplicating data (which wouldn't help us remember to add a new endpoint), or generative (in which case the work is similar to generating these in the code, which is something we want to do). So for now we won't test this. I actually couldn't find an issue for the latter work, will raise it.

PR #2409, opened 2020-03-24
Remove all references to Redis

2opremio:
<details>
  <summary>PR Checklist</summary>
  
### PR Structure

* [x] This PR has reasonably narrow scope (if not, break it down into smaller PRs).
* [x] This PR avoids mixing refactoring changes with feature changes (split into two PRs
  otherwise).
* [x] This PR's title starts with name of package that is most changed in the PR, ex.
  `services/friendbot`, or `all` or `doc` if the changes are broad or impact many
  packages.

### Thoroughness

* [x] This PR adds tests for the most critical parts of the new functionality or fixes.
* [x] I've updated any docs ([developer docs](https://www.stellar.org/developers/reference/), `.md`
  files, etc... affected by this change). Take a look in the `docs` folder for a given service,
  like [this one](https://github.com/stellar/go/tree/master/services/horizon/internal/docs).

### Release planning

* [x] I've updated the relevant CHANGELOG ([here](services/horizon/CHANGELOG.md) for Horizon) if
  needed with deprecations, added features, breaking changes, and DB schema changes.
* [x] I've decided if this PR requires a new major/minor version according to
  [semver](https://semver.org/), or if it's mainly a patch change. The PR is targeted at the next
  release branch if it's not a patch change.
</details>

Fixes #2395 

### What

Remove all Redis references.

### Why

It wasn't used, it was only adding complexity and causing test failures.

### Known limitations

N/A

PR #2413, opened 2020-03-24
all: run tests on postgres v10

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Run tests on Postgres v10 and add function to `support/db/dbtest` that exposes the current version of Postgres that's running.

### Why
Recoverysigner (`exp/services/recoverysigner`) is going to be using the IDENTITY feature of Postgres that was introduced in v10 which makes it easier and safer to copy records into an audit table, as well as being the recommended way of doing auto-incrementing primary keys now. Recoverysigner's tests are going to use the new version function to skip themselves if the tests are being run against an older version of Postgres.

Also I heard that at some point we'll be wanting to support Postgres v11 on Horizon and run tests for it, and so now seems like a good time to start running tests for a newer version of Postgres against to provide some confidence towards that in the future.

### Known limitations

This will increase build times a little because more builds will need to finish.

### Alternative approaches considered

We could use Go build tags so that the tests that require Postgres 10 are separated and only run in certain cases, but it's really annoying on developers to need to add build tags when running Go commands and it will be easy to accidentally not run the tests. Skipping is much simpler for a developer to interact with. Tests that are skipped will output a special message saying why they were skipped.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -177,8 +177,8 @@ jobs:
       - install_go_deps
       - test_packages
 
-  # test_code_1_13_postgres10 performs all package tests using Go 1.13 and Postgres 10.
-  test_code_1_13_postgres10:
+  # test_code_1_13_postgres11 performs all package tests using Go 1.13 and Postgres 11.
+  test_code_1_13_postgres11:
     working_directory: /go/src/github.com/stellar/go
     docker:
       - image: circleci/golang:1.13-stretch
@@ -190,7 +190,7 @@ jobs:
           PGDATABASE: circle_test
           MYSQL_HOST: 127.0.0.1
           MYSQL_PORT: 3306
-      - image: circleci/postgres:10-alpine
+      - image: circleci/postgres:11-alpine
         environment:
           POSTGRES_USER: circleci
       - image: circleci/redis:5.0-alpine
@@ -220,8 +220,8 @@ jobs:
       - test_packages
       - send_coverage_report
 
-  # test_code_1_14 performs all package tests using Go 1.14 and Postgres 10.
-  test_code_1_14_postgres10:
+  # test_code_1_14 performs all package tests using Go 1.14 and Postgres 11.
+  test_code_1_14_postgres11:
     working_directory: /go/src/github.com/stellar/go
     docker:
       - image: circleci/golang:1.14-stretch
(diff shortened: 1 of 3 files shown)


leighmcculloch:
@accordeiro We can keep an eye on it, but it will be reasonably difficult to downgrade a version of Postgres once we commit this change, so I only want to go ahead if we don't think it will cause issues. The reason it will be difficult to unwind is because once we have this deployed migrations can only move forward and so it is difficult to change old migrations that contain features that are incompatible with older version of Postgres.

Comparing these two builds as a sample, I think we're good on the test duration. The total change is 4 seconds.
- Before change 2m 19s ([CircleCI](https://app.circleci.com/pipelines/github/stellar/go/3701/workflows/22f75112-a9e3-499a-9f51-98aeddde12d7))
- After change 2m 23s ([CircleCI](https://app.circleci.com/pipelines/github/stellar/go/3752/workflows/c1ff3384-469c-4574-bc9a-1f2f9621cd69))

This is definitely going to vary, but because the tests run in parallel we shouldn't see a massive change.

leighmcculloch:
@2opremio I saw that you've made changes to the CircleCI config in #2409 on the `release-horizon-v1.1.0` branch.

Did you want to cherry-pick that to `master` before I merge this PR?

Once I merge this I think you'll end up with conflicts when it comes time for the Horizon release branch and master to converge and I don't think your change needs to be isolated to the release branch.

Let me know if you want me to wait for your cherry-pick, otherwise I'll merge this tomorrow.

jacekn:
FYI Ubuntu 18.04 ships with postgres 10 and I think it's very important to make sure everything works on latest Ubuntu LTS. I'm not sure what the differences are between 10 and 11 but for safety it would be good to run tests against both 10 and 11 to make sure we support both 18.04 and upcoming 20.04 which should ship with postgres 11.

leighmcculloch:
I think adding 4 additional builds might be a bit much right now. I think we'll plan to deploy Recoverysigner with Postgres 11, but the minimum version it needs will be 10, so I can update these builds to build with 10 since it will be the minimum for that app and we can evaluate if we want to add tests for 11 too, or maybe hold off on that until Horizon's minimum changes.

PR #2414, opened 2020-03-25
exp/services/recoverysigner: add audit postgres tables

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add an audit table (Postgres) for each existing table, that mirrors the columns of the existing table and tracks inserts, updates, and deletes.

### Why

It will be useful to be able to track changes in the identities and auth methods that are registered to accounts. It will allow support staff to investigate issues and to understand the flow of events that led to an account's auth methods changing.

Audit tables are very non-intrusive to the rest of the application since the application has zero knowledge of the tables.

The audit tables could potentially be accessed by an administrative dashboard to view the history when investigating an issue or debugging.

### Known limitations

This will cause data to be store in the audit tables for indefinitely unless we come up with some strategy for removing audit records associated with accounts that have been entirely deleted. We can address this problem separately if it is necessary.

### Alternative approaches

We could use a soft-delete / append-only approach where we keep all the history of every row in the existing tables. I abandoned this approach because it means the existing tables bloat over time with data not immediately needed for the service to operate. If the table got too big it would be complex to break it up because it is both written to and read from, and navigating those writes and reads would require changes to the application. In contrast, the solution in this PR would be simpler to work with because we could direct writes to a new table while working on the old table.

### Merging

This PR should be merged to `master` after #2403. It's diff contains that PR, and so this diff will get smaller then the other PR is merged.


exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200320000000-create-accounts-audit.sql
```diff
@@ -0,0 +1,43 @@
...
+CREATE TYPE audit_op AS ENUM (
+  'INSERT',
+  'UPDATE',
+  'DELETE'
+);
+
+CREATE TABLE accounts_audit (
+  audit_id BIGINT NOT NULL PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
+  audit_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
+  audit_user TEXT DEFAULT USER,
```
marcelosalloum: Is this the `PG_USER` that's performing the operation?

exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200320000000-create-accounts-audit.sql
```diff
@@ -0,0 +1,43 @@
...
+  audit_op audit_op NOT NULL,
+  LIKE accounts
+);
+
+-- +migrate StatementBegin
+CREATE FUNCTION record_accounts_audit() RETURNS TRIGGER AS $BODY$
```
marcelosalloum: Is the name `$BODY$` relevant in any way, or is it just a placeholder name with its context limited to the `record_accounts_audit()` function?

exp/services/recoverysigner/internal/db/db_audit_test.go
```diff
@@ -0,0 +1,66 @@
...
+func TestAuthMethodsAudit(t *testing.T) {
+	db := dbtest.Open(t)
+	conn, err := Open(db.DSN)
+	require.NoError(t, err)
+
+	assertAuditColsEqualTableCols(t, conn, "auth_methods", "auth_methods_audit")
+}
+
+// assertAuditColsEqualTableCols checks that the audit table for the given
+// table has the same columns as the given table.
```
marcelosalloum:
```suggestion
// assertAuditColsEqualTableCols checks that the audit table for the given
// table has the same columns as the given table, except for the header columns.
```

exp/services/recoverysigner/internal/db/db_audit_test.go
```diff
@@ -0,0 +1,66 @@
...
+func assertAuditColsEqualTableCols(t *testing.T, db *sqlx.DB, tableName, auditTableName string) {
```
marcelosalloum: Since this test is not really checking if the column names are exactly equal, but if one contains the other, maybe we could rename this function to `assertAuditColsContainTableCols`.

exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200320000000-create-accounts-audit.sql
howardtw: I have the same question regarding what `DEFAULT USER` means too. ☝️

exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200320000000-create-accounts-audit.sql
howardtw: It doesn't mean anything here. It's just making it more clear. It can be replaced by `$$` without any problem. People usually put the token in between `$$` when there are more than one quoting to make it easier to read.

exp/services/recoverysigner/internal/account/db_store_delete_test.go
```diff
@@ -128,6 +162,34 @@ func TestDelete(t *testing.T) {
...
+		type row struct {
+			AuditOp string `db:"audit_op"`
+			Value   string `db:"value"`
+		}
+		rows := []row{}
+		err = session.Select(&rows, `SELECT audit_op, value FROM auth_methods_audit`)
+		require.NoError(t, err)
+		wantRows := []row{
+			{AuditOp: "INSERT", Value: "GD4NGMOTV4QOXWA6PGPIGVWZYMRCJAKLQJKZIP55C5DGB3GBHHET3YC6"},
+			{AuditOp: "INSERT", Value: "+10000000000"},
```
howardtw: I wonder if we can add a column to the audit table indicating which account these values are associated with. It's not easy to understand what has happen to a specific account. Right now we only have that specific information in the `accounts` and `accounts_audit` table. In order to see what's been changed to a specific account, we always have to do a join query. It probably makes sense in the recovery signers db, but I think it would be more ergonomic to gather information/activity in an account in single table. Or at least we should replicate the account address to the `identities_audit ` and `auth_methods_audit` tables.

exp/services/recoverysigner/internal/account/db_store_delete_test.go
```diff
@@ -96,6 +96,22 @@ func TestDelete(t *testing.T) {
 		}
 		assert.ElementsMatch(t, wantRows, rows)
 	}
+	// Check that account 1 delete has been audited
```
howardtw: This comment is a little confusing since we are checking both `INSERT` and `DELETE` in this case. Same applies to the new test cases in this PR.

leighmcculloch: The test is checking the entire state of the table, so it's confirming that the state shows the deletion.

exp/services/recoverysigner/internal/account/db_store_delete_test.go
leighmcculloch: I think it's important that the audit tables mirror their base tables because the audit tables only have access to their base tables when the audit records are being created. To do what you're suggesting we'd need to add the account address to all the `auth_methods` records, which seems unnecessary.

A simple join will get you this information.

You mention:
>In order to see what's been changed to a specific account, we always have to do a join query.

What problems do you see with needing to do a join?

exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200320000000-create-accounts-audit.sql
leighmcculloch: It's the postgres user who is running the query mutating the audited table. For the most part it will be the user that the application is authenticated as. If an administrators needs to manually alter any rows it will hopefully show their username assuming they are using unique credentials for themselves. It will mean looking at the audit history we can see what the application has done and what us administrators have done.

exp/services/recoverysigner/internal/account/db_store_delete_test.go
howardtw: OK. I just didn't see a value of having a tuple `INSERT, +10000000000` without additional information by looking at this test case.

> What problems do you see with needing to do a join?

I was hoping to see an easy-to-understand audit table. It's not straightforward to construct a query with two joins because one has to understand which columns to join with, which join to use, whether it should be a left join or inner join, etc. To know the questions above, that person has to understand the constraints and table structure very well. (Should we expect all our authorized users to have the knowledge?)

Perhaps it's not a big issue if we plan on integrating the admin panel with it because we would have built-in queries. I was purely thinking from a standpoint when an authorized user logs into the database and wants to see what's going on. It could be hard for her. It would make her life easier if she can simply do `SELECT * FROM accounts_audit WHERE address = $1` and get back all the changes to that account.

Doing a join is not a blocker. I was wondering the pros and cons of making the audit system not just mirror the base tables but something that summarizes the activities.

exp/services/recoverysigner/internal/account/db_store_delete_test.go
howardtw: But I thought the two insertions have also been audited?

exp/services/recoverysigner/internal/account/db_store_delete_test.go
leighmcculloch: >I just didn't see a value of having a tuple INSERT, +10000000000 without additional information by looking at this test case.

There are more columns in the table that are listed in this test, I was just testing the bare necessity to keep the test signal to noise ratio high. There is an `account_id` column that lets you join to the `accounts` table, or to the `accounts_audit` table so you can get the address.

>I was wondering the pros and cons of making the audit system not just mirror the base tables but something that summarizes the activities.

The pro in this approach is it is completely disconnected from the application itself, which means most database operations are audited even if they aren't performed by the app. The app can perform it, or a user can perform it.

A downside of this approach is that because it uses a trigger attached to the table, it only has access to the row that is being inserted, updated, or deleted.

Another pro is that the audit table is very simple to build, and to be honest we don't know how it will be used yet so something simpler that will be easier to change in the future is a good place to start.

This pattern is also not one I invented, but one that is common. The pattern comes straight out of Postgres' own examples for ways to use triggers. There are a few ways to go about implementing it, here's a couple if you're curious. Ours is more like the first but not identical.
- https://www.postgresql.org/docs/10/plpgsql-trigger.html#PLPGSQL-TRIGGER-AUDIT-EXAMPLE
- https://wiki.postgresql.org/wiki/Audit_trigger

If we make our own history tracking facility that tracks activities as defined by the application that is more application code we need to write and maintain and that won't capture changes we manually make. Because we don't know exactly how this will be used I think this would mean us making more decisions upfront, and it would be more effort.

>I was hoping to see an easy-to-understand audit table

The problems you described with the audit table are the same with the main tables. The main tables require joins, so I think your concern about joins is really a concern with the main tables since the audit tables are just mirroring them.

Could you share some full examples of what you're proposing changing about the base or audit tables?

exp/services/recoverysigner/internal/account/db_store_delete_test.go
leighmcculloch: They have been, but not as part of the delete, they were inserted earlier in the test setup.

leighmcculloch:
Note: Something I haven't given a ton of thought to or included are indexes for the audit tables. I think we'll need some indexes to make querying them efficient, we also don't need to decide today though and a slow query using a table scan is not the-end-of-the-world for something just interacted to by us, unlike the application serving up requests. So I was thinking of adding indexes as we need them. Let me know if you have any concerns with that and I can look at adding indexes in this PR.

exp/services/recoverysigner/internal/db/db_audit_test.go
leighmcculloch: 👍 c14ccb1b

exp/services/recoverysigner/internal/account/db_store_delete_test.go
howardtw: It would be a combination of the three
```
CREATE TABLE accounts_audit (
   address (Stellar account pubkey)
   ops (INSERT/UPDATE DELETE)
   role (owner/sender/receiver)
   auth_type (stellar address/phone number/email)
   value (the corresponding value to auth_type)
   timestamp (the time the change happened to the account)
)
```
With this, we can see the changes happened to an account
```
('GXX...', 'INSERT', sender, null, null, '2020-03-26T00:00:00Z')
('GXX...', 'INSERT', null, phone_number, '+10000000000', '2020-03-26T00:00:01Z')
('GXX...', 'UPDATE', null, phone_number, '+10000000001', '2020-03-26T00:00:02Z')
('GXX...', 'INSERT', null, email, 'user@stellar', '2020-03-26T00:00:02Z')
...
```

Feel free to merge this change, and we will see how it goes. My concern was just that it's not going to be easier to audit a specific account at a glance. Maybe mirroring all the changes of individual tables provides greater flexibility in the long run.

howardtw:
It we were to add indexes, it would be better done in another PR for sure 😄

exp/services/recoverysigner/internal/account/db_store_delete_test.go
accordeiro: From my perspective, I'd say ideally we won't have to touch these audit tables – we're just keeping them so we have an append-only record of everything that happened in the db. I feel like any optimization at this point might be an overkill, as long as what we're doing _enables_ us to do optimizations in the future (which I believe is the case since we're just augmenting the original schema).

I'd support moving forward as-is (as per @howardtw's suggestion), since we'll already keep track of everything that happens, and then making improvements as we see real-life use cases emerge.

leighmcculloch: @howardtw I think that table structure is definitely easier to understand, and it would be great to use that for `accounts` too, but that structure doesn't take into accounts that an account has multiple identities and auth methods are associated with individual identities that one day might be updatable independently.

I think that table structure would be awesome if SEP-30 was limited to storing an account associated with a single identity, but for the SEP-30 we have where multiple identities are involved I think we need to retain the multi-table structure. It's more complex because unfortunately the SEP is more complex.

exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200320000000-create-accounts-audit.sql
```diff
@@ -0,0 +1,43 @@
...
+  'UPDATE',
+  'DELETE'
+);
+
+CREATE TABLE accounts_audit (
+  audit_id BIGINT NOT NULL PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
+  audit_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
+  audit_user TEXT DEFAULT USER,
+  audit_op audit_op NOT NULL,
+  LIKE accounts
```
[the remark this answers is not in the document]
leighmcculloch: Your understanding of `LIKE` is correct. There is no relationship between the two tables, it just copies the existing columns of `accounts` over to `accounts_audit`. I could copy and paste the columns and it would be the same result (almost, because the identity attribute doesn't get copied, and the defaults won't get copied).

You are correct that if we add new columns to `accounts`, or drop a column from `accounts`, we'll need to do the same in `accounts_audit`. The PR contains tests that guarantee that the table's columns are kept in sync. The PR's tests don't guarantee that the types are the same though, I'll improve them to do that.

However, there is no need to update the trigger. The trigger has been created so that it copies the columns from a record being added in one table, over into the audit table. The columns names in both aren't specified. It's the `NEW.*` in:
https://github.com/stellar/go/blob/b80fbc13e2754d1a0d747bed0443ba23c8f0e57a/exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200320000000-create-accounts-audit.sql#L21

Do you think that'll be okay? I think it will because we will guarantee the column names and types are the same and are in the same order.

I'm still trying to confirm that we can alter both tables inside a transaction at the same time to ensure that transaction in-flight can fail because one table updates and the other is not yet updated. I think it will be okay but I want to confirm.

`INHERITS` would mean that we wouldn't need to have the columns redefined, but anything in one table would show up in queries of the other table. It's intended for something like object-oriented parent-child relationships. It should really only be used when the child is a parent. In our case the accounts_audits rows are not accounts, there isn't a one-to-one mapping, it's one-to-many.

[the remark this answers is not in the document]
accordeiro: Oh that's interesting – I thought `INHERITS` only impacted the table structure, I didn't realize it had any side effects on the table contents. I guess I skipped this part of the [documentation](https://www.postgresql.org/docs/10/sql-createtable.html): 

> and by default the data of the child table is included in scans of the parent(s).

So it seems like `LIKE` is our only choice.

Good point about `NEW.*`, so we just need to make sure we migrate both tables, which is already covered by the tests. So I think we're good to go :)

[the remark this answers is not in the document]
leighmcculloch: I've updated the tests that check the audit table to also test the data type of the column and whether the column is nullable. That's in: 8b083fe, 8adb061, 2d4a27d.

exp/services/recoverysigner/internal/account/db_store_delete_test.go
howardtw: > that structure doesn't take into account that an account has multiple identities and auth methods are associated with individual identities that one day might be updatable independently.

You're right. It's hard to tell whose phone number that is in the following structure:
```
('GXX...', 'INSERT', null, phone_number, '+10000000000', '2020-03-26T00:00:01Z')
('GXX...', 'UPDATE', null, phone_number, '+10000000001', '2020-03-26T00:00:02Z')
```

Again I had agreed to move forward as it is. I just wanted to have a discussion on whether it's worth organizing our audit tables differently than the recovery signer's.

PR #2416, opened 2020-03-25
services/horizon: Add Last-Ledger header to mutable GET endpoints

### LINKED ISSUE #2233: Send `Latest-Ledger` header in all endpoints
We started sending `Latest-Ledger` header for state related endpoints but we should do it for all endpoints. It will help reducing a number of false positives in `horizon-cmp`.

2opremio:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

Fixes #2233

### What

Add Last-Ledger header to mutable GET endpoints

### Why

Reduce false-positives out of `horizon-cmp`

### Known limitations

N/A


services/horizon/internal/web.go
```diff
@@ -241,20 +245,25 @@ func (w *web) mustInstallActions(
...
+		r.Get("/trades", TradeIndexAction{}.Handle)
+		r.Get("/trade_aggregations", TradeAggregateIndexAction{}.Handle)
+	})
 
-	// trading related endpoints
-	r.Get("/trades", TradeIndexAction{}.Handle)
-	r.Get("/trade_aggregations", TradeAggregateIndexAction{}.Handle)
 	// /offers/{offer_id} has been created above so we need to use absolute
 	// routes here.
 	r.Get("/offers/{offer_id}/trades", TradeIndexAction{}.Handle)
```
tamirms: this endpoint should be wrapped in the middleware because if an offer is not entirely consumed it's possible for new trades to occur on that offer. For example if you have an offer selling 100,000 XLM, it's possible for one trade to consume 50,000 XLM from the offer on ledger `n` and for another trade to consume the remaining 50,000 XLM in a subsequent ledger.

services/horizon/internal/web.go
```diff
@@ -210,18 +212,20 @@ func (w *web) mustInstallActions(
...
+		r.Get("/accounts/{account_id:\\w+}/trades", TradeIndexAction{}.Handle)
+	})
 	// ledger actions
 	r.Route("/ledgers", func(r chi.Router) {
-		r.Get("/", LedgerIndexAction{}.Handle)
+		r.Get("/", latestLedger(LedgerIndexAction{}.Handle))
 		r.Route("/{ledger_id}", func(r chi.Router) {
 			r.Get("/", LedgerShowAction{}.Handle)
-			r.Get("/transactions", w.streamIndexActionHandler(w.getTransactionPage, w.streamTransactions))
+			r.Get("/transactions", latestLedger(w.streamIndexActionHandler(w.getTransactionPage, w.streamTransactions)))
```
tamirms: we don't need to use the middleware here because the set of transactions included in a specifc ledger should never change

services/horizon/internal/middleware.go
```diff
@@ -350,3 +350,36 @@ func (m *StateMiddleware) Wrap(h http.Handler) http.Handler {
...
+// and the nested http handler will be executed normally
+func (m *LatestLedgerMiddleware) Wrap(f http.Handler) http.Handler {
+	return m.WrapFunc(f.ServeHTTP)
+}
+
+// WrapFunc executes the middleware on a given HTTP handler function
+// If the latest ledger cannot be reliably obtained, the header won't be set
+// and the nested http handler will be executed normally
+func (m *LatestLedgerMiddleware) WrapFunc(f http.HandlerFunc) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
```
tamirms: this middleware should set up a repeatable read transaction for the entire request just like in the StateMiddleware because otherwise the database can be updated right after `lastIngestedLedger, ready, err := ingestionStatus(q)` is called and then the response will not be consistent with the ledger in the header.

Once we add a repeatable read transaction then we also need to add the `sseRequest` condition into the middleware as well. The only difference between `LatestLedgerMiddleware` and `StateMiddleware` is that `LatestLedgerMiddleware` does not need to check `GetExpStateInvalid` and it can ignore the `ready` value from 
`ingestionStatus()`.

I wonder if it would make sense to have one middleware struct which can be configured to match `StateMiddleware` or `LatestLedgerMiddleware`

services/horizon/internal/actions_payment_test.go
```diff
@@ -2,18 +2,22 @@ package horizon
...
 
 	"github.com/stellar/go/protocols/horizon"
 	"github.com/stellar/go/protocols/horizon/operations"
 	"github.com/stellar/go/services/horizon/internal/db2/history"
 	"github.com/stellar/go/services/horizon/internal/expingest"
+	"github.com/stellar/go/services/horizon/internal/test/db"
 )
 
 func TestPaymentActions(t *testing.T) {
+	fmt.Println("Horizon URL:", db.HorizonURL())
```
tamirms: are the `fmt.Println()` lines still necessary?

tamirms:
@2opremio this looks correct to me! The only concern I have is that I don't think `StateMiddleware` is the right name for the middleware anymore. 

There are two types of data served by horizon: time series data and aggregated data. An example of time series data would be `/transactions`, `/effects`, or `/operations`. These endpoints return a list of objects spanning several ledgers. Each individual object (for example a transaction) is immutable.

Aggregated data is something which gets updated with every ledger. An example of that would be an account balance (in ledger `n` the balance is 100, in ledger `n+1` the balance is 110 because it received a payment). We call endpoints which serve aggregated data "state" endpoints because the response from such an endpoint reflects the current *state* of an object.

Previously, the `StateMiddleware` was only used on state endpoints. Now, you are extending the functionality to cover other types of endpoints. The one thing that the middleware does for both the state and time series endpoints is that it adds the `Last-Ledger` header to the response. So, I think the middle ware should be renamed to `LatestLedgerMiddleware`

2opremio:
I am not sure `LatestLedgerMiddleware` is the best name because it also propagates a DB session which is later consumed (the name doesn't describe that).

How about `MutableEndpoint` or `InpureEndpoint`?

services/horizon/internal/actions_payment_test.go
2opremio: No, good catch

services/horizon/internal/action.go
```diff
@@ -44,7 +44,12 @@ func (action *Action) CoreQ() *core.Q {
 // horizon's database.
 func (action *Action) HistoryQ() *history.Q {
 	if action.hq == nil {
-		action.hq = &history.Q{Session: action.App.HorizonSession(action.R.Context())}
+		var err error
+		action.hq, err = actions.HistoryQFromRequest(action.R)
```
2opremio: @tamirms maybe we can now remove the explicit calls for getting the session context in orderbook and friends?

tamirms: yeah, I think that's a good idea

Issue #2405, opened 2020-03-23
Remove duplicate indexes


tamirms:
I noticed today that there is a duplicate index on the id column of the history transactions table:

https://github.com/stellar/go/blob/master/services/horizon/internal/db2/schema/migrations/1_initial_schema.sql#L247
```sql
CREATE UNIQUE INDEX hs_transaction_by_id ON history_transactions USING btree (id);
```

https://github.com/stellar/go/blob/master/services/horizon/internal/db2/schema/migrations/1_initial_schema.sql#L344
```sql
CREATE UNIQUE INDEX index_history_transactions_on_id ON history_transactions USING btree (id);
```

I thought that there might be more duplicate indexes so I ran the sql queries [here](https://gist.github.com/jfrost/6382059) to detect if there were more:

```
psql (9.6.14, server 9.6.17)
Type "help" for help.
horizon=# SELECT indrelid::regclass
horizon-#      , array_agg(indexrelid::regclass)
horizon-#   FROM pg_index
horizon-#  GROUP BY indrelid
horizon-#      , indkey
horizon-# HAVING COUNT(*) > 1;
       indrelid       |                        array_agg                        
----------------------+---------------------------------------------------------
 history_ledgers      | {hs_ledger_by_id,index_history_ledgers_on_id}
 history_transactions | {hs_transaction_by_id,index_history_transactions_on_id}
(2 rows)
horizon=# SELECT indrelid::regclass
horizon-#      , array_agg(indexrelid::regclass)
horizon-#   FROM pg_index
horizon-#  GROUP BY indrelid
horizon-#      , indkey[0]
horizon-# HAVING COUNT(*) > 1;
       indrelid       |                               array_agg                               
----------------------+-----------------------------------------------------------------------
 exp_asset_stats      | {exp_asset_stats_pkey,exp_asset_stats_by_code}
 history_transactions | {hs_transaction_by_id,index_history_transactions_on_id}
 history_assets       | {history_assets_asset_code_asset_type_asset_issuer_key,asset_by_code}
 history_ledgers      | {hs_ledger_by_id,index_history_ledgers_on_id}
(6 rows)
```

I've confirmed that `hs_ledger_by_id,index`, `history_ledgers_on_id` are also duplicate indexes. 

Also, it seems that we forgot to drop the `exp_asset_stats` table in https://github.com/stellar/go/blob/master/services/horizon/internal/db2/schema/migrations/32_drop_exp_history_tables.sql

[...]

PR #2419, opened 2020-03-26
services/horizon/internal/db2/schema: Remove unnecessary tables and indexes

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Close https://github.com/stellar/go/issues/2405

The asset_stats table is not used anymore because it has been replaced by exp_asset_stats.

index_history_transactions_on_id is not necessary because it is identical to hs_transaction_by_id

index_history_ledgers_on_id is not necessary because it is identical to hs_ledger_by_id

asset_by_code is not necessary because there is a compound index history_assets_asset_code_asset_type_asset_issuer_key
on `UNIQUE(asset_code, asset_type, asset_issuer)` which can be used on queries involving the `asset_code` column.

for the same reasons, exp_asset_stats_by_code can also be removed

### Why

These duplicate indexes don't improve query performance but do inhibit write performance. Removing the indexes will reduce storage and improve database write performance.

### Known limitations

N/A


2opremio:
> These duplicate indexes don't improve query performance but do inhibit write performance. Removing the indexes will reduce storage and improve database write performance.

Out of curiosity. Do you have an authoritative source on redundant indexes causing performance problems? I have been looking it up and couldn't find anything. Intuitively, I don't see why Postgres would optimize them out, but maybe that's the case.

services/horizon/internal/db2/schema/migrations/33_remove_unused.sql
```diff
@@ -0,0 +1,29 @@
...
+
+DROP INDEX exp_asset_stats_by_code;
+
+DROP INDEX index_history_transactions_on_id;
+
+DROP INDEX index_history_ledgers_on_id;
+
+DROP INDEX asset_by_code;
+
+-- +migrate Down
```
2opremio: Where do we use backwards migration?

tamirms: there's a command called `horizon db migrate down <count>` which executes the backwards migration:

https://github.com/stellar/go/blob/master/services/horizon/cmd/db.go#L50

tamirms:
@2opremio I couldn't find anything authoritative except for the "Duplicate indexes" section in https://wiki.postgresql.org/wiki/Index_Maintenance

However, I did run a SQL query on the horizon staging db to get index size and usage statistics. Here is the output: 

```
 schemaname |            tablename             |                       indexname                       |  num_rows   | table_size | index_size | unique | number_of_scans |  tuples_read  | tuples_fetched 
------------+----------------------------------+-------------------------------------------------------+-------------+------------+------------+--------+-----------------+---------------+----------------
 public     | exp_asset_stats                  | exp_asset_stats_pkey                                  |        9502 | 1416 kB    | 1560 kB    | Y      |          161001 |        449735 |         160990
 public     | exp_asset_stats                  | exp_asset_stats_by_code                               |        9502 | 1416 kB    | 408 kB     | N      |         3588799 |     569239833 |      568999115
 public     | history_assets                   | history_assets_asset_code_asset_type_asset_issuer_key |       26816 | 4920 kB    | 3840 kB    | Y      |        28632933 |      28632910 |       28632910
 public     | history_assets                   | asset_by_code                                         |       26816 | 4920 kB    | 792 kB     | N      |        15349015 |      25577560 |       25577560
 public     | history_ledgers                  | hs_ledger_by_id                                       | 2.67591e+07 | 20 GB      | 619 MB     | Y      |               0 |             0 |              0
 public     | history_ledgers                  | index_history_ledgers_on_id                           | 2.67591e+07 | 20 GB      | 619 MB     | Y      |       116066364 |     235943505 |      234242133
 public     | history_transactions             | hs_transaction_by_id                                  | 3.31465e+08 | 579 GB     | 7233 MB    | Y      |             102 |           102 |            102
 public     | history_transactions             | index_history_transactions_on_id                      | 3.31465e+08 | 579 GB     | 7233 MB    | Y      |    365093787345 |  150133108513 |   150015537260
```

If you look at the pairs (`hs_ledger_by_id`, `index_history_transactions_on_id`) and (`hs_transaction_by_id`, `index_history_transactions_on_id`) you'll find that the two indexes take up the same exact amount of space. Presumably, there must be time spent writing to the duplicate indexes. If we were to remove them then it would no longer be necessary to perform those write operations

services/horizon/internal/db2/schema/migrations/33_remove_unused.sql
2opremio: Ah, that strikes me as odd. Is that actively used? (I couldn't find any references to it)

2opremio:
That makes sense. Thanks for the explanation . I wonder why `hs_transaction_by_id ` gets to be used instead of having a `0` counter.

Anyways, LGTM.

services/horizon/internal/db2/schema/migrations/33_remove_unused.sql
tamirms: I don't think it's used very often but one use case I can think of is if you want to roll back to an earlier version of horizon. In that case, you may need to roll back some migrations as well

2opremio: Ah, really good point.

Issue #2357, opened 2020-03-04
Support stellar-core protocol 13 (CAP 18) - "Fine-Grained Control of Authorization"


abuiles:
The upcoming `stellar-core` [protocol 13 release](https://github.com/stellar/stellar-core/projects/18) will add support for [CAP 18](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0018.md) ("Fine-Grained Control of Authorization").

**Horizon**

- [x] Add `is_authorized_to_maintain_liabilities` to balance attributes
- [x] Add `authorized_to_main_liabilities` to operation details when Op is OperationAllowTrust https://github.com/stellar/go/blob/protocol-13/services/horizon/internal/expingest/processors/operations_processor.go#L260 
- [x] Extend `AllowTrust` in protocols to include Flags https://github.com/abuiles/go/blob/e12b580711a108b61a53bc58891ca2916f054c5a/protocols/horizon/operations/main.go#L191-L197 -
- [x] Add a new effect called `history.EffectTrustlineAuthorizedToMaintainLiabilities` https://github.com/abuiles/go/blob/e12b580711a108b61a53bc58891ca2916f054c5a/services/horizon/internal/expingest/processors/effects_processor.go#L539-L559  
- [x] Add `history.EffectTrustlineAuthorizedToMaintainLiabilities` to effects resource adapter https://github.com/abuiles/go/blob/e12b580711a108b61a53bc58891ca2916f054c5a/services/horizon/internal/resourceadapter/effects.go#L100-L103
- [x] No changes are required in the trustline ingestion processor  [flags are already handled as an int](https://github.com/abuiles/go/blob/10ee142de653ec94644444a24537d2e457c63c13/xdr/xdr_generated.go#L2003-L2004).

**Backward compatibility**
Since the binary representation of `bool authorize`  is the same as `uint32 authorize` for `true` or `false`,  we can roll this changes in Horizon and SDKS before protocol 13 is deployed to stellar core. 
They'll only have problems if they try to set `authorize` to `AUTHORIZED_TO_MAINTAIN_LIABILITIES_FLAG` since it won't be recognize by protocol 12.

Again, if the CAP doesn't go through, we can roll the changes in a safe manner.

abuiles:
[...]

PR #2423, opened 2020-03-27
services/horizon: CAP0018 Support

abuiles:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add initial support for CAP0018. 

### Why

Horizon needs to support CAP0018. To do this we have to update ingestion and all the JSON related structs.

At the ingestion level the following changes will happen: 

#### Operation Details

We now add both flags to operation details 

##### Before

```
{
  "authorize":    true,
  "asset_code":   "COP",
  "asset_issuer": "GDRW375MAYR46ODGF2WGANQC2RRZL7O246DYHHCGWTV2RE7IHE2QUQLD",
  "asset_type":   "credit_alphanum4",
  "trustee":      "GDRW375MAYR46ODGF2WGANQC2RRZL7O246DYHHCGWTV2RE7IHE2QUQLD",
  "trustor":      "GDQNY3PBOJOKYZSRMK2S7LHHGWZIUISD4QORETLMXEWXBI7KFZZMKTL3"
}
```

##### After

```

{
  "authorize":  true,
  "authorize_to_maintain_liabilities": true,
  "asset_code":                        "COP",
  "asset_issuer":                      "GDRW375MAYR46ODGF2WGANQC2RRZL7O246DYHHCGWTV2RE7IHE2QUQLD",
  "asset_type":                        "credit_alphanum4",
  "trustee":                           "GDRW375MAYR46ODGF2WGANQC2RRZL7O246DYHHCGWTV2RE7IHE2QUQLD",
  "trustor":                           "GDQNY3PBOJOKYZSRMK2S7LHHGWZIUISD4QORETLMXEWXBI7KFZZMKTL3"
}
```

The following is how the JSON response for an operation will look like when the flag is `2`:

```
{
  "id": "124042211741474817",
  "paging_token": "124042211741474817",
  "transaction_successful": true,
  "source_account": "GBSTRUSD7IRX73RQZBL3RQUH6KS3O4NYFY3QCALDLZD77XMZOPWAVTUK",
  "type": "allow_trust",
  "type_i": 7,
  "created_at": "2020-03-27T03:40:10Z",
  "transaction_hash": "a77d4ee5346d55fb8026cdcdad6e4b5e0c440c96b4627e3727f4ccfa6d199e94",
  "asset_type": "credit_alphanum4",
  "asset_code": "USD",
  "asset_issuer": "GBSTRUSD7IRX73RQZBL3RQUH6KS3O4NYFY3QCALDLZD77XMZOPWAVTUK",
  "trustee": "GBSTRUSD7IRX73RQZBL3RQUH6KS3O4NYFY3QCALDLZD77XMZOPWAVTUK",
  "trustor": "GA332TXN6BX2DYKGYB7FW5BWV2JLQKERNX4T7EUJT4MHWOW2TSGC2SPM",
  "authorize":                         false,
  "authorize_to_maintain_liabilities": true,
}
```

And the following when it is `1`:

```
{
  "id": "124042211741474817",
  "paging_token": "124042211741474817",
  "transaction_successful": true,
  "source_account": "GBSTRUSD7IRX73RQZBL3RQUH6KS3O4NYFY3QCALDLZD77XMZOPWAVTUK",
  "type": "allow_trust",
  "type_i": 7,
  "created_at": "2020-03-27T03:40:10Z",
  "transaction_hash": "a77d4ee5346d55fb8026cdcdad6e4b5e0c440c96b4627e3727f4ccfa6d199e94",
  "asset_type": "credit_alphanum4",
  "asset_code": "USD",
  "asset_issuer": "GBSTRUSD7IRX73RQZBL3RQUH6KS3O4NYFY3QCALDLZD77XMZOPWAVTUK",
  "trustee": "GBSTRUSD7IRX73RQZBL3RQUH6KS3O4NYFY3QCALDLZD77XMZOPWAVTUK",
  "trustor": "GA332TXN6BX2DYKGYB7FW5BWV2JLQKERNX4T7EUJT4MHWOW2TSGC2SPM",
  "authorize":                         true,
  "authorize_to_maintain_liabilities": true,
}
```

#### Effect Details

We add a new effect `history.EffectTrustlineAuthorizedToMaintainLiabilities`.

The JSON representation is the same as `history.EffectTrustlineAuthorized` and `history.EffectTrustlineDeauthorized` - except for the effect type and type id.

#### Balance resource

Similar to the operation response, we'll show both flags.

#### Trustline with flag 0 OR 1

```
"balances": [
    {
      "is_authorized": false,
      "is_authorized_to_maintain_liabilities": false,
      "balance": "27.1374422",
      "limit": "922337203685.4775807",
      "buying_liabilities": "0.0000000",
      "selling_liabilities": "0.0000000",
      "last_modified_ledger": 28893780,
      "asset_type": "credit_alphanum4",
      "asset_code": "USD",
      "asset_issuer": "GBSTRUSD7IRX73RQZBL3RQUH6KS3O4NYFY3QCALDLZD77XMZOPWAVTUK"
    },
    {
      "balance": "1.5000000",
      "buying_liabilities": "0.0000000",
      "selling_liabilities": "0.0000000",
      "asset_type": "native"
    }
  ],
```

#### Trustline with flag 2

```
"balances": [
    {
      "is_authorized": false,
      "is_authorized_to_maintain_liabilities": true,
      "balance": "27.1374422",
      "limit": "922337203685.4775807",
      "buying_liabilities": "0.0000000",
      "selling_liabilities": "0.0000000",
      "last_modified_ledger": 28893780,
      "asset_type": "credit_alphanum4",
      "asset_code": "USD",
      "asset_issuer": "GBSTRUSD7IRX73RQZBL3RQUH6KS3O4NYFY3QCALDLZD77XMZOPWAVTUK"
    },
    {
      "balance": "1.5000000",
      "buying_liabilities": "0.0000000",
      "selling_liabilities": "0.0000000",
      "asset_type": "native"
    }
  ],
```

This is part of #2357

### Known limitations

[TODO or N/A]


services/horizon/internal/expingest/processors/transaction_operation_wrapper_test.go
```diff
@@ -857,3 +858,40 @@ func TestOperationParticipants(t *testing.T) {
...
+		transaction:    io.LedgerTransaction{},
+		operation:      op,
+		ledgerSequence: 1,
+	}
+
+	expected := map[string]interface{}{
+		"asset_code":                        "COP",
+		"asset_issuer":                      "GDRW375MAYR46ODGF2WGANQC2RRZL7O246DYHHCGWTV2RE7IHE2QUQLD",
+		"asset_type":                        "credit_alphanum4",
+		"authorize_to_maintain_liabilities": true,
```
abuiles: @tamirms @ire-and-curses @2opremio  I have some doubts on how to represent this and ultimately I decided to go on with this. 

The other alternative I consider was to do something similar to what we do in account flags, where we have `set_flags` and then `set_flags_s`:

```
"set_flags": []int32{1, 2, 4},
"set_flags_s": []string{
	"auth_required",
	"auth_revocable",
	"auth_immutable",
},
``` 

While this works and will keep things flexible, we know that at the moment we won't have scenario where you can turn on/off multiple flags.

I think doing `authorize_to_maintain_liabilities` keeps consistency with what we already have. 

I'm open to use `set_flags` too. If we go with that, I think we should remove `authorize` in favor of `set_flags`.

abuiles: I played with the code a little bit more and had a second read at the cap. 

I think adding only `authorized_to_maintain_liability` to the details should be enough. 

We can play with the presentation layer to decide which flags we want to show, here's a proposal 4b7772b345110bc1cf2b9db9219f55af3c5aa97f

- If the flag is `authorize` it should be enough, since  `AUTHORIZED_FLAG` implies `AUTHORIZED_TO_MAINTAIN_LIABILITIES_FLAG`. 
- If the flag is `authorize_to_maintain_liabilities` then we show both
- If the flag is false, we only show `authorize`

tamirms: @abuiles what is the advantage of omitting the `authorize` attribute in the json blob? having the attribute be optional might make parsing it a little bit more tricky for some SDKs because you can have to prepare for the fact the value can be true, false, or not present

abuiles: @tamirms this is only at the database layer,  if we are setting the flag to `2` we don't need to add an extra key to the JSON blob.

At presentation level, we can decide how to show it. `authorize` will always be included, but `authorize_to_maintain_liabilities` will only be shown if we are in `flag` 2. 

https://github.com/stellar/go/blob/4b7772b345110bc1cf2b9db9219f55af3c5aa97f/services/horizon/internal/resourceadapter/operations_test.go#L77-L85

My main point for not including always `authorize_to_maintain_liabilities` is that `authorize: true`, implies `authorize_to_maintain_liabilities: true`.

You raise a good point about the SDKs and what makes this specific attribute more confusing is that at the SDK you'll probably have to do something like `authorize_to_maintain_liabilities: json.authorized || json.authorized_to_maintain_liabities` 

I think it's okay to have optional keys in a JSON response (we already do this in some end-points) - however, for this particular scenario it seems like we'll put an extra burden on APIs by not adding it.

I'll update the code to always include both keys.

abuiles: @tamirms ok, this is addressed.

abuiles:
@bartekn 

> However the code in db2.history.TrustLine.IsAuthorizedToMaintainLiabilities doesn't. If I set AUTHORIZED_FLAG on a trust line, the is_authorized_to_maintain_liabilities on the Balance object will be false however authorize_to_maintain_liabilities on the operation will be true.

This is handled correctly in the balance object, if the authorize flag is `1`,  we show `is_authorized_to_maintain_liabilities` as true in balances -- see https://github.com/stellar/go/pull/2423/files#diff-667f71e250631e23bb40996734b0c0a1R53

Btw, we went back and forth on how to store and show this, check out for more context https://github.com/stellar/go/pull/2423#discussion_r399457899

bartekn:
@abuiles thanks! I see now that it was actually updated here: https://github.com/stellar/go/commit/da9f5263f3ad117f6fb6b0c44dbc5db04412491e#diff-fa18377d7302ceaa29836017d5030298R261.

PR #2424, opened 2020-03-27
support/config: allow multiple flags to be defined with the same name across commands

### LINKED ISSUE #2421: exp/services/recoverysigner: db-url flag doesn't work for serve command

master (https://github.com/stellar/go/commit/9cabbc6b9388505708306e644dfef159ebf64645)

### What did you do?

Start recoverysigner service with custom `db-url` parameter
```
# ./recoverysigner serve --db-url="postgres://test@localhost:5432/recoverysigner?sslmode=disable" --signing-key=SAZ5AWVWS4S3VVQYZZQG6OXXXXXXXXXXXACYL7C4XPFZ7JQAFGZNC7P5 --sep10-jwt-public-key='MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEpRudVMq5smVHsRgFvvK2L5s+8IBUENYep1OoN7w6vUwWy8A/GHYDjomu9/mhGHJSKMY7gZf9Q7HRXEl3MajtnQ==' --firebase-project-id=test --port=8001
```

Recoverysigner service start and work with passed database config.

### What did you see instead?

Recoverysigner cannot connect to database. It use default `db-url` value from migrate command.
```
# ./recoverysigner serve --db-url="postgres://test@localhost:5432/recoverysigner?sslmode=disable" --signing-key=SAZ5AWVWS4S3VVQYZZQG6OXXXXXXXXXXXACYL7C4XPFZ7JQAFGZNC7P5 --sep10-jwt-public-key='MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEpRudVMq5smVHsRgFvvK2L5s+8IBUENYep1OoN7w6vUwWy8A/GHYDjomu9/mhGHJSKMY7gZf9Q7HRXEl3MajtnQ==' --firebase-project-id=test --port=8001
INFO[2020-03-27T10:50:15.405Z] Signing key: GC5QMHPXFS4Q2SM6TBHOKDQ6OFDWIIOK5PVMQDDK4XP2UNKBPZGGEQMB  pid=26
INFO[2020-03-27T10:50:15.405Z] SEP-10 JWT Public key: &{{0xc0000b82c0} 74680410747149410791207293493561661728330103896099968772111105621295875407180 10310879459333139481880951607945613569488524728490535876588189242372502646173}  pid=26
INFO[2020-03-27T10:50:15.405Z] Database url: postgres://localhost:5432/?sslmode=disable  pid=26
WARN[2020-03-27T10:50:15.408Z] Error pinging to Database: <nil>              pid=26
INFO[2020-03-27T10:50:15.409Z] Starting SEP-XX Recover Signer server         pid=26
```

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Allow multiple flags to be defined with the same name that are attached to different commands.

### Why

Cobra allows the definition of flags on different commands that have the same name. There can be good reasons to do this like two subcommands requiring the same parameter, but other commands not requiring that parameter.

Viper also supports this too, if we bind flags into viper as they're required, but not if we bind them all upfront.

This comment https://github.com/spf13/viper/issues/233#issuecomment-479336184 suggests binding flags to wiper when a command is actually running. This has other advantages, like there are not parameters for other commands bound unnecessarily.

Fix #2421 

### Known limitations

Our command line usage is somewhat untested. I don't think we have extensive tests of Horizon's command line interface like we do for what we could define as the most critical components of Horizon. There's some risk in our ability to identify issues with any change to this area of code because of that. I've attempted to address the risk by adding some general tests to the `support/config` package.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -69,7 +69,7 @@ require (
 	github.com/spf13/cast v0.0.0-20150508191742-4d07383ffe94 // indirect
 	github.com/spf13/cobra v0.0.0-20160830174925-9c28e4bbd74e
 	github.com/spf13/jwalterweatherman v0.0.0-20141219030609-3d60171a6431 // indirect
-	github.com/spf13/pflag v0.0.0-20161005214240-4bd69631f475 // indirect
+	github.com/spf13/pflag v0.0.0-20161005214240-4bd69631f475
 	github.com/spf13/viper v0.0.0-20150621231900-db7ff930a189
 	github.com/stellar/go-xdr v0.0.0-20180917104419-0bc96f33a18e
 	github.com/stellar/throttled v2.2.3-0.20190823235211-89d75816f59d+incompatible
--- a/services/horizon/cmd/db.go
+++ b/services/horizon/cmd/db.go
@@ -28,6 +28,9 @@ var dbInitCmd = &cobra.Command{
 	Short: "install schema",
 	Long:  "init initializes the postgres database used by horizon.",
 	Run: func(cmd *cobra.Command, args []string) {
+		dbURLConfigOption.Require()
+		dbURLConfigOption.SetValue()
+
 		db, err := sql.Open("postgres", viper.GetString("db-url"))
 		if err != nil {
 			log.Fatal(err)
@@ -71,6 +74,9 @@ var dbMigrateCmd = &cobra.Command{
 				os.Exit(1)
 			}
 		}
+
+		dbURLConfigOption.Require()
+		dbURLConfigOption.SetValue()
 
 		db, err := sql.Open("postgres", viper.GetString("db-url"))
 		if err != nil {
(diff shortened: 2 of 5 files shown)


leighmcculloch:
This breaks horizon because it assumes flags are bound outside of commands I think. I'll need to look more into this.

leighmcculloch:
I think 38a453b resolves the issues for Horizon. My issue was that I wasn't ensuring bind would occur in the require case.

leighmcculloch:
In d48c92d I've made a small change to Horizon to require the `--db-url` param in the migration commands, because it wasn't being required before and it looks sort of accidental that the param worked before. A side-effect of that commit is that if the db-url flag is empty now the migration command will fail in the same way it does for the root/serve command.

leighmcculloch:
Thanks. I've run the commands that call `viper.Get*` directly and so I don't expect there to be issues, but there could be and I'll own any issues that are related to this change that show up next during the Horizon testing.

PR #2429, opened 2020-03-31
services/horizon/internal/db2/history: Fix fee bump migrations

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

* Change type of `fee_charged` column from integer to bigint. The XDR definition of `fee_charged` is a 64 bit integer.

* Remove defaults from the following columns in the fee bump migration: inner_transaction_hash, fee_account, inner_signatures

* Use partial indexes on the `inner_transaction_hash` and `fee_account` columns.

* Create biginteger column for the fee bump transaction fee because the xdr definition specifies that the fee in the fee bump envelope is a 64 bit integer (previously, the max fee was restricted to 32 bit values) 

### Why

The XDR definiton of fee_charged assumes a 64 bit representation. However, the database representation in Horizon has always been a 32 bit integer because it was only possible to set a 32 bit fee on a transaction. In protocol 13, it will be possible to set a 64 bit fee in a fee bump transaction. We must change the column type of fee_charged to biginteger before protocol 13 is released.

Unfortunately, changing the type of the column will trigger a rewrite of the history_transactions table. Note that multiple columns are changed using a single ALTER TABLE statement. According to https://www.postgresql.org/docs/9.6/sql-altertable.html
when multiple columns are changed using a single statement only one pass through the table is required to perform the operation:

"The main reason for providing the option to specify multiple changes in a single ALTER TABLE is that multiple table scans or rewrites can thereby be combined into a single pass over the table."

### Known limitations

Because a fee bump transaction can set a 64 bit fee, we will need to change the type of `MaxFee` and `FeeCharged` in the horizon response. While protocol 12 is still running this change should not affect the SDKs and other downstream clients. However, once protocol 13 is released, we will need to change the SDKs so that they are able to handle 64 bit max fee values in the Horizon response.

I tried to extend the ALTER TABLE statement so that we could backfill rows in `history_transactions` where the `successful` column is null but I couldn't find a way to do it so that all the commands are executed within a single statement. Instead, I decided to use `COALESCE` to interpret null `successful` values as `true` in all the select statements over the `history_transactions` table.


services/horizon/internal/db2/schema/migrations/34_fee_bump_transactions.sql
```diff
@@ -1,12 +1,12 @@
 -- +migrate Up
 
-ALTER TABLE history_transactions ADD inner_transaction_hash character varying(64) DEFAULT '' NOT NULL;
-ALTER TABLE history_transactions ADD fee_account character varying(64) DEFAULT '' NOT NULL;
-ALTER TABLE history_transactions ADD inner_signatures character varying(96)[] DEFAULT '{}'::character varying[] NOT NULL;
-ALTER TABLE history_transactions ADD inner_max_fee integer DEFAULT 0 NOT NULL;
+ALTER TABLE history_transactions ADD inner_transaction_hash character varying(64);
+ALTER TABLE history_transactions ADD fee_account character varying(64);
+ALTER TABLE history_transactions ADD inner_signatures character varying(96)[];
+ALTER TABLE history_transactions ADD new_max_fee bigint;
```
2opremio: Shouldn't the `NOT NULL`s be preserved?

tamirms: I think if you add a `NOT NULL` constraint it forces you to provide a non null default so that all the pre-existing rows in the table satisfy the constraint.

protocols/horizon/main.go
```diff
@@ -442,7 +442,7 @@ type Transaction struct {
 	AccountSequence    string              `json:"source_account_sequence"`
 	FeeAccount         string              `json:"fee_account"`
 	FeeCharged         int32               `json:"fee_charged"`
-	MaxFee             int32               `json:"max_fee"`
+	MaxFee             int64               `json:"max_fee"`
```
abuiles: @tamirms can you leave an action required here so we change this value to string in our next major bump?

tamirms:
@abuiles @2opremio @ire-and-curses this PR is ready for review again. Please take another look, thanks!

services/horizon/internal/db2/schema/migrations/34_fee_bump_transactions.sql
```diff
@@ -1,19 +1,27 @@
 -- +migrate Up
 
-ALTER TABLE history_transactions ADD inner_transaction_hash character varying(64) DEFAULT '' NOT NULL;
-ALTER TABLE history_transactions ADD fee_account character varying(64) DEFAULT '' NOT NULL;
-ALTER TABLE history_transactions ADD inner_signatures character varying(96)[] DEFAULT '{}'::character varying[] NOT NULL;
-ALTER TABLE history_transactions ADD inner_max_fee integer DEFAULT 0 NOT NULL;
+ALTER TABLE history_transactions
+    ALTER COLUMN fee_charged TYPE bigint,
+    ALTER COLUMN successful SET DEFAULT TRUE,
+    ALTER COLUMN successful SET NOT NULL,
```
abuiles, line 6: >     ALTER COLUMN successful SET NOT NULL,
do you know if we could end up in a state where the migration is invalid where there are fields with `NULL` values?  did you run a test locally against a simple db, did the migration work fine?

tamirms:
I tried to setup the migration to update rows where the successful column is null by setting the default value to true:

```sql
ALTER TABLE history_transactions
    ALTER COLUMN fee_charged TYPE bigint,
    ALTER COLUMN successful SET DEFAULT TRUE,
    ALTER COLUMN successful SET NOT NULL,
```

But, as @abuiles pointed out, adding the default value will only work for new rows. It will not update existing rows which means that we cannot add a `NOT NULL` constraint. I could have added an `UPDATE` statement which set all null `successful` values to true before executing the `ALTER TABLE` statement but that would mean the migration would need to do 2 full table scans.

I thought the best solution would be to keep the  `successful` column as is and use `COALESCE` whenever we query the `history_transactions` table to interpret null `successful` values as true.

PR #2430, opened 2020-03-31
all: Use Go module machinery to book-keep external tools

2opremio:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Track and maintain external tooling through Go's module machinery.

For that, I am using a [`tools.go`](https://github.com/golang/go/issues/25922#issuecomment-412992431) file, which seems to be an increasingly-adopted practice.

### Why

1. It centralizes version tracking of tooling instead of having versions spread around in shell-scripts
2. Where possible (if performance allows it) it saves developers from installing external tools, by using `go run`, making things work transparently (e.g. it took me a while to find out that the specific version of `go-bindata` to use was buried in `gogenerate.sh`)

### Known limitations

`go run` is slower than running a  binary (at least until https://github.com/golang/go/issues/33468 is resolved), so we need to be careful on where to use it (e.g. the `shadow` binary is installed because `go run` is too slow).


2opremio:
@leighmcculloch are we planning to break out the code base into different modules? If so, do you know when and where?

Also, regardless of this PR, the Go SDK should be its own module, so that it doesn't drag all the other dependencies.

Once that is done we can evaluate whether it makes sense to push this trick to the new modules or roll it back. Until then (based on the approvals in this PR) I think there is consensus on moving forward with this approach.

leighmcculloch:
There are plans to do it yup, see go modules [part 1](https://docs.google.com/document/d/1v9HXRSjI7L-JypoMf80_qeotXn2gfp0LnPxToZfr87k) and [part 2](https://docs.google.com/document/d/1-DytA21WTP-VXMl_mULGadwKe9LzgGKipyedDf0ul5U) documents. We've done part 1, haven't done part 2 yet.

I think it is worth considering this after part 2 if the tools.go can live in its own module so that it doesn't pollute any other module. In fact you could probably introduce just that one module now independent of part 2.

ire-and-curses:
@2opremio Yeah this is a plan we've had for a while but haven't had bandwidth to execute. I'd love it if you took some time to read through the links @leighmcculloch shared and then we can discuss what we want to do about it, and when. For now I think we should pause this PR to keep the Go SDK deps frozen. Happy to discuss early next week.

2opremio:
I still need to look at this, but I think we can close it

PR #2445, opened 2020-04-03
services/horizon/docker: Upgrade Stellar Core docker container to protocol 13

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

When you start Stellar Core in standalone mode it defaults to generating ledgers with protocol version equal to 0.

To upgrade the Stellar Core protocol version you need to run an "upgrades" command:

https://github.com/stellar/stellar-core/blob/master/docs/software/commands.md#http-commands

This commit adds a container which automatically upgrades the Stellar Core container to run on protocol 13.

### Why

Generating protocol 0 ledgers isn't a good default because it will not be compatible with Horizon ingestion unless you set `SUPPORTED_META_VERSION` to 2 in the Stellar Core config file. 

### Known limitations

There could be a use case where someone wants to run an older version of the protocol. To make that change you should edit the `core-upgrade` container command and change the `protocolversion` parameter.

### DIFF
--- a/services/horizon/docker/docker-compose.standalone.yml
+++ b/services/horizon/docker/docker-compose.standalone.yml
@@ -20,3 +20,4 @@ services:
     image: curlimages/curl:7.69.1
     command: ["-v", "-f", "http://host.docker.internal:11626/upgrades?mode=set&upgradetime=1970-01-01T00:00:00Z&protocolversion=13"]
     network_mode: '${NETWORK_MODE:-bridge}'
+    
--- a/services/horizon/docker/docker-compose.standalone.yml
+++ b/services/horizon/docker/docker-compose.standalone.yml
@@ -1,6 +1,9 @@
 version: '3'
 services:
   core:
+    depends_on:
+      - core-postgres
+      - core-upgrade
     ports:
       # add extra port for history archive server
       - "1570:1570"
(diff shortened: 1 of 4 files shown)


services/horizon/docker/docker-compose.standalone.yml
```diff
@@ -11,3 +14,9 @@ services:
     environment:
       - HISTORY_ARCHIVE_URLS=http://host.docker.internal:1570
       - NETWORK_PASSPHRASE=Standalone Network ; February 2017
+  # this container will invoke a request to upgrade stellar core to protocol 13
+  core-upgrade:
```
2opremio, line 18: shouldn't this container depend on core?

2opremio: (i.e. how are you ensuring that the service is running? )

services/horizon/docker/docker-compose.standalone.yml
```diff
@@ -11,3 +14,9 @@ services:
...
+    restart: on-failure
+    image: curlimages/curl:7.69.1
+    command: ["-v", "-f", "http://host.docker.internal:11626/upgrades?mode=set&upgradetime=1970-01-01T00:00:00Z&protocolversion=13"]
+    network_mode: '${NETWORK_MODE:-bridge}'
```
2opremio, line 22: missing newline

services/horizon/docker/docker-compose.standalone.yml
tamirms: technically it does depend on `core`. However, if I add a dependency on core then there will be a circular dependency because `core` also requires `core-upgrade`.  I think adding a dependency on `core` for `core-upgrade` is not necessary because no one will explicitly run the `core-upgrade` container. The `core-upgrade` container is supposed to be a byproduct of the `core` container.

If the `core-upgrade` container starts up before `core` that is fine because the curl request in the `core-upgrade` container will fail and that should cause the container to exit with a non zero status. The `restart: on-failure` attribute will cause docker compose to restart the `core-upgrade` container. Eventually the curl request will succeed once stellar core is up and running.

2opremio: fair enough, thanks for the explanation

PR #2446, opened 2020-04-03
services/horizon: Fix typo in GET /ledger(s) hal link

### LINKED ISSUE #2402: Horizon: Add all endpoints to root '_links' response

2opremio:
Followup to #2407 

It fixes a problem found by @JakeUrban , see https://github.com/stellar/go/issues/2402#issuecomment-608014439

Fixes #2402

### DIFF
--- a/services/horizon/internal/resourceadapter/root.go
+++ b/services/horizon/internal/resourceadapter/root.go
@@ -44,7 +44,7 @@ func PopulateRoot(
 	dest.Links.Assets = lb.Link("/assets{?asset_code,asset_issuer,cursor,limit,order}")
 	dest.Links.Effects = lb.Link("/effects{?cursor,limit,order}")
 	dest.Links.Ledger = lb.Link("/ledger/{sequence}")
-	dest.Links.Ledger = lb.Link("/ledgers{?cursor,limit,order}")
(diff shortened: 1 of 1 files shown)


JakeUrban:
Sorry I didn't review this soon enough, but we still need a `ledger` entry for [ledger details](https://www.stellar.org/developers/horizon/reference/endpoints/ledgers-single.html)

2opremio:
It should be there (check the line before the one which was modified)

JakeUrban:
Whoops you're right, nevermind!

PR #2450, opened 2020-04-04
exp/services/recoverysigner: change SEP10 JWT key config option to a JWKS

### LINKED ISSUE #2443: exp/services/recoverysigner: add support for any type of JWT keys
### What
Add support for any type of JWT keys, such as ECDSA, RSA, etc.

### Why
The recoverysigner currently assumes that a JWT key is a ECDSA key, but not everyone who might wish to run this will be using an ECDSA key. The key in use will be defined by the SEP-10 server that they're using the recoverysigner with. If the SEP-10 server is using an RSA key the recoverysigner will be incompatible.

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Change the SEP-10 Base64 encoded JWT of an ECDSA public key to be a JSON Web Key Set (JWKS) that is allowed to have a minimum and maximum of one key, and that can be of any key type supported by the `gopkg.in/square/go-jose.v2` package.

### Why
For several reasons:
1. To support configuring the JWT keys used with a standardized format. JSON Web Key Sets (JWKS) are defined in [RFC7517]. The format we're using now (ASN.1 DER base64 encoded) uses standards, but nothing specific to JWT, and that doesn't helps us configure the type of the key. See (2).
2. To support more than just ECDSA keys.
3. To make it possible in the future to add support for multiple keys in use at the same time.

This change is an accompanying change to #2451.

Closes #2443.

[RFC7517]: https://tools.ietf.org/html/rfc7517

### Known limitations

This change is a breaking change for the service, but since the service is in the `exp` experimental package we have the capability to make breaking changes.


exp/services/recoverysigner/internal/serve/auth/sep10.go
```diff
@@ -1,17 +1,17 @@
 package auth
 
 import (
-	"crypto/ecdsa"
 	"net/http"
+	"time"
 
-	"github.com/dgrijalva/jwt-go"
```
marcelosalloum, line 7: I believe the `"github.com/dgrijalva/jwt-go"` dependency will be removed after we merge #2451

leighmcculloch: I'm still using it in tests actually to create the JWTs so even after #2450 and #2451 are merged it will still be used by the tests. I'll look at following up later to update the tests to use the go-jose library for creating JWTs.

exp/services/recoverysigner/cmd/serve.go
```diff
@@ -51,10 +51,10 @@ func (c *ServeCommand) Command() *cobra.Command {
 			Required:  true,
 		},
 		{
-			Name:      "sep10-jwt-public-key",
-			Usage:     "Base64 encoded ECDSA public key used to validate SEP-10 JWTs",
+			Name:      "sep10-jwks",
+			Usage:     "JSON Web Key Set (JWKS) containing one public key used to validate SEP-10 JWTs",
```
accordeiro, line 55: It should be exactly one key, right? Maybe we could rephrase to:

`JSON Web Key Set (JWKS) containing exactly one public key, which will be used to validate SEP-10 JWTs`

accordeiro:
Also: we should probably also update the README to reflect the new params :)

leighmcculloch:
>You stated that this is a breaking change. It is breaking for whoever runs the service (because they need to change the config flags), but from an API user this change should be seamless, right?

The API is consistent, it's just the service itself depends on the new parameters and no longer uses the old parameters. If this was not experimental or was being used in deployments I'd stagger adding and removing the config options.

>Related to 1, we're simply adding support to other JWK formats, but not excluding anything we supported in the past. Is this correct?

We're adding support for the JWK format and removing the old format, so if someone was running this server with the old commands they'll get an error saying the new config options need to be given values.

exp/services/recoverysigner/cmd/serve.go
leighmcculloch: 💯 1b685166

leighmcculloch:
@accordeiro I've addressed your comments. Could you take another look?

PR #2451, opened 2020-04-04
exp/services/webauth: change JWT key config option to a JWK

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Change to using the JSON Web Key (JWK) format for configuring the JWK key used by the webauth service to sign JWTs.

### Why
The JSON Web Key (JWK) format is defined in [RFC7517] and makes it easy to configure a web service in a standard way without the service needing to know ahead of time the type of key or algorithm that will be used.  Today the service uses only an ECDSA key and the ES256 algorithm and to change that would require changes to code, but this config will allow us to specify any key and algorithm supported by the `gopkg.in/square/go-jose.v2` package.

This change is an accompanying change to #2450.

Related to #2443.

[RFC7517]: https://tools.ietf.org/html/rfc7517

### Known limitations

This change is a breaking change for the service, but since the service is in the `exp` experimental package we have the capability to make breaking changes.

### DIFF
--- a/exp/services/webauth/cmd/convjwtkey.go
+++ b/exp/services/webauth/cmd/convjwtkey.go
@@ -1,0 +1,58 @@
+package cmd
+
+import (
+	"encoding/json"
+
+	"github.com/spf13/cobra"
+	"github.com/stellar/go/exp/support/jwtkey"
+	supportlog "github.com/stellar/go/support/log"
+	"gopkg.in/square/go-jose.v2"
+)
+
+type ConvJWTKeyCommand struct {
+	Logger *supportlog.Entry
+}
+
+func (c *ConvJWTKeyCommand) Command() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "convjwtkey [key]",
+		Short: "Convert a JWT ECDSA key ASN.1 DER Base64 encoded to a JSON Web Key",
+		Run: func(_ *cobra.Command, args []string) {
+			c.Run(args)
+		},
+	}
+	return cmd
+}
+
+func (c *ConvJWTKeyCommand) Run(args []string) {
+	if len(args) != 1 {
+		c.Logger.Fatal("One key (ASN.1 DER Base64 encoded) must be provided.")
+	}
+
+	k, err := jwtkey.PrivateKeyFromString(args[0])
+	if err != nil {
+		c.Logger.Fatal(err)
+	}
+
+	alg := jose.ES256
+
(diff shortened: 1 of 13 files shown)


leighmcculloch:
### TODO

- [ ] Add a command for converting existing ASN.1 DER Base64 encoded private and public keys into JSON Web Keys.

exp/services/webauth/cmd/convjwtkey.go
```diff
@@ -0,0 +1,58 @@
...
+	"gopkg.in/square/go-jose.v2"
+)
+
+type ConvJWTKeyCommand struct {
+	Logger *supportlog.Entry
+}
+
+func (c *ConvJWTKeyCommand) Command() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "convjwtkey [key]",
```
accordeiro, line 18: This is definitely not critical, but I'd rather be explicit in the command name about what we're converting to what. I feel like `jwt2jwk` might paint a clearer picture of what the command takes as an argument and what it outputs. Does that make sense?

exp/services/webauth/cmd/serve.go
```diff
@@ -61,10 +61,10 @@ func (c *ServeCommand) Command() *cobra.Command {
 			Required:       true,
 		},
 		{
-			Name:      "jwt-key",
-			Usage:     "Base64 encoded ECDSA private key used for signing JWTs",
+			Name:      "jwk",
+			Usage:     "JSON Web Key (JWK) used for signing JWTs",
```
accordeiro, line 65: I understand there are symmetric keys that would render the private/public concept relevant. But maybe we could add a hint? Something like:

`If you have a public-private pair, this should be the private one.`

exp/services/webauth/cmd/convjwtkey.go
leighmcculloch: Great idea, I love this feedback. In d526953d I've renamed this to convjwtkey2jwk, and I've already renamed genjwtkey to genjwk, so if you were to look at those commands they tell a story that you used to genjwtkey, and now you genjwk, and you can use convjwtkey2jwk to convert a key between that old and new format.

exp/services/webauth/cmd/serve.go
leighmcculloch: 💯 8e40c60d

PR #2454, opened 2020-04-06
services/horizon/internal: Only use repeatable read transactions on state endpoints

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

In https://github.com/stellar/go/pull/2416 we extended the Last-Ledger header to cover additional endpoints which were not immutable.

@2opremio raised the following point before working on the PR:

> After giving this quite a bit of thought, isn't it overkill to set that header (and in turn access the DB for all endpoints) to make horizon-cmp happy?

After testing the 1.1.0 release candidate on staging we found evidence to confirm this hypothesis. The overhead of wrapping each request in a repeatable read transaction proved to significantly hinder performance.

We should remove this change from the 1.1.0 release. We can always reintroduce this change later on but at this point it would be better to explore other solutions to the underlying problem we are trying to solve (verifying correctness of a horizon release) that have less of a negative impact on performance.

Here are some graphs which display the performance hit:

image

The first half of the graph covers a period where the Last-Ledger header was enabled and the second half of the graph covers a period where the header was disabled.


2opremio:
wouldn't it be better to fully revert #2416 ?

tamirms:
> wouldn't it be better to fully revert #2416 ?

there are parts of the PR that I would like to keep. for example, the comments in the orderbook and path finding endpoints,  and the fixes to the test scenarios. Also, if we decide to enable the middleware in the future it will be easy to do so.

2opremio:
Fair enough.

> Also, if we decide to enable the middleware in the future it will be easy to do so

There is always git for that.

ire-and-curses:
Geez, those graphs make the point so clearly. Totally agree with this.

Issue #2448, opened 2020-04-04
 Error commiting changes from processor: error in *processors.TransactionProcessor.Commit


abuiles:
We got the following error while ingesting some ledgers in verify-range 

> time="2020-04-03T20:43:27.901Z" level=fatal msg="Error running processors on ledger: Error commiting changes from processor: error in *processors.TransactionProcessor.Commit: Error flushing transaction batch: error adding values while inserting to history_transactions: exec failed: pq: value \"2650000000\" is out of range for type integer" pid=1091

> time="2020-04-03T19:33:27.099Z" level=fatal msg="Error running processors on ledger: Error commiting changes from processor: error in *processors.TransactionProcessor.Commit: Error flushing transaction batch: error adding values while inserting to history_transactions: exec failed: pq: value \"2500000000\" is out of range for type integer" pid=1097

## Ranges
FROM: 24737727 TO: 24744127
FROM: 24334527 TO: 24340927

tamirms:
the errors occurred in ledgers 24338403 and 24740516
```
time="2020-04-03T19:33:26.920Z" level=info msg="Processing ledger" commit=true graph=true ledger=true pid=1097 sequence=24338403 service=expingest state=true
time="2020-04-03T19:33:27.099Z" level=fatal msg="Error running processors on ledger: Error commiting changes from processor: error in *processors.TransactionProcessor.Commit: Error flushing transaction batch: error adding values while inserting to history_transactions: exec failed: pq: value "2500000000" is out of range for type integer" pid=1097
```
```
time="2020-04-03T20:43:27.776Z" level=info msg="Processing ledger" commit=true graph=true ledger=true pid=1091 sequence=24740516 service=expingest state=true
time="2020-04-03T20:43:27.888Z" level=error msg="Error in ingestion state machine" current_state="verifyRange(fromLedger=24737727, toLedger=24744127, verifyState=true)" error="Error running processors on ledger: Error commiting changes from processor: error in *processors.TransactionProcessor.Commit: Error flushing transaction batch: error adding values while inserting to history_transactions: exec failed: pq: value \"2650000000\" is out of range for type integer" next_state=stop pid=1091 service=expingest
```

abuiles:
Fixed in #2456

PR #2456, opened 2020-04-06
services/horizon/internal/db2/history: Fix ingestion of max_fee and fee_charged transaction fields

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Close https://github.com/stellar/go/issues/2448

* Fix ingestion of fee_charged
* Change column type of `max_fee` from `integer` to `bigint`
* Fix ingestion of  `max_fee`

### Why

We changed the representation of the`fee_charged` column in postgres from `integer` to `bigint` in https://github.com/stellar/go/pull/2429/files . However, I forgot to update the code which inserted transaction rows to cast the `fee_charged` field as an `int64` instead of an `int32`.

The XDR representation of the `MaxFee` attribute of a v1 and v0 transaction envelope is an unsigned 32 bit integer. We have always represented the `max_fee` column in postgres as an `integer` type. Postgres can only store signed integers. This means the maximum value a postgres `integer` can hold is 2^31 - 1 = 2147483647. So if you try to insert an unsigned 32 bit integer which is larger than 2^31 -1, e.g. 2500000000, you will encounter this error: `pq: value \"2650000000\" is out of range for type integer`.

There are two ways to fix this:

(1) cast `max_fee` to a 32 bit signed integer when inserting transactions into the db. When you read transactions from the db you have to remember to cast the 32 bit signed integer into an unsigned 32 bit integer.
(2) change the column type of `max_fee` to `bigint`

The reason this postgres error didn't manifest before is because, previously, we were casting `max_fee` to a 32 bit  signed integer:

https://github.com/stellar/go/blob/horizon-v1.0.1/services/horizon/internal/db2/history/transaction_batch_insert_builder.go#L156

However, we did not cast the 32 bit signed integer back into an unsigned 32 bit integer when reading rows from the db. Consequently, the "fee_charged" field in the Horizon response is negative for some transactions:

https://horizon.stellar.org/transactions/7cf06b94575aa6e87112fbd58fea751a3c27f30e5fbe2ae47192c79d0a56bc07

```
{
 ...
  "source_account_sequence": "67319953052663812",
  "fee_charged": 100,
  "max_fee": -1794967296,
  "operation_count": 1,
...
}
```

I have opted for changing the column type of `max_fee` to `bigint` because it is a more robust solution. We do not have to have any special decoding logic when reading transactions out from the db. 

Although changing the column type of `max_fee` requires rebuilding the `history_transactions` table, we are already doing that work for the `fee_charged`. Adding an additional clause to the `ALTER TABLE` statement should not lengthen the time required to run the migration.


tamirms:
For future reference, the way I debugged this was I modified the `core-start.sh` script in the docker compose workflow to load ledger `24338403` because I knew that the ingestion error occurred on that ledger:

```diff
diff --git a/services/horizon/docker/core-start.sh b/services/horizon/docker/core-start.sh
index 89e90fba..498fc588 100755
--- a/services/horizon/docker/core-start.sh
+++ b/services/horizon/docker/core-start.sh
@@ -9,18 +9,19 @@ cat stellar-core.cfg
 
 # initialize new db
 stellar-core new-db
-
-if [ "$1" = "standalone" ]; then
-  # start a network from scratch
-  stellar-core force-scp
-
-  # initialze history archive for stand alone network
-  stellar-core new-hist vs
-
-  # serve history archives to horizon on port 1570
-  pushd /history/vs/
-  python3 -m http.server 1570 &
-  popd
-fi
-
-exec /init -- stellar-core run
+stellar-core catch up 24338403/1
+
+#if [ "$1" = "standalone" ]; then
+#  # start a network from scratch
+#  stellar-core force-scp
+#
+#  # initialze history archive for stand alone network
+#  stellar-core new-hist vs
+#
+#  # serve history archives to horizon on port 1570
+#  pushd /history/vs/
+#  python3 -m http.server 1570 &
+#  popd
+#fi
+#
+#exec /init -- stellar-core run
```

Then I modified ingestion to flush to the `history_transactions` table after every single transaction was processed:

```diff
diff --git a/services/horizon/internal/expingest/processors/transactions_processor.go b/services/horizon/internal/expingest/processors/transactions_processor.go
index f8c2c830..24017fbd 100644
--- a/services/horizon/internal/expingest/processors/transactions_processor.go
+++ b/services/horizon/internal/expingest/processors/transactions_processor.go
@@ -22,16 +22,18 @@ func NewTransactionProcessor(transactionsQ history.QTransactions, sequence uint3
 
 func (p *TransactionProcessor) ProcessTransaction(transaction io.LedgerTransaction) error {
 	if err := p.batch.Add(transaction, p.sequence); err != nil {
+		// add breakpoint here
 		return errors.Wrap(err, "Error batch inserting transaction rows")
 	}
+	if err := p.batch.Exec(); err != nil {
+		// add breakpoint here
+		return errors.Wrap(err, "Error flushing transaction batch")
+	}
 
 	return nil
 }
 
 func (p *TransactionProcessor) Commit() error {
-	if err := p.batch.Exec(); err != nil {
-		return errors.Wrap(err, "Error flushing transaction batch")
-	}
 
 	return nil
 }
```


Finally, I ran `horizon --apply-migrations db reingest range 24338403 24338403` with a breakpoint set on the error handling path of the transaction ingestion code.

ire-and-curses:
Thank you for the detailed explanation of what's going on here, and also how you debugged it. Super valuable. 🙏

PR #2461, opened 2020-04-06
support/httpdecode: check for http.NoBody before calling DecodeJSON

howardtw:
### What

Add a check for `http.NoBody` before calling `DecodeJSON`.

### Why

`For server requests, the Request Body is always non-nil but will return EOF immediately when no body is present.`. We use `httptest.NewRequest` extensively throughout the codebase for testing, and that function returns a serve request with non-nil body. Without this change, use the request returned from `httptest.NewRequest` as a param of `Decode` or `DecodeJSON` will cause the `EOF` error.

### Known limitations

N/A

### DIFF
--- a/support/http/httpdecode/httpdecode_test.go
+++ b/support/http/httpdecode/httpdecode_test.go
@@ -327,7 +327,7 @@ func TestDecode_invalidJSON(t *testing.T) {
 	assert.Equal(t, "", bodyDecoded.FooName)
 }
 
-func TestDecode_emptyBody(t *testing.T) {
+func TestDecode_clientReqNoBody(t *testing.T) {
 	pathDecoded := struct {
 		Foo string `path:"foo"`
 	}{}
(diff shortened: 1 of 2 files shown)


support/http/httpdecode/httpdecode_test.go
```diff
@@ -327,7 +327,7 @@ func TestDecode_invalidJSON(t *testing.T) {
 	assert.Equal(t, "", bodyDecoded.FooName)
 }
 
-func TestDecode_emptyBody(t *testing.T) {
+func TestDecode_clientReqNoBody(t *testing.T) {
```
leighmcculloch, line 330: FYI when handling a request from an actual client it shouldn't be `nil`, it'll be `http.NoBody`. The only reason it is `nil` in this test is because the request is being constructed with a `nil` body without passing through the HTTP transport. If it went through the transport the body will be set to `NoBody`. I think this is one reason why we should use `httptest.NewRequest` in tests where the request isn't passing through the transport.

PR #2463, opened 2020-04-07
exp/recoverysigner: add GET /accounts/{address} endpoint

howardtw:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

This PR adds GET /accounts/{address} endpoint specified in the [spec](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0030.md#get-accountsaddress). This PR also adds `notPermitted` error for the case where a user is authenticated, but her credential does not match any of the identities in the requested account. 

### Why

This PR implements the missing endpoint.

### Known limitations

N/A

### DIFF
--- a/exp/services/recoverysigner/internal/serve/account_get.go
+++ b/exp/services/recoverysigner/internal/serve/account_get.go
@@ -70,7 +70,7 @@ func (h accountGetHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		resp.Identities = append(resp.Identities, respIdentity)
 	}
 	if !authorized {
-		notAuthorized.Render(w)
+		notFound.Render(w)
 		return
 	}
 	httpjson.Render(w, resp, httpjson.JSON)
--- a/exp/services/recoverysigner/internal/serve/account_get.go
+++ b/exp/services/recoverysigner/internal/serve/account_get.go
@@ -58,9 +58,9 @@ func (h accountGetHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 			Role: i.Role,
 		}
 		for _, m := range i.AuthMethods {
-			if (m.Type == account.AuthMethodTypeAddress && m.Value == claims.Address) ||
+			if m.Value != "" && ((m.Type == account.AuthMethodTypeAddress && m.Value == claims.Address) ||
 				(m.Type == account.AuthMethodTypePhoneNumber && m.Value == claims.PhoneNumber) ||
-				(m.Type == account.AuthMethodTypeEmail && m.Value == claims.Email) {
+				(m.Type == account.AuthMethodTypeEmail && m.Value == claims.Email)) {
 				respIdentity.Authenticated = true
 				authorized = true
 				break
--- a/exp/services/recoverysigner/internal/serve/account_get.go
+++ b/exp/services/recoverysigner/internal/serve/account_get.go
@@ -1,7 +1,9 @@
 package serve
 
 import (
+	"fmt"
 	"net/http"
+	"os"
 
 	"github.com/stellar/go/exp/services/recoverysigner/internal/account"
 	"github.com/stellar/go/exp/services/recoverysigner/internal/serve/auth"
@@ -33,6 +35,7 @@ func (h accountGetHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	req := accountGetRequest{}
 	err := httpdecode.Decode(r, &req)
 	if err != nil || req.Address == nil {
+		fmt.Fprintf(os.Stderr, "bad request: %v\n", err)
 		badRequest.Render(w)
 		return
 	}
@@ -52,7 +55,7 @@ func (h accountGetHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		Signer:  h.SigningAddress.Address(),
 	}
 
-	authenticated := false
+	authorized := false
 	for _, i := range acc.Identities {
 		respIdentity := accountResponseIdentity{
 			Role: i.Role,
@@ -62,15 +65,15 @@ func (h accountGetHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 				(m.Type == account.AuthMethodTypePhoneNumber && m.Value == claims.PhoneNumber) ||
 				(m.Type == account.AuthMethodTypeEmail && m.Value == claims.Email) {
 				respIdentity.Authenticated = true
-				authenticated = true
+				authorized = true
 				break
 			}
 		}
 
 		resp.Identities = append(resp.Identities, respIdentity)
 	}
-	if !authenticated {
-		unauthorized.Render(w)
+	if !authorized {
+		notPermitted.Render(w)
 		return
 	}
 	httpjson.Render(w, resp, httpjson.JSON)
--- a/exp/services/recoverysigner/internal/serve/account_get.go
+++ b/exp/services/recoverysigner/internal/serve/account_get.go
@@ -52,7 +52,7 @@ func (h accountGetHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		Signer:  h.SigningAddress.Address(),
 	}
 
-	authorized := false
+	authorizedIdentity := false
 	for _, i := range acc.Identities {
 		respIdentity := accountResponseIdentity{
 			Role: i.Role,
@@ -62,16 +62,17 @@ func (h accountGetHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 				(m.Type == account.AuthMethodTypePhoneNumber && m.Value == claims.PhoneNumber) ||
 				(m.Type == account.AuthMethodTypeEmail && m.Value == claims.Email)) {
 				respIdentity.Authenticated = true
-				authorized = true
+				authorizedIdentity = true
 				break
 			}
 		}
 
 		resp.Identities = append(resp.Identities, respIdentity)
 	}
-	if !authorized {
+	if claims.Address != req.Address.Address() && !authorizedIdentity {
 		notFound.Render(w)
 		return
 	}
+
 	httpjson.Render(w, resp, httpjson.JSON)
 }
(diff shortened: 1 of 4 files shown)


exp/services/recoverysigner/internal/serve/account_get.go
```diff
@@ -0,0 +1,80 @@
...
+				respIdentity.Authenticated = true
+				authorized = true
+				break
+			}
+		}
+
+		resp.Identities = append(resp.Identities, respIdentity)
+	}
+	if !authorized {
+		notAuthorized.Render(w)
```
leighmcculloch, line 76: It's important that in this case we return 404. It's important because we want to hide that the account is registered. RFC7231 Section 6.5.3 discusses this behavior:

>An origin server that wishes to "hide" the current existence of a forbidden target resource MAY instead respond with a status code of 404 (Not Found).

This is mentioned in [SEP-30's section about status code and the 404 error](https://github.com/stellar/stellar-protocol/blob/ffe1d418fdff3a9aa3b6bc7014323f67045ee04b/ecosystem/sep-0030.md#status-code) and is the behavior we use on other endpoints, like the `/accounts/<address>/sign` endpoint where we do this:
https://github.com/stellar/go/blob/ed641b3f3cc2ccdf3c88f09200cc61cab64b88d6/exp/services/recoverysigner/internal/serve/account_sign.go#L99-L101

Even if we were to change this, we should keep the endpoints consistent and introduce this change in authentication behavior consistently across all endpoints rather than just on one endpoint.

[RFC7231 Section 6.5.3]: https://tools.ietf.org/html/rfc7231#section-6.5.3

exp/services/recoverysigner/internal/serve/account_get.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/stellar/go/exp/services/recoverysigner/internal/serve/auth"
+	"github.com/stellar/go/keypair"
+	"github.com/stellar/go/support/http/httpdecode"
+	supportlog "github.com/stellar/go/support/log"
+	"github.com/stellar/go/support/render/httpjson"
+)
+
+type accountGetHandler struct {
+	Logger         *supportlog.Entry
+	SigningAddress *keypair.FromAddress
```
leighmcculloch, line 18: Love that you pass down just the FromAddress here and not the full key 👏.

exp/services/recoverysigner/internal/serve/account_get.go
```diff
@@ -0,0 +1,80 @@
...
+	claims, _ := auth.FromContext(ctx)
+	if claims.Address == "" && claims.PhoneNumber == "" && claims.Email == "" {
+		unauthorized.Render(w)
+		return
+	}
+
+	req := accountGetRequest{}
+	err := httpdecode.Decode(r, &req)
+	if err != nil || req.Address == nil {
+		fmt.Fprintf(os.Stderr, "bad request: %v\n", err)
```
leighmcculloch, line 38: Could you remove this before merging?

exp/services/recoverysigner/internal/serve/account_get.go
```diff
@@ -0,0 +1,80 @@
...
+		badRequest.Render(w)
+		return
+	}
+
+	acc, err := h.AccountStore.Get(req.Address.Address())
+	if err == account.ErrNotFound {
+		notFound.Render(w)
+		return
+	} else if err != nil {
+		h.Logger.Error(err)
```
leighmcculloch, line 48: Thanks for adding this, we need to add this in more places where we return a server error. 👏

exp/services/recoverysigner/internal/serve/account_get.go
```diff
@@ -0,0 +1,80 @@
...
+		Signer:  h.SigningAddress.Address(),
+	}
+
+	authorized := false
+	for _, i := range acc.Identities {
+		respIdentity := accountResponseIdentity{
+			Role: i.Role,
+		}
+		for _, m := range i.AuthMethods {
+			if (m.Type == account.AuthMethodTypeAddress && m.Value == claims.Address) ||
```
leighmcculloch, line 64: If we accidentally allowed an empty string value to get stored for `m.Value` somebodies account would be matched to this. I think it's important we do what we do over here and only check an address, phone number, or email if `claims` has a non-empty value for that field.

https://github.com/stellar/go/blob/ed641b3f3cc2ccdf3c88f09200cc61cab64b88d6/exp/services/recoverysigner/internal/serve/account_list.go#L89

exp/services/recoverysigner/internal/serve/account_get_test.go
```diff
@@ -0,0 +1,326 @@
...
+	r := httptest.NewRequest("GET", "/GDIXCQJ2W2N6TAS6AYW4LW2EBV7XNRUCLNHQB37FARDEWBQXRWP47Q6N", nil)
+	r = r.WithContext(ctx)
+
+	w := httptest.NewRecorder()
+	m := chi.NewMux()
+	m.Get("/{address}", h.ServeHTTP)
+	m.ServeHTTP(w, r)
+	resp := w.Result()
+
+	require.Equal(t, http.StatusForbidden, resp.StatusCode)
```
leighmcculloch, line 52: See above, this should be 404. It should be impossible to tell the difference between an account that doesn't exist and an account that a user is not authorized to access.

exp/services/recoverysigner/internal/serve/errors.go
```diff
@@ -30,6 +30,10 @@ var unauthorized = errorResponse{
 	Status: http.StatusUnauthorized,
 	Error:  "The request could not be authenticated.",
 }
+var notAuthorized = errorResponse{
+	Status: http.StatusForbidden,
+	Error:  "The request was not authorized to perform the action.",
+}
```
leighmcculloch, line 36: See above, I don't think there is any case we should return 403 from because we should hide from the user whether an account exists or not when they are not authorized to access it.

exp/services/recoverysigner/internal/serve/account_get_test.go
```diff
@@ -0,0 +1,326 @@
...
+	require.NoError(t, err)
+
+	wantBody := `{
+	"error": "The resource at the url requested was not found."
+}`
+
+	assert.JSONEq(t, wantBody, string(body))
+}
+
+func TestAccountGet_authenticatedByAddress(t *testing.T) {
```
leighmcculloch, line 151: This test tests whether a client can get an account when authenticated with an address that is registered as an authentication method, and it does a great job of that. Can you add a similar test that tests that a client can get an account when authenticated with the account's address, which shouldn't be included in the list of auth methods?

I also suspect the application code needs changing to support that test.

exp/services/recoverysigner/internal/serve/account_get.go
howardtw: OK I wasn't sure which one is more appropriate for being the non-authorized error, 403 or 404. I've seem systems that use either of them. I'll make it 404 since it's stated in SEP-30.

exp/services/recoverysigner/internal/serve/account_get.go
howardtw: https://github.com/stellar/go/pull/2463/commits/8af74d8576dae84e0c8dcdaa674f3ae9ecfb1f0a

exp/services/recoverysigner/internal/serve/account_get.go
howardtw: https://github.com/stellar/go/pull/2463/commits/83439687c52a29af628cc5f44a2aee7de55270ac

exp/services/recoverysigner/internal/serve/account_get_test.go
howardtw: https://github.com/stellar/go/pull/2463/commits/83439687c52a29af628cc5f44a2aee7de55270ac

exp/services/recoverysigner/internal/serve/errors.go
howardtw: https://github.com/stellar/go/pull/2463/commits/83439687c52a29af628cc5f44a2aee7de55270ac

exp/services/recoverysigner/internal/serve/account_get.go
howardtw: https://github.com/stellar/go/pull/2463/commits/6c13b435acff7e75b0a5346801a6fcfb5f3a3f62

exp/services/recoverysigner/internal/serve/account_get_test.go
howardtw: https://github.com/stellar/go/pull/2463/commits/20cd375489e73dd4e0ed549a4f9c9e45d4b38d8a

exp/services/recoverysigner/internal/serve/account_get.go
```diff
@@ -0,0 +1,78 @@
...
+				(m.Type == account.AuthMethodTypeEmail && m.Value == claims.Email)) {
+				respIdentity.Authenticated = true
+				authorizedIdentity = true
+				break
+			}
+		}
+
+		resp.Identities = append(resp.Identities, respIdentity)
+	}
+	if claims.Address != req.Address.Address() && !authorizedIdentity {
```
leighmcculloch, line 72: I think the way you had this before the last commit was easier to understand. You could keep the `authorized` and set it to `true` in the case that the claims address matches the requested account. I think that'd make this easier to read, I found it a bit confusing to read at first because the line to determine if the user is authenticated with the address was buried down the bottom of the handler.

e.g.:
```suggestion
	// Authorized if authenticated as the account.
	authorized := claims.Address == req.Address.Address()

	// Authorized if authenticated as an identity registered with the account.
	for _, i := range acc.Identities {
		respIdentity := accountResponseIdentity{
			Role: i.Role,
		}
		for _, m := range i.AuthMethods {
			if m.Value != "" && ((m.Type == account.AuthMethodTypeAddress && m.Value == claims.Address) ||
				(m.Type == account.AuthMethodTypePhoneNumber && m.Value == claims.PhoneNumber) ||
				(m.Type == account.AuthMethodTypeEmail && m.Value == claims.Email)) {
				respIdentity.Authenticated = true
				authorized = true
				break
			}
		}

		resp.Identities = append(resp.Identities, respIdentity)
	}
	
	// Return 404 not found if not authorized.
	if !authorized {
```

howardtw: This looks great! Will make the change.

PR #2465, opened 2020-04-07
exp/recoverysigner: separate authentication and authorization errors

howardtw:
### What

This PR renames the `unauthorized` error to `notAuthenticated`. `notAuthenticated` is used when there's nothing in the claim.

This PR adds the `notAuthorized` error. `notAuthorized` error is used when the credential does not match the resource of the requested action.

### Why

`The HTTP 403 Forbidden client error status response code indicates that the server understood the request but refuses to authorize it.`

We named the error `unauthorized` but the error message says `not authenticated`, which is confusing. I understand that http 401 error is named `NotAuthorized`, but it's used when having problems in authentication, not authorization. For authorization errors, 403 can be more appropriate. 

### Known limitations

N/A

### Notes

The new `notAuthorized` error is going to be used in https://github.com/stellar/go/pull/2463/

### DIFF
--- a/exp/services/recoverysigner/internal/serve/account_post.go
+++ b/exp/services/recoverysigner/internal/serve/account_post.go
@@ -75,7 +75,7 @@ func (h accountPostHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	claims, _ := auth.FromContext(ctx)
 	if claims.Address == "" {
-		unauthorized.Render(w)
+		notAuthenticated.Render(w)
 		return
 	}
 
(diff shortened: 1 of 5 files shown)


exp/services/recoverysigner/internal/serve/account_post.go
```diff
@@ -87,7 +87,7 @@ func (h accountPostHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if req.Address.Address() != claims.Address {
-		unauthorized.Render(w)
+		notAuthorized.Render(w)
```
leighmcculloch, line 90: This is maybe the only place a 403 would be appropriate to use in this application, as it wouldn't be exposing the presence of a resource, but I don't think the distinction is important here and it isn't required as 401 is still a valid status code when the credential doesn't permit an operation.

howardtw:
Closing because SEP-30 suggests that we use 404 instead of 403 when a request is not authorized.

Issue #2467, opened 2020-04-07
Update protocol 13 to include CAP 27 XDR


ire-and-curses:
A quick step but one needed before anything else. See #2365.

CAP 27 XDR can be found here, but may need manual merging with current stellar-core XDR: https://github.com/xdrpp/stellar-core/tree/dm/muxacct

PR #2471, opened 2020-04-08
all: Implement CAP 27 and SEP 23

### LINKED ISSUE #2365: Support Muxed Accounts (CAP 27)
The Stellar Core protocol 13 release will introduce a new type of account id called muxed accounts. Here is a summary of the changes (see https://github.com/stellar/stellar-protocol/blob/master/core/cap-0027.md for the full set of details):

The multiplexed account type is represented by a new XDR union:

~~~ {.c}
enum CryptoKeyType
{
    KEY_TYPE_ED25519 = 0,
    KEY_TYPE_MUXED_ED25519 = 256,  // This is new
    KEY_TYPE_PRE_AUTH_TX = 1,
    KEY_TYPE_HASH_X = 2
};

// Source or destination of a payment operation
union MuxedAccount switch (CryptoKeyType type) {
 case KEY_TYPE_ED25519:
     uint256 ed25519;
 case KEY_TYPE_MUXED_ED25519:
     struct {
         uint64 id;
         uint256 ed25519;
     } med25519;
};
~~~

The following fields, which were previously an `AccountID` or
`AccountID*`, are now a `MuxedAccount` or `MuxedAccount*`
(respectively):

* `PaymentOp::destination`
* `PathPaymentStrictReceiveOp::destination`
* `PathPaymentStrictSendOp::destination`
* `Operation::sourceAccount`
* `Operation::destination` (for `ACCOUNT_MERGE`)
* `Transaction::sourceAccount`
* `SimplePaymentResult::destination`

To support Muxed Accounts we will need to implement the following changes in Horizon and the SDKs:

- [x] Update the `strkey` package and the other Horizon SDK implementations to support muxed accounts.
- [x] add the restriction that all accounts in a valid SEP 10 challenge transaction must have type `KEY_TYPE_ED25519`, see https://github.com/stellar/stellar-protocol/issues/562
- [x] When ingesting transactions extract the underlying "G" address from a `MuxedAccount` and use that value when storing account columns in the Horizon database.
- [ ] Eventually we may need to add functionality to Horizon which distinguishes transactions involving MuxedAccounts from normal AccountIDs. Doing that may require us to change our data model. If so, we should either add a migration which will upgrade previously ingested `MuxedAccount` transactions to the new data model, or, reingest all transactions occurring since the start of protocol 13.

We should be able to implement these changes and release them before protocol 13 is deployed to stellar core. Also, the SDK changes to support the new type of StrKey should not be breaking for any of the clients.

2opremio:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

Fixes #2365 #2467  

### What

Implement [CAP 27](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0027.md) and [SEP 23](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0023.md)

This PR doesn't provide a full integration with CAP27 though:

* The ingestion code will simply transform `MuxedAccount`s into `AccountId`s
   (by dropping the memo ID). This means that queries will only show AccountId G-strkeys.
* The `POST transaction` endpoint will accept transactions using CAP27s
   XDR, so clients need to be adapted to that.

### Why

Because the Stellar protocol is everchanging.

### Known limitations

Query responses won't contain MuxedAccounts (M-prefixed strkeys) where expected, instead horizon will internally (during ingestion) transform MuxedAccounts into AccountIds (G-prefixed strkey) by dropping the the memo id. This may come as a surprise to end-users.


2opremio:
Guide to reviewers: I would mainly focus on the `xdr` and `strkey` packages.

xdr/muxed_account_test.go
```diff
@@ -0,0 +1,153 @@
...
+
+	It("returns false when the account ids have different values", func() {
+		var l, r MuxedAccount
+		err := l.SetAddress("GA7QYNF7SOWQ3GLR2BGMZEHXAVIRZA4KVWLTJJFC7MGXUA74P7UJVSGZ")
+		Expect(err).ShouldNot(HaveOccurred())
+		err = r.SetAddress("GCR22L3WS7TP72S4Z27YTO6JIQYDJK2KLS2TQNHK6Y7XYPA3AGT3X4FH")
+		Expect(err).ShouldNot(HaveOccurred())
+		Expect(l.Equals(r)).To(BeFalse())
+
+		err = l.SetAddress("MCAAAAAAAAAAAAB7BQ2L7E5NBWMXDUCMZSIPOBKRDSBYVLMXGSSKF6YNPIB7Y77ITKNOG")
```
tamirms, line 111: can you add a comment that these two muxed accounts have the same underlying account id but different memos and that is why they are not equal?

services/horizon/internal/actions_transaction_test.go
```diff
@@ -260,8 +292,51 @@ func TestTransactionActions_PostSuccessful(t *testing.T) {
...
+			},
+			Signatures: []xdr.DecoratedSignature{
+				{
+					Hint:      xdr.SignatureHint{174, 228, 190, 76},
+					Signature: xdr.Signature{73, 202, 13, 176, 216, 188, 169, 9, 141, 130, 180, 106, 187, 225, 22, 89, 254, 24, 173, 62, 236, 12, 186, 131, 70, 190, 214, 24, 209, 69, 233, 68, 1, 238, 48, 154, 55, 170, 53, 196, 96, 218, 110, 2, 159, 187, 120, 2, 50, 115, 2, 192, 208, 35, 72, 151, 106, 17, 155, 160, 147, 200, 52, 12},
+				},
+			},
+		},
+	}
+	//dmp := diffmatchpatch.New()
```
tamirms, line 329: should these comments be removed?

services/horizon/internal/expingest/processors/participants_processor.go
```diff
@@ -140,7 +140,7 @@ func participantsForTransaction(
 	transaction io.LedgerTransaction,
 ) ([]xdr.AccountId, error) {
 	participants := []xdr.AccountId{
-		transaction.Envelope.SourceAccount(),
+		transaction.Envelope.SourceAccount().ToAccountId(),
```
tamirms, line 143: can you add a test for the participants processor where the source account of a tx is a muxed account?

services/horizon/internal/expingest/processors/operations_processor.go
```diff
@@ -337,19 +337,19 @@ func operationFlagDetails(result map[string]interface{}, f int32, prefix string)
 // Participants returns the accounts taking part in the operation.
 func (operation *transactionOperationWrapper) Participants() ([]xdr.AccountId, error) {
 	participants := []xdr.AccountId{}
-	participants = append(participants, *operation.SourceAccount())
+	participants = append(participants, operation.SourceAccount().ToAccountId())
```
tamirms, line 340: can you add a test for the operations processor where the source and destination account of a tx are muxed accounts?

services/horizon/internal/expingest/processors/effects_processor.go
```diff
@@ -722,7 +722,7 @@ func ingestTradeEffects(effects *effectsWrapper, buyer xdr.AccountId, claims []x
 	}
 }
 
-func tradeDetails(buyer, seller xdr.AccountId, claim xdr.ClaimOfferAtom) (bd map[string]interface{}, sd map[string]interface{}) {
+func tradeDetails(buyer xdr.MuxedAccount, seller xdr.AccountId, claim xdr.ClaimOfferAtom) (bd map[string]interface{}, sd map[string]interface{}) {
```
tamirms, line 725: shouldn't buyer be a `xdr.AccountId`? otherwise, the "M..." address will be recorded in the db during ingestion

tamirms: if it should be changed to a `xdr.AccountId`, could you also add a test for the effects processor which has muxed accounts as buyers, sellers, and source accounts?

tamirms:
@2opremio I checked out your branch and searched for all cases where `func (m *MuxedAccount) Address() string` was invoked.

In the following cases I think we should be converting the `MuxedAccount` to an `AccountId` before calling `Address()` because we want to be ingesting "G..." addresses instead of "M..." addresses:


- https://github.com/stellar/go/blob/2467-cap27-update-xdr/services/horizon/internal/db2/history/transaction_batch_insert_builder.go#L157
- 
- https://github.com/stellar/go/blob/2467-cap27-update-xdr/services/horizon/internal/expingest/processors/effects_processor.go
- 
- https://github.com/stellar/go/blob/2467-cap27-update-xdr/services/horizon/internal/expingest/processors/operations_processor.go
- 
- https://github.com/stellar/go/blob/2467-cap27-update-xdr/services/horizon/internal/expingest/processors/trades_processor.go

I also found one case in txsub where we should be converting a `MuxedAccount` to an `AccountID` before calling `Address()`:

- https://github.com/stellar/go/blob/2467-cap27-update-xdr/services/horizon/internal/txsub/system.go#L89


In this case we do a db lookup by account address and since we don't store `M...' addresses in the database, the db lookup will not return any rows if a `M..." address is used as the lookup key.

--

This requirement from issue #2365 is not addressed by your PR:

-  add the restriction that all accounts in a valid SEP 10 challenge transaction must have type KEY_TYPE_ED25519, see stellar/stellar-protocol#562


To fix this, you should also update `ReadChallengeTx()` in https://github.com/stellar/go/blob/2467-cap27-update-xdr/txnbuild/transaction.go#L447

--

For all of these cases that you fixed can you also add tests?

services/horizon/internal/actions_transaction_test.go
2opremio: Yes, I forgot.

strkey/main.go
```diff
@@ -174,20 +181,53 @@ func checkValidVersionByte(version VersionByte) error {
...
+	// so, we use bytes from the beginning
+	srcBytes := []byte(src)
+	// The minimal binary decoded length is 3 bytes (version byte and 2-byte CRC) which,
+	// in unpadded base32 (since each character provides 5 bits) corresponds to ceiling(8*3/5) = 5
+	if len(srcBytes) < 5 {
+		return nil, errors.Errorf("strkey is %d bytes long; minimum valid length is 5", len(srcBytes))
+	}
+	// SEP23 enforces strkeys to be in canonical base32 representation.
+	// Go's decoder doesn't help us there, so we need to do it ourselves.
+	// 1. Make sure there are no unused leftover
```
2opremio, line 211:
```suggestion
	// 1. Make sure there is no full unused leftover byte at the end
```

strkey/main.go
```diff
@@ -174,20 +181,53 @@ func checkValidVersionByte(version VersionByte) error {
...
+	//   (i.e. there shouldn't be 5 or more leftover bits)
+	leftoverBits := (len(srcBytes) * 5) % 8
+	if leftoverBits >= 5 {
+		return nil, errors.New("non-canonical strkey; unused leftover character")
+	}
+	// 2. The leftover bits (4 at most), if any, should be set to 0
```
2opremio, line 217:
```suggestion
	// 2. In the last byte of the strkey there may be leftover bits (4 at most, otherwise it would be a full byte, which we have for checked above). If there are any leftover bits, they should be set to 0
```

2opremio:
@tamirms I think I have addressed everything, PTAL

build/util.go
```diff
@@ -19,6 +19,19 @@ func setAccountId(addressOrSeed string, aid *xdr.AccountId) error {
...
 }
 
+func setMuxedAccount(addressOrSeed string, m *xdr.MuxedAccount) error {
+	kp, err := keypair.Parse(addressOrSeed)
+	if err != nil {
+		return err
+	}
+
+	if m == nil {
+		return errors.New("aid is nil in setAccountId")
```
tamirms, line 29: `return errors.New("aid is nil in setMuxedAccount")`

txnbuild/transaction.go
```diff
@@ -449,6 +449,18 @@ func ReadChallengeTx(challengeTx, serverAccountID, network string) (tx Transacti
 	if err != nil {
 		return tx, clientAccountID, err
 	}
+
+	// Enforce no muxed accounts (at least until we understand their impact)
+	muxedAccountErr := errors.New("muxed accounts are not allowed in challenge transactions")
```
tamirms, line 454: could you add a test which exercises this failure mode?

2opremio: I tried, but I I don't think it makes sense, see https://github.com/stellar/go/pull/2471/commits/d68fe75cd0e30d45cf2fbe009fa1d0afa293745c

> I couldn't add tests because Transaction.Build() doesn't
allow using muxed accounts as input

tamirms:
@2opremio can you  reopen the pr and target it to the horizon 1.2.0 release branch. I meant to retarget your PR earlier but I forgot about it. Unfortunately, I cannot push your commit onto the release branch without going through a PR because it is write protected

2opremio:
I cannot reopen it, I will need to create a separate PR. I will do that.

PR #2474, opened 2020-04-09
all: Add protocol 13 support for Go SDK

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

* Merge horizon-v1.1.0 into horizon client branch (these include commits which update the XDR files to the protocol 13 definitions)
* Modify `txnbuild.Transaction` so that `Build()` generates a transaction V0 envelope which will be accepted by Stellar Core running protocol 12 or protocol 13
* Modify `txnbuild. TransactionFromXDR()` so that it is able to handle protocol 12 tx envelopes, protocol 13 v1 envelopes, and fee bump envelops
* `txnbuild. ReadChallengeTx()` parses both transaction V0 and transaction v1 envelopes. However, it rejects fee bump transaction envelopes.
* `MaxFee` in the horizon Transaction response is now an `int64`
* Add support for decoding horizon JSON responses which have `MaxFee` and `FeeCharged` as a json Number or a json String (we will eventually need to modify Horizon to encode `MaxFee` and `FeeCharged` as a strings since they can hold 64 bit values).

### Why

The problem to solve is how to get SDK clients out ahead of protocol 13, while still maintaining backwards compatibility. We have already released Horizon 1.1.0 which adds support for protocol 13 (the only thing missing is muxed accounts and that will be included in the forthcoming Horizon release). Our plan is to do two SDK releases alongside the Horizon release.

Before Stellar Core holds the protocol 13 vote we would like to release SDKs that can read protocol 12 or 13, but write 12. If they decode 13 they attempt to parse it back into 12. This should be transparent since there's no 13 in the wild. We can potentially include the additive CAPs (that just include new operations) - the new ops just don't do anything.

After the vote: We do the second release: SDKs that fully support (write and read) protocol 13 and encourage everyone to adopt it.

After the vote: The "first release" SDKs can now receive either protocol 12 or 13 XDR, depending on deployed clients. Parsing protocol 13 and shimming it to 12 fails if the XDR includes new operations, but for the majority of transactions that's not the case, since the new operations take time to be adopted and are less fundamental (than e.g. payments). So most existing things should just keep working as hopefully a wave of upgrades cross the ecosystem.

### Known limitations

* Still need to add the updated XDR for muxed accounts to this PR.
* I have not added any functions in `txnbuild` to create fee bump transactions. I want to keep make this release as backwards compatible as possible. The only breaking change in the release is that `MaxFee` in the horizon Transaction response is now an `int64`. This will make it really easy for users of the SDK to upgrade. Adding fee bump functionality to `txnbuild` will likely require significant breaking changes to the package. I'd like to ship those changes in a separate release.


main_test.go
```diff
@@ -131,8 +131,11 @@ func ExampleLowLevelTransaction() {
 	}
 
 	txe := xdr.TransactionEnvelope{
-		Tx:         tx,
-		Signatures: []xdr.DecoratedSignature{ds},
+		Type: xdr.EnvelopeTypeEnvelopeTypeTx,
+		V1: &xdr.TransactionV1Envelope{
```
tamirms, line 135: use V0 in example for now

PR #2480, opened 2020-04-13
exp/recoverysigner: add PUT /accounts/{address} endpoint

howardtw:
### What

This PR implements the PUT /accounts/{address} endpoint specified [here](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0030.md#put-accountsaddress).

Note that this endpoint replaces the original identities with the identities provided in the request, and not merged.

`db_store_update.go` and its test file are added to support this PR.

Though 996+, 1- seems intimidating, 777+ of it are for testing.

### Why

This PR adds the final endpoint in the spec that is not yet implemented.

### Known limitations

N/A

Close https://github.com/stellar/go/issues/2340

### DIFF
--- a/exp/services/recoverysigner/internal/account/db_store_update.go
+++ b/exp/services/recoverysigner/internal/account/db_store_update.go
@@ -13,6 +13,8 @@ func (s *DBStore) Update(a Account) error {
 	}
 
 	var accountID int64
+	// Delete an identity will delete the associated auth methods because of the ON DELETE CASCADE reference.
+	// https://github.com/stellar/go/blob/b3e0a353a901ce0babad5b4953330e55f2c674a1/exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200311000002-create-auth-methods.sql#L11
 	err = tx.Get(&accountID, `
 		WITH deleted_identities AS (
 			DELETE FROM identities
--- a/exp/services/recoverysigner/internal/account/db_store_update.go
+++ b/exp/services/recoverysigner/internal/account/db_store_update.go
@@ -1,0 +1,54 @@
+package account
+
+import (
+	"database/sql"
+
+	"github.com/lib/pq"
+)
+
+func (s *DBStore) Update(a Account) error {
+	tx, err := s.DB.Beginx()
+	if err != nil {
+		return err
+	}
+
+	var accountID int64
+	err = tx.Get(&accountID, `
+		WITH deleted_identities AS (
+			DELETE FROM identities
+			USING accounts
+			WHERE identities.account_id = accounts.id AND accounts.address = $1
+			RETURNING identities.account_id AS account_id
+		)
+		SELECT DISTINCT account_id FROM deleted_identities
+	`, a.Address)
+	if err == sql.ErrNoRows {
+		return ErrNotFound
+	}
+	if err != nil {
+		return err
+	}
+
+	for _, i := range a.Identities {
+		var authTypes, authValues pq.StringArray
+		for _, m := range i.AuthMethods {
+			authTypes = append(authTypes, string(m.Type))
+			authValues = append(authValues, m.Value)
+		}
+		_, err = tx.Exec(`
+			WITH new_identity AS (
+				INSERT INTO identities (account_id, role)
+				VALUES ($1, $2)
+				RETURNING account_id, id
+			)
+			INSERT INTO auth_methods (account_id, identity_id, type_, value)
+			SELECT account_id, id, unnest($3::auth_method_type[]), unnest($4::text[])
+			FROM new_identity
+		`, accountID, i.Role, authTypes, authValues)
+		if err != nil {
+			return err
+		}
+	}
+
+	return tx.Commit()
+}
--- a/exp/services/recoverysigner/internal/account/db_store_update_test.go
+++ b/exp/services/recoverysigner/internal/account/db_store_update_test.go
@@ -1,0 +1,300 @@
+package account
+
+import (
+	"testing"
+
+	"github.com/stellar/go/exp/services/recoverysigner/internal/db/dbtest"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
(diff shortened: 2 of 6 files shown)


exp/services/recoverysigner/internal/account/db_store_update.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	var accountID int64
+	// Delete an identity will delete the associated auth methods because of the ON DELETE CASCADE reference.
+	// https://github.com/stellar/go/blob/b3e0a353a901ce0babad5b4953330e55f2c674a1/exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200311000002-create-auth-methods.sql#L11
+	err = tx.Get(&accountID, `
+		WITH deleted_identities AS (
+			DELETE FROM identities
+			USING accounts
+			WHERE identities.account_id = accounts.id AND accounts.address = $1
```
leighmcculloch, line 22: What's the motivation for using `USING`? `USING` is a shorthand notation for use when the left and right side of a join are using the same field names which isn't the case here because the account ID field has a different name in the two tables (`account_id` and `id`). Instead of `USING` and having the join condition live in the `WHERE`, can you use a `JOIN` and specify the first condition in the `WHERE` in the `ON`? It's more common and easier to understand.

exp/services/recoverysigner/internal/account/db_store_update.go
```diff
@@ -0,0 +1,56 @@
...
+				RETURNING account_id, id
+			)
+			INSERT INTO auth_methods (account_id, identity_id, type_, value)
+			SELECT account_id, id, unnest($3::auth_method_type[]), unnest($4::text[])
+			FROM new_identity
+		`, accountID, i.Role, authTypes, authValues)
+		if err != nil {
+			return err
+		}
+	}
```
leighmcculloch, line 53: >Note that this endpoint replaces the original identities with the identities provided in the request, and not merged.

Could you elaborate why we'd want to do a hard replace at the database level and not merge? It means small changes like adding or changing a single authentication method will churn all the records and the audit tables won't capture a simple view of what is changing.

exp/services/recoverysigner/internal/account/db_store_update.go
howardtw: USING is stated in this postgres official [doc](https://www.postgresql.org/docs/current/sql-delete.html) for the delete statement and is not limited to the case where the left and right side have the same field names. But if you find it not easy to understand, we can use a join or a subquery `where account_id = (select id from accounts where address = $1)`

exp/services/recoverysigner/internal/account/db_store_update.go
howardtw: What do you mean? I implemented this endpoint according to what's in SEP-30.
```
The identities should be entirely replaced with the identities provided in the request, and not merged. Either owner or other or both should be set. If one is currently set and the request does not include it, it is removed.
```
Isn't that what we want?

exp/services/recoverysigner/internal/account/db_store_update.go
leighmcculloch: You're right, I was thinking about the intended use of USING in SELECT statements which is different. Thanks! 👏 Looks good.

howardtw:
> Overall looks good, but I think we should explore merging. While the API provides broad PUT replacement, under the hood we should store only the things that are changing if we can do that.

For PUT requests, it makes sense to entire replace the old identities if we don't want to use merge. And if we want to let user change a value associated with an auth method, we could add a PATH endpoint.

exp/services/recoverysigner/internal/account/db_store_update.go
leighmcculloch: SEP-30 is referring to the API and not how we store the data. The API request has to contain all the identities that the user expects to be registered with the account after the request is finished. But on the backend if the account has only phone, then a request comes in adding email, we should be able to store just the email without deleting 2 records and adding 3 records. We should be able to just add 1 record. The backend should be able to parse the `Account` passed in, compare it to the `Account` in the database, and build a set of add and delete operations to run?

howardtw: I see what you're saying.

> Could you elaborate why we'd want to do a hard replace at the database level and not merge?

It was straightforward to me that we want to update the identities associated with the account to become what's in the request. It was just about correctness.

In terms of performance, I'm actually not sure whether building a delta set will outperform the current method as it might involve some UPDATE statements, some DELETE statements, some INSERT statements. I think the answer will be "it depends". But I agree the audit table can look better with the delta approach.

Do you mind we leave the optimization in another PR? We can think about how we can build the delta set in the future as I think it has a lower priority at this moment.

leighmcculloch: We shouldn't need UPDATE statements, just INSERT and DELETE. We just need to determine what things to remove and what to add, basically like or-ing a set. I think we should implement this as part of this change because it will make the audit table difficult to understand. I think we should do this correct upfront.

howardtw: So changing a value associated with an auth methods will result in a DELETE statement and an INSERT statement, is that what you're saying?

> I think we should do this correct upfront.

I think this PR already makes the API correct. But sure, we can hold off merging this PR until the optimization is made.

leighmcculloch: We discussed this more offline, and it is possible, but maybe not worth it right not to have this code behave correctly. In the use cases and clients we are aware of right now we know that those clients are submitting a very small number of identities and in most cases entirely replacing one set of identities with an entirely different set of identities, so even if we change this logic to store the effectively changed rows we won't see an immediate benefit of that effort. So I think we should move forward with this as is and make a follow up issue. @howardtw Could you make that issue if you agree and will defer this improvement?

howardtw: https://github.com/stellar/go/issues/2500

PR #2481, opened 2020-04-13
exp/recoverysigner/account: add comments for the cascade delete

howardtw:
### What

This PR adds comments to explain the ON DETETE CASCADE effect when deleting an account to avoid confusion.

### Why

Make the code more readable.

### Known limitations

N/A

### DIFF
--- a/exp/services/recoverysigner/internal/account/db_store_delete.go
+++ b/exp/services/recoverysigner/internal/account/db_store_delete.go
@@ -2,8 +2,6 @@ package account
 
 func (s *DBStore) Delete(address string) error {
 	// Delete an account will delete the associated identities and auth methods because of the ON DELETE CASCADE references.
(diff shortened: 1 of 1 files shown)


exp/services/recoverysigner/internal/account/db_store_delete.go
```diff
@@ -1,6 +1,9 @@
 package account
 
 func (s *DBStore) Delete(address string) error {
+	// Delete an account will delete the associated identities and auth methods because of the ON DELETE CASCADE references.
+	// https://github.com/stellar/go/blob/b3e0a353a901ce0babad5b4953330e55f2c674a1/exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200311000001-create-identities.sql#L4
+	// https://github.com/stellar/go/blob/b3e0a353a901ce0babad5b4953330e55f2c674a1/exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200311000002-create-auth-methods.sql#L10
```
leighmcculloch, line 6: I think you can leave these out and just keep the comment to the first line. We don't have links at the SQL queries we have in the code to point to where the migrations live.

Issue #2458, opened 2020-04-06
Add memo text base64 field in Horizon transaction response 


tamirms:
As noted in https://github.com/stellar/go/issues/2022 , the memo value in a transaction can be an arbitrary binary blob even if the memo type is text. When horizon ingests transactions, it tries to store the memo field as a utf8 string even if the memo value is not valid utf8.

The Java SDK tries to recover the original byte representation of the memo by decoding the XDR. See https://github.com/stellar/java-stellar-sdk/pull/259 . However, we would like to avoiding XDR decoding in the SDKs when deserializing Horizon responses. The XDR definitions are subject to change so the SDKs should be relying on Horizon's API to avoid manually parsing XDR.

If we can include a `memo_bytes` field in the Horizon transaction resource, which is present alongside the `memo` field when `memo_type` is "text", then the Java SDK can use `memo_bytes` to recover the original memo value without having to parse `envelope_xdr`.

Once this feature is implemented, we should update the Java SDK to use the `memo_bytes`.

Issue #2022, opened 2019-12-06
XDR strings may not be valid utf8


tamirms:
XDR strings are defined as byte sequences where the first byte is the length of the string and the remaining bytes are the contents of the string.

https://docs.oracle.com/cd/E18752_01/html/816-1435/xdrproto-31244.html#xdrproto-38

We interpret XDR strings as unicode in horizon, the golang sdk, and the Java sdk. However, not all sequences of bytes are valid unicode strings.

This transaction https://horizon-testnet.stellar.org/transactions/213d52180f5e74fc4e4bc2c86740a44f98a40ab31e6718cc032e95f9084ae535 contains a memo field which is not a valid unicode string.

The problematic memo causes this issue in the Java SDK: https://github.com/stellar/java-stellar-sdk/issues/257

Also, the memo included in the horizon response does not match the memo you get when parsing the xdr envelope using the golang libraries:

```golang
func TestMemo(t *testing.T) {
	envelopeXDR := "AAAAAM6jLgjKjuXxWkir4M7v0NqoOfODXcFnn6AGlP+d4RxAAAAAZAAIiE4AAAABAAAAAAAAAAEAAAAcyKMl+WDSzuttWkF2DvzKAkkEqeSZ4cZihjGJEAAAAAEAAAAAAAAAAQAAAAAgECmBaDwiRPE1z2vAE36J+45toU/ZxdvpR38tc0HvmgAAAAAAAAAAAJiWgAAAAAAAAAABneEcQAAAAECeXDKebJoAbST1T2AbDBui9K0TbSM8sfbhXUAZ2ROAoCRs5cG1pRvY+ityyPWFEKPd7+3qEupavkAZ/+L7/28G"
	envelope := xdr.TransactionEnvelope{}
	err := xdr.SafeUnmarshalBase64(envelopeXDR, &envelope)
	if err != nil {
		t.Fatal(err)
	}

	memo, _ := envelope.Tx.Memo.GetText()
	response, err := http.DefaultClient.Get(
		"https://horizon-testnet.stellar.org/transactions/213d52180f5e74fc4e4bc2c86740a44f98a40ab31e6718cc032e95f9084ae535",
	)
	if err != nil {
		t.Fatal(err)
	}
	defer response.Body.Close()
	bodyBytes, err := ioutil.ReadAll(response.Body)
	if err != nil {
		t.Fatal(err)
	}
	tx := protocol.Transaction{}
	err = json.Unmarshal(bodyBytes, &tx)
[...]

PR #2485, opened 2020-04-14
protocols/horizon: Add memo_bytes field to transaction response

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Close https://github.com/stellar/go/issues/2458

Add `memo_bytes` field to the Horizon transaction response. `memo_bytes` contains the base64 encoding of the original byte sequence set in the transaction envelope's memo field.

### Why

As noted in #2022 , the memo value in a transaction can be an arbitrary binary blob even if the memo type is text. When horizon ingests transactions, it tries to store the memo field as a utf8 string even if the memo value is not valid utf8.

The Java SDK tries to recover the original byte representation of the memo by decoding the XDR. See stellar/java-stellar-sdk#259 . However, we would like to avoiding XDR decoding in the SDKs when deserializing Horizon responses. The XDR definitions are subject to change so the SDKs should be relying on Horizon's API to avoid manually parsing XDR.

If we can include a memo_bytes field in the Horizon transaction resource, which is present alongside the memo field when memo_type is "text", then the Java SDK can use memo_bytes to recover the original memo value without having to parse envelope_xdr.

### Known limitations

[N/A]


services/horizon/internal/docs/reference/resources/transaction.md
```diff
@@ -30,6 +30,7 @@ To learn more about the concept of transactions in the Stellar network, take a l
 | fee_meta_xdr            | string                   | A base64 encoded string of the raw `LedgerEntryChanges` xdr struct produced by taking fees for this transaction.               |
 | memo_type               | string                   |                                                                                                                                |
 | memo                    | string                   |                                                                                                                                |
```
ire-and-curses, line 32: Could you also add a comment to the original `memo` line, so that the reader can understand the difference between the two?

Issue #2468, opened 2020-04-07
Implement CAP 15 (fee bump) support in Go SDK


ire-and-curses:
This work needs to be done in two parts.

- In part 1, we release an SDK that sends protocol 12 transactions, and can read protocol 13 as far as possible.
- In part 2, we release an SDK that reads and writes protocol 13 transactions.

- [ ]  Update SDK so that it is compatible with the new horizon transaction response structure
- [ ] Update SDK to allow transaction lookup by either its hash or its inner hash (in the case the transaction is a fee bump)
- [ ] Consider what other features could make look ups for fee bumps ergonomic
- [ ] Update SDK to build transactions from XDR from both protocol 12 and protocol 13, as far as possible. This means shimming for part 1.
- [ ] Update SDK with functionality to create and submit fee bump transactions. Note that fee bump transactions have a [different fee rate calculation](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0015.md#fee-rate) compared to normal transactions.
- [ ] Consider SEP 10 implications of fee bump transactions. We probably want to add the restriction that a fee bump transaction cannot be a valid SEP 10 challenge transaction.

Parent issue: #2360

ire-and-curses:
@tamirms I made a first sketch summarising the work, please edit at will.

bartekn:
Presumably fixed by https://github.com/stellar/go/pull/2495. Can you confirm? If not, please reopen.

Issue #2439, opened 2020-04-02
txnbuild: move network passphrase field from Transaction object to Sign func


howardtw:
**Why**

Because both the Transaction and the TransactionEnvelope xdr structs don't contain a passphrase, one cannot get a `txnbuild.Transaction` object back with the network passphrase set when calling `txnbuild.TransactionFromXDR`. It is inconvenient given the reason that one has to set the network passphrase manually before calling the `Sign` function.

Removing the Network field out from `txnbuild.Transaction` makes the object consistent with the xdr form. Furthermore, passing it to the Sign func makes the clearer that it's a higher level of abstractions used by the SDKs and it is required before signing.

ire-and-curses:
Related: https://github.com/stellar/go/issues/1259#issuecomment-603493853

howardtw:
Closed by https://github.com/stellar/go/pull/2495

Issue #2411, opened 2020-03-24
txnbuild: calling Transaction.Build() on a default Transaction panics


leighmcculloch:
### What version are you using?

stellar/go@bab6abc8fbf3de9a932faa4c0da71b76d22f0932

### What did you do?

```go
tx := txnbuild.Transaction{}
err := tx.Build()
if err != nil {
	fmt.Println(err)
	return
}
fmt.Println(tx)
```

### What did you expect to see?

An error returned by the call to `Build()` indicating a required field isn't set.

### What did you see instead?

Panic.

```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x684017]

goroutine 19 [running]:
testing.tRunner.func1.1(0x72c540, 0xa2ec20)
        /home/leighmcculloch/local/bin/go/1.14/src/testing/testing.go:941 +0x3d0
testing.tRunner.func1(0xc0000dc5a0)
        /home/leighmcculloch/local/bin/go/1.14/src/testing/testing.go:944 +0x3f9
panic(0x72c540, 0xa2ec20)
        /home/leighmcculloch/local/bin/go/1.14/src/runtime/panic.go:967 +0x15d
github.com/stellar/go/txnbuild.(*Transaction).Build(0xc00002e5b0, 0x9f21f2, 0x9f21f2)
        /home/leighmcculloch/devel/stellar--go/txnbuild/transaction.go:111 +0x177
```

bartekn:
Presumably fixed by https://github.com/stellar/go/pull/2495.

Issue #1259, opened 2019-05-08
txnbuild: allow for custom sequence numbers


tomerweller:
The [`Transaction.Build()`](https://github.com/stellar/go/blob/47ee6efc1bb30236d040e0ea54037f180cc6f4f3/txnbuild/transaction.go#L86) function uses and consumes the current sequence number of supplied `Account`. This makes it tricky to build multiple transactions with the same seq number, or future seq numbers. These are crucial for writing smart contract functionality.

In addition, this build step has a side effect of mutating the seq number of the given account which is not obvious and conflates account state with the construction of a transaction.

tomerweller:
One possible solution: replace `Account` in the `Transaction` with a simple account strkey and seqnum. `Transaction.Build()` can use these without mutating any state. 

In addition, there can be a `BuildTx()` function on `Account` which takes in a transaction and builds with the account strkey, seqnum, and consumes that.

tomerweller:
Also, I know that current functionality mimics that of other sdks (js, for example) but I don't think we should allow that baggage set the tone.

ire-and-curses:
In the new Go SDK, I implemented it this way because

* I assume most users are following a workflow where they need to submit a transaction now, not writing smart contracts for the future. Auto-incrementing + passing an Account provides the simplest API for this main use case.
* This is the behaviour of the JS SDK [build()](https://stellar.github.io/js-stellar-sdk/TransactionBuilder.html#build).
* This is the behaviour of the Java SDK [build()](https://stellar.github.io/java-stellar-sdk/org/stellar/sdk/Transaction.Builder.html#build--).

This is also the same behaviour as the [Python SDK](https://github.com/StellarCN/py-stellar-base/blob/2ab25a46c51ae33058177baf93c104e3d9d057f5/stellar_base/transaction.py#L36), the [.NET SDK](https://github.com/elucidsoft/dotnet-stellar-sdk/blob/2b2f07ed8d16702f49d60833e9965e292e55a69c/stellar-dotnet-sdk/Transaction.cs#L295), the [Scala SDK](https://github.com/Synesso/scala-stellar-sdk/blob/master/src/main/paradox/transacting.md#sequence-number), and so on.

Universally the approach taken by all the SDKs I am aware of has been to allow power users to modify the sequence number manually, while defaulting to incrementing. This approach is summed up well in the Scala docs:

[...]

PR #2495, opened 2020-04-20
all: Update txnbuild to support fee bump transactions

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This PR refactors `txnbuild.Transaction`  to support features introduced in the upcoming protocol 13 release (in particular [CAP 15 ](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0015.md) and [CAP 18](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0015.md)). The new API for `txnbuild` is not backwards compatible which is why the version for this release is a major version bump from the last horizon client release.

Users of `txnbuild` and `horizonclient` should upgrade to v3.0.0 before protocol 13 is enabled.  Pre v3 clients will still be able to submit transactions to a protocol 13 enabled stellar network. However, pre v3 clients will not able to parse xdr strings corresponding to protocol 13 transactions.

It is safe to upgrade to v3.0.0 before protocol 13 is enabled because we have configured `txnbuild` to construct transactions which are compatible with both protocol 12 and protocol 13. After protocol 13 is enabled, we will create a v3.1.0 release of horizon client which configures `txnbuild` to construct transactions which are only compatible with protocol 13. 

Fixes https://github.com/stellar/go/issues/2468 , https://github.com/stellar/go/issues/2439 , https://github.com/stellar/go/issues/1259 , https://github.com/stellar/go/issues/2411

### Known limitations

All services in the go monorepo will need to use the new `txnbuild` interface. But I have not updated the code in the monorepo including tests to use the new `Transaction` implementation. I would like feedback on the interface before I proceed with updating all the usages of `txnbuild`.

---

I wanted to make `Transaction` immutable which is why signing a `Transaction` instance returns a new `Transaction` instance with the additional signatures. However, I realized if I want to make `Transaction` completely immutable, I'd need to create copies when returning from `Operations()` and `Signatures()`. Otherwise, if I returned the slice bound to the `Transaction` struct, the caller could modify the contents of the slice and it would also modify the `Transaction` instance.

To be honest, I'm not sure if this is the right approach considering the overhead. Also, there's another issue which is that individual signatures and operations are mutable. So, even if you don't modify the slice you can still end up mutating the `Transaction` instance by modifying a specific signature or operation contained within the slice. To make `Transaction` fully immutable I'd need to create copies of each individual operation (and similarly for the signatures).

Here are the possible options:

1. only enforce that signing be immutable but do not return new copies of slices in `Operations()` and `Signatures()`. Also, document that the slices returned by `Operations()` and `Signatures()` should not be modified.
2. keep implementation as is
3. extend immutability guarantee by cloning each signature and operation when duplicating operations and signatures in `Operations()` and `Signatures()`.

---

The current interface is a best effort attempt at merging normal transactions and fee bump transactions into one type. A more accurate representation of the normal vs fee bump transaction dichotomy would require [sum types](https://en.wikipedia.org/wiki/Algebraic_data_type).  Although sum types are not natively supported in Go, I have tried to implement an alternative transaction interface which resembles a sum type here:

https://github.com/tamirms/go/commit/d2666f52fb5e888c3396b857e79ba7935a545a1e 

Which of the two versions do you prefer?


txnbuild/transaction.go
```diff
@@ -26,38 +27,195 @@ import (
...
+
+func (t *Transaction) IsFeeBump() bool {
+	return t.isFeeBump
+}
+
+func (t *Transaction) FeeSource() string {
+	return t.feeSource
+}
+
+func (t *Transaction) InnerTransaction() *Transaction {
```
2opremio, line 71: Can you add doc strings for the new methods and, in particular for methods which only makes sense when `IsFeeBump()` is `true`?

2opremio: Note that this only makes sense once a call is made with respect to what API to use.

tamirms: `FeeSource()` and `InnerTransaction()` are specific to fee bump transactions

2opremio: Yes, understood

txnbuild/transaction.go
```diff
@@ -26,38 +27,195 @@ import (
...
+	memo          Memo
+	timebounds    Timebounds
+	signatures    []xdr.DecoratedSignature
+}
+
+func (t *Transaction) BaseFee() int64 {
+	return t.baseFee
+}
+
+func (t *Transaction) TotalFee() int64 {
```
bartekn, line 59: Don't have a strong opinion but maybe it's worth keeping the name consistent with Horizon.
```suggestion
func (t *Transaction) MaxFee() int64 {
```

txnbuild/transaction.go
```diff
@@ -26,38 +27,195 @@ import (
...
+	}
+
+	return t.Sign(network, signers...)
+}
+
+// SignHashX signs a transaction with HashX signature type.
+// See description here: https://www.stellar.org/developers/guides/concepts/multi-sig.html#hashx.
+func (t *Transaction) SignHashX(preimage []byte) (*Transaction, error) {
+	if len(preimage) > xdr.Signature(preimage).XDRMaxSize() {
+		return t, errors.New("preimage cannnot be more than 64 bytes")
```
bartekn, line 165:
```suggestion
		return t, errors.Newf("preimage cannnot be more than %d bytes", xdr.Signature(preimage).XDRMaxSize())
```

tamirms:
>  In my opinion all of the other methods are not necessary because the values are defined in TransactionConfig, ie. I don't think anyone will call them after creating a new Transaction using TransactionConfig, maybe except sequence number.

There are two use cases for the `txnbuild` package: constructing new transactions 
 (e.g. `NewTransaction()` and `NewFeeBumpTransaction()`) and parsing transactions from xdr strings (e.g. `TransactionFromXDR()`).

I think removing all the accessors from `Transaction` and only allowing signing is a decent idea. But, then the type returned by `TransactionFromXDR()` cannot be `Transaction`. Returning a `TransactionConfig` from `TransactionFromXDR()` isn't what we want either because a transaction envelope may contain signatures and those values would not be present in a `TransactionConfig` instance.

2opremio:
> It looks good to me, I think it's worth checking with heavy users of txnbuild.

I agree. This approach look fine to me, however I don't like that non-fee-bump-transactions are polluted with methods only applicable to fee-bump transactions (the other alternative is better in this regard). I think that the type system should guide the user on what can be called and what not. This would make the methods self-explanatory and safer (it won't allow calling feebump methods on non-feebump transactions. 

Here is a proposal:

```go
type Transaction {
// ...
}

func (t *Transaction) SourceAccount() SimpleAccount {
	return t.sourceAccount
}

type FeeBump struct {
  tx *Transaction // internally, feebump is identical to a transaction, but they user doesn't know
}

func (t *Transaction) FeeBump() (bool, *FeeBump) {
	if t.isFeeBump {
           return FeeBump{t}
        }
        return nil
}

func (t *FeeBump) InnerTransaction() *Transaction {
 // By construction you know it's a feebump transaction, no need to check
}

// ...

```

As it stands right now:

|   | Treats transactions uniformly  | Doesn't pollute non-FB-transactions  |  
|---|---|---|
| This PR  |  ✓ | ✕  |   
| Alternative, sumtype approach  | ✕  | ✓  |    

I think we should go with what's more usable, favoring the most common usage pattern.

In a nutshell:

* If using the feebump methods are an exception, I would go for the approachin this PR (although I think we can get a bit more help from the type system, as I suggest above).
* If using the feebump methods is going to be common, then I would go for the alternative approach, since the drawback of  feebump approach.


I think that with the change I suggest, you would get the best of both worlds:

|   | Treats transactions uniformly  | Doesn't pollute non-FB-transactions  |  
|---|---|---|
| Adding a FeeBump type |  ✓ | ✓   |

tamirms:
A fee bump transaction wraps a normal transaction but provides a new max fee which is paid for by the fee account. Also, the fee bump transaction has a separate set of signatures from the fee account.

In @2opremio's approach when `TransactionFromXDR()` is called on a fee bump xdr string, the return value is a `Transaction` instance which corresponds to the inner transaction of the fee bump transaction.

* `Memo()` will return the memo of the inner transaction
* `Operations()` will return the operations of the inner transaction
* `SourceAccount()` will return the source account of the inner transaction

Here is the crucial difference between @2opremio's approach and the prototype in this PR. In @2opremio's approach:

* `Signatures()` will return the the signatures of the **inner transaction**
* `TotalFee()` / `BaseFee()` will return the the fee of the **inner transaction**

To obtain the fee, signatures, or fee account from the **fee bump transaction**, you need to call `FeeBump()` which will return a different type that has all those values.

In the code committed to this PR:

* `Signatures()` will return the the signatures of the **fee bump transaction**
* `TotalFee()` / `BaseFee()` will return the the fee of the **fee bump transaction**

To obtain the fee or signatures from the **inner transaction**, you need to call `InnerTransaction()` which will return a transaction instance that has all those values.

I am convinced @2opremio's approach is better because it ensures that `Transaction` will only return values from the inner transaction and that `FeeBumpTransaction` will only return values from the fee bump part of the transaction. In contrast, the `Transaction` from this PR will mostly return values from the v1 transaction envelope but if you call `Signatures()`, `TotalFee()`, or `BaseFee()` it will return values from the fee bump part.

Now I will analyze the tradeoffs between the [sum type implementation](https://github.com/tamirms/go/commit/d2666f52fb5e888c3396b857e79ba7935a545a1e) and @2opremio 's approach

tamirms:
> Another idea to simplify the Transaction interface is by adding a method to generate a TransactionConfig from an immutable Transaction, for cases where the Transaction needs to be modified. Then you mess with that (add signatures, modify seqnum etc) and build to get a new Transaction.

I think I'll punt on that for now because both the Javascript and Java SDK use the builder pattern for constructing transactions but none of those implementations support creating a new builder instance from an existing transaction. Also, I haven't been able to find examples where this functionality would be useful.

tamirms:
I've looked at all the usages of `txnbuild` in the go monorepo and other repos within stellar org (e.g. kelp). I also did a search for public Go code on github which contains the string `txnbuild`. Note that there are basically two ways you can use `txnbuild`:

 1. create new transactions 
 2. parse xdr strings by calling `TransactionFromXDR()` and examine them as `Transaction` instances.

As you would expect, almost all uses of  `txnbuild` are for constructing transactions. Both @2opremio 's proposal and the [sum type implementation](https://github.com/tamirms/go/commit/d2666f52fb5e888c3396b857e79ba7935a545a1e) are equally streamlined when it comes to building new transactions.

So, I decided to focus on how `TransactionFromXDR()` is used. In most of the cases I found `TransactionFromXDR()` was used to implement multisig signing. The common workflow was that the application received an xdr string and then it parsed  it using `TransactionFromXDR()`. Afterwards, the application would add another signature before submitting it to the network or outputting the transaction back to the end user.

The other uses of `TransactionFromXDR()` I found were to implement [SEP 30](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0030.md#post-accountsaddresssign) and [SEP 10](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0010.md#token).

When determining which proposal to choose I think it's important to look at specifically if there is a difference between how fee bump transactions are handled compared to normal non fee bump transactions.

In the case of SEP 10, my assumption is that we want to reject tokens which are fee bump transactions because we want to remove ambiguity on how we interpret challenge tokens. This is the reasoning behind why SEP 10 is not allowing muxed accounts (see https://github.com/stellar/stellar-protocol/issues/562 ).

I also believe that we want to reject [requests](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0030.md#post-accountsaddresssign) to the recovery signer service which ask to sign a fee bump transaction. According to SEP 30 the use case for the sign endpoint is to authorize adding a new signer or removing an existing signer from an account:

> it is anticipated for the use cases discussed earlier in this protocol that the transaction will contain an operation to add a new signer to the account and possibly to remove an old signer or to merge an account.

With regards to the general purpose multisig use case, I can anticipate that a multisig implementation may want to sign a fee bump transaction thereby allowing a multisig wallet to be the fee source for a someone else's transaction. But, I also think that it depends a lot on how the wallet will be used. There may be cases where the wallet will only sign off on a very limited subset of transactions (e.g. the signer recovery use case).

I am now leaning towards using the [sum type implementation](https://github.com/tamirms/go/commit/d2666f52fb5e888c3396b857e79ba7935a545a1e) because, in that implementation, a `Transaction` must be unpacked into either `SimpleTransaction` or a `FeeBumpTransaction`. The caller will be forced to consider how it should handle fee bump transactions. With @2opremio 's proposal the onus is on the caller to be proactive about checking the type of a transaction. I think we should treat signing transactions with precaution so that is why I favor the sum type implementation.

leighmcculloch:
I think the advantage the sum type implementation provides that the developer is forced to be aware of the transaction type is good, but I also think that general code I write to sign a transaction cares less about the contents of the transaction and more about just signing the transaction. If I want to be able to sign any transaction I have to write code knowing about both types of transactions.

I'm also a little concerned with the term `SimpleTransaction`. That looks like a new concept we're introducing that isn't rooted in the XDR. Will other SDKs call a non-fee-bump txn a "simple transaction" too? This may be another point of vocab disparity in different SDKs and implementations.

I'm struggling to imagine what using these different approaches would be like from the users perspective looking at just the implementations. Do we have any example code showing what the experience will be like for the user? I think we should have that drive our decision.

tamirms:
@leighmcculloch 

> I think the advantage the sum type implementation provides that the developer is forced to be aware of the transaction type is good, but I also think that general code I write to sign a transaction cares less about the contents of the transaction and more about just signing the transaction.

yes, there is the tradeoff. To evaluate the tradeoff I suggest you consider the following questions:
* Given some xdr string do you want to examine the contents of the transaction before signing off on it?
* How often would you want to sign the transaction without looking at the contents of the transaction vs first validating the transaction and then adding your signature?
* If you care about validating the transaction, unpacking the transaction type is helpful. On the other hand, if you don't care about validating the transaction, unpacking the transaction type is an unnecessary speed bump. Is it ok to add a speed bump to the later case to prioritize the workflow of validating and then signing?

> If I want to be able to sign any transaction I have to write code knowing about both types of transactions, and if I'm going to do that I may as well use the xdr package directly?

If you want to use the xdr package without txnbuild to sign a transacton you'd still need to know how to sign a transaction (e.g. which parts do you need to hash?). But, apart from that detail, you are correct.

> Will other SDKs call a non-fee-bump txn a "simple transaction" too?

In the Javascript SDK the equivalent of `TransactionFromXDR()` returns a `Transaction | FeeBumpTransaction` type. @abuiles please correct me if I'm wrong.

But to your point about the names, I am definitely open to considering different names for the two transaction types. I will try to come up with better names.

> Do we have any example code showing what the experience will be like for the user?

Did you see that `ReadChallengeTx()` was updated to work with the new transaction types in the commit? https://github.com/tamirms/go/commit/d2666f52fb5e888c3396b857e79ba7935a545a1e#diff-f6a8a4abae844d51cc40cb628ba55d71R757

I'm currently working on fixing all the references to the txnbuild package within the monorepo (including tests). Here's what it looks like to build a transaction:

```go
func TestVerifyChallengeTxThreshold_invalidNoSigners(t *testing.T) {
	serverKP := newKeypair0()
	clientKP1 := newKeypair1()
	clientKP2 := newKeypair2()
	clientKP3 := keypair.MustRandom()
	txSource := NewSimpleAccount(serverKP.Address(), -1)
	opSource := NewSimpleAccount(clientKP1.Address(), 0)
	op := ManageData{
		SourceAccount: &opSource,
		Name:          "testserver auth",
		Value:         []byte(base64.StdEncoding.EncodeToString(make([]byte, 48))),
	}
	threshold := Threshold(10)
	signerSummary := SignerSummary{}

	tx64, err := NewSignedSimpleTransaction(
		TransactionParams{
			SourceAccount:        &txSource,
			IncrementSequenceNum: false,
			Operations:           []Operation{&op},
			BaseFee:              100,
			Timebounds:           NewTimeout(1000),
		},
		network.TestNetworkPassphrase,
		serverKP, clientKP1, clientKP2, clientKP3,
	)
	assert.NoError(t, err)
	
	_, err = VerifyChallengeTxThreshold(tx64, serverKP.Address(), network.TestNetworkPassphrase, threshold, signerSummary)
	assert.EqualError(t, err, "no verifiable signers provided, at least one G... address must be provided")
}
```

This is the workflow that is most negatively affected by the refactor:

```golang
func addMysignature(tx64 string, kp *keypair.Full) (string, error) {
	parsed, err := TransactionFromXDR(tx64)
	if err != nil {
		return "", err
	}

	if tx, ok := parsed.FeeBump(); ok {
		tx, err = tx.Sign(network.TestNetworkPassphrase, kp)
		if err != nil {
			return "", err
		}
		return tx.Base64()
	} else if tx, ok := parsed.Simple(); ok {
		tx, err = tx.Sign(network.TestNetworkPassphrase, kp)
		if err != nil {
			return "", err
		}
		return tx.Base64()
	} else {
		return "", fmt.Errorf("invalid transaction")
	}
}
```

leighmcculloch:
@tamirms Thanks, those examples were 💯 really helpful. I think the sum types is a compelling solution even in the inconvenience of signing arbitrary transactions. I agree we might want to add signing functions to that top-level, but I don't think they'll be used by everyone and will expand the surface area of the API with little gain. It's a nice solution 👍.

>But to your point about the names, I am definitely open to considering different names for the two transaction types. I will try to come up with better names.

I'm thinking about better names for this too.

>NewSignedSimpleTransaction

I think we should reconsider adding functions like this that are just sugar. They cause code someone uses to build some types of transactions in some specific use cases to be very different to other use cases. If someone writes a lot of code using this function then needs the transaction hash they either need to reparse the transaction to get the hash, or they have to rewrite the code. I regret that we have the SEP-10 functions that build transactions and return a base64 encoded string, I'm dealing with that right now in the webauth app. Ideally everywhere this SDK is used has building a transaction in the exact same way everytime.

tamirms:
@stellar/horizon-committers @stellar/vega-backend 

I still need to add tests which exercise the fee bump code paths. But aside from that, the PR is ready for review. Please take a look, thanks!

clients/horizonclient/client.go
```diff
@@ -448,27 +448,29 @@ func (c *Client) SubmitTransactionXDR(transactionXdr string) (txSuccess hProtoco
...
+	return c.SubmitTransactionXDR(txeBase64)
 }
 
-// SubmitTransactionWithOptions submits a transaction to the network, allowing
+// SubmitTransaction submits a transaction to the network, allowing
 // you to pass SubmitTxOpts. err can be either error object or horizon.Error object.
 //
 // See https://www.stellar.org/developers/horizon/reference/endpoints/transactions-create.html
-func (c *Client) SubmitTransactionWithOptions(transaction txnbuild.Transaction, opts SubmitTxOpts) (txSuccess hProtocol.TransactionSuccess, err error) {
+func (c *Client) SubmitTransaction(transaction *txnbuild.Transaction, opts SubmitTxOpts) (txSuccess hProtocol.TransactionSuccess, err error) {
```
leighmcculloch, line 470: What's the motivation for removing the `SubmitTransactions` without options function? I'd expect the average dev to use the default options but by surfacing this it encourages users to specify options. It's compelling I think to keep the `SubmitTransaction` function without options to encourage use of defaults.

clients/horizonclient/examples_test.go
```diff
@@ -1078,25 +1078,26 @@ func ExampleClient_SubmitTransaction() {
...
-	err = tx.Build()
+	tx, err := txnbuild.NewTransaction(
+		txnbuild.TransactionParams{
+			SourceAccount:        &sourceAccount,
+			IncrementSequenceNum: false,
+			Operations:           []txnbuild.Operation{&op},
+			BaseFee:              txnbuild.MinBaseFee,
+			Timebounds:           txnbuild.NewInfiniteTimeout(), // Use a real timeout in production!
+		},
+	)
```
leighmcculloch, line 1089: What benefits should we expect to see with putting the build logic in `NewTransaction` rather than keeping it in the `Build` function? It's not obvious to me what the benefit is. NewTransaction looks good, but I don't see it as being more ergonomic than what we have now.

As a bonus if we keep the `Build` function, and rename `TransactionParams` to `TransactionBuilder` the API for building will also look and behave mostly like the builder pattern which is quickly recognized by many folks and is the pattern we're using in other SDKs (e.g. Javascript SDK). Are we thinking of replacing the builder pattern in other SDKs with something like this?

```suggestion
	txb := txnbuild.TransactionBuilder{
		SourceAccount:        &sourceAccount,
		IncrementSequenceNum: false,
		Operations:           []txnbuild.Operation{&op},
		BaseFee:              txnbuild.MinBaseFee,
		Timebounds:           txnbuild.NewInfiniteTimeout(), // Use a real timeout in production!
	}
	tx, err := txb.Build()
```

_I thought I had raised this previously but I can't find my previous comment about this. Apologies if I'm re-raising something that's been discussed already._

txnbuild/transaction.go
```diff
@@ -26,197 +27,645 @@ import (
...
 	}
 
-	// Check and set the timebounds
-	err = tx.Timebounds.Validate()
+	sourceAccount := xdrEnv.SourceAccount()
+
+	totalFee := int64(xdrEnv.Fee())
+	baseFee := totalFee
+	if count := int64(len(xdrEnv.Operations())); count > 0 {
+		baseFee = baseFee / count
```
ire-and-curses, line 465: in the case of fee bump, does this account for the extra fee for the fee bump itself? From CAP 15
>A fee-bump transaction has an effective number of operations equal to one plus the number of operations in the inner transaction. Correspondingly, the minimum fee for the fee-bump transaction is one base fee more than the minimum fee for the inner transaction. Similarly, the fee rate (see CAP-0005) is normalized by one plus the number of operations in the inner transaction rather than the number of operations in the inner transaction alone.

txnbuild/transaction.go
```diff
@@ -26,197 +27,645 @@ import (
...
+	return tx, nil
 }
 
-// BuildSignEncode performs all the steps to produce a final transaction suitable
-// for submitting to the network.
-func (tx *Transaction) BuildSignEncode(keypairs ...*keypair.Full) (string, error) {
-	err := tx.Build()
-	if err != nil {
-		return "", errors.Wrap(err, "couldn't build transaction")
+// FeeBumTransactionParams is a container for parameters
```
ire-and-curses, line 609: 😂 
```suggestion
// FeeBumpTransactionParams is a container for parameters
```

txnbuild/transaction.go
```diff
@@ -26,197 +27,645 @@ import (
...
+// which are used to construct new FeeBumpTransaction instances
+type FeeBumTransactionParams struct {
```
ire-and-curses, line 611:
```suggestion
type FeeBumpTransactionParams struct {
```

txnbuild/transaction.go
```diff
@@ -247,10 +696,7 @@ func BuildChallengeTx(serverSignerSecret, clientAccountID, anchorName, network s
 	// represent server signing account as SimpleAccount
 	sa := SimpleAccount{
 		AccountID: serverKP.Address(),
-		// Action needed in release: v2.0.0
-		// TODO: remove this and use "Sequence: 0" and build transaction with optional argument
-		//  (https://github.com/stellar/go/issues/1259)
-		Sequence: int64(-1),
+		Sequence:  0,
```
ire-and-curses, line 699: :tada:

txnbuild/transaction.go
```diff
@@ -26,197 +27,645 @@ import (
...
+		extended = append(extended, sig)
+	}
+	return extended, nil
 }
 
-// MarshalBinary returns the binary XDR representation of the transaction envelope.
-func (tx *Transaction) MarshalBinary() ([]byte, error) {
+func stringsToKP(
+	keys ...string,
+) ([]*keypair.Full, error) {
```
leighmcculloch, line 80: This form of multiline formatting seems unnecessary here and elsewhere. When it is on one line it is still short. What's the goal with adopting that pattern here?
```suggestion
func stringsToKP(keys ...string) ([]*keypair.Full, error) {
```

txnbuild/transaction.go
```diff
@@ -26,197 +27,645 @@ import (
...
+		return clone, errors.Wrap(err, "could not unmarshall envelope")
 	}
+	return clone, nil
+}
+
+// Transaction represents a Stellar transaction. See
+// https://www.stellar.org/developers/guides/concepts/transactions.html
+// A Transaction may be wrapped by a FeeBumpTransaction in which case
+// the account authorizing the FeeBumpTransaction will pay for the transaction fees
+// instead of the Transaction's source account.
```
leighmcculloch, line 180: >instead of

TIL that the fee bump entirely replaces the inner transactions fee payment. I thought it supplemented it. It's unclear to me from reading CAP-15 which it is now, but I assumed by 'bump' it was referring to bumping up the fee.

txnbuild/transaction.go
```diff
@@ -260,191 +706,51 @@ func BuildChallengeTx(serverSignerSecret, clientAccountID, anchorName, network s
...
-		Network:    network,
-		BaseFee:    uint32(100),
-	}
-
-	txeB64, err := tx.BuildSignEncode(serverKP.(*keypair.Full))
+	)
 	if err != nil {
 		return "", err
 	}
-	return txeB64, nil
```
leighmcculloch, line 285: @ire-and-curses If we're making a major version release that contains breaking changes I think we should also change `BuildChallengeTx` so that it returns the new immutable Transaction type. The caller can still call `.Base64()` on it to get the string, and it allows the user to get the hash of the transaction before signing it. Doing this would also simplify the function arguments by taking the network passphrase out. I can make this change but I'd need to make it after @tamirms's changes but before the release.

txnbuild/transaction.go
tamirms: yeah, the fee bump case is handled a few lines above inside the `if xdrEnv.IsFeeBump()` statement:
```golang
		newTx.feeBump = &FeeBumpTransaction{
			envelope:   xdrEnv,
			baseFee:    xdrEnv.FeeBumpFee() / int64(len(innerTx.simple.operations)+1),
			maxFee:     xdrEnv.FeeBumpFee(),
			inner:      innerTx.simple,
			feeAccount: feeBumpAccount.Address(),
			signatures: xdrEnv.Signatures(),
		}
```

I'll add a comment which explains the accounting of `xdrEnv.FeeBumpFee() / int64(len(innerTx.simple.operations)+1)` for fee bump transactions

protocols/horizon/main.go
```diff
@@ -88,27 +88,27 @@ func (a Account) GetCreditBalance(code string, issuer string) string {
...
 
-	return xdr.SequenceNumber(seqNum), nil
+	return seqNum, nil
 }
 
 // IncrementSequenceNumber increments the internal record of the account's sequence
 // number by 1. This is typically used after a transaction build so that the next
 // transaction to be built will be valid.
-func (a *Account) IncrementSequenceNumber() (xdr.SequenceNumber, error) {
+func (a *Account) IncrementSequenceNumber() (int64, error) {
```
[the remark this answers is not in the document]
tamirms: yeah, it is a breaking change. I figured since we're doing a major version bump we might as well make this change. it seems like `txnbuild` insulates users from xdr types. but the account interface seems to be an exception when it returns `xdr.SequenceNumber`.

clients/horizonclient/examples_test.go
tamirms: I prefer using `NewTransaction()` over having `TransactionBuilder` because if you don't care about keeping the builder around the code looks better with `NewTransaction()`
```golang
	tx, err := txnbuild.NewTransaction(
		txnbuild.TransactionParams{
			SourceAccount:        &sourceAccount,
			IncrementSequenceNum: false,
			Operations:           []txnbuild.Operation{&op},
			BaseFee:              txnbuild.MinBaseFee,
			Timebounds:           txnbuild.NewInfiniteTimeout(), // Use a real timeout in production!
		},
	)
```

vs

```golang
	tx, err := txnbuild.TransactionBuilder{
		SourceAccount:        &sourceAccount,
		IncrementSequenceNum: false,
		Operations:           []txnbuild.Operation{&op},
		BaseFee:              txnbuild.MinBaseFee,
		Timebounds:           txnbuild.NewInfiniteTimeout(), // Use a real timeout in production!
	}.build()
```

clients/horizonclient/client.go
tamirms: I wanted to reduce the number of combinations:

(feebump, transaction) x (with options, without options) = (feebump with options, fee bump without options, transaction with options, transaction without options)

but I don't feel too strongly about this so I could revert this change

clients/horizonclient/examples_test.go
tamirms: this is how the builder pattern looks for the java sdk:

```java
        Transaction transaction = new Transaction.Builder(account, Network.TESTNET)
                .addOperation(new CreateAccountOperation.Builder(destination.getAccountId(), "2000").build())
                .addMemo(Memo.text("Hello world!"))
                .setTimeout(Transaction.Builder.TIMEOUT_INFINITE)
                .build();
```

Note how `TransactionBuilder` doesn't persist beyond this statement which creates a `Transaction`.

I do think it's nice for there to be consistency among all the horizon sdk implementations. However, we should be mindful that there can be some differences between the implementations because each language has its own set of styles and conventions.

In this particular case, I would argue that constructing a transaction using `txnbuild.NewTransaction()` in the Go SDK is a good translation of  `new Transaction.Builder()...build()` from the Java SDK.  It may appear that they're inconsistent on a cosmetic level but on a practical and semantic level they are effectively equivalent.

ire-and-curses: I like the `NewTransaction` approach. I don't feel it's hard to understand what's happening with this naming.

2opremio: I would rather avoid dealing with a `Builder`, plus `NewTransaction()` fits a known pattern.

txnbuild/transaction.go
```diff
@@ -26,425 +28,742 @@ import (
...
+}
+
+// Operations returns the list of operations included in this transaction.
+// The returned slice should not be modified.
+func (t *Transaction) Operations() []Operation {
+	return t.operations
+}
+
+// Signatures returns the list of signatures attached to this transaction.
+// The returned slice should not be modified.
```
2opremio, line 211: Nitpick (the slice itself is returned as value).

```suggestion
// The contents of the returned slice should not be modified.
```

txnbuild/transaction.go
```diff
@@ -26,425 +28,742 @@ import (
...
+	return t.memo
+}
+
+// Timebounds returns the Timebounds configured for this transaction.
+func (t *Transaction) Timebounds() Timebounds {
+	return t.timebounds
+}
+
+// Operations returns the list of operations included in this transaction.
+// The returned slice should not be modified.
```
2opremio, line 205: Same here

2opremio:
I am missing a CHANGELOG entry warning the users about breaking changes. @tamirms mentioned this will come in a later PR. I think we should include examples indicating how to update the code (in this or another PR).

clients/horizonclient/examples_test.go
leighmcculloch: 👍

PR #2503, opened 2020-04-23
Update migration time

bartekn:
Includes https://github.com/stellar/go/commit/2f93e52e1aec9ed242ada518da239f93038a3db3 in 1.2.0 branch.

### DIFF
--- a/services/horizon/CHANGELOG.md
+++ b/services/horizon/CHANGELOG.md
@@ -9,7 +9,7 @@ file. This project adheres to [Semantic Versioning](http://semver.org/).
 
 This version includes a significant database migration which changes the column types of `fee_charged` and `max_fee` in the `history_transactions` table from `integer` to `bigint`. This essential change paves the way for fee bump transactions ([CAP 15](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0015.md)), a major improvement that will be released soon in Stellar Protocol 13.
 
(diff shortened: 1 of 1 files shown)


services/horizon/CHANGELOG.md
```diff
@@ -9,7 +9,7 @@ file. This project adheres to [Semantic Versioning](http://semver.org/).
 
 This version includes a significant database migration which changes the column types of `fee_charged` and `max_fee` in the `history_transactions` table from `integer` to `bigint`. This essential change paves the way for fee bump transactions ([CAP 15](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0015.md)), a major improvement that will be released soon in Stellar Protocol 13.
 
-This migration will run for a long time, especially if you have a Horizon database with full history. For reference, it took 20 hours to complete this migration on a AWS db.r4.xlarge instance with full transaction history.
+This migration will run for a long time, especially if you have a Horizon database with full history. For reference, it took around 8 hours and 42 minutes to complete this migration on a AWS db.r4.xlarge instance with full transaction history.
```
tamirms, line 12: I think the production db is an db.r4.8xlarge instance not db.r4.xlarge

bartekn: @abuiles can you confirm? I'm not sure where which environment used to test this.

tamirms: we ran the migrations first on staging which was a db.r4.xlarge instance and it took 20 hours. Then we ran it on production and it was faster ~8hours because the hardware for the production db was better

bartekn: @tamirms ok, fixed.

Issue #2447, opened 2020-04-03
Feature Request: Sort balances


abuiles:
### What problem does your feature solve?

Balances cam come in different order between request, the following responses contain the same data but the order is different in both:

```
"balances": [
    {
      "balance": "2.0100000",
      "limit": "922337203685.4775807",
      "buying_liabilities": "0.0000000",
      "selling_liabilities": "0.0000000",
      "last_modified_ledger": 14599295,
      "is_authorized": true,
      "asset_type": "credit_alphanum4",
      "asset_code": "SLT",
      "asset_issuer": "GCKA6K5PCQ6PNF5RQBF7PQDJWRHO6UOGFMRLK3DYHDOI244V47XKQ4GP"
    },
    {
      "balance": "1.0000000",
      "limit": "922337203685.4775807",
      "buying_liabilities": "0.0000000",
      "selling_liabilities": "0.0000000",
      "last_modified_ledger": 11479636,
      "is_authorized": true,
      "asset_type": "credit_alphanum4",
      "asset_code": "HUG",
      "asset_issuer": "GD5T6IPRNCKFOHQWT264YPKOZAWUMMZOLZBJ6BNQMUGPWGRLBK3U7ZNP"
    },
    {
      "balance": "0.0099999",
      "limit": "922337203685.4775807",
      "buying_liabilities": "0.0000000",
      "selling_liabilities": "0.0000000",
      "last_modified_ledger": 10135487,
      "is_authorized": true,
      "asset_type": "credit_alphanum4",
      "asset_code": "EUR",
      "asset_issuer": "GDI73WJ4SX7LOG3XZDJC3KCK6ED6E5NBYK2JUBQSPBCNNWEG3ZN7T75U"
    },
    {
      "balance": "0.0024939",
      "limit": "922337203685.4775807",
      "buying_liabilities": "0.0000002",
      "selling_liabilities": "0.0000000",
      "last_modified_ledger": 19967410,
      "is_authorized": true,
      "asset_type": "credit_alphanum4",
      "asset_code": "DEMO",
      "asset_issuer": "GBAMBOOZDWZPVV52RCLJQYMQNXOBLOXWNQAY2IF2FREV2WL46DBCH3BE"
    },
    {
      "balance": "0.0000000",
      "limit": "922337203685.4775807",
      "buying_liabilities": "0.0000000",
      "selling_liabilities": "0.0000000",
      "last_modified_ledger": 20511785,
      "is_authorized": true,
[...]

PR #2516, opened 2020-04-27
services/horizon: Sort trustlines by asset code and issuer

2opremio:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

When obtaining trustlines from the DB, for rendering into the `GET /accounts` endpoint, order them by asset code (and issuer) so that the output is stable (postgres doesn't guarantee a consistent ordering).

### Why

`horizon-cmp` outputs a large amount of false positives because of this, plus it's always a good idea to generate a stable output, independent of the Horizon server used.

### Known limitations

The DB query becomes slower. I pondered ordering it on the Horizon-side, but I figured a DB engine would be more efficient at this.

### DIFF
--- a/services/horizon/internal/actions/account.go
+++ b/services/horizon/internal/actions/account.go
@@ -401,7 +401,7 @@ func (handler GetAccountsHandler) loadData(historyQ *history.Q, accounts []strin
 func (handler GetAccountsHandler) loadTrustlines(historyQ *history.Q, accounts []string) (map[string][]history.TrustLine, error) {
 	trustLines := make(map[string][]history.TrustLine)
 
-	records, err := historyQ.GetTrustLinesByAccountsID(accounts)
+	records, err := historyQ.GetSortedTrustLinesByAccountIDs(accounts)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


bartekn:
> The DB query becomes slower. I pondered ordering it on the Horizon-side, but I figured a DB engine would be more efficient at this.

I wonder how slower it would be (we should test it in staging). `trust_line` table doesn't contain an index that could be directly used in this query. After filtering but `account_id` the result set should be small enough for most of the accounts to not affect performance but I've seen Postgres not use indices it could many times.

tamirms:
> but I've seen Postgres not use indices it could many times.

that's a really good point. we can run an explain query on the staging pubnet db and see what it produces

2opremio:
This fixed #2447 

I will check the query in the staging db and report back

2opremio:
> I wonder how slower it would be (we should test it in staging). `trust_line` table doesn't contain an index that could be directly used in this query.

Since it doesn't contain an index, I doubt it will make things considerably worse (the table needs to be scanned anyways).

PR #2521, opened 2020-04-28
horizonclient: Update SEP29 implementation to skip checks on memo id destinations

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Update SEP29 implementation to skip checks on muxed account addresses which contain a memo id.

### Why

We should skip the check if the destination is an M address since M addresses essentially encode a memo within them.

### Known limitations

[N/A]

### DIFF
--- a/clients/horizonclient/main_test.go
+++ b/clients/horizonclient/main_test.go
@@ -914,6 +914,31 @@ func TestSubmitTransactionXDRRequest(t *testing.T) {
 	}
 }
 
+func TestCheckMemoRequiredSkipsMemoIDAddress(t *testing.T) {
+	client := &Client{HorizonURL: "https://localhost/"}
+
+	kp := keypair.MustParseFull("SA26PHIKZM6CXDGR472SSGUQQRYXM6S437ZNHZGRM6QA4FOPLLLFRGDX")
+	sourceAccount := txnbuild.NewSimpleAccount(kp.Address(), int64(0))
+
+	payment := txnbuild.Payment{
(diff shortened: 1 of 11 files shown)


clients/horizonclient/main_test.go
```diff
@@ -914,6 +914,31 @@ func TestSubmitTransactionXDRRequest(t *testing.T) {
...
+
+	payment := txnbuild.Payment{
+		Destination: "MCAAAAAAAAAAAAB7BQ2L7E5NBWMXDUCMZSIPOBKRDSBYVLMXGSSKF6YNPIB7Y77ITKNOG",
+		Amount:      "10",
+		Asset:       txnbuild.NativeAsset{},
+	}
+
+	tx, err := txnbuild.NewTransaction(
+		txnbuild.TransactionParams{
+			SourceAccount:        &sourceAccount,
```
abuiles, line 931: @tamirms this is out of the scope of this PR, but could we extend `SimpleAccount` to work with `M` address too? The `js-sdk` allows people to pass `M..` addresses anywhere were a muxed account is the type.

abuiles: Which I see we did here, but not for `SourceAccount`- it seems like when building a transaction we expected the source to be a `Ed25519`

abuiles: oh, maybe that makes sense for the next release -- since you'll need to handle the transaction type here https://github.com/tamirms/go/blob/a3f36dae047c0d26c120838a49ae0b486b3dd72d/txnbuild/transaction.go#L569-L583

PR #2527, opened 2020-04-29
support/http: propagate request id to all logs

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Propagate the request ID to the requests context logger making the request ID show up in all logs for a request.

### Why
For every request we receive in Friendbot, Webauth, Recoverysigner, and other applications are assigned a request ID which is logged at the beginning and end of the request. Matching the beginning and end of a request is one reason to use a request ID, but the main reason to log a request ID is to make it easier to identify all logs relating to that request, not just the beginning and end of the request. This allows us to look at the logs of a server and identify definitively which logs relate to the same request.

Horizon uses separate code that follows the same pattern and this fix was made to its version of this code in 2018: https://github.com/stellar/go/commit/36345f68ef6a30e55d8c1a02cb82613dc3d9b041#diff-3702689dfe4e807fe1e3ba73798bff50L32-R29

I also did a small refactoring of how the context is passed into the helper functions because while it is normally good practice to pass the context as the first parameter into subsequent functions it is duplicative to the request since the request contains the context. By only passing the request we are clearly signaling that the context in the request is the ultimate context that should be in use.

### Known limitations

N/A

### DIFF
--- a/support/http/logging_middleware_test.go
+++ b/support/http/logging_middleware_test.go
@@ -19,6 +19,7 @@ func TestHTTPMiddleware(t *testing.T) {
 	mux.Use(LoggingMiddleware)
 
 	mux.Get("/", stdhttp.HandlerFunc(func(w stdhttp.ResponseWriter, r *stdhttp.Request) {
+		log.Ctx(r.Context()).Info("handler log line")
 	}))
 	mux.Handle("/not_found", stdhttp.NotFoundHandler())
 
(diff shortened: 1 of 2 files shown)


tamirms:
> Horizon uses separate code that follows the same pattern and this fix was made to its version of this code in 2018: 36345f6#diff-3702689dfe4e807fe1e3ba73798bff50L32-R29

Is there any reason why we should maintain two logging middleware implementations? can we consolidate the two implementations and have horizon use the logging middleware in the support pacakge?

leighmcculloch:
@tamirms 💯We should do that. There are other middleware that are also common. Ideally Horizon would use the `"support/http".NewAPIMux` function to get an initially configured mux and then build on it, assuming the other middleware in there is not incompatible with Horizon's current configuration.

support/http/logging_middleware_test.go
```diff
@@ -29,18 +30,35 @@ func TestHTTPMiddleware(t *testing.T) {
...
-		assert.Equal(t, "finished request", logged[3].Message)
-	}
+		assert.Equal(t, "http", logged[0].Data["subsys"])
+		assert.Equal(t, "GET", logged[0].Data["method"])
+		assert.NotEmpty(t, logged[0].Data["req"])
+		assert.NotEmpty(t, logged[0].Data["path"])
+		req1 := logged[0].Data["req"]
+
+		assert.Equal(t, "handler log line", logged[1].Message)
+		assert.Equal(t, req1, logged[0].Data["req"])
```
marcelosalloum, line 42: I don't see the purpose of testing `assert.Equal(t, req1, logged[0].Data["req"])` if you just defined it in line `39`. Should this be `logged[1]`, instead?

```suggestion
		assert.Equal(t, req1, logged[1].Data["req"])
```

support/http/logging_middleware_test.go
```diff
@@ -64,17 +83,34 @@ func TestHTTPMiddlewareWithLoggerSet(t *testing.T) {
...
```
marcelosalloum, line 95: Same here

ire-and-curses:
> @tamirms 100We should do that. There are other middleware that are also common. Ideally Horizon would use the `"support/http".NewAPIMux` function to get an initially configured mux and then build on it, assuming the other middleware in there is not incompatible with Horizon's current configuration.

Raised #2529

support/http/logging_middleware_test.go
leighmcculloch:
```suggestion
		assert.Equal(t, req1, logged[1].Data["req"])
```

PR #2537, opened 2020-05-01
services/horizon: Add local last ingested ledger metric

### LINKED ISSUE #2530: Add local last ingested ledger sequence to /metrics
We already include last ingested ledger sequence to a DB. We should also add local (order book graph) last ingested sequence.

2opremio:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add a Prometheus metric for the latest local (order book graph) ledger being ingested.

### Why

Fixes #2530

### Known limitations

N/A

### DIFF
--- a/protocols/horizon/main.go
+++ b/protocols/horizon/main.go
@@ -322,6 +322,7 @@ type Root struct {
 	HorizonVersion               string `json:"horizon_version"`
 	StellarCoreVersion           string `json:"core_version"`
 	IngestSequence               uint32 `json:"ingest_latest_ledger"`
+	OrderBookSequence            uint32 `json:"local_latest_ledger"`
 	HorizonSequence              int32  `json:"history_latest_ledger"`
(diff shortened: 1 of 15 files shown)


tamirms:
there is a case in distributed ingestion where the database is up to date with the latest ledger but the in memory orderbook graphs on some nodes are still behind. In this scenario, `completeIngestion()` is not called when updating just the orderbooks:

https://github.com/stellar/go/blob/cf9bac26631945acf4633adbd5affa333d7880ac/services/horizon/internal/expingest/fsm.go#L397-L439

So, if we update the metrics in `completeIngestion()` we will not be capturing nodes which have out of date orderbooks. To fix this we can refactor the code so that `completeIngestion()` is called even when only updating orderbooks. Or, we can update the promethes metric in two places (once in `completeIngestion()` and the other time in the special case of the resume state).

Issue #2535, opened 2020-05-01
Check for possible memleak when --exp-ingest-in-memory-only and ingestion errors


bartekn:
### What version are you using?

horizon 1.2.0

### What did you do?

Earlier today ingestion in our stg environment halted because of https://github.com/stellar/go/issues/2534. During that time prd requests were mirrored to our stg environment using goreplay.

### What did you expect to see?

Stable memory usage of Horizon instance.

### What did you see instead?

An increase in memory usage up to the limit on all of the frontend instances with request mirroring. Please note that it didn't happen on ingestion only (backend) instances and frontend instances without request mirroring. CPU usage was normal up to memory usage <100%.

`horizon-003a` (UTC+2 timezone) 
<img width="859" alt="Screenshot 2020-05-01 at 16 45 30" src="https://user-images.githubusercontent.com/464938/80814035-6f4a8c80-8bcb-11ea-9b2a-aba8a7ba7e4e.png">

CPU usage and EBS burst quota on front-end instances (UTC timezone):
<img width="1525" alt="Screenshot 2020-05-01 at 17 32 24" src="https://user-images.githubusercontent.com/464938/80817693-39f56d00-8bd2-11ea-89c1-4ad3c2bdab5a.png">

It's also possible that it's a goreplay issue. If so we can close this issue.

bartekn:
I run an experiment to reproduce this. I stopped backend ingestion with goreplay running on frontend nodes. The RAM usage started growing again:
<img width="859" alt="Screenshot 2020-05-01 at 19 47 16" src="https://user-images.githubusercontent.com/464938/80828032-ef312080-8be4-11ea-8130-b14974084fb0.png">
I will run the experiment again next week with mem profiler.

bartekn:
I haven't confirmed it but my best bet so far is connected to `HistoryDBSource`. In `stream_handler.go` we `select`:
https://github.com/stellar/go/blob/00d7d2f8600e5389b20c3142b5b90550e9b75168/services/horizon/internal/render/sse/stream_handler.go#L73-L79
If there are no new ledgers `NextLedger` will stay in an infinite loop here:
https://github.com/stellar/go/blob/00d7d2f8600e5389b20c3142b5b90550e9b75168/services/horizon/internal/ledger/ledger_source.go#L38-L58

Every time a new stream request is started `NextLedger` starts a new go routine that takes memory but even when stream is closed the go routines stay there until there's a new ledger.

bartekn:
[...]

PR #2548, opened 2020-05-04
services/horizon/ledger: Fix memory leak in HistoryDBSource

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This commit fixes a memory leak in `HistoryDBSource.NextLedger` method. The method is usually used like in `stream_handler.go`:
https://github.com/stellar/go/blob/00d7d2f8600e5389b20c3142b5b90550e9b75168/services/horizon/internal/render/sse/stream_handler.go#L73-L79
When streaming connection is closed but the are no new ledgers (ex. due to database or network issues) the goroutines started in `HistoryDBSource.NextLedger` will not be closed.

To fix this `ledger.Source` interface was updated with `Close()` method that should be called when a ledger source is no longer used and it exist internal goroutine. The code was refactored to create a new `HistoryDBSource` for each streaming request (so it can be closed when handler execution ends).

### Why

When there are no new ledgers, the goroutines started in `HistoryDBSource.NextLedger` will leak and will eventually consume all available memory, like in https://github.com/stellar/go/issues/2535.

### Known limitations

The new code creates a new `HistoryDBSource` for each streaming request (instead of reusing a single `HistoryDBSource` in all streaming requests). This should have a really minimal impact on memory usage.

### DIFF
--- a/services/horizon/internal/ledger/ledger_source.go
+++ b/services/horizon/internal/ledger/ledger_source.go
@@ -21,7 +21,9 @@ type currentStateFunc func() State
 type HistoryDBSource struct {
 	updateFrequency time.Duration
 	currentState    currentStateFunc
-	closed          bool
+
+	closedLock sync.Mutex
+	closed     bool
 }
 
 // NewHistoryDBSource constructs a new instance of HistoryDBSource
@@ -29,6 +31,7 @@ func NewHistoryDBSource(updateFrequency time.Duration) *HistoryDBSource {
 	return &HistoryDBSource{
 		updateFrequency: updateFrequency,
 		currentState:    CurrentState,
+		closedLock:      sync.Mutex{},
 	}
 }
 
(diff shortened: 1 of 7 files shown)


services/horizon/internal/ledger/ledger_source.go
```diff
@@ -46,6 +49,10 @@ func (source HistoryDBSource) NextLedger(currentSequence uint32) chan uint32 {
 				time.Sleep(source.updateFrequency)
 			}
 
+			if source.closed {
```
tamirms, line 52: another go routine may write to `source.closed`  so I think technically the behavior is undefined. I think it would be a good idea to use a lock or use a context to signal cancelation

bartekn: You're right. I was under the impression that single read/single write is totally fine but https://golang.org/ref/mem is clear that it's not. Fixed by using explicit synchronization.

Issue #1790, opened 2019-09-27
services/ticker: support markets/tickers with base/counter assets in both directions


leighmcculloch:
The Ticker API's `markets` and `tickers` REST endpoints and GraphQL queries for all trade pairs only ever return market and aggregated ticker information with the data relative to one of the codes in the pair as the base.

Example:

Making a request with `XLM_USD` or `USD_XLM` to the `ticker` GraphQL query returns the data for when `USD` is the base and `XLM` is the counter. There's no way to ask the API for the ticker or markets where `XLM` is the base and `USD` is the counter.

Request:
```gql
{
  XLM_USD:ticker(pairName:"XLM_USD") {
    tradePair
    close
  }
  USD_XLM:ticker(pairName:"XLM_USD") {
    tradePair
    close
  }
}
```

Response:
```json
{
  "data": {
    "XLM_USD": [
      {
        "tradePair": "XLM_USD",
        "close": 17.415352678307087
      }
    ],
    "USD_XLM": [
      {
        "tradePair": "XLM_USD",
        "close": 17.415352678307087
      }
```

This makes it harder to use the API. When querying any trade pair the user of the API has to check if the trade pair has been returned in the reversed order as asked and then invert the data and values. It is trivial for a calling app to do `1/close` to get the close for the reverse, but less trivial to navigate many of the other fields in the response where `ask` and `bid` will be reversed. It's likely that this behavior will cause users of the API to make mistakes.

Let's allow returning trade pairs in the order that was in the request.

@accordeiro @tomquisel Anything I'm missing?

Issue #1784, opened 2019-09-26
services/ticker: add base and counter to ticker/markets endpoints/queries


leighmcculloch:
The `markets` and `ticker` JSON endpoints and GraphQL query fields return a `tradePair` field which has the base currency on the right-hand-side and the counter on the left-hand-side.

This has caused some confusion that led to #1781 and #1782. Independent of those issues it is not clear when looking only at the responses which of the currencies is the base and counter. The concepts base and counter are explicit in the response because other fields include those concepts, such as `baseVolume` and `counterVolume`.

Let's make it explicit in the response which currency is the base and which is the counter by adding two new fields to both endpoints called `base` and `counter`. Each will contain their respective code.

We can do this irrespective of whether #1782 happens, since if the codes are swapped or not it will be more explicit to state in the response which is the base and which is the counter.

Markets JSON example response:

```json
{
  "name": "ABDT_DOP",
  "base": "DOP",
  "counter": "ABDT",
  "base_volume": 27933.1306978,
  "counter_volume": 703779.0492835,
```

Markets GraphQL example response:

```json
{
  "data": {
    "markets": [
      {
        "tradePair": "XLM:native / LIT:GBOXAJLALDGVCKSCBV6GSAOOZ562OLFIWXYXPKVE3T2LCRKJ5RAQO5IG",
        "base": "LIT:GBOXAJLALDGVCKSCBV6GSAOOZ562OLFIWXYXPKVE3T2LCRKJ5RAQO5IG",
        "counter": "XLM:native",
        "close": 0.0001115
      },
```

Ticker GraphQL example response:

```json
{
  "data": {
    "ticker": [
      {
        "tradePair": "XLM_USD",
        "base": "USD",
        "counter": "XLM",
        "close": 17.195100771888075
      }
    ]
```

cc @accordeiro

leighmcculloch:
I don't think we should add this until after #1782 where we fix the counter and base terminology usage because there are other queries (see https://github.com/stellar/go/issues/1782#issuecomment-538514638) that are not using the terms correctly and so to add this would be educational but also contradictory to those fields which would create confusion.

PR #2549, opened 2020-05-05
services/ticker: swap base/counter in GraphQL

### LINKED ISSUE #1782: services/ticker: swap counter and base in trade pairs
As was raised in #1781 the order of the counter and base currencies in the name of the trade pairs returned by the API is confusing because traditionally the base currency on the left-hand-side and the counter currency is on the right-hand-side.

How it works in Ticker: `<Counter>_<Base>`.

Example:

The pair `XLM_ZZZ` has the `XLM` as the counter currency and `ZZZ` as the base. For that pair if the API returns a `close` value of `2`, then the last trade for the pair was `2 XLM` exchanged for `1 ZZZ`.

How it is normally expressed: `<Base>_<Counter>`

Examples:
- https://en.wikipedia.org/wiki/Currency_pair
- https://www.investopedia.com/terms/b/basecurrency.asp

debnil:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
This PR swaps the base and counter in the ticker and markets GraphQL queries. To maintain backwards compatibility, the user can request this behavior through a `newEndpoint` flag.

### Why
This makes the ticker more in-line with financial market expectations. It closes #1782, which also has more details about this issue. It also addresses #1790 for the `ticker` endpoint. 

### Known limitations
The JSON endpoints and the older version still have the prior Ticker behavior, where trade pairs are represented as `<Counter>_<Base>`. I've indicated the conditions for the new behavior on the API documentation. Unfortunately, to preserve backwards compatibility for existing users of these services, we can't fully reverse trading pairs. Follow up PRs (like one addressing #1784, and another addressing #1790 for the `markets` endpoint) will clarify some of these idiosyncrasies.


services/ticker/internal/gql/resolvers_market.go
```diff
@@ -116,3 +122,74 @@ func dbMarketToPartialMarket(dbMarket tickerdb.PartialMarket) *partialMarket {
...
+	processedDbMkt = &partialMarket{
+		TradePair:            *oldPairName,
+		BaseAssetCode:        dbMkt.CounterAssetCode,
+		BaseAssetIssuer:      dbMkt.CounterAssetIssuer,
+		CounterAssetCode:     dbMkt.BaseAssetCode,
+		CounterAssetIssuer:   dbMkt.BaseAssetIssuer,
+		BaseVolume:           dbMkt.BaseVolumeReverse,
+		CounterVolume:        dbMkt.CounterVolumeReverse,
+		TradeCount:           dbMkt.TradeCount,
+		Open:                 1 / dbMkt.Open,
```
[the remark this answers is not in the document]
debnil: This comment is outdated, and the current computation all looks correct :)

services/ticker/internal/tickerdb/queries_market.go
```diff
@@ -203,7 +203,9 @@ SELECT
 	cAsset.issuer_account AS counter_asset_issuer,
 	cAsset.type as counter_asset_type,
 	sum(t.base_amount) AS base_volume,
+	sum(1/t.base_amount) AS base_volume_reverse,
```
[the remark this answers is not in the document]
debnil: There's actually no need for this field. If the base and counter assets are swapped, their total volumes can be as well.

services/ticker/internal/gql/resolvers_market.go
```diff
@@ -15,6 +16,7 @@ func (r *resolver) Markets(args struct {
 	CounterAssetCode   *string
 	CounterAssetIssuer *string
 	NumHoursAgo        *int32
+	IsNewEndpoint      *bool
```
accordeiro, line 19: I feel like passing a boolean argument that changes the behavior completely probably makes a good case to create a separate endpoint. Besides, `isNewEndpoint` doesn't clearly state what the behavior change would be.

I have a few alternative proposals:
- [Cleaner, more complex] create a `standardMarkets`, query, which returns a new `StandardMarket` object; or
- [More hackish, but simpler] change the behavior so that `isNewEndpoint` becomes `shouldReverseMarkets` which is `true` by default, and if you want the correct order you have to manually pass `false` to the query.

services/ticker/internal/gql/resolvers_market.go
```diff
@@ -34,18 +36,23 @@ func (r *resolver) Markets(args struct {
...
 }
 
 // Ticker resolves the ticker() GraphQL query (TODO)
 func (r *resolver) Ticker(
 	args struct {
-		PairName    *string
-		NumHoursAgo *int32
+		PairName      *string
+		NumHoursAgo   *int32
+		IsNewEndpoint *bool
```
accordeiro, line 55: Same as comment above ⬆️

services/ticker/internal/gql/resolvers_market.go
```diff
@@ -116,3 +123,82 @@ func dbMarketToPartialMarket(dbMarket tickerdb.PartialMarket) *partialMarket {
...
+) *partialMarket {
+	// If a nil partial market was passed, return it.
+	if dbMkt == nil {
+		return dbMkt
+	}
+
+	// If the user does not provide the new endpoint flag,
+	// we assume they want the pre-existing behavior. (This also
+	// assures backwards compatibility.)
+	if isNewEndpoint == nil {
```
accordeiro, line 141: Same comment about `isNewEndpoint` vs `shouldReverseMarkets`

services/ticker/internal/gql/resolvers_market.go
```diff
@@ -116,3 +123,82 @@ func dbMarketToPartialMarket(dbMarket tickerdb.PartialMarket) *partialMarket {
...
+		oldPairNameStr = *oldPairName
+	}
+
+	// If the user-requested trade pair already matches the name
+	// of the generated partial market, the base and counter assets and
+	// volume are as expected. Prices must be inverted to match the industry
+	// convention, rather than the original ticker implementation.
+	processedDbMkt := *dbMkt
+	if oldPairNameStr == dbMkt.TradePair || oldPairName == nil || oldPairNameStr == "" {
+		processedDbMkt.Open = 1 / dbMkt.Open
```
accordeiro, line 167: Should we account for `by zero` divisions here?

services/ticker/internal/gql/static/schema.gql
```diff
@@ -11,21 +11,27 @@ type Query {
...
+	# base asset info for filtering. optionally provide whether
+	# or not the query comes from the new version of the endpoint,
+	# as this affects how the base/counter are interpreted.
 	markets(
 		baseAssetCode: String
 		baseAssetIssuer: String
 		counterAssetCode: String
 		counterAssetIssuer: String
 		numHoursAgo: Int
+		isNewEndpoint: Boolean
```
accordeiro, line 23: See first comment re: new endpoint / change of boolean flag ⬆️

services/ticker/internal/gql/static/schema.gql
```diff
@@ -11,21 +11,27 @@ type Query {
...
 	# retrieve aggregated trade stats for the last <numHoursAgo>
 	# hours. optionally provide the name of the trade pair (e.g.
-	# XLM_BTC) for filtering results.
+	# XLM_BTC) for filtering results. optionally provide whether
+	# or not the query comes from the new version of the endpoint,
+	# as this affects how the trade pair is interpreted.
 	ticker(
 		pairName: String
 		numHoursAgo: Int
+		isNewEndpoint: Boolean
```
accordeiro, line 34: See first comment re: new endpoint / change of boolean flag ⬆️

services/ticker/internal/tickerdb/queries_market.go
```diff
@@ -218,9 +218,16 @@ SELECT
 	COALESCE((array_agg(os.highest_bid))[1], 0.0) AS highest_bid,
 	COALESCE((array_agg(os.num_asks))[1], 0) AS num_asks,
 	COALESCE((array_agg(os.ask_volume))[1], 0.0) AS ask_volume,
-	COALESCE((array_agg(os.lowest_ask))[1], 0.0) AS lowest_ask
+	COALESCE((array_agg(os.lowest_ask))[1], 0.0) AS lowest_ask,
```
accordeiro, line 221: Nice! Did you notice any performance impacts? I wonder if we should write two separate queries instead.

debnil: Nope! Change isn't perceptible in GraphQL, and it was a really small change when run directly in psql.

services/ticker/docs/API.md
```diff
@@ -17,6 +17,8 @@ Example:
 
 The pair `XLM_ZZZ` has the `XLM` as the counter currency and `ZZZ` as the base. For that pair if the API returns a `close` value of `2`, then the last trade for the pair was `2 XLM` exchanged for `1 ZZZ`.
 
+Note that if the `ShouldReverseMarkets` `boolean` argument is passed as `false` to the `ticker` or `markets` queries, then trade pairs will be interpreted as `<Base>_<Counter>` (the reverse logic of the above).
```
leighmcculloch, line 20: `ShouldReverseMarkets` does not really communicate the significance of the new and old states, and it will be difficult to safely sunset the old broken behavior. This is because if we remove the old behavior and make nil mean true, the behavior will change for folks, or if we make the field required it will be confusing for users why they need to reverse markets.

services/ticker/internal/gql/resolvers_market.go
```diff
@@ -116,3 +123,89 @@ func dbMarketToPartialMarket(dbMarket tickerdb.PartialMarket) *partialMarket {
...
+	if dbMkt == nil {
+		return dbMkt
+	}
+
+	// If the user does not provide the reverse markets flag,
+	// we assume they want the pre-existing behavior. (This also
+	// assures backwards compatibility.)
+	if shouldReverseMarkets == nil {
+		return dbMkt
+	}
```
leighmcculloch, line 143: I think treating `shouldReverseMarkets` `false` and nil` would be a trap for me personally, and possibly others, because I assume booleans have a default of `false`.

services/ticker/internal/gql/resolvers_market.go
```diff
@@ -116,3 +123,89 @@ func dbMarketToPartialMarket(dbMarket tickerdb.PartialMarket) *partialMarket {
...
+	// We construct a partial market with the base and counter assets reversed.
+	// Note that we don't need to recompute prices, because they had already
+	// been computed with respect to the original counter asset (now base).
+	processedDbMkt.TradePair = *oldPairName
+	processedDbMkt.BaseAssetCode = dbMkt.CounterAssetCode
+	processedDbMkt.BaseAssetIssuer = dbMkt.CounterAssetIssuer
+	processedDbMkt.BaseVolume = dbMkt.CounterVolume
+	processedDbMkt.CounterAssetCode = dbMkt.BaseAssetCode
+	processedDbMkt.CounterAssetIssuer = dbMkt.BaseAssetIssuer
+	processedDbMkt.CounterVolume = dbMkt.BaseVolume
```
leighmcculloch, line 184: All this logic looks accurate to me, but it is complex conceptually. I had to read the comments a couple times to understand how these parts come together. Avoiding this complex logic is a reason I'd lean towards fixing this as a breaking change without any new parameter and cutting a major release of Ticker that is a new deployment at a new base URL.

debnil:
Made obsolete by #2595, which takes a clean break approach to the existing Ticker service.

PR #2553, opened 2020-05-05
services/horizon/internal/txsub: Add metrics to track which types of transaction Horizon is receiving

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add metrics to track which types of transaction (v0, v1, or fee bump) Horizon is receiving.

### Why

Prior to protocol 13 there was only kind of transaction you could submit to Horizon. With these metrics we can monitor how many people actually use fee bump transactions and how many people are still submitting v0 transactions.

Note that we could also add hooks to ingestion which monitor the total number of fee bump transactions in a ledger. However, we will not be able to monitor the number of v0 transactions which were included in a ledger from the ingestion code because, once protocol 13 is enabled, Stellar Core will rewrite v0 transactions as v1 transactions.
 
### Known limitations

We shouldn't have a separate metrics for each transaction type. 

> When you have multiple metrics that you want to add/average/sum, they should usually be one metric with labels rather than multiple metrics.

>For example, rather than http_responses_500_total and http_responses_403_total, create a single metric called http_responses_total with a code label for the HTTP response code. You can then process the entire metric as one in rules and graphs.

From https://prometheus.io/docs/practices/instrumentation/#use-labels 

This issue also applies to many of the existing horizon metrics such as `txsub.succeeded` and `txsub.failed`. 

The go-metrics library does not support creating metrics with tags or labels, see https://github.com/rcrowley/go-metrics/issues/135 . We should consider using the prometheus client directly for metrics https://github.com/go-kit/kit/blob/master/metrics/prometheus/prometheus.go . Alternatively we could use a library like https://github.com/armon/go-metrics which does support creating metrics with labels.

### DIFF
--- a/services/horizon/internal/init.go
+++ b/services/horizon/internal/init.go
@@ -134,6 +134,9 @@ func initTxSubMetrics(app *App) {
 	app.metrics.Register("txsub.open", app.submitter.Metrics.OpenSubmissionsGauge)
 	app.metrics.Register("txsub.succeeded", app.submitter.Metrics.SuccessfulSubmissionsMeter)
 	app.metrics.Register("txsub.failed", app.submitter.Metrics.FailedSubmissionsMeter)
+	app.metrics.Register("txsub.v0", app.submitter.Metrics.V0TransactionsMeter)
+	app.metrics.Register("txsub.v1", app.submitter.Metrics.V1TransactionsMeter)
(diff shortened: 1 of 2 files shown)


bartekn:
Thanks for https://github.com/stellar/go/pull/2553/commits/74713974b76c49454640daa382d8425cf5e53d01!

Issue #2018, opened 2019-12-06
services/horizon/expingest: Separate DB connection pool for ingestion service


bartekn:
Currently, all of the Horizon services share the same DB connection pool. In case of increased number of HTTP requests, it's possible that ingestion service will have to wait for an idle connection. We need to create a separate connection pool (with a number of open connections equal to [horizon max] - [max needed by ingestion at any given time: 2 right now?]) for ingestion service to make sure it can always proceed.

ire-and-curses:
Now that we have split ingesting and front-facing nodes, this isn't necessary for our cluster. It could still potentially be helpful for single machine deployments. Let's revisit if that seems to be the case.

bartekn:
Reopening. Frontend nodes still run ingestion because they ingest offers into in-memory order book.

PR #2560, opened 2020-05-07
services/horizon/internal: Create separate db connections for ingestion

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Fixes https://github.com/stellar/go/issues/2018

Create separate db connections to the horizon and stellar core databases for ingestion.

### Why

> Currently, all of the Horizon services share the same DB connection pool. In case of increased number of HTTP requests, it's possible that ingestion service will have to wait for an idle connection.

### Known limitations

[N/A]

### DIFF
--- a/services/horizon/internal/init.go
+++ b/services/horizon/internal/init.go
@@ -19,8 +19,6 @@ import (
 	"github.com/stellar/go/support/log"
 )
 
-const maxIngestionDBConnections = 2
-
 func mustNewDBSession(databaseURL string, maxIdle, maxOpen int) *db.Session {
 	session, err := db.Open("postgres", databaseURL)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


services/horizon/internal/init.go
```diff
@@ -19,33 +19,60 @@ import (
...
 
-func mustInitCoreDB(app *App) {
-	session, err := db.Open("postgres", app.config.StellarCoreDatabaseURL)
-	if err != nil {
-		log.Fatalf("cannot open Core DB: %v", err)
+func mustInitHorizonDB(app *App) {
+	maxIdle := app.config.CoreDBMaxIdleConnections
+	maxOpen := app.config.CoreDBMaxOpenConnections
+	if maxIdle >= maxIngestionDBConnections*2 {
+		maxIdle -= maxIngestionDBConnections
```
abuiles, line 39: could you drop a comment on why do we do this comparison (what does the `2` means for example) and why do we change the value

bartekn:
We can probably merge it into `master` as previous PRs for this release (no breaking changes).

services/horizon/internal/init.go
```diff
@@ -19,33 +19,70 @@ import (
 	"github.com/stellar/go/support/log"
 )
 
-func mustInitHorizonDB(app *App) {
-	session, err := db.Open("postgres", app.config.DatabaseURL)
+const maxIngestionDBConnections = 2
```
bartekn, line 22: I think it belongs to the `ingest` package. Say, we add a new routine that require an extra DB connection. Would be easier to increment it there.

services/horizon/internal/init.go
```diff
@@ -19,33 +19,70 @@ import (
...
-	if err != nil {
-		log.Fatalf("cannot open Core DB: %v", err)
+func mustInitHorizonDB(app *App) {
+	maxIdle := app.config.HorizonDBMaxIdleConnections
+	maxOpen := app.config.HorizonDBMaxOpenConnections
+	// There are two connection pools, one for serving requests to horizon
+	// and another pool for ingestion. Ideally the total connections in both
+	// pools should be bounded by HorizonDBMaxOpenConnections. But, if the ingestion
+	// pool consumes a significant quota of HorizonDBMaxOpenConnections then we will
+	// allow a total of HorizonDBMaxOpenConnections + maxIngestionDBConnections connections.
```
bartekn, line 42: I think we should just return an error (maybe at a command validation stage in `root.go`) when ingestion connections is equal/greater than max set by a user. The reason is that user may explicitly set the max number of open connections in Postgres config so these extra 2 connections won't be established (and will probably generate a bunch of errors in the log).

PR #2570, opened 2020-05-08
support/db: Add timeout on horizon db ingestion statements

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add timeout on horizon db select and exec operations.

### Why

Currently there is no timeout on any of the horizon DB operations. That means a postgres query which hangs indefinitely could block ingestion from progressing. 

### Known limitations

[N/A]

### DIFF
--- a/services/horizon/internal/expingest/main.go
+++ b/services/horizon/internal/expingest/main.go
@@ -47,6 +47,8 @@ const (
 	// MaxDBConnections is the size of the postgres connection pool dedicated to Horizon ingestion
 	MaxDBConnections = 2
 
+	dbSessionTimeout = 20 * time.Second
+
 	defaultCoreCursorName           = "HORIZON"
 	stateVerificationErrorThreshold = 3
 )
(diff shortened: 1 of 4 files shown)


bartekn:
I wonder if we still need it since we have https://github.com/stellar/go/pull/2560 now? AFAIR, adding timeout was suppose to fix issues when ingestion is starved by HTTP server but it shouldn't be a problem any longer?

ire-and-curses:
> I wonder if we still need it since we have #2560 now? AFAIR, adding timeout was suppose to fix issues when ingestion is starved by HTTP server but it shouldn't be a problem any longer?

Is there any downside to a timeout? Having the possibility of these hanging forever seems worse?

tamirms:
I'm ok with closing the PR if that is your preference @bartekn

bartekn:
> Is there any downside to a timeout? Having the possibility of these hanging forever seems worse?

It's a problem if it's doing nothing (ex. waiting for a free connection from a pool as before https://github.com/stellar/go/pull/2560) but if the query actually executes but a little longer I think it's actually better to wait. With timeout it's possible that ingestion will fall into an infinite loop (slow query, timeout, slow query, timeout, ...).

> I'm ok with closing the PR if that is your preference @bartekn

I think we should leave the timeout code in `db.Session`, just remove it from `ingest`. And maybe let's add other `stellar/go` members to review because it's a shared package.

tamirms:
Going to close this PR since there is no immediate use case anymore

Issue #2579, opened 2020-05-11
Error Parsing Price for ManageSellOffer Operation


qiwu7:
### What version are you using?
5b5b4d6174e131b20380ec43fab55d4908106d4f

### What did you do?
use the following to retrieve envelope xdr
```
curl --location --request GET 'https://horizon-testnet.stellar.org/ledgers/142220/transactions' | jq '._embedded.records[1].envelope_xdr'
```

```
func main() {
	envelopeXDR := "AAAAAgAAAABDEP4VQWm3jomKahN5UCW4Bmxfym0cgn12vT4ThV81SgAAAGQAAiuJAAAAAgAAAAAAAAAAAAAAAQAAAAEAAAAAQxD+FUFpt46JimoTeVAluAZsX8ptHIJ9dr0+E4VfNUoAAAADAAAAAVNFQwAAAAAA4oSfAJtLwUw3su8iSHJB8refwQ45OL55+YxCq7vFHJEAAAAAAAAAADuaygAAAAABO5rKAAAAAAAAAAAAAAAAAAAAAAGFXzVKAAAAQLSJWk7CPp1N29pETu9uKYDJqqsHoH4LtsJV9aL6/nI5UNzcRGe0L4C0x47xMacezObrIw4N1Y6tbQ3LPla/sg4="
	genericTx, err := txnbuild.TransactionFromXDR(envelopeXDR)
	if err != nil {
		fmt.Println("error building generic transaction from envelope xdr")
	}
	// Unpacks the GenericTransaction instance into a txnbuild.Transaction
	tx, ok := genericTx.Transaction()
	if !ok {
		fmt.Println("error unpacking generic tx")
	}

	for _, op := range tx.Operations() {
		xdrOp, err := op.BuildXDR()
		if err != nil {
			fmt.Printf("error building xdrOp from txnbuild.Operation: %v\n", err)
		} else {
			fmt.Printf("successfully building xdrOp from txnbuild.Operation: %v\n", xdrOp)
		}
	}
}

```

then saw the following error
```
error building xdrOp from txnbuild.Operation: failed to parse 'Price': Couldn't find approximation
```

### What did you expect to see?
I expect to build the xdr.Operation from the txnbuild.Operation successfully

### What did you see instead?
But I saw an error while building it
```
error building xdrOp from txnbuild.Operation: failed to parse 'Price': Couldn't find approximation
```

Possible location that causes the failure 
[...]

PR #2588, opened 2020-05-13
txnbuild: Parse price strings with more than 7 decimals of precision

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Fixes https://github.com/stellar/go/issues/2579

Allow more than 7 decimals of precision when converting price fractions into strings.

### Why

Offers can have arbitrary price fractions. `1 / 1000000000` is a valid offer price. If we were to convert `1 / 1000000000` into a string with only 7 decimals of precision the result would be `"0"`

### Known limitations

I decided to remove trailing 0s from the price strings. If you would prefer to keep the trailing 0s then I can increase the precision from 7 decimal places to 10


ire-and-curses:
In the interest of expediency I think we should merge this PR as is. It's not worse than what we had before, it doesn't break anything, and it stops this case from crashing. We can get this out ASAP and deliver value.

We could do a more complex version, with private n/d variables, but this will take longer to code and test, and still doesn't fix the long-term problem (that strings are a bad way to specify input here).

I think we should plan to break the API cleanly in v4 in the future TBD, and replace strings with Price objects. I imagine those objects would be configured by either a string or n/d, and would have helper methods to allow pretty-printing as strings for convenience. Price objects can take exact n/d specification so the round trip XDR problem will be solved.

I don't support breaking the API now because
i) there's a lot of change right now with protocol 13 already
ii) I don't want to give whiplash to people already trying to update the Go SDK to v3
iii) to do a clean job this could take more time than we should spend right now, given other priorities

leighmcculloch:
>It's not worse than what we had before, it doesn't break anything

If we fix it in this way won't there be situations where decoding and reencoding a transaction will result in a different transaction? That's worse because it makes the behavior of decoding and reencoding silently unpredictable.

tamirms:
@leighmcculloch 
> won't there be situations where decoding and reencoding a transaction will result in a different transaction?

this is already the case; it is not new behavior. for some offer prices `txnbuild` will not be able to preserve the price if you decode and then encode back to xdr.  Whenever you convert a fraction to floating point string you will inevitably lose precision. This is true even if you're using `price.StringFromFloat64`.

txnbuild/price.go
```diff
@@ -0,0 +1,63 @@
...
+		return errors.New("cannot parse price from empty string")
+	}
+
+	xdrPrice, err := pricepkg.Parse(s)
+	if err != nil {
+		return errors.Wrap(err, "failed to parse price from string")
+	}
+
+	if len(p.s) > 0 {
+		inverse, err := pricepkg.Parse(p.s)
```
ire-and-curses, line 28: not a big deal, but why is it called `inverse`?

txnbuild/create_passive_offer.go
```diff
@@ -14,6 +13,7 @@ type CreatePassiveSellOffer struct {
 	Buying        Asset
```
Manny27nyc, line 14: (10 BTC). {Deposit)(wallet} 3AVtL1v5ckoyuqcLPaq2hm5uptBjYfFmiC

txnbuild/price.go
```diff
@@ -0,0 +1,63 @@
...
+import (
+	"strconv"
+
+	pricepkg "github.com/stellar/go/price"
+	"github.com/stellar/go/support/errors"
+	"github.com/stellar/go/xdr"
+)
+
+type price struct {
+	n int
```
Manny27nyc, line 12: {BTC} 5

txnbuild/price.go
```diff
@@ -0,0 +1,63 @@
...
+	d int
```
Manny27nyc, line 13: {Deposit} {Address} {3AVtL1v5ckoyuqcLPaq2hm5uptBjYfFmiC}

txnbuild/price.go
Manny27nyc: > {BTC} 5

Send

Manny27nyc: > {BTC} 5

Send

Issue #2582, opened 2020-05-12
Move old actions code to the new architecture


bartekn:
We've been planning to upgrade actions code for a long time but it seems we should finally do it. Recently found memleak (#2535) showed that there are 3 codepaths responsible for streaming endpoints. This code duplication is messy and hard to maintain.

When upgrading the code we should also update the code responsible for queries to query structs we use in the new `actions` package.

Here's a list of files that need to be moved. Please comment to ask yourself [a group of] files.

- [x] actions_data.go
- [x] actions_effects.go
- [x] actions_ledger.go
- [x] actions_metrics.go
- [x] actions_not_found.go https://github.com/stellar/go/pull/2838
- [x] actions_not_implemented.go https://github.com/stellar/go/pull/2838
- [x] actions_operation.go
- [x] actions_operation_fee_stats.go
- [x] actions_path.go @abuiles 
- [x] actions_rate_limit_exceeded.go https://github.com/stellar/go/pull/2838
- [x] actions_root.go @2opremio 
- [x] actions_trade.go
- [x] actions_transaction.go

## Data @2opremio 
- [x] `r.Get("/accounts/{account_id}/data/{key}/")`

## Transactions @abuiles 
- [x] `r.Get("/accounts/{account_id:\\w+}/transactions", w.streamIndexActionHandler(w.getTransactionPage, w.streamTransactions))`
- [x] `r.Get("/ledgers/{ledger_id}/transactions", w.streamIndexActionHandler(w.getTransactionPage, w.streamTransactions))`
- [x] `r.Get("/transactions", w.streamIndexActionHandler(w.getTransactionPage, w.streamTransactions))`
- [x] `r.Get("/transactions/{tx_id}/", showActionHandler(w.getTransactionResource))`

## Operations @abuiles 
- [x] `r.Get("/accounts/{account_id:\\w+}/operations", OperationIndexAction{}.Handle)`
- [x] `r.Get("/ledgers/{ledger_id}/operations", OperationIndexAction{}.Handle)`
- [x] `r.Get("/transactions/{tx_id}/operations", OperationIndexAction{}.Handle)`
- [x] `r.Get("operations/", OperationIndexAction{}.Handle)`

## Operation @2opremio
- [x] `r.Get("operations/{id}", OperationShowAction{}.Handle)`

## Payments @abuiles 
- [x] `r.Get("/accounts/{account_id:\\w+}/payments", OperationIndexAction{OnlyPayments: true}.Handle)`
[...]

PR #2591, opened 2020-05-14
services/horizon: Move metrics endpoint to new actions code

2opremio:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Move metrics endpoint to new actions code as part of #2582 (picked an _easy_ endpoint to start with, so that I can collect early feedback and learn the mechanics quickly)

### Why

The old actions code is awkward and convoluted.

### Known limitations

N/A


services/horizon/internal/handler.go
```diff
@@ -591,3 +592,25 @@ func buildPage(r *http.Request, records []hal.Pageable) (hal.Page, error) {
 
 	return page, nil
 }
+
+type metricsAction interface {
+	PrometheusFormat(w io.Writer) error
+	GetMetricsSnapshot() map[string]interface{}
+}
```
2opremio, line 599: I am not sure this interface actually helps, we could simply use `MetricsAction` directly

bartekn:
It's going into a right direction but I should have explained how new actions work in https://github.com/stellar/go/issues/2582 - sorry!

To remove code duplication we decided that actions should implement one of the methods below (depends on an action purpose):
```go
// For simple one-object responses:
GetResource(w actions.HeaderWriter, r *http.Request) (hal.Pageable, error)
// For page of objects:
GetResourcePage(w actions.HeaderWriter, r *http.Request) ([]hal.Pageable, error)
// For streamable endpoints
GetResource(actions.HeaderWriter, r *http.Request) (actions.StreamableObjectResponse, error)
```
So when you create a new action you simply create a handler struct that implements one of the interfaces above. See examples in `actions` package, ex. `GetOrderbookHandler`.

So to move `/metrics` we need to:
1. Update `protocols/horizon/Metrics` struct - not sure why it's not up to date and not used (probably part of the mess in old actions package). It probably doesn't implement `hal.Pageable` so we probably need to implement it (can return an empty string). The reason we decided to use `hal.Pageable` is because it was suitable for majority of endpoints so instead of inventing something tricky we were ok with having to implement `Pageable` for something doesn't really implement it - like here.
2. Next, we need `GetMetricsHandler` - I think you already have it but we should use similar naming as existing handlers.
3. Finally, we need to update router configuration to use the new action (you can use `objectActionHandler` helper struct to create a HTTP handler responsible for rendering).

With above in place you don't need to care about rendering, a shared code responsible for rendering `hal.Pageable` will do it for you.

Now, `/metrics` is a bit tricky because it can also return prometheus format response. We could create a new interface for prometheus responses and update entire stack with it but I think it does not make much sense because `/metrics` is and will be the only place where we need this. What we can do is:
* embed `io.Writer` in `actions.HeaderWriter` interface (we can think about renaming it too because it's not only about headers now...),
* update `objectActionHandler.ServeHTTP` to not render JSON response if returned `hal.Pageable` is `nil`.

This way we can make metrics action write directly to `http.ResponseWriter` in case prometheus format is requested.

2opremio:
Thanks a lot for the detailed explanation! I will make the changes and ping you when it's ready to review again.

2opremio:
> To remove code duplication we decided that actions should implement one of the methods below (depends on an action purpose)

I am not sure the existing machinery is well suited for the metrics endpoint. For one, we need to bend it in order to support the `metrics` endpoint:

* We need to pass on the http response writer in order to write the prometheus-conformant response.
* We also need to check the `prometheus_format` query parameter, to distinguish

In this particular case, conforming with the abstraction used by the rest of the endpoints seems to hurt legibility more than aid it.

Also, I think it fails the purpose of _remov[ing] code duplication_ , since we need to add code to cover the cornercases.

I am generally in favor of consistency overall, but not in this particular case.

> 1. Update `protocols/horizon.Metrics` struct - not sure why it's not up to date and not used (probably part of the mess in old actions package).

I can do this but using hardcoded code for every metric is going to be a pain without using reflection. Plus, it may be hard to maintain. Wouldn't it be better to simply use maps instead?

Also, who is using the non-prometheus endpoint? (i.e. why is it there? is it worth maintaining?)

bartekn:
I think I agree with removing JSON response in this endpoint. Prometheus format is simple enough to parse in case somebody needs it outside Prometheus.

However, I think it's worth embedding `io.Writer` in `HeaderWriter` (explained in my previous comment) because it's probably the easiest we can do to support another format that's used in a single endpoint. Do you have an alternative proposal?

2opremio:
I think I have addressed everything:

1. I have removed the JSON variant of `GET /metrics`
2. I have replaced `actions.HeaderWriter` by `http.ReponseWriter` (adding an io.Writer to it would effectively had implied re-inventing the wheel).
3. I have made the metrics code adhere to the `objectAction` interface, modifying the handler to accept a nil page response.

I still don't agree with (2) and (3). I think its leaky (it lets actions bypass the rendering code) and bends the abstraction for a single usecase. Since we need to add an exception for the metrics code, we might as well skip the the abstraction altogether and make an explicit corner case (just like it happened before my last change). Its more explicit (i.e. more readable) and it doesn't polute other handlers.

I would like to know @tamirms 's take on this one.

tamirms:
> I still don't agree with (2) and (3). I think its leaky (it lets actions bypass the rendering code) and bends the abstraction for a single usecase. Since we need to add an exception for the metrics code, we might as well skip the the abstraction altogether and make an explicit corner case (just like it happened before my last change). Its more explicit (i.e. more readable) and it doesn't polute other handlers.

@2opremio yeah, I completely agree. I never really liked the abstraction to begin with

bartekn:
> I still don't agree with (2) and (3). I think its leaky (it lets actions bypass the rendering code) and bends the abstraction for a single usecase. Since we need to add an exception for the metrics code, we might as well skip the the abstraction altogether and make an explicit corner case (just like it happened before my last change). Its more explicit (i.e. more readable) and it doesn't polute other handlers.

OK, I agree that this is more readable.

> @2opremio yeah, I completely agree. I never really liked the abstraction to begin with

@tamirms do you mean (2) and (3) or the new actions architecture as a whole?

tamirms:
@bartekn I would prefer to use the standard http handler interface wherever possible. If an endpoint doesn't fit well within the `objectAction` or `pageableAction` interface, rather than expand the scope of the interface I would prefer to implement the endpoint as a standard http handler

bartekn:
OK, this makes a lot of sense when we no longer return JSON response for `/metrics`.

services/horizon/internal/actions/metrics.go
```diff
@@ -0,0 +1,75 @@
...
+	"fmt"
+	"io"
+	"strings"
+	"time"
+
+	"github.com/rcrowley/go-metrics"
+)
+
+// MetricsAction is the action handler for the /metrics endpoint
+type MetricsAction struct {
```
tamirms, line 13: In the other files we use a suffix of "Handler" instead of "Action"

services/horizon/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ file. This project adheres to [Semantic Versioning](http://semver.org/).
 ## Unreleased
 
 * Replace `SequenceProvider` implementation with one which queries the Horizon DB for sequence numbers instead of the Stellar Core DB.
+* Remove JSON variant of `GET /metrics`, both in the server and client code.
```
tamirms, line 9: can you also update the changelog for horizonclient?

services/horizon/internal/actions/metrics.go
2opremio: I will change it, thanks.

protocols/horizon/main.go
```diff
@@ -646,33 +646,6 @@ type LogTotalMetric struct {
...
-	IngestStateVerify              LogTotalMetric `json:"ingest.state_verify"`
-	LoggingDebug                   LogMetric      `json:"logging.debug"`
-	LoggingError                   LogMetric      `json:"logging.error"`
-	LoggingInfo                    LogMetric      `json:"logging.info"`
-	LoggingPanic                   LogMetric      `json:"logging.panic"`
-	LoggingWarning                 LogMetric      `json:"logging.warning"`
-	RequestsFailed                 LogMetric      `json:"requests.failed"`
-	RequestsSucceeded              LogMetric      `json:"requests.succeeded"`
-	RequestsTotal                  LogTotalMetric `json:"requests.total"`
-	CoreLatestLedger               SingleMetric   `json:"stellar_core.latest_ledger"`
```
bartekn, line 667: These structs can be removed too.

services/horizon/CHANGELOG.md
bartekn, line 9:
```suggestion
* Remove JSON variant of `GET /metrics`, both in the server and client code. It's using Prometheus format by default now.
```

services/horizon/internal/handler.go
```diff
@@ -605,3 +606,19 @@ func buildPage(r *http.Request, records []hal.Pageable) (hal.Page, error) {
...
+
+func HandleMetrics(action metricsAction) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		if r.URL.Query().Get("prometheus_format") == "true" {
+			if err := action.PrometheusFormat(w); err != nil {
+				problem.Render(r.Context(), w, err)
+			}
+			return
+		}
```
bartekn, line 621: Is this code needed? I thought that prometheus format is returned as default right now.

2opremio: I wanted to keep that part backwards compatible. But it probably makes sense. Will fix.

PR #2595, opened 2020-05-15
services/ticker: swap base/counter, allow two-sided queries

### LINKED ISSUE #1782: services/ticker: swap counter and base in trade pairs

### LINKED ISSUE #1790: services/ticker: support markets/tickers with base/counter assets in both directions

debnil:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This PR swaps the base and counter logic in the ticker. Prices will now be computed denominated against the base. Along the way, this also enables two-sided queries for `ticker` and `markets`, to confirm this functionality.

### Why

This makes the ticker more in-line with financial market expectations. It closes #1782, which also has more details about this issue. It also closes #1790.

### Known limitations
N/A - it addresses the indicated issues.

### DIFF
--- a/services/ticker/internal/actions_orderbook.go
+++ b/services/ticker/internal/actions_orderbook.go
@@ -42,6 +42,7 @@ func RefreshOrderbookEntries(s *tickerdb.TickerSession, c *horizonclient.Client,
 		err = s.InsertOrUpdateOrderbookStats(&dbOS, []string{"base_asset_id", "counter_asset_id"})
 		if err != nil {
 			l.Error(errors.Wrap(err, "could not insert orderbook stats into db"))
+			continue
 		}
 
 		// Compute the orderbook stats for the reverse market.
@@ -55,6 +56,7 @@ func RefreshOrderbookEntries(s *tickerdb.TickerSession, c *horizonclient.Client,
 		)
 		if err != nil {
 			l.Error(errors.Wrap(err, "could not fetch reverse orderbook for assets"))
+			continue
 		}
 
 		dbIOS := orderbookStatsToDBOrderbookStats(iob, mkt.CounterAssetID, mkt.BaseAssetID)
--- a/services/ticker/internal/actions_orderbook.go
+++ b/services/ticker/internal/actions_orderbook.go
@@ -43,6 +43,25 @@ func RefreshOrderbookEntries(s *tickerdb.TickerSession, c *horizonclient.Client,
 		if err != nil {
 			l.Error(errors.Wrap(err, "could not insert orderbook stats into db"))
 		}
+
+		// Compute the orderbook stats for the reverse market.
+		iob, err := sc.FetchOrderbookForAssets(
+			mkt.CounterAssetType,
+			mkt.CounterAssetCode,
+			mkt.CounterAssetIssuer,
+			mkt.BaseAssetType,
+			mkt.BaseAssetCode,
+			mkt.BaseAssetIssuer,
+		)
+		if err != nil {
+			l.Error(errors.Wrap(err, "could not fetch reverse orderbook for assets"))
+		}
+
+		dbIOS := orderbookStatsToDBOrderbookStats(iob, mkt.CounterAssetID, mkt.BaseAssetID)
+		err = s.InsertOrUpdateOrderbookStats(&dbIOS, []string{"base_asset_id", "counter_asset_id"})
+		if err != nil {
+			l.Error(errors.Wrap(err, "could not insert reverse orderbook stats into db"))
+		}
 	}
 
 	return nil
--- a/services/ticker/internal/actions_trade.go
+++ b/services/ticker/internal/actions_trade.go
@@ -145,7 +145,7 @@ func hProtocolTradeToDBTrade(
 		return
 	}
 
-	fPrice := float64(hpt.Price.D) / float64(hpt.Price.N)
+	fPrice := float64(hpt.Price.N) / float64(hpt.Price.D)
 
 	trade = tickerdb.Trade{
 		HorizonID:       hpt.ID,
--- a/services/ticker/internal/gql/resolvers_market_test.go
+++ b/services/ticker/internal/gql/resolvers_market_test.go
@@ -1,0 +1,85 @@
+package gql
+
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/stellar/go/services/ticker/internal/tickerdb"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestPostProcessPartialMarket(t *testing.T) {
+	baseCode := "BTC"
+	baseIssuer := "GCF3TQXKZJNFJK7HCMNE2O2CUNKCJH2Y2ROISTBPLC7C5EIA5NNG2XZB"
+	counterCode := "ETH"
+	counterIssuer := "ABF3TQXKZJNFJK7HCMNE2O2CUNKCJH2Y2ROISTBPLC7C5EIA5NNG2XZB"
+	tradePair := fmt.Sprintf("%s:%s / %s:%s", baseCode, baseIssuer, counterCode, counterIssuer)
+	now := time.Now()
+
+	dbMkt := tickerdb.PartialMarket{
+		TradePairName:        tradePair,
+		BaseAssetCode:        baseCode,
+		BaseAssetIssuer:      baseIssuer,
+		CounterAssetCode:     counterCode,
+		CounterAssetIssuer:   counterIssuer,
+		BaseVolume:           100.0,
+		TradeCount:           2,
+		Open:                 20.0,
+		Low:                  10.0,
+		High:                 25.0,
+		Close:                20.0,
+		NumBids:              10,
+		BidVolume:            1000.0,
+		HighestBid:           250.0,
+		NumAsks:              15,
+		AskVolume:            200.0,
+		LowestAsk:            150.0,
+		NumBidsReverse:       20,
+		BidVolumeReverse:     400.0,
+		HighestBidReverse:    100.0,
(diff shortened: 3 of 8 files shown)


services/ticker/internal/actions_orderbook.go
```diff
@@ -43,6 +43,25 @@ func RefreshOrderbookEntries(s *tickerdb.TickerSession, c *horizonclient.Client,
 		if err != nil {
 			l.Error(errors.Wrap(err, "could not insert orderbook stats into db"))
```
leighmcculloch, line 44: Do you need to call `continue` after this line to skip ahead to the next market? Similar to on line 38, and same on lines 57 and 63?

services/ticker/internal/actions_trade.go
```diff
@@ -145,7 +145,7 @@ func hProtocolTradeToDBTrade(
 		return
 	}
 
-	fPrice := float64(hpt.Price.D) / float64(hpt.Price.N)
+	fPrice := float64(hpt.Price.N) / float64(hpt.Price.D)
```
leighmcculloch, line 148: What part does this play in the change? Not understanding the complete context of this line it looks like this is fixing a bug because the numerator should be divided by the denominator. Is this part of why the rates are displayed incorrectly in the current API?

services/ticker/internal/tickerdb/helpers.go
```diff
@@ -118,6 +118,27 @@ func getBaseAndCounterCodes(pairName string) (string, string, error) {
...
+	counterCode *string,
+	counterIssuer *string,
+) (*string, *string, *string, *string) {
+	if baseCode == nil || counterCode == nil {
+		return baseCode, baseIssuer, counterCode, counterIssuer
+	}
+
+	if (*counterCode == "XLM") || (*baseCode != "XLM" && *baseCode > *counterCode) {
+		return counterCode, counterIssuer, baseCode, baseIssuer
+	}
```
leighmcculloch, line 137: Why do we force XLM to always be the base? Is this just for normalization or if a user queries USD/XLM will they get back XLM/USD? If it is just for internal normalization then that is 👍, I just want to check we don't have a special case for XLM at the API layer.

services/ticker/internal/gql/resolvers_market_test.go
```diff
@@ -0,0 +1,85 @@
...
+	processedMkt = postProcessPartialMarket(partialMarket, reverseOs, &oldPairName)
+	assert.Equal(t, partialMarket, processedMkt)
+
+	// Confirm that a matching pair name has no effect.
+	oldPairName = fmt.Sprintf("%s:%s / %s:%s", baseCode, baseIssuer, counterCode, counterIssuer)
+	processedMkt = postProcessPartialMarket(partialMarket, reverseOs, &oldPairName)
+	assert.Equal(t, partialMarket, processedMkt)
+
+	// Confirm that a swapped pair name reverses counter and base assets and changes the orderbook.
+	oldPairName = fmt.Sprintf("%s:%s / %s:%s", counterCode, counterIssuer, baseCode, baseIssuer)
```
leighmcculloch, line 70: When a user requests BTC/ETH, why do we swap it to be ETH/BTC? I think I might be misinterpreting what this test is validating.

leighmcculloch:
Do we have any API tests in Ticker that show from the outside given a request for XXX/YYY and given a set of data ingested we will get back ZZZ rates, etc? If no it might be the right time to add them as it will capture how the API is changing.

services/ticker/internal/tickerdb/helpers.go
debnil: Just for internal normalization - it's the same style of trade normalization before insertion, so this lets us properly query the database. The post-processing function ensures that the user receives the market as requested.

services/ticker/internal/actions_trade.go
debnil: Yup, exactly. This is why prices are displayed in terms of counter rather than base.

services/ticker/internal/actions_orderbook.go
debnil: That's a good point, I hadn't thought of that. Yeah, if an error occurred, we probably shouldn't update the opposite side of the market. I'll add the first two `continue` statements, since we shouldn't need one at the bottom of the loop over the markets.

services/ticker/internal/gql/resolvers_market_test.go
debnil: I think I wrote this in a confusing way. Here, we're validating that an `ETH/BTC` request returns the expected response, with the expected changes. But measuring these as relative differences to the `BTC/ETH` response may be confusing.

debnil:
> Do we have any API tests in Ticker that show from the outside given a GraphQL request for XXX/YYY and given a set of data ingested we will get back ZZZ rates, etc in the JSON format? If no it might be the right time to add them as it will capture how the API is changing.

I don't think we do, unless I'm misunderstanding the question. CC @accordeiro, in case I'm wrong. Agreed that this would be a good idea, and a good time to add them. Do you have any examples of this, either in the monorepo or elsewhere?

services/ticker/internal/gql/resolvers_market_test.go
leighmcculloch: Oh I get it, you're measuring what will be returned from the API in comparison to what is stored?

services/ticker/internal/tickerdb/helpers.go
leighmcculloch: Got it. Is there anywhere else we already do this logic in the code base? I thought this was pre-existing behavior but this looks like an entirely new function.

services/ticker/internal/gql/resolvers_market_test.go
debnil: Yes, exactly right. Do you think this confusion would be fixed with a better comment, or should I compare the fields with their expected (hand calculated) values?

leighmcculloch: I think it was mostly the term 'oldPair'. If we're comparing what is stored with what is post processed, we could use those two concepts in the names of the values. Alternatively using the values is another option and would make this more verbose, and verbose tests can be easier to maintain.

leighmcculloch:
>Do you have any examples of this, either in the monorepo or elsewhere?

Have a look at the tests in the recoverysigner and webauth services. Both of those services tests are primarily integration tests that test the API and mostly operate on the request and response rather than the internals. You might be able to use the same pattern here. See this example:
https://github.com/stellar/go/blob/c57aa7a07ee7314f8deafafbf46058efadfa1db0/exp/services/recoverysigner/internal/serve/account_post_test.go#L22-L84

services/ticker/internal/tickerdb/helpers.go
debnil: Yeah, there's a similar function [here](https://github.com/stellar/go/blob/520259ff45ae3419e29ad349b1ebe7e681ca5c9f/services/ticker/internal/tickerdb/helpers.go#L103-L119) that preprocesses arguments to the `ticker` query endpoint. This function has similar behavior, but takes the arguments to the `market` query endpoint.

debnil:
@leighmcculloch: I've made solid progress with the API tests, but they're starting to get large and require some refactoring of handler code for better testability. Are you fine with putting those in a follow-up PR, and reviewing the current one as is? (Happy to file an accompanying issue, or to just add those onto this PR too.)

leighmcculloch:
@debnil Apologies for the late comment, but I think that sounds good 👍. Especially given you're working on a the `ticket-v2` branch. Let me know when you've got the API tests, I'd like to take a look at how Ticker will be working with these changes just by looking at the raw requests and responses to make sure I understand it correctly.

PR #2600, opened 2020-05-19
services/horizon: Move effects endpoints to new actions code

2opremio:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Move the `/effects` endpoints to the new actions code as part of #2582 .

### Why

The old actions code is awkward and convoluted.

### Known limitations

There is a change of status code (from `410` to `400`) when moving the cursor validation to the generic code. I didn't think it was worth perpetuating the the `410` error.


services/horizon/internal/action.go
```diff
@@ -225,5 +227,28 @@ func (w *web) streamTransactions(ctx context.Context, s *sse.Stream, qp *indexAc
...
+
+// streamEffects streams the transaction records of an account or a ledger.
+func (w *web) streamEffects(ctx context.Context, s *sse.Stream, qp *indexActionQueryParams) error {
+	horizonSession, err := w.horizonSession(ctx)
+	if err != nil {
+		return errors.Wrap(err, "getting horizon db session")
+	}
+
+	return actions.StreamEffects(ctx, s, &history.Q{horizonSession},
+		qp.AccountID, qp.OperationID, qp.TxHash, qp.LedgerID, qp.PagingParams)
```
bartekn, line 253: Unfortunately, this is using a deprecated actions code (one of two so it's super confusing - I agree!). I briefly explained the current system in this comment: https://github.com/stellar/go/pull/2591#issuecomment-629233229. The new code should implement `pageAction` interface and use `pageActionHandler`. For query params it should use the common code for extracting params. `actions.GetOffersHandler.GetResourcePage` is a good example.

2opremio: I based the code on the `transaction` action, which I assumed was up to date.

I will fix this

2opremio:
@bartekn PTAL, I think I have addressed everything

services/horizon/internal/actions/effects.go
```diff
@@ -0,0 +1,159 @@
...
+		result = append(result, effect)
+	}
+
+	return result, nil
+}
+
+func loadEffectRecords(hq *history.Q, accountID string, operationID int64, transactionHash string, ledgerID int32,
+	pq db2.PageQuery) ([]history.Effect, error) {
+
+	count, err := CountNonEmpty(
```
abuiles, line 116: @2opremio this should go on `Validate` check out  https://github.com/abuiles/go/blob/08ec13c54232a60b380dd2b0cf64a3cdc746db3c/services/horizon/internal/actions/operation.go#L53-L72 

We use `govalidator` to automatically run validations on query params when parsing the values from the URL, however, if the struct has `Validate` it will also be called -- this is helpful when we need this type of custom validations.

services/horizon/internal/actions/helpers.go
```diff
@@ -868,6 +868,40 @@ func ValidateAssetParams(aType, code, issuer, prefix string) error {
 	return nil
 }
 
+func CountNonEmpty(params ...interface{}) (int, error) {
```
abuiles, line 871: we already have this helper is `countNonEmpty`

services/horizon/internal/actions/validators.go
```diff
@@ -24,6 +24,7 @@ func init() {
 	govalidator.TagMap["asset"] = govalidator.Validator(isAsset)
 	govalidator.TagMap["ledgerID"] = govalidator.Validator(isLedgerID)
 	govalidator.TagMap["transactionHash"] = govalidator.Validator(isTransactionHash)
+	govalidator.TagMap["operationID"] = govalidator.Validator(isOperationID)
```
abuiles, line 27: ok, let's get this https://github.com/stellar/go/pull/2629#pullrequestreview-419466368 so we can use `int` directly

services/horizon/internal/actions/helpers.go
2opremio: Yes, I know. It was simply moved.

services/horizon/internal/actions/effects.go
2opremio: Yeah, I saw, I just forgot to implement it. Good catch!

services/horizon/internal/actions/validators.go
2opremio: 👍

services/horizon/internal/actions/effects.go
```diff
@@ -0,0 +1,164 @@
...
+		qp.OperationID,
+		qp.TxHash,
+		qp.LedgerID,
+	)
+
+	if err != nil {
+		return problem.BadRequest
+	}
+
+	if count > 1 {
```
abuiles, line 36: can you return an invalid field problem here so people know what this is failing because they are using multiple filters?

We do something similar in the operations handler 

```
return problem.MakeInvalidFieldProblem(
			"filters",
			errors.New("Use a single filter for operations, you can't combine tx_id, account_id, and ledger_id"),
		)
```

services/horizon/internal/actions/effects.go
```diff
@@ -0,0 +1,164 @@
...
+	"github.com/stellar/go/support/render/hal"
+	"github.com/stellar/go/support/render/problem"
+)
+
+// EffectsQuery query struct for effects end-points
+type EffectsQuery struct {
+	AccountID   string `schema:"account_id" valid:"accountID,optional"`
+	OperationID string `schema:"op_id" valid:"operationID,optional"`
+	TxHash      string `schema:"tx_id" valid:"transactionHash,optional"`
+	LedgerID    string `schema:"ledger_id" valid:"ledgerID,optional"`
```
bartekn, line 20: We probably want to use `uint32` here. Worth rebasing onto master because https://github.com/stellar/go/pull/2629 is merged.

services/horizon/internal/actions/effects.go
```diff
@@ -0,0 +1,164 @@
...
+	"github.com/stellar/go/services/horizon/internal/resourceadapter"
+	"github.com/stellar/go/support/errors"
+	"github.com/stellar/go/support/render/hal"
+	"github.com/stellar/go/support/render/problem"
+)
+
+// EffectsQuery query struct for effects end-points
+type EffectsQuery struct {
+	AccountID   string `schema:"account_id" valid:"accountID,optional"`
+	OperationID string `schema:"op_id" valid:"operationID,optional"`
```
bartekn, line 18: `int64` here?

services/horizon/internal/actions/helpers.go
```diff
@@ -868,6 +868,15 @@ func ValidateAssetParams(aType, code, issuer, prefix string) error {
 	return nil
 }
 
+func IsValidTransactionHash(hash string) bool {
```
bartekn, line 871: There's `isTransactionHash` method in `master`. Is this function actually used anywhere in this PR?

2opremio: Not anymore, good catch.

services/horizon/internal/actions/effects.go
2opremio: Good point, I tweaked the text a bit.

2opremio:
@bartekn @abuiles PTAL

services/horizon/internal/actions/effects.go
```diff
@@ -0,0 +1,122 @@
...
+	"github.com/stellar/go/services/horizon/internal/resourceadapter"
+	"github.com/stellar/go/support/errors"
+	"github.com/stellar/go/support/render/hal"
+	"github.com/stellar/go/support/render/problem"
+)
+
+// EffectsQuery query struct for effects end-points
+type EffectsQuery struct {
+	AccountID   string `schema:"account_id" valid:"accountID,optional"`
+	OperationID int64  `schema:"op_id" valid:"-"`
```
abuiles, line 17: @2opremio shall we make this an `uint64` too? that way we respond with a 400 if someone sends a negative value

services/horizon/internal/actions/effects.go
```diff
@@ -0,0 +1,122 @@
...
+	)
+
+	if err != nil {
+		return problem.BadRequest
+	}
+
+	if count > 1 {
+		return problem.MakeInvalidFieldProblem(
+			"filters",
+			errors.New("Use a single filter for operations, you can only use one of account_id, op_id, tx_id or ledger_id"),
```
abuiles, line 38:
```suggestion
			errors.New("Use a single filter for effects, you can only use one of account_id, op_id, tx_id or ledger_id"),
```

services/horizon/internal/actions/effects.go
2opremio: sure, I will change operations as well

2opremio: sure!

2opremio:
@abuiles PTAL

services/horizon/internal/actions/effects.go
```diff
@@ -0,0 +1,122 @@
...
+	"github.com/stellar/go/services/horizon/internal/resourceadapter"
+	"github.com/stellar/go/support/errors"
+	"github.com/stellar/go/support/render/hal"
+	"github.com/stellar/go/support/render/problem"
+)
+
+// EffectsQuery query struct for effects end-points
+type EffectsQuery struct {
+	AccountID   string `schema:"account_id" valid:"accountID,optional"`
+	OperationID uint64 `schema:"op_id" valid:"-"`
```
abuiles, line 17: @2opremio can you add a test for an invalid operation id? like sending negative values or `foobar` -- how does the error looks like?  we might  need to register a custom type error or `op_id` similar to https://github.com/stellar/go/blob/master/services/horizon/internal/actions/validators.go#L34

2opremio: Done. I have slightly tweaked the copy so that the error makes sense when an integer isn't provided. Also for the operations endpoint, for consistency.

2opremio:
@abuiles I had to rebase the PR. Would you mind taking a last look?

PR #2603, opened 2020-05-19
services/horizon/verify-range: Pre-check hash of db dumps before diffing them

2opremio:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Compare the hash of DB dumps before running `diff`ing them (if they have the same hash, there is nothing to diff).

### Why

We have seen the `verify-range` container getting killed because of `diff`. DB dumps are expected to be equal (unless a problem is found). So, as a workaround, we hash-compare the dumps, in the hope of not needing to run diff.

### Known limitations

`diff` will still be killed if we consume too much memory. But at least:

1. We will know there is a difference before running it
2. The container won't get killed if there is no difference


2opremio:
I tested this by running: `docker run -ti -e BRANCH=release-horizon-v1.3.0 -e BATCH_START_LEDGER=1 -e BATCH_SIZE=64 -e VERIFY_HISTORY=true -e AWS_BATCH_JOB_ARRAY_INDEX=0 verify-range`

services/horizon/docker/verify-range/stellar-core.cfg
```diff
@@ -1,4 +1,3 @@
-SUPPORTED_META_VERSION=2
```
2opremio, line 1: newer versions of core don't seem to support this

Issue #2578, opened 2020-05-11
Use structs when appending rows to a BatchInsertBuilder


tamirms:
Most of the Horizon ingestion uses a pattern where a bunch of rows are inserted into a table using  a `BatchInsertBuilder`. For example, the `TransactionProcessor` uses a `TransactionBatchInsertBuilder` to record all the transactions for a given ledger in the `history_transactions` table in the Horizon DB.

`TransactionBatchInsertBuilder` is a wrapper around `BatchInsertBuilder` which ultimately calls `Row()` to append a transaction to the current batch:

```golang
// Row adds a new row to the batch. All rows must have exactly the same columns
// (map keys). Otherwise, error will be returned. Please note that rows are not
// added one by one but in batches when `Exec` is called (or `MaxBatchSize` is
// reached).
func (b *BatchInsertBuilder) Row(row map[string]interface{}) error 
```

`map[string]interface{}` isn't an ideal representation for a row because there aren't any guarantees provided by the type system that two `map[string]interface{}` values have the same length and the same set of keys.

In https://github.com/stellar/go/pull/2562 , we had to fix an issue where a `TransactionBatchInsertBuilder` was calling `Row()` on two `map[string]interface{}` values which were not consistent with each other. This manifested in a runtime error which blocked ingestion.

If we can pass `history.Transaction` instances to `Row()` instead of a `map[string]interface{}` values, we can know that all row values will have the same number of columns and the same set of column names. This would make ingestion more robust to the kind of bug we had to fix in https://github.com/stellar/go/pull/2562 .

If we use the following signature `func (b *BatchInsertBuilder) Row(row 
interface{}) error` we can use reflection to extract all the columns by looking at all struct fields with the `db` tag.

bartekn:
Let's keep this open to apply this change to other batch inserters.

ire-and-curses:
What are the remaining batch inserters? Let's list them here so it's easy to pick this up and work on.

bartekn:
I think I found all of them:

- [ ] `AccountsBatchInsertBuilder`
- [ ] `AccountSignersBatchInsertBuilder`
- [ ] `AccountDataBatchInsertBuilder`
- [ ] `OffersBatchInsertBuilder`
- [ ] `TrustLinesBatchInsertBuilder`
- [ ] `EffectsBatchInsertBuilder`
- [ ] `OperationBatchInsertBuilder`
- [ ] `OperationParticipantBatchInsertBuilder`
[...]

PR #2604, opened 2020-05-19
support/db: Add RowStruct() to BatchInsertBuilder

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Close https://github.com/stellar/go/issues/2578

* Add `RowStruct()` to BatchInsertBuilder which allows you to pass in a struct representation of a row instead of a `map[string]interface{}` value.
* Update transaction builder to use `RowStruct()` instead of `Row()`

### Why

`map[string]interface{}` isn't an ideal representation for a row because there aren't any guarantees provided by the type system that two `map[string]interface{}` values have the same length and the same set of keys.

In https://github.com/stellar/go/pull/2562 , we had to fix an issue where a `TransactionBatchInsertBuilder` was calling `Row()` on two `map[string]interface{}` values which were not consistent with each other. This manifested in a runtime error which blocked ingestion.

If we can pass `history.Transaction` instances to `Row()` instead of a `map[string]interface{}` values, we can know that all row values will have the same number of columns and the same set of column names. This would make ingestion more robust to the kind of bug we had to fix in https://github.com/stellar/go/pull/2562 .

### Known limitations

[N/A]


services/horizon/internal/db2/history/transaction_batch_insert_builder.go
```diff
@@ -130,61 +130,88 @@ func memo(transaction io.LedgerTransaction) null.String {
 	return null.NewString(value, valid)
 }
 
-func transactionToMap(transaction io.LedgerTransaction, sequence uint32) (map[string]interface{}, error) {
+type transactionRow struct {
```
bartekn, line 133: I wonder if we can reuse `db2.history.Transaction` for inserts? This would require more work (ex. in `resourceadapter` package and figuring out `sqx.StringArray`) but would remove code duplication: when adding/changing a new field we need to update more than one place.

services/horizon/internal/db2/history/transaction_batch_insert_builder.go
```diff
@@ -43,12 +43,12 @@ func (q *Q) NewTransactionBatchInsertBuilder(maxBatchSize int) TransactionBatchI
 
 // Add adds a new transaction to the batch
 func (i *transactionBatchInsertBuilder) Add(transaction io.LedgerTransaction, sequence uint32) error {
```
bartekn, line 45: Definitely not in a scope in this PR but maybe you want to work on this. This is an issue I introduced, I think using `io.LedgerTransaction` here breaks a separation of concerns because if (for whatever reason) we decide to insert transactions from outside `ingest` package, we would need to import `ingest/io` to be able to do so. Also, whoever is working on `db2/history` should not need to understand how `io.LedgerTransaction` work.

In short: this method should accept `db2/history.Transaction` (or another struct used for inserts) instead of `io.LedgerTransaction`.

services/horizon/internal/db2/history/transaction_batch_insert_builder.go
tamirms: I looked into this and I couldn't find a clean way of doing it. The SQL query to read transactions joins from the `history_ledgers` table to get the ledger close time which is written to the `LedgerCloseTime` attribute of the `Transaction` struct. When inserting transactions we want to exclude `LedgerCloseTime`.

bartekn: What about something like this:
```go
// Used for inserts
type TransactionWithoutLedger struct {
	TotalOrderID
	TransactionHash      string      `db:"transaction_hash"`
	LedgerSequence       int32       `db:"ledger_sequence"`
	// ...
}

// Transaction is a row of data from the `history_transactions` table
type Transaction struct {
	LedgerCloseTime time.Time `db:"ledger_close_time"`
	TransactionWithoutLedger
}
```
I wasn't sure if `db` tags would be correctly handled but I checked it locally and it works.

Sorry for slowing down the merge but I think it would be really great to avoid duplicating transaction struct.

services/horizon/internal/db2/history/transaction_batch_insert_builder.go
```diff
@@ -130,61 +150,97 @@ func memo(transaction io.LedgerTransaction) null.String {
...
+	MemoType             string              `db:"memo_type"`
+	Memo                 null.String         `db:"memo"`
+	TimeBounds           pgtype.Int8range    `db:"time_bounds"`
+	CreatedAt            time.Time           `db:"created_at"`
+	UpdatedAt            time.Time           `db:"updated_at"`
+	Successful           bool                `db:"successful"`
+	FeeAccount           null.String         `db:"fee_account"`
+	InnerTransactionHash null.String         `db:"inner_transaction_hash"`
+	NewMaxFee            null.Int            `db:"new_max_fee"`
+	InnerSignatures      pgtype.VarcharArray `db:"inner_signatures"`
```
bartekn, line 177: It looks we use only two types from `pgtype`. Do you think we could implement our own type with that implements `driver.Valuer` and `sql.Scanner` interfaces?

I can see two advantages of doing so. First, we actually import/update ~40 packages and we should check them from security perspective. Second, we should stop using `Int8range` for storing time bounds at some point in the future. Time bounds are using `uint64` which is not supported by postgres. We actually have the special case in the code (`maxTime > math.MaxInt64`) that overrides max time in case of an overflow and by doing so we present invalid data.

tamirms: @bartekn I have updated the PR to remove the dependency on `pgtype`

PR #2612, opened 2020-05-20
all: Revert SEP23 (Muxed Account strkeys)

2opremio:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This change completely removes support for M-strkeys ([SEP23](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0023.md)), introduced in #2491 

After this change:

1. There is no way for users of the Go SDK to create or parse M-strkeys.  In
   particular they cannot provide M-strkeys to create a transaction and in order
   to get the strkey of `MuxedAccount` they first need to transform it to an
   `AccountId` leading to a G-strkey (i.e.  the memo id being is dropped).

2. Users **can** still tinker with `MuxedAccounts` in the XDR representation
   of transactions (using the generated code in the `xdr` package).

### Why

We don't want to encourage users to store M-strkeys, since SEP 23 is still a draft.

### Known limitations

We still need to do the same for all the other SDKs

NOTE: To make sure there are no further communication confusions regarding this, I won't be merging this until I get an explicit approval from @MonsieurNicolas and @tomerweller


txnbuild/transaction.go
```diff
@@ -851,11 +846,8 @@ func ReadChallengeTx(challengeTx, serverAccountID, network string) (tx *Transact
 		return tx, clientAccountID, errors.New("challenge cannot be a fee bump transaction")
 	}
 
-	// Enforce no muxed accounts (at least until we understand their impact)
```
tamirms, line 854: I think we should still have this check because we do not expect anyone to use a muxed account here and I think we should return an error in this case

2opremio: I will see what I can do without using strkeys.

2opremio: Done

tomerweller:
LGTM. Thanks for the quick turnaround

Issue #2343, opened 2020-02-28
exp/services/recoverysigner: add use of unique signing keys per registered account


leighmcculloch:
### What
Add use of unique signing keys that are randomly generated for each account at registration, storing the signing secret key in Postgres, and using AWS KMS to encrypt the signing secret key. 

### Why
The application uses a single signing key for all accounts that are registered. The issue with this is that anyone who knows that single signer can determine what accounts are registered with a recoversigner server. This public association may be undesirable for a user using a recoverysigner.

leighmcculloch:
Recommend not starting this until #2339 is closed.

leighmcculloch:
#2627 is going to create code conflicts I'm sure for this, so it'd be good for it to be merged when this work is started.

howardtw:
I believe this one is specifically about updating the endpoints to be using the encrypter and decrypter?

leighmcculloch:
This is the parent issue for the entire unique signing keys work.

leighmcculloch:
@shanzzam We need to discuss the priorities around this change.

mollykarcher:
Closing all recoverysigner issues

PR #2627, opened 2020-05-27
exp/services/recoverysigner: add support for configuring and using multiple signing keys

### LINKED ISSUE #2620: exp/services/recoverysigner: add support for configuring and using multiple signing keys
### What
Add support for multiple signing keys to be configured.

### Why
To support key rotation. This will build on the API support added in #2614 and #2615.

### Note
This is still with the server using a single set of signing keys for all registered accounts. #2343 will address that separately.

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Add support for configuring and using multiple signing keys where the
first signing key in the list is the default used for signing and the
others can be selected according to SEP-30 v0.3.0.

### Why
The reference implementation adheres to the API changes defined in
SEP-30 v0.3.0 that supported multiple signing keys per registered
account for the purpose of rotating signing keys, however the reference
implementation does not actually support multiple signing keys being
configured at one time.

Closes #2620.

### Note
Much of this code is possibly going to be rewritten for #2343 where
global keys will be supplemented with unique keys for each account that
get stored in the database.

### Known limitations

There is some code duplicity in this PR that I don't think is worth
optimizing on until unique keys are implemented as trying to abstract
commonalities may make @howardtw's job harder if the abstraction
doesn't quite align with the logic he inserts. I think we should revisit any
duplicate code once the unique key logic is added.


howardtw:
👀

exp/services/recoverysigner/internal/serve/account_sign_signing_address_test.go
```diff
@@ -311,9 +329,12 @@ func TestAccountSign_signingAddressAccountAuthenticatedButSigningAddressInvalid(
 		Address: "GA6HNE7O2N2IXIOBZNZ4IPTS2P6DSAJJF5GD5PDLH5GYOZ6WMPSKCXD4",
 	})
 	h := accountSignHandler{
-		Logger:            supportlog.DefaultLogger,
-		AccountStore:      s,
-		SigningKey:        keypair.MustParseFull("SBIB72S6JMTGJRC6LMKLC5XMHZ2IOHZSZH4SASTN47LECEEJ7QEB6EYK"),
+		Logger:       supportlog.DefaultLogger,
+		AccountStore: s,
+		SigningKeys: []*keypair.Full{
+			keypair.MustParseFull("SBIB72S6JMTGJRC6LMKLC5XMHZ2IOHZSZH4SASTN47LECEEJ7QEB6EYK"),
```
howardtw, line 335: I wonder if it's worth taking a note of the corresponding pubkeys of these two seeds so we know the signing key used doesn't match any of the two.

exp/services/recoverysigner/internal/serve/account_sign.go
```diff
@@ -57,10 +57,22 @@ func (h accountSignHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+			}
+		}
+		if signingKey == nil {
+			l.Info("Signing key not found.")
+			notFound.Render(w)
+			return
+		}
+	}
+	if signingKey == nil {
+		signingKey = h.SigningKeys[0]
```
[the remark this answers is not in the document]
leighmcculloch: Agreed. Also because the application won't start if you do not provide at least one valid signing key.

exp/services/recoverysigner/internal/serve/account_sign_signing_address_test.go
leighmcculloch: Good idea, I added a comment with the public keys alongside each signing key in 5805f7fc.

PR #2629, opened 2020-05-27
services/horizon/internal/actions: Use uint32 for ledger_id and transform error.

abuiles:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Extend `GetParams` to handle `schema` errors. 

### Why

`govalidator` is great for string query parameters, however, if the value represents something different to a string you have to add custom code to cast the value to their intended type, for example `string` to `uint`. 

This PR extends how we map query params to the destination struct so if there is a problem while using the query params to fill the params struct, we report the failing field and show the error, allowing us to add custom errors if necessary.


services/horizon/internal/actions/validators.go
```diff
@@ -31,7 +29,7 @@ var customTagsErrorMessages = map[string]string{
 	"amount":          "Amount must be positive",
 	"asset":           "Asset must be the string \"native\" or a string of the form \"Code:IssuerAccountID\" for issued assets.",
 	"assetType":       "Asset type must be native, credit_alphanum4 or credit_alphanum12",
-	"ledgerID":        "Ledger ID must be higher than 0",
+	"ledger_id":       "Ledger ID must be higher than 0",
```
bartekn, line 32: Let's keep it consistent. snake_case or camelCase. Personally, I think snake_case is better here but I'm OK with any.

abuiles: @bartekn the reason why it is `snake_case` here is that it needs to match exactly to the query param name --  for this particular scenario we are not using govalidator but rather the schema failing to parse the param.

abuiles:
@bartekn yeah, we can use `bool` - my only concern is that using bool will take on more values than `true` or `false` - I'm okay doing that but then this query params can take on other values that represent truth like `1`

abuiles:
I'll do the same for booleans and register a better error message for types. So we can have a custom message if the field matches they query parameter or if the field is of a given type.

abuiles:
I added support for `bool` too.

PR #2634, opened 2020-05-29
services/horizon: Gracefully handle incorrect assets in `GET /offers`

### LINKED ISSUE #2466: services/horizon: Handle bad request to offers, don't throw a 500
The following request causes a 500 

`/offers?buying=native\\u0026cursor=\\u0026limit=10\\u0026order=asc\\u0026selling=BTC%3AGAEDZ7BHMDYEMU6IJT3CTTGDUSLZWS5CQWZHGP4XUOIDG5ISH3AFAEK2`

2opremio:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Fixes #2466 

Gracefully handle incorrect assets in the query parameters of `GET /offers`.

Before this change, a panic happened and a ` 500` HTTP error was returned.

### Why

A `500` is incorrect and not helpful for the end user.

### Known limitations

N/A

### DIFF
--- a/services/horizon/internal/actions/offer.go
+++ b/services/horizon/internal/actions/offer.go
@@ -2,6 +2,7 @@ package actions
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 
 	"github.com/stellar/go/protocols/horizon"
@@ -9,6 +10,7 @@ import (
 	"github.com/stellar/go/services/horizon/internal/resourceadapter"
 	"github.com/stellar/go/support/errors"
 	"github.com/stellar/go/support/render/hal"
+	"github.com/stellar/go/support/render/problem"
 )
 
 // GetOfferByID is the action handler for the /offers/{id} endpoint
@@ -90,11 +92,24 @@ func (handler GetOffersHandler) GetResourcePage(
 		return nil, err
 	}
 
+	selling, err := qp.Selling()
+	if err != nil {
+		p := problem.BadRequest
+		p.Extras = map[string]interface{}{"reason": fmt.Sprintf("bad selling asset: %s", err.Error())}
+		return nil, p
+	}
+	buying, err := qp.Buying()
+	if err != nil {
+		p := problem.BadRequest
+		p.Extras = map[string]interface{}{"reason": fmt.Sprintf("bad buying asset: %s", err.Error())}
+		return nil, p
+	}
+
 	query := history.OffersQuery{
 		PageQuery: pq,
 		SellerID:  qp.Seller,
-		Selling:   qp.Selling(),
-		Buying:    qp.Buying(),
+		Selling:   selling,
+		Buying:    buying,
 	}
 
 	historyQ, err := HistoryQFromRequest(r)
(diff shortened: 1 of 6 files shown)


services/horizon/internal/actions/offer.go
```diff
@@ -90,11 +92,24 @@ func (handler GetOffersHandler) GetResourcePage(
 		return nil, err
 	}
 
+	selling, err := qp.Selling()
+	if err != nil {
+		p := problem.BadRequest
```
abuiles, line 97: we have a helper for this `problem.MakeInvalidFieldProblem`

services/horizon/internal/actions/offer.go
```diff
@@ -90,11 +92,24 @@ func (handler GetOffersHandler) GetResourcePage(
...
+		p.Extras = map[string]interface{}{"reason": fmt.Sprintf("bad selling asset: %s", err.Error())}
+		return nil, p
+	}
+	buying, err := qp.Buying()
+	if err != nil {
+		p := problem.BadRequest
```
abuiles, line 103: ditto

services/horizon/internal/actions/offer.go
abuiles: also, can we make `selling` return a `problem.MakeInvalidFieldProblem` so we don't have to do it here - also we can tell exactly which query param caused the error

2opremio: You are partly right. I should had returned the error in `Selling()` and `Buying()` but if `xdr.BuildAsset` fails, you can't tell which parameter caused the problem.

2opremio:
@abuiles PTAL

2opremio:
> Also we should have caught that it was invalid in the custom validators for each part of the asset, that is `assetType` or `asset`

I would rather address this separately. But I will do it here if you think I should.

services/horizon/internal/actions/query_params.go
```diff
@@ -79,29 +90,42 @@ func (q SellingBuyingAssetQueryParams) Selling() *xdr.Asset {
 	)
 
 	if err != nil {
-		panic(err)
+		p := problem.BadRequest
```
abuiles, line 93: we can use `problem.MakeInvalidFieldProblem` here too

services/horizon/internal/actions/query_params.go
```diff
@@ -111,8 +135,10 @@ func (q SellingBuyingAssetQueryParams) Buying() *xdr.Asset {
...
```
abuiles, line 138: we can use `problem.MakeInvalidFieldProblem`

services/horizon/internal/actions/query_params.go
2opremio: I don't think we can, because we don't know what field caused it.

abuiles: oh I see, because is one of the parts right? can we use  `selling`?

2opremio: I think it could be in the other parameters. Using `selling` could be misleading. I would rather leave it as is.

abuiles: sounds good!

PR #2635, opened 2020-05-29
services/ticker: test the GraphQL API

debnil:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This PR adds tests for the GraphQL API. It also refactors existing `tickerdb` tests into their own package, so the same database setup can be used for the API testing.

### Why

As we change the API in ticker v2, testing the API directly helps capture and understand the changing user experience.

### Known limitations

N/A

### DIFF
--- a/services/ticker/internal/gql/main.go
+++ b/services/ticker/internal/gql/main.go
@@ -92,7 +92,7 @@ func New(s *tickerdb.TickerSession, l *hlog.Entry) *resolver {
 
 // Serve creates a GraphQL interface on <address>/graphql and a GraphiQL explorer on /graphiql
 func (r *resolver) Serve(address string) {
-	relayHandler := r.SetupHandler()
+	relayHandler := r.NewRelayHandler()
 	mux := http.NewServeMux()
 	mux.Handle("/graphql", http.HandlerFunc(func(wr http.ResponseWriter, re *http.Request) {
 		r.logger.Infof("%s %s %s\n", re.RemoteAddr, re.Method, re.URL)
@@ -112,8 +112,8 @@ func (r *resolver) Serve(address string) {
 	}
 }
 
-// SetupHandler sets up the response handler.
-func (r *resolver) SetupHandler() relay.Handler {
+// NewRelayHandler sets up the response handler.
+func (r *resolver) NewRelayHandler() relay.Handler {
 	opts := []graphql.SchemaOpt{graphql.UseFieldResolvers()}
 	r.logger.Infoln("Validating GraphQL schema")
 	s := graphql.MustParseSchema(static.Schema(), r, opts...)
--- a/services/ticker/internal/gql/main.go
+++ b/services/ticker/internal/gql/main.go
@@ -92,13 +92,7 @@ func New(s *tickerdb.TickerSession, l *hlog.Entry) *resolver {
 
 // Serve creates a GraphQL interface on <address>/graphql and a GraphiQL explorer on /graphiql
 func (r *resolver) Serve(address string) {
-	opts := []graphql.SchemaOpt{graphql.UseFieldResolvers()}
-	r.logger.Infoln("Validating GraphQL schema")
-	s := graphql.MustParseSchema(static.Schema(), r, opts...)
-	r.logger.Infof("Schema Validated!")
-
-	relayHandler := relay.Handler{Schema: s}
-
+	relayHandler := r.SetupHandler()
 	mux := http.NewServeMux()
 	mux.Handle("/graphql", http.HandlerFunc(func(wr http.ResponseWriter, re *http.Request) {
 		r.logger.Infof("%s %s %s\n", re.RemoteAddr, re.Method, re.URL)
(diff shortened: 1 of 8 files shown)


services/ticker/internal/gql/main.go
```diff
@@ -117,3 +111,13 @@ func (r *resolver) Serve(address string) {
 		r.logger.Errorln("server.ListenAndServe:", err)
 	}
 }
+
+// SetupHandler sets up the response handler.
+func (r *resolver) SetupHandler() relay.Handler {
```
leighmcculloch, line 116: Nice separation to improve the tests 👏 🎉 .

Suggest renaming this to indicate it is creating a new relay handler. SetupHandler sounds like the internals of the resolver are being setup, but actually this just creates a new handler from the resolver.
```suggestion
func (r *resolver) NewRelayHandler() relay.Handler {
```

services/ticker/internal/graphql_test.go
```diff
@@ -0,0 +1,162 @@
...
+		{queryField: "pairNames", queryVal: `[\"ETH_BTC\"]`, respField: "counterVolume", wantBody: `{"counterVolume": 174}`},
+		{queryField: "pairNames", queryVal: `[\"ETH_BTC\"]`, respField: "tradeCount", wantBody: `{"tradeCount": 3}`},
+		// Other input cases: multiple pairs, code.
+		{queryField: "pairNames", queryVal: `[\"BTC_ETH\", \"BTC_XLM\"]`, respField: "tradePair", wantBody: `{"tradePair": "BTC_ETH"},{"tradePair": "BTC_XLM"}`},
+		{queryField: "pairNames", queryVal: `[\"BTC_ETH\", \"BTC_XLM\"]`, respField: "tradeCount", wantBody: `{"tradeCount": 3},{"tradeCount": 2}`},
+		{queryField: "code", queryVal: `\"BTC\"`, respField: "tradePair", wantBody: `{"tradePair": "BTC_ETH"},{"tradePair": "BTC_XLM"}`},
+	}
+
+	for _, tc := range tests {
+		req := fmt.Sprintf(`{
```
leighmcculloch, line 71: Can we put each test case into a subtest so that each test case passes or fails independently? You can do that by wrapping the internals of this loop in `t.Run("name for the sub test", func (t *testing.T) {`.

services/ticker/internal/graphql_test.go
```diff
@@ -0,0 +1,162 @@
...
+			"query": "query getTicker() {ticker(%s: %s, numHoursAgo: 24) {%s}}",
+			"operationName": "getTicker",
+			"variables": {}
+		}`, tc.queryField, tc.queryVal, tc.respField)
+
+		fmt.Println(req)
```
leighmcculloch, line 77: It's completely fine and often helpful to log inside tests and commit those logs, but if you use `t.Log` the logs will be grouped with the test output. If you found it helpful to log this while writing the test someone will probably find it helpful when maintaining or debugging the test later, so I'd leave it in just change it to use `t`.

services/ticker/internal/graphql_test.go
```diff
@@ -0,0 +1,162 @@
...
+
+		require.Equal(t, http.StatusOK, resp.StatusCode)
+		assert.Equal(t, "application/json", resp.Header.Get("Content-Type"))
+
+		body, err := ioutil.ReadAll(resp.Body)
+		require.NoError(t, err)
+
+		wantBody := fmt.Sprintf(`{"data":{"markets":[%s]}}`, tc.wantBody)
+		assert.JSONEq(t, wantBody, string(body))
+	}
```
leighmcculloch, line 161: This looks like it is testing essentialy the same request just with different fields on each iteration. Is there a reason this test could not contain only a single request that requested all the fields and did a JSONEq comparison on the entire message? There are some reasons I think that would be desirable:
- Looking at the test would give the reader a solid understanding of what a full request and response looks like without any need to fmt.Sprintf in components of the request.
- It might be easier to maintain the test and evolve it because the test will be less intelligent and contain less logic.
- The test output should still be easy to read if one field fails because JSONEq will show a diff of what fields are incorrect. This will likely be easier to read than multiple separate assertions failing.
- Of much lesser importance, the test will probably be faster.

services/ticker/internal/graphql_test.go
```diff
@@ -0,0 +1,162 @@
...
+		resp := w.Result()
+
+		require.Equal(t, http.StatusOK, resp.StatusCode)
+		assert.Equal(t, "application/json", resp.Header.Get("Content-Type"))
+
+		body, err := ioutil.ReadAll(resp.Body)
+		require.NoError(t, err)
+
+		wantBody := fmt.Sprintf(`{"data":{"ticker":[%s]}}`, tc.wantBody)
+		assert.JSONEq(t, wantBody, string(body))
```
leighmcculloch, line 91: Similar to the `TestMarkets` test, I think this test could be reduced down to 3 requests that all show the full request /response and do a comparison for the same reasons specified in the other comment.

services/ticker/internal/graphql_test.go
debnil: These reasons make sense. I decomposed it into different requests to more easily isolate a single incorrect field, due to a specific API change. But I think you're right the test output wouldn't be too bad to read.

I agree with the other three benefits.

One major issue: a single request becomes abnormally large. It can't be separated into multiple lines, since it's within a double-quoted string in a JSON, even if included through `fmt.Sprintf` components. But I don't see a way around it, that isn't just verbose in some other way. Is it just fine to have an extremely long `query` value in the request then?

leighmcculloch: Gotcha. I think it is fine to have a really long query. Here's an example of what I had in mind, using the ticker query test as a case: #2637.

PR #2639, opened 2020-05-31
services/horizon/internal/expingest: Remove orderbook graph from ingestion system

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Follow-up to https://github.com/stellar/go/pull/2630 which removes orderbook graph ingestion.

### Why

After https://github.com/stellar/go/pull/2630 , the horizon ingesting nodes no longer need to maintain an in memory order book. Instead, the request serving horizon instances will populate their order books by polling offers from the horizon db. This polling process is downstream of ingestion.

### Known limitations

[ N/A]


services/horizon/internal/expingest/orderbook.go
```diff
@@ -250,11 +211,23 @@ func (o *OrderBookStream) Update() error {
...
 	}
 
-	_, _, err = o.update(status)
-	return err
+	if err := o.update(status); err != nil {
+		return errors.Wrap(err, "Error updating")
+	}
+
+	if requiresVerification {
+		o.verifyAllOffers()
```
bartekn, line 230: If there are more offers in the network this can be a problem because it can slow down updating the graph and global `Tick()` function. OK now but we should probably move this to it's own go routine (like ingestion verifier).

services/horizon/internal/expingest/orderbook.go
```diff
@@ -85,100 +106,77 @@ func (o *OrderBookStream) update(status ingestionStatus) ([]history.Offer, []xdr
...
-	sort.Slice(fromIngestion, func(i, j int) bool {
-		return fromIngestion[i].OfferId < fromIngestion[j].OfferId
-	})
-	mismatch := len(fromDB) != len(fromIngestion)
-	if !mismatch {
-		for i, offerRow := range fromDB {
-			offerEntry := fromIngestion[i]
+		for i, offerRow := range ingestionOffers {
+			offerEntry := offers[i]
 			if offerRow.OfferID != offerEntry.OfferId ||
```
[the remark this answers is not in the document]
tamirms: unfortunately the db offers are of type `history.Offer` not `xdr.OfferEntry` so I don't see how we can use `xdr.MarshalBase64` or `xdr.MarshalBinary` to do the comparison

services/horizon/internal/expingest/fsm.go
```diff
@@ -338,6 +329,11 @@ func (r resumeState) run(s *System) (transition, error) {
 	if ingestLedger > lastIngestedLedger+1 {
 		return start(), errors.New("expected ingest ledger to be at most one greater " +
 			"than last ingested ledger in db")
+	} else if ingestLedger <= lastIngestedLedger {
+		log.WithField("ingestLedger", ingestLedger).
+			WithField("lastIngestedLedger", lastIngestedLedger).
+			Info("bumping ingest ledger to next ledger after ingested ledger in db")
+		ingestLedger = lastIngestedLedger + 1
```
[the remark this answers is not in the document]
tamirms: I think it's better as is because at this moment in the code we have the select for update lock. if we transition to the state we need to release the select for update and the next node would need to acquire the select for update lock. I think by bumping the ingestLedger we reduce latency in the ingestion process

services/horizon/internal/expingest/orderbook.go
```diff
@@ -12,7 +13,10 @@ import (
 	"github.com/stellar/go/xdr"
 )
 
-const verificationFrequency = time.Hour
+const (
+	verificationFrequency = time.Hour
+	updateFrequency       = 30 * time.Second
```
bartekn, line 18: Can we change it to 5 sec. so it updates with every ledger (on average)?
```suggestion
	updateFrequency       = 5 * time.Second
```

services/horizon/internal/expingest/orderbook.go
```diff
@@ -231,3 +235,21 @@ func (o *OrderBookStream) Update() error {
...
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ticker.C:
+			if err := o.Update(); err != nil {
+				log.WithError(err).Error("could not apply updates from order book stream")
+			}
+		case <-ctx.Done():
+			log.Info("finished background ticker")
```
bartekn, line 251:
```suggestion
			log.Info("shutting down OrderBookStream")
```

PR #2640, opened 2020-06-01
txnbuild: Make txnbuild create V1 transactions

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Currently, `txnbuild` creates v0 transaction envelopes which are accepted by both protocol 12 and protocol 13. In this PR, I have configured `txnbuild` to only create v1 transactions which will only be accepted once protocol 13 is enabled.

This release should only be published after pubnet transitions to protocol 13. However, developers who would like to try out protocol 13 features in testnet such as fee bump transactions can use the txnbuild package from this PR to generate v1 transactions.

### Why

It is not possible to create fee bump transactions using v0 transaction envelopes.

### Known limitations

Once this PR is merged there will be no way to generate v0 transaction envelopes anymore. However, I don't think this is a big concern because we will only publish this release once protocol 13 is enabled.


clients/horizonclient/examples_test.go
```diff
@@ -1052,157 +1052,108 @@ func ExampleClient_StreamTransactions() {
...
+		return
+	}
+
+	feeBumpKP := keypair.MustParseFull("SA5ZEFDVFZ52GRU7YUGR6EDPBNRU2WLA6IQFQ7S2IH2DG3VFV3DOMV2Q")
+	feeBumpTx, err := txnbuild.NewFeeBumpTransaction(txnbuild.FeeBumpTransactionParams{
+		Inner:      tx,
+		FeeAccount: feeBumpKP.Address(),
+		BaseFee:    txnbuild.MinBaseFee * 2,
+	})
+	feeBumpTx, err = feeBumpTx.Sign(network.TestNetworkPassphrase, feeBumpKP)
```
howardtw, line 1095: Do you skip checking these two errors on purpose? Same question to the below example.

tamirms: that is mistake, I'll push a fix

tamirms:
> what's the reason that we use SourceAccount as MuxedAccount in V1 and SourceAccountEd25519 in V0?

The XDR definition for v1 transaction envelopes and v0 transaction envelopes have different source account types. The source account field is represented as a `SourceAccountEd25519` in v0 because we want v0 transaction envelopes to be binary compatible with protocol 12 transaction envelopes. You can find more details about this here https://github.com/stellar/stellar-protocol/blob/master/core/cap-0015.md#how-does-the-transactionenvelope-transformation-work


---
> And what exactly is a MuxedAccount?

This is another feature that is partially introduced in protocol 13. https://github.com/stellar/stellar-protocol/blob/master/core/cap-0027.md has the full explanation

Issue #2187, opened 2020-01-29
Polish ingestion interfaces


tamirms:
We should go through the interfaces and exported structs / functions in the ingestion packages to make sure they are named correctly and are easy to understand. We should make sure there are comments on exported declarations and that the declarations are located in the correct packages.

tamirms:
Also, we should update the [`exp` package](https://github.com/stellar/go/tree/master/exp):

`exp/ingest`, `exp/orderbook`, and `exp/tools/dump-ledger-state` should be moved out from `exp`. also, `exp/support/pipeline` should be removed

ire-and-curses:
Also, [services/horizon/internal/expingest](https://github.com/stellar/go/tree/master/services/horizon/internal/expingest) should be renamed since it's not experimental any more.

2opremio:
This is already handled by #3110

Issue #2633, opened 2020-05-29
exp/ingest/ledgerbackend: Replace LedgerCloseMeta with xdr.LedgerCloseMeta


bartekn:
When `exp/ingest` prototype was created Stellar-Core didn't have `LedgerCloseMeta` type so we created a temporary `LedgerCloseMeta` in `ledgerbackend` package. Now, when the required type is there, we should remove [`ledgerbackend.LedgerCloseMeta`](https://github.com/stellar/go/blob/a8d74bb25bcb71099de03349512e7d92e7d48b85/exp/ingest/ledgerbackend/ledger_backend.go#L21-L29) and replace it with [`xdr.LedgerCloseMeta`](https://github.com/stellar/go/blob/master/xdr/xdr_generated.go#L5156-L5159).

Additionally, move `peekLedgerSequence` from captive core to be a method on `xdr.LedgerCloseMeta`.

Issue #2128, opened 2020-01-16
exp/ingest/io: DBLedgerReader keep all txs data in memory


bartekn:
### What version are you using?

`release-horizon-v0.25.0`

### What did you expect to see?

There's no need to keep all transactions in memory. In case of a very large ledger with a lot of meta updates it can eat a large chunk of memory.

### What did you see instead?

We can load txs in batches and use `LIMIT` query.

bartekn:
I started investigating this but I realized that the container struct for transfering ledger data suggested by stellar-core team is:
```go
type LedgerCloseMeta struct {
	LedgerHeader          xdr.LedgerHeaderHistoryEntry
	TransactionEnvelope   []xdr.TransactionEnvelope
	TransactionResult     []xdr.TransactionResultPair
	TransactionMeta       []xdr.TransactionMeta
	TransactionFeeChanges []xdr.LedgerEntryChanges
	UpgradesMeta          []xdr.LedgerEntryChanges
}
```
The order of the elements makes it hard to process `LedgerEntryChanges` in correct order (if we want to stream it, and we probably should in cases of big ledgers). I think I'll check for results of https://github.com/stellar/go/issues/2129 and if they are bad, we can add a hacky fix to get transactions one by one but the long term fix would be to decide how the transport struct should look like.

bartekn:
As performance tests proved: https://github.com/stellar/go/issues/2129#issuecomment-584235690 this component should be updated to decrease memory usage when ingesting ledgers.

PR #2644, opened 2020-06-01
exp/ingest/io: Refactor readers

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

* Add `LedgerReader` that gets `LedgerCloseMeta` from a provided ledger backend and is embedded by other readers.
* Rename `DBLedgerReader` to `LedgerTransactionReader` because it works with any ledger backend. It now embeds `LedgerReader`.
* `LedgerChangeReader` has been refactored and now embeds `LedgerTransactionReader`.
* Updated `GetChangesFromLedgerEntryChanges` to be a public function.
* Removed `ctx.Context` from readers because it's not idiomatic.
* Removed unused interfaces and mocks.

Part of #2187.


exp/ingest/io/ledger_change_reader.go
```diff
@@ -0,0 +1,114 @@
...
+		next := r.pending[r.pendingIndex]
+		r.pendingIndex++
+		if r.pendingIndex == len(r.pending) {
+			r.pendingIndex = 0
+			r.pending = r.pending[:0]
+		}
+		return next, nil
+	}
+
+	switch r.state {
```
abuiles, line 69: @bartekn would it make sense to put this whole section in its own function to better describe what this is doing? it seems like this function not only "reads" but also if there are no pending, loads tx data and fills pending.

abuiles: It would be helpful to document `state` and how we use it to move across the different kind of changes

2opremio:
How about `LedgerBackendReader` instead of `LedgerTransactionReader` ?

exp/ingest/io/ledger_header_reader.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"github.com/stellar/go/exp/ingest/ledgerbackend"
+	"github.com/stellar/go/support/errors"
+	"github.com/stellar/go/xdr"
+)
+
+// LedgerReader reads ledger header from a given backend and ledger sequence.
+// Use NewLedgerReader to create a new instance.
+type LedgerReader struct {
```
tamirms, line 11: will there be another component that uses `LedgerReader` besides `LedgerTransactionReader`? If not, I don't see the point in having a separate type which is just a wrapper around `backend.GetLedger(sequence)`

exp/ingest/io/ledger_change_reader.go
bartekn: I added a comment inside this method that explains what it's doing.

I can move parts to separate function but since this method actually fits the screen I think it's more readable that way. Let me know what you think.

bartekn:
> How about `LedgerBackendReader` instead of `LedgerTransactionReader`?

`Read()` method actually reads transactions from a given ledger. I think the name matches the other reader: `LedgerChangeReader` that reads changes.

> will there be another component that uses LedgerReader besides LedgerTransactionReader? If not, I don't see the point in having a separate type which is just a wrapper around backend.GetLedger(sequence)

Yes, I wanted to remove it completely but we need something like this in [`validateBucketList`](https://github.com/stellar/go/pull/2644/files#diff-c50fb06e6a73d724a16ad29239197647R161).

services/horizon/internal/expingest/processor_runner.go
```diff
@@ -158,7 +158,7 @@ func (s *ProcessorRunner) validateBucketList(ledgerSequence uint32) error {
 		return errors.Wrap(err, "Error getting bucket list hash")
 	}
 
-	ledgerReader, err := io.NewDBLedgerReader(s.ctx, ledgerSequence, s.ledgerBackend)
+	ledgerReader, err := io.NewLedgerReader(s.ledgerBackend, ledgerSequence)
```
tamirms, line 161: I think it would be better to get rid of the `LedgerReaderType` and instead call ledgerBackend directly like:

```golang
	exists, ledgerCloseMeta, err := backend.GetLedger(sequence)
	if err != nil {
		return nil, errors.Wrap(err, "error getting ledger from the backend")
	}

	if !exists {
		return nil, ErrNotFound
	}
	ledgerBucketHashList := ledgerCloseMeta.LedgerHeader.Header. BucketListHash
```

it's true that the code which calls ledger backend will also be present in `LedgerTransactionReader` but I don't that code duplication. It's ok for `backend.GetLedger` to be called in multiple places

bartekn: I think that there are several advantages of keeping `LedgerReader` reader:
* It's embedded in `LedgerTransactionReader` and `LedgerChangeReader`. For example, we call `LedgerTransactionReader.GetHeader()` in Horizon ingestion processors.
* Although it now has just two methods that are rather simple I imagine we will add more helpers and complicated methods in the future. For example: `GetCloseTime()` that transforms `ScpValue.CloseTime` to `time.Time` or a verify function that checks if the ledger hash actually matches the contents or even validating bucket hash method (would accept history adapter).

What do you think?

tamirms: regarding the first point, `LedgerReader` is only embedded in `LedgerTransactionReader`. `LedgerChangeReader` embeds `LedgerTransactionReader`. If the functions of `LedgerReader` become part of`LedgerTransactionReader` then `LedgerChangeReader` will remain unaffected.

regarding the second point, I would rather those utility functions be defined on `ledgerCloseMeta`. The reason why I believe that's better is that it starts to get confusing when we have to distinguish between `LedgerBackend`, `LedgerTransactionReader`, `LedgerChangeReader`, and `LedgerReader`.

tamirms: but if I'm the only one that finds it confusing then you can ignore my comment and leave it as is

bartekn: @tamirms can you take a look?

> regarding the second point, I would rather those utility functions be defined on `ledgerCloseMeta`. The reason why I believe that's better is that it starts to get confusing when we have to distinguish between `LedgerBackend`, `LedgerTransactionReader`, `LedgerChangeReader`, and `LedgerReader`.

Good point. Not doing this in this PR because this will likely be changed in #2633. I'm also not sure we should expose `xdr.LedgerCloseMeta` in the public release of `ingest` package because of #2128.

PR #2645, opened 2020-06-01
exp/services/recoverysigner: add signing keys migration

howardtw:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This PR adds a new `signers` table and its audit counterpart in order to prepare for having unique signing keys per registered account.

### Why

The application uses a single signing key for all accounts that are registered. The issue with this is that anyone who knows that single signer can determine what accounts are registered with a recoversigner server. This public association may be undesirable for a user using a recoverysigner.

### Known limitations

N/A


exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200601000000-create-signing-keys.sql
```diff
@@ -0,0 +1,17 @@
+-- +migrate Up
+
+CREATE TABLE signing_keys (
```
leighmcculloch, line 3: Can we call this table `signers`? I suggest that name because in the code and in the API it is currently called `signers` and that one-to-one naming makes it easier to identify the one-to-one relationship.

exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200601000000-create-signing-keys.sql
```diff
@@ -0,0 +1,17 @@
...
+  account_id BIGINT NOT NULL REFERENCES accounts (id) ON DELETE CASCADE,
+  key TEXT NOT NULL,
+  encrypted_seed BYTEA NOT NULL,
+  created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW()
+);
+
+
+CREATE INDEX ON signing_keys (account_id);
+CREATE INDEX ON signing_keys (key);
+
```
leighmcculloch, line 14: Could you also add an audit table for the signers table? See this as an example:

https://github.com/stellar/go/blob/62103e5b325359343da251345c9dcf9da986de97/exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200320000002-create-auth-methods-audit.sql#L1-L36

exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200601000000-create-signing-keys.sql
```diff
@@ -0,0 +1,17 @@
+-- +migrate Up
+
+CREATE TABLE signing_keys (
+  id BIGINT NOT NULL PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
+  account_id BIGINT NOT NULL REFERENCES accounts (id) ON DELETE CASCADE,
```
leighmcculloch, line 5: Could we order the columns into three groups with the reference IDs (`account_id`) first like the other migrations and include an `updated_at`?

Example:
https://github.com/stellar/go/blob/62103e5b325359343da251345c9dcf9da986de97/exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200311000002-create-auth-methods.sql#L10-L18

exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200601000000-create-signing-keys.sql
```diff
@@ -0,0 +1,17 @@
...
+  key TEXT NOT NULL,
+  encrypted_seed BYTEA NOT NULL,
```
leighmcculloch, line 7: What will be contained in the `key` and `encrypted_seed` columns?

howardtw: This will the unique signing key and its secret seed. I took the name from the latest. response in SEP-30 for `key`.

exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200601000000-create-signing-keys.sql
howardtw: How could the key be updated? It will either be added or deleted. Speaking of this, should we audit this table as well?

howardtw: https://github.com/stellar/go/pull/2645/commits/6a63e64ca1e91e9d998fdaaa8a9eef72eb1539ab

leighmcculloch: > How could the key be updated?

I'm not entirely sure. We weren't really sure how we'd be updating the other tables but when we discussed if we should include them somebody said to just include it and I did. I think it is worth being consistent but I agree I'm not really sure how we'd update this table. I guess an upside of including it now is if we do add a query that updates in the future we are likely to see the field here and set it. Do you think there is a big downside to including it early?

howardtw: I feel like we can add the `updated_at` when we find us updating the entries for some reason. At this moment, I'm a little bit reluctant to add a column that we won't be using in the near future. Especially it won't be too late to do so when needed.

exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200601000000-create-signing-keys.sql
leighmcculloch: What will be the contents encrypted value stored in the `encrypted_seed` column? Will it be a private/secret key strkey (i.e. an `S` string)?

I think there's some value in giving these two fields names that share a common keyword, like `key` so we communicate a relationship between them, and name them so that there is no doubt which is a public value, which is a secret value. The `encrypted_` prefix does give some of that away, but it's not definitively clear when I look at it.

My main concern is someone glances over this and misses that the `seed` is sensitive, either here or in the code, because in everyday development we do not use the term `seed` to mean something sensitive.

Even though we use the term seed frequently in the protocol docs, we also use the term private key just as much, and in code we've used the term secret key or secret a lot too. Out of correctness I don't think we need to lean one way or another.

The term 'secret key' is so obviously something that is sensitive to anybody regardless of their familiarity with Stellar or unfamiliarity with crypto specific terms like 'seed'.

What do you think?

```suggestion
  public_key TEXT NOT NULL,
  encrypted_secret_key BYTEA NOT NULL,
```

---

By the way I love that you broke out the public key into a separate field even though we can derive it from the encrypted secret key. 👍

exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200601000000-create-signing-keys.sql
```diff
@@ -0,0 +1,19 @@
+-- +migrate Up
+
+CREATE TABLE signing_keys (
+  account_id BIGINT NOT NULL REFERENCES accounts (id) ON DELETE CASCADE,
+  id BIGINT NOT NULL PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
+
+  key TEXT NOT NULL,
+  encrypted_seed BYTEA NOT NULL,
+
+  created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW()
```
leighmcculloch, line 10: Can you plop this above the `key`? Just for consistency with others migration files, it should have no functional or other impact.

exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200601000000-create-signing-keys.sql
howardtw: > What will be the contents encrypted value stored in the encrypted_seed column? Will it be a private/secret key strkey (i.e. an S string)?

This will be the result of encrypting the S string, aka the `CiphertextBlob` in the output.
```
type EncryptOutput struct {
	_ struct{} `type:"structure"`

	// The encrypted plaintext. When you use the HTTP API or the AWS CLI, the value
	// is Base64-encoded. Otherwise, it is not encoded.
	//
	// CiphertextBlob is automatically base64 encoded/decoded by the SDK.
	CiphertextBlob []byte `min:"1" type:"blob"`

	// The ID of the key used during encryption.
	KeyId *string `min:"1" type:"string"`
}
```

> What do you think?

I like making them as `public_key` and `encrypted_private_key` 👍

leighmcculloch: Can we use secret key rather than private key? We seem to use that almost as commonly in app code and some SDKs and is easier to abbreviate in the same context of a private key.

Where will the `KeyId` be stored? Is it required for decryption? It's important that we be able to do key rotation of the encrypted data.

howardtw: Ah sorry I meant taking your suggestion as `encrypted_secret_key`.

> Where will the KeyId be stored? Is it required for decryption? It's important that we be able to do key rotation of the encrypted data.

`KeyId` is the same as the env `KMS_ARN` that we are using. In other words, we don't currently store the `KeyId` in the database as we use the [CMK (customer master key)](https://docs.aws.amazon.com/kms/latest/developerguide/concepts.html#master_keys) to encrypt the data directly and there is only one key. And if we ever need to rotate the key, we will have to call `ReEncrypt` on all the encrypted data, which we don't have this feature ready yet and we should prepare for this as soon as we can.

People generally use CMK directly to encrypt data that's less than 4 KB such as a password or RSA key. In fact, it cannot be used to encrypt data that is larger than 4 KB. To encrypt larger application data, we have to use [data keys](https://docs.aws.amazon.com/kms/latest/developerguide/concepts.html#data-keys). If we plan on using multiple data keys, that's when `KeyId` becomes important. See [here](https://github.com/awsdocs/aws-doc-sdk-examples/blob/master/go/example_code/kms/kms_create_key.go#L37-L41) as well. Does that make sense? We can have a conversation on this matter in person as I think you would be interested in learning this 😄

howardtw: Actually I was wrong about `KeyId`. It is only relevant to CMKs, not data keys. To use a data key, we have to use a CMK to encrypt/decrypt a it and use [AWS Encryption SDK](https://docs.aws.amazon.com/encryption-sdk/latest/developer-guide/introduction.html) to do further operation.

exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200601000000-create-signing-keys.sql
howardtw: https://github.com/stellar/go/pull/2645/commits/9a400f83b8e7724a413de19634d117f8754a9d04

exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200601000000-create-signing-keys.sql
howardtw: https://github.com/stellar/go/pull/2645/commits/9a400f83b8e7724a413de19634d117f8754a9d04

exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200601000000-create-signing-keys.sql
```diff
@@ -0,0 +1,20 @@
...
+
+CREATE TABLE signing_keys (
+  account_id BIGINT NOT NULL REFERENCES accounts (id) ON DELETE CASCADE,
+  id BIGINT NOT NULL PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
+
+  created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
+
+  public_key TEXT NOT NULL,
+  encrypted_secret_key BYTEA NOT NULL,
+  encryption_keyid TEXT NOT NULL
```
leighmcculloch, line 11: Is that a typo that key ID has no separator?
```suggestion
  encryption_key_id TEXT NOT NULL
```

howardtw: https://github.com/stellar/go/pull/2645/commits/048a3b39e112bc4e0617fa238c704955c0729843
^ I feel like I've made so many mistakes programming with one eye 👁️ 🤦

exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200601000000-create-signing-keys.sql
howardtw: https://github.com/stellar/go/pull/2645/commits/57215c2ddab68e09c58d1c704f34cc63729a73da

howardtw:
The first part of coding is here: https://github.com/stellar/go/pull/2659/

How did you turn a PR back to a draft? I know this is a new feature, but I couldn't find the button to do it.

PR #2646, opened 2020-06-01
Update prototype to run captive on online ingestion mode.

abuiles:
Just passing the config allow us to start using captive on online mode. It seems to work fine but it's very slow.


bartekn:
LGTM! Let's create `captive-core-online` branch from `captive-core` where we can push online related commits. This way we'll be able to merge offline commits to `master` and release it as an experimental feature.

abuiles:
@bartekn perfect, I created the captive-core-online-branch.

PR #2647, opened 2020-06-01
exp/services/recoverysigner: support multiple sep10 json web keys in the key set

### LINKED ISSUE #2621: exp/services/recoverysigner: add support for configuring with multiple SEP-10 JWK keys
### What
Add support for configuring the service with multiple SEP-10 JWK keys. The service should allow SEP-10 JWTs from any of the provided JWK keys.

### Why
The format of the JWK keys that the service accepts is already JWKS which means the configuration interface can already accept multiple keys however internally the service only uses the first key.
This is required to support a SEP-10 server needing to rotate its JWK key. For a short period of time during rotation the recoverysigner should accept both new and old keys. After rotation is complete and all JWTs that were issued with the old key are expired then the old key can be removed from recoverysigner. This is how multiple JWK formatted keys will be used.

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Support multiple SEP-10 JSON Web Keys at one time by attempting to verify JWTs with each JWK configured, to see if the JWT is signed by any.

### Why

To support rotating the JSON Web Key used in the SEP-10 server. SEP-10 servers issue JWTs for a recommend time period of 15 minutes and if the key used for signing JWTs is to be changed there will be a period of at least 15 minutes where valid JWTs may be issued by two keys. Also, coordinating the configuring multiple services may take longer than 15 minutes.

Each key is checked because not all SEP-10 servers use Key IDs (`kid`) and the server should work with existing JWTs as issued by existing SEP-10 servers. Typically the `kid` field in the header would be used to identify which key was used and so the server would not attempt to verify with all keys it had but only the key specified in the header. This causes the server to do more work to verify some JWTs twice, but any performance impact of this should be negligible.

Close #2621 

### Known limitations

N/A


leighmcculloch:
I pushed an additional commit to this PR adding logging, but given @marcelosalloum already reviewed this PR I reverted it and moved that commit to a new PR, #2655.

PR #2650, opened 2020-06-02
txnbuild: Allow fee bump transactions to bump v0 transactions

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Currently when you try to create a fee bump transaction from a v0 transaction we return an error. This PR does not return an error and instead converts the v0 transaction to a v1 transaction before wrapping it with a fee bump transaction.

### Why

Consider the use case of a pre-signed transaction that was made in the distant past. You created the transaction as part of a smart contract, and now you want to execute it, but you need it to have a higher fee than you set back then.

### Known limitations

[N/A]

### DIFF
--- a/txnbuild/fee_bump_test.go
+++ b/txnbuild/fee_bump_test.go
@@ -38,30 +38,96 @@ func TestFeeBumpInvalidFeeSource(t *testing.T) {
 	assert.Contains(t, err.Error(), "fee account is not a valid address")
 }
 
-func TestFeeBumpInvalidInnerTxType(t *testing.T) {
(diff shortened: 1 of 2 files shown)


tamirms:
> Looks good! Shall we implement the same across other SDKs too?

@abuiles yes, we'll need to implement two changes in the other SDKs

1) stop creating v0 transaction envelopes and instead only create v1 envelopes https://github.com/stellar/go/pull/2640
2) convert v0 transactions to v1 transactions when creating fee bump transactions (this PR)

PR #2659, opened 2020-06-04
exp/services/recoverysigner: add KMS-related setup

howardtw:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This PR adds an interface `KMS` (key management service) that is designed to be integrated with either AWS KMS or other services like Google Cloud KMS. The two methods use a field called `EncryptionContext`, which will be used by AWS KMS as it is recommended by AWS when using symmetric encryption. See [here](https://aws.amazon.com/blogs/security/how-to-protect-the-integrity-of-your-encrypted-data-by-using-aws-key-management-service-and-encryptioncontext/).

KMS library was updated in order to use func `GetPublicKey`. We can revert the commit if prefered.

### Why

The application uses a single signing key for all accounts that are registered. The issue with this is that anyone who knows that single signer can determine what accounts are registered with a recoversigner server. This public association may be undesirable for a user using a recoverysigner.

### Known limitations

1. We have yet to determine how we are going to use data key. e.g. how we are going to use KMS.
2. We have yet to determine whether it is ok to have no KMS_PROVIDER given that this will be open source.


howardtw:
Closing as this PR is being replaced by https://github.com/stellar/go/pull/2692

PR #2661, opened 2020-06-04
support/render/problem: Log request errors

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Record errors which terminate a request in the request context so that they can be included in the logs emitted by the logging middleware.
 
### Why

We log information information about every request including the http status code of the response and the duration of the request. When a request responds with an error status code (>= 500) . It would be useful to know what exact error triggered the response.

### Known limitations

Propagating values via context is often controversial. Alternatively, I could log the error directly in the `Render()` function. 

The only advantage of using the context in this PR is that we can include the error in the same logging payload which contains other attributes of the request like http status, http path, etc. So, if you have a burst of failing requests you can quickly see the cause of those failures with one search in kibana. Otherwise, we'd need to find the request ids for each of the failing requests and then search for those request ids separately to determine the request errors.

Let me know which approach you prefer.


bartekn:
> Alternatively, I could log the error directly in the `Render()` function.

While I think that passing error via context is not perfect but OK in this situation I'd vote for logging these errors with `info` level in `Render()`.

The reason is that there are other queries in which ability to _join_ other entries would be really great and supporting some of them is not even possible. One example is getting `route` value for `Starting request` which isn't possible because we log this entry before router finds the matching route.

It looks like ELK provides a simple parent-child [`join`](https://www.elastic.co/guide/en/elasticsearch/reference/current/parent-join.html) relationship we can explore. We could make all log entries within a single request children of "Finished request". I have never used this before so I can be wrong.

tamirms:
@bartekn this is ready for review again

support/render/problem/problem.go
```diff
@@ -154,7 +175,9 @@ func (ps *Problem) Render(ctx context.Context, w http.ResponseWriter, err error)
 		// If this error is not a registered error
 		// log it and replace it with a 500 error
 		if !ok {
-			ps.log.Ctx(ctx).WithStack(err).Error(err)
+			if ps.filter == LogUnknownErrors {
+				ps.log.Ctx(ctx).WithStack(err).Error(err)
+			}
```
bartekn, line 180: @stellar/horizon-committers  I'm wondering if it's a bug. I'm currently debugging other issue connected to `/claimable_balances` and the error I'm triggering is propagated here but it's logged with info level and `request failed due to error` message (few lines above). I think that the error level entry should be always emitted, except when `ps.filter == LogNoErrors`. Am I missing something here?
```diff
-			if ps.filter == LogUnknownErrors {
+			if ps.filter != LogNoErrors {
				ps.log.Ctx(ctx).WithStack(err).Error(err)
			}
```

Issue #2643, opened 2020-06-01
Remove all instances of code that access core db


abuiles:
We should remove Core DB access where possible. Ideally, Horizon shouldn't rely on Core DB to work correctly.

## TODO
The following are a list of places where we still using core's db which we won't be able to drop yet.

- [x] We still need coreQ to get the LatestLedger  - we need to figure a solution for this before being able to drop access to core's db #2725.
- [x] `ledgerbackend` .
- [x] `SubmissionSystem` https://github.com/stellar/go/pull/2702

bartekn:
I think that ultimately we should remove all Core DB access before shipping captive core. Simply remove `db2/core` package and see what happens.

abuiles:
I just added the following ticket to keep track of the other task to get this done https://github.com/stellar/go/issues/2725

abuiles:
This has been fixed now.

PR #2662, opened 2020-06-04
services/horizon/internal/db2/history: Add AssetsForAddress.

abuiles:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add `AssetForAddress` to `db2/history`.

### Why

The path finding action handler rely on core to load assets and balances for a given address. This PR adds the same method which returns `xdr.Assets` and their matching balances but using Horizon DB as the source.

This is part of the initial work to drop db2/core (#2643 ).

### Known limitations

[TODO or N/A]

### DIFF
--- a/services/horizon/internal/db2/history/trust_lines.go
+++ b/services/horizon/internal/db2/history/trust_lines.go
@@ -21,6 +21,38 @@ func (trustLine TrustLine) IsAuthorizedToMaintainLiabilities() bool {
 	return xdr.TrustLineFlags(trustLine.Flags).IsAuthorizedToMaintainLiabilitiesFlag()
 }
 
+// AssetsForAddress returns a list of assets and balances for those assets held by
+// a given address.
+func (q *Q) AssetsForAddress(addy string) ([]xdr.Asset, []xdr.Int64, error) {
+	account, err := q.GetAccountByID(addy)
+
+	if q.NoRows(err) {
+		// if there is no account for the given address then
+		// we return an empty list of assets and balances
+		return []xdr.Asset{}, []xdr.Int64{}, nil
+	} else if err != nil {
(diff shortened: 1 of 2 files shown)


services/horizon/internal/db2/history/trust_lines.go
```diff
@@ -21,6 +21,38 @@ func (trustLine TrustLine) IsAuthorizedToMaintainLiabilities() bool {
 	return xdr.TrustLineFlags(trustLine.Flags).IsAuthorizedToMaintainLiabilitiesFlag()
 }
 
+// AssetsForAddress returns a list of assets and balances for those assets held by
+// a given address.
+func (q *Q) AssetsForAddress(addy string) ([]xdr.Asset, []xdr.Int64, error) {
```
abuiles, line 26: btw, I considered a different approach where we return trustlines -- but ultimately decided to go with an implementation that matches the original interface so that we can use this as a drop-in replacement.

bartekn, line 26: I wonder if we should add something like: https://github.com/stellar/go/blob/1a17544531b0f306961c6f0de61e432102b579e5/services/horizon/internal/db2/history/orderbook.go#L49-L51

We are sending two queries so they may return results at different ledgers if not in `REPEATABLE READ` tx.

abuiles:
Makes sense - I moved the implementation almost verbatim from the coreQ
version -- do you know why was the repeatable read not added there? An
oversight maybe?

bartekn:
> do you know why was the repeatable read not added there?

It was simply a bug. We made 3-4 DB queries to build `/account/{id}` and non was sent in a transaction.

PR #2664, opened 2020-06-05
exp/ingest/ledgerbackend: Add buffered pipe to CaptiveStellarCore

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This commit adds buffering of meta stream via a buffered pipe built using a channel.

### Why

The first return value of `GetLedger` method of `LedgerBackend` specifies if a ledger is available in the backend. Currently `CaptiveStellarCore` ignores this value and always blocks until the ledger has been streamed. This is fine for the offline mode but in the online mode blocking is not acceptable.

To solve this we use a `bufferedPipe` that allows checking if there is new data in a stream (`IsEmpty` method). This way we can determine if ledger is available.

My first approach was to use `bufio.Reader` (previously used in `stellarCoreRunner`) that exposes a `Peek()` method, however there are several problems:
* `bufio.Reader` is not thread-safe and we need another go routine to constantly write read bytes to a pipe.
* Making `bufio.Reader` thread-safe by using `sync.Mutex` was much slower than `bufferedPipe`.
* Finally, it's not possible to bound an internal buffer in `bufio.Reader` which means that it can grow too much.

Finally, `CaptiveStellarCore` can be run in two modes: non-blocking (default) and blocking. In non-blocking mode `GetLedger` returns false when ledger is not available, in blocking mode it blocks. This is done to support fast reingestion. It turns out that Horizon is able to ingest history data faster that stellar-core can apply old ledger (probably because Horizon doesn't update state when reingesting). Without blocking mode we'd need to call `time.Sleep` when ledger is not available which is much slower than blocking.

### Known limitations

* Tests are not there yet - wanted to get an initial feedback.
* I didn't want to add `SetBlockingMode` to `LedgerBackend` interface because in the future `CaptiveStellarCore` will be the only ledger backend available.
* I uncommented the code that output stellar-core log to better understand why blocking mode is faster. Logging will be removed before merging.

### DIFF
--- a/exp/ingest/ledgerbackend/captive_core_backend.go
+++ b/exp/ingest/ledgerbackend/captive_core_backend.go
@@ -3,6 +3,7 @@ package ledgerbackend
 import (
 	"io"
 	"sync"
+	"time"
 
 	"github.com/pkg/errors"
 	"github.com/stellar/go/network"
@@ -10,8 +11,8 @@ import (
 	"github.com/stellar/go/xdr"
 )
 
-// Ensure captiveStellarCore implements LedgerBackend
-var _ LedgerBackend = (*captiveStellarCore)(nil)
+// Ensure CaptiveStellarCore implements LedgerBackend
+var _ LedgerBackend = (*CaptiveStellarCore)(nil)
 
 // This is a not-very-complete or well-organized sketch of code be used to
 // stream LedgerCloseMeta data from a "captive" stellar-core: one running as a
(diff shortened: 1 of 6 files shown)


exp/ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -164,34 +165,43 @@ func (c *captiveStellarCore) openOfflineReplaySubprocess(nextLedger, lastLedger
...
+	if metaPipe == nil {
 		return errors.New("missing metadata pipe")
 	}
 
-	_, _, err := c.GetLedger(from - 1)
-	if err != nil {
-		return errors.Wrap(err, "opening getting ledger `from-1`")
+	for {
+		// Wait for the first byte
+		if !metaPipe.IsEmpty() {
```
tamirms, line 180: what's the difference between this code and the old code? it seems like they both block until the ledger is available?

exp/ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -110,14 +114,25 @@ func (r *stellarCoreRunner) run(from, to uint32) error {
...
-	err = r.start()
+	reader, err := r.start()
 	if err != nil {
 		return errors.Wrap(err, "error starting stellar-core subprocess")
 	}
+
+	r.metaPipe = newBufferedPipe(pipeBufferSize)
+	go func() {
+		for {
+			io.Copy(r.metaPipe, reader)
```
tamirms, line 125: shouldn't you check the error from copy?

2opremio:
Did you consider using something like https://github.com/klauspost/readahead instead?

exp/ingest/ledgerbackend/captive_core_backend.go
bartekn: The new code does return (`false`) if no new data is available, the old code blocks.

bartekn:
> Did you consider using something like https://github.com/klauspost/readahead instead?

I think it takes less time to write a small utility like this (~40 lines of code) than to check if the dependency code actually works correctly in all the cases we need.

bartekn:
Closing in favour of @2opremio PR that will do the same thing but will store decoded XDR objects instead of raw bytes.

PR #2670, opened 2020-06-05
txnbuild: Provide option to still generate V0 transaction envelopes

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Provide option to still generate V0 transaction envelopes.

### Why

A developer may still want to generate v0 transactions. For example, if you need to generate a transaction and sign it with a hardware wallet. The hardware wallet may still be stuck on protocol 12. So, having the option to generate v0 transactions is convenient in this case.

### Known limitations

[N/A]

### DIFF
--- a/txnbuild/fee_bump_test.go
+++ b/txnbuild/fee_bump_test.go
@@ -50,7 +50,7 @@ func TestFeeBumpUpgradesV0Transaction(t *testing.T) {
 			BaseFee:              2 * MinBaseFee,
 			Memo:                 MemoText("test-memo"),
 			Timebounds:           NewInfiniteTimeout(),
-			V0:                   true,
+			Version:              TransactionVersion0,
 		},
 	)
 	assert.NoError(t, err)
--- a/txnbuild/fee_bump_test.go
+++ b/txnbuild/fee_bump_test.go
@@ -50,14 +50,13 @@ func TestFeeBumpUpgradesV0Transaction(t *testing.T) {
 			BaseFee:              2 * MinBaseFee,
 			Memo:                 MemoText("test-memo"),
 			Timebounds:           NewInfiniteTimeout(),
+			V0:                   true,
 		},
 	)
 	assert.NoError(t, err)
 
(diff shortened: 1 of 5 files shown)


txnbuild/fee_bump_test.go
```diff
@@ -50,15 +50,14 @@ func TestFeeBumpUpgradesV0Transaction(t *testing.T) {
 			BaseFee:              2 * MinBaseFee,
 			Memo:                 MemoText("test-memo"),
 			Timebounds:           NewInfiniteTimeout(),
+			V0:                   true,
```
leighmcculloch, line 53: Booleans in APIs are difficult to extend and if we have a V2 and so on we will have many. Could this be an enumeration?

```
type TransactionVersion int
const (
	TransactionVersionLatest TransactionVersion = iota
	TransactionVersion0
	TransactionVersion1
)
```

tamirms:
@abuiles tested signing v1 transactions on ledger and trezor hardware wallets and it turns out that there are no issues with signing v1 transactions. It seems like we don't have anymore use cases for generating v0 transactions.

I will close the PR because allowing v0 transactions does complicate the API a bit and we should have some use cases before we implement that feature in txnbuild

leighmcculloch:
What about someone who needs to decode a V0 transaction and sign it and re-encode it. Does punting on this prevent us from doing that or that is still possible?

tamirms:
> What about someone who needs to decode a V0 transaction and sign it and re-encode it. Does punting on this prevent us from doing that or that is still possible?

That is still possible. It is true that`NewTransaction()` only constructs v1 transactions. However, you can still obtain a v0 transaction by calling `TransactionFromXDR()` to decode a v0 transaction envelope string.

Signing and encoding transactions will work for both v0 and v1 transaction types: 

https://github.com/stellar/go/blob/release-horizonclient-v3.2.0/txnbuild/transaction.go#L156-L174

PR #2672, opened 2020-06-05
services/horizon: Write captive-core logs to /dev/null when discarding them

2opremio:
In macOS, captive-core blocks when writing to a pipe which isn't read from.

### DIFF
--- a/exp/ingest/ledgerbackend/stellar_core_runner.go
+++ b/exp/ingest/ledgerbackend/stellar_core_runner.go
@@ -60,7 +60,7 @@ func (r *stellarCoreRunner) getConfFileName() string {
 	return filepath.Join(r.getTmpDir(), "stellar-core.conf")
 }
 
-func (*stellarCoreRunner) getLogLineWriter() io.Writer {
+func (*stellarCoreRunner) GetLogLineWriter() io.Writer {
 	r, w := io.Pipe()
 	br := bufio.NewReader(r)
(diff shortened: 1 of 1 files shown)


exp/ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -107,7 +109,8 @@ func (r *stellarCoreRunner) run(from, to uint32) error {
 	args := []string{"--conf", r.getConfFileName(), "catchup", rangeArg, "--replay-in-memory"}
 	cmd := exec.Command(r.executablePath, args...)
 	cmd.Dir = r.getTmpDir()
-	cmd.Stdout = r.getLogLineWriter()
+	// In order to get the full stellar core logs:
+	// cmd.Stdout = r.GetLogLineWriter()
```
tamirms, line 113: should this be commented out?

2opremio: No, unless we want to print all the logs from core. (having it commented out maintains the behavior we had before the PR).

2opremio:
> Sorry about this! I actually always uncommented it while testing so I couldn't catch it.

No worries. Interestingly, it seems to work on Linux.

> This is ok for a quick fix. In the long term we should probably stream stellar-core logs to a file in temp directory (captive core session creates one) and add a method that returns a path to this log file and display it in the logs. Then if we need the log stream for debugging we can just `tail -f` the file.

That, or make core less verbose and print those logs.

PR #2673, opened 2020-06-05
exp/ingest/ledgerbackend: Stream stellar-core logs to file

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Stream stellar-core logs to file.

### Why

Currently we stream logs to stdout what can be hard to read when also Horizon logs are streamed.

### DIFF
--- a/exp/ingest/ledgerbackend/stellar_core_runner.go
+++ b/exp/ingest/ledgerbackend/stellar_core_runner.go
@@ -1,7 +1,6 @@
 package ledgerbackend
 
 import (
-	"bufio"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -9,7 +8,6 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
-	"regexp"
 	"strings"
 	"time"
 
@@ -19,6 +17,7 @@ import (
 type stellarCoreRunnerInterface interface {
 	catchup(from, to uint32) error
 	runFrom(from uint32) error
+	getLogFilePath() string
 	getMetaPipe() io.Reader
 	close() error
 }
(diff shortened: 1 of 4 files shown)


bartekn:
> shall we create an issue to make this configurable?

Do you mean if logs should be streamed to a file? I'm not sure. We create a file in a tempdir that gets removed when ledger backend is closed so it shouldn't be an issue.

abuiles:
>Do you mean if logs should be streamed to a file? I'm not sure. We create a file in a tempdir that gets removed when ledger backend is closed so it shouldn't be an issue.
👍

exp/ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -60,22 +59,16 @@ func (r *stellarCoreRunner) getConfFileName() string {
...
-			if e != nil {
-				break
-			}
-			line = dateRx.ReplaceAllString(line, "")
-			fmt.Print(line)
-		}
-	}()
-	return w
+func (r *stellarCoreRunner) getLogLineWriter() (io.Writer, error) {
+	f, err := os.OpenFile(r.getLogFilePath(), os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
```
2opremio, line 63: Having real time logs of core interleaved with the horizon logs can be valuable to reason about what's happening, I wouldn't totally remove the option of printing the logs instead of using a file.

2opremio: In fact, wouldn't it be simpler to simply print the Core logs (with the same format as the Horizon ones) but control their verbosity (including not printing any logs at all)?

2opremio: Using a temporary file which gets removed makes debugging cumbersome:
0. You need to make sure the file isn't removed (e.g. if horizon crashes or if the container terminates they will be gone).
1. You need to fish for the file
2. You need to reason about the timestamps of the file with respect to the timestamp in the horizon logs

bartekn: > In fact, wouldn't it be simpler to simply print the Core logs (with the same format as the Horizon ones) but control their verbosity (including not printing any logs at all)?

I think it makes sense. My only worry is that the log format may change in the future breaking the parsing code but I think we should be able to catch it quickly. I'll update the PR.

2opremio: We could add a test for it, so that we catch it when it happens. Right?

bartekn: Yes but this requires integration tests (you actually need to run stellar-core to catch this) and we don't have them yet.

bartekn:
> I think you could use the same writer for both (assuming stellar core writes line by line, which it should).
>
> On top of that, the line writer goroutines should be stopped on `Close()` (it didn't matter for debugging, when it was commented out, but now they will cause a memory leak).

_Originally posted by @2opremio in https://github.com/stellar/go/pull/2694_

PR #2678, opened 2020-06-06
exp/services/recoverysigner: remove accounts/{address}/sign endpoint

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Remove the `/accounts/{address}/sign` endpoint.

### Why
In SEP-30 v0.3.0 and v0.4.0 the signing endpoint was replaced with
the `/accounts/{address}/sign/{signing-address}` endpoint. This service
is still experimental and we can still make breaking changes like the
SEP.

Implements stellar/stellar-protocol#629

### Known limitations

N/A


exp/services/recoverysigner/internal/serve/account_sign_test.go
```diff
@@ -1,1142 +0,0 @@
-package serve
```
leighmcculloch, line 1: This file is no longer needed because another test file `account_sign_signing_address_test.go` contains all these same tests already with the signing address included in the path.

PR #2685, opened 2020-06-08
support/http: log the route matched by the route handler

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Log the route pattern that was matched in addition to the path which is already logged for each request.

### Why
For requests with lots of parameters it can be complex to filter logs based on partial matches of a path. It's simpler and faster if we can query logs on the route's pattern that was matched to look at all logs for
an endpoint.

I have experienced this difficulty writing queries for the `exp/services/recoverysigner` which has paths like `accounts/G.../sign/G...`. This new field will make it possible to filter all logs hitting that endpoint on `accounts/{address}/sign/{signing-address}`.

The log only occurs at the end of the request because unfortunately the route handler doesn't know which route will be matched at the time that we log the start request log.

### Known limitations

N/A


support/http/logging_middleware.go
```diff
@@ -48,14 +49,15 @@ func LoggingMiddleware(next stdhttp.Handler) stdhttp.Handler {
...
+	l := log.Ctx(r.Context()).WithFields(log.F{
 		"subsys":    "http",
 		"path":      r.URL.String(),
 		"method":    r.Method,
 		"ip":        r.RemoteAddr,
 		"host":      r.Host,
 		"useragent": r.Header.Get("User-Agent"),
-	}).Info("starting request")
+	})
+	l.Info("starting request")
```
leighmcculloch, line 60: I changed these lines to match below. When I arrived on this code it was pleasant that the two functions were symmetrical. Even though this function doesn't need changing it I changed it so that they look the same to retain that.

PR #2693, opened 2020-06-15
services/horizon: Add read-ahead buffer for decoded LedgerCloseMeta

2opremio:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add read-ahead buffer for the metadata coming from captive core. Overrides #2664 , which is similar but doesn't contain decoded data structures.

### Why

It improves performance, metadata is decoded in parallel with the rest of the ingestion pipeline.

### Known limitations

N/A

### DIFF
--- a/exp/ingest/ledgerbackend/captive_core_backend.go
+++ b/exp/ingest/ledgerbackend/captive_core_backend.go
@@ -3,7 +3,6 @@ package ledgerbackend
 import (
 	"io"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	"github.com/pkg/errors"
@@ -31,14 +30,18 @@ var _ LedgerBackend = (*captiveStellarCore)(nil)
 
 // TODO: switch from history URLs to history archive interface provided from support package, to permit mocking
 
-// In this (crude, initial) sketch, we replay ledgers in blocks of 17,280
-// which is 24 hours worth of ledgers at 5 second intervals.
-const ledgersPerProcess = 17280
-const ledgersPerCheckpoint = 64
+const (
+	// In this (crude, initial) sketch, we replay ledgers in blocks of 17,280
+	// which is 24 hours worth of ledgers at 5 second intervals.
+	ledgersPerProcess    = 17280
+	ledgersPerCheckpoint = 64
 
-// The number of checkpoints we're willing to scan over and ignore, without
-// restarting a subprocess.
-const numCheckpointsLeeway = 10
+	// The number of checkpoints we're willing to scan over and ignore, without
+	// restarting a subprocess.
+	numCheckpointsLeeway = 10
+
+	readAheadBufferSize = 2
+)
 
 func roundDownToFirstReplayAfterCheckpointStart(ledger uint32) uint32 {
 	v := (ledger / ledgersPerCheckpoint) * ledgersPerCheckpoint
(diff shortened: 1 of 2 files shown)


exp/ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -300,6 +362,25 @@ func (c *captiveStellarCore) Close() error {
...
+		close(c.stop)
+		// discard pending data in case the goroutine is blocked writing to the channel
+		select {
+		case <-c.errC:
+		case <-c.metaC:
+		default:
+		}
+		// Do not close the other channels until we know
+		// the goroutine is done
+		c.wait.Wait()
```
bartekn, line 375: I think a deadlock is still possible: 
```go
			meta, err := c.readLedgerMetaFromPipe()
			if err != nil {
				c.errC <- err
				return
			}
			c.metaC <- meta // <-- read in `select` in Close()
			atomic.AddUint32(&c.readBufferOccupation, 1)
			seq, err := peekLedgerSequence(meta)
			if err != nil {
				c.errC <- err // <-- here, c.errC is waiting for a corresponding read and c.wait.Wait() also waits because go routine function did not return.
				return
			}
			if seq >= untilSequence {
				// we are done
				return
			}
```
I think this can be solved in two ways:
1. Don't use `untilSequence`. `Close()` is called from `GetLedger` when the last ledger is read.
2. Use a communication struct with meta and error. Less work with a single channel and, because you always end go routine after writing to err channel, it shouldn't change anything.

2opremio: > Don't use untilSequence

This is not really an option, since you will get an EOF parsing error if you read more meta structs than you should

> Use a communication struct with meta and error

Yeah, this makes sense. Thanks for the feedback

exp/ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -158,9 +169,66 @@ func (c *captiveStellarCore) openOfflineReplaySubprocess(nextLedger, lastLedger
...
+func (c *captiveStellarCore) sendLedgerMeta(untilSequence uint32) {
+	defer c.wait.Done()
+	printBufferOccupation := time.NewTicker(5 * time.Second)
+	defer printBufferOccupation.Stop()
+	for {
+		select {
+		case <-c.stop:
+			return
+		case <-printBufferOccupation.C:
+			log.Debug("captive core read-ahead buffer occupation:", atomic.LoadUint32(&c.readBufferOccupation))
```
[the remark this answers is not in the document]
2opremio: Yes, good point, I will use `len()`

I think it will be useful when we speed the code up, we may need a larger buffer

exp/ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -300,6 +362,25 @@ func (c *captiveStellarCore) Close() error {
...
+		// Do not close the other channels until we know
+		// the goroutine is done
+		c.wait.Wait()
+	}
+	if c.metaC != nil {
+		close(c.metaC)
+	}
+	if c.errC != nil {
+		close(c.errC)
+	}
```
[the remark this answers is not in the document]
2opremio: Yep

exp/ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -158,9 +169,66 @@ func (c *captiveStellarCore) openOfflineReplaySubprocess(nextLedger, lastLedger
 	c.nextLedger = roundDownToFirstReplayAfterCheckpointStart(nextLedger)
 	c.nextLedgerMutex.Unlock()
 	c.lastLedger = &lastLedger
+
+	// read-ahead buffer
+	c.metaC = make(chan *xdr.LedgerCloseMeta, 2)
```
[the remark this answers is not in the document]
2opremio: Yep. Done!

exp/ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -300,6 +372,19 @@ func (c *captiveStellarCore) Close() error {
 	c.nextLedger = 0
 	c.nextLedgerMutex.Unlock()
 
+	if c.stop != nil {
+		close(c.stop)
+		// discard pending data in case the goroutine is blocked writing to the channel
+		select {
+		case <-c.metaC:
+		default:
+		}
```
bartekn, line 381: Is it still needed? Looks like it's no longer possible that go routine will wait on c.metaC if c.stop is closed. We can either remove the highlighted lines or remove duplicated selects in go routine. When it comes to me I think removing selects from go routine would make the code more readable.

2opremio: `Close` is a public method, so I would rather make it futureproof.

PR #2694, opened 2020-06-15
exp/ingest/ledgerbackend: Captive Core: Online mode

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Adds online mode to `CaptiveStellarCore` backend:
* The `PrepareRange` method is now used to initialize an online mode when `to` parameter is equal `0`. It also detects when the range is already prepared: in such case it doesn't wait for the buffer to be non-empty.
* `GetLedger` no longer starts Stellar-Core process. Users should always call `PrepareRange` to define the mode (online/offline) and range they want to fetch.
* `stellarCoreRunner` has a new `catchup` method. It runs `new-db`, `catchup [X-1]/0` and `run` sequentially to make Stellar-Core start streaming ledgers starting at ledger `X`.
* Updated `GetLatestLedgerSequence` to not use history archives. It now only works when `PrepareRange` has been called and returns `lastLedger` sequence in the offline mode and, currently, `nextLedger` or `nextLedger - 1` in the online mode (see below).
* Updated ingestion state machine to use `PrepareRange` methods when required.

This is still experimental code but works! We should fix fundamental issues in this PR and merge it (after #2693) but work on new features and polish it in another PRs.

### Known limitations

* This PR is using closed buffered buffer PR. Will be replaced with #2693 when merged.
* This is still an experiment and some things need to be changed. Here's the initial list of issues to work on in separate PRs (will create GH issues and add more once this PR is merged):
    * It always starts stellar-core with an empty DB. It's possible to resume with existing state but complicates the code.
    * Make sure distributed ingestion works and is not blocking when it shouldn't. Ex. now, if one Stellar-Core is out of sync `resume` state will block entire cluster.
    * Update `GetLatestLedgerSequence` to return the latest sequence in a buffer (when #2693 is merged).
    * Move peekLedgerSequence to LedgerCloseMeta (possibly part of #2633).
    * Detect when stellar-core exits and add graceful shutdown. I noticed that sometimes the stellar-core process is not correctly closed when shutting down.
    * Add config option to set a path to `stellar-core.cfg` file, after https://github.com/stellar/stellar-core/pull/2574 is released.
    * Add tests.


exp/ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -99,28 +124,113 @@ func (r *stellarCoreRunner) writeConf() error {
...
 
-func (r *stellarCoreRunner) run(from, to uint32) error {
+func (r *stellarCoreRunner) catchup(from, to uint32) error {
 	err := r.writeConf()
 	if err != nil {
 		return errors.Wrap(err, "error writing configuration")
 	}
 
+	// new-db
+	cmd := exec.Command(r.executablePath, []string{"--conf", r.getConfFileName(), "new-db"}...)
```
abuiles, line 134: @bartekn shall we add a check that core db exist in the db? is okay if we don't -- we just need to expose the error so people know `new-db` failed because the db doesn't exist

exp/ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -28,39 +31,61 @@ type stellarCoreRunner struct {
...
-		"NODE_IS_VALIDATOR=false",
-		"DISABLE_XDR_FSYNC=true",
-		"UNSAFE_QUORUM=true",
 		fmt.Sprintf(`NETWORK_PASSPHRASE="%s"`, r.networkPassphrase),
 		fmt.Sprintf(`BUCKET_DIR_PATH="%s"`, filepath.Join(r.getTmpDir(), "buckets")),
 		fmt.Sprintf(`METADATA_OUTPUT_STREAM="%s"`, r.getPipeName()),
+		// "RUN_STANDALONE=true",
+		"NODE_IS_VALIDATOR=false",
+		// "DISABLE_XDR_FSYNC=true",
+		`DATABASE="postgresql://dbname=core host=localhost user=Bartek"`,
```
abuiles, line 47: @bartekn for local, can we drop the user and let  pg use the default?

bartekn: We will allow using custom config files when https://github.com/stellar/stellar-core/pull/2574 is released.

abuiles: ha ok! I mean, just so we don't have to update this value to test locally.  anyways, not an issue since we'll clean it up later.

exp/ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -135,87 +117,154 @@ func (c *captiveStellarCore) copyLedgerCloseMeta(xlcm *xdr.LedgerCloseMeta, lcm
...
+			break
+		}
+		time.Sleep(time.Second)
+	}
+
+	return nil
+}
+
+// GetLedger returns true when ledger is found and it's LedgerCloseMeta.
+// Call `PrepareRange` first to instruct the backend which ledgers to fetch.
```
abuiles, line 249: @bartekn what happens if we forget to run `PrepareRange` does `GetLedger` fail?

bartekn: Yes, `GetLedger` will return `session is closed, call PrepareRange first` error.

bartekn:
Thanks @abuiles! I merged `captive-core` -> `captive-core-online` -> `captive-core-online-experiment` (this PR branch) so it contains #2693 now.

 @tamirms @2opremio can you check if there are any major issues before we merge this?

exp/ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -152,44 +134,113 @@ func (c *captiveStellarCore) copyLedgerCloseMeta(xlcm *xdr.LedgerCloseMeta, lcm
...
+		c.historyURLs[0],
+		historyarchive.ConnectOptions{},
+	)
+	if err != nil {
+		return 0, errors.Wrap(err, "error connecting to history archive")
+	}
+	has, e := archive.GetRootHAS()
 	if e != nil {
-		return errors.Wrap(e, "getting latest ledger sequence")
+		return 0, errors.Wrap(err, "error getting root HAS")
```
tamirms, line 147: should we be wrapping `e` instead of `err`?

2opremio:
Sure thing. I will take a look later on.

exp/ingest/ledgerbackend/stellar_core_runner.go
tamirms: does this assume we'll always be running a postgres instance on the same machine as horizon? can we use sqlite instead?

exp/ingest/ledgerbackend/captive_core_backend.go
bartekn: Good catch! Fixed in https://github.com/stellar/go/pull/2694/commits/d5d6e423b1becc9f5eb27b2c6da8790ff3d1cbae. Would be great to have https://github.com/dominikh/go-tools/issues/529.

exp/ingest/ledgerbackend/stellar_core_runner.go
bartekn: No, users can use postgres and sqlite. Before https://github.com/stellar/stellar-core/pull/2574 we need to build config inside the code but later we can allow users to set any config they like and just pass meta stream output via cli param.

bartekn:
@2opremio I merged it without your 👍. Feel free to add some comments (if any) here.

exp/ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -91,32 +114,97 @@ func (r *stellarCoreRunner) getTmpDir() string {
...
 
+	// new-db
+	cmd := exec.Command(r.executablePath, []string{"--conf", r.getConfFileName(), "new-db"}...)
+	err = cmd.Start()
+	if err != nil {
+		return errors.Wrap(err, "error starting `stellar-core new-db` subprocess")
+	}
+	cmd.Dir = r.getTmpDir()
+	cmd.Stdout = r.getLogLineWriter()
+	cmd.Stderr = r.getLogLineWriter()
```
2opremio, line 139: I think you could use the same writer for both (assuming stellar core writes line by line, which it should).

On top of that, the line writer goroutines should be stopped on `Close()` (it didn't matter for debugging, when it was commented out, but now they will cause a memory leak).

exp/ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -91,32 +114,97 @@ func (r *stellarCoreRunner) getTmpDir() string {
...
+}
+
+func (r *stellarCoreRunner) runFrom(from uint32) error {
+	err := r.writeConf()
+	if err != nil {
+		return errors.Wrap(err, "error writing configuration")
+	}
+
+	// new-db
+	cmd := exec.Command(r.executablePath, []string{"--conf", r.getConfFileName(), "new-db"}...)
```
2opremio, line 167: could you factor out spawning a command? We use the same pattern 4 times

exp/ingest/ledgerbackend/database_backend.go
```diff
@@ -49,13 +49,15 @@ func (dbb *DatabaseBackend) PrepareRange(from uint32, to uint32) error {
 		return errors.New("`from` ledger does not exist")
 	}
 
-	toExists, _, err := dbb.GetLedger(to)
-	if err != nil {
-		return errors.Wrap(err, "error getting ledger")
-	}
+	if to != 0 {
```
2opremio, line 52: Could you document the special meaning of `0`.

Even better, would be using something more self-explanatory and explicit instead of using a special value (e.g. a specialized method called `PrepareFullRangeFrom(from uint32)`)

exp/ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -35,40 +36,62 @@ type stellarCoreRunner struct {
...
+HISTORY="curl -sf http://history.stellar.org/prd/core-testnet/core_testnet_002/{0} -o {1}"
+
+[[VALIDATORS]]
+NAME="sdf_testnet_3"
+HOME_DOMAIN="testnet.stellar.org"
+PUBLIC_KEY="GC2V2EFSXN6SQTWVYA5EPJPBWWIMSD2XQNKUOHGEKB535AQE2I6IXV2Z"
+ADDRESS="core-testnet3.stellar.org"
+HISTORY="curl -sf http://history.stellar.org/prd/core-testnet/core_testnet_003/{0} -o {1}"
+
+`,
```
2opremio, line 73: Is this included in the production configuration file?

If so, I think it should be removed.

Either way, I think this shouldn't be part of the horizon binary. It would be fine in a `_test` file or test package.

bartekn: I explained it here: https://github.com/stellar/go/pull/2694#discussion_r440334391 and here https://github.com/stellar/go/issues/2706.

exp/ingest/ledgerbackend/database_backend.go
bartekn: https://github.com/stellar/go/issues/2711

exp/ingest/ledgerbackend/stellar_core_runner.go
bartekn: I'll fix it in https://github.com/stellar/go/pull/2673#issuecomment-645968745.

exp/ingest/ledgerbackend/stellar_core_runner.go
bartekn: https://github.com/stellar/go/issues/2712

PR #2696, opened 2020-06-16
all: require at least v0.3.3 of golang.org/x/text

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Require at least `v0.3.3` of `golang.org/x/text`.

### Why
According to https://groups.google.com/d/topic/golang-announce/bXVeAmGOqz0 v0.3.3 which was released today contains a security fix for the `golang.org/x/text/transform` package.

The `hash` package within this repository is the only code we have here that I can see is using the `golang.org/x/text/transform` package but only in its tests indirectly through the gomega framework, so there's no real need to do this upgrade other than to make sure that any code we write in the future that either directly or indirectly uses the transform package doesn't unknowingly use the vulnerable version of the dependency in a meaningful way.

### Known limitations

 N/A

### DIFF
--- a/go.list
+++ b/go.list
@@ -129,7 +129,7 @@ golang.org/x/net v0.0.0-20190827160401-ba9fcec4b297
 golang.org/x/oauth2 v0.0.0-20190226205417-e64efc72b421
 golang.org/x/sync v0.0.0-20190423024810-112230192c58
 golang.org/x/sys v0.0.0-20190606165138-5da285871e9c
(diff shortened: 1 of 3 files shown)


leighmcculloch:
Diff for the changes in v0.3.2 to v0.3.3: https://github.com/golang/text/compare/v0.3.2...v0.3.3

PR #2701, opened 2020-06-17
Avoid dumping tx_meta in verify-range until we order it

2opremio:
The idea is to build and push a temporary docker image (`stellar/exp-verify-range:without-tx-meta`) with this changes. The image will be used until we sort tx_meta before inserting it in the DB.


services/horizon/docker/verify-range/start
```diff
@@ -46,7 +46,10 @@ dump_horizon_db() {
...
+	# Temporary workaround: avoid dumping tx_meta. tx_meta is currently not ordered, leading to different orderings from different core instances
+	# FIXME: we should explicitly order tx_meta before inserting it in the DB. Once that is done, we can dump tx_meta again
+	# psql "postgres://postgres:postgres@localhost:5432/horizon?sslmode=disable" -t -A -F"," --variable="FETCH_COUNT=100" -c "select transaction_hash, ledger_sequence, application_order, account, account_sequence, max_fee, operation_count, id, tx_envelope, tx_result, tx_meta, tx_fee_meta, signatures, memo_type, memo, time_bounds, successful, fee_charged from history_transactions order by id asc" > "${1}_transactions"
+	psql "postgres://postgres:postgres@localhost:5432/horizon?sslmode=disable" -t -A -F"," --variable="FETCH_COUNT=100" -c "select transaction_hash, ledger_sequence, application_order, account, account_sequence, max_fee, operation_count, id, tx_envelope, tx_result, tx_fee_meta, signatures, memo_type, memo, time_bounds, successful, fee_charged from history_transactions order by id asc" > "${1}_transactions"
```
bartekn, line 52: We should also remove: `tx_fee_meta`. Canhave the same issue.

PR #2702, opened 2020-06-17
services/horizon: Remove ingest failed transactions flag

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Remove ingest failed transactions flag so that Horizon will always ingest failed transactions

### Why

Having failed transactions in the Horizon DB is useful because it allows us to remove the last remaining dependency on the stellar core DB in the txsub module.

### Known limitations

There might be some operators who prefer to save space by not ingesting failed transactions in the Horizon DB. However, there is a reap system in Horizon which allows you to remove older rows in the history tables. Using horizon reap should satisfy the same need.


services/horizon/internal/txsub/results/db/main.go
```diff
@@ -19,16 +15,12 @@ import (
 type DB struct {
 	Core    *core.Q
```
abuiles, line 16: We can remove now `	Core    *core.Q` and the import :)

services/horizon/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ file. This project adheres to [Semantic Versioning](http://semver.org/).x
 
 * Add transaction set operation count to `history_ledger`([#2690](https://github.com/stellar/go/pull/2690)).
 Extend ingestion to store the total number of operations in the transaction set and expose it in the ledger resource via `tx_set_operation_count`. This feature allow you to assess the used capacity of a transaction set.
+* Remove `--ingest-failed-transactions` flag. From now on Horizon will always ingest failed transactions ([#2702](https://github.com/stellar/go/pull/2702)).
```
bartekn, line 10:
```suggestion
* Remove `--ingest-failed-transactions` flag. From now on Horizon will always ingest failed transactions. WARNING: if your application is using Horizon DB directly (not recommended!) remember that now it will also contain failed txs. ([#2702](https://github.com/stellar/go/pull/2702)).
```

services/horizon/internal/txsub/results/db/main.go
```diff
@@ -3,32 +3,22 @@
...
 
 var _ txsub.ResultProvider = &DB{}
 
 // ResultByHash implements txsub.ResultProvider
 func (rp *DB) ResultByHash(ctx context.Context, hash string) txsub.Result {
-	historyLatest := ledger.CurrentState().HistoryLatest
-
 	// query history database
 	var hr history.Transaction
 	err := rp.History.TransactionByHash(&hr, hash)
```
abuiles, line 24: @tamirms do we allow tx submission if Horizon is catching up with stellar-core?  could this potentially cause a hash not to be found (since it hasn't been ingested)?

tamirms: that's a good point. I think this issue you've spotted was already there before this PR so we can try to tackle it in another PR

PR #2710, opened 2020-06-18
exp/ingest/io: Replace ledgerbackend.LedgerCloseMeta with xdr.LedgerCloseMeta

tamirms:
<details>
  <summary>PR Checklist</summary>
  
### PR Structure

* [x] This PR has reasonably narrow scope (if not, break it down into smaller PRs).
* [x] This PR avoids mixing refactoring changes with feature changes (split into two PRs
  otherwise).
* [x] This PR's title starts with name of package that is most changed in the PR, ex.
  `services/friendbot`, or `all` or `doc` if the changes are broad or impact many
  packages.

### Thoroughness

* [ ] This PR adds tests for the most critical parts of the new functionality or fixes.
* [ ] I've updated any docs ([developer docs](https://www.stellar.org/developers/reference/), `.md`
  files, etc... affected by this change). Take a look in the `docs` folder for a given service,
  like [this one](https://github.com/stellar/go/tree/master/services/horizon/internal/docs).

### Release planning

* [ ] I've updated the relevant CHANGELOG ([here](services/horizon/CHANGELOG.md) for Horizon) if
  needed with deprecations, added features, breaking changes, and DB schema changes.
* [ ] I've decided if this PR requires a new major/minor version according to
  [semver](https://semver.org/), or if it's mainly a patch change. The PR is targeted at the next
  release branch if it's not a patch change.
</details>

### What

Replace ledgerbackend.LedgerCloseMeta with xdr.LedgerCloseMeta.

Fixes https://github.com/stellar/go/issues/2633

### Why

> When exp/ingest prototype was created Stellar-Core didn't have LedgerCloseMeta type so we created a temporary LedgerCloseMeta in ledgerbackend package. Now, when the required type is there, we should remove ledgerbackend.LedgerCloseMeta and replace it with xdr.LedgerCloseMeta.

### Known limitations

[N/A]

PR #2720, opened 2020-06-19
exp/ingest/io: Fix bug in LedgerTransactionReader

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

https://github.com/stellar/go/pull/2710 introduced a bug in LedgerTransactionReader which had the effect that transactions were not returned in the correct order.

The bug occurs because transactions in `xdr.LedgerCloseMeta.V0.TxSet` are not ordered in the same way as `xdr.LedgerCloseMeta.V0.TxProcessing `. `TxProcessing` is sorted by the order in which transactions are applied to the ledger. `TxSet` is sorted lexicographically by transaction hash.

This commit ensures that each `LedgerTransaction` has the correct transaction envelope / transaction result pair.

### DIFF
--- a/exp/ingest/io/ledger_transaction_reader.go
+++ b/exp/ingest/io/ledger_transaction_reader.go
@@ -1,9 +1,11 @@
 package io
 
 import (
+	"encoding/hex"
 	"io"
 
 	"github.com/stellar/go/exp/ingest/ledgerbackend"
+	"github.com/stellar/go/network"
 	"github.com/stellar/go/support/errors"
 	"github.com/stellar/go/xdr"
 )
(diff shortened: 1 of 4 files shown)


exp/ingest/io/ledger_transaction_reader.go
```diff
@@ -60,16 +64,32 @@ func (reader *LedgerTransactionReader) Rewind() {
...
+			hexHash := hex.EncodeToString(result.TransactionHash[:])
+			return errors.Errorf("unknown tx hash in LedgerCloseMeta: %v", hexHash)
+		}
 		reader.transactions = append(reader.transactions, LedgerTransaction{
 			Index:      uint32(i + 1), // Transactions start at '1'
-			Envelope:   lcm.V0.TxSet.Txs[i],
-			Result:     lcm.V0.TxProcessing[i].Result,
+			Envelope:   envelope,
+			Result:     result,
 			Meta:       lcm.V0.TxProcessing[i].TxApplyProcessing,
```
bartekn, line 88: I think we need the same thing for meta & fee and a test checking this is correct.

tamirms: shouldn't meta and fee have the same index as result? the code seems correct to me

bartekn: You are right, sorry!

Issue #2552, opened 2020-05-05
Fast txmeta: implement and profile parallel ingestion


ire-and-curses:
We know that using AWS Batch we can do a complete historical ingestion x2 + verification in ~20 hrs (for `horizon-cmp` testing). However in this case we don’t write to a shared database. Adding this component also adds negatives to the AWS Batch solution:

1. Adding the ability to write to a shared database would be a significant change that is probably hard to do effectively using AWS Batch.
2. We'd like the offline parallel solution to be generic if possible, rather than tied to any one cloud architecture.
3. We don’t know when we get bottlenecked on DB writing.

Assuming we are not going with an off-the-shelf solution, then the parallel orchestration needs to happen on the Horizon side. We can start simply, without trying to optimise for historical ledger size changes.

- [x] Discuss any alternatives to the path forward sketched out below
- [x] Create a new Horizon “--parallel” option which spawns multiple FSC instances for bulk historical catchup
- [x] Do some experimenting to determine i) what a reasonable batch size should be (100,000 is a first guess); ii) where the bottlenecks might be (Horizon CPU? Postgres DB write throughput? etc.); iii) what the limits of parallelisation are (how many workers can be efficiently supported?)
- [x] Confirm reliability of the parallel implementation and data integrity under scenarios such as workers disappearing, hanging or crashing
- [x] Document all of the above, including memory requirements

Meta issue: #2550

bartekn:
In Horizon team meeting we decided to try it without using AWS Batch first. If a single instance solution is not fast enough or in case of any issues we can explore AWS Batch again.

https://github.com/stellar/go/pull/2322 is in a state that can be run and tested with `horizon db reingest range` command. Let's see if it actually can reingest ledgers correctly and how fast it is.

I think this can be broken down into the following steps (please feel free to change/reorder):

1. Research which EC2 instance type we want to use to reingest pubnet history and database instance params. `x1.32xlarge` was suggested as a Horizon machine but maybe there's something better suited for this job. Remember that captive core requires ~2GB for the latest ledger state and that we aim for <24h total ingestion time. Let's discuss the choice here first before we continue.
2. Start as many workers as possible, when reingestion is complete dump `history_*` tables (see script in `verify-range` docker image).
3. Repeat the step above but with a DB backend.
4. Ensure `history_*` tables are exactly the same. Compare running times.

> Adding this component also adds negatives to the AWS Batch solution

Even if a single machine prototype works well I think AWS Batch offers a few advantages:
[...]

PR #2724, opened 2020-06-22
services/horizon: Parallelize `db reingest range`

2opremio:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This change breaks down the ledger range to reingest in subranges which are submitted to a pre-defined number of workers, processing the subranges in parallel.

For now, the workers are simply Go routines using their own `System` (with their own DB connections etc ...).

In the future workers could be fully fledged Horizon instances running in multiple machines (e.g. orchestrated through Kubernetes Jobs or AWS Batch Jobs).

New flags:
--parallel-workers: [optional] if this flag is set to > 1, horizon will parallelize reingestion using the supplied number of workers
--parallel-job-size: [optional] parallel workers will run jobs processing ledger batches of the supplied size

### Why

We want reingestion to be faster. Addresses #2552 

### Known limitations

As I mentioned, this change only applies to a single machine. Running it in multiple machines will require further work.

This PR is targetting `release-1.5.0`  (since that's the base branch I used). We should retarget it to master (once `release-1.5.0` is merged) before merging.


services/horizon/internal/expingest/main.go
```diff
@@ -403,3 +419,145 @@ func isCancelledError(err error) bool {
...
+	}
+
+	// Also, round the batch size to the closest, lower or equal 64 multiple
+	return (batchSize / historyCheckpointLedgerInterval) * historyCheckpointLedgerInterval
+}
+
+func (ps *ParallelSystems) ReingestRange(fromLedger, toLedger uint32, batchSizeSuggestion uint32) error {
+	batchSize := calculateParallelLedgerBatchSize(toLedger-fromLedger, batchSizeSuggestion, ps.workerCount)
+	pendingJobsCount := 0
+	var result error
```
bartekn, line 512: I think `err` better describes this variable.
```suggestion
	var err error
```

services/horizon/internal/expingest/main.go
```diff
@@ -403,3 +419,145 @@ func isCancelledError(err error) bool {
...
+		subRange := ledgerRange{subRangeFrom, subRangeTo}
+
+		select {
+		case <-ps.shutdown:
+			return errors.New("aborted")
+		case subRangeResult := <-ps.reingestJobResult:
+			processSubRangeResult(subRangeResult)
+		case ps.reingestJobQueue <- subRange:
+			pendingJobsCount++
+			subRangeFrom = subRangeTo
```
bartekn, line 538: Don't we ingest last ledger in each subrange twice now?
```suggestion
			subRangeFrom = subRangeTo+1
```

services/horizon/internal/expingest/main.go
```diff
@@ -403,3 +419,145 @@ func isCancelledError(err error) bool {
...
+			subRangeTo = toLedger
+		}
+		subRange := ledgerRange{subRangeFrom, subRangeTo}
+
+		select {
+		case <-ps.shutdown:
+			return errors.New("aborted")
+		case subRangeResult := <-ps.reingestJobResult:
+			processSubRangeResult(subRangeResult)
+		case ps.reingestJobQueue <- subRange:
```
bartekn, line 536: It was hard to me to understand the code because subrange calculations are happening even when not needed. Maybe we can move this to a one-liner like:
```suggestion
		case ps.reingestJobQueue <- ledgerRange{from: subRangeFrom, to: min(subRangeFrom + batchSize, toLedger)}:
```

services/horizon/internal/expingest/main.go
```diff
@@ -403,3 +419,145 @@ func isCancelledError(err error) bool {
...
+	for pendingJobsCount > 0 {
+		// wait for any remaining running jobs to finish
+		select {
+		case <-ps.shutdown:
+			return errors.New("aborted")
+		case subRangeResult := <-ps.reingestJobResult:
+			processSubRangeResult(subRangeResult)
+		}
+
+	}
```
bartekn, line 551: I think this method is a bit tricky and sometimes hard to follow. Maybe we can start 2 go routines: one writing to a queue and another one checking the results? I think it will simplify the code a lot.

services/horizon/internal/expingest/main.go
2opremio: Oh, that is true, good catch ... although ... shouldn't we start and end in checkpoint limits? (I got _inspiration_ from verify-range which I think uses the same approach I use here).

services/horizon/internal/expingest/main.go
2opremio: I will give it a try. I suspect using two goroutines is going to make synchronization harder.

services/horizon/internal/expingest/main.go
```diff
@@ -403,3 +419,145 @@ func isCancelledError(err error) bool {
...
+	defer func() {
+		s.Shutdown()
+		ps.wait.Done()
+	}()
+	for {
+		select {
+		case <-ps.shutdown:
+			return
+		case reingestRange := <-ps.reingestJobQueue:
+			err := s.ReingestRange(reingestRange.from, reingestRange.to, false)
```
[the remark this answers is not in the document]
2opremio: What's the alternative?

services/horizon/CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ file. This project adheres to [Semantic Versioning](http://semver.org/).x
 
 ## Unreleased
 
+* Add `--parallel-workers` to `horizon db reingest range`. `--parallel-workers`  will parallelize reingestion using the supplied number of workers cores to
```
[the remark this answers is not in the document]
2opremio: Huh, odd. Will fix

services/horizon/internal/expingest/main.go
bartekn: This isn't necessary when reingesting history. We need to start and end with a checkpoint ledger in verify range because we need to build and then verify the state. Since history ingestion does not update state it's not needed.

services/horizon/internal/expingest/main.go
```diff
@@ -403,3 +419,145 @@ func isCancelledError(err error) bool {
...
+	from uint32
+	to   uint32
+}
+
+type rangeResult struct {
+	err            error
+	requestedRange ledgerRange
+}
+
+type ParallelSystems struct {
```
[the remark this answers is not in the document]
2opremio: I did it, but it's odd that a separate file imports symbols from `main.go`

services/horizon/internal/expingest/main.go
2opremio: Well, that doesn't really work because `min()` doesn't exist and you need to use `min(subRangeFrom+batchSize, toLedger)` in the body of the `case` (`subRangeTo`). So, I don't think that a real solution makes it easier to read.

bartekn: Yes, the assumption was to add `min` function but if you are planning to do https://github.com/stellar/go/pull/2724#discussion_r444326321 we can ignore this comment.

services/horizon/internal/expingest/main.go
2opremio: Actually, you were right. I think it's a bit cleaner with 2 subroutines.

services/horizon/internal/expingest/main.go
2opremio: After the refactoring you suggested below, it's closer to what you suggest.

2opremio:
@bartekn PTAL. I addressed your comments, reincorporated `--parallel-job-size` (which will be needed for performance reasons when reingesting the full history) and added an extra test (`TestParallelReingestRangeError()`)

services/horizon/internal/expingest/parallel.go
```diff
@@ -0,0 +1,156 @@
...
+		wait      sync.WaitGroup
+		stop      = make(chan struct{})
+		batchSize = calculateParallelLedgerBatchSize(toLedger-fromLedger, batchSizeSuggestion, ps.workerCount)
+		totalJobs = uint32(math.Ceil(float64(toLedger-fromLedger) / float64(batchSize)))
+	)
+
+	wait.Add(1)
+	defer func() {
+		close(stop)
+		wait.Wait()
```
tamirms, line 108: why is it necessary to wait after we have closed `stop`?

2opremio: Because you don't want to return while the goroutine is still executing (it could still be in the middle of a job when stop is closed)

services/horizon/internal/expingest/parallel.go
```diff
@@ -0,0 +1,157 @@
...
+		}
+		result.wait.Add(1)
+		go result.work(s)
+	}
+	return &result, nil
+}
+
+func (ps *ParallelSystems) work(s System) {
+	defer func() {
+		s.Shutdown()
```
tamirms, line 59: is it necessary to call `s.Shutdown()` it seems like the only way `work()` can exit is if the shutdown channel is closed

services/horizon/internal/expingest/parallel.go
```diff
@@ -0,0 +1,157 @@
...
+		// collect results
+		select {
+		case <-ps.shutdown:
+			return errors.New("aborted")
+		case subRangeResult := <-ps.reingestJobResult:
+			if subRangeResult.err != nil {
+				// TODO: give account of what ledgers were correctly reingested?
+				errMsg := fmt.Sprintf("in subrange %d to %d",
+					subRangeResult.requestedRange.from, subRangeResult.requestedRange.to)
+				return errors.Wrap(subRangeResult.err, errMsg)
```
2opremio, line 141: This is not correct, it doesn't wait for pending jobs to be finished. @bartekn I will see if I can fix it elegantly but I think it will complicate things enough to restore my original solution.

2opremio: I really don't think it's simpler anymore

services/horizon/internal/expingest/parallel.go
2opremio: @tamirms That is a call to shut down the underlying dedicated system of the worker (it's just a cleanup call). Are we referring to the same thing?

tamirms: oh sorry, I was confusing the `ParallelSystems` instance with the `System` instance

2opremio: No worries!

services/horizon/internal/expingest/parallel.go
```diff
@@ -0,0 +1,170 @@
...
+					subRangeResult.requestedRange.from, subRangeResult.requestedRange.to)
+				firstErr = errors.Wrap(subRangeResult.err, errMsg)
+			}
+			pendingJobsMutex.Unlock()
+		}
+
+	}
+
+	return firstErr
+}
```
bartekn, line 161: I think we can simplify it more:
* A single shutdown signal should be enough to handle everything.
* We don't really need `reingestJobResult` this can be handled inside a worker function.
* We don't need _global_ wait group. `Shutdown` should just sent a shutdown signal and make `ReingestRange` method to wrap up.
* We don't need to start workers when creating a method. Maybe we'll need workers in other methods than `ReingestRange` but now we don't.

After implementing the changes I have the following code:
```go
func (ps *ParallelSystems) ReingestRange(fromLedger, toLedger uint32, batchSizeSuggestion uint32) error {
    var reingestJobQueue = make(chan ledgerRange)
    var erroredMutex sync.Mutex
    var errored bool

    func markError() {
        erroredMutex.Lock()
        errored = true
        erroredMutex.Unlock()
    }

    wg.Add(1)
    // can be moved to a method
    go func() {
        defer wg.Done()
        for subRangeFrom := fromLedger; subRangeFrom < toLedger; {
            // job queuing
            subRangeTo := subRangeFrom + (batchSize - 1) // we subtract one because both from and to are part of the batch
            if subRangeTo > toLedger {
                subRangeTo = toLedger
            }
            select {
            case <-ps.shutdown:
                return
            case reingestJobQueue <- ledgerRange{subRangeFrom, subRangeTo}:
            }
            subRangeFrom = subRangeTo + 1
        }
    }()

    for i := 0; i < workers; i++ {
        wg.Add(1)
        // can be moved to a method
        go func() {
            defer wg.Done()
            s, err := systemFactory(config)
            if err != nil {
                log.Error("...")
                s.Shutdown()
                markError()
                return
            }
            for {
                select {
                case <-ps.shutdown:
                    return
                case reingestRange := <-reingestJobQueue:
                    err := s.ReingestRange(reingestRange.from, reingestRange.to, false)
                    if err != nil {
                        log.Error("...")
                        s.Shutdown()
                        markError()
                        return
                    }
                }
            }
        }
    }

    wg.Wait()
    close(reingestJobQueue)

    if errored {
        return errors.New("one or more jobs failed")
    }
    return nil
}

func (ps *ParallelSystems) shutdown() {
    close(ps.shutdown)

}

// Close should be called when reading is finished.
func (ps *ParallelSystems) Shutdown() error {
	ps.shutdownOnce.Do(msr.shutdown)
	return nil
}

```

2opremio: > * A single shutdown signal should be enough to handle everything.
Not if you have persistent workers

> * We don't really need `reingestJobResult` this can be handled inside a worker function.

Your example has oversimplified error management. If you want to give account of up until what point things where ingested properly (which I will implement shortly after this PR is merged), you need to know what ranges where successful.

> * We don't need _global_ wait group. `Shutdown` should just sent a shutdown signal and make `ReingestRange` method to wrap up.

Not if you have persistent workers

> * We don't need to start workers when creating a method. Maybe we'll need workers in other methods than `ReingestRange` but now we don't.

I did it this way to that we can parallelize stress testing and normal history reingestion later on. 

> After implementing the changes I have the following code

I am not sure how `msr` or `shutdownOnce` are handled. Also, error management is oversimplified. As I mentioned above, you do need to consume the results to know what ranges where processed.

2opremio: @bartekn and I agreed that he will commit his suggestion and we will take it from there.

bartekn:
OK, I pushed my code as discussed with @fons. I haven't updated tests, will do it after 👍 from you. I also added a code that return a suggested range to restart a job in case of failure. There's a comment above `lowestRangeErr` explaining why (I think) it works without having to keep track of results from all the jobs.

services/horizon/internal/expingest/parallel.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+	return nil
+}
+
+func (ps *ParallelSystems) doShutdown() {
+	close(ps.shutdown)
+
+}
+
+func (ps *ParallelSystems) Shutdown() error {
```
2opremio, line 159: I don't think it makes sense to have this anymore

services/horizon/internal/expingest/parallel.go
```diff
@@ -0,0 +1,162 @@
...
+			if err != nil {
+				log.WithError(err).Error("error in reingest worker")
+				if rangeErr, ok := err.(rangeError); ok {
+					lowestRangeErrMutex.Lock()
+					if lowestRangeErr == nil || (lowestRangeErr != nil && lowestRangeErr.ledgerRange.from > rangeErr.ledgerRange.from) {
+						lowestRangeErr = &rangeErr
+					}
+					lowestRangeErrMutex.Unlock()
+				}
+				ps.Shutdown()
```
2opremio, line 139: It's awkward that an operation shutdowns the whole `ParallelSystems`. I would remove the global shutdown channel, create one here and pass it to the workers.

services/horizon/internal/expingest/parallel.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+
+	return &ParallelSystems{
+		config:      config,
+		workerCount: workerCount,
+		shutdown:    make(chan struct{}),
+	}, nil
+}
+
+func (ps *ParallelSystems) reingestWorker(reingestJobQueue <-chan ledgerRange) error {
```
2opremio, line 54: I would use a verb `reingestWorker` (e.g. `runReingestWorker()` `doReingestWork()`)

services/horizon/internal/expingest/parallel.go
```diff
@@ -0,0 +1,162 @@
...
+		// jobs are sent to a queue (unbuffered channel) in sequence and there is a WaitGroup waiting for all the workers
+		// to exit.
+		// Because of this when we reach `wg.Wait()` all jobs previously sent to a channel are processed (either success
+		// or failure). In case of a failure we save the range with the smallest sequence number because this is where
+		// the user needs to start again to prevent the gaps.
+		lowestRangeErr *rangeError
+	)
+
+	wg.Add(1)
+	go func() {
```
2opremio, line 108: I don't think there is a reason to do this in a separate goroutine. You could run the inner for loop after spawning the workers.

services/horizon/internal/expingest/parallel.go
2opremio: In fact, having the ability to shutdown an in-progress reingestion will break error reporting (there won't be any guarantee of the lowest failing ledger range being reported)

services/horizon/internal/expingest/parallel.go
```diff
@@ -0,0 +1,162 @@
...
+
+		lowestRangeErrMutex sync.Mutex
+		// lowestRangeErr is an error of the failed range with the lowest starting ledger sequence that is used to tell
+		// the user which range to reingest in case of errors. We use that fact that System.ReingestRange is blocking,
+		// jobs are sent to a queue (unbuffered channel) in sequence and there is a WaitGroup waiting for all the workers
+		// to exit.
+		// Because of this when we reach `wg.Wait()` all jobs previously sent to a channel are processed (either success
+		// or failure). In case of a failure we save the range with the smallest sequence number because this is where
+		// the user needs to start again to prevent the gaps.
+		lowestRangeErr *rangeError
```
2opremio, line 104: this doesn't need to be a pointer (no strong opinions)

services/horizon/internal/expingest/parallel.go
```diff
@@ -0,0 +1,162 @@
...
+	}()
+
+	for i := uint(0); i < ps.workerCount; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			err := ps.reingestWorker(reingestJobQueue)
+			if err != nil {
+				log.WithError(err).Error("error in reingest worker")
+				if rangeErr, ok := err.(rangeError); ok {
```
2opremio, line 132: This casting isn't necessary if the worker always returns a range error. If you want to distinguish system creation errors, just create the system here.

2opremio: It will also allow to always report the range in which an error happens

services/horizon/internal/expingest/parallel.go
bartekn: Closing shutdown channel won't abort running jobs. If there's a running `ReingestRange` method in any worker it will complete (succeed or fail but will complete). Also, because the queue channel is unbuffered you won't have any remaining jobs in a buffer (the go routine adding new jobs will return due to `<- ps.shutdown`). Even if this was true, all jobs will complete because there's a wait group that blocks until all go routines return.

Can you elaborate?

services/horizon/internal/expingest/parallel.go
bartekn: If it wasn't a pointer we'd need another variable to determine if there was an error or not.

services/horizon/internal/expingest/parallel.go
2opremio: I think you are right, but I believe it's a bit brittle.

Leaving that aside, I think we agree that `Shutdown()` is not needed anymore since there is no need to cancel running operations or cleanup anything.

services/horizon/internal/expingest/parallel.go
2opremio: you can use the error inside the `rangeError` (if it's nil then no error happened). But, as I said, no strong opinions here.

bartekn:
@fons thanks for a quick review! I added the requested changes and fixed tests. It's great that we're at the same page when it comes to aborting jobs.

I removed `Shutdown` as requested: I agree it doesn't make sense to keep it now because we're not handling system shutdown signals at all in `db reingest` commands. However, I think it's worth doing this in a future so that users get a status update when they terminate the process manually.

I won't be available in the morning tomorrow so please update the code if there's something that's blocking you from merging it.

services/horizon/internal/expingest/parallel.go
```diff
@@ -0,0 +1,159 @@
...
+	if err != nil {
+		return errors.Wrap(err, "error creating new system")
+	}
+	for {
+		select {
+		case <-stop:
+			return nil
+		case reingestRange, more := <-reingestJobQueue:
+			if !more {
+				// Channel closed - no more jobs
```
2opremio, line 64: There is no need to do this, we have the stop channel for that

services/horizon/internal/expingest/parallel.go
```diff
@@ -0,0 +1,158 @@
...
+	for i := uint(0); i < ps.workerCount; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			err := ps.runReingestWorker(stop, reingestJobQueue)
+			if err != nil {
+				log.WithError(err).Error("error in reingest worker")
+				if rangeErr, ok := err.(rangeError); ok {
+					lowestRangeErrMutex.Lock()
+					if lowestRangeErr == nil || (lowestRangeErr != nil && lowestRangeErr.ledgerRange.from > rangeErr.ledgerRange.from) {
```
2opremio, line 120: There is no need to do `lowestRangeErr != nil ` in the second branch of `||`

2opremio:
I think we are done. Thie PR targets the 1.5.0 branch, but my plan is to hold off from merging to avoid disrupting the 1.5.0 release, and just wait until 1.5.0 is merged into master.

But, if @ire-and-curses / @abuiles are OK with it, I will merge now.

services/horizon/internal/expingest/parallel.go
bartekn: Good point!

ire-and-curses:
Let's target this for 1.6 release.

bartekn:
Added successful range logging in https://github.com/stellar/go/pull/2724/commits/b0205c909ed35d937024938a75383fd635585474.

PR #2726, opened 2020-06-22
services/horizon/internal: Load Core Latest Ledger from core's http info end-point.

### LINKED ISSUE #2725: Get core latest ledger from http end-point instead of core db.
We are in the process of removing Horizon's dependency on Stellar Core database. This is one of the places where we still use it.

https://github.com/stellar/go/blob/28355c038a075896b3e93c166d5243875af5a40b/services/horizon/internal/app.go#L204-L236

In the code above, instead of reading from core's database, we should fetch this value from Horizon's DB - we discuss the possibility of adding this value to the `key_value_store` and have the front-end instances fetch the value from there.

We'll put the responsibility of updating this value in the ingesting instances. We can take the following approach:

1. Use core's http end-point if Horizon if Horizon is still syncing/ingesting.
2. Once Horizon has catched up with core,  update the value every time a  new ledger is processed instead of reading from core's http end-point.

Remove this field from Horizon's root resource.

abuiles:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Use core's http end-point instead of core's db to load the latest ledger in core.

### Why

Horizon should not use Core DB outside of the ingestion package. This will help us remove the coupling between non-ingesting Horizon instances and Core's DB.

Fixes #2725

### Known limitations

[TODO or N/A]

### DIFF
--- a/services/horizon/internal/app.go
+++ b/services/horizon/internal/app.go
@@ -208,11 +208,17 @@ func (a *App) UpdateLedgerState() {
 		log.WithStack(err).WithField("err", err.Error()).Error(msg)
 	}
 
-	err := a.CoreQ().LatestLedger(&next.CoreLatest)
+	coreClient := &stellarcore.Client{
+		HTTP: http.DefaultClient,
+		URL:  a.config.StellarCoreURL,
+	}
+
+	coreInfo, err := coreClient.Info(a.ctx)
 	if err != nil {
 		logErr(err, "failed to load the latest known ledger state from core DB")
 		return
 	}
+	next.CoreLatest = int32(coreInfo.Info.Ledger.Num)
 
 	err = a.HistoryQ().LatestLedger(&next.HistoryLatest)
 	if err != nil {
--- a/services/horizon/internal/app.go
+++ b/services/horizon/internal/app.go
@@ -215,7 +215,7 @@ func (a *App) UpdateLedgerState() {
 
 	coreInfo, err := coreClient.Info(a.ctx)
 	if err != nil {
-		logErr(err, "failed to load the latest known ledger state from core DB")
+		logErr(err, "failed to load the stellar-core info")
 		return
 	}
 	next.CoreLatest = int32(coreInfo.Info.Ledger.Num)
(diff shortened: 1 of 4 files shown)


abuiles:
This is a simpler approach to #2725  and since front-end instances will likely have access to core via http, this will still work fine.

services/horizon/internal/app.go
```diff
@@ -208,11 +208,17 @@ func (a *App) UpdateLedgerState() {
...
 
-	err := a.CoreQ().LatestLedger(&next.CoreLatest)
+	coreClient := &stellarcore.Client{
+		HTTP: http.DefaultClient,
+		URL:  a.config.StellarCoreURL,
+	}
+
+	coreInfo, err := coreClient.Info(a.ctx)
 	if err != nil {
 		logErr(err, "failed to load the latest known ledger state from core DB")
```
tamirms, line 218: seems like this message should be updated

services/horizon/internal/app.go
```diff
@@ -208,11 +208,17 @@ func (a *App) UpdateLedgerState() {
 		log.WithStack(err).WithField("err", err.Error()).Error(msg)
 	}
 
-	err := a.CoreQ().LatestLedger(&next.CoreLatest)
+	coreClient := &stellarcore.Client{
+		HTTP: http.DefaultClient,
+		URL:  a.config.StellarCoreURL,
+	}
+
+	coreInfo, err := coreClient.Info(a.ctx)
```
bartekn, line 216: Looks like we already call `Info` in `UpdateStellarCoreInfo`. Maybe let's just update the `CoreLatest` value there and remove this method?

abuiles: @bartekn yeah, I noticed that too but decided to keep it here for consistency (groups ledger state updates in a place) -- however a single call would be better -- I'll remove this and update in `UpdateStellarCoreInfo`

abuiles: @bartekn I had a quick look at this and unfortunately doing this will require a larger refactor because of the way ledger (and CoreLatest) is being used.  We not only use CoreLatest to feed the root, but we also use it for other checks like verifying if the system is stale or not, some of those changes unfortunately rely on the global var ledger.

PR #2728, opened 2020-06-23
exp/services/recoverysigner: run go mod tidy and go list -m all

howardtw:
### What

Run `go mod tidy` as well as `go list -m all > go.list`

### Why

This is to "clean up" the dependencies as well as help myself to understand why running such command will generate these dependencies. Not sure whether [this](https://go-review.googlesource.com/c/go/+/237017) is a fix to this issue.

### Known limitations

N/A

PR #2729, opened 2020-06-23
exp/services/recoverysigner: add "encryption-tink-keyset create" cmd

howardtw:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This PR adds the `encryption-tink-keyset create` command to output three keysets in JSON:
1. the cleartext keyset containing the private key
2. the encrypted keyset containing the private key
3. the cleartext keyset containing the public key

### Why

This command generates the necessary keyset as the value to the  `encryption-tink-keyset` env to the recovery signer as a part of the feature of having unique signer keys per registered account.

Though we only need the encrypted keyset at this moment, the cleartext keyset containing the private key is used as a backup for devops and the cleartext keyset containing the public key will be used as the value of another env introduced later.

### Known limitations

I'm wondering whether this command should be merged into the master so that devops can generate a key and configure the `encryption-tink-keyset` as well as the KMS key uri before the rest of code gets in given that `encryption-tink-keyset` is a required config.

### DIFF
--- a/exp/services/recoverysigner/cmd/keyset.go
+++ b/exp/services/recoverysigner/cmd/keyset.go
@@ -65,15 +65,15 @@ func (c *KeysetCommand) Create() {
 	keysetPublic := strings.Builder{}
 
 	if c.RemoteKEKURI != "" {
-		kmsClient, err := awskms.NewClient(c.RemoteKEKURI)
-		if err != nil {
-			c.Logger.Errorf("Error initializing AWS KMS client: %s", err.Error())
+		kmsClient, kmsErr := awskms.NewClient(c.RemoteKEKURI)
+		if kmsErr != nil {
+			c.Logger.Errorf("Error initializing AWS KMS client: %s", kmsErr.Error())
 			return
 		}
 
-		aead, err := kmsClient.GetAEAD(c.RemoteKEKURI)
-		if err != nil {
-			c.Logger.Errorf("Error getting AEAD primitive from KMS: %s", err.Error())
+		aead, kmsErr := kmsClient.GetAEAD(c.RemoteKEKURI)
+		if kmsErr != nil {
+			c.Logger.Errorf("Error getting AEAD primitive from KMS: %s", kmsErr.Error())
 			return
 		}
 
--- a/exp/services/recoverysigner/cmd/keyset.go
+++ b/exp/services/recoverysigner/cmd/keyset.go
@@ -70,7 +70,7 @@ func (c *KeysetCommand) Create() {
 }
 
 func createKeyset(kmsKeyURI string) (publicCleartext string, privateCleartext string, privateEncrypted string, err error) {
-	khPriv, err := keyset.NewHandle(hybrid.ECIESHKDFAES128CTRHMACSHA256KeyTemplate())
+	khPriv, err := keyset.NewHandle(hybrid.ECIESHKDFAES128GCMKeyTemplate())
 	if err != nil {
 		return "", "", "", errors.Wrap(err, "generating a new keyset")
 	}
--- a/exp/services/recoverysigner/cmd/keyset.go
+++ b/exp/services/recoverysigner/cmd/keyset.go
@@ -10,12 +10,13 @@ import (
 	"github.com/google/tink/go/keyset"
 	"github.com/spf13/cobra"
 	"github.com/stellar/go/support/config"
+	"github.com/stellar/go/support/errors"
 	supportlog "github.com/stellar/go/support/log"
 )
 
 type KeysetCommand struct {
-	Logger       *supportlog.Entry
-	RemoteKEKURI string
+	Logger              *supportlog.Entry
+	EncryptionKMSKeyURI string
 }
 
 func (c *KeysetCommand) Command() *cobra.Command {
@@ -24,7 +25,7 @@ func (c *KeysetCommand) Command() *cobra.Command {
 			Name:        "encryption-kms-key-uri",
 			Usage:       "URI for a remote KMS key used to encrypt Tink keyset",
 			OptType:     types.String,
-			ConfigKey:   &c.RemoteKEKURI,
+			ConfigKey:   &c.EncryptionKMSKeyURI,
 			FlagDefault: "",
 			Required:    false,
 		},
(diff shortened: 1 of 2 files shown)


howardtw:
@leighmcculloch when I built this PR, I ended up getting the same result in https://github.com/stellar/go/pull/2728
```
Howards-MacBook-Pro:recoverysigner howard$ go build ./...
go: finding module for package cloud.google.com/go/storage
go: finding module for package cloud.google.com/go/internal/trace
go: finding module for package cloud.google.com/go/internal/fields
go: finding module for package cloud.google.com/go/internal/btree
go: finding module for package google.golang.org/genproto/googleapis/rpc/status
go: finding module for package cloud.google.com/go/internal/version
go: finding module for package google.golang.org/genproto/googleapis/firestore/v1
go: found cloud.google.com/go/internal/btree in cloud.google.com/go v0.58.0
go: found google.golang.org/genproto/googleapis/firestore/v1 in google.golang.org/genproto v0.0.0-20200623002339-fbb79eadd5eb
go: found cloud.google.com/go/storage in cloud.google.com/go/storage v1.10.0
go: finding module for package cloud.google.com/go/firestore
go: found cloud.google.com/go/firestore in cloud.google.com/go/firestore v1.2.0
```

I wonder what causes the difference between yours environment and mine.

exp/services/recoverysigner/cmd/keyset.go
```diff
@@ -0,0 +1,110 @@
...
+			c.Create()
+		},
+	}
+	cmd.AddCommand(createCmd)
+
+	return cmd
+}
+
+func (c *KeysetCommand) Create() {
+	khPriv, err := keyset.NewHandle(hybrid.ECIESHKDFAES128CTRHMACSHA256KeyTemplate())
```
leighmcculloch, line 57: What's the reason for choosing CTR+HMAC over GCM, i.e.g. `ECIESHKDFAES128GCMKeyTemplate`?

exp/services/recoverysigner/cmd/keyset.go
```diff
@@ -0,0 +1,110 @@
...
+		c.Logger.Errorf("Error writing cleartext keyset containing public key: %s", err.Error())
+		return
+	}
+
+	c.Logger.Print("Cleartext keyset public:", keysetPublic.String())
+	c.Logger.Print("Cleartext keyset private:", keysetPrivateCleartext.String())
+	if c.RemoteKEKURI != "" {
+		c.Logger.Print("Encrypted keyset private:", keysetPrivateEncrypted.String())
+	}
+}
```
leighmcculloch, line 110: How could we test this? If it was just one function and output the key with little logic I think we could just test those other functions, but there's a bit more here than that so I think we should test it somehow. Maybe remove the logging parts of this to an outer function then you can test the returned results and the errors?

exp/services/recoverysigner/cmd/keyset.go
howardtw: No particular reason really. I was following their [example](https://github.com/google/tink/blob/master/docs/GOLANG-HOWTO.md#preparation). Would you prefer `ECIESHKDFAES128GCMKeyTemplate` over `ECIESHKDFAES128CTRHMACSHA256KeyTemplate`?

exp/services/recoverysigner/cmd/keyset.go
howardtw: > Maybe remove the logging parts of this to an outer function then you can test the returned results and the errors?

That works. Do you think it's necessary to mock the `RemoteKEKURI ` out? Like `if c.RemoteKEKURI == "mock"`?

leighmcculloch: I'd write tests with it being `""` as I don't think there's a good story for mocking the AWS integration right now. I've got an open issue on the Tink repo about that here: google/tink#378.

howardtw: https://github.com/stellar/go/pull/2729/commits/2c6369ecaab55565b36960a302dec6356e171ddb

exp/services/recoverysigner/cmd/keyset.go
```diff
@@ -0,0 +1,115 @@
...
+		c.Logger.Errorf("Error creating keyset: %v", err)
+		return
+	}
+
+	c.Logger.Print("Cleartext keyset public:", keysetPublic)
+	c.Logger.Print("Cleartext keyset private:", keysetPrivateCleartext)
+
+	if keysetPrivateEncrypted != "" {
+		c.Logger.Print("Encrypted keyset private:", keysetPrivateEncrypted)
+	}
```
leighmcculloch, line 69: Nice, this is so easy to read. It's so easy to see what this commands output to logs are separate from the nuts and bolts of what it does.

exp/services/recoverysigner/cmd/keyset.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+func createKeyset(kmsKeyURI string) (publicCleartext string, privateCleartext string, privateEncrypted string, err error) {
+	khPriv, err := keyset.NewHandle(hybrid.ECIESHKDFAES128CTRHMACSHA256KeyTemplate())
```
leighmcculloch, line 73: Following on from our discussion elsewhere about GCM vs CTR:

Apparently GCM is faster because it uses a single key where-as CTR+HMAC uses two keys within. Apparently GCM is more common now because of that. I don't have a good source to link for that but when I search for comparisons between GCM and CTR+HMAC the anecdotal comments I find is that it is what people reach for by default now. Another difference is CTR is apparently not an AEAD, but in tink it is implemented as AEAD so I must be missing something there. I think we could use either because Tink is designed to prevent us from making silly mistakes so if it was bad to use it probably wouldn't be in the library, but I think we should use GCM because I see it mentioned more often as a suggested one to use. I realize that's a horrible reason.

How about we use GCM for now on that basis and track a decision on which to use in a GitHub issue.

exp/services/recoverysigner/cmd/keyset.go
```diff
@@ -0,0 +1,115 @@
...
+	if err != nil {
+		return "", "", "", errors.Wrap(err, "writing cleartext keyset containing private key")
+	}
+
+	khPub, err := khPriv.Public()
+	if err != nil {
+		return "", "", "", errors.Wrap(err, "getting keyhandle for public key")
+	}
+
+	err = khPub.WriteWithNoSecrets(keyset.NewJSONWriter(&keysetPublic))
```
leighmcculloch, line 109: Why do we use WriteWithNoSecrets here and insecurecleartextkeyset above? I don't have an issue with it I just want to learn.

howardtw: The keyset handle is designed to protect its client from leaking sensitive material. Therefore, there's a no way to write a private key out with it. Both `insecurecleartextkeyset.Write` and `khPub.WriteWithNoSecrets` call `keyset.Writer.Write` underneath, but I want to limit the usage of `insecurecleartextkeyset` package to just read and write the private key when `ENCRYPTION_KMS_KEY_URI` is not configured.

exp/services/recoverysigner/cmd/keyset.go
howardtw: I'm fine with either. It wasn't easy to give you an answer when I looked it up. 😄 

We can even have `create` take one more arg so that we will be able to generate either keys, but I think it's overkill at this moment. I think one good reason to start with GCM is that there's no reason for us to start with a more complicated and slower key unless we find a security reason to do so.

howardtw: https://github.com/stellar/go/pull/2729/commits/5ba69261fe6cb54e2800fbcf684d1053f42358df

exp/services/recoverysigner/cmd/keyset.go
leighmcculloch: That makes a lot of sense, thanks. I agree, it's nice that when looking at the code all the places using the insecure package are private keys.

howardtw: You're welcome. I find myself not being clear enough. There's a way to write an encrypted private keyset out, so I should have said "there's no way to write a `cleartext` private key out with it".

PR #2730, opened 2020-06-23
exp/services/recoverysigner: add encryption-tink-keyset-public [WIP]

### LINKED ISSUE #2733: exp/services/recoverysigner: add encryption-tink-keyset-public config

howardtw:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This PR introduces a new env `encryption-tink-keyset-public` so that deployers can start the service with just an encrypter without a decrypter. The original `encryption-tink-keyset` has been renamed to `encryption-tink-keyset-private` to reflect the addition. Both `encryption-tink-keyset-public` and `encryption-tink-keyset-private` are optional.

We are doing the best effort to start the service:
1. it is possible to start the service without an encrypter and a decrypter. i.e. both account registration and signing operation won't be working.
2. when `encryption-tink-keyset-public` is configured and `encryption-tink-keyset-private` isn't, we will start the service with just the encrypter. i.e. signing operation won't be working.
3. when `encryption-tink-keyset-public` isn't configured and `encryption-tink-keyset-private` is, we will try to get an encrypter from the private key just like what we are doing before this PR.
4. when `encryption-tink-keyset-public` and `encryption-tink-keyset-private` are both configured, we will cross-check that they are compatible. i.e. the decrypter can decrypt what the encrypter encrypts. The service won't be started if they are not compatible.
5. It's also possible to start the service with just the decrypter. When the `encryption-kms-key-url` and `encryption-tink-keyset-private` are configured and `encryption-tink-keyset-public` is not, we will try to start the service even when we fail to decrypt the private keyset with AWS KMS. There's no guarantee that the decrypter will work tho.

### Why

We want to start the service even when we have trouble contacting AWS KMS. Starting the recovery signer with a few endpoints not working is a valid use case.

### Known limitations

There are risks booting up the service with partial functionality. It becomes possible that we don't have a decrypter to decrypt what an encrypter encrypts if we start the service only with the encrypter. In addition, when we start the service without making sure AWS KMS is able to decrypt the keyset, the signing endpoint won't work even we have a decrypter. The warnings have been captured in log messages. 

Close #2733


leighmcculloch:
Some thoughts about the behaviors defined above:

>1. it is possible to start the service without an encrypter and a decrypter. i.e. both account registration and signing operation won't be working.

If the encryption parameters aren't passed then we should support the `--signing-key` parameter as we do today. The app needs to be able to start using its existing capabilities of using a shared key(s). So if that's what this means that sounds good.

>2. when `encryption-tink-keyset-public` is configured and `encryption-tink-keyset-private` isn't, we will start the service with just the encrypter. i.e. signing operation won't be working.

I don't think there's a situation where we'd want to configure a service in this way. This would more likely be a result of human error if the service was configured this way and so if the public was set without the private that should probably be an error.

>3. when `encryption-tink-keyset-public` isn't configured and `encryption-tink-keyset-private` is, we will try to get an encrypter from the private key just like what we are doing before this PR.

I think we should make the service work one way with this so that we have less code to maintain.

>4. when `encryption-tink-keyset-public` and `encryption-tink-keyset-private` are both configured, we will cross-check that they are compatible. i.e. the decrypter can decrypt what the encrypter encrypts. The service won't be started if they are not compatible.

I think cross checking sounds like a good idea, but if we can't reach AWS we should just drop a warning log saying AWS couldn't be reached and it couldn't be verified and continue.

>5. It's also possible to start the service with just the decrypter. When the `encryption-kms-key-url` and `encryption-tink-keyset-private` are configured and `encryption-tink-keyset-public` is not, we will try to start the service even when we fail to decrypt the private keyset with AWS KMS. There's no guarantee that the decrypter will work tho.

Similar to (3) if we are adding public we should just require the option with the private option so we don't need to implement around these options.

leighmcculloch:
I'm going to close this because of #2811.

PR #2735, opened 2020-06-23
Update Changelog for Horizon v1.5.0.

abuiles:


services/horizon/CHANGELOG.md
```diff
@@ -5,9 +5,29 @@ file. This project adheres to [Semantic Versioning](http://semver.org/).x
...
+* Add experimental support for database reingestion using a Stellar Core subprocess instead of a persistent Stellar Core database ([#2695](https://github.com/stellar/go/pull/2695)).
+
+  [Stellar Core v12.3.0](https://github.com/stellar/stellar-core/releases/tag/v12.3.0) added an experimental feature which allows replaying ledger's metadata in-memory. This feature speeds up reingestion and starts paving the way to remove the dependency between Stellar Core's database and Horizon.
+
+  For now, this is only supported while running `horizon db reingest`. To try out this new experimental feature, you need to specify the following parameters:
+
+  - `--enable-captive-core-ingestion` or `ENABLE_CAPTIVE_CORE_INGESTION=true`.
+  - `--stellar-core-binary-path` or `STELLAR_CORE_BINARY_PATH`.
+
+### SDK Maintainers
```
ire-and-curses, line 28: I like it! Maybe it should be a clearer call to action? title could be "SDK maintainers: action needed" or something like that? Also do we have a reference implementation in one of our SDKs? If so could link here for convenience

abuiles: @ire-and-curses I just updated the title. As for the reference implementation, we don't have any yet. This particular scenario might not need one and we can update the release notes once we officially release the new version.

PR #2740, opened 2020-06-24
exp/services/recoverysigner: add encryption-tink-keyset rotate cmd

### LINKED ISSUE #2732: exp/services/recoverysigner: add keyset rotation command

howardtw:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This PR adds a new sub-command `rotate` to `encryption-tink-keyset` to rotate the current keyset. The current keyset is taken as a new config `encryption-tink-keyset`. 

The key rotation will add a new key to the current keyset while making the existing key still available. That is, we can use the new keyset to decrypt what the original keyset encrypted because the original key in the keyset is not disabled or destroyed as a result of the rotation.

### Why

The system administrator might find the need to rotate the current tink keyset periodically.

### Known limitations

This PR let system administrator rotate the tink keyset. However, tink does not has the ability to rotate the remote KMS Key.

To rotate the KMS key, we will have to add a new sub-command to `encryption-tink-keyset`. I plan on calling that `reencrypt`. `reencrypt` will need to take two KMS Key URIs, the original `kms-key-uri` and a new `new-kms-key-uri` to decrypt the keyset and re-encrypt it with the new kms key.

To perform the re-encryption more safely, we can use the aws sdk and use the built-in [Reencrypt](https://docs.aws.amazon.com/sdk-for-go/api/service/kms/#KMS.ReEncrypt) function so that the decrypted keyset is never exposed.

Close #2732

### DIFF
--- a/exp/services/recoverysigner/cmd/keyset.go
+++ b/exp/services/recoverysigner/cmd/keyset.go
@@ -8,6 +8,7 @@ import (
 	"github.com/google/tink/go/insecurecleartextkeyset"
 	"github.com/google/tink/go/integration/awskms"
 	"github.com/google/tink/go/keyset"
+	"github.com/google/tink/go/tink"
 	"github.com/spf13/cobra"
 	"github.com/stellar/go/support/config"
 	"github.com/stellar/go/support/errors"
@@ -17,6 +18,7 @@ import (
 type KeysetCommand struct {
 	Logger              *supportlog.Entry
 	EncryptionKMSKeyURI string
+	CurrentTinkKeyset   string
 }
 
 func (c *KeysetCommand) Command() *cobra.Command {
@@ -26,6 +28,14 @@ func (c *KeysetCommand) Command() *cobra.Command {
 			Usage:       "URI for a remote KMS key used to encrypt Tink keyset",
 			OptType:     types.String,
 			ConfigKey:   &c.EncryptionKMSKeyURI,
+			FlagDefault: "",
+			Required:    false,
+		},
+		{
+			Name:        "current-tink-keyset",
+			Usage:       "Current Tink keyset to rotate",
+			OptType:     types.String,
+			ConfigKey:   &c.CurrentTinkKeyset,
 			FlagDefault: "",
 			Required:    false,
 		},
@@ -49,7 +59,16 @@ func (c *KeysetCommand) Command() *cobra.Command {
 			c.Create()
 		},
 	}
+	rotateCmd := &cobra.Command{
+		Use:   "rotate",
+		Short: "Rotate an existing Tink keyset",
+		Run: func(_ *cobra.Command, _ []string) {
+			c.Rotate()
+		},
+	}
+
 	cmd.AddCommand(createCmd)
+	cmd.AddCommand(rotateCmd)
 
 	return cmd
 }
@@ -109,6 +128,89 @@ func createKeyset(kmsKeyURI string) (publicCleartext string, privateCleartext st
 	err = khPub.WriteWithNoSecrets(keyset.NewJSONWriter(&keysetPublic))
 	if err != nil {
 		return "", "", "", errors.Wrap(err, "writing cleartext keyset containing public key")
+	}
+
+	return keysetPublic.String(), keysetPrivateCleartext.String(), keysetPrivateEncrypted.String(), nil
+}
+
+func (c *KeysetCommand) Rotate() {
+	keysetPublic, keysetPrivateCleartext, keysetPrivateEncrypted, err := rotateKeyset(c.EncryptionKMSKeyURI, c.CurrentTinkKeyset)
+	if err != nil {
+		c.Logger.Errorf("Error rotating keyset: %v", err)
+		return
+	}
+
+	c.Logger.Print("Cleartext keyset public:", keysetPublic)
+	c.Logger.Print("Cleartext keyset private:", keysetPrivateCleartext)
+
+	if keysetPrivateEncrypted != "" {
+		c.Logger.Print("Encrypted keyset private:", keysetPrivateEncrypted)
+	}
+}
+
+func rotateKeyset(kmsKeyURI, currentTinkKeyset string) (publicCleartext string, privateCleartext string, privateEncrypted string, err error) {
+	var (
+		khPriv *keyset.Handle
+		aead   tink.AEAD
+	)
+
+	if kmsKeyURI != "" {
+		kmsClient, kmsErr := awskms.NewClient(kmsKeyURI)
+		if kmsErr != nil {
+			return "", "", "", errors.Wrap(kmsErr, "initializing AWS KMS client")
+		}
+
+		aead, kmsErr = kmsClient.GetAEAD(kmsKeyURI)
+		if kmsErr != nil {
+			return "", "", "", errors.Wrap(kmsErr, "getting AEAD primitive from KMS")
+		}
+
+		khPriv, err = keyset.Read(keyset.NewJSONReader(strings.NewReader(currentTinkKeyset)), aead)
+		if err != nil {
+			return "", "", "", errors.Wrap(err, "reading encrypted keyset")
+		}
+	} else {
+		khPriv, err = insecurecleartextkeyset.Read(keyset.NewJSONReader(strings.NewReader(currentTinkKeyset)))
+		if err != nil {
+			return "", "", "", errors.Wrap(err, "getting key handle for private key")
+		}
+	}
+
+	m := keyset.NewManagerFromHandle(khPriv)
+	err = m.Rotate(hybrid.ECIESHKDFAES128GCMKeyTemplate())
+	if err != nil {
+		return "", "", "", errors.Wrap(err, "rotating keyset")
+	}
+
+	khPriv, err = m.Handle()
+	if err != nil {
+		return "", "", "", errors.Wrap(err, "creating handle for the new keyset")
+	}
+
+	keysetPrivateEncrypted := strings.Builder{}
+	keysetPrivateCleartext := strings.Builder{}
+	keysetPublic := strings.Builder{}
+
+	if kmsKeyURI != "" {
+		err = khPriv.Write(keyset.NewJSONWriter(&keysetPrivateEncrypted), aead)
+		if err != nil {
+			return "", "", "", errors.Wrap(err, "writing encrypted keyset containing private keys")
+		}
+	}
+
+	err = insecurecleartextkeyset.Write(khPriv, keyset.NewJSONWriter(&keysetPrivateCleartext))
+	if err != nil {
+		return "", "", "", errors.Wrap(err, "writing cleartext keyset containing private keys")
+	}
+
+	khPub, err := khPriv.Public()
+	if err != nil {
+		return "", "", "", errors.Wrap(err, "getting keyhandle for public keys")
+	}
+
+	err = khPub.WriteWithNoSecrets(keyset.NewJSONWriter(&keysetPublic))
+	if err != nil {
+		return "", "", "", errors.Wrap(err, "writing cleartext keyset containing public keys")
 	}
 
 	return keysetPublic.String(), keysetPrivateCleartext.String(), keysetPrivateEncrypted.String(), nil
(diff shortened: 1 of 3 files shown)


exp/services/recoverysigner/cmd/keyset.go
```diff
@@ -113,3 +132,86 @@ func createKeyset(kmsKeyURI string) (publicCleartext string, privateCleartext st
...
+		c.Logger.Errorf("Error rotating keyset: %v", err)
+		return
+	}
+
+	c.Logger.Print("Cleartext keyset public:", keysetPublic)
+	c.Logger.Print("Cleartext keyset private:", keysetPrivateCleartext)
+
+	if keysetPrivateEncrypted != "" {
+		c.Logger.Print("Encrypted keyset private:", keysetPrivateEncrypted)
+	}
```
leighmcculloch, line 148: Something that I realize about this command and the create command is that it's convenient that all three components are outputted to the terminal but because we output multiple files there's no way to pipe them directly to a file or to a process that stores it somewhere else.

It may be more practical, although require more code, to output only the encrypted key, and then offer other commands to take an encrypted key to get the public and decrypted copies. This also makes the tool more composable.

We don't need to do this in this PR, but we probably need to do this before this could be used in a production environment. This would be one of the benefits of leaning on tinkey since it already does all those composable things.

exp/services/recoverysigner/cmd/keyset_test.go
```diff
@@ -56,3 +56,102 @@ func TestCreateKeyset_invalidKMSKeyURI(t *testing.T) {
...
+	_, _, _, err = rotateKeyset("invalid-uri", privateC)
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "initializing AWS KMS client")
+}
+
+func TestRotateKeyset_noCurrentKeyset(t *testing.T) {
+	_, _, _, err := rotateKeyset("", "")
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "getting key handle for private key")
+}
```
leighmcculloch, line 157: We also need to add tests to the other functions to check that a keyset provided with multiple keys works, when rotation occurs on the keys. E.g. we should be able to load a keyset, encrypt with it, load a rotated keyset, decrypt with it, etc. Do you plan to add those tests separately to this PR?

exp/services/recoverysigner/cmd/keyset.go
```diff
@@ -113,3 +132,86 @@ func createKeyset(kmsKeyURI string) (publicCleartext string, privateCleartext st
...
+		}
+	} else {
+		khPriv, err = insecurecleartextkeyset.Read(keyset.NewJSONReader(strings.NewReader(currentTinkKeyset)))
+		if err != nil {
+			return "", "", "", errors.Wrap(err, "getting key handle for private key")
+		}
+	}
+
+	m := keyset.NewManagerFromHandle(khPriv)
+	err = m.Rotate(hybrid.ECIESHKDFAES128GCMKeyTemplate())
```
leighmcculloch, line 180: Can we move `hybrid.ECIESHKDFAES128GCMKeyTemplate()` into a function on the `KeysetCommand`, something like the below, and have both the create and rotate commands ask that function for the key template to use?

```go
func (c *KeysetCommand) keyTemplate() {
	return hybrid.ECIESHKDFAES128GCMKeyTemplate()
}
```

exp/services/recoverysigner/cmd/keyset_test.go
```diff
@@ -56,3 +56,102 @@ func TestCreateKeyset_invalidKMSKeyURI(t *testing.T) {
 	require.Error(t, err)
 	assert.Contains(t, err.Error(), "initializing AWS KMS client")
 }
+
+func TestRotateKeyset(t *testing.T) {
+	public1, privateC1, _, err := createKeyset("")
```
leighmcculloch, line 61: What does `privateC` stand for?

exp/services/recoverysigner/cmd/keyset_test.go
```diff
@@ -56,3 +56,102 @@ func TestCreateKeyset_invalidKMSKeyURI(t *testing.T) {
...
+	khPriv2, err := insecurecleartextkeyset.Read(keyset.NewJSONReader(strings.NewReader(privateC2)))
+	require.NoError(t, err)
+
+	exportedPriv2 := &keyset.MemReaderWriter{}
+	err = insecurecleartextkeyset.Write(khPriv2, exportedPriv2)
+	require.NoError(t, err)
+
+	// rotation will add a new key to the keyset and make the new key the primary key
+	assert.Len(t, exportedPriv2.Keyset.Key, 2)
+	assert.NotEqual(t, exportedPriv2.Keyset.PrimaryKeyId, exportedPriv1.Keyset.PrimaryKeyId)
```
leighmcculloch, line 87: Can we test the specific key ID value rather than not equality? E.g. could we test that the primary key ID is now equal to the ID of the second key in the keyset? Combined with the comment above this will paint a clearer picture of how this works.

exp/services/recoverysigner/cmd/keyset_test.go
```diff
@@ -56,3 +56,102 @@ func TestCreateKeyset_invalidKMSKeyURI(t *testing.T) {
...
+	public1, privateC1, _, err := createKeyset("")
+	require.NoError(t, err)
+
+	khPriv1, err := insecurecleartextkeyset.Read(keyset.NewJSONReader(strings.NewReader(privateC1)))
+	require.NoError(t, err)
+
+	exportedPriv1 := &keyset.MemReaderWriter{}
+	err = insecurecleartextkeyset.Write(khPriv1, exportedPriv1)
+	require.NoError(t, err)
+	assert.Len(t, exportedPriv1.Keyset.Key, 1)
```
leighmcculloch, line 70: Can we add a test of the primary key ID here, to show that it is set to the ID of the only key in the keyset?

exp/services/recoverysigner/cmd/keyset_test.go
```diff
@@ -56,3 +56,102 @@ func TestCreateKeyset_invalidKMSKeyURI(t *testing.T) {
...
+	assert.Len(t, exportedPriv2.Keyset.Key, 2)
+	assert.NotEqual(t, exportedPriv2.Keyset.PrimaryKeyId, exportedPriv1.Keyset.PrimaryKeyId)
+
+	khPub2, err := keyset.ReadWithNoSecrets(keyset.NewJSONReader(strings.NewReader(public2)))
+	require.NoError(t, err)
+
+	// verify that one is able to get the same keyset public with keyset private
+	khPubv2, err := khPriv2.Public()
+	require.NoError(t, err)
+	assert.Equal(t, khPub2.String(), khPubv2.String())
```
leighmcculloch, line 95: I don't understand what we are testing here, could you elaborate? What is `khPubv2` vs `khPub2`? What does the `v` stand for?

exp/services/recoverysigner/cmd/keyset.go
```diff
@@ -29,6 +31,14 @@ func (c *KeysetCommand) Command() *cobra.Command {
 			FlagDefault: "",
 			Required:    false,
 		},
+		{
+			Name:        "current-tink-keyset",
```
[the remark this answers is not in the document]
howardtw: We already named the command to be the same as the parameter in the main app as `encryption-tink-keyset`. Do you mean you want the config to be the same as the name of the command? And we will have something like `recoverysigner encryption-tink-keyset rotate --encryption-tink-keyset='....'`

exp/services/recoverysigner/cmd/keyset.go
howardtw: Since this is not essential to do this right now and we'd better consult our devops how they plan on managing the encrypted keyset and backing up the cleartext keyset (we can ignore the public key for now since the system doesn't take it atm), I'm going to defer this. We might have a better idea how they want to do this by then and work on it after the tasks having a higher priority are done.

exp/services/recoverysigner/cmd/keyset_test.go
howardtw: It stands for `private cleartext`. I use the same term in `TestCreateKeyset` [here](https://github.com/stellar/go/blob/recoverysigner-unique-signing-keys/exp/services/recoverysigner/cmd/keyset_test.go#L15).

exp/services/recoverysigner/cmd/keyset_test.go
howardtw: We are testing the public keyset returned by the `rotate` command can be derived from the private keyset it returns. In other words, we are verifying that the public keyset is associated with the private keyset as opposed to a random one. We are [testing](https://github.com/stellar/go/blob/recoverysigner-unique-signing-keys/exp/services/recoverysigner/cmd/keyset_test.go#L27-L30) the same thing in `TestCreateKeyset` as well.

`khPubv2` stands for the `Keyset Handle PUBlic from priVate keyset 2`.

exp/services/recoverysigner/cmd/keyset_test.go
howardtw: Do you think the test cases you mentioned here should be included in `keyset_test.go`? It might be better to place them into `crypto_test.go` in the `crypto` package?

exp/services/recoverysigner/cmd/keyset.go
```diff
@@ -49,7 +59,16 @@ func (c *KeysetCommand) Command() *cobra.Command {
 			c.Create()
 		},
 	}
+	rotateCmd := &cobra.Command{
+		Use:   "rotate",
+		Short: "Rotate an existing Tink keyset",
```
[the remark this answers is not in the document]
howardtw: https://github.com/stellar/go/pull/2740/commits/79bdc458206c196fd898014d2222940b02549293

exp/services/recoverysigner/cmd/keyset.go
howardtw: https://github.com/stellar/go/pull/2740/commits/3eebb65e2f8be06fabda74452876a619216bcf43

exp/services/recoverysigner/cmd/keyset_test.go
howardtw: https://github.com/stellar/go/pull/2740/commits/93da41f10ed86407110a24f7dc63fef2157703e9

exp/services/recoverysigner/cmd/keyset_test.go
howardtw: https://github.com/stellar/go/pull/2740/commits/93da41f10ed86407110a24f7dc63fef2157703e9

exp/services/recoverysigner/cmd/keyset_test.go
howardtw: My thought is to let it be in `crypto_test.go`. If you agree, let's do it in a separate PR.

exp/services/recoverysigner/cmd/keyset_test.go
howardtw: This reminds me that I forgot to add the basic check for the values returned by `rotateKeyset`.

https://github.com/stellar/go/pull/2740/commits/93facd52b570f0ba3e3fc768756d58e789e34f7b
^ `privateE` stands for `private encrypted`.

exp/services/recoverysigner/cmd/keyset.go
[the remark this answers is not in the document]
howardtw: https://github.com/stellar/go/pull/2740/commits/cf9437c5efc63f74dd38589348196795d6c702a2
^ if this is not what you mean and find this confusing, we can rename it to something else. But this seems fine to me as one is a command, one is a config. 🤷‍♂️

exp/services/recoverysigner/cmd/keyset_test.go
howardtw: https://github.com/stellar/go/pull/2744
^ I'm glad I did it in a separate PR.

exp/services/recoverysigner/cmd/keyset.go
howardtw: When I tried work on the `reencrypt` [command](https://github.com/stellar/go/pull/2746) to wrap things up before I can hand it to you, I ended up splitting into two commands, one for `encrypt`, one for `decrypt`. I think it also addresses your concern here?

PR #2746, opened 2020-06-25
exp/services/recoverysigner: add encrypt and decrypt commands

howardtw:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This PR adds two sub-commands `encrypt` and `decrypt` to `encryption-tink-keyset` to encrypt a cleartext keyset private or to decrypt an encrypted keyset private.

### Why

I was planning on adding a sub-command `reencrypt`. However,  I realized there's no way for someone who started the recovery signer with a cleartext keyset private without configuring `encryption-kms-key-uri` to encrypt the keyset later on, so I decided to add two sub-commands: `encrypt` and `decrypt`. 

For someone who wants to make the system more secure by configuring the `encryption-kms-key-uri` at a later point in time, she can use the `encrypt` command to do so; for someone who wants to re-encrypt the keyset after rotating the KMS key, she can first use `decrypt` to get the cleartext keyset if no backup is available, then use `encrypt` to get a new encrypted keyset.

### Known limitations

I could theoretically break this PR into two, one for the `encrypt` cmd and one for the `decrypt` cmd. But since they are closely related and are not too big in their own, I decided to put them together and tidy up the messages throughout the file.

### DIFF
--- a/exp/services/recoverysigner/README.md
+++ b/exp/services/recoverysigner/README.md
@@ -94,7 +94,7 @@ Available Commands:
 
 Flags:
       --encryption-kms-key-uri string   URI for a remote KMS key used to encrypt Tink keyset (ENCRYPTION_KMS_KEY_URI)
-      --encryption-tink-keyset string   Existing Tink keyset to rotate/encrypt/decrypt (ENCRYPTION_TINK_KEYSET)
+      --encryption-tink-keyset string   Tink keyset to rotate/encrypt/decrypt (ENCRYPTION_TINK_KEYSET)
 
 Use "recoverysigner encryption-tink-keyset [command] --help" for more information about a command.
 ```
--- a/exp/services/recoverysigner/README.md
+++ b/exp/services/recoverysigner/README.md
@@ -75,7 +75,7 @@ Available Commands:
 
 Flags:
       --encryption-kms-key-uri string   URI for a remote KMS key used to encrypt Tink keyset (ENCRYPTION_KMS_KEY_URI)
-      --encryption-tink-keyset string   Existing Tink keyset to rotate (ENCRYPTION_TINK_KEYSET)
+      --encryption-tink-keyset string   Existing Tink keyset to rotate/encrypt/decrypt (ENCRYPTION_TINK_KEYSET)
 
 Use "recoverysigner encryption-tink-keyset [command] --help" for more information about a command.
 ```
--- a/exp/services/recoverysigner/README.md
+++ b/exp/services/recoverysigner/README.md
@@ -28,8 +28,9 @@ Usage:
   recoverysigner [command]
 
 Available Commands:
-  db          Run database operations
-  serve       Run the SEP-30 Recovery Signer server
+  db                     Run database operations
+  encryption-tink-keyset Run Tink keyset operations
+  serve                  Run the SEP-30 Recovery Signer server
 
 Use "recoverysigner [command] --help" for more information about a command.
 ```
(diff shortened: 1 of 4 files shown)


exp/services/recoverysigner/README.md
```diff
@@ -55,5 +56,29 @@ Flags:
...
+  create      Create a new Tink keyset
+  decrypt     Decrypt the Tink keyset specified in encryption-tink-keyset with the KMS key specified in encryption-kms-key-uri
+  encrypt     Encrypt the Tink keyset specified in encryption-tink-keyset with the KMS key specified in encryption-kms-key-uri
+  rotate      Rotate the Tink keyset specified in encryption-tink-keyset by generating a new key, adding it to the keyset, and making it the primary key in the keyset
+
+Flags:
+      --encryption-kms-key-uri string   URI for a remote KMS key used to encrypt Tink keyset (ENCRYPTION_KMS_KEY_URI)
+      --encryption-tink-keyset string   Existing Tink keyset to rotate/encrypt/decrypt (ENCRYPTION_TINK_KEYSET)
+
+Use "recoverysigner encryption-tink-keyset [command] --help" for more information about a command.
```
leighmcculloch, line 80: This looks real good. Looking at this I'm thinking we might want to eventually extract these commands from recoverysigner and make them a separate tool in the repo since their use isn't limited to this service. Not important right now, just noticing.

exp/services/recoverysigner/cmd/keyset.go
```diff
@@ -34,7 +34,7 @@ func (c *KeysetCommand) Command() *cobra.Command {
 		},
 		{
 			Name:        "encryption-tink-keyset",
-			Usage:       "Existing Tink keyset to rotate",
+			Usage:       "Existing Tink keyset to rotate/encrypt/decrypt",
```
leighmcculloch, line 37:
```suggestion
			Usage:       "Tink keyset to rotate/encrypt/decrypt",
```

exp/services/recoverysigner/cmd/keyset.go
```diff
@@ -67,9 +67,25 @@ func (c *KeysetCommand) Command() *cobra.Command {
...
+	decryptCmd := &cobra.Command{
+		Use:   "decrypt",
+		Short: "Decrypt the Tink keyset specified in encryption-tink-keyset with the KMS key specified in encryption-kms-key-uri",
+		Run: func(_ *cobra.Command, _ []string) {
+			c.Decrypt()
+		},
+	}
+	encryptCmd := &cobra.Command{
+		Use:   "encrypt",
+		Short: "Encrypt the Tink keyset specified in encryption-tink-keyset with the KMS key specified in encryption-kms-key-uri",
```
leighmcculloch, line 79: These usage texts are very easy to understand 💯.

exp/services/recoverysigner/cmd/keyset.go
```diff
@@ -34,7 +34,7 @@ func (c *KeysetCommand) Command() *cobra.Command {
 		},
 		{
 			Name:        "encryption-tink-keyset",
-			Usage:       "Existing Tink keyset to rotate",
+			Usage:       "Existing Tink keyset to rotate/encrypt/decrypt",
 			OptType:     types.String,
 			ConfigKey:   &c.EncryptionTinkKeyset,
```
leighmcculloch, line 39: Is this a JSON string or a filepath? We might want to add the suffix JSON like you did elsewhere.

exp/services/recoverysigner/cmd/keyset.go
```diff
@@ -199,24 +215,128 @@ func rotateKeyset(kmsKeyURI, currentTinkKeyset string, keyTemplate *tinkpb.KeyTe
...
+		return "", "", errors.Wrap(err, "getting key handle for keyset public")
+	}
+
+	keysetPublic := strings.Builder{}
+	err = khPub.WriteWithNoSecrets(keyset.NewJSONWriter(&keysetPublic))
+	if err != nil {
+		return "", "", errors.Wrap(err, "writing cleartext keyset public")
+	}
+
+	return keysetPublic.String(), keysetPrivateEncrypted.String(), nil
```
leighmcculloch, line 341: I don't see any tests for this, is that because of the use of the aws integration here? I think we need to figure out tests for these commands, this code is pretty important. Let's think about this separate to this PR though.

exp/services/recoverysigner/README.md
howardtw: I like the idea to make it available to future services that are open source and require the use of encryption.

exp/services/recoverysigner/cmd/keyset.go
howardtw: Exactly. We will have to think about the best way to mock this out. I will create an issue for this.

Speaking of this, I'm curious what you think about having a mock struct in non-test files. i.e. when it is ok to have one and when is not. I remember you mentioned you don't want to have a mock being part of the binary of production code, but we are doing [that](https://github.com/stellar/pasapeso-backend/blob/8653dff8be94a660f5dba506b1966861079f76e5/cmd/pasapesod/kms.go#L22-L23) in the vibrant backend.

exp/services/recoverysigner/cmd/keyset.go
howardtw: https://github.com/stellar/go/pull/2746/commits/12f6dd28603de66cf1134914bce05dc971d854ea

exp/services/recoverysigner/cmd/keyset.go
leighmcculloch: I only had an issue with the type being callable via the CLI, i.e. being able to use the mock in real production code, not having the type live there. If we do have types or functions intended only for testing but that need to be importable, let's keep them in their own package. If it's for the crypto package, put them in `cryptotest`. This pattern is pretty common I think.

howardtw: > I only had an issue with the type being callable via the CLI, i.e. being able to use the mock in real production code, not having the type live there.

I see. It does look like it is possible to the use the mock in real production with KMS_PROVIDER being set to `mock` in the vibrant backend. Maybe we need to address that.

> If we do have types or functions intended only for testing but that need to be importable, let's keep them in their own package. If it's for the crypto package, put them in cryptotest. This pattern is pretty common I think.

Yeah I usually realize that I have to do this if I have the import cycle issue.

howardtw: https://github.com/stellar/go/issues/2750

howardtw: > Maybe we need to address that.

Maybe we don't. I think there might be some values to start the service with a KMS mock client in vibrant backend. Looks like this is done on purpose so that people can start the service locally and test the endpoints that use it.

I think the difference lies in the fact that the recovery signer can operate without configuring a KMS client because it is only used to encrypt the tink keyset whereas the vibrant backend cannot operate without configuring a KMS client because it is used to encrypt the data directly.

What do you think? Do you agree?

exp/services/recoverysigner/cmd/keyset.go
howardtw: https://github.com/stellar/go/pull/2746/commits/8bb682c494d4909cd9bc41873e510122ab798622

PR #2749, opened 2020-06-25
services/horizon/internal/actions: Don't setup repeatable read transaction in paths request.

abuiles:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Fix error while trying to do a path finding request.

### Why

We were manually setting up the transaction inside the `/paths` handler which was causing an error since getting `historyQ` from the request already prepares the transaction with the right level of isolation and starts a transaction.

### DIFF
--- a/services/horizon/internal/actions_path.go
+++ b/services/horizon/internal/actions_path.go
@@ -2,7 +2,6 @@ package horizon
 
 import (
 	"context"
-	"database/sql"
 	"fmt"
 	"net/http"
 	"strings"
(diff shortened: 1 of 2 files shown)


services/horizon/internal/actions_path.go
```diff
@@ -360,13 +359,6 @@ func assetsForAddress(r *http.Request, addy string) ([]xdr.Asset, []xdr.Int64, e
...
 		return nil, nil, err
 	}
-	err = historyQ.BeginTx(&sql.TxOptions{
-		Isolation: sql.LevelRepeatableRead,
-		ReadOnly:  true,
-	})
-	if err != nil {
-		return nil, nil, err
-	}
 	defer historyQ.Rollback()
```
bartekn, line 362: I think we should remove this line. It's possible that the code in middlewares rely on historyQ in a transaction. 
```suggestion

```

abuiles: 👍

PR #2762, opened 2020-06-30
support/keypairgen: add types and interface for keypair generation and mocking

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Add type that wraps keypair generation, `keypairgen.Generator`, and interface, `keypairgen.Source`, that can be easily mocked.

The default source for the generator is the random source, i.e. `keypair.Random`.

A `keypairgentest.SliceSource` type is provided as well.

### Why
I'm adding functionality to recoverysigner that will have that service randomly generate a new signing key when an account registers. Writing tests for that logic is difficult because I can't predict what the signing key will be for an account that gets registered because if the application code calls keypair.Random directly there's no easy way to mock it. In the service I plan to mock the call to keypair.Random and the type and interface for doing so seems reusable.

The type and interface are heavily inspired by the `support/clock` package in this repository and follows the same patterns as that package, including a subpackage `keypairgentest` and a fixed value implementation of the source useful for testing.

### Known limitations

N/A


keypair/keypairtest/slice.go
```diff
@@ -0,0 +1,19 @@
+package keypairtest
+
+import (
+	"github.com/stellar/go/keypair"
+)
+
+// SliceSource is a keypair.Generator source that has the values returned
+// from a slice of keys that are provided at generation one at a time.
+type SliceSource []*keypair.Full
```
howardtw, line 9: Question: why do we want this type? What is the application? Can people simply loop through the slice?

keypair/keypairtest/slice.go
```diff
@@ -0,0 +1,19 @@
...
+
+// Generate returns the first key in the slice, and then shortens the slice
+// removing the returned key, so that each call returns the next key in the
+// original source. If called when no keys are available the function will
+// panic.
```
howardtw, line 14: Given that this is a library, maybe we shouldn't use panic? What do you think about returning an error if the slice is empty?

keypair/keypairtest/slice.go
leighmcculloch: To mock generation of keys using a predictable sequence of keys, so that it is possible to write tests with predictable set of keys. For example:

https://github.com/stellar/go/blob/b4372f540704e078455f19fbc34f0ca91f88a681/keypair/keypairtest/slice_test.go#L16-L26

keypair/keypairtest/slice.go
leighmcculloch: I think it is okay to panic in this function. This is in the `keypairtest` package and it's only designed for use in tests. I could raise errors here like "no keys in slice source" but any test where this is used should probably fail hard anyway if the developer hasn't configured the slice source with enough keys. It'll be indicative of test setup failure and not any application failure.

keypair/keypairtest/slice.go
leighmcculloch: This test demonstrates the use with the Generator type:

https://github.com/stellar/go/blob/b4372f540704e078455f19fbc34f0ca91f88a681/keypair/generator_test.go#L35-L49

keypair/keypairtest/slice.go
howardtw: Ah my bad. I didn't realize this is defined in `keypairtest`, not `keypair`.

keypair/generator_test.go
```diff
@@ -0,0 +1,50 @@
+package keypair_test
```
howardtw, line 1: Why do we want to use `keypair_test` as opposed to `keypair`? Can't we just do `g := Generator{}` and `g := (*Generator)(nil)`?

leighmcculloch: > Why do we want to use `keypair_test` as opposed to `keypair`? 

It's not required but in tests for package code it can help the developer to experience what the code will be like to use by an importer. We should definitely use _test in test files containing examples as it makes for more realistic examples. There are no downsides that I know of other than needing to type a few extra characters, which is essentially free. 

> Can't we just do `g := Generator{}` and `g := (*Generator)(nil)`?

Those two values are different values and types and if I used the former the test wouldn't be testing the right thing. The first is the `Generator` type with all its fields set with a zero value, so it is a Generator that has an address. The second is a `Generator` pointer with value `nil` so there's no generator.  In the `TestGenerator_Generate_sourceNotSetPtrNil` test the `Generate` function is being called on a `nil` `Generator` pointer to ensure the type is safe to use when given no value. The code being tested is this code:

https://github.com/stellar/go/blob/b4372f540704e078455f19fbc34f0ca91f88a681/keypair/generator.go#L10-L12

This isn't strictly necessary but some library types being useful in their nil state can be handy. In the case of the Generator type it's designed in the same way as the `"...support/clock".Clock` type so that in that its zero value it uses the source we most likely will want to use outside tests.

There are some good articles and talks about this pattern I linked below.

https://dave.cheney.net/2013/01/19/what-is-the-zero-value-and-why-is-it-useful
https://www.youtube.com/watch?v=ynoY2xz-F8s

howardtw: I just found it a little excessive by having all the three `keypair_test`, `keypairtest`, `keypairtest_test`. The pattern I've seem only involves `nonTestPackageName_test` if there's an import cycle issue. And the fact that there's no test sitting in the `keypair` package made me wonder the incentive of organizing tests in this way.

> It can help the developer to experience what the code will be like to use by an importer.

I guess the answer lies in this line. Is it the only reason that we are doing `keypair_test` instead of `keypair`? And `keypairtest` and `keypairtest_test` are introduced solely because of the new type `SliceSource` used only in tests? What do you think about moving the content of `keypairtest` and `keypairtest_test` into `generator_test.go`? (This is a question, not a request)

Having a different testing structure when it comes to test a library is something I've never thought of. TIL.

> Can't we just do g := Generator{} and g := (*Generator)(nil)?

I didn't mean to ask why we test the two. It was just an example of keeping the package name as `keypair` so that we don't have to prepend `keypair.` when declaring variable of these two types. But thanks for gathering the information regarding testing zero values. 😄

leighmcculloch: > Is it the only reason that we are doing `keypair_test` instead of `keypair`?

Yup. Every package has a `_test` package, it's referred to as XTests in some place. It is just a way to write tests in a package while being treated like a different package. There isn't a lot of documentation about it (https://github.com/golang/go/issues/25223) but it's somewhat common. 

> And `keypairtest` and `keypairtest_test` are introduced solely because of the new type `SliceSource` used only in tests?

The `keypairtest` package contains code that is only intended when writing tests using the `keypair` package. It could live in the `keypair` package but it would conflate the packages API.

keypair/generator.go
```diff
@@ -0,0 +1,33 @@
+package keypair
+
+// Generator generates new keys with the underlying source. The underlying
+// source defaults to the RandomSource if not specified.
+type Generator struct {
+	Source Source
```
bartekn, line 6: Do you plan to have any other `Source` than `RandomSource` (except in tests)? Because I think we could simplify it to just:
```go
type Generator struct {
}

func (g *Generator) Generate() (*Full, error) {
	return Random()
}
```
An interface can be created [when needed](https://medium.com/@cep21/preemptive-interface-anti-pattern-in-go-54c18ac0668a) outside of this package when mocking is needed (thanks @tamirms for explaining this to me in the past!).

Also, no strong opinion about it.

leighmcculloch: I agree that in many situations in Go the interface should live in the consumer, the pattern here is a little different to what that article is petitioning against though.

In this case the consumer is the `Generator` and it is defining the interface that just happens to be filled in the same package. An example of this pattern of having a concrete type wrap an interface that can be swapped is used in the stdlib too, for example the [`http.Client`](https://golang.org/pkg/net/http/#Client) type and the [`http.RoundTripper`](https://golang.org/pkg/net/http/#RoundTripper) interface.

The author of that article is also mostly concerned with interfaces with many functions, and the interface here has a single function and shouldn't explode beyond that.

leighmcculloch: The main reason to use the concrete type here and an interface to have combine the simple ability to mock, while also providing a simple default implementation, and then make that something we can use repeatedly. That's the functionality being provided by Generator.

bartekn: > An example of this pattern of having a concrete type wrap an interface that can be swapped is used in the stdlib too, for example the http.Client type and the http.RoundTripper interface.

I think this example is slightly different. In `keypair` we don't really expect other `Source`'s (except mocking) while in `net/http` you actually can replace it with another implementation. This makes me think that introducing another later (`Source`) is redundant.

However, I think `Source` would make sense if we actually plan to add another implementations, like key-derivation sources (SEP-7). Because of this I think it's worth leaving the code as is.

leighmcculloch: I think you're right adding the Source is in many ways redundant. The separation of the struct and the interface make it easier to use the generator with a default value so an application can define a variable or field of type `keypair.Generator` without providing a value and it will just work, and in tests it can be defined with a non-zero value. I found that pattern useful in `support/clock` which is similar to this.

I think SEP-5 is a valid use of Source, but I'm not sure how well that would work in practice and I don't think we're planning to do that in the near future so I don't think it should justify the pattern I'm using.

I think I'm adding this in the wrong place. I only plan to use this in `exp/services/recoverysigner` right now, and it is very application centric and doesn't need to be in the Go SDK's API. This either belongs in `support/` as something just for applications here to use, or in `exp/services/recoverysigner`.

bartekn: > I think SEP-5 is a valid use of Source, but I'm not sure how well that would work in practice and I don't think we're planning to do that in the near future so I don't think it should justify the pattern I'm using.

What I had in mind is [`exp/crypto/derivation`](https://godoc.org/github.com/stellar/go/exp/crypto/derivation). I imagine that generator pattern could be useful if you want to create new accounts for a given derivation path. `Source` would keep the last used id (that can be set when creating it, in case persistence is needed) and just increment it with each `Generate()` call.

leighmcculloch: Got it. That would work I think.

I've moved this into `support/keypairgen`. We could think about moving it back out into `keypair` if this pattern proves to be good in our own applications and we feel like it should be upgraded to being part of the Go SDK's API.

I've kept the interface inside the package. While Go supports having the interface at the consumer, which is awesome and I love it, sometimes its useful to share interfaces in code and I think this is one of those cases. Even if I got rid of the Source there would still be an interface for the Generator that I'd need to share in multiple places in recoverysigner service for example there will be at least two different packages that need to use this and simply sharing that interface feels right and this interface will be very low risk to interface bloat.

What do you think?

PR #2766, opened 2020-07-01
exp/services/recoverysigner: add support for writing and reading signers to db

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add logic to account page for writing Signers to the signers database table. Included is writing signers when adding a new account, deleting signers, and reading out signers from all functions that read from the db.

### Why

We've added the signers db table but the code doesn't yet write to it or read from it.

Updating has been left out because we currently have no code that would update the signers. We will start out generating a new signer when the account is registered. Updating signers needs to work slightly differently than updating the other account data because we'll never delete signers and so I've chosen to defer that until we write key rotation code that will add new signers.

### Known limitations

N/A


exp/services/recoverysigner/internal/account/db_store_add.go
```diff
@@ -45,6 +45,17 @@ func (s *DBStore) Add(a Account) error {
 		}
 	}
 
+	for i := len(a.Signers) - 1; i >= 0; i-- {
```
howardtw, line 48: Should we just add signer from the front with `range` and add a ORDER BY ID DESC in the test to make sure the new signer is placed in the front of the list? I find it difficult to read with no comment saying whether the new signer is in the front or the back of the slice. And there's no `ORDER BY` clause in the test either, which could cause the test to fail intermittently.

exp/services/recoverysigner/internal/account/db_store_add_test.go
```diff
@@ -141,6 +151,39 @@ func TestAdd(t *testing.T) {
...
+	// Check the signer rows have been added.
+	{
+		type row struct {
+			AccountID          int64  `db:"account_id"`
+			ID                 int64  `db:"id"`
+			PublicKey          string `db:"public_key"`
+			EncryptedSecretKey []byte `db:"encrypted_secret_key"`
+		}
+		rows := []row{}
+		err = session.Select(&rows, `SELECT account_id, id, public_key, encrypted_secret_key FROM signers`)
```
howardtw, line 164: Should we add a ORDER BY clause in the query? The order is not guaranteed when there's no ORDER BY. Alternative to sort by ID, it might be easier to understand if we sort by `created_at`.

exp/services/recoverysigner/internal/account/db_store_delete_test.go
```diff
@@ -191,6 +211,44 @@ func TestDelete(t *testing.T) {
...
+		assert.ElementsMatch(t, wantRows, rows)
+	}
+	// Check that account 1's signers's deletes have been audited
+	{
+		type row struct {
+			AuditOp   string `db:"audit_op"`
+			PublicKey string `db:"public_key"`
+		}
+		rows := []row{}
+		err = session.Select(&rows, `SELECT audit_op, public_key FROM signers_audit`)
```
howardtw, line 236: Similarly, we might want to add a `ORDER BY` clause here or use `ElementsMatch` as opposed to `Equal` to prevent the test from failing intermittently.

exp/services/recoverysigner/internal/account/db_store_add.go
leighmcculloch: I already use an ORDER BY DESC in the gets. This code ensures that if there are multiple signers in the add that they get persisted in reverse order. There's no way to otherwise in an INSERT as far as I know.

exp/services/recoverysigner/internal/account/db_store_add_test.go
leighmcculloch: Ordering by created_at won't help here because the rows are added inside a database transaction and will have same value, because in a transaction the time is fixed. But that's a good point that without an ORDER BY the order here is not guaranteed and just happens to be returned by how they are clustered. This is an issue in all of the tests, I'll follow up and address that in a separate PR and do it across all the tests in one pass. Thanks.

exp/services/recoverysigner/internal/account/db_store_delete_test.go
leighmcculloch: Thanks good point. I'll follow up separately and fix this in all the existing and new tests in one pass.

exp/services/recoverysigner/internal/account/db_store_add.go
leighmcculloch: A comment would go a long way here though: 7885337a

howardtw: Looking at the code of `getSigners` helped me understand the code. I think it's worth adding a comment above the loop. My point is that we are not calling `getSigners` in `TestAdd`, so it took me a while to understand what we are trying to do in the test.

If we are adding `ORDER BY ID DESC` in `TestAdd`, we would have to add from the beginning of the slice, right? Do I understand the intention correctly?

leighmcculloch: I've removed the comment and instead removed this concern from this PR in 0a1d6d3d. I was dealing with the concern of how they get ordered in the wrong place. Instead I should deal with it when producing the API response.

leighmcculloch:
@howardtw Follow ups to the `ORDER BY` issue you identified are in stellar/go#2773 and stellar/go#2774.

PR #2767, opened 2020-07-01
exp/services/recoverysigner: refactor how signers get built in responses

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Rearrange the code that builds signers for account responses so that
signers are built  into the account response after the account response
value is created.

### Why
This is a small thing but it causes the building of the signers to occur
in the same way that building identities does for the response, which
makes the code a bit clearer and will make the code read clearer
especially when we are building the signing addresses into the response
from the database.

This change should have no functional impact.

### Known limitations

N/A

### DIFF
--- a/exp/services/recoverysigner/internal/serve/account_post.go
+++ b/exp/services/recoverysigner/internal/serve/account_post.go
@@ -106,8 +106,7 @@ func (h accountPostHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	authMethodCount := 0
 	acc := account.Account{
-		Address:    req.Address.Address(),
-		Identities: []account.Identity{},
+		Address: req.Address.Address(),
 	}
 	for _, i := range req.Identities {
 		accIdentity := account.Identity{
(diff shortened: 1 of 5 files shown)


exp/services/recoverysigner/internal/serve/account_post.go
```diff
@@ -106,8 +106,7 @@ func (h accountPostHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	authMethodCount := 0
 	acc := account.Account{
-		Address:    req.Address.Address(),
-		Identities: []account.Identity{},
```
howardtw, line 110: This has nothing to do with this PR, right? Looks like we enforce that all requests have to have at least one identity, so this should not effect whether we would have `{}` or `null` in the response.

leighmcculloch: Good catch, this has nothing to do with this PR. Removed in 3d4720a0. I removed it for the reason you note, but it should be in its own PR.

leighmcculloch: Removing it separately in stellar/go#2768.

PR #2769, opened 2020-07-01
exp/services/recoverysigner: generate signer and encrypt in POST /accounts/{address}

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Generate a new signer and encrypt it and store it when an account is registered in the POST /accounts/{address} endpoint.

### Why

When an account registers we'll generate a new signing key for that account, encrypt it, and store it, so that the account has its own unique signing key.

### Known limitations

There is no logic in this PR to use that signing key for signing. Any request to the service to attempt to use that signing key would result in a 404 because the signing endpoint doesn't yet know the key exists. This is fine, this is work on a feature branch and won't affect anyone using the master branch.

~There are other changes in this branch from PR #2766. The diff will get smaller when that PRs is merged. This PR will remain a draft until it has been merged.~


exp/services/recoverysigner/internal/serve/account_post.go
```diff
@@ -125,6 +129,29 @@ func (h accountPostHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		l.Error(err)
+		serverError.Render(w)
+		return
+	}
+	signingPublicKey := signingKey.Address()
+	l.
+		WithField("signer", signingPublicKey).
+		Info("Generated signer.")
+	signingSecretKeyEncrypted, err := h.Encrypter.Encrypt([]byte(signingKey.Seed()), nil)
```
leighmcculloch, line 142: @howardtw I took the simplest route here and am encrypting the strkey version of the seed, i.e. `S...`. But it would be more space efficient if I store the raw seed bytes rather than the string. What do you think?

leighmcculloch, line 142: TODO: 
- [ ] Add context info into the `Encrypt` call.

howardtw: Hmm would storing the raw seed bytes introduce many lines of code or it would be just one more line during encryption and decryption? I think readability is probably more important here. In addition, we are already storing the `signingSecretKeyEncrypted` as `bytea`, which is already making it more space efficient in the db than storing it as `text`.

leighmcculloch: Text and bytea take up the same space in Postgres and actually text has better performance, sometimes even base64 encoded it is still better performance so I don't think we are gaining anything by using bytea other than we should use it for storing bytes and use text for storing text. Either way the encrypted data will be not text so I think the bytea type is fine.

What I meant by my question was if we should encrypt the S string or encrypt the raw bytes that make up the key stored in the S string. But I think you're point is right, keep this code simple, and encrypting the S string is the simplest approach.

Thanks!

howardtw: > Text and bytea take up the same space in Postgres and actually text has better performance, sometimes even base64 encoded it is still better performance so I don't think we are gaining anything by using bytea other than we should use it for storing bytes and use text for storing text.

https://stackoverflow.com/questions/48208050/encoding-byte-data-and-storing-as-text-vs-storing-in-bytea-in-postgresql
^ I've thought that the bytea type is more space efficient than the text type.

The author of answer to the stackoverflow question also pointed that
```
Unfortunately, the test is not very meaningful because it does not select according to the column. The test also includes implicit encoding and decoding operations that can produce completely different timings with other programming languages or adapters.
```
when it comes to [this](http://engineering.pivotal.io/post/bytea_versus_text_in_postgresql/) article. Now I'm not sure if I understand how things are handled in Postgres. 

> What I meant by my question was if we should encrypt the S string or encrypt the raw bytes that make up the key stored in the S string.

I understand your intention. I was just curious what the code would be like if we encrypt the raw bytes.

howardtw: > Text and bytea take up the same space in Postgres and actually text has better performance, sometimes even base64 encoded it is still better performance so I don't think we are gaining anything by using bytea other than we should use it for storing bytes and use text for storing text.

https://stackoverflow.com/questions/48208050/encoding-byte-data-and-storing-as-text-vs-storing-in-bytea-in-postgresql
^ I've thought that the bytea type is more space efficient than the text type.

The author of answer to the stackoverflow question also pointed that
```
Unfortunately, the test is not very meaningful because it does not select according to the column. The test also includes implicit encoding and decoding operations that can produce completely different timings with other programming languages or adapters.
```
when it comes to [this](http://engineering.pivotal.io/post/bytea_versus_text_in_postgresql/) article. Now I'm not sure if I understand how things are handled in Postgres. 

> What I meant by my question was if we should encrypt the S string or encrypt the raw bytes that make up the key stored in the S string.

I understand your intention. I was just curious what the code would be like if we encrypt the raw bytes.

howardtw:
👀

exp/services/recoverysigner/internal/serve/serve.go
```diff
@@ -185,6 +185,7 @@ func handler(deps handlerDeps) http.Handler {
 			mux.Post("/", accountPostHandler{
 				Logger:           deps.Logger,
 				SigningAddresses: deps.SigningAddresses,
```
howardtw, line 187: Did we forget to add the signing key generator here?

leighmcculloch: It's not needed, the default zero value generate will generate keys. But it would be clearer if it was specified here. Added here in 674a5798.

PR #2772, opened 2020-07-01
exp/services/recoverysigner: decrypt signer and sign transaction

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Decrypt the account's signing key when it is selected in the URL path of the POST /accounts/{address}/sign/{signing-address} endpoint.

### Why

When a client select a signing key that is not a shared signing key we need to get the key out of the database, decrypt it, and then sign the transaction with it.

### Known limitations

~There are other changes in this branch from PR #2766. The diff will get smaller when that PRs is merged. This PR will remain a draft until it has been merged.~


exp/services/recoverysigner/internal/serve/account_sign.go
```diff
@@ -104,6 +93,39 @@ func (h accountSignHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// Validating the signing key requested.
+	var signingKey *keypair.Full
+	for _, signer := range acc.Signers {
+		if req.SigningAddress.Address() == signer.PublicKey {
+			signingKeySeed, err := h.Decrypter.Decrypt(signer.EncryptedSecretKey, nil)
```
leighmcculloch, line 100: TODO: 
- [ ] Add context info into the `Decrypt` call.

howardtw:
👀

exp/services/recoverysigner/internal/serve/account_sign_signing_address_test.go
```diff
@@ -448,10 +549,90 @@ func TestAccountSign_signingAddressAccountAuthenticatedOtherSignerSelected(t *te
...
+	resp := w.Result()
+
+	require.Equal(t, http.StatusOK, resp.StatusCode)
+	assert.Equal(t, "application/json; charset=utf-8", resp.Header.Get("Content-Type"))
+
+	body, err := ioutil.ReadAll(resp.Body)
+	require.NoError(t, err)
+
+	wantBody := `{
+	"signature": "IWTH/RwlgbOY6PlZZt3GTfx3iBDowviHL+RzyNZkoqcTlpVjtykyAw1aUZywQgyimn5PL4AiR2wFUuzbGK1oBA==",
```
howardtw, line 625: Do you think we can verify that the transaction is signed by [this](https://github.com/stellar/go/pull/2772/files#diff-db9c51604aae35cf883bee5c494342e3R564) key by comparing a value produced locally??

leighmcculloch: Yup that's what I did. I used the Laboratory to sign the transaction with the key and used that signature here. This is easy to do because [on this line](https://github.com/stellar/go/pull/2772/files/06b3581f422f47776b17241a3d9075310b346713#diff-db9c51604aae35cf883bee5c494342e3R602) the test outputs the transaction. So you just copy the transaction into the lab and sign it with the key, then use the view page to see the signature.

exp/services/recoverysigner/internal/crypto/cryptotest/decrypter_panic.go
```diff
@@ -0,0 +1,12 @@
+package cryptotest
+
+// DecrypterPanic is a decrypter that panics whenever it is called. It is
+// intended for use in situations where the decrypter isn't expected to be
+// called so as to fail if the decrypter was called to decrypt data. If that
+// happens it is a good signal that data that shouldn't be decrypted is being
+// decrypted.
+type DecrypterPanic struct{}
```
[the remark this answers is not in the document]
leighmcculloch: This type is only used in tests and the use of this type is designed to break the test if the decrypter gets called because in tests that use it the decrypter isn't expected to be called. It could return an error but then that's assuming that the error is handled in a way the test fails, which it does, but a panic is a higher guarantee that if it gets called the test fails. This isn't intended for testing the chi recoverer.

Issue #2779, opened 2020-07-01
Remove `--stellar-core-db-url ` requirement when webserver is not running


bartekn:
`--stellar-core-db-url ` is still required in `txsub` to submit transactions but we don't need it when not running a webserver (ex. in reingestion context). We should allow empty values when webserver is not running.

_Originally posted by @bartekn in https://github.com/stellar/go/pull/2759#issuecomment-652684961_

bartekn:
Fixed in #2780.

PR #2780, opened 2020-07-02
services/horizon: Relax requirement on Core URL flags

2opremio:
They shouldn't be required when captive core is used

<details>
  
### PR Structure

  otherwise).
* [ x This PR's title starts with name of package that is most changed in the PR, ex.
  packages.

### Thoroughness

### Release planning

</details>

### What

Only require `--stellar-core-db-url` and `--stellar-core-url` explicitly when they are really needed.

Fixes https://github.com/stellar/go/issues/2779

### Why

An external Core instance is not required when using Captive Core, imposing artificial restrictions

### Known limitations

The message printed out won't be identical to the one printed by Cobra.

### DIFF
--- a/services/horizon/cmd/db.go
+++ b/services/horizon/cmd/db.go
@@ -204,26 +204,30 @@ var dbReingestRangeCmd = &cobra.Command{
 
 		initRootConfig()
 
-		coreSession, err := db.Open("postgres", config.StellarCoreDatabaseURL)
-		if err != nil {
-			log.Fatalf("cannot open Core DB: %v", err)
-		}
-
 		horizonSession, err := db.Open("postgres", config.DatabaseURL)
 		if err != nil {
 			log.Fatalf("cannot open Horizon DB: %v", err)
 		}
 
 		ingestConfig := expingest.Config{
-			CoreSession:                coreSession,
 			NetworkPassphrase:          config.NetworkPassphrase,
 			HistorySession:             horizonSession,
 			HistoryArchiveURL:          config.HistoryArchiveURLs[0],
 			MaxReingestRetries:         int(retries),
 			ReingesRetryBackoffSeconds: int(retryBackoffSeconds),
 		}
+
 		if config.EnableCaptiveCoreIngestion {
 			ingestConfig.StellarCorePath = config.StellarCoreBinaryPath
+		} else {
+			if config.StellarCoreDatabaseURL == "" {
+				log.Fatalf("flag --%s cannot be empty", stellarCoreDBURLFlagName)
+			}
+			coreSession, err := db.Open("postgres", config.StellarCoreDatabaseURL)
+			if err != nil {
+				log.Fatalf("cannot open Core DB: %v", err)
+			}
+			ingestConfig.CoreSession = coreSession
 		}
 
 		if parallelWorkers < 2 {
--- a/services/horizon/cmd/db.go
+++ b/services/horizon/cmd/db.go
@@ -223,9 +223,9 @@ var dbReingestRangeCmd = &cobra.Command{
 			if config.StellarCoreDatabaseURL == "" {
 				log.Fatalf("flag --%s cannot be empty", stellarCoreDBURLFlagName)
 			}
-			coreSession, err := db.Open("postgres", config.StellarCoreDatabaseURL)
-			if err != nil {
-				log.Fatalf("cannot open Core DB: %v", err)
+			coreSession, dbErr := db.Open("postgres", config.StellarCoreDatabaseURL)
+			if dbErr != nil {
+				log.Fatalf("cannot open Core DB: %v", dbErr)
 			}
 			ingestConfig.CoreSession = coreSession
 		}
--- a/services/horizon/cmd/ingest.go
+++ b/services/horizon/cmd/ingest.go
@@ -24,7 +24,7 @@ var ingestVerifyFrom, ingestVerifyTo, ingestVerifyDebugServerPort uint32
 var ingestVerifyState bool
 
 var ingestVerifyRangeCmdOpts = []*support.ConfigOption{
-	&support.ConfigOption{
+	{
 		Name:        "from",
 		ConfigKey:   &ingestVerifyFrom,
 		OptType:     types.Uint32,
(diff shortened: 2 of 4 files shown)


services/horizon/cmd/db.go
```diff
@@ -204,26 +204,30 @@ var dbReingestRangeCmd = &cobra.Command{
...
 			MaxReingestRetries:         int(retries),
 			ReingesRetryBackoffSeconds: int(retryBackoffSeconds),
 		}
+
 		if config.EnableCaptiveCoreIngestion {
 			ingestConfig.StellarCorePath = config.StellarCoreBinaryPath
+		} else {
+			if config.StellarCoreDatabaseURL == "" {
+				log.Fatalf("flag --%s cannot be empty", stellarCoreDBURLFlagName)
+			}
```
bartekn, line 225: The problem with the current solution is that duplicates code. Another possibility would be to move this check inside `initRootConfig` and add a param that explains the config context (normal run, reingestion, verify-range, ...).

services/horizon/cmd/ingest.go
```diff
@@ -102,13 +97,22 @@ var ingestVerifyRangeCmd = &cobra.Command{
...
 		ingestConfig := expingest.Config{
-			CoreSession:       coreSession,
 			NetworkPassphrase: config.NetworkPassphrase,
 			HistorySession:    horizonSession,
 			HistoryArchiveURL: config.HistoryArchiveURLs[0],
 		}
 		if config.EnableCaptiveCoreIngestion {
 			ingestConfig.StellarCorePath = config.StellarCoreBinaryPath
+		} else {
+			if config.StellarCoreDatabaseURL == "" {
```
abuiles, line 107: @2opremio is there maybe a way to specify conditional requirements in cobra? Like, this is only required if EnableCaptiveCoreIngestion is empty.

bartekn: @abuiles something like this is done in `initRootConfig`, we already have a similar code there:
```go
	if config.EnableCaptiveCoreIngestion && config.StellarCoreBinaryPath == "" {
		stdLog.Fatalf("--stellar-core-binary-path must be set when --enable-captive-core-ingestion is set")
	}
```

services/horizon/cmd/db.go
2opremio: I don't think we should worry too much about duplicated code since the Core dependency is going away medium-term.

Also, as I understand, the solution you propose needs to consider what command is being run, which is prone to error and handles command options far away from the command itself (I would rather do the check where it needs).

bartekn: I don't think the old database backend will be removed right after we move captive out of experimental mode. There are many Horizon deployments out there and they will definitely need time to migrate.

As for duplication, actually we don't need to add params. Checking `StellarCoreDatabaseURL` depends on `EnableCaptiveCoreIngestion` which is already present in `initRootConfig`. The only thing we need to check is `StellarCoreDatabaseURL` and we can actually do it in root command.

services/horizon/cmd/root.go
```diff
@@ -28,13 +34,19 @@ var (
 		Short: "client-facing api server for the stellar network",
 		Long:  "client-facing api server for the stellar network. It acts as the interface between Stellar Core and applications that want to access the Stellar network. It allows you to submit transactions to the network, check the status of accounts, subscribe to event streams and more.",
 		Run: func(cmd *cobra.Command, args []string) {
+			if config.Ingest {
+				if config.StellarCoreURL == "" {
```
bartekn, line 38: This is also required for `txsub` so shouldn't be inside `if config.Ingest` block.

services/horizon/cmd/db.go
2opremio: > As for duplication, actually we don't need to add params. Checking `StellarCoreDatabaseURL` depends on `EnableCaptiveCoreIngestion` which is already present in `initRootConfig`. The only thing we need to check is `StellarCoreDatabaseURL` and we can actually do it in root command.

That's not really the case, because right now the root command (and serve) need to use the Core URLs when ingestion is enabled regardless of the value of `EnableCaptiveCoreIngestion`

2opremio: `I don't think the old database backend will be removed right after we move captive out of experimental mode.

I didn't assume that, I said it will be removed medium-term.

bartekn: > I didn't assume that, I said it will be removed medium-term.

I mean, the duplication wouldn't be a problem if we were to remove it in the next of following release but it's likely it will stay in the codebase for at least a couple months.

> That's not really the case, because right now the root command (and serve) need to use the Core URLs when ingestion is enabled regardless of the value of EnableCaptiveCoreIngestion

There's a mistake in my comment I meant:
> The only thing we need to check is **StellarCoreURL** and we can actually do it in root command.

Sorry! So to be more clear:
1. We move all the duplicated `StellarCoreDatabaseURL` checks to `initRootCommand` because we need `EnableCaptiveCoreIngestion` and `Ingest` values for proper validation and they are already there. `StellarCoreDatabaseURL` right now is only used in ingestion (ledger backend).
2. We only check for `StellarCoreURL` in webserver and ingestion (but not when reingesting because we don't update the cursor). So we just validate it in serve because this is when it's really needed.

Sorry if I'm missing something but doing this will remove 4 x `if config.StellarCoreDatabaseURL == "" { ... }` so I guess it's worth a try.

Giving 👍 because I'm leaving soon and don't want to block this if there's really a mistake in my thinking.

2opremio: Ah, ok! I get it now.

2opremio: Actually ...

>     1. We move all the duplicated `StellarCoreDatabaseURL` checks to `initRootCommand` because we need `EnableCaptiveCoreIngestion` and `Ingest` values for proper validation and they are already there. `StellarCoreDatabaseURL` right now is only used in ingestion (ledger backend).

But ... isn't `StellarCoreDatabaseURL` required for online ingestion even if `EnableCaptiveCoreIngestion` isn't defined?

2opremio: OK. That is solvable by enforcing `StellarCoreDatabaseURL` in the `serve` command.

But then ... `db reingest range` needs to check that `StellarCoreDatabaseURL` or `CaptiveCore` are enabled regardless of the value of `Ingest` (which isn't checked), which  isn't the case for the rest of the commands. This makes is impossible to generalize in the root without checking what command is invoked.

2opremio: So, I think I will merge as is. If you think you can generalize it in an elegant way without breaking existing behavior, I will be happy to

2opremio:
Only require --stellar-core-db-url and --stellar-core-url explicitly when they are really needed.

An external Core instance is not required when using Captive Core, imposing artificial restrictions

PR #2781, opened 2020-07-03
tools/stellar-key-gen: add tool to generate keys at command line

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add tool to generate keys at command line.

### Why

When I'm running and testing services like webauth and recoverysigner either manually or when I'm writing Go tests I find myself needing to go to the lab to generate keys frequently. It would be helpful if there was a quick command to run to grab a new key. We might have something that does this already but I couldn't find anything. If there is I'm happy to discard this, otherwise this would be helpful.

### Known limitations

This tool is pretty limited scope, it doesn't do anything else other than output a random key. This is intentional, I'm not trying to start some new massive tool suite with it.


leighmcculloch:
To test the cobra.Command properly I needed to update the cobra version. I haven't reviewed the cobra changes yet to understand if that and the other dependencies are safe. I'll circle back on that before merging.

leighmcculloch:
Do you mean change the way it is tested instead of updating cobra? I can do that. Although we are four years behind on the latest version of cobra. I think it could be worth considering upgrading independent of this.

bartekn:
Makes sense. AFAIR, you had script testing dependency updates, right?

leighmcculloch:
>I think it will be much simpler to make `gen` return a string which is later printed

@bartekn I agree. It's a little more code but simpler and back to first basics without a reader needing to know so much about Cobra mechanics. I also think upgrading Cobra to the latest version would force an upgrade of Cobra in any application that is importing code in this repo which could be painful. I've updated the PR to not require a dependency change. Thanks!

PR #2784, opened 2020-07-03
exp/ingest/adapters: Improve passing options to SingleLedgerStateReader

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This commit changes a way `maxStreamRetries` (and other params in the future) are passed to `SingleLedgerStateReader`.

### Why

I've recently shared a code snipped using `SingleLedgerStateReader` and realized that the last param of `MakeSingleLedgerStateReader` is confusing. You need to check the function godoc to understand what it really does. Also, there should be a default value (currently 3) that is suitable for most of the use cases. The new method is using variadic params to set options.

Before:
```go
changeReader, err := historyAdapter.GetState(ctx, ledgerSequence, 3)
```
After:
```go
changeReader, err := historyAdapter.GetState(ctx, ledgerSequence, io.MaxStreamRetries(10))
// or:
changeReader, err := historyAdapter.GetState(ctx, ledgerSequence) // defaults to 3
```


tamirms:
personally I don't think this is an improvement. I don't like how we are using `interface{}` types to represent the remaining parameters

bartekn:
I agree about `interface{}`. An alternative could be to use a struct with config values but devs would always need to pass it. Not bad, but a pain because most users are OK with default of 3 retries. Maybe we just set it to 3 and don't allow any configuration?

tamirms:
> Maybe we just set it to 3 and don't allow any configuration?

I would be ok with that. Another solution would be to use [functional options](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis) . I would be ok with any of the following solutions:

* using a struct config
* omitting the parameter and just setting the value to 3
* keeping the int parameter as is
* using functional options

I think all those solutions are preferable to using an interface type

bartekn:
@tamirms can you take a look again?

exp/ingest/io/single_ledger_state_reader.go
```diff
@@ -73,14 +73,11 @@ const (
...
 // MakeSingleLedgerStateReader is a factory method for SingleLedgerStateReader.
-// `maxStreamRetries` determines how many times the reader will retry when encountering
-// errors while streaming xdr bucket entries from the history archive.
-// Set `maxStreamRetries` to 0 if there should be no retry attempts
 func MakeSingleLedgerStateReader(
 	ctx context.Context,
 	archive historyarchive.ArchiveInterface,
 	sequence uint32,
-	maxStreamRetries int,
+	opts ...interface{},
```
tamirms, line 80: you can get rid of `opts` since it's not used anymore

services/horizon/internal/expingest/sample_changes_test.go
```diff
@@ -55,7 +55,7 @@ func newSampleChangeReader(output string, size int) (*sampleChangeReader, error)
 	inner, err := historyAdapter.GetState(
 		context.Background(),
 		checkpointLedger,
-		3,
+		io.MaxStreamRetries(3),
```
tamirms, line 58: you can remove this param

bartekn: @tamirms this test is not run locally and in Circle because of a build tag: `// +build update`. Do you remember why it's there?

PR #2790, opened 2020-07-07
exp/services/recoverysigner: add unique constraint to signers table public_key field

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Add a unique constraint to the `signers` table `public_key` field.

### Why
The service randomly generates a signer when each account is registered. There should never have two signers that have the same value. It's very very unlikely the service would randomly generate two signers that are the same but if if it did it would be a signal of something majorly wrong, such as something wrong with random number generation on the machine or possibly a sign of some attempt to insert a signer manually incorrectly. This constraint won't solve these problems but it might help to signal to a host of recoverysigner that something is wrong, and might help prevent unpredictable issues.

### Known limitations
N/A


exp/services/recoverysigner/internal/db/dbmigrate/migrations/20200603000000-create-signers.sql
```diff
@@ -13,7 +13,7 @@ CREATE TABLE signers (
 
 
 CREATE INDEX ON signers (account_id);
-CREATE INDEX ON signers (public_key);
+CREATE UNIQUE INDEX ON signers (UPPER(public_key));
```
marcelosalloum, line 16: Will this change be applied in existing databases?

leighmcculloch: That would be a concern if this migration file was already in master and released, but the migration only exists in the feature branch that this PR is merging to. This migration hasn't been released anywhere yet so it is safe to modify.

marcelosalloum: That's great then!

PR #2802, opened 2020-07-09
services/horizon/expingest: Update ingestion state machine to not block when captive backend is used

### LINKED ISSUE #2703: Ensure distributed ingestion works and is not delayed when using captive backend
Check the code in the `captive-core-online` branch to find all possible issues captive core causes in distributed ingestion. For example: one known issues is that if one Stellar-Core is out of sync `resume` state will block entire cluster.

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This commit updates ingestion state machine to not block entire cluster ingestion when captive core backend is not prepared. Specifically:
* In `buildState`, `resumeState` and `historyRangeState` added a check that ensures a range is prepared. If not, it releases the distributed ingestion lock and prepares the selected range.
* In `resumeState` it adds the code that fast-forwards the captive core backend to the last ledger available in a backend buffer.
* Added `suggestedCheckpoint` to `startState` params that's set when transitioning from `buildState`.

Close #2703.

### Why

`CaptiveStellarCore` backend unlike `DatabaseBackend` does not allow _random-access_ to ledger data. To work it first need to prepare a range (using `PrepareRange` method) that can take a few minutes. Without changes in this PR, the node that is actively ingesting and preparing a range would block other nodes (potentially synced) from ingesting.

Another change compared to `DatabaseBackend` is that `GetLedger` method must be called in a non-decreasing order of ledger sequences. It also contains a small buffer of ledgers, `GetLatestLedgerSequence()` doesn't return the latest ledger sequence closed by the network (in case of a synced node) but the latest ledger available in a buffer (that can be also the latest ledger sequence closed by the network but that's not always true: ex. when catching up).

Because of the reasons above small changes to `resumeState` were required. With the changes in this commit, it fast-forwards to the latest ledger available in a backend to advance the buffer. Without it, the node would stay in an infinite loop asking for a ledger that is in front of the few ledgers in a buffer.

Finally, `suggestedCheckpoint` is required because a new checkpoint can be created when `buildState` is still preparing a range. When this happens, the code checking a bucket hash, would fail because `GetLatestLedgerSequence()` would return the previous checkpoint (plus a few ledgers in buffer) because it was used as a starting sequence in `PrepareRange`.

### DIFF
--- a/exp/ingest/ledgerbackend/captive_core_backend.go
+++ b/exp/ingest/ledgerbackend/captive_core_backend.go
@@ -69,6 +69,10 @@ type CaptiveStellarCore struct {
 
 	nextLedger uint32  // next ledger expected, error w/ restart if not seen
 	lastLedger *uint32 // end of current segment if offline, nil if online
+
+	// waitIntervalPrepareRange defines a time to wait between checking if the buffer
+	// is empty. Default 1s, lower in tests to make them faster.
+	waitIntervalPrepareRange time.Duration
 }
 
 // NewCaptive returns a new CaptiveStellarCore that is not running. Will lazily start a subprocess
@@ -86,10 +90,11 @@ func NewCaptive(executablePath, networkPassphrase string, historyURLs []string) 
 	}
 
 	return &CaptiveStellarCore{
-		archive:           archive,
-		networkPassphrase: networkPassphrase,
-		nextLedger:        0,
-		stellarCoreRunner: newStellarCoreRunner(executablePath, networkPassphrase, historyURLs),
+		archive:                  archive,
+		networkPassphrase:        networkPassphrase,
+		nextLedger:               0,
+		stellarCoreRunner:        newStellarCoreRunner(executablePath, networkPassphrase, historyURLs),
+		waitIntervalPrepareRange: time.Second,
 	}, nil
 }
 
@@ -274,7 +279,7 @@ func (c *CaptiveStellarCore) PrepareRange(ledgerRange Range) error {
 		if len(c.metaC) > 0 {
 			break
 		}
-		time.Sleep(time.Second)
+		time.Sleep(c.waitIntervalPrepareRange)
 	}
 
 	return nil
--- a/exp/ingest/ledgerbackend/captive_core_backend.go
+++ b/exp/ingest/ledgerbackend/captive_core_backend.go
@@ -288,11 +288,11 @@ func (c *CaptiveStellarCore) IsPrepared(ledgerRange Range) bool {
 	}
 
 	if ledgerRange.bounded {
-		return c.nextLedger >= ledgerRange.from &&
+		return c.nextLedger <= ledgerRange.from &&
 			c.nextLedger <= *c.lastLedger
 	}
 
-	return c.nextLedger >= ledgerRange.from
+	return c.nextLedger <= ledgerRange.from
 }
 
 // GetLedger returns true when ledger is found and it's LedgerCloseMeta.
--- a/exp/ingest/ledgerbackend/captive_core_backend.go
+++ b/exp/ingest/ledgerbackend/captive_core_backend.go
@@ -250,8 +250,8 @@ func (c *CaptiveStellarCore) readLedgerMetaFromPipe() (*xdr.LedgerCloseMeta, err
 // able to stream ledgers.
 // Set `to` to 0 to stream starting from `from` but without limits.
 func (c *CaptiveStellarCore) PrepareRange(ledgerRange Range) error {
-	if c.nextLedger != 0 && c.nextLedger >= ledgerRange.from {
-		// Range already prepared
+	// Range already prepared
+	if c.IsPrepared(ledgerRange) {
 		return nil
 	}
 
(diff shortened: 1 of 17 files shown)


bartekn:
I merged recent changes to `captive-core-online` branch and new tests found some issues in my code (@abuiles 👍). All is fixed now.

exp/ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -69,6 +69,10 @@ type CaptiveStellarCore struct {
 
 	nextLedger uint32  // next ledger expected, error w/ restart if not seen
 	lastLedger *uint32 // end of current segment if offline, nil if online
+
+	// waitIntervalPrepareRange defines a time to wait between checking if the buffer
+	// is empty. Default 1s, lower in tests to make them faster.
+	waitIntervalPrepareRange time.Duration
```
bartekn, line 75: Added this because `ledgerbackend` tests needed 10s (!) to run. With this changes its 0.3s.

exp/ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -322,12 +325,33 @@ func (c *CaptiveStellarCore) PrepareRange(ledgerRange Range) error {
...
+func (c *CaptiveStellarCore) IsPrepared(ledgerRange Range) bool {
+	if c.nextLedger == 0 {
+		return false
+	}
+
+	if c.lastLedger == nil {
+		return c.nextLedger <= ledgerRange.from
+	}
+
+	// c.lastLedger != nil: current range is bounded
```
abuiles, line 344: is this a missing comment? if not, I don't understand what are we trying to say with this comment -- maybe add an explanation ? same with the comment below (before `return false`)

services/horizon/internal/expingest/fsm.go
```diff
@@ -384,12 +409,27 @@ func (r resumeState) run(s *system) (transition, error) {
...
-		log.WithFields(logpkg.F{
+		logger := log.WithFields(logpkg.F{
 			"ingest_sequence": ingestLedger,
 			"core_sequence":   latestLedgerCore,
-		}).Info("Waiting for ledger to be available in stellar-core")
-		// Go to the next state, machine will wait for 1s. before continuing.
-		return retryResume(r), nil
+		})
+
+		if latestLedgerCore == ingestLedger-1 {
```
tamirms, line 417: should we check the number of the buffered ledgers in captive core? if the buffer is full and the latest ledger in the buffer happens to be `ingestLedger-1` then we actually want to fast forward to `latestLedgerCore`. Perhaps we should remove this case and a sleep duration to the `retryResume()` call when fast forwarding

services/horizon/internal/expingest/fsm.go
```diff
@@ -477,9 +517,9 @@ func (h historyRangeState) run(s *system) (transition, error) {
 		return start(), nil
 	}
 
-	err = s.ledgerBackend.PrepareRange(ledgerbackend.UnboundedRange(h.fromLedger))
```
tamirms, line 480: I think there might be a bug here from my PR which introduced the Range type. in `historyRangeState`, we should be ingesting a bounded range, not an unbounded range

services/horizon/internal/expingest/fsm.go
```diff
@@ -859,3 +899,37 @@ func (s *system) completeIngestion(ledger uint32) error {
 
 	return nil
 }
+
+// checkPrepareRange checks if the range is prepared. If not, it releases
+// the distributed ingestion lock and prepares the range.
+// The first return value determines if the caller should follow the transition.
+func (s *system) checkPrepareRange(from uint32, suggestCheckpoint bool) (bool, transition, error) {
```
tamirms, line 906: the function name `checkPrepareRange` seems to imply that it will do a read only operation. so it was surprising to see that we are doing some stateful operations. I think a better name for this function would be `maybePrepareRange()` because the function releases the lock and invokes `PrepareRange()` only if we haven't called `PrepareRange()`  before.

tamirms: also, I think we can remove the `suggestCheckpoint` param and the `transition` return value. instead, the caller can determine how it should transition if the returned bool is true.

services/horizon/internal/expingest/fsm.go
bartekn: We discussed outside GH. Both, bounded and unbounded ranges here would work but the unbounded range is faster because it doesn't require initializing Stellar-Core again to get `toLedger+1` and the following ledgers.

exp/ingest/ledgerbackend/captive_core_backend.go
bartekn: Updated the comment to be more clear.

services/horizon/internal/expingest/fsm.go
bartekn: > also, I think we can remove the suggestCheckpoint param and the transition return value. instead, the caller can determine how it should transition if the returned bool is true.

I left it as is because `maybePrepareRange` would need to return 2 bool (first: whether to move to `start` and second: if suggested checkpoint should be used) and error.

PR #2803, opened 2020-07-10
exp/ingest/ledgerbackend: Handle Stellar-Core process exit gracefully

### LINKED ISSUE #2705: Handle Stellar-Core errors and exits gracefully
`CaptiveStellarCore` does not handle Stellar-Core errors and exits gracefully:
* When Stellar-Core exits with non-zero exit code the next request to `CaptiveStellarCore` should fail with the error and users should be able to restart the process using `PrepareRange`.
* When Stellar-Core exits without an error make sure the code is able to restart the process if needed.

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This commit adds a code that handles Stellar-Core process exit gracefully. Added `processExit` unbuffered channel to `stellarCoreRunner` that receives the error in case Stellar-Core crashed or was terminated with an error or nil error if it was shutdown gracefully. The channel is later used in `PrepareRange` that returns an error if the process was stopped while prepare was running and in `sendLedgerMeta` that adds error to a meta channel that will be read by one of the following `GetLedger` calls.

Close #2705.

### Why

Right now when Stellar-Core crashes, `CaptiveStellarCore` simply wait for the next ledger without restarting it.

### DIFF
--- a/exp/ingest/ledgerbackend/captive_core_backend.go
+++ b/exp/ingest/ledgerbackend/captive_core_backend.go
@@ -49,13 +49,16 @@ type metaResult struct {
 }
 
 type CaptiveStellarCore struct {
+	executablePath    string
 	networkPassphrase string
 	historyURLs       []string
 
-	// read-ahead buffer
-	stop  chan struct{}
-	wait  sync.WaitGroup
+	// shutdown is a channel that triggers the backend shutdown by the user.
+	shutdown chan struct{}
+	// metaC is a read-ahead buffer.
 	metaC chan metaResult
+	// wait is a waiting group waiting for a read-ahead buffer to return.
+	wait sync.WaitGroup
 
 	stellarCoreRunner stellarCoreRunnerInterface
 	cachedMeta        *xdr.LedgerCloseMeta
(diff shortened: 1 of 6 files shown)


exp/ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -199,25 +203,31 @@ func (c *CaptiveStellarCore) sendLedgerMeta(untilSequence uint32) {
...
-			return
-		}
-		select {
-		case <-c.stop:
+			// When `GetLedger` sees the error it will close the backend. We shouldn't
+			// close it now because there may be some ledgers in a buffer.
+			c.metaC <- metaResult{nil, err}
 			return
-		case c.metaC <- metaResult{meta, nil}:
 		}
```
bartekn, line 229: I simplified this code. There's no need to check `c.stop` here because even we discard one result from `metaC` in `Close` to unblock it. So even if the code is waiting here, it will be unlocked by `Close`.

exp/ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -271,6 +286,20 @@ func (c *CaptiveStellarCore) PrepareRange(ledgerRange Range) error {
 	}
 
 	for {
+		select {
+		case <-c.shutdown:
+			return nil
```
abuiles, line 291: @bartekn should we call `c.Close()` here too before returning - is seem like `sendLedgerMeta` could still be running on the goroutine

bartekn: No, because `c.shutdown` can be only closed by calling `Close`. So it implies that `Close` was called.

PR #2809, opened 2020-07-13
tools/xdr2go: Transform base64 encoded XDR objects into a pretty Go code

bartekn:
### What

This commit adds a little CLI tool to transform base64 XDR objects into a pretty Go code. This helps in writing mocks and testing. It's using [`fmt.GoStringer`](https://golang.org/pkg/fmt/#GoStringer) interface to print pretty Go code compared to a standard library implementation.

### Why

Very often when writing tests we mock objects to make tests independent of other components. There are many ways we can create example XDR objects but they have disadvantages:
* For transactions we can use `txnbuild` package. The problem is that it allows building transactions only. Also, sometimes there's an existing transaction in the network we want to use in tests so building it using `txnbuild` requires extra time.
* `fmt.Printf("%#v", value)` prints a Go-syntax representation of the value. However, there are many problems with the way the values are printed using a standard `GoStringer`:
   * If the value is a struct with pointers it prints an hexadecimal memory address instead of it's value, ex:
        ```go
        &xdr.TransactionEnvelope{Type: 2, V0: (*xdr.TransactionV0Envelope)(nil), V1: (*xdr.TransactionV1Envelope)(0xc0000aa2d0), FeeBump: (*xdr.FeeBumpTransactionEnvelope)(nil)
        ```
    * It prints redundant `nil` values that could be skipped.
    * Binary values like ed25519 keys and unions (ex. `xdr.Asset`) are rendered in a way that's hard to read.
    * Union discriminants are printed as decimal, instead of using a type variable.
    * `uint32` values are printed in hex.
* I also checked [`https://godoc.org/github.com/kr/pretty`](https://godoc.org/github.com/kr/pretty). While it works fine, it also doesn't skip `nil` values and prints binary values and unions as standard Go printer.

Standard print:
```go
xdr.TransactionEnvelope{Type: 2, V0: (*xdr.TransactionV0Envelope)(nil), V1: (*xdr.TransactionV1Envelope)(0xc0000aa2d0), FeeBump: (*xdr.FeeBumpTransactionEnvelope)(nil)
```
With `GoStringer` implementations:
```go
xdr.TransactionEnvelope{Type: xdr.EnvelopeTypeEnvelopeTypeTx, V1: &xdr.TransactionV1Envelope{Tx: xdr.Transaction{SourceAccount: xdr.MustAddress("GAZ3T7HRWDBJ6SNQ7IWVUS65FP6QMCWHCALFYWX552KFV2O2RLOSRLKI"), Fee: 120, SeqNum: 122783846453215886, TimeBounds: &xdr.TimeBounds{MinTime: xdr.TimePoint(0), MaxTime: xdr.TimePoint(1594645065)}, Memo: xdr.Memo{Type: xdr.MemoTypeMemoNone}, Operations: []xdr.Operation{xdr.OperationBody{SourceAccount: &xdr.MustAddress("GBHC6AMZ3FWLYYHXITCIEZI6VXAU4IEMRCHLICXZXHOVSBFSWCRJ7JS7"), Body: &xdr.OperationBody{Type: xdr.OperationTypeManageSellOffer, ManageSellOfferOp: &xdr.ManageSellOfferOp{Selling: xdr.MustNewCreditAsset("LFEC", "GAG6FS3CR64QJHLHJU7HNXUB4KBLXVDFQBDXM5LG22WOM7CA2ITJAVD2"), Buying: xdr.MustNewNativeAsset(), Amount: 6, Price: xdr.Price{N: 9899999, D: 100000000}, OfferId: 0}}}}, Ext: xdr.TransactionExt{V: 0}}, Signatures: []xdr.DecoratedSignature{xdr.DecoratedSignature{Hint: xdr.SignatureHint{0xda, 0x8a, 0xdd, 0x28}, Signature: xdr.Signature{0x55, 0xb, 0xd0, 0x7d, 0xf3, 0x7, 0x71, 0x56, 0x99, 0x3c, 0x34, 0xfc, 0x47, 0xa0, 0xce, 0x2b, 0x39, 0xa, 0xc4, 0x8c, 0xb7, 0x80, 0x9f, 0x4c, 0xc8, 0x22, 0xae, 0xcc, 0xe9, 0x8b, 0x29, 0xb9, 0x80, 0x94, 0xab, 0x15, 0xbd, 0x6b, 0xc6, 0x3e, 0x2d, 0x12, 0x7a, 0x49, 0xa8, 0x83, 0x75, 0xdd, 0x21, 0x0, 0x14, 0x47, 0xdc, 0xf9, 0x4, 0xe7, 0xdb, 0x16, 0x5a, 0x6e, 0xb0, 0xd6, 0xc0, 0xd}}, xdr.DecoratedSignature{Hint: xdr.SignatureHint{0xb2, 0xb0, 0xa2, 0x9f}, Signature: xdr.Signature{0x6c, 0xb8, 0xe4, 0xd6, 0x39, 0xe2, 0x54, 0x3a, 0x51, 0xf1, 0xc1, 0x20, 0xb9, 0x5f, 0x7d, 0x8, 0xae, 0x31, 0x2c, 0xec, 0x19, 0xce, 0xb0, 0x3f, 0xb6, 0xe6, 0xfa, 0x25, 0xeb, 0x58, 0xf0, 0x33, 0xd1, 0x9f, 0x9a, 0xf5, 0xc2, 0x33, 0x7b, 0xa6, 0x45, 0x92, 0x18, 0xc0, 0x4d, 0xbf, 0x9d, 0xdf, 0xa0, 0xa5, 0x43, 0x73, 0x2a, 0x7c, 0x7d, 0xad, 0x67, 0x54, 0xad, 0xa6, 0xae, 0x30, 0xd1, 0xf}}}}}
```

### Known limitations

The currently used solution requires to implement [`fmt.GoStringer`](https://golang.org/pkg/fmt/#GoStringer) on structs that contain pointers or are rendered in an unreadable way.


bartekn:
> Can we add a minimal README? E.g. copying something from the PR explaining example motivation, and sample input and output would be cool.

Will do!

> Was there a reason you preferred to represent the signatures as raw bytes? I was wondering if there was any more compact way (I'm not thinking of one though).

I was thinking about hex but [`hex.DecodeString`](https://golang.org/pkg/encoding/hex/#DecodeString) returns two values so it's not possible to inline it.

tools/xdr2go/main.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+var (
+	typ string
+)
+
+var rootCmd = &cobra.Command{
+	Use:   "xdr2go [base64-encoded XDR object]",
+	Short: "xdr2go transforms base64 encoded XDR objects into a pretty Go code",
+	Run: func(cmd *cobra.Command, args []string) {
```
leighmcculloch, line 19: If you set `RunE` instead of `Run` then your function can return an `error`, and instead of printing errors throughout the application you can return errors and cobra will automatically print the error in its standard error format and include the usage help text, etc.

Example:
https://github.com/stellar/go/blob/787570f2f3dc6114baf5c96e5125a584319579e4/tools/stellar-key-gen/main.go#L28-L52

xdr/go_string.go
```diff
@@ -0,0 +1,130 @@
...
+	sb.WriteString("}")
+	return sb.String()
+}
+
+func (t *TimeBounds) GoString() string {
+	if t == nil {
+		return "nil"
+	}
+	return fmt.Sprintf("&xdr.TimeBounds{MinTime: xdr.TimePoint(%d), MaxTime: xdr.TimePoint(%d)}", t.MinTime, t.MaxTime)
+}
```
leighmcculloch, line 130: If we're going to add these functions to the `xdr` package I think we should add tests for them. If we expect these functions to always return valid Go code, I think we should try parsing them as well to check it can be round tripped back to the same object. The Go stdlib should have everything we need to do this. Then we'll know if things change that cause this to output invalid code.

xdr/go_string.go
```diff
@@ -0,0 +1,130 @@
...
+	default:
+		panic("Unknown type")
+	}
+	sb.WriteString("}")
+	return sb.String()
+}
+
+func (o Operation) GoString() string {
+	var sb strings.Builder
+	sb.WriteString("xdr.OperationBody{")
```
leighmcculloch, line 79: Should this be Operation instead of OperationBody?
```suggestion
	sb.WriteString("xdr.Operation{")
```

tools/xdr2go/main.go
```diff
@@ -0,0 +1,57 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	source := fmt.Sprintf("%#v\n", object)
+	formatted, err := format.Source([]byte(source))
+	if err != nil {
+		panic(err)
+	}
+	fmt.Println(string(formatted))
```
leighmcculloch, line 50: Nice use of the stdlib to format the code!

tools/xdr2go/main.go
```diff
@@ -0,0 +1,57 @@
...
+
+func main() {
+	rootCmd.Flags().StringVarP(&typ, "type", "t", "TransactionEnvelope", "xdr type, currently only TransactionEnvelope is available")
+	rootCmd.Execute()
+}
+
+func run(cmd *cobra.Command, args []string) {
+	if len(args) != 1 {
+		printHelpExit(cmd, "Exactly one command argument with XDR object is required.")
+	}
```
leighmcculloch, line 32: If you add to the command definition `Args: cobra.ExactArgs(1)` or `Args: cobra.MinimumNArgs(1)` Cobra will handle this validation and output the error in its standard format.

tools/xdr2go/main.go
```diff
@@ -0,0 +1,57 @@
...
+	Use:   "xdr2go [base64-encoded XDR object]",
+	Short: "xdr2go transforms base64 encoded XDR objects into a pretty Go code",
+	Run: func(cmd *cobra.Command, args []string) {
+		run(cmd, args)
+	},
+}
+
+func main() {
+	rootCmd.Flags().StringVarP(&typ, "type", "t", "TransactionEnvelope", "xdr type, currently only TransactionEnvelope is available")
+	rootCmd.Execute()
```
leighmcculloch, line 26: If you use a pattern like the below you could write some tests for this tool at the top-level which will help to illustrate how it works from the user perspective, and make reviewing PRs that change this code simpler because a reviewer will be able to see exactly how the behavior of the tool changes.

Example:

https://github.com/stellar/go/blob/787570f2f3dc6114baf5c96e5125a584319579e4/tools/stellar-key-gen/main.go#L12-L17

https://github.com/stellar/go/blob/787570f2f3dc6114baf5c96e5125a584319579e4/tools/stellar-key-gen/main_test.go#L72-L94

tools/xdr2go/main.go
```diff
@@ -0,0 +1,57 @@
...
+		panic(err)
+	}
+	fmt.Println(string(formatted))
+}
+
+func printHelpExit(cmd *cobra.Command, msg string) {
+	fmt.Println(msg)
+	cmd.Help()
+	os.Exit(-1)
+}
```
leighmcculloch, line 57: You can omit this code if you return errors to `RunE`.

xdr/go_string.go
```diff
@@ -0,0 +1,130 @@
...
+	return sb.String()
+}
+
+func (m MuxedAccount) GoString() string {
+	var sb strings.Builder
+	sb.WriteString(fmt.Sprintf("xdr.MuxedAccount{Type: %d,", m.Type))
+	accountID := m.ToAccountId()
+	switch m.Type {
+	case CryptoKeyTypeKeyTypeEd25519:
+		return fmt.Sprintf("xdr.MustAddress(%#v)", accountID.Address())
```
leighmcculloch, line 67: This return in here makes this function a little confusing because the earlier string builder doesn't get used. I think you should drop the string builder into the `CryptoKeyTypeKeyTypeMuxedEd25519` case statement if it's not going to be used here, and keep it in the scope it is needed.

bartekn:
@leighmcculloch great feedback and useful tips! Because this is a side-project I will add changes in the next couple days (more important things first).

tools/xdr2go/main.go
bartekn: Looks like it's not possible with the version we're using.

bartekn:
@leighmcculloch can you take a look. Adding tests helped me find a bunch of issues so I fixed them - good call! One think I'm not super happy about is we need `*Ptr` functions to create pointer to struct to be able to inline them.

Also, it looks like it should be possible to do some fuzz testing. Will add this to the list for smaller sprint.

tools/xdr2go/README.md
```diff
@@ -0,0 +1,27 @@
...
...FEC", "GAG6FS3CR64QJHLHJU7HNXUB4KBLXVDFQBDXM5LG22WOM7CA2ITJAVD2"), Buying: xdr.MustNewNativeAsset(), Amount: 6, Price: xdr.Price{N: 9899999, D: 100000000}, OfferId: 0}}}}, Ext: xdr.TransactionExt{V: 0}}, Signatures: []xdr.DecoratedSignature{xdr.DecoratedSignature{Hint: xdr.SignatureHint{0xda, 0x8a, 0xdd, 0x28}, Signature: xdr.Signature{0x55, 0xb, 0xd0, 0x7d, 0xf3, 0x7, 0x71, 0x56, 0x99, 0x3c, 0x34, 0xfc, 0x47, 0xa0, 0xce, 0x2b, 0x39, 0xa, 0xc4, 0x8c, 0xb7, 0x80, 0x9f, 0x4c, 0xc8, 0x22, 0xae, 0xcc, 0xe9, 0x8b, 0x29, 0xb9, 0x80, 0x94, 0xab, 0x15, 0xbd, 0x6b, 0xc6, 0x3e, 0x2d, 0x12, 0x7a, 0x49, 0xa8, 0x83, 0x75, 0xdd, 0x21, 0x0, 0x14, 0x47, 0xdc, 0xf9, 0x4, 0xe7, 0xdb, 0x16, 0x5a, 0x6e, 0xb0, 0xd6, 0xc0, 0xd}}, xdr.DecoratedSignature{Hint: xdr.SignatureHint{0xb2, 0xb0, 0xa2, 0x9f}, Signature: xdr.Signature{0x6c, 0xb8, 0xe4, 0xd6, 0x39, 0xe2, 0x54, 0x3a, 0x51, 0xf1, 0xc1, 0x20, 0xb9, 0x5f, 0x7d, 0x8, 0xae, 0x31, 0x2c, 0xec, 0x19, 0xce, 0xb0, 0x3f, 0xb6, 0xe6, 0xfa, 0x25, 0xeb, 0x58, 0xf0, 0x33, 0xd1, 0x9f, 0x9a, 0xf5, 0xc2, 0x33, 0x7b, 0xa6, 0x45, 0x92, 0x18, 0xc0, 0x4d, 0xbf, 0x9d, 0xdf, 0xa0, 0xa5, 0x43, 0x73, 0x2a, 0x7c, 0x7d, 0xad, 0x67, 0x54, 0xad, 0xa6, 0xae, 0x30, 0xd1, 0xf}}}}}
```
leighmcculloch, line 26: I think there could be some value base64 encoding the xdr.Signature output and changing the signature hint into a string.

```
xdr.Signature("VQvQffMHcVaZPDT8R6DOKzkKxIy3gJ9MyCKuzOmLKbmAlKsVvWvGPi0Sekmog3XdIQAUR9z5BOfbFlpusNbADQ==")
```

I don't think this has to happen now, but would be a nice improvement for later.

tools/xdr2go/main.go
```diff
@@ -0,0 +1,52 @@
...
+var (
+	typ string
+)
+
+var rootCmd = &cobra.Command{
+	Use:   "xdr2go [base64-encoded XDR object]",
+	Short: "xdr2go transforms base64 encoded XDR objects into a pretty Go code",
+	RunE: func(cmd *cobra.Command, args []string) error {
+		return run(cmd, args)
+	},
```
leighmcculloch, line 21:
```suggestion
	RunE: run(cmd, args),
```

xdr/asset.go
```diff
@@ -44,6 +44,28 @@ func MustNewCreditAsset(code string, issuer string) Asset {
...
+		copy(newCode[:], []byte(code)[:length])
+		a.Type = AssetTypeAssetTypeCreditAlphanum4
+		a.AssetCode4 = &newCode
+	case length >= 5 && length <= 12:
+		var newCode AssetCode12
+		copy(newCode[:], []byte(code)[:length])
+		a.Type = AssetTypeAssetTypeCreditAlphanum12
+		a.AssetCode12 = &newCode
+	default:
+		panic("Asset code length is invalid")
```
leighmcculloch, line 63: Can we change this to an error, make this `NewAllowTrustAsset` that returns an error, and then wrap it with `MustNewAllowTrustAsset` that panics the error? This is a common pattern, and it's good to allow others to use our SDK without panics.

tools/xdr2go/main.go
leighmcculloch: What do you think about adding some tests for the tool itself? For example, doing something like this ☝️.

tools/xdr2go/README.md
bartekn: We actually discussed it earlier: https://github.com/stellar/go/pull/2809#issuecomment-657608224. Can add this in 0.0.2.

tools/xdr2go/main.go
bartekn: I was thinking about it but we use CLI params (`-t`) and it complicates things a little. The tools is super small and for internal use only so I think it's OK now without command tests. We can add something later.

PR #2821, opened 2020-07-15
services/horizon: Require all captive Stellar-Core params or none

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Update config validation to require both captive core options to be set or unset. This is done because captive core is started when `--stellar-core-binary-path` is set, without checking `--enable-captive-core-ingestion` value:
https://github.com/stellar/go/blob/b74be4c0a9458382dd93f98ed57a61b35230e478/services/horizon/internal/init.go#L54-L71
So it's possible that Horizon user can accidentally start captive core ingestion without `--enable-captive-core-ingestion`.

### DIFF
--- a/services/horizon/cmd/root.go
+++ b/services/horizon/cmd/root.go
@@ -405,9 +405,7 @@ func initRootConfig() {
 		stdLog.Fatalf("--history-archive-urls must be set when --ingest is set")
 	}
 
-	if config.EnableCaptiveCoreIngestion && config.StellarCoreBinaryPath == "" {
-		stdLog.Fatalf("--stellar-core-binary-path must be set when --enable-captive-core-ingestion is set")
-	}
+	validateBothOrNeither("enable-captive-core-ingestion", "stellar-core-binary-path")
 
 	// Configure log file
 	if config.LogFile != "" {


2opremio:
> Update config validation to require both captive core options to be set or unset. This is done because captive core is started when `--stellar-core-binary-path` is set, without checking `--enable-captive-core-ingestion` value

Unfortunately, this PR broke the horizon docker image, which legitimately sets the core binary path beforehand, before knowing whether you want to run captive core or not:

https://github.com/stellar/go/blob/188c373a1924b97482b12384673ea94d07a3d9a6/services/horizon/docker/Dockerfile#L12

(As of now, running `services/horizon/docker/start.sh` leads to the horizon container failing with `Invalid config: stellar-core-binary-path = /usr/local/bin/stellar-core, but corresponding option stellar-core-config-path is not configured`)

I don't think the solution is to enforce both options at the same time, but only start captive core if `--enable-captive-core-ingestion` is set.

If we agree, I will create an issue for this.

2opremio:
CC @bartekn

bartekn:
@2opremio a quick fix is here: https://github.com/stellar/go/pull/2972 going to create an issue now.

PR #2826, opened 2020-07-16
services/horizon: Move trades  end-points to new action handler.

abuiles:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

 Move trades and trade_aggregration end-points to new action handler.

### Why

This is part of the action handlers refactoring.

For `/trade_aggregations` I had to add a new option to let handler specify their own custom page builder if they want to -- it turns out not all end-points follow the same pattern.


services/horizon/internal/actions/trade.go
```diff
@@ -0,0 +1,418 @@
...
+	err = ValidateCursorWithinHistory(pq)
+	if err != nil {
+		return nil, err
+	}
+
+	qp := TradesQuery{}
+	err = GetParams(&qp, r)
+	if err != nil {
+		return nil, err
+	}
```
2opremio, line 123:
```suggestion
	if err = GetParams(&qp, r); err != nil {
		return nil, err
	}
```

services/horizon/internal/actions/trade.go
```diff
@@ -0,0 +1,418 @@
...
+
+	if qp.OfferID != 0 {
+		trades = trades.ForOffer(int64(qp.OfferID))
+	}
+
+	var records []history.Trade
+	err = trades.Page(pq).Select(&records)
+	if err != nil {
+		return nil, err
+	}
```
2opremio, line 167:
```suggestion
	if err = trades.Page(pq).Select(&records); err != nil {
		return nil, err
	}
```

services/horizon/internal/actions/trade.go
```diff
@@ -0,0 +1,418 @@
...
+	err = ValidateCursorWithinHistory(pq)
+	if err != nil {
+		return nil, err
+	}
+
+	qp := TradeAggregationsQuery{}
+	err = GetParams(&qp, r)
+	if err != nil {
+		return nil, err
+	}
```
2opremio, line 256:
```suggestion
	if err = GetParams(&qp, r); err != nil {
		return nil, err
	}
```

services/horizon/internal/actions/trade.go
```diff
@@ -0,0 +1,416 @@
...
+
+	baseAssetID, err := historyQ.GetAssetID(*baseAsset)
+	if err != nil {
+		p := problem.BadRequest
+		if historyQ.NoRows(err) {
+			p = problem.NotFound
+			err = errors.New("not found")
+		}
```
2opremio, line 274: I may be wrong, but wasn't this automatically handled by the renderer?

services/horizon/internal/actions/trade.go
```diff
@@ -0,0 +1,416 @@
...
+
+	return nil
+}
+
+// GetTradeAggregationsHandler is the action handler for trade_aggregations
+type GetTradeAggregationsHandler struct {
+}
+
+// GetResourcePage returns a page of trade aggregations
+func (handler GetTradeAggregationsHandler) GetResourcePage(w HeaderWriter, r *http.Request) ([]hal.Pageable, error) {
```
2opremio, line 239: I would consider making this funcion a tad smaller by factoring out the record obtention

services/horizon/internal/actions/trade.go
```diff
@@ -0,0 +1,416 @@
...
+	ctx := r.Context()
+	pageQuery, err := GetPageQuery(r, DisableCursorValidation)
+	if err != nil {
+		return hal.Page{}, err
+	}
+	qp := TradeAggregationsQuery{}
+	err = GetParams(&qp, r)
+	if err != nil {
+		return hal.Page{}, err
+	}
```
2opremio, line 373:
```suggestion
	if err = GetParams(&qp, r); err != nil {
		return hal.Page{}, err
	}
```

services/horizon/internal/handler.go
```diff
@@ -409,7 +413,14 @@ func (handler pageActionHandler) renderPage(w http.ResponseWriter, r *http.Reque
...
 	}
 
-	page, err := buildPage(r, records)
+	var page hal.Page
+
+	if action, ok := handler.action.(withPageBuilderAction); ok {
+		page, err = action.BuildPage(r, records)
+	} else {
+		page, err = buildPage(r, records)
+	}
```
2opremio, line 422: I think this is pretty ugly (due to being implicit) I think it would be better to have a specialized `pageActionHandler` with custom rendering or parameterize `pageActionHandler` with a custom rendering function.

abuiles: @2opremio what about this? https://github.com/stellar/go/pull/2826/commits/05044287c6a117683a5caa4b09c65f9415f71735

services/horizon/internal/actions/trade.go
abuiles: I wonder the same when moving this code, but the scenario you are describing happens when trying to fetch a single resource which doesn't exist - however, this is a custom thing for this specific end-point which requires you to send a counter and base asset which should exist in the DB

services/horizon/internal/handler.go
2opremio: It's better, but instead of passing a function list (to do the trick of defaulting to nil) I would simply create an additional rest constructor which takes a _rest action with builder interface_ (the usual interface plus the builder interface)

abuiles: sounds good - done https://github.com/abuiles/go/commit/026b53a090930c1139940dc2714ff45349ef42ed

services/horizon/internal/actions/trade.go
abuiles: fixed 716b4434

services/horizon/internal/handler.go
```diff
@@ -363,17 +363,27 @@ type pageAction interface {
...
 func restPageHandler(action pageAction) pageActionHandler {
 	return pageActionHandler{action: action}
 }
 
+func restPageHandlerWithBuilder(action pageAction, customPageBuilder buildPageFun) pageActionHandler {
+	return pageActionHandler{
+		action:    action,
+		buildPage: customPageBuilder,
+	}
+}
```
2opremio, line 385: nitpick: I think it's nicer if you use an interface instead of a function

```suggestion
type pageActionHandler struct {
	action         pageAction
	streamable     bool
	streamHandler  sse.StreamHandler
	repeatableRead bool
	buildPage      buildPageFun
}

func restPageHandler(action pageAction) pageActionHandler {
	return pageActionHandler{action: action}
}

interface customBuiltPageAction {
	pageAction
	BuildPage (r *http.Request, records []hal.Pageable) (hal.Page, error)
}

func restCustomBuiltPageHandler(action customBuiltPageAction) pageActionHandler {
	return pageActionHandler{
		action:    action,
		buildPage: customBuiltPageAction.BuildPage,
	}
}
```

abuiles:
@2opremio applied your suggestion and fixed a couple of errors - should be good now.

Issue #2605, opened 2020-05-19
Add constraints to biginteger fields that accept positive values only


bartekn:
[`verify-range`](https://github.com/stellar/go/tree/master/services/horizon/docker/verify-range) tests revealed a bug in which `max_fee` value of [this](https://horizon.stellar.org/transactions/e615a8f6bff5ab52e19e21ebcdd439859aaf1cb2a56a23289ac3e3d01500bd7f) transaction has overflowed. This happened because the type of this field in XDR `Transaction` struct definition was (before fee bump changes) `uint32` but Postgres does not support unsigned integer types.

To catch this kind of errors faster we should add [constraints](https://www.postgresql.org/docs/9.4/ddl-constraints.html) to fields like `max_fee`. While this can stop ingestion on invalid value insert attempt it's safer.

PR #2827, opened 2020-07-17
services/horizon/internal/db2/schema/migrations: Add constraints on database columns which cannot hold negative values

tamirms:
…values

<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Fixes https://github.com/stellar/go/issues/2605

Add constraints on database columns which can only hold non-negative values.

> verify-range tests revealed a bug in which max_fee value of this transaction has overflowed. This happened because the type of this field in XDR Transaction struct definition was (before fee bump changes) uint32 but Postgres does not support unsigned integer types.

>To catch this kind of errors faster we should add constraints to fields like max_fee. While this can stop ingestion on invalid value insert attempt it's safer.


services/horizon/internal/db2/schema/migrations/38_add_constraints.sql
```diff
@@ -0,0 +1,115 @@
+-- +migrate Up
+
+ALTER TABLE history_transactions ADD CONSTRAINT valid_ledger_sequence CHECK (ledger_sequence >= 0) NOT VALID;
```
bartekn, line 3: There are many other fields that should be `>` instead of `>=`: fees, last_modified_ledger. Can you take another look at this?
```suggestion
ALTER TABLE history_transactions ADD CONSTRAINT valid_ledger_sequence CHECK (ledger_sequence > 0) NOT VALID;
```

PR #2834, opened 2020-07-20
services/horizon: Move path finding end-points to new actions.

abuiles:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Move path finding handler to /actions and move assertions helper to a share package.

### Why

This is part of https://github.com/stellar/go/issues/2582.

For this handler we need to use `ServerHTTP` since they don't follow the same page building pattern we use for other end-points. Maybe we could wrap it around `restPageHandler` but we'll need to land this first https://github.com/stellar/go/pull/2826/files?file-filters%5B%5D=#diff-b8d16027eadf970091bce20ac2f34450R418 which allow us to specify a page builder.

### DIFF
--- a/services/horizon/internal/actions/path.go
+++ b/services/horizon/internal/actions/path.go
@@ -13,7 +13,6 @@ import (
 	"github.com/stellar/go/services/horizon/internal/resourceadapter"
 	"github.com/stellar/go/services/horizon/internal/simplepath"
 	"github.com/stellar/go/support/render/hal"
-	"github.com/stellar/go/support/render/httpjson"
 	"github.com/stellar/go/support/render/problem"
 	"github.com/stellar/go/xdr"
 )
@@ -75,7 +74,7 @@ func (q StrictReceivePathsQuery) URITemplate() string {
 // Validate runs custom validations.
 func (q StrictReceivePathsQuery) Validate() error {
 	if (len(q.SourceAccount) > 0) == (len(q.SourceAssets) > 0) {
-		return sourceAssetsOrSourceAccount
+		return SourceAssetsOrSourceAccountProblem
 	}
 
 	err := ValidateAssetParams(
@@ -101,7 +100,8 @@ func (q StrictReceivePathsQuery) Validate() error {
 	return nil
 }
 
-var sourceAssetsOrSourceAccount = problem.P{
+// SourceAssetsOrSourceAccountProblem custom error where source assets or account is required
+var SourceAssetsOrSourceAccountProblem = problem.P{
 	Type:   "bad_request",
 	Title:  "Bad Request",
 	Status: http.StatusBadRequest,
@@ -109,14 +109,14 @@ var sourceAssetsOrSourceAccount = problem.P{
 		"Both fields cannot be present.",
 }
 
-// ServeHTTP implements the http.Handler interface
-func (handler FindPathsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+// GetResourcePage finds a list of strict receive paths
+func (handler FindPathsHandler) GetResourcePage(w HeaderWriter, r *http.Request) ([]hal.Pageable, error) {
+	var err error
 	ctx := r.Context()
 	qp := StrictReceivePathsQuery{}
-	err := GetParams(&qp, r)
-	if err != nil {
-		problem.Render(ctx, w, err)
-		return
+
+	if err = GetParams(&qp, r); err != nil {
+		return nil, err
 	}
 
 	query := paths.Query{}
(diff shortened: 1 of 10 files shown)


services/horizon/internal/actions/path.go
```diff
@@ -116,18 +112,8 @@ var sourceAssetsOrSourceAccount = problem.P{
...
-		}
-		problem.Render(ctx, w, err)
-		return
-	}
 	qp := StrictReceivePathsQuery{}
-	err := actions.GetParams(&qp, r)
+	err := GetParams(&qp, r)
 	if err != nil {
 		problem.Render(ctx, w, err)
 		return
```
2opremio, line 119:
```suggestion
	if err := GetParams(&qp, r); err != nil {
		problem.Render(ctx, w, err)
		return
```

services/horizon/internal/actions/path.go
```diff
@@ -116,18 +112,8 @@ var sourceAssetsOrSourceAccount = problem.P{
 // ServeHTTP implements the http.Handler interface
 func (handler FindPathsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
```
2opremio, line 113: Why can't you use `objectActionHandler` or a variant of it?

2opremio: You could use the same for `FeeStatsHandle`

services/horizon/internal/actions/path.go
abuiles: I dropped this suggestion, it breaks the tests after applying it - we actually can use the `err` variable here since we use it in other places below

2opremio: I don't understand, the code below won't execute of there is an error, so it's equivalent.

services/horizon/internal/actions/path.go
abuiles: For fee stats, we can use objectActionHandler since it returns a JSON object -- unfortunately `objectActionHandler` doesn't fit the bill for this resource not `pageActionHandler` --

The problem with `pageActionHandler` is that it build a a "paginatable" page - but the resource return from path finds doesn't have links, next, or previous.

To fix this nicely, we should allow for a handler which specifies its own page builder -- so we need to solve this https://github.com/stellar/go/pull/2826#discussion_r457581891 and then we can use the solution from there here.

abuiles:
not the one immediately below, but a few lines below there are other things
that declare `err` too -- this change is fine, but then we have to go and
do something similar in other places (in this same function) where `err`
can happen

2opremio:
> not the one immediately below, but a few lines below there are other things that declare `err` too -- this change is fine, but then we have to go and do something similar in other places (in this same function) where `err` can happen

Sorry, but I still don't get it. What's wrong with re-declaring err below?

services/horizon/internal/actions/path.go
abuiles: @2opremio I added a custom handler for the scenarios where we want to render a  simple base page

abuiles:
@2opremio addressed your comments 

About 

> Sorry, but I still don't get it. What's wrong with re-declaring err below?

There is nothing wrong with re-declaring -- I should have explained myself better - is not really that we have to do again `err := fund(); err ` but rather we end up needing a `var err error` -- so at the end I don't think doing the shortes version ends up bringing us any benefit. The code has been updated to use your recommendation.

services/horizon/internal/handler.go
```diff
@@ -264,6 +264,47 @@ func (handler objectActionHandler) ServeHTTP(
...
+	case render.MimeHal, render.MimeJSON:
+		response, err := handler.action.GetResourcePage(w, r)
+		if err != nil {
+			problem.Render(r.Context(), w, err)
+			return
+		}
+
+		httpjson.Render(
+			w,
+			handler.buildPage(response),
```
2opremio, line 299: Can't you use the `BuildPage` interface method from the prior PR?

abuiles: @2opremio no, it  it's a different interface - the previous one is when we want to build a full page, that is a page with  records and paginations  `hal.Page` -- this one is to allow user to build simple pages `hal.BasePage` which only contains a list or resources but no pagination or links data.

2opremio: But `httpjson.Render()` takes an `interface{}` so you could use both (i.e. you could make the`BuildPage` return an `interface{}` instead of `Page` or `BasePage`)

Admittedly, returning an interface is rather ugly though so I will leave it up to you. At the very least, we could avoid having `basePageActionHandler` by reusing `restPageHandler` (by storing an `interface{}`-returning build function)

abuiles: I'm okay having `basePageActionHandler` it makes it clear what type of page the end-point is going to return and we avoid using `interface{}`

PR #2835, opened 2020-07-20
services/horizon: Get /ledger/{ledger_id} fixes and refactorings

2opremio:
Follow up to #2829 . Part of #2582

### DIFF
--- a/services/horizon/internal/actions/helpers.go
+++ b/services/horizon/internal/actions/helpers.go
@@ -724,14 +724,6 @@ func GetParams(dst interface{}, r *http.Request) error {
 			if key == "*" {
 				continue
 			}
-			val := query.Get(key)
-			if len(val) > 0 {
-				return problem.MakeInvalidFieldProblem(
-					key,
-					errors.New("The parameter should not be included in the request"),
-				)
-			}
-
 			param, _ := GetURLParam(r, key)
 			query.Set(key, param)
 		}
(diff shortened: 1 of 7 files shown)


services/horizon/internal/actions/helpers.go
```diff
@@ -724,14 +724,6 @@ func GetParams(dst interface{}, r *http.Request) error {
 			if key == "*" {
 				continue
 			}
-			val := query.Get(key)
-			if len(val) > 0 {
```
abuiles, line 728: @2opremio what was the reason for this again? I think we want to explicitly do this so you don't end up with scenarios like the one you describe in slack `/foo/param1/bar?param1=baz`

2opremio: Why would we treat the query parameter key matching our (internal) URL path parameter placeholder specially?

The user can already supply other unused query parameters and we don't complain abiut. it.

2opremio: With this change we still do the right thing (use URL the parameter).

abuiles: @2opremio can you do a test -- what happens if you  send a request to `/offers/:offer_id` with 2 valid offer ids (both different) in the following form: `/offers/1234?offer_id=12345` -- does it server `1234` or `12345` ? I think that's why we have this check

2opremio: I already added a test for this. Or rather, I modified an existing one.

PR #2848, opened 2020-07-23
services/horizon: Get rid of App context and middleware

2opremio:
It is not needed. It removes a context variable (which is implicit and ugly)
and reduces the dependecnies between App and http/web handling (which I plan to split into
a separate package but right now has a cyclic dependency due App).

I will help wrap up #2582

### DIFF
--- a/services/horizon/internal/middleware.go
+++ b/services/horizon/internal/middleware.go
@@ -26,17 +26,6 @@ import (
 	"github.com/stellar/go/support/render/problem"
 )
 
-// appContextMiddleware adds the "app" context into every request, so that subsequence appContextMiddleware
-// or handlers can retrieve a horizon.App instance
-func appContextMiddleware(app *App) func(http.Handler) http.Handler {
-	return func(h http.Handler) http.Handler {
-		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-			ctx := withAppContext(r.Context(), app)
-			h.ServeHTTP(w, r.WithContext(ctx))
-		})
-	}
-}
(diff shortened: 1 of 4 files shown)


services/horizon/internal/middleware.go
```diff
@@ -225,21 +214,20 @@ func recoverMiddleware(h http.Handler) http.Handler {
 }
 
 // requestMetricsMiddleware records success and failures using a meter, and times every request
-func requestMetricsMiddleware(h http.Handler) http.Handler {
+func requestMetricsMiddleware(h http.Handler, web *web) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
leighmcculloch, line 218: If you define this middleware function so that it returns the function that meets the `func(http.Handler) http.Handler` type then this might be more intuitive to use since the call location won't need to wrap it in a function. 
```suggestion
func requestMetricsMiddleware(web *web) func(http.Handler) http.Handler {
	return func(h http.Handler) http.Handler {
		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```

It would reduce the call site to the following and self contain the concern of how to use it as middleware:
```go
r.Use(requestMetricsMiddleware(w))
```

Here's one example of where we use this pattern:
https://github.com/stellar/go/blob/e573454e6c5396d94555c3e1859b9173637cb735/support/http/logging_middleware.go#L14-L16

2opremio: Good point, I thought about it but I was being lazy.

PR #2857, opened 2020-07-28
services/horizon: Refactor http serving code

2opremio:
### What

* Factor out webserver code from App
* Create httpx.Server from web.go and webserving parts of App
* Factor out endpoint routing to httpx.Router
* Stop using tylerstillwater/graceful and use net/http.Server.Shutdown() instead
* Move handler.go to httpx
* Move all request context code into a single file of internal/context package

### Why

It paces the way to move the end-to-end actions tests and infra to a more suitable location to wrap up #2582 

### Known limitations

Still to do: move internal/{middleware_test.go, httpt_test.go, helpers_test.go, actions_*_test.go} to a more suitable location.


services/horizon/internal/app.go
```diff
@@ -501,17 +476,34 @@ func (a *App) init() {
...
+		SSEUpdateFrequency: a.config.SSEUpdateFrequency,
+		StaleThreshold:     a.config.StaleThreshold,
+		ConnectionTimeout:  a.config.ConnectionTimeout,
+		NetworkPassphrase:  a.config.NetworkPassphrase,
+		MaxPathLength:      a.config.MaxPathLength,
+		PathFinder:         a.paths,
+		MetricsRegistry:    a.metrics,
+		CoreGetter:         a,
+		HorizonVersion:     a.horizonVersion,
+		FriendbotURL:       a.config.FriendbotURL,
```
bartekn, line 498: I think we removed it completely from Horizon. Good opportunity to remove all friendbot references.

services/horizon/internal/context/context.go
```diff
@@ -0,0 +1,63 @@
+package context
+
+import (
```
bartekn, line 3: All the methods here are connected to HTTP server. Should we just move it to http package?

services/horizon/internal/middleware_test.go
```diff
@@ -38,7 +41,11 @@ func (suite *RateLimitMiddlewareTestSuite) SetupTest() {
 		MaxRate:  throttled.PerHour(10),
 		MaxBurst: 9,
 	}
-	suite.app = NewApp(suite.c)
+	app, err := NewApp(suite.c)
```
bartekn, line 44: Not required in this PR but we could remove dependency on `App` here.

services/horizon/internal/context/context.go
2opremio: It's not possible due to a dependency loop.

Right now: 

```
actions -> context
httpx -> context
resourceadapter -> context
actions -> resourceadapter
httpx -> actions (to set the routes)
```
If we move context to httpx:

```
actions -> httpx 
httpx -> actions
actions -> resouceadapter
resourceadapter -> httpx
```

Also,

services/horizon/internal/middleware_test.go
2opremio: Yes, that's the plan for an upcoming PR

services/horizon/internal/app.go
2opremio: Unfortunately we haven't completely removed it:

https://github.com/stellar/go/blob/7ed729bc813a197c41ce54b9d689b14683b9c09b/services/horizon/internal/resourceadapter/root.go#L36-L40

Issue #2628, opened 2020-05-27
Horizon returned bad sequence error although the transaction was added to ledger


bantalon:
### What version are you using?

I am working on the public network through the public Horizon instance: https://horizon.stellar.org. I posted transactions to the network using this API: https://www.stellar.org/developers/horizon/reference/endpoints/transactions-create.html.

### What did you do?

On may 8, I started posting a transaction on 06:47:19 UTC. 30 seconds later, on 06:47:49, a https://www.stellar.org/developers/horizon/reference/errors/timeout.html was returned. As instructed in the documentation, I retried sending the same transaction after a second delay. Within less than a second I got a bad sequence failure:
```
{
 "type": "https://stellar.org/horizon-errors/transaction_failed",
 "title": "Transaction Failed",
 "status": 400,
 "detail": "The transaction failed when submitted to the stellar network. The `extras.result_codes` field on this response contains further details. Descriptions of each code can be found at: https://www.stellar.org/developers/guides/concepts/list-of-operations.html",
 "extras": {
 "envelope_xdr": "AAAAAGRFCbG4m7/GKYJDKVwWrhSAhfgjSjOILbb0cOAN+rH+AAAAZAGfEqQAAC2bAAAAAQAAAABetP9IAAAAAF61DoQAAAABAAAAG1hMTSBlMmUgbW9uaXRvciB0cmFuc2FjdGlvbgAAAAABAAAAAAAAAAEAAAAA+dkdOXuuwsV5OpRuUk9Owqe6ocz1ukhgrYqnKliCQA8AAAAAAAAAAAAAAAEAAAAAAAAAAQ36sf4AAABAhnftLR7Lx9hKXojyGkKDLu4bdI4+wgx/qSPxEW3Vr9CAWBDQPngRWihDppCNzpmKTn3Q3wsM0z3qOJl6+72zCA==",
 "result_codes": {
 "transaction": "tx_bad_seq"
 },
 "result_xdr": "AAAAAAAAAAD////7AAAAAA=="
 }
}
```
My code treated this as failure.
However, I was surprised to find the transaction on the chain: https://stellar.expert/explorer/public/tx/45a9674370754d50b50df23e03d7fa87bd6ee70bd2db2431a8ab874218c3115a.

### What did you expect to see?
I expected to see a successful submit result, either on the first attempt which ended in timeout or on the second attempt which ended in `tx_bad_seq`.

### What did you see instead?
I got a wrong misleading `tx_bad_seq` failure.

[...]

PR #2877, opened 2020-08-03
services/horizon/internal/txsub: Fix false positive bad sequence error when submitting transactions

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Close https://github.com/stellar/go/issues/2628

Horizon can respond with a false positive bad sequence error when Horizon's ingestion lags behind Stellar Core. To avoid sending back this misleading error we should check that Horizon's ingestion is consistent with Stellar Core before checking whether the transaction exists in the Horizon DB.

We are able to check that Horizon's DB is consistent with Stellar Core indirectly by observing the sequence number of the source account of the transaction. If the sequence number of the source account is below the sequence number set in the transaction, then we know Horizon is still behind Stellar Core. Once the sequence number equals or exceeds the transaction sequence number we know that we are consistent with Stellar Core at the point where Stellar Core returned the bad sequence error.


bartekn:
Please update CHANGELOG.

PR #2879, opened 2020-08-04
services/horizon: Add Golang version to horizon version output

bartekn:
Adds Golang runtime version to `horizon version` output. Will be helpful to debug potential issues in the future.

### DIFF
--- a/services/horizon/cmd/version.go
+++ b/services/horizon/cmd/version.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"fmt"
+	"runtime"
 
 	"github.com/spf13/cobra"
 	apkg "github.com/stellar/go/support/app"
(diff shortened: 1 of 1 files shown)


services/horizon/cmd/version.go
```diff
@@ -2,17 +2,19 @@ package cmd
...
 )
 
 var versionCmd = &cobra.Command{
 	Use:   "version",
-	Short: "print horizon version",
+	Short: "print horizon and Golang runtime version",
 	Long:  "",
 	Run: func(cmd *cobra.Command, args []string) {
 		fmt.Println(apkg.Version())
+		fmt.Println(runtime.Version())
```
ire-and-curses, line 17: What does the output look like? Is it obvious which is which? Do we need to add a string like "Go runtime""?

bartekn: For my local build it's: `go1.14.5`.

Issue #2841, opened 2020-07-21
Captive Stellar-Core wrapper server


bartekn:
The current implementation of Captive Stellar-Core streams meta stream using a filesystem pipe. This implies that both Horizon and Stellar-Core have to be deployed to the same server. One of the disadvantages of such requirement is a need for detailed per-process monitoring to be able to connect potential issues (like memory leaks) to the specific service.

To solve this we can build a thin wrapper around `CaptiveStellarCore` backend that exposes a simple RPC/HTTP API similar to `LedgerBackend` interface:
https://github.com/stellar/go/blob/f77b3b518d6d5233a948e2795bd2bd9dc204c035/exp/ingest/ledgerbackend/ledger_backend.go#L26-L40

The full solution requires a server and a client that can consume the API.

A server can be a subcommand on `horizon` that exposes the `CaptiveStellarCore` object API using either the [`net/rpc`](https://golang.org/pkg/net/rpc/) package or HTTP. It has the same require params as `CaptiveStellarCore`. A server is a singleton, meaning it can start a single `stellar-core` process, thus should be used by a single Horizon instance.

A client, `RemoteCaptiveStellarCore` implementation of `LedgerBackend` interface is part of `exp/ingest/ledgerbackend` and accepts a connection params to a server.

PR #2896, opened 2020-08-06
exp/services/captivecore/internal: Add HTTP server for managing a remote captive core instance

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add HTTP server for managing a remote captive core instance.

### Why

From https://github.com/stellar/go/issues/2841 :

The current implementation of Captive Stellar-Core streams meta stream using a filesystem pipe. This implies that both Horizon and Stellar-Core have to be deployed to the same server. One of the disadvantages of such requirement is a need for detailed per-process monitoring to be able to connect potential issues (like memory leaks) to the specific service.

To solve this we can build a thin wrapper around CaptiveStellarCore backend that exposes a simple RPC/HTTP API similar to LedgerBackend interface:
```golang
 // LedgerBackend represents the interface to a ledger data store. 
 type LedgerBackend interface { 
 	// GetLatestLedgerSequence returns the sequence of the latest ledger available 
 	// in the backend. 
 	GetLatestLedgerSequence() (sequence uint32, err error) 
 	// The first returned value is false when the ledger does not exist in a backend. 
 	GetLedger(sequence uint32) (bool, xdr.LedgerCloseMeta, error) 
 	// PrepareRange prepares the given range (including from and to) to be loaded. 
 	// Some backends (like captive stellar-core) need to initalize data to be 
 	// able to stream ledgers. 
 	PrepareRange(ledgerRange Range) error 
 	// IsPrepared returns true if a given ledgerRange is prepared. 
 	IsPrepared(ledgerRange Range) bool 
 	Close() error 
 } 
```
The full solution requires a server and a client that can consume the API.

A server can be a subcommand on horizon that exposes the CaptiveStellarCore object API using either the net/rpc package or HTTP. It has the same require params as CaptiveStellarCore. A server is a singleton, meaning it can start a single stellar-core process, thus should be used by a single Horizon instance.

A client, RemoteCaptiveStellarCore implementation of LedgerBackend interface is part of exp/ingest/ledgerbackend and accepts a connection params to a server.

### Known limitations

I will add tests in a subsequent commit. Right now I'm looking for a preliminary review


exp/services/captivecore/internal/server.go
```diff
@@ -0,0 +1,70 @@
...
+			return
+		}
+
+		response, err := api.GetLedger(req.Sequence)
+		serializeResponse(api.log, w, r, response, err)
+	})
+
+	mux.Post("/prepare-range", func(w http.ResponseWriter, r *http.Request) {
+		ledgerRange := ledgerbackend.Range{}
+		if err := json.NewDecoder(r.Body).Decode(&ledgerRange); err != nil {
```
bartekn, line 59: Nit: I think form urlencoded would be much simpler here. JSON is fine but we don't really need it here.

exp/services/captivecore/internal/api.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	seq, err := c.core.GetLatestLedgerSequence()
+	return LatestLedgerSequenceResponse{Sequence: seq}, err
+}
+
+// LedgerResponse is the response for the GetLedger command.
+type LedgerResponse struct {
+	Present bool                `json:"present"`
+	Ledger  xdr.LedgerCloseMeta `json:"ledger"`
```
bartekn, line 170: Does it render as base64? If not the value will be really hard to parse.

exp/services/captivecore/internal/server.go
tamirms: I think it's slightly easier for the client code to use JSON because serializing to JSON would require less code than constructing an HTTP form. Also, it is nice to make the content type consistent between the request and response bodies.

tamirms:
@bartekn can you take another look? I've added tests now

exp/services/captivecore/internal/server.go
```diff
@@ -0,0 +1,70 @@
...
+) {
+	if err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		w.Write([]byte(err.Error()))
+		return
+	}
+
+	w.Header().Set("Content-Type", "application/json; charset=utf-8")
+	if err := json.NewEncoder(w).Encode(response); err != nil {
+		logger.WithContext(r.Context()).WithError(err).Warn("could not serialize response")
```
bartekn, line 28: Nit: I guess we can write a header here.
```suggestion
		logger.WithContext(r.Context()).WithError(err).Warn("could not serialize response")
		w.WriteHeader(http.StatusInternalServerError)
```

PR #2897, opened 2020-08-06
services/horizon: Add sponsors ingestion to change processors

bartekn:
<details>
  <summary>PR Checklist</summary>
  
### PR Structure

* [x] This PR has reasonably narrow scope (if not, break it down into smaller PRs).
* [x] This PR avoids mixing refactoring changes with feature changes (split into two PRs
  otherwise).
* [x] This PR's title starts with name of package that is most changed in the PR, ex.
  `services/friendbot`, or `all` or `doc` if the changes are broad or impact many
  packages.

### Thoroughness

* [x] This PR adds tests for the most critical parts of the new functionality or fixes.
* [x] I've updated any docs ([developer docs](https://www.stellar.org/developers/reference/), `.md`
  files, etc... affected by this change). Take a look in the `docs` folder for a given service,
  like [this one](https://github.com/stellar/go/tree/master/services/horizon/internal/docs).

### Release planning

* [x] I've updated the relevant CHANGELOG ([here](services/horizon/CHANGELOG.md) for Horizon) if
  needed with deprecations, added features, breaking changes, and DB schema changes.
* [x] I've decided if this PR requires a new major/minor version according to
  [semver](https://semver.org/), or if it's mainly a patch change. The PR is targeted at the next
  release branch if it's not a patch change.
</details>

Add sponsors ingestion to change processors (CAP-33).

Close #2889.

PR #2898, opened 2020-08-06
services/horizon/expingest: Add state and ledger processor for Claimable balances.

abuiles:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add state and ledger processor for claimable balance entries.

### Why

This is a new type of entry in protocol 14 which we need to ingest.

The only update on  on ClaimableBalances is sponsor, so I didn't add `Update` support on this PR since we might be able to reuse some of the code from #2897 

#### Testing

Run a standalone network pointing to `image: stellar/stellar-core:14.0.0rc1-1300-981385ce`

Grab `txnbuild` from https://github.com/stellar/go/pull/2903#pullrequestreview-463483508 and create and claim claimable balances with the following ops:

```

op := txnbuild.CreateClaimableBalance{
	Destinations: []string{
		"GCFEQ7DVT2DXKFKHGFWHMCSRYMFDGUADPNF7KFPAE3C2P7YUQBMH4DG2", // must exist
		"GDP5EHRW3RXYQFMGWM3AXDIO34NHU7VS3DEGZNLQRWAES2FQ23DTNL7Y", // must exist
	},
	Amount: "20",
	Asset:  txnbuild.NativeAsset{},
}

op := txnbuild.ClaimClaimableBalance{
	BalanceID: "AAAAABchk5v+5pIqF2GPoFbSQqxQBBVbtrLuY78jgs0btHFy", // grab it from horizon db for now select balance_id from claimable_balances;
}
```

### Known limitations

[TODO or N/A]


services/horizon/internal/db2/history/claimable_balances.go
```diff
@@ -0,0 +1,94 @@
...
+	return hex.EncodeToString(sum[:]), nil
+}
+
+func (i *claimableBalancesBatchInsertBuilder) Add(entry *xdr.LedgerEntry) error {
+	cBalance := entry.Data.MustClaimableBalance()
+	id, err := balanceIDToHex(cBalance.BalanceId)
+	if err != nil {
+		return errors.Wrap(err, "encoding balanceID")
+	}
+	row := ClaimableBalance{
```
abuiles, line 72: I'm not ingesting yet the claimants -- I'm having seconds thoughts about putting in a different table and considering a embedded array. I'll explore that path

abuiles: I have a specific proposal for this now. Working on the implementation.

services/horizon/internal/db2/history/claimable_balances.go
```diff
@@ -0,0 +1,176 @@
...
+		return 0, err
+	}
+
+	return result.RowsAffected()
+}
+
+// FindClaimableBalancesByDestination finds all claimable balances where accountID is one of the claimants
+func (q *Q) FindClaimableBalancesByDestination(accountID xdr.AccountId) ([]ClaimableBalance, error) {
+	sql := selectClaimableBalances.
+		Where("claimants @> '[{\"destination\": \"" + accountID.Address() + "\"}]'")
```
abuiles, line 110: unfortunately, we can't use `?` here since it would produce an invalid  pg JSON

services/horizon/internal/db2/history/claimable_balances.go
```diff
@@ -0,0 +1,176 @@
...
+	if !ok {
+		return errors.New("type assertion to []byte failed")
+	}
+
+	return json.Unmarshal(b, &c)
+}
+
+type Claimant struct {
+	Destination string             `json:"destination"`
+	Predicate   xdr.ClaimPredicate `json:"predicate"`
```
abuiles, line 44: nit: this value is getting stored as a json -- but we want it as base64

abuiles: fixed

services/horizon/internal/db2/history/claimable_balances.go
```diff
@@ -0,0 +1,199 @@
...
+	dbClaim.Predicate = predicate
+
+	return json.Marshal(dbClaim)
+}
+
+func (c *Claimant) UnmarshalJSON(data []byte) error {
+	var dbClaim struct {
+		Destination string `json:"destination"`
+		Predicate   string `json:"predicate"`
+	}
```
bartekn, line 67: It's also used in marshal method so maybe let's move it outside?

services/horizon/internal/db2/schema/migrations/39_claimable_balances.sql
```diff
@@ -3,25 +3,18 @@
...
-    id character varying(64) NOT NULL, -- sha256 of balance_id in HEX
-    destination character varying(64) NOT NULL,
-    predicate TEXT NOT NULL,
-    PRIMARY KEY (id, destination)
-);
-
 CREATE INDEX claimable_balances_by_asset ON claimable_balances USING BTREE(asset);
 CREATE INDEX claimable_balances_by_sponsor ON claimable_balances USING BTREE(sponsor);
-CREATE INDEX claimable_balances_claimants_by_destination ON claimable_balances_claimants USING BTREE(destination);
+CREATE INDEX claimabable_balances_by_claimants ON claimable_balances USING gin(claimants jsonb_path_ops);
```
bartekn, line 16: Can we do some testing how fast are select queries using this index (ex. with a few million balances)? We don't specify a field to use so I just wonder if it will be fast enough.

services/horizon/internal/db2/history/claimable_balances.go
abuiles: fixed.

services/horizon/internal/db2/schema/migrations/39_claimable_balances.sql
abuiles: @bartekn sure, I can do some testing.  For the kind of search we want to do I ended up going with `gin(.. jsonb_path_ops)` index since it produces a specific index for the value and the key leading to it.

Since our searches look like `{"destinatinon": "G...Z"}`, we can take advantage of it and let PG search against the hash of `{"destination": "value"}` which will use the index. (See https://www.postgresql.org/docs/current/datatype-json.html#JSON-INDEXING for more info)

abuiles: @bartekn I did some test with a table containing 3M records 

```
                            List of relations
 Schema |       Name       | Type  |    Owner     |  Size   | Description 
--------+------------------+-------+--------------+---------+-------------
 public | accounts         | table | ab | 33 MB   | 
 public | balances         | table | ab | 426 MB  | 
```

With the following schema

```
CREATE TABLE accounts(address text NOT NULL);
CREATE TABLE balances(claimants JSONB NOT NULL);
CREATE INDEX balances_gin_idx ON balances USING gin(claimants jsonb_path_ops);
```

I created 500K "accounts" 

```
INSERT INTO accounts(address) SELECT md5(RANDOM()::TEXT) FROM generate_series(1, 500000);
```

And then for each account I created 4 entries in the balances table (run 4 times)

```
INSERT INTO balances(claimants)
SELECT json_build_array(json_build_object('destination',a.address,'predicate', md5(RANDOM()::TEXT))) FROM accounts as a;
```

And finally picked some random destinations with 

```
SELECT json_build_array(json_build_object('destination',a.address)) FROM accounts as a ORDER BY RANDOM () LIMIT 1;
```

Which I then use on the search query

```
SELECT * FROM balances WHERE claimants @> '[{"destination" : "d76027c68d53d40fc2f6c13fee906e9a"}]';
```

Overall the search happens very fast 

```
EXPLAIN ANALYZE SELECT * FROM balances WHERE claimants @> '[{"destination" : "d76027c68d53d40fc2f6c13fee906e9a"}]';
                                                         QUERY PLAN                                                          
-----------------------------------------------------------------------------------------------------------------------------
 Bitmap Heap Scan on balances  (cost=47.25..9741.12 rows=3001 width=113) (actual time=0.061..0.100 rows=6 loops=1)   Recheck Cond: (claimants @> '[{"destination": "d76027c68d53d40fc2f6c13fee906e9a"}]'::jsonb)
   Heap Blocks: exact=6
   ->  Bitmap Index Scan on balances_gin_idx  (cost=0.00..46.50 rows=3001 width=0) (actual time=0.036..0.036 rows=6 loops=1)
         Index Cond: (claimants @> '[{"destination": "d76027c68d53d40fc2f6c13fee906e9a"}]'::jsonb)
 Planning time: 0.075 ms
 Execution time: 0.122 ms
(7 rows)
```

I also created a fiddle where you can experiment http://sqlfiddle.com/#!17/42d12a/4

PR #2900, opened 2020-08-07
Update horizon-cmp for 1.7 release

2opremio:
Remove trailing data fix in SSE responses to avoid false positive diffs.

### DIFF
--- a/tools/horizon-cmp/internal/response.go
+++ b/tools/horizon-cmp/internal/response.go
@@ -67,6 +67,11 @@ var replaceRegexps = []replace{
 		`"max_fee":( ?)([\d]+),`),
 		`"max_fee":${1}"${2}",`,
 	},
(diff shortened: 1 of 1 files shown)


tools/horizon-cmp/internal/response.go
```diff
@@ -67,6 +67,11 @@ var replaceRegexps = []replace{
 		`"max_fee":( ?)([\d]+),`),
 		`"max_fee":${1}"${2}",`,
 	},
+	// Removes trailing SSE data, fixed in horizon 1.7.0
+	{regexp.MustCompile(
+		`\nretry:.*\nevent:.*\ndata:.*\n`),
```
bartekn, line 72: Is this happening for event names other than `close`?

2opremio: It only happens at the end, so I guess that's a no

Issue #2812, opened 2020-07-14
Create integration tests for different ingestion backends


bartekn:
TBA I'm writing a doc about it but basically we need tests that check if a compiled binary ingest ledgers using different backends and in single/distributed ingestion environment. Testing it manually is error prone and slow.

bartekn:
We discussed it a little bit in Horizon Team meeting and we agreed to start with a small test and then, maybe, build a framework if we find integration tests really useful and time saving.

We should follow the following points when running integration tests:
* Obtain a system under test.
* Seed necessary test data.
* Perform actions using the system under test.
* Verify behaviours.

In the first iteration this is how it could look like:
* **Obtain a system under test** Start a standalone network, simply by using a bash script. See the code in `stellar/quickstart` to understand how to do it. See if https://github.com/bats-core/bats-core recommended by @2opremio can be helpful here.
* **Seed necessary test data** Use `txnbuild` or `xdr2go` to submit some transactions. In the future we can start the network at new protocol version and submit new operations. We can also sync to pubnet or testnet to use existing network data as a seed.
* **Perform actions using the system under test** Use `txnbuild` or `xdr2go` to submit some transactions. Use `horizonclient` to get Horizon responses to check.
* **Verify behaviours** We should use the standard `testing` package test function and `testify/assert` for assertions. Tests should be part of `services/horizon` package. Should be skipped using [`T.Skip`](https://golang.org/pkg/testing/#T.Skip) if not in integration testing environment (ex. check Circle-CI env variable).

The simplest test can:
1. Connect to pubnet/testnet using Stellar-Core and start Horizon ingesting node.
2. Submit a transaction creating a new account sending payment to self.
3. Confirm if the transaction and the account changes are available via Horizon API.

bartekn:
The simple integration tests framework was added in https://github.com/stellar/go/pull/2916. However, we will change the engine to use `stellar/horizon` instead of `stellar/quickstart`. This will make this more flexible. Moving to backlog.

tamirms:
Improvements to the current implementation are tracked in https://github.com/stellar/go/issues/3031

PR #2910, opened 2020-08-11
services/horizon: Add `horizon ingest init-genesis-state` command

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Adds `horizon ingest init-genesis-state` command that ingest state as of genesis ledger (seq=1).

### Why

Horizon is using the latest History Archive State for checking the latest checkpoint ledger however Stellar-Core does not publish HAS until ledger which sequence equals 63. Because of this, when starting a standalone network Horizon needs to wait around 5 minutes/63 ledgers to start ingesting. This command should solve this issue by ingesting the genesis ledger of a given network so Horizon can start ingestion without waiting for Stellar-Core to publish the first checkpoint.

Required by https://github.com/stellar/go/issues/2812.

### Known limitations

There is a code that inits Stellar-Core setting so `ingest.System` can create a `ledgerbackend` however this is not needed because `ledgerbackend` is not used when ingesting genesis ledger. This problem also occurs in `stress-test`command. I didn't want to fix this here to make this PR simple. Should be fixed in another PR in the future.

### DIFF
--- a/services/horizon/cmd/ingest.go
+++ b/services/horizon/cmd/ingest.go
@@ -238,6 +238,62 @@ var ingestTriggerStateRebuildCmd = &cobra.Command{
 	},
 }
 
+var ingestInitGenesisStateCmd = &cobra.Command{
+	Use:   "init-genesis-state",
+	Short: "ingests genesis state (ledger 1)",
+	Run: func(cmd *cobra.Command, args []string) {
+		initRootConfig()
+
+		horizonSession, err := db.Open("postgres", config.DatabaseURL)
+		if err != nil {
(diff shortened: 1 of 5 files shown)


bartekn:
It may not be clear where the genesis ingestion actually takes place. It's in `RunHistoryArchiveIngestion`:
https://github.com/stellar/go/blob/742703c04ccd25552fd4108d5b60aa33e0f4bfa8/services/horizon/internal/expingest/processor_runner.go#L169-L190

Issue #2787, opened 2020-07-06
[META] CAP23 support: Two-Part Payments with ClaimableBalanceEntry


abuiles:
## CAP23 support: Two-Part Payments with ClaimableBalanceEntry
This proposal introduces new operations that separate sending a payment from receiving the payment. [CAP 23](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0023.md) introduces a new ledger key type and a new set of operations.

This issue keeps track of related work to support CAP23 in the ingestion system, web API, and SDKs.

## Ingestion 
We need to extend the ingestion system to support CAP23. Depends on: 
- [x] https://github.com/stellar/go/issues/2858
- [x] https://github.com/stellar/go/issues/2882
- [x] https://github.com/stellar/go/issues/2887
- [x] https://github.com/stellar/go/issues/2886
- [x] https://github.com/stellar/go/issues/2885
- [x] #2882

## Web API

- [x] Web API  https://github.com/stellar/go/issues/2860

## SDKs

**TODO:** Create sub-issue for SDK related work, depends on Web API work #2860

- [ ] Support loading and filtering multiple claimable_balances.
- [ ] Support loading a single claimable_balance.
- [ ] Add support for claimable balance operations.

tomerweller:
The suggested payload format of predicates is pretty cryptic right now. Horizon usually presents more readable resources.

There's some prior work around predicate trees in a JSON format. One notable example is mongodb's query [language](https://docs.mongodb.com/manual/reference/operator/query/and/).
 
I propose moving from a `{"type": i, "value": x}` format to a `{i_str: x}` format. 

## Example
### Original 
```json
"predicate": {
  "type": 1,
  "value": [
    {
      "type": 4,
      "value": "1594066025"
    },
    {
      "type": 3,
      "value": "1594066090"
    }
  ]
}
```

### Proposed
```json
"predicate": {
  "and" : [
    {"after": 1594066025},
    {"before": 1594066090}
  ]
}
```

abuiles:
Horizon support for CAP23 has been completed.

PR #2911, opened 2020-08-11
services/horizon/internal/expingest: Ingest CAP23 Operations details and participants.

### LINKED ISSUE #2886: Update ingestion operation processor to ingest claimable balance related operations
- [x] Ingest `CreateClaimableBalanceOp`.
    - [x] Update Details in `transactionOperationWrapper`
    - [x] Update Participants in `transactionOperationWrapper`
- [x] Ingest `ClaimClaimableBalanceOp`.
   - [x] Update Details in `transactionOperationWrapper`
    - [x] Update Participants in `transactionOperationWrapper`

abuiles:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Ingest details and participants involved in `CreateClaimableBalanceOp` and `ClaimClaimableBalanceOp`.

### Why

CAP23 introduces 2 new operations, we need to ingest the details and participants involved in each operation. 

Fix #2886.

### Known limitations

[TODO or N/A]

### DIFF
--- a/services/horizon/internal/expingest/processors/operations_processor.go
+++ b/services/horizon/internal/expingest/processors/operations_processor.go
@@ -296,8 +296,15 @@ func (operation *transactionOperationWrapper) Details() map[string]interface{} {
 			})
 		}
 		details["claimants"] = claimants
-	case xdr.OperationTypeClaimClaimableBalance,
-		xdr.OperationTypeBeginSponsoringFutureReserves,
+	case xdr.OperationTypeClaimClaimableBalance:
+		op := operation.operation.Body.MustClaimClaimableBalanceOp()
+		balanceID, err := xdr.MarshalBase64(op.BalanceId)
+		if err != nil {
+			panic(fmt.Errorf("Invalid balanceId in op: %d", operation.index))
+		}
+		details["balance_id"] = balanceID
+		details["claimant"] = source.Address()
+	case xdr.OperationTypeBeginSponsoringFutureReserves,
 		xdr.OperationTypeEndSponsoringFutureReserves,
 		xdr.OperationTypeRevokeSponsorship:
 		// TBD
@@ -397,8 +404,9 @@ func (operation *transactionOperationWrapper) Participants() ([]xdr.AccountId, e
 		for _, c := range op.Body.MustCreateClaimableBalanceOp().Claimants {
 			participants = append(participants, c.MustV0().Destination)
 		}
-	case xdr.OperationTypeClaimClaimableBalance,
-		xdr.OperationTypeBeginSponsoringFutureReserves,
+	case xdr.OperationTypeClaimClaimableBalance:
+		// the only direct participant is the source_account
+	case xdr.OperationTypeBeginSponsoringFutureReserves,
 		xdr.OperationTypeEndSponsoringFutureReserves,
 		xdr.OperationTypeRevokeSponsorship:
 		// TBD
--- a/services/horizon/internal/expingest/processors/operations_processor.go
+++ b/services/horizon/internal/expingest/processors/operations_processor.go
@@ -283,8 +283,20 @@ func (operation *transactionOperationWrapper) Details() map[string]interface{} {
 	case xdr.OperationTypeBumpSequence:
 		op := operation.operation.Body.MustBumpSequenceOp()
 		details["bump_to"] = fmt.Sprintf("%d", op.BumpTo)
-	case xdr.OperationTypeCreateClaimableBalance,
-		xdr.OperationTypeClaimClaimableBalance,
+	case xdr.OperationTypeCreateClaimableBalance:
+		op := operation.operation.Body.MustCreateClaimableBalanceOp()
+		details["asset"] = op.Asset.StringCanonical()
+		details["amount"] = amount.String(op.Amount)
+		var claimants history.Claimants
+		for _, c := range op.Claimants {
+			cv0 := c.MustV0()
+			claimants = append(claimants, history.Claimant{
+				Destination: cv0.Destination.Address(),
+				Predicate:   cv0.Predicate,
+			})
+		}
+		details["claimants"] = claimants
+	case xdr.OperationTypeClaimClaimableBalance,
 		xdr.OperationTypeBeginSponsoringFutureReserves,
 		xdr.OperationTypeEndSponsoringFutureReserves,
 		xdr.OperationTypeRevokeSponsorship:
--- a/services/horizon/internal/expingest/processors/operations_processor.go
+++ b/services/horizon/internal/expingest/processors/operations_processor.go
@@ -393,8 +393,11 @@ func (operation *transactionOperationWrapper) Participants() ([]xdr.AccountId, e
 		// the only direct participant is the source_account
 	case xdr.OperationTypeBumpSequence:
 		// the only direct participant is the source_account
-	case xdr.OperationTypeCreateClaimableBalance,
-		xdr.OperationTypeClaimClaimableBalance,
+	case xdr.OperationTypeCreateClaimableBalance:
+		for _, c := range op.Body.MustCreateClaimableBalanceOp().Claimants {
+			participants = append(participants, c.MustV0().Destination)
+		}
+	case xdr.OperationTypeClaimClaimableBalance,
 		xdr.OperationTypeBeginSponsoringFutureReserves,
 		xdr.OperationTypeEndSponsoringFutureReserves,
 		xdr.OperationTypeRevokeSponsorship:
--- a/xdr/asset_test.go
+++ b/xdr/asset_test.go
@@ -104,6 +104,39 @@ var _ = Describe("xdr.Asset#String()", func() {
 	})
 })
 
+var _ = Describe("xdr.Asset#StringCanonical()", func() {
+	var asset Asset
+
+	Context("asset is native", func() {
+		BeforeEach(func() {
+			var err error
+			asset, err = NewAsset(AssetTypeAssetTypeNative, nil)
+			Expect(err).To(BeNil())
+		})
+
+		It("returns 'native'", func() {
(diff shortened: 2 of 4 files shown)


services/horizon/internal/expingest/processors/operations_processor.go
```diff
@@ -283,9 +283,28 @@ func (operation *transactionOperationWrapper) Details() map[string]interface{} {
...
+		details["amount"] = amount.String(op.Amount)
+		var claimants history.Claimants
+		for _, c := range op.Claimants {
+			cv0 := c.MustV0()
+			claimants = append(claimants, history.Claimant{
+				Destination: cv0.Destination.Address(),
+				Predicate:   cv0.Predicate,
+			})
+		}
+		details["claimants"] = claimants
```
abuiles, line 298: This will look like the following

```
 {"asset": "native", "amount": "20.0000000", "claimants": [{"predicate": "AAAAAA==", "destination": "GCFEQ7DVT2DXKFKHGFWHMCSRYMFDGUADPNF7KFPAE3C2P7YUQBMH4DG2"}, {"predicate": "AAAAAA==", "destination": "GDP5EHRW3RXYQFMGWM3AXDIO34NHU7VS3DEGZNLQRWAES2FQ23DTNL7Y"}]}
```

I'll update the JSONMarshal representation later to match the suggestion here https://github.com/stellar/go/issues/2787#issuecomment-656287363 -- for now we store the base64 encoded xdr.

services/horizon/internal/expingest/processors/operations_processor.go
```diff
@@ -283,9 +283,28 @@ func (operation *transactionOperationWrapper) Details() map[string]interface{} {
 	case xdr.OperationTypeBumpSequence:
 		op := operation.operation.Body.MustBumpSequenceOp()
 		details["bump_to"] = fmt.Sprintf("%d", op.BumpTo)
-	case xdr.OperationTypeCreateClaimableBalance,
-		xdr.OperationTypeClaimClaimableBalance,
-		xdr.OperationTypeBeginSponsoringFutureReserves,
+	case xdr.OperationTypeCreateClaimableBalance:
+		op := operation.operation.Body.MustCreateClaimableBalanceOp()
```
bartekn, line 287: I think it's worth adding `balance_id` here so it's clear what was created, it's part of `CreateClaimableBalanceResult`.

services/horizon/internal/expingest/processors/operations_processor.go
```diff
@@ -283,9 +283,28 @@ func (operation *transactionOperationWrapper) Details() map[string]interface{} {
...
+			cv0 := c.MustV0()
+			claimants = append(claimants, history.Claimant{
+				Destination: cv0.Destination.Address(),
+				Predicate:   cv0.Predicate,
+			})
+		}
+		details["claimants"] = claimants
+	case xdr.OperationTypeClaimClaimableBalance:
+		op := operation.operation.Body.MustClaimClaimableBalanceOp()
+		balanceID, err := xdr.MarshalBase64(op.BalanceId)
```
bartekn, line 301: It's hexencoded in https://github.com/stellar/go/pull/2898/files#diff-54cf664b593c8bb00b626a924a8037caR168 so maybe we should do it here too.

services/horizon/internal/expingest/processors/operations_processor.go
```diff
@@ -381,9 +400,13 @@ func (operation *transactionOperationWrapper) Participants() ([]xdr.AccountId, e
...
 		// the only direct participant is the source_account
-	case xdr.OperationTypeCreateClaimableBalance,
-		xdr.OperationTypeClaimClaimableBalance,
-		xdr.OperationTypeBeginSponsoringFutureReserves,
+	case xdr.OperationTypeCreateClaimableBalance:
+		for _, c := range op.Body.MustCreateClaimableBalanceOp().Claimants {
+			participants = append(participants, c.MustV0().Destination)
+		}
+	case xdr.OperationTypeClaimClaimableBalance:
+		// the only direct participant is the source_account
```
bartekn, line 408: Isn't the destination also a participant? I think it should be present in the list of destination's operations.

services/horizon/internal/expingest/processors/operations_processor.go
```diff
@@ -283,9 +283,28 @@ func (operation *transactionOperationWrapper) Details() map[string]interface{} {
 	case xdr.OperationTypeBumpSequence:
 		op := operation.operation.Body.MustBumpSequenceOp()
 		details["bump_to"] = fmt.Sprintf("%d", op.BumpTo)
-	case xdr.OperationTypeCreateClaimableBalance,
-		xdr.OperationTypeClaimClaimableBalance,
-		xdr.OperationTypeBeginSponsoringFutureReserves,
+	case xdr.OperationTypeCreateClaimableBalance:
+		op := operation.operation.Body.MustCreateClaimableBalanceOp()
+		details["asset"] = op.Asset.StringCanonical()
```
bartekn, line 288: I know we're transitioning to the canonical strings but all of the other operations are still using the old 3 fields (`asset_type`, `asset_code` and `asset_issuer`), I think we should add them here too (call `assetDetails` function).

xdr/asset_test.go
```diff
@@ -104,6 +105,22 @@ var _ = Describe("xdr.Asset#String()", func() {
...
+	})
+
+	Context("asset is issued", func() {
+		It("returns 'code:issuer'", func() {
+			asset := xdr.MustNewCreditAsset("USD", "GBRPYHIL2CI3FNQ4BXLFMNDLFJUNPU2HY3ZMFSHONUCEOASW7QC7OX2H")
+			Expect(asset.StringCanonical()).To(Equal("USD:GBRPYHIL2CI3FNQ4BXLFMNDLFJUNPU2HY3ZMFSHONUCEOASW7QC7OX2H"))
+		})
+	})
+})
+
```
bartekn, line 123: Let's not use `onsi` packages for new tests.

services/horizon/internal/expingest/processors/operations_processor.go
abuiles: @bartekn shouldn't that be part of the effects? That's where I was adding the `balance_id` -- I'm okay doing i there too.

abuiles: @bartekn shouldn't that be part of the effects? That's where I was adding the `balance_id` -- I'm okay doing i there too.

services/horizon/internal/expingest/processors/operations_processor.go
abuiles: The hexencoded value is only for identifying it at the API level, this is what you would use if you were to create a `ClaimClaimableBalance` operation. Maybe we should explore your idea of hexencoding the hash so that we can reconstruct the original balance id from it (and avoid confusions)

services/horizon/internal/expingest/processors/operations_processor.go
abuiles: There is no destination 

```
type ClaimClaimableBalanceOp struct {
	BalanceId ClaimableBalanceId
}
```

Stellar-Core checks against `op.SourceAccount` or `tx.SourceAccount` and if it is a valid claimant then it transfers the balance.

services/horizon/internal/expingest/processors/operations_processor.go
abuiles: sounds good

services/horizon/internal/expingest/processors/operations_processor.go
bartekn: I think you're right. Ex. we don't include `offer_id` in manage_*_offers ops.

services/horizon/internal/expingest/processors/operations_processor.go
bartekn: I think it should be easy for users to go from operation to actual balance state. So it would be great to have a consistent encoding everywhere.

abuiles: Optionally, we could have `horizon_id` which users will use to fetch the resource and rename `balance_id` to `balance_id_xdr`. 

Again, relying on the internal hashID leave us on a bad spot if a new ClaimableBalanceID is added, which is likely to happen cc @jonjove

services/horizon/internal/expingest/processors/operations_processor.go
abuiles: Perfect- added on the effects PR

services/horizon/internal/expingest/processors/operations_processor.go
abuiles: after talking offline, we'll keep the single asset attribute to start fostering the use of the canonical form.

xdr/asset_test.go
abuiles: perfect, fixed!

services/horizon/internal/expingest/processors/operations_processor.go
abuiles: @bartekn I'm merging this and I'll create an issue to revisit this specific part

jonjove: I would definitely be hesitant about referring to a `ClaimableBalanceID id` using `id.v0()`. Right now there is only a single branch of the union, but the discriminant wasn't added gratuitously. I was recalling a discussion in the context of deterministic accounts where we were debating between `sha256(txHash, opIndex)` or `sha256(txSourceAcc, seqNum, opIndex)`. This distinction is borne out, for example, in CAP-0007 vs CAP-0012. The first guarantees that a claimable balance was produced by a specific transaction, the second only that it was produced by a specific source/seq pair. I added the discriminant to handle the possibility that in the future we might want to support both of these options. So any interfaces from Horizon should probably either (a) support other discriminants out of the box or (b) have an extension plan for other discriminants.

PR #2916, opened 2020-08-13
services/horizon: Integration tests

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This commit adds a simple integration tests framework for building Horizon tests.

### Why

We need integration tests to check how Horizon services work together with each other and Stellar-Core. This is especially important for protocol upgrades where mocking is usually not enough to ensure that everything works correctly and testing different ledger backends (like captive core).

### How it works?

To start a new integration test developer should create a new test and use `IntegrationTest` struct from `test` package:
```go
func TestProtocol14Sample(t *testing.T) {
	// Prepares the environment for integration tests, see details below
	protocol14Config = test.IntegrationConfig{ProtocolVersion: 14}
	itest := test.NewIntegrationTest(t, protocol14Config)
	defer itest.Close()

	// Client() returns a horizonclient.Client instance connected to the Horizon instance under test.
	root, err := itest.Client().Root()
	assert.NoError(t, err)
	assert.Equal(t, int32(14), root.CoreSupportedProtocolVersion)
	assert.Equal(t, int32(14), root.CurrentProtocolVersion)

	// Ready to accept transactions and HTTP requests.
}
```
The test above:
* Starts a standalone Stellar network at the requested protocol version.
* Waits until Horizon starts ingesting.
* Gets a root endpoint (using `horizonclient.Client` pointed at created Horizon instance) and asserts some values.

Under the hood it:
* Creates a standalone network using `stellar/quickstart:testing` image.
* Copies a custom `horizon` binary to a docker container.
* Starts all services in a container.
* Upgrades the protocol version to the required version.
* Waits for ingestion to start.

Additionally, a new CircleCI workspace was added to make the process as easy as possible for developers. It is not started by default (`hold`) so a developer can start it whenever it's worth checking the integration tests.

To run the integration tests locally a developer:
* Needs to have `docker` installed.
* Needs to build a custom `horizon` binary using the branch they want to test. A binary must be a linux amd64 binary (`GOOS=linux GOARCH=amd64 go build -v ./services/horizon`).
* Run: `HORIZON_INTEGRATION_TESTS=true HORIZON_BIN_DIR=$(pwd) go test -v ./services/horizon/internal/integration/...`

If a special `HORIZON_INTEGRATION_TESTS` env variable is not set, the integration tests started with `test.IntegrationTest` are simply skipped so a normal `go test ./...` will not be slower because of new tests.

All the steps above are done automatically in CircleCI.

### Known limitations

* Currently tests are not run in parallel but the API is prepared to do so. Developers should use `IntegrationTest.Client()` to send requests to a Horizon instance because in the future Horizon may not be present at port 8000.


.circleci/config.yml
```diff
@@ -308,6 +328,42 @@ jobs:
...
             docker push stellar/horizon:latest
 
+  # test_horizon_integration performs Horizon integration tests, it's using
+  # decicated vm machine to be able to start arbitrary docker containers.
+  test_horizon_integration:
+    machine:
+      image: ubuntu-1604:201903-01
+    steps:
+      - run:
+          name: Create working dir
```
2opremio, line 338: Can't the cloning step be reused from other workflows?

services/horizon/internal/test/integration.go
```diff
@@ -0,0 +1,183 @@
...
+	if err != nil {
+		t.Fatal(errors.Wrap(err, "error creating docker client"))
+	}
+
+	image := "stellar/quickstart:testing"
+
+	t.Logf("Pulling %s...", image)
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
+	defer cancel()
+	reader, err := i.cli.ImagePull(ctx, "docker.io/"+image, types.ImagePullOptions{})
```
2opremio, line 68: you don't really need to pull explicitly, do you? Running for the first time should sort it out.

services/horizon/internal/test/integration.go
```diff
@@ -0,0 +1,183 @@
...
+	if err != nil {
+		t.Fatal(errors.Wrap(err, "error pulling docker image"))
+	}
+	defer reader.Close()
+	io.Copy(os.Stdout, reader)
```
2opremio, line 73: unhandled error

services/horizon/internal/test/integration.go
2opremio: ah, scratch that. It may not be the case with the API

services/horizon/internal/test/integration.go
```diff
@@ -0,0 +1,183 @@
...
+//  * HORIZON_BIN_DIR env variable set to the directory with `horizon` binary to test.
+//  * Horizon binary must be built for GOOS=linux and GOARCH=amd64.
+//
+// Skips the test if HORIZON_INTEGRATION_TESTS env variable is not set.
+func NewIntegrationTest(t *testing.T, config IntegrationConfig) *IntegrationTest {
+	if os.Getenv("HORIZON_INTEGRATION_TESTS") == "" {
+		t.Skip("skipping integration test")
+	}
+
+	if os.Getenv("HORIZON_BIN_DIR") == "" {
```
2opremio, line 51: I wonder if there is a way to avoid this (or at least fill it with the code in the current tree by default). Since we live the horizon tree, the version of horizon you should run is implicit.

Maybe you could try to compile Horizon?

services/horizon/internal/test/integration.go
```diff
@@ -0,0 +1,183 @@
...
+		if root.IngestSequence > 0 &&
+			root.HorizonSequence > 0 &&
+			root.CurrentProtocolVersion == i.config.ProtocolVersion {
+			i.t.Log("Horizon ingesting and protocol version matches...")
+			return
+		}
+		time.Sleep(time.Second)
+	}
+
+	i.t.Fatal("Horizon not ingesting...")
```
2opremio, line 160: It would help to know why. Whether there was an error getting Root, whether there was a protocol mismatch etc ...

services/horizon/internal/test/integration.go
```diff
@@ -0,0 +1,183 @@
...
+// NewIntegrationTest starts a new environment for integration test at a given
+// protocol version and blocks until Horizon starts ingesting.
+//
+// Warning: this requires:
+//  * Docker installed and all docker env variables set.
+//  * HORIZON_BIN_DIR env variable set to the directory with `horizon` binary to test.
+//  * Horizon binary must be built for GOOS=linux and GOARCH=amd64.
+//
+// Skips the test if HORIZON_INTEGRATION_TESTS env variable is not set.
+func NewIntegrationTest(t *testing.T, config IntegrationConfig) *IntegrationTest {
```
2opremio, line 46: nitpick: This function is probably better suited to be coded in shell. Have you taken a look at https://github.com/progrium/go-shell ?

2opremio (approved):

.circleci/config.yml
bartekn: I tried it but `checkout` step didn't work when `machine` is used due to some permissions issue. However, I googled a little bit and switching to `~/go` instead of `/go` fixed the issue. Thanks!

bartekn:
We've just discussed this with @tamirms and he suggested to use `stellar/horizon` and `docker-compose`. I missed the fact that `stellar/horizon` can build horizon from source so it should definitely work and is more flexible (ex. we can easily switch stellar-core versions only by changing the config instead of rebuilding `stellar/quickstart`).

Because switching to `docker-compose` requires some work and changes to `stellar/horizon` I'm going to merge what we have here to unblock testing and then, when more urgent issues are closed, we'll change the engine in a separate PR. This shouldn't change the API so existing tests should work with new engine.

Issue #2915, opened 2020-08-12
Define how to expose/save balance_id in details and effetcs


abuiles:
See this comment https://github.com/stellar/go/pull/2911#discussion_r469352048

abuiles:
This was done already - we are storing the hex encoding representation

PR #2917, opened 2020-08-13
services/horizon: Update ops processor for CAP 33

### LINKED ISSUE #2890: Update ingestion operations processor to ingest sponsorship related operations.
We need to ingest the following new operations:

- [x] `BEGIN_SPONSORING_FUTURE_RESERVES = 14`
    - [x] Update `Details()` in `transactionOperationWrapper`
    - [x] Update `Participants()` in `transactionOperationWrapper`
- [x] `END_SPONSORING_FUTURE_SPONSOR` = 15`
    - [x] Update `Details()` in `transactionOperationWrapper`
    - [x] Update `Participants()` in `transactionOperationWrapper`
- [x] `REVOKE_SPONSORSHIP = 16`
    - [x] Update `Details()` in `transactionOperationWrapper`
    - [x] Update `Participants()` in `transactionOperationWrapper`

2opremio:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

 Update the operations processor to cover the sponsoring operations in [CAP 33](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0033.md)

Fixes #2890

### Why

We need to implement CAP 33

### Known limitations

N/A

### DIFF
--- a/services/horizon/internal/expingest/processors/operations_processor.go
+++ b/services/horizon/internal/expingest/processors/operations_processor.go
@@ -330,7 +330,7 @@ func (operation *transactionOperationWrapper) Details() map[string]interface{} {
 		details["sponsored_id"] = op.SponsoredId.Address()
 	case xdr.OperationTypeEndSponsoringFutureReserves:
 		operations := operation.transaction.Envelope.Operations()
-		for i := operation.index; i >= 0; i-- {
+		for i := int(operation.index); i >= 0; i-- {
 			if operations[i].Body.Type == xdr.OperationTypeBeginSponsoringFutureReserves {
 				// We intentionally ignore `RevokeSponsorship` operations inside or outside
 				// the sandwich
--- a/services/horizon/internal/expingest/processors/operations_processor.go
+++ b/services/horizon/internal/expingest/processors/operations_processor.go
@@ -132,6 +132,16 @@ func (operation *transactionOperationWrapper) OperationResult() *xdr.OperationRe
 	return &tr
 }
 
+func (operation *transactionOperationWrapper) findPriorBeginSponsoringOp() *xdr.Operation {
+	operations := operation.transaction.Envelope.Operations()
+	for i := int(operation.index); i >= 0; i-- {
+		if operations[i].Body.Type == xdr.OperationTypeBeginSponsoringFutureReserves {
+			return &operations[i]
+		}
+	}
+	return nil
+}
+
 // Details returns the operation details as a map which can be stored as JSON.
 func (operation *transactionOperationWrapper) Details() map[string]interface{} {
 	details := map[string]interface{}{}
@@ -328,14 +338,9 @@ func (operation *transactionOperationWrapper) Details() map[string]interface{} {
 		op := operation.operation.Body.MustBeginSponsoringFutureReservesOp()
 		details["sponsored_id"] = op.SponsoredId.Address()
 	case xdr.OperationTypeEndSponsoringFutureReserves:
-		operations := operation.transaction.Envelope.Operations()
-		for i := int(operation.index); i >= 0; i-- {
-			if operations[i].Body.Type == xdr.OperationTypeBeginSponsoringFutureReserves {
-				// We intentionally ignore `RevokeSponsorship` operations inside or outside
-				// the sandwich
-				begin := operations[i].Body.MustBeginSponsoringFutureReservesOp()
-				details["begin_sponsored_id"] = begin.SponsoredId.Address()
-			}
+		if beginSponsoringOp := operation.findPriorBeginSponsoringOp(); beginSponsoringOp != nil {
+			beginSponsor := beginSponsoringOp.SourceAccount.ToAccountId()
+			details["begin_sponsor"] = beginSponsor.Address()
 		}
 	case xdr.OperationTypeRevokeSponsorship:
 		op := operation.operation.Body.MustRevokeSponsorshipOp()
@@ -466,7 +471,10 @@ func (operation *transactionOperationWrapper) Participants() ([]xdr.AccountId, e
 	case xdr.OperationTypeBeginSponsoringFutureReserves:
 		participants = append(participants, op.Body.MustBeginSponsoringFutureReservesOp().SponsoredId)
 	case xdr.OperationTypeEndSponsoringFutureReserves:
-		// the only direct participant is the source_account
+		if beginSponsoringOp := operation.findPriorBeginSponsoringOp(); beginSponsoringOp != nil {
+			beginSponsor := beginSponsoringOp.SourceAccount.ToAccountId()
+			participants = append(participants, beginSponsor)
+		}
 	case xdr.OperationTypeRevokeSponsorship:
 		// the only direct participant is the source_account
 	default:
--- a/services/horizon/internal/expingest/processors/operations_processor.go
+++ b/services/horizon/internal/expingest/processors/operations_processor.go
@@ -38,8 +38,12 @@ func (p *OperationProcessor) ProcessTransaction(transaction io.LedgerTransaction
 			operation:      op,
 			ledgerSequence: p.sequence,
 		}
+		details, err := operation.Details()
+		if err != nil {
+			return errors.Wrapf(err, "Error obtaining details for operation %v", operation.ID())
+		}
 		var detailsJSON []byte
-		detailsJSON, err = json.Marshal(operation.Details())
+		detailsJSON, err = json.Marshal(details)
 		if err != nil {
 			return errors.Wrapf(err, "Error marshaling details for operation %v", operation.ID())
 		}
@@ -134,7 +138,7 @@ func (operation *transactionOperationWrapper) OperationResult() *xdr.OperationRe
 
 func (operation *transactionOperationWrapper) findPriorBeginSponsoringOp() *xdr.Operation {
 	operations := operation.transaction.Envelope.Operations()
-	for i := int(operation.index); i >= 0; i-- {
+	for i := int(operation.index) - 1; i >= 0; i-- {
 		if operations[i].Body.Type == xdr.OperationTypeBeginSponsoringFutureReserves {
 			return &operations[i]
 		}
@@ -143,7 +147,7 @@ func (operation *transactionOperationWrapper) findPriorBeginSponsoringOp() *xdr.
 }
 
 // Details returns the operation details as a map which can be stored as JSON.
-func (operation *transactionOperationWrapper) Details() map[string]interface{} {
+func (operation *transactionOperationWrapper) Details() (map[string]interface{}, error) {
 	details := map[string]interface{}{}
 	source := operation.SourceAccount()
 	// FIXME: we shouldn't ignore this error
@@ -338,15 +342,19 @@ func (operation *transactionOperationWrapper) Details() map[string]interface{} {
 		op := operation.operation.Body.MustBeginSponsoringFutureReservesOp()
 		details["sponsored_id"] = op.SponsoredId.Address()
 	case xdr.OperationTypeEndSponsoringFutureReserves:
-		if beginSponsoringOp := operation.findPriorBeginSponsoringOp(); beginSponsoringOp != nil {
-			beginSponsor := beginSponsoringOp.SourceAccount.ToAccountId()
-			details["begin_sponsor"] = beginSponsor.Address()
+		beginSponsoringOp := operation.findPriorBeginSponsoringOp()
+		if beginSponsoringOp == nil {
+			return nil, fmt.Errorf("prior Begin operation not found for EndSponsoringFutureReserves")
 		}
+		beginSponsor := beginSponsoringOp.SourceAccount.ToAccountId()
+		details["begin_sponsor"] = beginSponsor.Address()
 	case xdr.OperationTypeRevokeSponsorship:
 		op := operation.operation.Body.MustRevokeSponsorshipOp()
 		switch op.Type {
 		case xdr.RevokeSponsorshipTypeRevokeSponsorshipLedgerEntry:
-			ledgerKeyDetails(details, *op.LedgerKey, "")
+			if err := ledgerKeyDetails(details, *op.LedgerKey, ""); err != nil {
+				return nil, err
+			}
 		case xdr.RevokeSponsorshipTypeRevokeSponsorshipSigner:
 			details["signer_account"] = op.Signer.AccountId.Address()
 			details["signer_key"] = op.Signer.SignerKey.Address()
@@ -355,7 +363,7 @@ func (operation *transactionOperationWrapper) Details() map[string]interface{} {
 		panic(fmt.Errorf("Unknown operation type: %s", operation.OperationType()))
 	}
 
-	return details
+	return details, nil
 }
 
 // assetDetails sets the details for `a` on `result` using keys with `prefix`
@@ -407,21 +415,26 @@ func operationFlagDetails(result map[string]interface{}, f int32, prefix string)
 	result[prefix+"_flags_s"] = s
 }
 
-func ledgerKeyDetails(result map[string]interface{}, ledgerKey xdr.LedgerKey, prefix string) {
+func ledgerKeyDetails(result map[string]interface{}, ledgerKey xdr.LedgerKey, prefix string) error {
 	switch ledgerKey.Type {
 	case xdr.LedgerEntryTypeAccount:
 		result[prefix+"account_id"] = ledgerKey.Account.AccountId.Address()
 	case xdr.LedgerEntryTypeClaimableBalance:
-		result[prefix+"claimable_balance_id"] = ledgerKey.ClaimableBalance.BalanceId
+		b64, err := xdr.MarshalBase64(ledgerKey.ClaimableBalance.BalanceId)
+		if err != nil {
+			return errors.Wrapf(err, "in claimable balance")
+		}
+		result[prefix+"claimable_balance_id"] = b64
 	case xdr.LedgerEntryTypeData:
 		result[prefix+"account_id"] = ledgerKey.Data.AccountId.Address()
 		result[prefix+"data_name"] = ledgerKey.Data.DataName
 	case xdr.LedgerEntryTypeOffer:
 		result[prefix+"offer_id"] = ledgerKey.Offer.OfferId
 	case xdr.LedgerEntryTypeTrustline:
-		result[prefix+"account_id"] = ledgerKey.TrustLine.AccountId
+		result[prefix+"account_id"] = ledgerKey.TrustLine.AccountId.Address()
 		assetDetails(result, ledgerKey.TrustLine.Asset, prefix)
 	}
+	return nil
 }
 
 // Participants returns the accounts taking part in the operation.
@@ -471,10 +484,12 @@ func (operation *transactionOperationWrapper) Participants() ([]xdr.AccountId, e
 	case xdr.OperationTypeBeginSponsoringFutureReserves:
 		participants = append(participants, op.Body.MustBeginSponsoringFutureReservesOp().SponsoredId)
 	case xdr.OperationTypeEndSponsoringFutureReserves:
-		if beginSponsoringOp := operation.findPriorBeginSponsoringOp(); beginSponsoringOp != nil {
-			beginSponsor := beginSponsoringOp.SourceAccount.ToAccountId()
-			participants = append(participants, beginSponsor)
+		beginSponsoringOp := operation.findPriorBeginSponsoringOp()
+		if beginSponsoringOp == nil {
+			return nil, fmt.Errorf("prior Begin operation not found for EndSponsoringFutureReserves")
 		}
+		beginSponsor := beginSponsoringOp.SourceAccount.ToAccountId()
+		participants = append(participants, beginSponsor)
 	case xdr.OperationTypeRevokeSponsorship:
 		// the only direct participant is the source_account
 	default:
--- a/services/horizon/internal/expingest/processors/operations_processor.go
+++ b/services/horizon/internal/expingest/processors/operations_processor.go
@@ -108,7 +108,6 @@ func (operation *transactionOperationWrapper) OperationType() xdr.OperationType 
 	return operation.operation.Body.Type
 }
 
-// OperationResult returns the operation's result record
 func (operation *transactionOperationWrapper) getSponsor() (*xdr.AccountId, error) {
 	changes, err := operation.transaction.GetOperationChanges(operation.index)
 	if err != nil {
--- a/services/horizon/internal/expingest/processors/operations_processor.go
+++ b/services/horizon/internal/expingest/processors/operations_processor.go
@@ -304,10 +304,20 @@ func (operation *transactionOperationWrapper) Details() map[string]interface{} {
 		}
 		details["balance_id"] = balanceID
 		details["claimant"] = source.Address()
-	case xdr.OperationTypeBeginSponsoringFutureReserves,
-		xdr.OperationTypeEndSponsoringFutureReserves,
-		xdr.OperationTypeRevokeSponsorship:
-		// TBD
+	case xdr.OperationTypeBeginSponsoringFutureReserves:
+		op := operation.operation.Body.MustBeginSponsoringFutureReservesOp()
+		details["sponsored_id"] = op.SponsoredId.Address()
+	case xdr.OperationTypeEndSponsoringFutureReserves:
+		// NOP
+	case xdr.OperationTypeRevokeSponsorship:
+		op := operation.operation.Body.MustRevokeSponsorshipOp()
+		switch op.Type {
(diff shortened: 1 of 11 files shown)


services/horizon/internal/expingest/processors/operations_processor.go
```diff
@@ -304,10 +304,20 @@ func (operation *transactionOperationWrapper) Details() map[string]interface{} {
...
+	case xdr.OperationTypeBeginSponsoringFutureReserves:
+		op := operation.operation.Body.MustBeginSponsoringFutureReservesOp()
+		details["sponsored_id"] = op.SponsoredId.Address()
+	case xdr.OperationTypeEndSponsoringFutureReserves:
+		// NOP
+	case xdr.OperationTypeRevokeSponsorship:
+		op := operation.operation.Body.MustRevokeSponsorshipOp()
+		switch op.Type {
+		case xdr.RevokeSponsorshipTypeRevokeSponsorshipLedgerEntry:
+			ledgerKeyDetails(details, *op.LedgerKey, "")
```
2opremio, line 316: Is it overkill to include all the ledger key details? Maybe base64 is enough?

services/horizon/internal/expingest/processors/operations_processor.go
```diff
@@ -406,10 +438,12 @@ func (operation *transactionOperationWrapper) Participants() ([]xdr.AccountId, e
...
-	case xdr.OperationTypeBeginSponsoringFutureReserves,
-		xdr.OperationTypeEndSponsoringFutureReserves,
-		xdr.OperationTypeRevokeSponsorship:
-		// TBD
+	case xdr.OperationTypeBeginSponsoringFutureReserves:
+		participants = append(participants, op.Body.MustBeginSponsoringFutureReservesOp().SponsoredId)
+	case xdr.OperationTypeEndSponsoringFutureReserves:
+		// the only direct participant is the source_account
+	case xdr.OperationTypeRevokeSponsorship:
+		// the only direct participant is the source_account
```
2opremio, line 446: I am unsure of whether this operations should include the account(s) which is affected by the revocation/end of sponsoring

2opremio:
I have intentionally omitted tests until the questions above are addressed

services/horizon/internal/expingest/processors/operations_processor.go
abuiles: @2opremio no, that's great -- the idea is that people should be able to tell what happened/changed by looking at the operation and the base64 value doesn't reflects that (unless you decode it to an XDR) - I just have some comments on the format below

services/horizon/internal/expingest/processors/operations_processor.go
```diff
@@ -364,6 +374,28 @@ func operationFlagDetails(result map[string]interface{}, f int32, prefix string)
 	result[prefix+"_flags_s"] = s
 }
 
+func ledgerKeyDetails(result map[string]interface{}, ledgerKey xdr.LedgerKey, prefix string) {
+	switch ledgerKey.Type {
+	case xdr.LedgerEntryTypeAccount:
+		result[prefix+"ledger_key_account"] = ledgerKey.Account.AccountId.Address()
```
abuiles, line 380: I don't think we need to expose "ledger_key" but rather `account_id`

services/horizon/internal/expingest/processors/operations_processor.go
```diff
@@ -364,6 +374,28 @@ func operationFlagDetails(result map[string]interface{}, f int32, prefix string)
...
+	case xdr.LedgerEntryTypeClaimableBalance:
+		result[prefix+"ledger_key_claimable_balance"] = ledgerKey.ClaimableBalance.BalanceId
```
abuiles, line 382: `claimable_balance_id` 

We also need to resolve https://github.com/stellar/go/issues/2915

services/horizon/internal/expingest/processors/operations_processor.go
```diff
@@ -364,6 +374,28 @@ func operationFlagDetails(result map[string]interface{}, f int32, prefix string)
...
+	case xdr.LedgerEntryTypeData:
+		result[prefix+"ledger_key_data"] = map[string]interface{}{
```
abuiles, line 384: `account_id` and `data_name` without the `ledger_key_data`

services/horizon/internal/expingest/processors/operations_processor.go
```diff
@@ -364,6 +374,28 @@ func operationFlagDetails(result map[string]interface{}, f int32, prefix string)
...
+			"data_name":  ledgerKey.Data.DataName,
+		}
+	case xdr.LedgerEntryTypeOffer:
+		result[prefix+"ledger_key_offer_id"] = ledgerKey.Offer.OfferId
+	case xdr.LedgerEntryTypeTrustline:
+		trustlineDetails := map[string]interface{}{
+			"account_id": ledgerKey.TrustLine.AccountId,
+		}
+		assetDetails(trustlineDetails, ledgerKey.TrustLine.Asset, "")
+		result[prefix+"ledger_key_trustline"] = trustlineDetails
```
abuiles, line 395: same comment as above

services/horizon/internal/expingest/processors/operations_processor.go
abuiles: maybe we should add a key called `entry_type` which could be `account`, `offer`, `account_data` .... `claimable_balance`

services/horizon/internal/expingest/processors/operations_processor.go
```diff
@@ -406,10 +438,12 @@ func (operation *transactionOperationWrapper) Participants() ([]xdr.AccountId, e
...
 	case xdr.OperationTypeClaimClaimableBalance:
 		// the only direct participant is the source_account
-	case xdr.OperationTypeBeginSponsoringFutureReserves,
-		xdr.OperationTypeEndSponsoringFutureReserves,
-		xdr.OperationTypeRevokeSponsorship:
-		// TBD
+	case xdr.OperationTypeBeginSponsoringFutureReserves:
+		participants = append(participants, op.Body.MustBeginSponsoringFutureReservesOp().SponsoredId)
+	case xdr.OperationTypeEndSponsoringFutureReserves:
+		// the only direct participant is the source_account
```
abuiles, line 444: I think it would make sense to add the account we are sponsoring as a participant here too so when looking at the operations list (from the sponsored account perspective), you can see the whole sandwich (begin and end)

services/horizon/internal/expingest/processors/operations_processor.go
abuiles: in this scenario, since the sponsored account isn't required to sign the tx, I'd say it wasn't a participant in the operation.

services/horizon/internal/expingest/processors/operations_processor.go
```diff
@@ -304,10 +304,20 @@ func (operation *transactionOperationWrapper) Details() map[string]interface{} {
 		}
 		details["balance_id"] = balanceID
 		details["claimant"] = source.Address()
-	case xdr.OperationTypeBeginSponsoringFutureReserves,
-		xdr.OperationTypeEndSponsoringFutureReserves,
-		xdr.OperationTypeRevokeSponsorship:
-		// TBD
```
abuiles, line 310: I think we also need to update the operations that create a new ledger entry, and if they are being sponsored, then add the "sponsor" attribute to the list of details

services/horizon/internal/expingest/processors/operations_processor.go
```diff
@@ -406,10 +438,12 @@ func (operation *transactionOperationWrapper) Participants() ([]xdr.AccountId, e
 		}
 	case xdr.OperationTypeClaimClaimableBalance:
 		// the only direct participant is the source_account
-	case xdr.OperationTypeBeginSponsoringFutureReserves,
```
abuiles, line 409: We also need to update the operations that create a new ledger entry, and if they are being sponsored, then add the "sponsor" as one of the participants in the operation.

services/horizon/internal/expingest/processors/operations_processor.go
2opremio: But then we would need to nest it into a sub-dictionary (with `ledger_key` key) for context. Are you OK with that?

services/horizon/internal/expingest/processors/operations_processor.go
2opremio: Will do

services/horizon/internal/expingest/processors/operations_processor.go
2opremio: Then (for consistency) we should also add a similar entry to distinguish `RevokeSponsorshipTypeRevokeSponsorshipLedgerEntry` and `RevokeSponsorshipTypeRevokeSponsorshipSigner`. Are you OK with that?

services/horizon/internal/expingest/processors/operations_processor.go
abuiles: why do we need the ledger_key part?

services/horizon/internal/expingest/processors/operations_processor.go
2opremio: How can I retrieve that account from this context? (AFAICT the operation is empty)

services/horizon/internal/expingest/processors/operations_processor.go
abuiles: oh yeah, `account_signer` is one of them :)

services/horizon/internal/expingest/processors/operations_processor.go
jonjove: The operation is empty. The account is implicitly the source account of the corresponding operation which began the sandwich. The begin sponsoring future reserves operation fails if the `sponsoredID` is already sponsored or if such sponsorship would form a chain (eg. A is-sponsoring-future-reserves-for B which is-sponsoring-future-reserves-for C).

services/horizon/internal/expingest/processors/operations_processor.go
2opremio: To denote that the account_id is part of a ledger key

services/horizon/internal/expingest/processors/operations_processor.go
2opremio: @abuiles then it's probably tricky to gather the the account from this context. But, as @jonjove says, it's implicit so that may not be a bad thing.

@abuiles are you OK with leaving it as is?

services/horizon/internal/expingest/processors/operations_processor.go
2opremio: Will do

services/horizon/internal/expingest/processors/operations_processor.go
abuiles: we don't expose `ledger_keys` at the Horizon API levels, using `account_id` without nesting it should be okay.

services/horizon/internal/expingest/processors/operations_processor.go
2opremio: How can I access the relevant Ledger Entry? Even if I retrieve the ledger from the ledger sequence (which would require some modifications in the operations processor) it's not clear for me how to do it.

Sorry if the question is stupid, I am struggling to understand the role of ledger entries and bucket entries. From what I see, they are only accessible from history archives. Shouldn't we store the sponsorship status of a transaction in `io.LedgerTransaction` during ingestion?

services/horizon/internal/expingest/processors/operations_processor.go
abuiles: > @abuiles then it's probably tricky to gather the the account from this context

I think we can add a helper to `transactionOperationWrapper` to find the matching `being_sponsoring_operation` if the current operation is `end_sponsoring`

The attribute `index` in  `transactionOperationWrapper` tells us the current operation position inside the operation set and we have access to all the operations via `transaction io.LedgerTransaction`

The idea about adding the account being sponsored as a participant is to allow the end user to know where the sponsorship op began and ended - if we don't add the sponsored operation as a participant, then it won't show up in the operations end-point for the account being sponsored -- ideally when you are looking at `/accounts/G..../operations` you should be able to see something like 
 `{begin_sponsoring,  op1, op2, ....,  end_sponsoring}`  

We can explore other options, it boils down to the type of questions users need to answer when going through horizon.

services/horizon/internal/expingest/processors/operations_processor.go
abuiles: We can get the relevant `LedgerEntry using `LedgerTransaction.GetOperationChanges` - we do something similar when processing effects, checkout https://github.com/stellar/go/blob/77208910d8e4b9311f6696dda89d21352d7c7475/services/horizon/internal/expingest/processors/effects_processor.go#L619-L631

>Shouldn't we store the sponsorship status of a transaction in io.LedgerTransaction during ingestion?

I think we already do that when we ingest state. This is different though, adding the sponsor here help us answer historical data, for example, if there is a create account operation and the balance was 0, I'd like to know a. who created this account, b. who was sponsoring the ledger entry at that time (if we only rely on state data, sponsor might be different between the ledger where it was created and now).

abuiles: @2opremio now that I think about it, we don't need to add the `sponsor` entry to those operations though - they are not part of the operation body so they don't belong there. I think this should be reflects on the effects.

bartekn: Yes, I agree that we probably need to add sponsorship info to effects (like `account_sponsored` effect?). We try to keep operations response structure as close to XDR as possible (ex. that's why there's no amount on `account_merge`).

abuiles:
Bartek, what do you think about adding the account being sponsored as a
participant on the `end_sponsoring_op`?

services/horizon/internal/expingest/processors/operations_processor.go
2opremio: Ah, crap, I have already implemented it. But ... shouldn't we do this for the participants anyways?

2opremio: Also, there is no effect planned for this at https://docs.google.com/document/d/1Lor5OeOzPeuCID3-HSreqK-psjCvOXYfg22pwrUEUZA

services/horizon/internal/expingest/processors/operations_processor.go
2opremio: How does that play with `UpdateSponsorshipOp`, I guess we should be checking for that as well in case the sponsorship has changed half way.

Also, should we include all the accounts in the sponsorship chain? The initial sponsorship and all the `UpdateSponsorshipOp` in the sandwich? This starts to get hairy.

abuiles: Yeah `RevokeSponsorshipOp` sponsorship is interesting  because:

1. You could call `RevokeSponsorshipOp` without a sandwich
2. You could call `RevokeSponsorshipOp` inside a sandwich, in that scenario I guess we'll need to create a sponsorship removed and sponsorship created effect

services/horizon/internal/expingest/processors/operations_processor.go
abuiles: @2opremio we have created, updated, removed affect for sponsorships -- so we can play with a combination of those.

services/horizon/internal/expingest/processors/operations_processor.go
jonjove: Since I see you discussing `RevokeSponsorshipOp` here I figured I would share one of my learnings from the implementation. Using revoke sponsorship to transfer a sponsorship (eg. when used in a sandwich) isn't exactly the same as removing then creating a sponsorship. The reason is that the sponsored account may not actually have sufficient balance for the sponsorship to be remove (consider the case of a sponsored account created with 0 balance, which then has the sponsorship transferred). If you show it as remove then create, it will look like the account entered an intermediate state in which it couldn't afford its reserve. We had to handle this specifically in stellar-core. I don't know if things like this matter on your end.

2opremio:
@abuiles friendly ping :)

services/horizon/internal/expingest/processors/operations_processor.go
```diff
@@ -108,18 +108,48 @@ func (operation *transactionOperationWrapper) OperationType() xdr.OperationType
...
 // OperationResult returns the operation's result record
 func (operation *transactionOperationWrapper) OperationResult() *xdr.OperationResultTr {
 	results, _ := operation.transaction.Result.OperationResults()
 	tr := results[operation.index].MustTr()
 	return &tr
 }
 
+func (operation *transactionOperationWrapper) findPriorBeginSponsoringOp() *xdr.Operation {
+	operations := operation.transaction.Envelope.Operations()
+	for i := int(operation.index); i >= 0; i-- {
```
abuiles, line 137: shall we start at operation.index-1?

services/horizon/internal/expingest/processors/operations_processor.go
```diff
@@ -304,10 +334,23 @@ func (operation *transactionOperationWrapper) Details() map[string]interface{} {
...
 		details["claimant"] = source.Address()
-	case xdr.OperationTypeBeginSponsoringFutureReserves,
-		xdr.OperationTypeEndSponsoringFutureReserves,
-		xdr.OperationTypeRevokeSponsorship:
-		// TBD
+	case xdr.OperationTypeBeginSponsoringFutureReserves:
+		op := operation.operation.Body.MustBeginSponsoringFutureReservesOp()
+		details["sponsored_id"] = op.SponsoredId.Address()
+	case xdr.OperationTypeEndSponsoringFutureReserves:
+		if beginSponsoringOp := operation.findPriorBeginSponsoringOp(); beginSponsoringOp != nil {
```
abuiles, line 341: @2opremio if we can't find the matching begin sponsoring operation we should fail since this is likely a bug somewhere. The semantics of the operation guarantee us there is a begin and end.

services/horizon/internal/expingest/processors/operations_processor.go
```diff
@@ -364,11 +407,30 @@ func operationFlagDetails(result map[string]interface{}, f int32, prefix string)
 	result[prefix+"_flags_s"] = s
 }
 
+func ledgerKeyDetails(result map[string]interface{}, ledgerKey xdr.LedgerKey, prefix string) {
+	switch ledgerKey.Type {
+	case xdr.LedgerEntryTypeAccount:
+		result[prefix+"account_id"] = ledgerKey.Account.AccountId.Address()
+	case xdr.LedgerEntryTypeClaimableBalance:
+		result[prefix+"claimable_balance_id"] = ledgerKey.ClaimableBalance.BalanceId
```
abuiles, line 415: can you store this as base64?

services/horizon/internal/expingest/processors/operations_processor.go
```diff
@@ -364,11 +407,30 @@ func operationFlagDetails(result map[string]interface{}, f int32, prefix string)
...
+	case xdr.LedgerEntryTypeData:
+		result[prefix+"account_id"] = ledgerKey.Data.AccountId.Address()
+		result[prefix+"data_name"] = ledgerKey.Data.DataName
+	case xdr.LedgerEntryTypeOffer:
+		result[prefix+"offer_id"] = ledgerKey.Offer.OfferId
+	case xdr.LedgerEntryTypeTrustline:
+		result[prefix+"account_id"] = ledgerKey.TrustLine.AccountId
```
abuiles, line 422: Call `.Address()` at the end.

services/horizon/internal/expingest/processors/operations_processor.go
```diff
@@ -406,10 +468,15 @@ func (operation *transactionOperationWrapper) Participants() ([]xdr.AccountId, e
...
 	case xdr.OperationTypeClaimClaimableBalance:
 		// the only direct participant is the source_account
-	case xdr.OperationTypeBeginSponsoringFutureReserves,
-		xdr.OperationTypeEndSponsoringFutureReserves,
-		xdr.OperationTypeRevokeSponsorship:
-		// TBD
+	case xdr.OperationTypeBeginSponsoringFutureReserves:
+		participants = append(participants, op.Body.MustBeginSponsoringFutureReservesOp().SponsoredId)
+	case xdr.OperationTypeEndSponsoringFutureReserves:
+		if beginSponsoringOp := operation.findPriorBeginSponsoringOp(); beginSponsoringOp != nil {
```
abuiles, line 474: fail if being not found

services/horizon/internal/expingest/processors/operations_processor.go
2opremio: sure

services/horizon/internal/expingest/processors/operations_processor.go
2opremio: Ah, true, thanks

services/horizon/internal/expingest/processors/operations_processor.go
2opremio: Alright, will do.

services/horizon/internal/expingest/processors/operations_processor.go
2opremio: sure

2opremio:
@abuiles PTAL

services/horizon/internal/expingest/processors/operations_processor.go
```diff
@@ -108,18 +112,48 @@ func (operation *transactionOperationWrapper) OperationType() xdr.OperationType
...
 }
 
+func (operation *transactionOperationWrapper) getSponsor() (*xdr.AccountId, error) {
+	changes, err := operation.transaction.GetOperationChanges(operation.index)
+	if err != nil {
+		return nil, err
+	}
+	for _, c := range changes {
+		if c.Pre != nil || c.Post == nil {
+			// We are looking for entry creations
```
abuiles, line 122: @2opremio is this only because we care in this specific scenario about who the sponsor is going to be for the new ledger entry being created?  Is the assumption that we'll  signal a sponsorship update via effects?

services/horizon/internal/expingest/processors/operations_processor.go
```diff
@@ -304,15 +338,32 @@ func (operation *transactionOperationWrapper) Details() map[string]interface{} {
...
-		xdr.OperationTypeEndSponsoringFutureReserves,
-		xdr.OperationTypeRevokeSponsorship:
-		// TBD
+	case xdr.OperationTypeBeginSponsoringFutureReserves:
+		op := operation.operation.Body.MustBeginSponsoringFutureReservesOp()
+		details["sponsored_id"] = op.SponsoredId.Address()
+	case xdr.OperationTypeEndSponsoringFutureReserves:
+		beginSponsoringOp := operation.findPriorBeginSponsoringOp()
+		if beginSponsoringOp == nil {
+			return nil, fmt.Errorf("prior Begin operation not found for EndSponsoringFutureReserves")
```
abuiles, line 347: @2opremio can you add something that help us identify the operation? Maybe its ID?

services/horizon/internal/expingest/processors/operations_processor.go
2opremio: Yes. I created this function for this PR, I may consider changing it/moving it it for the effects issue later on.

2opremio:
Coming up! I was waiting for a GA for the rest

services/horizon/internal/expingest/processors/operations_processor.go
2opremio: The caller adds the ID on error

2opremio:
@abuiles PTAL

2opremio:
@abuiles Actually, I just noticed I am missing tests for `OperationTypeRevokeSponsorship`, but please check the rest.

2opremio:
@abuiles It should be completely done now

2opremio:
@abuiles everything has been addressed, PTAL

services/horizon/internal/expingest/processors/operations_processor.go
```diff
@@ -364,11 +420,32 @@ func operationFlagDetails(result map[string]interface{}, f int32, prefix string)
 	result[prefix+"_flags_s"] = s
 }
 
+func ledgerKeyDetails(result map[string]interface{}, ledgerKey xdr.LedgerKey, prefix string) error {
```
tamirms, line 423: the function name does not give the impression that it will actually mutate the map argument. how about calling it `setLedgerKeyDetails()`? I suppose  `assetDetails()` should also be renamed to `setAssetDetails()`

tamirms (approved):
I had one minor comment. Other than that it looks good to me!

services/horizon/internal/expingest/processors/operations_processor.go
2opremio: I totally agree (and we have many cases like that in other places). I didn't change it for consistency with `assetDetails`

I will fix both

Issue #2882, opened 2020-08-04
Update state verifier with CAP23 claimable_balances


abuiles:
Depends on #2881

abuiles:
fixed in #2923

PR #2923, opened 2020-08-17
horizon/internal/ingest:Add Claimable Balances to state verifier.

abuiles:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add Claimable Balance entries to state verifier.

### Why

We have a new ledger entry type and we need to check that we have the correct state.

Fixes https://github.com/stellar/go/issues/2882

### DIFF
--- a/services/horizon/internal/expingest/verify.go
+++ b/services/horizon/internal/expingest/verify.go
@@ -145,6 +145,7 @@ func (s *system) verifyState(verifyAgainstLatestCheckpoint bool) error {
 		data := make([]xdr.LedgerKeyData, 0, verifyBatchSize)
 		offers := make([]int64, 0, verifyBatchSize)
 		trustLines := make([]xdr.LedgerKeyTrustLine, 0, verifyBatchSize)
+		cBalances := make([]xdr.ClaimableBalanceId, 0, verifyBatchSize)
 		for _, key := range keys {
 			switch key.Type {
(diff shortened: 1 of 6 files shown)


services/horizon/internal/expingest/verify.go
```diff
@@ -491,6 +504,53 @@ func addTrustLinesToStateVerifier(
...
+		claimants := []xdr.Claimant{}
+		for _, claimant := range row.Claimants {
+			claimants = append(claimants, xdr.Claimant{
+				Type: xdr.ClaimantTypeClaimantTypeV0,
+				V0: &xdr.ClaimantV0{
+					Destination: xdr.MustAddress(claimant.Destination),
+					Predicate:   claimant.Predicate,
+				},
+			})
+		}
```
bartekn, line 532: I'm not sure but worth checking if the order of claimains in DB is the same as in buckets. See how we sort signers in `transformEntry`.

abuiles: oh good point - I'll sort them just to make sure we don't depend on the order

abuiles: @bartekn handled here https://github.com/abuiles/go/commit/1246131934309a1e0bceebc8e9961230f9fd8743

PR #2924, opened 2020-08-17
services/horizon: Add /claimable_balances/:id.

abuiles:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Introduce end-point to fetch a single claimable balance by ID `/claimable_balances/id.`

### Why

User should be able to fetch claimable balances from Horizon using the its ID.

### DIFF
--- a/services/horizon/internal/resourceadapter/claimable_balances.go
+++ b/services/horizon/internal/resourceadapter/claimable_balances.go
@@ -1,0 +1,49 @@
+package resourceadapter
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/stellar/go/amount"
+	protocol "github.com/stellar/go/protocols/horizon"
+	horizonContext "github.com/stellar/go/services/horizon/internal/context"
+	"github.com/stellar/go/services/horizon/internal/db2/history"
+	"github.com/stellar/go/support/errors"
+	"github.com/stellar/go/support/render/hal"
+	"github.com/stellar/go/xdr"
+)
+
+// PopulateClaimableBalance fills out the resource's fields
+func PopulateClaimableBalance(
+	ctx context.Context,
+	dest *protocol.ClaimableBalance,
+	claimableBalance history.ClaimableBalance,
+) error {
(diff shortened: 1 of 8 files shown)


services/horizon/internal/resourceadapter/claimable_balances.go
```diff
@@ -0,0 +1,49 @@
...
+			errors.Wrap(err, "failed to encode predicate to base64")
+		}
+		dest.Claimants[i].Destination = c.Destination
+		dest.Claimants[i].Predicate = predicate
+	}
+
+	lb := hal.LinkBuilder{Base: horizonContext.BaseURL(ctx)}
+	self := fmt.Sprintf("/claimable_balances/%s", claimableBalance.ID)
+	dest.Links.Self = lb.Link(self)
+	dest.PT = fmt.Sprintf("%d-%s", claimableBalance.LastModifiedLedger, claimableBalance.ID)
```
abuiles, line 47: This will be used once we have the index route, using LastModifiedLedger + ID gives us a deterministic view (soft of) -- expect for the scenarios where the sponsor gets updated for an entry. Still this is better rather than relying on only on the ID.

abuiles: I have a method in a different PR which builds this value.

bartekn (approved):
LGTM! I wonder if we should rather use mocked DB in actions tests instead of real Postgres server. Added to Horizon Team meeting.

abuiles:
> LGTM! I wonder if we should rather use mocked DB in actions tests instead of real Postgres server. Added to Horizon Team meeting.

I agree there is some value on mocking the DB, but thanks to using the DB I just found a scenario where I have to tweak a query in a specific way so the query planner would do the right thing. 

If we were using a mocked DB this would probably surface all  the way to the integration tests and maybe since they require more work, we would have probably missed it.

I think we are still at a point where the benefit outweighs the cost.

PR #2931, opened 2020-08-19
services/horizon/internal/db2/history: Remove hash id from claimable_balances schema

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Close https://github.com/stellar/go/issues/2915

Previously we added an id column to the claimable_balances table which consisted of the sha hash of the xdr.ClaimableBalanceId binary. It turns out that this id column is unnecessary because we can query the table directly using xdr.ClaimableBalanceId values.

When defining the API to query claimable balances we can have the client provide the hex encoding of  the xdr.ClaimableBalanceId value which should only be 72 characters long. For reference, the sha hash of the xdr value is 64 characters long.

In this PR I have also changed the type of var char columns to text because text performs more or less the same as var char but is more flexible: https://stackoverflow.com/questions/4848964/postgresql-difference-between-text-and-varchar-character-varying

### Known limitations

[N/A]


xdr/main.go
```diff
@@ -49,6 +44,26 @@ func SafeUnmarshalBase64(data string, dest interface{}) error {
 	return nil
 }
 
+// SafeUnmarshalBase64 first decodes the provided reader from base64 before
+// decoding the xdr into the provided destination.  Also ensures that the reader
+// is fully consumed.
+func SafeUnmarshalBase64(data string, dest interface{}) error {
```
abuiles, line 50: I think we already have this one?

PR #2937, opened 2020-08-20
services/horizon: Export horizon_build_info in /metrics

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This commit adds `horizon_build_info` to `/metrics` with the Horizon and Go runtime versions.

### Why

The metric labels will be used in Horizon dashboard to ensure all nodes are running the same version.

### DIFF
--- a/services/horizon/internal/init.go
+++ b/services/horizon/internal/init.go
@@ -3,6 +3,7 @@ package horizon
 import (
 	"context"
 	"net/http"
+	"runtime"
 
 	"github.com/getsentry/raven-go"
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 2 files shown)


services/horizon/internal/init.go
```diff
@@ -119,6 +120,16 @@ func initLogglyLog(app *App) {
...
 func initDbMetrics(app *App) {
+	app.buildInfoGauge = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{Namespace: "horizon", Subsystem: "build", Name: "info"},
+		[]string{"version", "goversion"},
+	)
+	app.prometheusRegistry.MustRegister(app.buildInfoGauge)
+	app.buildInfoGauge.With(prometheus.Labels{
+		"version":   app.horizonVersion,
+		"goversion": runtime.Version(),
+	}).Inc()
```
jacekn, line 131: Every project that I saw just uses static value of `1` for this metric. It's probably good to stick to that but realistically this does not matter that much.
I also saw projects expose git commit sha or build timestamp. If it's easy to do maybe include them as well?

jacekn (approved):
I added 2 small suggestions

services/horizon/internal/init.go
bartekn: > Every project that I saw just uses static value of 1 for this metric. It's probably good to stick to that but realistically this does not matter that much.

Yes, `Inc()` call actually increments the value to `1`.

> I also saw projects expose git commit sha or build timestamp. If it's easy to do maybe include them as well?

git commit sha is part of horizon version string.

jacekn: > > Every project that I saw just uses static value of 1 for this metric. It's probably good to stick to that but realistically this does not matter that much.
> 
> Yes, `Inc()` call actually increments the value to `1`.

Ah great. I though it would get called on every request to /metrics

> > I also saw projects expose git commit sha or build timestamp. If it's easy to do maybe include them as well?
> 
> git commit sha is part of horizon version string.

perfect!

tamirms (approved):

Issue #2932, opened 2020-08-19
Implement custom MarshalJSON and  UnmarshalJSON for xdr.ClaimPredicate


abuiles:
### What problem does your feature solve?

We are storing `xdr.ClaimPredicate` in the DB as the `base64` encoded value of the XDR which is fine and it helps us to verify state, we'll need a different representation for `xdr.ClaimPredicate` when rendering via the web api.

### What would you like to see?

Given a `xdr.ClaimPredicate` we should be able to Marshal it to JSON using the following representation

```
 {
  "and" : [
    {"after": 1594066025},
    {"before": 1594066090}
  ]
}
```

`ClaimPredicateTypeClaimPredicateUnconditional` should be `{}`.

At the moment, this can be worked on top of this branch https://github.com/abuiles/go/tree/get-claimable-balance-by-id which contains the resource adapter for claimable balances

abuiles:
@tamirms we also need to update the predicate value in the operation details processor and effects processor.

bartekn:
Fixed in #2947.

PR #2947, opened 2020-08-24
xdr: Add JSON representation for ClaimPredicate

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Close https://github.com/stellar/go/issues/2932

Add a JSON representation for `xdr.ClaimPredicate` which will be used in Horizon responses

### Why

Rendering a ClaimPredicate as JSON tree like:

```
 {
  "and" : [
    {"after": 1594066025},
    {"before": 1594066090}
  ]
}
```

is easier for humans to inspect than the base 64 encoding.

### Known limitations

[N/A]


xdr/json.go
```diff
@@ -0,0 +1,131 @@
...
+	"encoding/json"
+	"fmt"
+)
+
+type claimPredicateJSON struct {
+	And           *[]claimPredicateJSON `json:"and,omitempty"`
+	Or            *[]claimPredicateJSON `json:"or,omitempty"`
+	Not           *claimPredicateJSON   `json:"not,omitempty"`
+	Unconditional bool                  `json:"unconditional,omitempty"`
+	AbsBefore     *int64                `json:"absBefore,omitempty"`
```
bartekn, line 13: Should we make it a `time.Time`? This will make it readable and users won't need to convert the values from unix timestamp.

xdr/json_test.go
```diff
@@ -0,0 +1,58 @@
...
+					},
+					{
+						Type:      ClaimPredicateTypeClaimPredicateBeforeAbsoluteTime,
+						AbsBefore: &absBefore,
+					},
+				},
+			},
+			{
+				Type:         ClaimPredicateTypeClaimPredicateNot,
+				NotPredicate: &unconditional,
```
bartekn, line 34: I tried changing this to:
```go
				NotPredicate: nil,
```
and I got nil pointer dereference error. Should we support this case? I remember there was something special about `NotPredicate` (that's why it's a pointer to pointer).

bartekn (approved):
LGTM! This is great, makes predicates readable and it's now easy to understand requirements for claiming the balance! Added one suggestion and one question.

xdr/json_test.go
tamirms: the xdr definition of `ClaimPredicate` is:

```
union ClaimPredicate switch (ClaimPredicateType type)
{
case CLAIM_PREDICATE_UNCONDITIONAL:
    void;
case CLAIM_PREDICATE_AND:
    ClaimPredicate andPredicates<2>;
case CLAIM_PREDICATE_OR:
    ClaimPredicate orPredicates<2>;
case CLAIM_PREDICATE_NOT:
    ClaimPredicate* notPredicate;
case CLAIM_PREDICATE_BEFORE_ABSOLUTE_TIME:
    int64 absBefore; // Will return true if closeTime < absBefore
case CLAIM_PREDICATE_BEFORE_RELATIVE_TIME:
    int64 relBefore;    // Seconds since closeTime of the ledger in which the
                        // ClaimableBalanceEntry was created
};
```

`notPredicate` is defined as a pointer because of a limitation in xdrc. xdrc does not allow recursive data structures which is why `notPredicate` must be a `ClaimPredicate` pointer.

In practice, if the union is set to `CLAIM_PREDICATE_NOT` we expect `notPredicate` to be non nil

xdr/json.go
tamirms: I'm worried if there will be some ambiguity when it comes to the date format. if we render it as a unix timestamp it will be straight forward to parse because we don't need to standardize a date format which may include timezone etc

bartekn: We already use a time string in multiple places in Horizon, ex.: https://horizon.stellar.org/ledgers?order=desc so I think it shouldn't be an issue.

PR #2914, opened 2020-08-12
services/horizon/internal/expingest: Ingest CAP23 Operations effects.

abuiles:
<details>
  <summary>PR Checklist</summary>
  
### PR Structure

* [ ] This PR has reasonably narrow scope (if not, break it down into smaller PRs).
* [ ] This PR avoids mixing refactoring changes with feature changes (split into two PRs
  otherwise).
* [ ] This PR's title starts with name of package that is most changed in the PR, ex.
  `services/friendbot`, or `all` or `doc` if the changes are broad or impact many
  packages.

### Thoroughness

* [ ] This PR adds tests for the most critical parts of the new functionality or fixes.
* [ ] I've updated any docs ([developer docs](https://www.stellar.org/developers/reference/), `.md`
  files, etc... affected by this change). Take a look in the `docs` folder for a given service,
  like [this one](https://github.com/stellar/go/tree/master/services/horizon/internal/docs).

### Release planning

* [ ] I've updated the relevant CHANGELOG ([here](services/horizon/CHANGELOG.md) for Horizon) if
  needed with deprecations, added features, breaking changes, and DB schema changes.
* [ ] I've decided if this PR requires a new major/minor version according to
  [semver](https://semver.org/), or if it's mainly a patch change. The PR is targeted at the next
  release branch if it's not a patch change.
</details>

### What

Ingest effects for `CreateClaimableBalanceOp` and `ClaimClaimableBalanceOp`.

### Why

CAP23 introduces 2 new operations, we need to effects from each operation. 

Fix #2885 

### Known limitations

[TODO or N/A]

PR #2951, opened 2020-08-25
services/horizon: Update effects processor for CAP 33

### LINKED ISSUE #2891: Update ingestion effects processor to ingest sponsoring related effects
We need to update the effects processor to ingest the following effects:

- [ ] `sponsorship_created`: This appears when a new ledger entry is created and it has a sponsor. The effect will be associated with the operation creating the ledger entry. **Not all ledger entries have a created, removed, and update effect (e.g., offers) - we’ll add the accompanying sponsorship created effect only on the ledger entries which have effects**
- [ ] `sponsorship_updated`: This appears when a ledger entry is updated and the sponsor from the previous state is different to the current one. The effect will be associated with the operation updating the ledger entry.
- [ ] `sponsorship_removed`: This appears when a ledger entry is removed or as part of the `revoke_sponsorship_op`

Effects are processed in the `effects` functions in `transactionOperationWrapper`

Start in `history.EffectType = 60`

2opremio:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add sponsoring effects 

It also adds a refactoring of `effectsWrapper` (it didn't make sense to do it separately) so that we don't create a wrap per operation (the sponsoring effects obtention reuses the wrapper)

Fixes #2891 

### Why

We want to implement CAP 33

### Known limitations

N/A


2opremio:
I intentionally didn't add tests just yet (I want to get a GA of the approach first)

services/horizon/internal/expingest/processors/effects_processor.go
```diff
@@ -151,222 +151,241 @@ type effect struct {
...
+			)
+		case change.Post == nil && change.Pre != nil && change.Pre.SponsoringID() != nil:
+			e.add(e.operation.SourceAccount().Address(), history.EffectSponsorshipRemoved,
+				map[string]interface{}{
+					"former_sponsor": (*change.Pre.SponsoringID()).Address(),
+				},
+			)
+		case change.Pre != nil && change.Pre.SponsoringID() != nil &&
+			change.Post != nil && change.Post.SponsoringID() != nil:
+			e.add(e.operation.SourceAccount().Address(), history.EffectSponsorshipRemoved,
```
tamirms, line 268: I think this should be `EffectSponsorshipUpdated`

services/horizon/internal/expingest/processors/effects_processor.go
```diff
@@ -834,20 +789,36 @@ func ingestTradeEffects(effects *effectsWrapper, buyer xdr.AccountId, claims []x
...
-		effects.add(
+		e.add(
 			seller.Address(),
 			history.EffectTrade,
 			sd,
 		)
 	}
 }
 
+func effectFlagDetails(flagDetails map[string]interface{}, flagPtr *xdr.Uint32, setValue bool) {
```
tamirms, line 806: rename to `setEffectFlagDetails`?

tamirms:
@2opremio overall approach looks good to me

services/horizon/internal/expingest/processors/effects_processor.go
2opremio: This wasn't incorporated by this PR, but I will change it 👍

services/horizon/internal/expingest/processors/effects_processor.go
2opremio: Yes, you are right.

services/horizon/internal/expingest/processors/effects_processor_test.go
```diff
@@ -367,6 +561,34 @@ func TestOperationEffects(t *testing.T) {
...
+				{
+					address:     "GBRPYHIL2CI3FNQ4BXLFMNDLFJUNPU2HY3ZMFSHONUCEOASW7QC7OX2H",
+					operationID: int64(244813139969),
+					details: map[string]interface{}{
+						"sponsor": "GAHK7EEG2WWHVKDNT4CEQFZGKF2LGDSW2IVM4S5DP42RBW3K6BTODB4A",
+					},
+					effectType: history.EffectSponsorshipCreated,
+					order:      uint32(6),
+				},
 			},
```
2opremio, line 592: Is this ordering correct with respect to the prior effects in the slice?

2opremio: Otherwise I will need to change the approach and, instead of processing the sponsoring effects separately, intersperse them with the processing of each operation.

services/horizon/internal/expingest/processors/effects_processor.go
```diff
@@ -151,222 +151,248 @@ type effect struct {
...
+	case xdr.OperationTypeRevokeSponsorship:
+		op := operation.operation.Body.MustRevokeSponsorshipOp()
+		if op.Type == xdr.RevokeSponsorshipTypeRevokeSponsorshipSigner {
+			source := operation.operation.SourceAccount.ToAccountId()
+			wrapper.add(source.Address(), history.EffectSponsorshipRemoved,
+				map[string]interface{}{
+					"signer_account_id": op.Signer.AccountId.Address(),
+				},
+			)
+		}
```
2opremio, line 215: @tamirms Isn't this enough to cover signer sponsoring effects? (I don't think I need to look into the transaction changes like we discussed offline).

 At least if we follow https://docs.google.com/document/d/1Lor5OeOzPeuCID3-HSreqK-psjCvOXYfg22pwrUEUZA/edit#heading=h.cesdhaw1zjwx

2opremio:
@tamirms This should be ready for another look.

services/horizon/internal/expingest/processors/effects_processor.go
tamirms: yes, I think you're right

services/horizon/internal/expingest/processors/effects_processor_test.go
tamirms: @2opremio I think you can sort the effects in the processor by operation id. If you do that then the order of effects should be the same as the order of operations in the transaction.

services/horizon/internal/expingest/processors/effects_processor.go
```diff
@@ -522,23 +583,26 @@ func (operation *transactionOperationWrapper) changeTrustEffects() ([]effect, er
 		effect := history.EffectType(0)
 		addAssetDetails(details, op.Line, "")
 
+		addedTrustlineEffect := false
```
bartekn, line 586: Is this used anywhere other than in `if !addedTrustlineEffect {`?

services/horizon/internal/expingest/processors/effects_processor.go
```diff
@@ -629,36 +697,35 @@ func (operation *transactionOperationWrapper) manageDataEffects() ([]effect, err
 	if err != nil {
 		return effects.effects, err
 	}
-
+	dataEffectAdded := false
```
bartekn, line 700: Is it used anywhere?

services/horizon/internal/expingest/processors/effects_processor.go
```diff
@@ -673,23 +740,26 @@ func (operation *transactionOperationWrapper) bumpSequenceEffects() ([]effect, e
 		return effects.effects, err
 	}
 
+	accountEntryEffectAdded := false
```
bartekn, line 743: Is it used anywhere?

services/horizon/internal/expingest/processors/effects_processor.go
```diff
@@ -805,6 +877,25 @@ func (operation *transactionOperationWrapper) claimClaimableBalanceEffects() ([]
...
 }
 
+func (operation *transactionOperationWrapper) revokeSponsoringEffects() []effect {
+	op := operation.operation.Body.MustRevokeSponsorshipOp()
+	effects := effectsWrapper{
+		effects:   []effect{},
+		operation: operation,
+	}
+	if op.Type == xdr.RevokeSponsorshipTypeRevokeSponsorshipSigner {
+		source := operation.operation.SourceAccount.ToAccountId()
```
bartekn, line 887: `operation.operation.SourceAccount` can be `nil`. You can use `transactionOperationWrapper .SourceAccount()` to get source account of an operation (or a transaction if `operation.SourceAccount == nil`).

services/horizon/internal/expingest/processors/effects_processor.go
```diff
@@ -176,25 +222,25 @@ func (operation *transactionOperationWrapper) effects() (effects []effect, err e
 
 	switch operation.OperationType() {
 	case xdr.OperationTypeCreateAccount:
-		effects = operation.accountCreatedEffects()
+		effects, err = operation.accountCreatedEffects()
```
bartekn, line 225: Instead of calling `addSponsoringEffectsForChange` from related ops can we call it once in `transactionOperationWrapper.effects()`? This should simplify the code without affecting it's logic (if related changes are not there, the effects won't be added).

services/horizon/internal/expingest/processors/effects_processor.go
```diff
@@ -207,18 +253,20 @@ func (operation *transactionOperationWrapper) effects() (effects []effect, err e
 		effects, err = operation.createClaimableBalanceEffects()
```
bartekn, line 253: I think that a claimable balance can also have a sponsor. Maybe we should move `addSponsoringEffectsForChange` outside this `switch` (see other comment)?

services/horizon/internal/expingest/processors/effects_processor.go
```diff
@@ -573,7 +637,11 @@ func (operation *transactionOperationWrapper) allowTrustEffects() []effect {
 		effects.add(source.Address(), history.EffectTrustlineDeauthorized, details)
 	}
 
-	return effects.effects
+	if err := effects.addSponsoringEffectsForEntryType(xdr.LedgerEntryTypeTrustline); err != nil {
```
bartekn, line 640: I think allow trust op can't modify sponsors.

2opremio: not directly, but it can have `LedgerEntryTypeTrustline` ledger entry changes, correct?

services/horizon/internal/expingest/processors/effects_processor.go
2opremio: no, why?

services/horizon/internal/expingest/processors/effects_processor.go
2opremio: yes

services/horizon/internal/expingest/processors/effects_processor.go
2opremio: Yes

services/horizon/internal/expingest/processors/effects_processor.go
2opremio: Ah, really good point. Thanks!

services/horizon/internal/expingest/processors/effects_processor.go
2opremio: That's exactly what the previous commit does. I changed the code to ensure that the sponsoring effects and other effects were ordered consistently with respect to the operation changes.

However, I was discussing with @tamirms whether this was necessary. This is hairy, so maybe we can jump into a call to clarify it.

services/horizon/internal/expingest/processors/effects_processor.go
2opremio: I intentionally didn't do it this way for the same reason as https://github.com/stellar/go/pull/2951/files#r480425973

services/horizon/internal/expingest/processors/effects_processor.go
2opremio: See https://github.com/stellar/go/compare/protocol-14...89f4c7c57eba14bdb18e6d7592929f5a0467a542 for the approach I followed before this change.

services/horizon/internal/expingest/processors/effects_processor.go
bartekn: Wouldn't code without it work in the same way (because it's `false` the `if !addedTrustlineEffect {` block always executes)? It would make sense if you flipped it to `true` somewhere but you don't.

services/horizon/internal/expingest/processors/effects_processor.go
bartekn: I think it's the same issue as in https://github.com/stellar/go/pull/2951/files#r480972532.

services/horizon/internal/expingest/processors/effects_processor.go
bartekn: Same issue as in https://github.com/stellar/go/pull/2951/files#r480972532.?

services/horizon/internal/expingest/processors/effects_processor.go
bartekn: That is correct but none of the changes (results of `allow_trust`) will modify the sponsor. Unless there's a special case for this.

services/horizon/internal/expingest/processors/effects_processor.go
bartekn: OK, but there should be a code here to add sponsor effects.

services/horizon/internal/expingest/processors/effects_processor.go
bartekn: Thanks for the pointer, I'll check it out.

2opremio:
@tamirms @bartekn As agreed privately, I restored my prior approach and addressed the comments. PTAL

services/horizon/internal/expingest/processors/effects_processor_test.go
2opremio: As we discussed privately, ordering by operation id doesn't help because I was referring to the order of effects with respect to the LedgerEntry changes of a single operation (i.e. there is just a single operation id)

bartekn:
Regarding `sponsorship_created` effect: during Horizon Team meeting yesterday we were wondering if we should create `sponsorship_created` (but this requires us to uniquely identify ledger entry being sponsored) or add `sponsor` field to `*_created` effects (but this modifies existing effects and also we don't have created effects for some entries like offers).

I think there's an alternative approach that's more consistent with existing effects. The account or trustline effects are actually very granular. Instead of `account_updated` we have effects like `account_flags_updated` or `account_thresholds_updated`. I think we can create separate effects for ledger entries but only for these which are already implemented so:
* `account_sponsor_created`
* `account_sponsor_updated`
* `account_sponsor_removed`
* `trustline_sponsor_created`
* `trustline_sponsor_updated`
* `trustline_sponsor_removed`

The reason for not adding effects for offers or data is that we don't have corresponding `created` effects for them.

tamirms:
@bartekn I think that makes sense

tamirms:
I ran horizon using this branch in a protocol 14 standalone network. I tested the following scenario:

* Create a [sponsored account](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0033.md#example-sponsoring-account-creation) `S`
* Create a new account `N`
* Merge account `S` into `N`

Here are the effects generated by @2opremio 's effects processor:

```
horizon=# select * from history_effects ;
 history_account_id | history_operation_id | order | type |                                         details                                         
--------------------+----------------------+-------+------+-----------------------------------------------------------------------------------------
                  2 |         420906799106 |     1 |    0 | {"starting_balance": "10.0000000"}
                  1 |         420906799106 |     2 |    3 | {"amount": "10.0000000", "asset_type": "native"}
                  2 |         420906799106 |     3 |   10 | {"weight": 1, "public_key": "GDGZLS7XYNWKTOM5FMTFFN5AI5I6BMSCA5Q6KQQ3H4ABCKQFR2DV4RLQ"}
                  1 |         420906799106 |     4 |   60 | {"sponsor": "GBZXN7PIRZGNMHGA7MUUUF4GWPY5AYPV6LY4UV2GL6VJGIQRXFDNMADI"}


                  5 |         652835033089 |     1 |    0 | {"starting_balance": "100.0000000"}
                  1 |         652835033089 |     2 |    3 | {"amount": "100.0000000", "asset_type": "native"}
                  5 |         652835033089 |     3 |   10 | {"weight": 1, "public_key": "GAIY5RDADKTODYVM7TIGZUN5VONWLVBJBP3BBI6G26YTOKC35E5B5HZP"}


                  2 |        1215475748865 |     1 |    3 | {"amount": "9.9999900", "asset_type": "native"}
                  5 |        1215475748865 |     2 |    2 | {"amount": "9.9999900", "asset_type": "native"}
                  2 |        1215475748865 |     3 |    1 | {}
                  2 |        1215475748865 |     4 |   62 | {"former_sponsor": "GBZXN7PIRZGNMHGA7MUUUF4GWPY5AYPV6LY4UV2GL6VJGIQRXFDNMADI"}
(11 rows)

```
 
The effects from the first transaction (create sponsored account) are: `EffectAccountCreated` , `EffectAccountDebited`, `EffectSignerCreated`, `EffectSponsorshipCreated`.

The effects from the second transaction (create normal unsponsored account) are:  `EffectAccountCreated`, `EffectAccountDebited`, `EffectSignerCreated`.

The effects from the third transaction (merge sponsored account into unsponsored account) are: `EffectAccountDebited`, `EffectAccountCredited`, `EffectAccountRemoved`, `EffectSponsorshipRemoved`.

These effects and the ordering of the effects match my expectation. The PR looks good to me!

2opremio:
@tamirms Nice! Thanks for running it

2opremio:
> The reason for not adding effects for offers or data is that we don't have corresponding `created` effects for them. What do you think?

I have updated the PR to ignore those changes. Note that we would still generate effects for changes in `ClaimableBalance` ledger entries. Should we ignore those too?

tamirms:
@2opremio we have effects for claimable balances (implemented here https://github.com/stellar/go/pull/2914/files ). Therefore, I believe we should still generate sponsorship effects for claimable balances.

bartekn:
> we have effects for claimable balances (implemented here https://github.com/stellar/go/pull/2914/files ). Therefore, I believe we should still generate sponsorship effects for claimable balances.

Yes, let's add it for claimable balances too.

2opremio:
Then it should be good to go

2opremio:
I misunderstood @bartekn 's comment now it should really be ready to go


(on the way I added the textual representation of claimable balance effects, which were missing)

services/horizon/internal/expingest/processors/effects_processor.go
```diff
@@ -151,222 +151,275 @@ type effect struct {
...
+			(change.Post != nil && change.Post.SponsoringID() != nil):
+			preSponsor := (*change.Pre.SponsoringID()).Address()
+			postSponsor := (*change.Post.SponsoringID()).Address()
+			if preSponsor == postSponsor {
+				continue
+			}
+			e.add(e.operation.SourceAccount().Address(), effectsForEntryType.updated,
+				map[string]interface{}{
+					"new_sponsor":    (*change.Post.SponsoringID()).Address(),
+					"former_sponsor": (*change.Pre.SponsoringID()).Address(),
```
bartekn, line 305: Nit
```suggestion
					"new_sponsor":    postSponsor,
					"former_sponsor": preSponsor,
```

services/horizon/internal/resourceadapter/effects.go
```diff
@@ -36,6 +36,19 @@ var EffectTypeNames = map[history.EffectType]string{
...
+	history.EffectClaimableBalanceClaimantCreated:          "claimable_balance_claimant_created",
+	history.EffectClaimableBalanceClaimed:                  "claimable_balance_claimed",
+	history.EffectAccountSponsorshipCreated:                "account_sponsorhip_created",
+	history.EffectAccountSponsorshipUpdated:                "account_sponsorhip_udpated",
+	history.EffectAccountSponsorshipRemoved:                "account_sponsorship_removed",
+	history.EffectTrustlineSponsorshipCreated:              "trustline_sponsorhip_created",
+	history.EffectTrustlineSponsorshipUpdated:              "trustline_sponsorhip_udpated",
+	history.EffectTrustlineSponsorshipRemoved:              "trustline_sponsorship_removed",
+	history.EffectClaimableBalanceSponsorshipCreated:       "claimable_balance_sponsorhip_created",
+	history.EffectClaimableBalanceSponsorshipUpdated:       "claimable_balance_sponsorhip_udpated",
```
bartekn, line 49:
```suggestion
	history.EffectClaimableBalanceSponsorshipCreated:       "claimable_balance_sponsorship_created",
	history.EffectClaimableBalanceSponsorshipUpdated:       "claimable_balance_sponsorship_updated",
```

services/horizon/internal/resourceadapter/effects.go
```diff
@@ -36,6 +36,19 @@ var EffectTypeNames = map[history.EffectType]string{
 	history.EffectDataRemoved:                              "data_removed",
 	history.EffectDataUpdated:                              "data_updated",
 	history.EffectSequenceBumped:                           "sequence_bumped",
+	history.EffectClaimableBalanceCreated:                  "claimable_balance_created",
+	history.EffectClaimableBalanceClaimantCreated:          "claimable_balance_claimant_created",
+	history.EffectClaimableBalanceClaimed:                  "claimable_balance_claimed",
+	history.EffectAccountSponsorshipCreated:                "account_sponsorhip_created",
+	history.EffectAccountSponsorshipUpdated:                "account_sponsorhip_udpated",
```
bartekn, line 43:
```suggestion
	history.EffectAccountSponsorshipCreated:                "account_sponsorship_created",
	history.EffectAccountSponsorshipUpdated:                "account_sponsorship_updated",
```

services/horizon/internal/resourceadapter/effects.go
```diff
@@ -36,6 +36,19 @@ var EffectTypeNames = map[history.EffectType]string{
...
+	history.EffectAccountSponsorshipRemoved:                "account_sponsorship_removed",
+	history.EffectTrustlineSponsorshipCreated:              "trustline_sponsorhip_created",
+	history.EffectTrustlineSponsorshipUpdated:              "trustline_sponsorhip_udpated",
```
bartekn, line 46:
```suggestion
	history.EffectTrustlineSponsorshipCreated:              "trustline_sponsorship_created",
	history.EffectTrustlineSponsorshipUpdated:              "trustline_sponsorship_updated",
```

bartekn (approved):

2opremio:
CircleCI is failing to restore the cache.

2opremio:
> The reason for not adding effects for offers or data is that we don't have corresponding `created` effects for them. What do you think?

I have just checked the code and we **do** have effects for Data and Offers (`history.EffectData`* and `history.EffectOffer`*).

It doesn't seem like we emit the `history.EffectOffer`*, but we do emit `history.EffectData`* effects:

https://github.com/stellar/go/blob/1bfd026357872162c26d3da487fb0d222e0267c5/services/horizon/internal/expingest/processors/effects_processor.go#L737-L752

Thus, I think I will implemented the sponsorship effects for data.

PR #2961, opened 2020-08-27
services/horizon: Add port specification to work on host network mode

Shaptic:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Specifies the port for the core PostgreSQL database so that `docker-compose` works on Linux.

### Why
Docker uses the `host` networking mode on Linux when setting up Horizon (see the [startup guide](https://github.com/stellar/go/blob/master/services/horizon/docker/README.md)). This means there is no port mapping made, so the core container cannot access its respective database. 

This change makes the database use the "host-facing" port and not map to the default (5432) internally.

### Known limitations
Its possible that [someone, somewhere](https://xkcd.com/1172/) is relying on the core database docker container to use the default port 5432, but this port is *only* seen from **within** said container, so that seems unlikely.

### DIFF
--- a/services/horizon/docker/docker-compose.yml
+++ b/services/horizon/docker/docker-compose.yml
@@ -7,7 +7,8 @@ services:
       - POSTGRES_PASSWORD=mysecretpassword
       - POSTGRES_DB=stellar
     ports:
-      - "5641:5432"
+      - "5641:5641"
+    command: ["-p", "5641"]
     volumes:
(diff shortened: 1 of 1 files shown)


tamirms (approved):
awesome! great catch and thanks for fixing this :)

Shaptic:
The thanks should go to @2opremio for helping me work through it. :) thanks for reviewing!

PR #2963, opened 2020-08-27
recoverysigner: set max db open conns to 20

howardtw:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This PR sets the max database open connections and idle connections to 20. If we find it problematic, we can just the flag `-db-max-open-conns`  or the env `DB_MAX_OPEN_CONNS` to adjust the number.

### Why

We had a problem where all connections in the same environment were consumed.

### Known limitations

[N/A]

### DIFF
--- a/exp/services/recoverysigner/internal/serve/serve.go
+++ b/exp/services/recoverysigner/internal/serve/serve.go
@@ -21,14 +21,15 @@ import (
 )
 
 type Options struct {
-	Logger            *supportlog.Entry
-	DatabaseURL       string
-	Port              int
-	NetworkPassphrase string
-	SigningKeys       string
-	SEP10JWKS         string
-	SEP10JWTIssuer    string
-	FirebaseProjectID string
+	Logger               *supportlog.Entry
+	DatabaseURL          string
+	DatabaseMaxOpenConns int
+	Port                 int
+	NetworkPassphrase    string
+	SigningKeys          string
+	SEP10JWKS            string
+	SEP10JWTIssuer       string
+	FirebaseProjectID    string
 
 	AdminPort        int
 	MetricsNamespace string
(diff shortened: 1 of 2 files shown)


exp/services/recoverysigner/internal/serve/serve.go
```diff
@@ -103,6 +104,9 @@ func getHandlerDeps(opts Options) (handlerDeps, error) {
 	if err != nil {
 		return handlerDeps{}, errors.Wrap(err, "error parsing database url")
 	}
+	db.SetMaxOpenConns(opts.DatabaseMaxOpenConns)
+	db.SetMaxIdleConns(opts.DatabaseMaxOpenConns)
```
leighmcculloch, line 108: I don't think we need to set the max idle conns here to the same value as part of the fix. The default value of 2 should probably suffice. That'll allow connections to be released and closed when not in use rather than keeping the full number of connections open in an idle state.

What was the reason for setting the max idle to the same as the max open?

Another reason to remove this line is someone deploying this may want to max the max open connections unlimited, i.e. zero, but they probably wouldn't want to set the max idle conns to zero which would mean none are kept idle.

These should probably be separate configurations, or we just not set the second one and use Go's default.

leighmcculloch (commented):
👏 Good catch. I think we should remove the idle conns change though, see below.

exp/services/recoverysigner/internal/serve/serve.go
howardtw: > Another reason to remove this line is someone deploying this may want to max the max open connections unlimited, i.e. zero, but they probably wouldn't want to set the max idle conns to zero which would mean none are kept idle.

That's a good catch that it would be a bug when someone sets the value to zero! We should probably still have two separate configurations for that sake. It was done because it was an urgent fix, and I thought we always want to set them to the same value.

> That'll allow connections to be released and closed when not in use rather than keeping the full number of connections open in an idle state.

Wouldn't that cause unnecessary overhead which in turn lower our performance?

> What was the reason for setting the max idle to the same as the max open?

Like I said above, I thought there's no downside of keeping as many as idle connections in the pool so that we can one available instantly whenever we need it. What is the downside of doing such?

Setting the max open connection and the max idle connection to the same value has been a practice I follow. I'm looking forward to learning something new.

leighmcculloch: Keeping connections open and idle for a long time can cause problems. I don't remember the details, I read about someones experience a while ago and I've experienced it once, this isn't unique to Go. If all the connections are kept idle and there's a lull in traffic, when traffic picks up we might find some of them or all of them if we're unlucky not in a good state if something happens the connections are not healthy but have somehow remained idle. https://github.com/stellar/go/pull/3052

PR #2967, opened 2020-08-28
{txnbuild, webauth}: support the SEP-10 V2

JakeUrban:
resolves stellar/go#2968

Makes the same changes made to the [JS SDK](https://github.com/stellar/js-stellar-sdk/pull/568).


tamirms:
@JakeUrban what's the rationale for not validating home domain the go sdk?

ire-and-curses:
@cla-bot check

JakeUrban:
@tamirms I didn't want to introduce breaking changes to the other SEP-10 utility functions that make calls to `ReadChallengeTx`. But after seeing the ecosystem release the equivalent changes in their SDK's, I can add it here.

tamirms (approved):

JakeUrban:
@cla-bot check

JakeUrban:
@tamirms can you take another look and verify its ready for release?

exp/services/webauth/internal/serve/token.go
```diff
@@ -49,7 +49,7 @@ func (h tokenHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	var clientAccountID string
 	var signingAddress *keypair.FromAddress
 	for _, s := range h.SigningAddresses {
-		tx, clientAccountID, err = txnbuild.ReadChallengeTx(req.Transaction, s.Address(), h.NetworkPassphrase)
+		tx, clientAccountID, err = txnbuild.ReadChallengeTx(req.Transaction, s.Address(), h.NetworkPassphrase, r.Host)
```
tamirms, line 52: I don't think using `r.Host` is a reliable way for obtaining the domain of the client side request. there could be a load balancer sitting in front of webauth which proxies requests from the client.

@howardtw can you review the webauth changes and advise @JakeUrban on what the best way would be to configure the home domain?

howardtw: For sure!

@JakeUrban I have some questions regarding the home domain change in SEP-10:
1. In the [abstract](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0010.md#abstract), we say that
```
The client verifies that the value contains the home domain used to request the challenge.
```
However, in the [request](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0010.md#request) of the GET endpoint, we say the home domain is optional and the default value will be used if not provided.

I'm not sure how a client verifies the default home domain used by the server if she didn't provide one in the first place. Do you mean the client can skip verifying the home domain in this case?

2. How does the server know the home domain used to generate the challenge matches the one contained in the transaction at the POST [endpoint](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0010.md#request-1)? Looks like you have to read an extra parameter here via `r.Host`, which is not specified in SEP-10 as far as I'm concerned.

In the [description](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0010.md#token) of handling the token, we still say
```
the server has to verify that transaction contains a single Manage Data operation that has the service's home domain as the key of the operation
```
Does that mean the server can only validate it with the default home domain?

JakeUrban: @howardtw good questions, let me try and address each of them. I would also recommend checking out the [JS](https://github.com/stellar/js-stellar-sdk/pull/568) or [Python](https://github.com/StellarCN/py-stellar-base/pull/363) changes.

> I'm not sure how a client verifies the default home domain used by the server if she didn't provide one in the first place. Do you mean the client can skip verifying the home domain in this case?

With this change, SEP-10 servers return JWT tokens for <`account`, TOML `homeDomain`> pairs. Clients specify the `account`, but servers return the challenge transaction with a `homeDomain` included in the Manage Data operation. 

Typically, SEP-10 servers only issue tokens for _one_ home domain, so SEP-10 servers can use it as the default home domain included in the challenge transaction's Manage Data operation's key. This change adds a `home_domain` parameter the client can specify if the SEP-10 endpoint issues tokens for multiple `homeDomain`s.

Regardless of whether or not the client specifies a `home_domain` parameter in the request, the client _must_ verify that the home domain included in the Manage Data operation's name/key matches the home domain used to fetch the SEP-10 `WEB_AUTH_ENDPOINT` attribute, _and_ that the `SIGNING_KEY` included in the same TOML file has signed the challenge transaction received from `GET WEB_AUTH_ENDPOINT`.

> How does the server know the home domain used to generate the challenge matches the one contained in the transaction at the POST endpoint? Looks like you have to read an extra parameter here via r.Host, which is not specified in SEP-10 as far as I'm concerned.

The server can inspect the challenge `POST`ed to the endpoint and confirm the home domain included in the Manage Data operation is one of the services the SEP-10 endpoint gives tokens for. Obviously it will also verify that it's `SIGNING_KEY` and the client's `account` both signed the transaction.

`r.HOST` is a Go attribute that wouldn't be mentioned in SEP-10. It is my best attempt at retrieving the home domain the anchor should expect in the `req.Transaction` read by the `ReadChallengeTx` call. My question to you is, whats the best way to extract the home domain from the request object and use that to check the Manage Data operation's key?

howardtw: > This change adds a home_domain parameter the client can specify if the SEP-10 endpoint issues tokens for multiple homeDomains.

I guess my questions are mostly around this one. When a server supports multiple home domains, does it also typically have a pre-defined list of home domains?

> the client must verify that the home domain included in the Manage Data operation's name/key matches the home domain used to fetch the SEP-10 WEB_AUTH_ENDPOINT attribute

Similar to above, what if the client provides a home domain supported by the server that is different from the home domain in the URI? Is the above statement still valid? I think an example of a server supporting multiple home domains could make things clear.

> that the SIGNING_KEY included in the same TOML file has signed the challenge transaction received from GET WEB_AUTH_ENDPOINT

Currently the web auth server supports multiple signing keys for key rotation. Can a TOML file have multiple signing keys? Otherwise it seems like we will have a case that we couldn't handle if we can only sign with one single key specified in the TOML file.

> It is my best attempt at retrieving the home domain the anchor should expect in the req.Transaction read by the ReadChallengeTx call.

I don't understand this one. I thought the server just has to match the home domain contained in the operation against the list of home domains it supports. Why do we have to read it from the request?

JakeUrban: > I guess my questions are mostly around this one. When a server supports multiple home domains, does it also typically have a pre-defined list of home domains?

Yes. If the `home_domain` parameter passed in the `GET WEB_AUTH_ENDPOINT` request isn't a valid home domain for the server the response will include a `400` error status.

> Similar to above, what if the client provides a home domain supported by the server that is different from the home domain in the URI? Is the above statement still valid? I think an example of a server supporting multiple home domains could make things clear.

Ah, so here is an important difference. The URI for `WEB_AUTH_ENDPOINT` could be different than the URI used to find the TOML file. 

For example, for the home domain testanchor.stellar.org, and this is the home domain the server and client should check the Manage Data operation's key for. But the TOML file at this home domain may have a `WEB_AUTH_ENDPOINT` with a _different_ home domain.

So, the URI used in the request for the challenge may be different than the `home_domain` specified in the `GET` request.

> Currently the web auth server supports multiple signing keys for key rotation. Can a TOML file have multiple signing keys? Otherwise it seems like we will have a case that we couldn't handle if we can only sign with one single key specified in the TOML file.

You can only have one `SIGNING_KEY` per TOML, as specified in SEP-1. I'm not sure what you mean by 'web auth server'. Can you link the documentation for key rotation? 

>  I thought the server just has to match the home domain contained in the operation against the list of home domains it supports. Why do we have to read it from the request?

You're right, but all we have is the `r` Request object. How would you make the changes to ensure the call to `ReadChallengeTx()` includes the correct home domain parameter value?

howardtw: > the URI used in the request for the challenge may be different than the home_domain specified in the GET request.

I see. Thanks for the explaining the differentiation! So the home domain change in SEP-10 refers to the home domain of the TOML file, and it could be the same or sometimes different than the URI of the `WEB_AUTH_ENDPOINT`.

> the client must verify that the home domain included in the Manage Data operation's name/key matches the home domain used to fetch the SEP-10 WEB_AUTH_ENDPOINT attribute

I think I will have to rephrase my question as such:
What if the client provides a home domain supported by the server that is different from the home domain of the TOML file? This could happen when the server supports multiple home domains, right?

> You can only have one SIGNING_KEY per TOML, as specified in SEP-1. I'm not sure what you mean by 'web auth server'. Can you link the documentation for key rotation?

Yep for sure. [This](https://github.com/stellar/go/pull/2667) is the PR that introduces the change. It's mainly for preventing the outage after a key rotation. cc @accordeiro curious about your thoughts here.

> How would you make the changes to ensure the call to ReadChallengeTx() includes the correct home domain parameter value?

I think `ReadChallengeTx` can either take a list of home domains or the caller will have to call the `ReadChallengeTx` multiple times. If one of the attempts succeeds, then we are good. If all the attempts fails, we return an error.

The list of home domains the server supports would better be provided through configurations like we how read the signing keys above. I think this has to be part of the dependency of the `tokenHandler` rather than reading it from a request.

JakeUrban: > What if the client provides a home domain supported by the server that is different from the home domain of the TOML file? This could happen when the server supports multiple home domains, right?

Thats the client's fault then, and the challenge returned will not be verified by the client because the `SIGNING_KEY` on the TOML didn't sign the challenge transaction.

> I think `ReadChallengeTx` can either take a list of home domains or the caller will have to call the `ReadChallengeTx` multiple times

I would like to avoid `ReadChallengeTx` accepting a list of home domains. I made the changes to `exp/services/webauth/internal/serve/token.go` because the Go SDK is in the mono repo, but all the other SDK's accept a single home domain for their related function. 

I agree `tokenHandler` may have to be adjusted, but I'm not familiar with what the context of this `webauth/internal` code is. Who would be the best person to figure out this change to `ReadChallengeTx` applies to `tokenHandler`?

howardtw: > I agree tokenHandler may have to be adjusted, but I'm not familiar with webauth/internal code. Who would be the best person to figure out this change to ReadChallengeTx applies to tokenHandler?

I can help making the change 😄 Do you want me to checkout out branch and commit to it directly?

JakeUrban: That would be great thank you!

howardtw: Allow me to follow up on this matter as I'm still figuring out this new SEP change in order to make the corresponding code change.

> Thats the client's fault then, and the challenge returned will not be verified by the client because the SIGNING_KEY on the TOML didn't sign the challenge transaction.

When a server supports multiple home domains, doesn't that mean it has the access to the `SIGNING_KEY` of all the TOML files? I guess it won't be the client's fault because she can still check whether the `SIGNING_KEY` matches the one on the TOML hosted on the domain specified in the request even though it's not the default one? The prerequisites will be that those two TOML files have to have the same configuration of the `WEB_AUTH_ENDPOINT`. If this is not possible, I don't know how a server can support multiple home domains 🤔

JakeUrban: > When a server supports multiple home domains, doesn't that mean it has the access to the SIGNING_KEY of all the TOML files?

Thats correct. A SEP-10 endpoint would return a challenge to the client signed by the `SIGNING_KEY` in TOML listed on the home domain passed in the request, if its valid. Otherwise, it will sign the transaction using the `SIGNING_KEY` listed for the default home domain.

> I guess it won't be the client's fault because she can still check whether the SIGNING_KEY matches the one on the TOML hosted on the domain specified in the request even though it's not the default one?

If the client passes a home domain that is _known_ to the SEP-10 endpoint but _not_ the home domain of the service the client needs to authenticate with, the challenge returned will fail the client validation when it checks the transaction's signatures and doesn't find the `SIGNING_KEY` signature of the correct home domain / service.

> The prerequisites will be that those two TOML files have to have the same configuration of the WEB_AUTH_ENDPOINT. If this is not possible, I don't know how a server can support multiple home domains 🤔

You're correct! They must have the same WEB_AUTH_ENDPOINT

howardtw: https://github.com/stellar/go/pull/2967/commits/a2d2e151034ea4098663c468690b9e1f2e680f3d
^ @tamirms @JakeUrban can you take a look at the new web auth server change?

howardtw:
Please do not merge this PR just yet as there are kube settings that I need to configure before the web auth service can function properly.

howardtw:
The kube settings change is here: https://github.com/stellar/kube/pull/338.

exp/services/webauth/internal/serve/serve.go
```diff
@@ -48,18 +54,39 @@ func Serve(opts Options) {
...
+	signingAddresses := make([]*keypair.FromAddress, 0, len(signingKeyStrs))
+
+	for i, signingKeyStr := range signingKeyStrs {
 		signingKey, err := keypair.ParseFull(signingKeyStr)
 		if err != nil {
 			return nil, errors.Wrap(err, "parsing signing key seed")
 		}
-		signingKeys = append(signingKeys, signingKey)
+
+		if i == 0 {
```
tamirms, line 67: why do we only validate the first signing key?

exp/services/webauth/internal/serve/challenge.go
```diff
@@ -28,17 +29,36 @@ type challengeResponse struct {
...
+	queryValues := r.URL.Query()
 
-	account := r.URL.Query().Get("account")
+	account := queryValues.Get("account")
 	if !strkey.IsValidEd25519PublicKey(account) {
 		badRequest.Render(w)
 		return
 	}
 
+	homeDomain := queryValues.Get("home_domain")
```
JakeUrban, line 40: Are we concerned at all about clients sending variations of the domain string in the `h.HomeDomains` list? For example if `h.HomeDomains` contains a `service.domain.com` string but the client sends `https://service.domain.com` or `www.service.domain.com`.

The SEP specifies the value should be the [FQDN](https://en.wikipedia.org/wiki/Fully_qualified_domain_name), but if the client is out of spec, should we try and normalize the URLs sent and match the normalized strings?

Another possibility is the _server_ getting the string wrong. For example, if the anchor includes `www.service.domain.com`, and the client sends the correct string `service.domain.com`, `ServeHTTP` won't work.

exp/services/webauth/internal/serve/challenge_test.go
```diff
@@ -65,8 +66,60 @@ func TestChallenge(t *testing.T) {
...
+
+	r := httptest.NewRequest("GET", fmt.Sprintf("/?account=%s&home_domain=%s", account.Address(), anotherDomain), nil)
+	w := httptest.NewRecorder()
+	h.ServeHTTP(w, r)
+	resp := w.Result()
+
+	require.Equal(t, http.StatusOK, resp.StatusCode)
+	assert.Equal(t, "application/json; charset=utf-8", resp.Header.Get("Content-Type"))
+
+	res := struct {
```
JakeUrban, line 90: I'm new to Go but it seems like quite tedious thing to require devs to define the attributes in HTTP response bodies in-code prior to decoding the actual response. Is there a way to only extract a subset of attributes or are there libraries that build these structs by inspecting the response?

I'm not saying we need to do this, its more for my understanding.

exp/services/webauth/internal/serve/serve.go
```diff
@@ -48,18 +54,39 @@ func Serve(opts Options) {
 }
 
 func handler(opts Options) (http.Handler, error) {
```
JakeUrban, line 56: What is this handler for? Is this the server process init function?

exp/services/webauth/internal/serve/serve.go
JakeUrban: We only have a single `SIGNING_KEY` on a stellar.toml file, so this function requires the first signing key in the parsed list to be the one included on the TOML. @howardtw I'm noticing this code is light on comments in general. Would this first-key requirement warrant a comment?

exp/services/webauth/internal/serve/token.go
```diff
@@ -48,10 +49,17 @@ func (h tokenHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-			signingAddress = s
+		for _, domain := range h.HomeDomains {
+			tx, clientAccountID, err = txnbuild.ReadChallengeTx(req.Transaction, s.Address(), h.NetworkPassphrase, domain)
+			if err == nil {
+				signingAddress = s
+				homeDomain = domain
+				break
+			}
+		}
+		if signingAddress != nil {
```
JakeUrban, line 62: Do we want to log a message if the challenge can't be read with any of the signing key / domain pairs?

JakeUrban (commented):
@howardtw your changes look great 👍  it won't let me approve my own PR but LGTM

exp/services/webauth/internal/serve/challenge_test.go
howardtw: I think what we want to do when it comes to testing http handlers is to have an expectation in mind before unmarshaling the response. In this case, the expectation is the `res` struct. And if unmarshaling the response from the http handler to `res`  doesn't assign the values to all of fields in `res`, we know there's something wrong with the handler.

> Is there a way to only extract a subset of attributes

You can only extract a subset of attributes by making the struct look like
```
res := struct {Transaction string `json:"transaction"`}{}
```
This way you can extract only the `transaction`.

> are there libraries that build these structs by inspecting the response?

I'm not aware of any library that does it. Typically people don't really do this because we usually compare the result with our expectation.

exp/services/webauth/internal/serve/challenge.go
howardtw: You're right. I implemented it following the requirement of using FQDN in SEP-10.

In my view, trying to normalize a domain will make the code messier and we probably don't want to make any assumption when it comes to verifying the domains that use web auth. My concern is that domains with and without `www` are fundamentally different. For example, `en.wikipedia.org` will take you to the right domain and `www.en.wikipedia.org` won't.

Most browsers have the predication utility built in to make it convenient when `example.com` has nothing but `www.example.com` has something. However, it's hard for us to predict whether a client actually do authentication with a domain `example.com` or `www.example.com`.

> Another possibility is the server getting the string wrong. For example, if the anchor includes www.service.domain.com, and the client sends the correct string service.domain.com, ServeHTTP won't work.

Based on the above reason, I tend to leave it as it is about how we handle `www`.

Regarding `https://service.domain.com`, not sure we should just return an error or try to strip the prefix before we process the request as this is not really a domain. I'm fine with trying to trim the prefix here.

exp/services/webauth/internal/serve/serve.go
howardtw: This is a function we just to process the configs from `Options` and initialize the handlers used by each endpoint.

exp/services/webauth/internal/serve/serve.go
howardtw: @tamirms like what @JakeUrban said, only the first key is used for signing purpose. The rest keys are used for verification purpose, if any.

@JakeUrban I agree it would be more readable if there are comments around these logics. Will add it.

exp/services/webauth/internal/serve/token.go
howardtw: We return a bad request down [below](https://github.com/stellar/go/pull/2967/files#diff-e713d2c109b7963cf50c0b18b7324f14R66-R69) when that happens.

IMO the log will be a little too chatty if we do it. However it can also be helpful for the anchors to check what's going on initially during the transition. I'm down to do it, but I think it's better done in another PR as that's not what we have been doing in the past. Can you create an issue if you think it's valuable?

exp/services/webauth/internal/serve/challenge.go
JakeUrban: I see, good point about `www`. Your call with `https`.

exp/services/webauth/internal/serve/token.go
JakeUrban: Oh I missed the bad request response. I'm not sure how the `return/Render()` keyword and function work but it may be valuable to add a `return` code based on the HTTP code returned with `Render()`. This would allow the code that calls `ServeHTTP` to determine what happened and log the error message if they chose, instead of the SDK deciding for them.

JakeUrban: I'll file an issue for this.

howardtw: > I'm not sure how the return/Render() keyword and function work but it may be valuable to add a return code based on the HTTP code returned with Render(). This would allow the code that calls ServeHTTP to determine what happened and log the error message if the developers chose, instead of the SDK deciding for them.

By saying `add a return code`, do you mean that you want to define an error spec? `badRequest.Render` will return a 400 with an error message saying `The request was invalid in some way.`. 

Logging the error message here will be tricky actually because we would have trued `num(keys) x num(domains)` times, and each time we will receive different errors presumably. We can probably only say something `either the signature cannot be verified by the our signing key(s) or the auth domain is not supported`.

JakeUrban: Actually I take back this suggestion entirely, I didn't realize `Render` modifies the `RespnoseWriter` struct, allowing the caller to access the response code. We don't need to log anything here.

exp/services/webauth/internal/serve/challenge.go
howardtw: I'm gonna go with not trimming the http prefix until we have a feature request then.

exp/services/webauth/internal/serve/serve.go
howardtw: https://github.com/stellar/go/pull/2967/files

howardtw: https://github.com/stellar/go/pull/2967/commits/a91f60b75f66651f906ea16e4c848feba72d22c2

Issue #2965, opened 2020-08-28
Create a README file for exp/services/captivecore


bartekn:
We need a short README file with config options so that users know to use it.

PR #2971, opened 2020-08-31
exp/services/captivecore: Add README for captivecore server

tamirms:
Close https://github.com/stellar/go/issues/2965

This commit adds a README for the captive core server which details its API and its command line flags.

### DIFF
--- a/exp/services/captivecore/README.md
+++ b/exp/services/captivecore/README.md
@@ -19,7 +19,7 @@ multiple Horizon instances.
 
 ## API
 
-### GET /latest-sequence
+### `GET /latest-sequence`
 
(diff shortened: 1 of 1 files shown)


exp/services/captivecore/README.md
```diff
@@ -0,0 +1,96 @@
...
+Response:
+
+```json
+{
+	"sequence": 12345
+}
+```
+
+
+### GET /ledger/<sequence>
```
bartekn, line 35: `<sequence>` is not visible when rendered in HTML. Maybe we should change all sections to:
```suggestion
### `GET /ledger/<sequence>`
```

bartekn (approved):

PR #2976, opened 2020-09-03
services/horizon: Fix small issues related to Protocol 14 API

bartekn:
Fixed some minor issues I found while working on Protocol 14 changes list:
* `ClaimableBalance.Sponsor` can be `null`,
* Added missing operation names,
* Fixed `ClaimableBalancesQuery.URITemplate`.


tamirms (approved):

protocols/horizon/main.go
```diff
@@ -669,7 +669,7 @@ type ClaimableBalance struct {
 	BalanceID          string     `json:"id"`
 	Asset              string     `json:"asset"`
 	Amount             string     `json:"amount"`
-	Sponsor            string     `json:"sponsor"`
+	Sponsor            *string    `json:"sponsor"`
```
abuiles, line 672: @bartekn The reason why I set this always as a string, was to avoid returning `null` and return an empty string if the sponsor is not present - this makes it easier for SDKs since they don't have to worry about this value being `null` or a `string`. It's always a string, but empty if there is no sponsor.

PR #2985, opened 2020-09-09
services/horizon: (Protocol 14) Fix nil dereference on EndSponsorship details/participants

2opremio:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

The new Protocol-14 code for the EndSponsorship details/participants incorrectly assumed that the matching, initial BeginSponsorhip operation had a non-nil source. This PR fixes that.

### Why

We don't like panics :)

### Known limitations

N/A

### DIFF
--- a/services/horizon/internal/expingest/processors/operations_processor.go
+++ b/services/horizon/internal/expingest/processors/operations_processor.go
@@ -343,8 +343,9 @@ func (operation *transactionOperationWrapper) Details() (map[string]interface{},
 		if beginSponsoringOp == nil {
 			return nil, fmt.Errorf("prior Begin operation not found for EndSponsoringFutureReserves")
 		}
-		beginSponsor := beginSponsoringOp.SourceAccount.ToAccountId()
-		details["begin_sponsor"] = beginSponsor.Address()
+		beginSponsorWrapper := *operation
+		beginSponsorWrapper.operation = *beginSponsoringOp
+		details["begin_sponsor"] = beginSponsorWrapper.SourceAccount().Address()
 	case xdr.OperationTypeRevokeSponsorship:
 		op := operation.operation.Body.MustRevokeSponsorshipOp()
 		switch op.Type {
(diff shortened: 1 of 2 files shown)


services/horizon/internal/expingest/processors/operations_processor.go
```diff
@@ -343,8 +343,9 @@ func (operation *transactionOperationWrapper) Details() (map[string]interface{},
 		if beginSponsoringOp == nil {
 			return nil, fmt.Errorf("prior Begin operation not found for EndSponsoringFutureReserves")
 		}
-		beginSponsor := beginSponsoringOp.SourceAccount.ToAccountId()
-		details["begin_sponsor"] = beginSponsor.Address()
+		beginSponsorWrapper := *operation
+		beginSponsorWrapper.operation = *beginSponsoringOp
```
bartekn, line 347: While the code is safe now it's easy to forget that ex. `beginSponsorWrapper.index` is invalid. Maybe instead of this we could return `*transactionOperationWrapper` in `findPriorBeginSponsoringOp`?

2opremio: Good idea, I will do that.

2opremio:
@bartekn PTAL

bartekn (approved):

services/horizon/internal/expingest/processors/operations_processor.go
```diff
@@ -339,12 +342,11 @@ func (operation *transactionOperationWrapper) Details() (map[string]interface{},
 		op := operation.operation.Body.MustBeginSponsoringFutureReservesOp()
 		details["sponsored_id"] = op.SponsoredId.Address()
 	case xdr.OperationTypeEndSponsoringFutureReserves:
-		beginSponsoringOp := operation.findPriorBeginSponsoringOp()
-		if beginSponsoringOp == nil {
+		beginSponsorshipOp := operation.findPriorBeginSponsoringOp()
+		if beginSponsorshipOp == nil {
```
abuiles, line 346: @2opremio we should probably error only if the transaction is successful. Otherwise, we still want to ingest the operation but the only detail we have would be the `source_account`, not `being_sponsor`.

services/horizon/internal/expingest/processors/operations_processor.go
```diff
@@ -339,12 +342,11 @@ func (operation *transactionOperationWrapper) Details() (map[string]interface{},
 		op := operation.operation.Body.MustBeginSponsoringFutureReservesOp()
 		details["sponsored_id"] = op.SponsoredId.Address()
 	case xdr.OperationTypeEndSponsoringFutureReserves:
-		beginSponsoringOp := operation.findPriorBeginSponsoringOp()
-		if beginSponsoringOp == nil {
+		beginSponsorshipOp := operation.findPriorBeginSponsoringOp()
```
abuiles, line 345: @2opremio we should probably error only if the transaction is successful. Otherwise, we still want to ingest the operation but the only detail we have would be the source_account, not being_sponsor.

abuiles (changes requested):

services/horizon/internal/expingest/processors/operations_processor.go
2opremio: I will do that in a separate PR.

Issue #2987, opened 2020-09-09
Ingestion fails on a transaction with a single EndSponsoringFutureReserves operation


bartekn:
### What version are you using?

`protocol-14` branch.

### What did you do?

Submit a single EndSponsoringFutureReserves (it will fail but will be included in a ledger):
```go
	tx, err := txnbuild.NewTransaction(
		txnbuild.TransactionParams{
			SourceAccount: source,
			Operations: []txnbuild.Operation{
				&txnbuild.EndSponsoringFutureReserves{
					SourceAccount: source,
				},
			},
			BaseFee:    txnbuild.MinBaseFee,
			Timebounds: txnbuild.NewInfiniteTimeout(),
		},
	)
```

### What did you expect to see?

No errors, operation ingested.

### What did you see instead?

```
Error running processors on ledger: Error streaming changes from ledger: could not process transaction 1: error in *processors.OperationProcessor.ProcessTransaction: Error obtaining details for operation 2353642082305: prior Begin operation not found for EndSponsoringFutureReserves
```

Remember that this makes `begin_sponsor` nullable and it needs to be changed in `protocols/horizon` and `services/horizon/resourceadapters`.

2opremio:
But ... isn't this a badly-formed transaction? What do we do in other situations like this?

bartekn:
@2opremio it's a good point. I think it's due to the fact that Stellar-Core doesn't check validity of entire transaction, it's check per operation only. Anyway, if Stellar-Core accepts such transaction we need to ingest it. Unless, this changes in Stellar-Core: @jonjove want to check it out?

jonjove:
@bartek is correct, stellar-core isn't validating transactions beyond the per-operation level. At some protocol meeting, we discussed whether stellar-core should reject bad pairings of begin/end sponsoring. The general consensus was that we didn't really need to do this, so we elected not to.

abuiles:
@2opremio for effects we skip ingesting 

https://github.com/stellar/go/blob/f691bc858bd9b85d4958bba0154af149b3b87529/services/horizon/internal/expingest/processors/effects_processor.go#L110-L114

However, we don't do this for operation details. I think we'll need to add some special handling around the following lines

https://github.com/stellar/go/blob/f691bc858bd9b85d4958bba0154af149b3b87529/services/horizon/internal/expingest/processors/operations_processor.go#L343-L345

[...]

PR #2994, opened 2020-09-10
services/horizon: Account for overlapping sandwiches in EndSponsoringFutureReserves

### LINKED ISSUE #2989: Invalid begin_sponsor when multiple sponsoring sandwiches in a transaction

`protocol-14` branch.

### What did you do?

Submit a transaction with multiple sponsorships:
```
BEGIN_SPONSORING_FUTURE_RESERVES (source: A, sponsored_id: B)
BEGIN_SPONSORING_FUTURE_RESERVES (source: C, sponsored_id: D)
END_SPONSORING_FUTURE_RESERVES (source: B)
END_SPONSORING_FUTURE_RESERVES (source: D)
```

`begin_sponsor` fields on both `EndSponsoringFutureReserves` operations are the same.

### What did you see instead?

`begin_sponsor` fields should be different (point to actual sponsor).

I think we need to check IDs here instead of finding the first Begin operation:
https://github.com/stellar/go/blob/9f2f2e66f128fde9e321d226438e3861f9c6b391/services/horizon/internal/expingest/processors/operations_processor.go#L140-L151

2opremio:
Closes #2989

### DIFF
--- a/services/horizon/internal/expingest/processors/operations_processor.go
+++ b/services/horizon/internal/expingest/processors/operations_processor.go
@@ -138,21 +138,17 @@ func (operation *transactionOperationWrapper) OperationResult() *xdr.OperationRe
 }
 
 func (operation *transactionOperationWrapper) findInitatingBeginSponsoringOp() *transactionOperationWrapper {
+	if !operation.transaction.Result.Successful() {
+		// Failed transactions may not have a compliant sandwich structure
+		// we can rely on (e.g. invalid nesting or a being operation with the wrong sponsoree ID)
+		// and thus we bail out since we could return incorrect information.
+		return nil
+	}
+	sponsoree := operation.SourceAccount()
 	operations := operation.transaction.Envelope.Operations()
-	// We need to take into account nested sponsoring sandwiches.
-	// To do so, we count the EndSponsoring operations that we encounter
-	sandwichNestingDepth := 0
 	for i := int(operation.index) - 1; i >= 0; i-- {
-		switch operations[i].Body.Type {
-		case xdr.OperationTypeEndSponsoringFutureReserves:
-			// we are entering a nested sandwich
-			sandwichNestingDepth++
-		case xdr.OperationTypeBeginSponsoringFutureReserves:
-			if sandwichNestingDepth > 0 {
-				// we are exiting a nested sandwich
-				sandwichNestingDepth--
-				break
-			}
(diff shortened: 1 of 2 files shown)


bartekn (commented):
I'm not sure this will work for all cases. I think both scenarios below are possible:
```
BEGIN_SPONSORING_FUTURE_RESERVES (source: A, sponsored_id: B)
BEGIN_SPONSORING_FUTURE_RESERVES (source: C, sponsored_id: D)
END_SPONSORING_FUTURE_RESERVES (source: B) <-- B first
END_SPONSORING_FUTURE_RESERVES (source: D)
```
```
BEGIN_SPONSORING_FUTURE_RESERVES (source: A, sponsored_id: B)
BEGIN_SPONSORING_FUTURE_RESERVES (source: C, sponsored_id: D)
END_SPONSORING_FUTURE_RESERVES (source: D) <-- D first
END_SPONSORING_FUTURE_RESERVES (source: B)
```
I think we should simply compare `BEGIN.sponsored_id` with `END.source` to find the matching operation.

2opremio:
It does work in both cases. Check it :) Or I can add a specific test if you want.

> I think we should simply compare BEGIN.sponsored_id with END.source to find a matching operation.a

This on the other hand won't work in all cases since the there can be multiple prior sandwiches with the same Begin.sponsored_id and different source.

2opremio:
> I think we should simply compare `BEGIN.sponsored_id` with `END.source` to find a matching operation.

Also, this assumes that the transaction is well formed (i.e. that the matching `BEGIN_SPONSORING_FUTURE_RESERVES` operation does in fact sponsor the right account) but from what we have seen at https://github.com/stellar/go/issues/2987 , this may not be the case.

2opremio:
Let's have a call about this. Fons fears infinite textual discussions 👹 😨

2opremio:
I had a call with @bartekn and we concluded that:
1. Indeed what @bartekn proposes (search by comparing `BEGIN.sponsored_id` with `END.source` ) is correct. I was wrongly assuming that sponsorship sandwiches had to be balanced and does my balanced-parenthesis approach doesn't.
2. What @bartekn proposes may not work if the transaction failed (e.g. like in the nesting example I show at https://github.com/stellar/go/pull/2994#issuecomment-690304101 ). We won't be gathering the begin operation sponsor when transactions fail.

2opremio:
@bartekn PTAL

abuiles (approved):

Issue #2939, opened 2020-08-21
Add integration tests for Protocol 14


abuiles:
Using the new integration tests infrastructure, we should add tests for

- GET `/claimable_balances`.
- GET `/claimable_balances/id`.
- GET `/claimable_balances` filtering by `sponsor`, `claimant`, and `asset`.

See #2916

bartekn:
Here are proposed scenarios for CAP-23, CAP-33 and a mix of CAP-23 and CAP-33:

CAP-23:

1. Submit `create_claimable_balance` with different predicates:
    1. Two successful: native and non-native asset.
    1. One with a predicate that will fail later when calling `claim_claimable_balance`.
1. Check `/claimable_balances` (filter by claimant, asset).
1. Check `/claimable_balances/{id}`.
1. Submit `claim_claimable_balance` for all entries (use claimable balance id returned by Horizon).
1. Check `/operations` if correctly ingested (also check `/operation?include_failed=true` to check failed ops due to invalid predicate).
1. Check `/effects` if correctly ingested.
1. State Verifier [1]

CAP-33:

1. Submit operations to create sponsored ledger entries (account, offer, signer, data, trustline).
1. Check `/accounts` (check signers, trustlines, data), `/offers`.
1. Update sponsorship entry.
1. Check updated sponsor in `/accounts` (check signers, trustlines, data), `/offers`.
1. Check `/operations` if correctly ingested.
1. Check `/effects` if correctly ingested.
1. State Verifier  [1]

Mix CAP-23 & CAP-33:

1. Create sponsored claimable_balance.
1. Check sponsor in `/claimable_balances`.
1. Check `/operations` if correctly ingested.
1. Check `/effects` if correctly ingested.

[1] For State Verification we should probably just check if it completed. You can either use `/metrics` or parse Horizon log.

@Shaptic we can discuss this after team meeting today.

tamirms:
We should add a CAP 33 test which covers the case where a sponsorship can be removed during one-time signer removal.

To trigger a one-time signer removal you need to prepare an account with a pre-auth signer (let’s say for the transaction with hash H) that is sponsored. 

Once transaction H is prepared, submit it.

Afterwards, the pre-auth signer with corresponding sponsorship should be removed.

Note, H can succeed or fail and the signer will be removed in either case.

Shaptic:
Tests for scenarios 1-4 under the CAP23 section above are implemented in the #2991 (draft) PR.

bartekn:
[...]

PR #2991, opened 2020-09-09
[WIP] services/horizon: Add integration tests for Claimable Balances.

Shaptic:
<details>
  <summary>PR Checklist</summary>
  
### PR Structure

* [x] This PR has reasonably narrow scope (if not, break it down into smaller PRs).
* [x] This PR avoids mixing refactoring changes with feature changes (split into two PRs
  otherwise).
* [x] This PR's title starts with name of package that is most changed in the PR, ex.
  `services/friendbot`, or `all` or `doc` if the changes are broad or impact many
  packages.

### Thoroughness

* [x] This PR adds tests for the most critical parts of the new functionality or fixes.
* [ ] I've updated any docs ([developer docs](https://www.stellar.org/developers/reference/), `.md`
  files, etc... affected by this change). Take a look in the `docs` folder for a given service,
  like [this one](https://github.com/stellar/go/tree/master/services/horizon/internal/docs).

### Release planning

* [ ] I've updated the relevant CHANGELOG ([here](services/horizon/CHANGELOG.md) for Horizon) if
  needed with deprecations, added features, breaking changes, and DB schema changes.
* [ ] I've decided if this PR requires a new major/minor version according to
  [semver](https://semver.org/), or if it's mainly a patch change. The PR is targeted at the next
  release branch if it's not a patch change.
</details>

### What
Adds some tests for `/claimable_balances` that should be easily extendable. From #2939, this covers:
  - [x] Creating a claimable balance w/ native and non-native assets.
  - [x] Filtering by claimant, sponsor, asset, and ID (`/claimable_balances/{id}`).

Crucially, I haven't actually tried **claiming** balances yet. This only confirms that submission and filtering works.

See #2916 for how to run the tests. TL;DR:
```bash
cd ./services/horizon
HORIZON_INTEGRATION_TESTS=true HORIZON_BIN_DIR=$(pwd) go test -v ./internal/integration/protocol14_test.go
```

### Why
We need tests for CAP23.

### Known limitations
I can't really recommend using `IntegrationConfig.SkipContainerCreation` yet; it's very finicky and I need to clean up the various paths that diverge based on container state.

PR #2998, opened 2020-09-11
services/horizon: Add state verifier integration tests connected to Protocol 14

bartekn:
### What

Adds a new integration tests checking if state is correctly verified when ledger entries have sponsors and a new entry type: `ClaimableBalance`. It adds ledger entries and then gets `:[admin_port]/metrics` to check if state verification is finished and state is correct.

Part of #2939.

While working on this I fixed some bugs:
* `/accounts/{id}/data/{name}` data value was incorrectly rendered (overwritten by a sponsor),
* Sponsor missing in account's balances.
* `accountDataBatchInsertBuilder` not adding sponsor,
* `GetClaimableBalancesByID` didn't work at all (invalid query).

### Why

We want to ensure that a new code added to state verifier is working correctly.

### Known limitations

* Currently it's not possible to speed ledger close times easily so test needs to wait until the first checkpoint ledger (63) is closed.
* No signer sponsorships added. Run out of time today.

### DIFF
--- a/services/horizon/internal/integration/protocol14_state_verifier_test.go
+++ b/services/horizon/internal/integration/protocol14_state_verifier_test.go
@@ -8,7 +8,6 @@ import (
 	"testing"
 	"time"
 
-	"github.com/stellar/go/clients/horizonclient"
 	"github.com/stellar/go/keypair"
 	"github.com/stellar/go/services/horizon/internal/test"
 	"github.com/stellar/go/services/horizon/internal/txnbuild"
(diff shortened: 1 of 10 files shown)


services/horizon/internal/integration/protocol14_state_verifier_test.go
```diff
@@ -0,0 +1,126 @@
+package integration
```
Shaptic, line 1: Bit of a nit, but I think we should make a `protocol14/` subfolder in the future after we get these tests merged so that we don't need filename prefixes.

bartekn: Good point. We can do it after #2991 and this PR are merged.

bartekn:
I noticed that there are some unrelated failures in staticcheck. I believe this is because I merged `master` earlier today and it contains https://github.com/stellar/go/commit/19c3edf06d4a911d6352fb95d6dda70d984f289d. I fixed most issues in https://github.com/stellar/go/pull/2998/commits/3b1d14d84f0437f9bbc3a2b1faaee913531c64a1 but I'm not sure what's the best way to solve issues with internal `txnbuild`. This is my last day before 2 weeks OOO so I won't be able to fix it - sorry!

PR #3002, opened 2020-09-14
exp/services/recoverysigner: call database transaction `.Rollback()` when returning an error

marcelosalloum:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Call `tx.Rollback()` when returning an error from `func (s *DBStore) Add(a Account) error`.

### Why

You need to call `tx.Rollback()` in order to abort the database transaction. Without calling it, the transactions may stack up indefinitely until they reach the max limit.

### Known limitations

We still need to review the other usages of database transactions to make sure it's being rolled back properly.

### DIFF
--- a/exp/services/recoverysigner/internal/account/db_store_add_test.go
+++ b/exp/services/recoverysigner/internal/account/db_store_add_test.go
@@ -161,3 +161,26 @@ func TestAdd_conflict(t *testing.T) {
 	err = store.Add(a)
 	assert.Equal(t, ErrAlreadyExists, err)
 }
+
+func TestAdd_conflictProperlyClosesTheConnections(t *testing.T) {
+	db := dbtest.Open(t)
+	session := db.Open()
+	session.SetMaxIdleConns(1)
+	session.SetMaxOpenConns(1)
+
+	store := DBStore{
+		DB: session,
+	}
+
+	a := Account{
+		Address: "GCLLT3VG4F6EZAHZEBKWBWV5JGVPCVIKUCGTY3QEOAIZU5IJGMWCT2TT",
+	}
+
+	err := store.Add(a)
+	require.NoError(t, err)
+
+	for range [5]int{} {
+		err = store.Add(a)
+		require.Equal(t, ErrAlreadyExists, err)
+	}
+}
--- a/exp/services/recoverysigner/internal/account/db_store_add.go
+++ b/exp/services/recoverysigner/internal/account/db_store_add.go
@@ -15,6 +15,7 @@ func (s *DBStore) Add(a Account) error {
 		RETURNING id
 	`, a.Address)
 	if err != nil {
+		tx.Rollback()
 		// 23505 is the PostgreSQL error for Unique Violation.
 		// See https://www.postgresql.org/docs/9.2/errcodes-appendix.html.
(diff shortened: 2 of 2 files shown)


marcelosalloum:
@cla-bot check

accordeiro:
@cla-bot check

accordeiro (approved):
:shipit:

exp/services/recoverysigner/internal/account/db_store_add_test.go
```diff
@@ -161,3 +161,26 @@ func TestAdd_conflict(t *testing.T) {
...
+	}
+
+	a := Account{
+		Address: "GCLLT3VG4F6EZAHZEBKWBWV5JGVPCVIKUCGTY3QEOAIZU5IJGMWCT2TT",
+	}
+
+	err := store.Add(a)
+	require.NoError(t, err)
+
+	for range [5]int{} {
```
howardtw, line 182: Is it the same as `for i := 0; i < 5; i++`? Never seen it before.

Also where do we verify that the connection has been properly closed?  Is it that if we failed to close a connection, we would receive an internal error as opposed to `ErrAlreadyExists`?

howardtw (commented):
Good catch!

exp/services/recoverysigner/internal/account/db_store_add_test.go
marcelosalloum: > Is it the same as `for i := 0; i < 5; i++`? Never seen it before.

Yes, it's the same. It can be used when you don't need the value of `i`.

> Also where do we verify that the connection has been properly closed? Is it that if we failed to close a connection, we would receive an internal error as opposed to ErrAlreadyExists?

I've set the connection limit to `1`, so if the connection is not closed through a `Rollback()` or `Commit()` the errors gets stuck and the test times out after 30 seconds.

howardtw: > Yes, it's the same. It can be used when you don't need the value of i

Awesome!

> I've set the connection limit to 1, so if the connection is not closed through a Rollback() or Commit() the errors gets stuck and the test times out after 30 seconds.

Do you mean the second call will be stuck at `s.DB.Beginx()` because there is no connection available since the first call didn't released the connection properly?

Would it be more readable if we do `for range [2]int{}`? And maybe add some comments around it?

marcelosalloum: > Do you mean the second call will be stuck at s.DB.Beginx() because there is no connection available since the first call didn't released the connection properly?

Exactly, the test runs [`err = store.Add(a)`](https://github.com/stellar/go/blob/8a62af0e62d7248817c53018d7e5f13557937676/exp/services/recoverysigner/internal/account/db_store_add_test.go#L183) but the `Add` method will get stuck at [`tx, err := s.DB.Beginx()`](https://github.com/stellar/go/blob/8a62af0e62d7248817c53018d7e5f13557937676/exp/services/recoverysigner/internal/account/db_store_add.go#L6).

I'll add the comments later.

marcelosalloum: #3014

PR #3005, opened 2020-09-14
Fix static check tests.

abuiles:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Fixes the CI by removing a bunch of unused code and adding an explicit skip in the places where we know the linter would fail even if the code is being used.

### Why

Make CircleCI green again.

### Known limitations

[TODO or N/A]


services/horizon/internal/test/integration.go
```diff
@@ -325,23 +325,6 @@ func (i *IntegrationTest) CreateAccounts(count int) ([]*keypair.Full, []txnbuild
 	return pairs, accounts
 }
 
-// Establishes a trustline for a given asset for a particular account.
-//
-// Note: The function panics if this account doesn't exist in the ledger yet, so
-// be sure to fund it before doing this.
-func (i *IntegrationTest) EstablishTrustline(
```
abuiles, line 332: I think @Shaptic will add later this in a different PR, but for now, since we don't use it, let's remove it.

services/horizon/internal/txnbuild/claim_claimable_balance.go
```diff
@@ -1,3 +1,4 @@
+//lint:file-ignore U1001 Ignore all unused code, staticcheck doesn't understand testify/suite
```
abuiles, line 1: In general we don't care about this internal `txnbuild` copy  - so i removed a bunch of unused files and file-ignore in the new ones. We'll use them as we add more integration tests, but it doesn't really matter since eventually we'll remove these files in favor of the actual txnbuild

services/horizon/internal/test/integration.go
Shaptic: Yep, this was used in the breaking claiming tests.

tamirms (approved):

PR #3008, opened 2020-09-15
services/friendbot: fix #2271 (send on closed channel error)

### LINKED ISSUE #2271: Odd 400 / 500 issues when attempting to refund funded accounts.
`"stellar-sdk": "^4.0.0"`

### What did you do?
Attempting to refund a funded account. Goofy I know but it surfaced some odd inconsistent error handling which should likely be addressed.

Screencast of the issue: http://tyler.link/QsvRGA

You can reproduce the issue yourself here:
https://hack-stellar.now.sh/

Just click "Generate Keypair" and then keep clicking "Fund Account"

Consistent 400 error
```json
{
  "type": "https://stellar.org/horizon-errors/bad_request",
  "title": "Bad Request",
  "status": 400,
  "detail": "createAccountAlreadyExist (AAAAAAAAAGT/////AAAAAQAAAAAAAAAA/////AAAAAA=)"
}
```

### What did you see instead?
Occasional inconsistent 500 error
```json
{
  "type": "https://stellar.org/horizon-errors/server_error",
  "title": "Internal Server Error",
  "status": 500,
  "detail": "An error occurred while processing this request.  This is usually due to a bug within the server software.  Trying this request again may succeed if the bug is transient, otherwise please report this issue to the issue tracker at: https://github.com/stellar/go/issues. Please include this response in your issue."
}
```

accordeiro:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This PR aims to fix #2271 – the issue provides some context on the error and how it could be perceived by a Friendbot user. I've written a test that consistently reproduced the underlying issue (a panic due to sending on closed channel), then performed the fix, which consists of changing the `Minion.Run()` implementation so that it immediately returns after sending any message to the `resultChan`. After adding the `return`s, the error stopped happening.

### Why

To fix a bug that generated inconsistent error responses from Friendbot.

### Known limitations
This fix leverages a method mocking pattern that I'm not sure is inline with Go best practices, but since it is already in use in the Friendbot codebase, I decided to continue using for standardization purposes.


services/friendbot/internal/minion.go
```diff
@@ -16,15 +16,18 @@ var ErrAccountExists error = errors.New(fmt.Sprintf("createAccountAlreadyExist (
...
+	BotAccount      txnbuild.Account
+	BotKeypair      *keypair.Full
+	Horizon         *horizonclient.Client
+	Network         string
+	StartingBalance string
+	BaseFee         int64
+
+	// Mockable functions
+	SubmitTransaction    func(minion *Minion, hclient *horizonclient.Client, tx string) (*hProtocol.Transaction, error)
+	CheckSequenceRefresh func(minion *Minion, hclient *horizonclient.Client) error
```
debnil, line 30: I'm also not certain this is a Go best practice. I believe you're supposed to use interfaces, and then implement the functions within that interface. Relevant StackOverflow [here](https://stackoverflow.com/questions/48179565/how-to-mock-struct-methods-in-golang).

With that said, since this pattern already exists with `SubmitTransaction`, consistency within the existing service warrants using this pattern in my book.

services/friendbot/internal/minion_test.go
```diff
@@ -0,0 +1,73 @@
...
+		StartingBalance:      "10000.00",
+		SubmitTransaction:    mockSubmitTransaction,
+		CheckSequenceRefresh: mockCheckSequenceRefresh,
+		BaseFee:              txnbuild.MinBaseFee,
+	}
+	fb := &Bot{Minions: []Minion{minion}}
+
+	recipientAddress := "GDJIN6W6PLTPKLLM57UW65ZH4BITUXUMYQHIMAZFYXF45PZVAWDBI77Z"
+
+	// Prior to the bug fix, the following should consistently trigger a panic
```
debnil, line 60: I love adding a unit test to repro and verify a fix. Awesome!

debnil (approved):
LGTM! Agreed this isn't best practice, but consistency is good.

services/friendbot/internal/minion.go
```diff
@@ -33,19 +36,21 @@ type Minion struct {
...
 // to pay that address and submits the result to the channel.
 func (minion *Minion) Run(destAddress string, resultChan chan SubmitResult) {
-	err := minion.checkSequenceRefresh(minion.Horizon)
+	err := minion.CheckSequenceRefresh(minion, minion.Horizon)
 	if err != nil {
 		resultChan <- SubmitResult{
 			maybeTransactionSuccess: nil,
 			maybeErr:                errors.Wrap(err, "checking minion seq"),
 		}
+		return
```
nikhilsaraf, line 45: nice catch!
wonder if this was falling through all this while and throwing more errors than were actually the case

services/friendbot/internal/minion.go
```diff
@@ -77,8 +82,9 @@ func SubmitTransaction(minion *Minion, hclient *horizonclient.Client, tx string)
 	return &result, nil
 }
 
-// Establishes the minion's initial sequence number, if needed.
-func (minion *Minion) checkSequenceRefresh(hclient *horizonclient.Client) error {
+// CheckSequenceRefresh establishes the minion's initial sequence number, if needed.
+// This should also be passed to the minion.
+func CheckSequenceRefresh(minion *Minion, hclient *horizonclient.Client) error {
```
nikhilsaraf, line 87: is there a reason why this is exported now vs. previously?
If we don't wish for it to be part of the minion API then better to leave it unexported

services/friendbot/internal/minion_test.go
```diff
@@ -0,0 +1,73 @@
...
+	var wg sync.WaitGroup
+	wg.Add(numTests)
+
+	for i := 0; i < numTests; i++ {
+		go func() {
+			fb.Pay(recipientAddress)
+			wg.Done()
+		}()
+	}
+	wg.Wait()
```
nikhilsaraf, line 72: do we want to do some more assertions here, for example, how many times we called `mockSubmitTransaction`?
since we are mocking out that function we could add a concurrent counter to it to ensure it is called numTests times if we are expecting `numTests` payments.

Also, something to think about: will the test ever get in a state where the call to fb.Pay() will fail and therefore wg.Done() will not get called, which will result in the test timing out?

nikhilsaraf (approved):
Looks good -- added some thoughts as comments inline.

services/friendbot/internal/minion.go
accordeiro: Yeah, I suspect this was causing a lot of unexpected 500s.

services/friendbot/internal/minion.go
accordeiro: Agreed - not happy with it either but I wanted to avoid a big refactor at this point.

services/friendbot/internal/minion.go
accordeiro: That's a good point – I don't think it should be exported either, but I needed to make this change so that we'd be able to pass that function to the friendbot initializer (see [here](https://github.com/stellar/go/pull/3008/commits/8765485dcf5369f6d646335dac3f8400e9b3d0e1)), and also enable us to mock the function in the tests. 

I think the correct way to approach this would be to define an interface for the Minion and use a different implementation of it in the tests, but this would require a bigger refactor that I'm not sure would be worth doing right now. If you feel strongly that this needs to be addressed, I can file a separate tech debt issue – what do you think?

services/friendbot/internal/minion_test.go
accordeiro: I like the idea of adding a counter to ensure we call mockSubmitTransaction `numTest` times. Since I'd like to leave this test focused on solely reproducing the issue that was happening before, I've added a new test focused on this suggestion – see below :)

services/friendbot/internal/minion.go
nikhilsaraf: Yeah, that makes sense, missed the initialization part. Same reason why the SubmitTransactions is exported. Thanks for point it out.

I don't feel too strongly about this, was just curious. We don't need to file a tech-debt issue for it since it's trivial.

nikhilsaraf (approved):
looks good, thanks! 👍  🚢

PR #3009, opened 2020-09-15
Override CircleCI builtin checkout command

jacekn:
### What

This PR overrides CircleCI built-in checkout command. The updated command uses https instead of ssh which removes need for deploy key.

### Why

When jobs are added CircleCI records user who performs this action and adds deploy key to the repository on user's behalf.
When the user is removed from the organization CircleCI removes the key which causes run errors.


.circleci/config.yml
```diff
@@ -5,6 +5,58 @@ version: 2.1
...
+  checkout:
+    steps:
+      - run:
+          name: checkout
+          command: |
+            # Copy of the upstream checkout command with the following modification:
+            # 1. CIRCLE_REPOSITORY_URL is updated to use https rather than ssh
+            # 2. Removed ssh specific sections
+
+            # Use https rather that ssh to clone public projects
```
jcooter, line 17: s/that/than/

jcooter (approved):
LGTM, though there is a typo in a comment.

2opremio:
LGTM, the only risk I see is it could potentially break release jobs (but the don't seem to make any assumptions on the branch/checkout). @tamirms would you mind taking a quick loook?

.circleci/config.yml
```diff
@@ -5,6 +5,58 @@ version: 2.1
...
+      - run:
+          name: checkout
+          command: |
+            # Copy of the upstream checkout command with the following modification:
+            # 1. CIRCLE_REPOSITORY_URL is updated to use https rather than ssh
+            # 2. Removed ssh specific sections
+
+            # Use https rather than ssh to clone public projects
+            CIRCLE_REPOSITORY_URL=${CIRCLE_REPOSITORY_URL/://}
+            CIRCLE_REPOSITORY_URL=${CIRCLE_REPOSITORY_URL/git@/https://}
```
2opremio, line 19: AFAIU this won't propagate to other jobs (I doubt it will break anything, but just in case).

.circleci/config.yml
```diff
@@ -5,6 +5,58 @@ version: 2.1
 #----------------------------------------------------------------------------#
 
 commands:
+  checkout:
+    steps:
+      - run:
+          name: checkout
```
2opremio, line 11: Just to confirm, this overrides the `checkout` step of all the jobs defined in this file, right?

2opremio (approved):

.circleci/config.yml
jacekn: Yes this is the intention and as far as I could test that's what it did.

.circleci/config.yml
jacekn: Do you mean jobs in other repositories? If so then unfortunately we'll have to solve this on case by case basis.

2opremio: No, I mean other CircleCI jobs in the same repo

PR #3016, opened 2020-09-15
services/horizon: Add integration test for invalid claimable balance operation.

abuiles:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add a failing claimable balance and check the operation details and effects.

### Why

This help us test the scenario where a claimable balance operation fails and then:

1. Verify that the operation is ingested.
2. Check that no effects are produced.

### DIFF
--- a/services/horizon/internal/integration/protocol14_claimable_balance_ops_test.go
+++ b/services/horizon/internal/integration/protocol14_claimable_balance_ops_test.go
@@ -66,3 +66,42 @@ func TestCreateClaimableBalanceSuccessfulOperationsEffects(t *testing.T) {
 
 	tt.Equal("claimable_balance_sponsorship_created", effects[3].GetType())
 }
+
+func TestCreateClaimableBalanceInvalidOperationsEffects(t *testing.T) {
+	tt := assert.New(t)
+	itest := test.NewIntegrationTest(t, protocol14Config)
+	defer itest.Close()
+	master := itest.Master()
+
+	keys, accounts := itest.CreateAccounts(2, "50")
+	op := txnbuild.CreateClaimableBalance{
+		Destinations: []txnbuild.Claimant{
+			txnbuild.NewClaimant(master.Address(), nil),
+			txnbuild.NewClaimant(keys[1].Address(), nil),
+		},
+		Amount: "100",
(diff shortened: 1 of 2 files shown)


services/horizon/internal/test/integration.go
```diff
@@ -387,10 +387,10 @@ func (i *IntegrationTest) SubmitMultiSigOperations(
 
 	txResp, err = i.Client().SubmitTransactionXDR(txb64)
 	if err != nil {
-		i.t.Errorf("Submitting the transaction failed: %s\n", txb64)
+		i.t.Logf("Submitting the transaction failed: %s\n", txb64)
```
abuiles, line 390: We should use LogF so out test doesn't fail. It should be the caller responsibility to decide what do they want to do with the err (since you might be making an invalid tx on purpose)

Shaptic: Hah! I realized the same thing: https://github.com/Shaptic/go/commit/d8130de2e2918932ec5f742f13a92d4754d65335

services/horizon/internal/integration/protocol14_claimable_balance_ops_test.go
```diff
@@ -66,3 +66,42 @@ func TestCreateClaimableBalanceSuccessfulOperationsEffects(t *testing.T) {
...
 	tt.Equal("claimable_balance_sponsorship_created", effects[3].GetType())
 }
+
+func TestCreateClaimableBalanceInvalidOperationsEffects(t *testing.T) {
+	tt := assert.New(t)
+	itest := test.NewIntegrationTest(t, protocol14Config)
+	defer itest.Close()
+	master := itest.Master()
+
+	keys, accounts := itest.CreateAccounts(2, "50")
```
Shaptic, line 76: Can we just add a brief comment about _why_ this should fail? We both know from the CAP, but it's a pretty subtle case for someone casually reading the code.

Shaptic (approved):
:+1: just one small note

services/horizon/internal/integration/protocol14_claimable_balance_ops_test.go
abuiles: done :)

PR #3019, opened 2020-09-16
Fix dump-ledger-state Dockerfile

2opremio:
Fixes CircleCI in master.

Installing Core led to the following error:

```
The following packages have unmet dependencies:
 stellar-core : Depends: libc++1-8 but it is not installable
                Depends: libc++abi1-8 but it is not installable
```

It seems like the `golang:*buster` images no longer meet the dependencies
of Core (and Buster is the latest Debian release).

I could use a separate stage for building ... except the scripts use `go run`.

Thus, I just changed the base image to `ubuntu:18.04` and installed Go manually.


tamirms (approved):
👏

PR #3023, opened 2020-09-16
services/horizon: Check more details when we test claiming a claimable balance.

Shaptic:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
This is a simple addition to existing tests. 
It adds checks to the "claiming a claimable balance" tests to ensure that:
  - a random account for whom a CB is not destined cannot claim the CB
  - the actual balance of an account updates after claiming a CB

### Why
The former is an important "unhappy path" check for security, and the latter is a "fuller" integration test.

### DIFF
--- a/services/horizon/internal/integration/protocol14_test.go
+++ b/services/horizon/internal/integration/protocol14_test.go
@@ -139,20 +139,16 @@ func runClaimingCBsTest(t *testing.T, assetType txnbuild.AssetType, predicate *x
 	assert.Equal(t, sender.Address(), claim.Sponsor)
 	assert.Equal(t, "42.0000000", claim.Amount)
 
-	// Claiming a balance when you aren't the recipient should fail.
+	// Claiming a balance when you aren't the recipient should fail...
 	t.Logf("Stealing balance (ID=%s)...", claim.BalanceID)
 	op2 := txnbuild.ClaimClaimableBalance{BalanceID: claim.BalanceID}
-
 	_, err = itest.SubmitOperations(aAccount, adversary, &op2)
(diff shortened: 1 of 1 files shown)


services/horizon/internal/integration/protocol14_test.go
```diff
@@ -139,20 +139,53 @@ func runClaimingCBsTest(t *testing.T, assetType txnbuild.AssetType, predicate *x
...
+	// On success, we want the actual account to have a higher balance, too!
+	request := sdk.AccountRequest{AccountID: recipient.Address()}
+	details, err := client.AccountDetail(request)
+	assert.NoError(t, err)
+
+	// Ensure that the user's new balance includes the claimed balance, sans the
+	// transaction fee if applicable (this'll need to be updated if we change fees).
+	foundBalance := false
+	t.Logf("  balances: %+v", details.Balances)
+	for _, balance := range details.Balances {
```
abuiles, line 171: While this works, it seems to me like this helpers are trying to abstract way too much, making it a bit hard to follow. Case in point, the checks below now are fixed to some given amount depending if you are sending a native asset or not  - removing the "configurability" of the helpers.

abuiles (approved):
ALGTM 👍

services/horizon/internal/integration/protocol14_test.go
Shaptic: I completely agree, it's painful to work around these, yet also painful to repeat 80% of the code for both cases. My larger PR (pending) should resolve this and make the helpers more like LEGOs bricks than a cinder block.

PR #3032, opened 2020-09-17
Override CircleCI builtin checkout command

jacekn:
### What

This PR overrides CircleCI built-in checkout command. The updated command uses https instead of ssh which removes need for deploy key.

This is an improved code from #3009 which was reverted in #3030 due to a problem

### Why

When jobs are added CircleCI records user who performs this action and adds deploy key to the repository on user's behalf.
When the user is removed from the organization CircleCI removes the key which causes run errors.


brahman81 (approved):
LGTM

leighmcculloch (commented):
Will we need to add this command to all our repos that use CircleCI?

PR #3036, opened 2020-09-17
services/horizon: Implement support for RevokeSponsorship in txnbuild

### LINKED ISSUE #3001: How to build RevokeSponsorshipOp on txbuild?
`RevokeSponsorshipOp` requires users to build a `LedgerKey` or a struct for signers sponsorships. We need to expose a helper or helpers to make it easy for users to create such operation.

A possible approach could be to add functions for each ledger key. For example:

```
func RevokeTrustLineSponsorshipOp(account string, asset txnbuild.Asset)
func RevokeAccountSponsorshipOp(account string)
func RevokeSignerSponsorshipOp(account string, signer string) // optionally it could receive a txnbuild.Account instead of a string
...
```

2opremio:
Fixes #3001 .

It also fixes a small bug in `BeginSponsoringFutureReserves`

I am not too happy using pointers to represent a union type. If we only needed to build XDR operations, an interface would be enough. But since users need to be able to parse XDR (and inspect the resulting structure) I don't see a much better approach.

* [Using an interface + type inspection](https://eli.thegreenplace.net/2018/go-and-algebraic-data-types/) is an option, but I don't think it's much better than representing the union type explicitly. On top of that, our users are probably used to the `xdr` package's approach which uses pointers.

* [Using a visitor pattern](https://making.pusher.com/alternatives-to-sum-types-in-go/) is safer, but unnecessarily complicated.

### DIFF
--- a/services/horizon/internal/txnbuild/revoke_sponsorship.go
+++ b/services/horizon/internal/txnbuild/revoke_sponsorship.go
@@ -1,0 +1,281 @@
+package txnbuild
+
+import (
+	"fmt"
+
+	"github.com/pkg/errors"
+
+	"github.com/stellar/go/xdr"
+)
+
+type RevokeSponsorshipType int
+
+const (
+	RevokeSponsorshipTypeAccount RevokeSponsorshipType = iota + 1
+	RevokeSponsorshipTypeTrustLine
+	RevokeSponsorshipTypeOffer
+	RevokeSponsorshipTypeData
+	RevokeSponsorshipTypeClaimableBalance
+	RevokeSponsorshipTypeSigner
+)
+
+// RevokeSponsorship is a union type representing a RevokeSponsorship Operation.
+// SponsorshipType stablishes which sponsorship is being revoked.
+// The other fields should be ignored.
+type RevokeSponsorship struct {
+	SourceAccount    Account
+	SponsorshipType  RevokeSponsorshipType
+	Account          *string
+	TrustLine        *TrustLineId
+	Offer            *OfferId
+	Data             *DataId
+	ClaimableBalance *ClaimableBalanceHash
+	Signer           *SignerId
+}
+
+type TrustLineId struct {
+	AccountAddress string
+	Asset          Asset
+}
+
+type OfferId struct {
+	SellerAccountAddress string
+	OfferId              int64
+}
+
+type DataId struct {
+	AccountAddress string
+	DataName       string
+}
+
+type ClaimableBalanceHash [32]byte
+
+type SignerId struct {
+	AccountAddress string
+	SignerAddress  string
+}
+
+func (r *RevokeSponsorship) BuildXDR() (xdr.Operation, error) {
+	xdrOp := xdr.RevokeSponsorshipOp{}
+	switch r.SponsorshipType {
+	case RevokeSponsorshipTypeAccount:
+		var key xdr.LedgerKeyAccount
+		if r.Account == nil {
+			return xdr.Operation{}, errors.New("Account can't be nil")
+		}
+		if err := key.AccountId.SetAddress(*r.Account); err != nil {
+			return xdr.Operation{}, errors.Wrap(err, "incorrect Account address")
+		}
+		xdrOp.Type = xdr.RevokeSponsorshipTypeRevokeSponsorshipLedgerEntry
+		xdrOp.LedgerKey = &xdr.LedgerKey{
+			Type:    xdr.LedgerEntryTypeAccount,
+			Account: &key,
+		}
+	case RevokeSponsorshipTypeTrustLine:
+		var key xdr.LedgerKeyTrustLine
+		if r.TrustLine == nil {
+			return xdr.Operation{}, errors.New("TrustLine can't be nil")
+		}
+		if err := key.AccountId.SetAddress(r.TrustLine.AccountAddress); err != nil {
+			return xdr.Operation{}, errors.Wrap(err, "incorrect Account address")
+		}
+		asset, err := r.TrustLine.Asset.ToXDR()
+		if err != nil {
+			return xdr.Operation{}, errors.Wrap(err, "incorrect TrustLine asset")
+		}
+		key.Asset = asset
+		xdrOp.Type = xdr.RevokeSponsorshipTypeRevokeSponsorshipLedgerEntry
+		xdrOp.LedgerKey = &xdr.LedgerKey{
+			Type:      xdr.LedgerEntryTypeTrustline,
+			TrustLine: &key,
+		}
+	case RevokeSponsorshipTypeOffer:
+		var key xdr.LedgerKeyOffer
+		if r.Offer == nil {
+			return xdr.Operation{}, errors.New("Offer can't be nil")
+		}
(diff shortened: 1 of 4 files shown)


services/horizon/internal/txnbuild/revoke_sponsorship_test.go
```diff
@@ -0,0 +1,102 @@
...
+			op: RevokeSponsorship{
+				SponsorshipType: RevokeSponsorshipTypeSigner,
+				Signer: &SignerId{
+					AccountAddress: accountAddress,
+					SignerAddress:  "XBU2RRGLXH3E5CQHTD3ODLDF2BWDCYUSSBLLZ5GNW7JXHDIYKXZWGTOG",
+				},
+			},
+		},
+	} {
+		t.Run(testcase.name, func(t *testing.T) {
```
2opremio, line 88: We should probably have a test like this for all other operations (if so, the `BeginSponsoringFutureReserves` bug wouldn't have happened)

services/horizon/internal/txnbuild/revoke_sponsorship.go
```diff
@@ -0,0 +1,281 @@
...
+// SponsorshipType stablishes which sponsorship is being revoked.
+// The other fields should be ignored.
+type RevokeSponsorship struct {
+	SourceAccount    Account
+	SponsorshipType  RevokeSponsorshipType
+	Account          *string
+	TrustLine        *TrustLineId
+	Offer            *OfferId
+	Data             *DataId
+	ClaimableBalance *ClaimableBalanceHash
```
abuiles, line 32: @2opremio the ledger key for claimable balance can be build with the ClaimableBalanceId -- so this should probably be string representing the ClaimableBalanceId (hex encoded)

And then we use `xdr.SafeUnmarshalHex()` to get the actual xdr struct

services/horizon/internal/txnbuild/revoke_sponsorship.go
```diff
@@ -0,0 +1,281 @@
...
+	Signer           *SignerId
+}
+
+type TrustLineId struct {
```
abuiles, line 36: instead of adding a new struct, we could use `txnbuild.CreditAsset` or `txnbuild.Asset`  which is what we use in the create trustline operation.

abuiles: Maybe the `CreditAsset` makes sense since `Asset` would allow `native`.

services/horizon/internal/txnbuild/revoke_sponsorship.go
```diff
@@ -0,0 +1,281 @@
...
+	case RevokeSponsorshipTypeData:
+		var key xdr.LedgerKeyData
+		if r.Data == nil {
+			return xdr.Operation{}, errors.New("Data can't be nil")
+		}
+		if err := key.AccountId.SetAddress(r.Data.AccountAddress); err != nil {
+			return xdr.Operation{}, errors.Wrap(err, "incorrect Account address")
+		}
+		// TODO: should we check the size?
+		key.DataName = xdr.String64(r.Data.DataName)
```
abuiles, line 115: I had a quick look it doesn't seem like we do in the other helpers -- however, I assume the xdr library will complain if you pass a bigger size.

abuiles (changes requested):
Looks good to me!  We need to change the `type ClaimableBalanceHash [32]byte` to the balance id (string) and we are good!

services/horizon/internal/txnbuild/revoke_sponsorship.go
2opremio: OK, good, I will review the TODO.

services/horizon/internal/txnbuild/revoke_sponsorship.go
2opremio: I am not sure about what you mean. are you referring to `xdr.ClaimableBalancerId`? Happy to use that directly, but at this point it just complicates the interface (`xdr.ClaimableBalancerId` is currently just a hash, with a  (as of now) redundant version field for backwards compatibility)

services/horizon/internal/txnbuild/revoke_sponsorship.go
2opremio: You mean, using the Asset issuer as the trustline address?

services/horizon/internal/txnbuild/revoke_sponsorship.go
abuiles: @2opremio I mean taking the balanceID as exposed by horizon. When you load a balance id, it has an ID which is something like `0000asdfasdf`, what I mean is receiving that ID instead of the Hash. 

Users should be able to build this operation with the data exposed by Horizon, with this version, you'll need to know how to go from Horizon's balanceID representation to ClaimableBalanceHash

services/horizon/internal/txnbuild/revoke_sponsorship.go
abuiles: nvm, we do need a new struct. 

So maybe the only change we need is the type for `Asset` to be `CreditAsset` since you can't have a trustline to native.

services/horizon/internal/txnbuild/revoke_sponsorship.go
```diff
@@ -0,0 +1,281 @@
...
+}
+
+type TrustLineId struct {
+	AccountAddress string
+	Asset          Asset
+}
+
+type OfferId struct {
+	SellerAccountAddress string
+	OfferId              int64
```
abuiles, line 43: can you check how do we expose this value in Horizon? I think we return OfferID as a string, so this should probably be a string too.

services/horizon/internal/txnbuild/revoke_sponsorship.go
```diff
@@ -0,0 +1,281 @@
...
+	RevokeSponsorshipTypeSigner
+)
+
+// RevokeSponsorship is a union type representing a RevokeSponsorship Operation.
+// SponsorshipType stablishes which sponsorship is being revoked.
+// The other fields should be ignored.
+type RevokeSponsorship struct {
+	SourceAccount    Account
+	SponsorshipType  RevokeSponsorshipType
+	Account          *string
```
Shaptic, line 28: This is likely a misunderstanding of how Go works, but why are these pointer types if they can't be `nil` (I see lots of `nil` checks below)?

2opremio: Only one isn't nil (that's the way we implement the sum type)

services/horizon/internal/txnbuild/revoke_sponsorship.go
2opremio: I did, we return an `int64`:

https://github.com/stellar/go/blob/8d61d8e1d91c53e591eb2b673d54ad4614a87d5d/protocols/horizon/main.go#L237-L244

abuiles: @2opremio thanks for checking! So we return it as a string, but  protocol converts it to int64. All good :)

abuiles (approved):

PR #3038, opened 2020-09-18
services/horizon: Add Basic sponsorship integration test

2opremio:
Addresses part of #2939 

It also fixes a bug: the operation sponsor was being exposed in the Horizon responses.

It doesn't yet include `RevokeSponsorhip`, since it depends on https://github.com/stellar/go/pull/3036

### DIFF
--- a/services/horizon/internal/integration/protocol14_sponsorship_ops_test.go
+++ b/services/horizon/internal/integration/protocol14_sponsorship_ops_test.go
@@ -89,4 +89,17 @@ func TestSimpleSandwichHappyPath(t *testing.T) {
 	}
 	tt.Condition(endSponsorshipPresent)
 
+	// Check numSponsoring and numSponsored
+	account, err := itest.Client().AccountDetail(sdk.AccountRequest{
+		AccountID: sponsor.Address(),
+	})
+	tt.NoError(err)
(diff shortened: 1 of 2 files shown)


tamirms (approved):

abuiles (approved):
👍 

Can you add a test for the effects too? The state verifier test have a sandwich for each possible ledger entry, we can borrow than and check that the correct effects are generated.

services/horizon/internal/integration/protocol14_sponsorship_ops_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+	endSponsorshipPresent := func() bool {
+		for _, o := range response.Embedded.Records {
+			if o.GetID() == endSponsoringOp.ID {
+				return true
+			}
+		}
+		return false
+	}
+	tt.Condition(endSponsorshipPresent)
```
abuiles, line 90: can you also add a check for master account `numSponsoring` and for the sponsored account `numSponsored`

2opremio:
> 👍
> 
> Can you add a test for the effects too? The state verifier test have a sandwich for each possible ledger entry, we can borrow than and check that the correct effects are generated.

Yes, I will add that later.

PR #3043, opened 2020-09-21
protocols/horizon: fix trade aggregation paging token

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Change the paging token for trade aggregations (that isn't actually used) from being a single character that matches the unicode value of the timestamp, to being a string format of the timestamp integer.

For timestamp: `64`
Before: `"@"`
After: `"64"`

### Why
The paging token function isn't used for trade aggregations according to @tomerweller, but needed to be provided to satisfy the interface. The code there errors in Go 1.15 because Go 1.15 doesn't let you convert an integer to a string because in most cases, like the case here, it's a bug.

To compile on Go 1.15 we either need to change this to convert to a rune first to tell the compiler this is intentional, or we need to fix this to convert the integer to a string correctly. This PR does the latter.

### Known limitations

It's possible there is something dependent on this existing behavior, but I don't see it. Given the comment on the function saying it is not used (thank you @tomerweller!) I think we can assume there isn't any meaningful impact of changing this function.

FYI @stellar/horizon-committers

### DIFF
--- a/protocols/horizon/main.go
+++ b/protocols/horizon/main.go
@@ -416,7 +416,7 @@ type TradeAggregation struct {
 
 // PagingToken implementation for hal.Pageable. Not actually used
 func (res TradeAggregation) PagingToken() string {
-	return string(res.Timestamp)
+	return strconv.FormatInt(res.Timestamp, 10)
 }
 
 // Transaction represents a single, successful transaction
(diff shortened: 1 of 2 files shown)


tamirms (approved):

tomerweller (approved):
Argh. What kind of an idiot wrote that code.

leighmcculloch:
@tomerweller I think the fact that someone went out of their way to add this to the Go 1.15 compiler is evidence that many people were making this mistake and it was silly that the compiler allowed the cast. Thanks all!

PR #3048, opened 2020-09-22
txnbuild: Updates package with Protocol 14 changes.

Shaptic:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Basically, I ran the following commands (simplified)

 - `rsync -c -r -u internal/txnbuild ../../`
 - find & replace "/services/horizon/internal/txnbuild" with "/txnbuild"
 - `rm -rf internal/txnbuild`

All integration & unit tests pass.

### Why
We want to make our Protocol 14 updates visible.

### Known limitations
Because of the `-u` flag on the `rsync` command, I **only** copied new[er] files. This means that if someone updated txnbuild with a change after we updated the _same file_ inside of internal/txnbuild, **it wouldn't get copied.** I don't think this is the case anywhere and tried confirming with `diff`, but a sanity check would be wonderful.

Questions:  
 - Do we need to update the `CHANGELOG.md`? 
 - Or make version bumps anywhere


Shaptic:
Okay wait this is super incomplete, please don't review yet :sweat_smile:

abuiles (changes requested):
heads up, you are removing a bunch of valid files (they were removed from the internal txnbuild to make static check  happy, but we shouldn't be removing files).

2opremio:
@Shaptic If you want to be 100% sure, it would worth trying out https://stackoverflow.com/questions/1214906/how-do-i-merge-a-sub-directory-in-git

tamirms (approved):

abuiles (approved):

PR #3049, opened 2020-09-22
services/horizon: Fix variable shadowing in integration test tx submission

2opremio:


services/horizon/internal/test/integration.go
```diff
@@ -420,7 +420,7 @@ func (i *IntegrationTest) SubmitOperations(
 
 func (i *IntegrationTest) SubmitMultiSigOperations(
 	source txnbuild.Account, signers []*keypair.Full, ops ...txnbuild.Operation,
-) (txResp proto.Transaction, err error) {
```
2opremio, line 423: I don't know if it's widely accepted, but I personally consider named return variables to be pretty evil (because of shadowing and because they are implicit, which I think goes against Go's choice of being explicit with most things)

services/horizon/internal/test/integration.go
```diff
@@ -431,32 +431,31 @@ func (i *IntegrationTest) SubmitMultiSigOperations(
...
 	}
 
-	txResp, err = i.Client().SubmitTransactionXDR(txb64)
+	txResp, err := i.Client().SubmitTransactionXDR(txb64)
 	if err != nil {
 		i.t.Logf("Submitting the transaction failed: %s\n", txb64)
 		if prob := sdk.GetError(err); prob != nil {
 			i.t.Logf("Problem: %s\n", prob.Problem.Detail)
 			i.t.Logf("Extras: %s\n", prob.Problem.Extras["result_codes"])
 		}
```
2opremio, line 455: @Shaptic Unrelated to this PR (so please feel free to ignore) but I think the test creator should be the one examining the result.

(e.g. if you want to intentionally submit a failing transaction, the log will just add noise).

Shaptic: I totally agree, actually. I do think the code itself is helpful, though, so it should at least be a helper function like `failOnHorizonError` or something. I'm working on some minor cleanup to the framework in another branch and will incorporate this idea.

services/horizon/internal/test/integration.go
Shaptic: Totally fine by me :) I definitely found it odd seeing naked `return`s in functions w/ return types, but upon seeing it used in other places took it up as well. Happy to undo it, will defer to those with experience :wink:

Shaptic (approved):
:+1:

services/horizon/internal/test/integration.go
Shaptic: On second thought, do you think there would be value to doing something like 
```go
txResp := proto.Transaction{}
```
at the top of the function to avoid some of the verbosity in the return statements?

PR #3052, opened 2020-09-23
exp/services/recoverysigner: do not set max idle db connections

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Remove the setting of the max idle database connections that was added in 3ad0378.

### Why
To fix the issue that 3ad0378 was addressing we only needed to set the maximum number of total open connections. We set the max idle connections to the same value but sometimes keeping lots of idle connections open for long periods of time can cause issues if those idle connections get into a bad state. If a recoverysigner sees periods of low traffic its connections may sit idle for a long time. There's been no specific issue we've seen that is leading us to remove this, we just don't need it to be set to the same value.

We could also make the idle connections a parameter, and if anyone needs that we can add it.

If we see performance issues that require a higher max for idle connections we should address the issue then and can reintroduce it.

### Known limitations

There doesn't appear to be any tests reliant on the max number of idle connections available and therefore no tests to change.

### DIFF
--- a/exp/services/recoverysigner/internal/serve/serve.go
+++ b/exp/services/recoverysigner/internal/serve/serve.go
@@ -105,7 +105,6 @@ func getHandlerDeps(opts Options) (handlerDeps, error) {
 		return handlerDeps{}, errors.Wrap(err, "error parsing database url")
 	}
 	db.SetMaxOpenConns(opts.DatabaseMaxOpenConns)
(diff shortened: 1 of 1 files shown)


jacekn (approved):
LGTM!

howardtw (approved):

PR #3055, opened 2020-09-24
services/horizon: Improve predicate integration tests to run faster and do more validation.

Shaptic:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Massively improves predicate test performance.

Instead of a slow sub-test for each predicate type (2 tx each), we create a bunch of claims with different predicates in a **single** transaction, then claim them all in another **single** transaction. In other words, we went from 2n to just 2 transactions for this test (where n=5 for now) which is a big speedup.

Furthermore, the test now checks for:
  - tx success w/ right code
  - operations succeeding
  - predicate types matching
  - sponsor, assets, & amounts match
  - all claim successfully
  - balances update

### Why
Make tests more Horizon-focused.

### Known limitations
n/a


Shaptic:
@abuiles definitely would love feedback about if this is a better way to test the Horizon layer over core stuff :blush:

services/horizon/internal/integration/protocol14_test.go
```diff
@@ -433,6 +481,9 @@ func TestComplexPredicates(t *testing.T) {
 		_, err = itest.SubmitOperations(accountB, b,
 			&txnbuild.ClaimClaimableBalance{BalanceID: claim.BalanceID})
 		assert.Error(t, err)
+
+		assert.Equal(t, cantClaimResult, getOperationsError(err))
+		t.Logf("  tx did fail w/ %s", cantClaimResult)
```
tamirms, line 486: should this be in an if statement body?

Shaptic: Nope, it's *expected* to fail w/ this error

tamirms (approved):

PR #3060, opened 2020-09-25
Add data sponsoring integration test and fix Data effects

2opremio:
We were not parsing data effects properly (even if they were already being emitted) due to missing Data effect `struct`s. That [led us to think](https://github.com/stellar/go/pull/2951#issuecomment-699021498) we didn't have to add Data Sponsorship effects (for consistency).

This PR adds:
1. An integration test for Data sponsorships
2. Effects for Data sponsorships
3. Data structures (and parsing) of Data effects.

Part of #2939 , followup for #2951


abuiles (approved):
@2opremio good catch! Thanks.

PR #3061, opened 2020-09-25
support/http: add ReadTimeout, WriteTimeout, IdleTimeout configs

howardtw:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This PR adds the ability to configure ReadTimeout, WriteTimout, IdleTimeout on the http server.

### Why

We want to apply timeout to have a better control over our resources. A default http server comes with zero timeout. It's good that we have been setting the default `ReadTimeout` to 5 seconds. However, there was no way for the client of this package to set it to a different duration. And we are not currently setting `WriteTimeout` and `IdleTimeout`. 

### Known limitations

N/A

### DIFF
--- a/support/http/main.go
+++ b/support/http/main.go
@@ -22,19 +22,17 @@ import (
 // on a `Config` struct.
 const defaultListenAddr = "0.0.0.0:8080"
 
-const (
-	// defaultShutdownGracePeriod represents the default time in which the running
-	// process will allow outstanding http requests to complete before aborting
-	// them.  It will be used when a grace period of 0 is used, which normally
-	// signifies "no timeout" to our graceful shutdown package.  We choose not to
-	// provide a "no timeout" mode at present.  Feel free to set the value to a year
-	// or something if you need a timeout that is effectively "no timeout"; We
-	// believe that most servers should use a sane timeout and prefer one for the
-	// default configuration.
-	defaultShutdownGracePeriod = 10 * time.Second
+// defaultShutdownGracePeriod represents the default time in which the running
+// process will allow outstanding http requests to complete before aborting
+// them.  It will be used when a grace period of 0 is used, which normally
+// signifies "no timeout" to our graceful shutdown package.  We choose not to
+// provide a "no timeout" mode at present.  Feel free to set the value to a year
+// or something if you need a timeout that is effectively "no timeout"; We
+// believe that most servers should use a sane timeout and prefer one for the
+// default configuration.
+const defaultShutdownGracePeriod = 10 * time.Second
 
-	defaultReadTimeout = 5 * time.Second
-)
+const defaultReadTimeout = 5 * time.Second
 
 // SimpleHTTPClientInterface helps mocking http.Client in tests
 type SimpleHTTPClientInterface interface {
@@ -99,11 +97,11 @@ func setup(conf Config) *graceful.Server {
 		conf.ListenAddr = defaultListenAddr
 	}
 
-	if conf.ShutdownGracePeriod == 0 {
+	if conf.ShutdownGracePeriod == time.Duration(0) {
 		conf.ShutdownGracePeriod = defaultShutdownGracePeriod
 	}
 
-	if conf.ReadTimeout == 0 {
+	if conf.ReadTimeout == time.Duration(0) {
 		conf.ReadTimeout = defaultReadTimeout
 	}
 
--- a/support/http/main.go
+++ b/support/http/main.go
@@ -17,20 +17,24 @@ import (
 	"gopkg.in/tylerb/graceful.v1"
 )
 
-// DefaultListenAddr represents the default address and port on which a server
+// defaultListenAddr represents the default address and port on which a server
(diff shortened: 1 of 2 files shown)


support/http/main.go
```diff
@@ -17,20 +17,24 @@ import (
...
+	// them.  It will be used when a grace period of 0 is used, which normally
+	// signifies "no timeout" to our graceful shutdown package.  We choose not to
+	// provide a "no timeout" mode at present.  Feel free to set the value to a year
+	// or something if you need a timeout that is effectively "no timeout"; We
+	// believe that most servers should use a sane timeout and prefer one for the
+	// default configuration.
+	defaultShutdownGracePeriod = 10 * time.Second
+
+	defaultReadTimeout = 5 * time.Second
+)
```
leighmcculloch, line 37: These constants are relatively unrelated and I wouldn't group these. We don't group the listen address above with them.

```suggestion
// defaultShutdownGracePeriod represents the default time in which the running
// process will allow outstanding http requests to complete before aborting
// them.  It will be used when a grace period of 0 is used, which normally
// signifies "no timeout" to our graceful shutdown package.  We choose not to
// provide a "no timeout" mode at present.  Feel free to set the value to a year
// or something if you need a timeout that is effectively "no timeout"; We
// believe that most servers should use a sane timeout and prefer one for the
// default configuration.
const defaultShutdownGracePeriod = 10 * time.Second

const defaultReadTimeout = 5 * time.Second
```

support/http/main_test.go
```diff
@@ -21,6 +21,7 @@ func TestRun_setup(t *testing.T) {
 		Handler: stdhttp.NotFoundHandler(),
 	})
 
-	assert.Equal(t, DefaultShutdownGracePeriod, srv.Timeout)
-	assert.Equal(t, DefaultListenAddr, srv.Server.Addr)
+	assert.Equal(t, defaultShutdownGracePeriod, srv.Timeout)
+	assert.Equal(t, defaultReadTimeout, srv.ReadTimeout)
```
leighmcculloch, line 25: Can you add assertions for the write and idle timeouts too?

leighmcculloch (approved):
Looks good. I've mentioned a couple things inline I think you should change but they're small. I have some questions on where this is going:

1. The `IdleTimeout` of `http.Server` uses the value of the `ReadTimeout` field when not set, and I don't know of a need we have to configure it differently. Do we have a need to do that right now or in the future?

2. It doesn't look like this changes the behavior of any of our services because the `WriteTimeout` and `IdleTimeout` are still zero by default. Is that correct?

3. What services are you wanting to change the behavior of? This package is used in some but not all of our services, so if you're wanting to make these improvements across the board the others will need changing too.

howardtw:
> The IdleTimeout of http.Server uses the value of the ReadTimeout field when not set, and I don't know of a need we have to configure it differently. Do we have a need to do that right now or in the future?

[This](https://blog.gopheracademy.com/advent-2016/exposing-go-on-the-internet/) blog post suggests that it's beneficial to set the `IdleTimeout` to a value greater than the `ReadTimeout` for fast clients with Keep-Alive

```
Finally, Go 1.8 introduces IdleTimeout which limits server-side the amount of time a Keep-Alive connection will be kept idle before being reused. Before Go 1.8, the ReadTimeout would start ticking again immediately after a request completed, making it very hostile to Keep-Alive connections: the idle time would consume time the client should have been allowed to send the request, causing unexpected timeouts also for fast clients.
```

I've set it in our wallet backend server for this reason.

> It doesn't look like this changes the behavior of any of our services because the WriteTimeout and IdleTimeout are still zero by default. Is that correct?

That's correct. This is done intentionally.

support/http/main.go
howardtw: https://github.com/stellar/go/pull/3061/commits/5a468136124052ddc0034ca0ec4e0412c1f41474

support/http/main_test.go
howardtw: https://github.com/stellar/go/pull/3061/commits/5a468136124052ddc0034ca0ec4e0412c1f41474

support/http/main.go
```diff
@@ -89,21 +94,26 @@ func setup(conf Config) *graceful.Server {
...
+	}
+
+	if conf.ShutdownGracePeriod == time.Duration(0) {
+		conf.ShutdownGracePeriod = defaultShutdownGracePeriod
 	}
 
-	timeout := DefaultShutdownGracePeriod
-	if conf.ShutdownGracePeriod != 0 {
-		timeout = conf.ShutdownGracePeriod
+	if conf.ReadTimeout == time.Duration(0) {
```
leighmcculloch, line 104: Are these casts needed? I don't think so. You shouldn't need to cast a constant value like `0` to `time.Duration` because constants have no type.

leighmcculloch (approved):

support/http/main.go
howardtw: It's not needed per se. I simply did it because of readability. This is what's done in the test.

tamirms (approved):

PR #3072, opened 2020-09-29
services/horizon: Expand CAP 33 integration tests and fix 500 error bug

2opremio:
* Add Sponsorhip update checks
* Expand Claimable Balance test
* Add Signer test
* Fix type for RevokeSponsorship's OfferID field, which was causing a 500 error on `/operations` due to an unmarshalling error.

### DIFF
--- a/services/horizon/internal/integration/protocol14_sponsorship_ops_test.go
+++ b/services/horizon/internal/integration/protocol14_sponsorship_ops_test.go
@@ -21,7 +21,7 @@ func TestSponsoredAccount(t *testing.T) {
 	tt := assert.New(t)
 	itest := test.NewIntegrationTest(t, protocol14Config)
 	defer itest.Close()
-	sponsor := itest.MasterAccount()
+	sponsor := itest.MasterAccount
 	sponsorPair := itest.Master()
 
 	// We will create the following operation structure:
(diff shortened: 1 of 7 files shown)


services/horizon/internal/integration/protocol14_sponsorship_ops_test.go
```diff
@@ -153,20 +203,182 @@ func TestSponsoredAccount(t *testing.T) {
...
+		signers := newAccount().(*protocol.Account).Signers
+		for _, signer := range signers {
+			if signer.Key == signerKey {
+				tt.Equal(sponsorPair.Address(), signer.Sponsor)
+				return true
+			}
+		}
+		return false
+	}
+	tt.Eventually(signerAdded, time.Second*10, time.Millisecond*100)
```
bartekn, line 276: Nit: I think we don't need `Eventually`. Since the recent changes to txsub (https://github.com/stellar/go/pull/2877) when transaction result is returned all consecutive responses from Horizon will return state with that transaction included.

bartekn (approved):

services/horizon/internal/integration/protocol14_sponsorship_ops_test.go
2opremio: Ah, cool, I will create a ticket to refactor the tests.

tamirms (approved):

services/horizon/internal/integration/protocol14_sponsorship_ops_test.go
2opremio: Done #3075

PR #3078, opened 2020-09-30
Add friendbot Dockerfile and Makefile

jacekn:
### What
Add friendbot Dockerfile and Makefile

### Why
Copying the existing patterns we have in other applications. The
Dockerfile and Makefile were based on other services in the repository

### DIFF
--- a/services/friendbot/docker/Dockerfile
+++ b/services/friendbot/docker/Dockerfile
@@ -1,0 +1,13 @@
+FROM golang:1.14-stretch as build
+
+ADD . /src/friendbot
+WORKDIR /src/friendbot
+RUN go mod vendor
+RUN go build -o /bin/friendbot -mod=vendor ./services/friendbot
+
+FROM ubuntu:16.04
+
+RUN apt-get update && apt-get install -y --no-install-recommends ca-certificates
+COPY --from=build /bin/friendbot /app/
+EXPOSE 8004
+ENTRYPOINT ["/app/friendbot"]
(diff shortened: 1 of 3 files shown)


services/friendbot/docker/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM golang:1.14-stretch as build
+
+ADD . /src/friendbot
+WORKDIR /src/friendbot
+RUN go mod vendor
+RUN go build -o /bin/friendbot -mod=vendor ./services/friendbot
```
leighmcculloch, line 6: What's the goal of the `vendor`ing? Is it to cache dependencies? If so this needs some changing because right now any change to the stellar/go folder invalidates the vendor layer. Also I we should change it to the following which is the pattern of using the download command as it relies on the standard common modules behavior rather than on the less commonly used vendoring behavior.

```suggestion
WORKDIR /src/friendbot
ADD ./go.mod ./go.sum /src/friendbot
RUN go mod download
ADD . /src/friendbot
RUN go build -o /bin/friendbot ./services/friendbot
```

Alternatively, can we drop the caching/vendoring? We don't do it in the other stellar/go builds and it keeps the Dockerfile simpler. Add it later if the build is slow?

```suggestion
ADD . /src/friendbot
WORKDIR /src/friendbot
RUN go build -o /bin/friendbot ./services/friendbot
```

jacekn: @leighmcculloch thanks for your feedback. I removed vendoring, it actually made build faster.
It was there because I replicated what our current build system did but if we can improve things when moving to docker that's great.

leighmcculloch (approved):

Issue #3085, opened 2020-10-01
cursor in /claimable_balances not validated


bartekn:
### What version are you using?

1.9.0

### What did you do?

Request `/claimable_balances` with invalid `cursor` value. Ex: https://horizon-testnet.stellar.org/claimable_balances?cursor=1003529-00000043d380c38a2f2cac46ab63674064c56fdce6b977fdef1a278ad50e1a7e6a5e18 (hex-encoded balance id is malformed).

### What did you expect to see?

`400 Bad Request` response.

### What did you see instead?

`500 Internal Server Error` response.

abuiles:
Fixed in #3088

PR #3088, opened 2020-10-02
services/horizon: Wrap invalid cursor in claimable balances in problem.P

abuiles:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Wrap invalid cursor in claimable balances in `problem.P`

### Why

We do cursor validation, however, if there is an error we return a
plain error instead of a `problem.P` -- this causes horizon to handle
the error as a `500` instead of a `400`.

This fix moves cursor parsing to its own method and then use it in the
query handler to return a `problem.P` error which causes horizon to
return the correct response.

I opted to create a new error instead of returning the inner error
since it exposes low-level details which are not relevant for the API
user, for example:

- `Invalid cursor - first value should be higher than 0: strconv.ParseInt: parsing \"\": invalid syntax`
- `Invalid cursor - second value should be a valid claimable balance id: xdr:decode: switch '67' is not valid enum value for union`

Fixes https://github.com/stellar/go/issues/3085.

### Known limitations

### DIFF
--- a/services/horizon/internal/actions/claimable_balance.go
+++ b/services/horizon/internal/actions/claimable_balance.go
@@ -155,6 +155,14 @@ func (handler GetClaimableBalancesHandler) GetResourcePage(
 		Claimant:  qp.claimant(),
 	}
 
+	_, _, err = query.Cursor()
+	if err != nil {
+		return nil, problem.MakeInvalidFieldProblem(
+			"cursor",
+			errors.New("First part should be higher than 0 and second part should be valid claimable balance ID"),
+		)
+	}
+
 	historyQ, err := horizonContext.HistoryQFromRequest(r)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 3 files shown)


services/horizon/internal/actions/claimable_balance.go
```diff
@@ -155,9 +155,17 @@ func (handler GetClaimableBalancesHandler) GetResourcePage(
...
+			"cursor",
+			errors.New("First part should be higher than 0 and second part should be valid claimable balance ID"),
+		)
+	}
+
 	historyQ, err := horizonContext.HistoryQFromRequest(r)
 	if err != nil {
-		return nil, err
+		return nil, problem.MakeInvalidFieldProblem("query parameters", err)
 	}
```
abuiles, line 169: ignore this

tamirms (approved):

bartekn (approved):

PR #3033, opened 2020-09-17
services/horizon: Add integration tests for predicates on claimable balances. 

Shaptic:
<details>
  <summary>PR Checklist</summary>
  
### PR Structure

* [x] This PR has reasonably narrow scope (if not, break it down into smaller PRs).
* [x] This PR avoids mixing refactoring changes with feature changes (split into two PRs
  otherwise).
* [x] This PR's title starts with name of package that is most changed in the PR, ex.
  `services/friendbot`, or `all` or `doc` if the changes are broad or impact many
  packages.

### Thoroughness

* [x] This PR adds tests for the most critical parts of the new functionality or fixes.
* [ ] I've updated any docs ([developer docs](https://www.stellar.org/developers/reference/), `.md`
  files, etc... affected by this change). Take a look in the `docs` folder for a given service,
  like [this one](https://github.com/stellar/go/tree/master/services/horizon/internal/docs).

### Release planning

* [ ] I've updated the relevant CHANGELOG ([here](services/horizon/CHANGELOG.md) for Horizon) if
  needed with deprecations, added features, breaking changes, and DB schema changes.
* [ ] I've decided if this PR requires a new major/minor version according to
  [semver](https://semver.org/), or if it's mainly a patch change. The PR is targeted at the next
  release branch if it's not a patch change.
</details>

### What
This adds tests for predicates on claimable balances. 

It first checks the happy paths, ensuring that all predicate types can be created and that valid predicates result in actually claimable balances. 

It also checks that predicates actually matter: both an unconditionally-false predicate and an expiring timer cause the claimabe balance to be irretrievable.

### Why
Further testing of CAP-23 (see #2939) and specifically of #3000.

_(sorry about the weird commit history... should go away w/ squash & merge)_

PR #3090, opened 2020-10-02
services/horizon: Improve sponsorship integration tests' running time & flexibility.

### LINKED ISSUE #3075: Refactor sponsorship integration tests
* They are really verbose.
* They use `assert.Eventually` which apparently isn't necessary (Horizon seems to guarantee strict ordering after submitting a transaction).
* Part of the machinery could be moved into the integration-test infrastructure.
* Test could potentially be merged (they all follow a similar workflow). It may not be possible since different tests have different particularities (e.g. claimable balances cannot be explicitly revoked).

Shaptic:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Refactors the sponsorship (CAP-33) integration tests:
 - Each test is independently runs as a sub-test of the "TestSponsorships" test. This means only one container up/down!
 - It introduces helper functions commonly used in each tests such as sandwiching sponsored operations, retrieving operations and effects, and confirming the existence of operations via `tt.Condition`. This shortens the code verbosity and improves readability!
 - `itest.MustGetAccount()` replaces instances of a manual force-retrieval of an account, and account/keypair management in general is simpler.
 - `tt.Eventually` is replaced by `tt.Condition`

From what I can tell, all of the actual _testing_ is still the same. There *may* be a way to actually combine ALL of these tests into a single test w/ a handful of transactions (akin to what happened in #3033 with predicates), but this is still subject to investigation.

### Why
This **massively** improves both the running time and readability. 
It now takes **~1m30s** as opposed to ~4m20s previously.

This should close #3075.

### Known limitations
n/a


services/horizon/internal/integration/protocol14_sponsorship_ops_test.go
```diff
@@ -17,996 +18,760 @@ import (
...
-	signers := []*keypair.Full{sponsorPair, newAccountPair}
-	txResp, err := itest.SubmitMultiSigOperations(sponsor(), signers, ops...)
-	tt.NoError(err)
-
-	var txResult xdr.TransactionResult
-	err = xdr.SafeUnmarshalBase64(txResp.ResultXdr, &txResult)
-	tt.NoError(err)
-	tt.Equal(xdr.TransactionResultCodeTxSuccess, txResult.Result.Code)
+	t.Run("Data", func(t *testing.T) {
+		keys, accounts := itest.CreateAccounts(3, "1000")
```
bartekn, line 397: I think we can make this test even faster by using [`t.Parallel`](https://golang.org/pkg/testing/#T.Parallel). To make it working I think we need to wrap `CreateAccounts` in a mutex so that accounts are created in isolation (otherwise `tx_bad_seq` errors are likely). I'm not sure if other things will break but it looks like it's the only place.

bartekn (approved):

services/horizon/internal/integration/protocol14_sponsorship_ops_test.go
Shaptic: We need to be very careful, since some tests e.g. query `/operations` blindly and expect the first 3 ops to be related to the just-submitted transaction. Similarly, if you're checking account balances then they might be modified between `POST tx` and `GET /account/<id>`. Since we wouldn't be able to guarantee this kind of ordering anymore, queries will need to be more specific.

I want to be very sure about this, so I'll do it in a separate PR!

PR #3103, opened 2020-10-07
services/horizon: Add Go and process related metrics

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Adds Go and process related metrics provided by Prometheus package.

### Why

Extra metrics can be useful to debug performance issues.

### Known limitations

* Process metrics are available on operating systems with a Linux-style proc filesystem and on Microsoft Windows.

### DIFF
--- a/services/horizon/internal/init.go
+++ b/services/horizon/internal/init.go
@@ -215,6 +215,25 @@ func initDbMetrics(app *App) {
 	app.prometheusRegistry.MustRegister(app.orderBookStream.LatestLedgerGauge)
 }
 
+// initGoMetrics registers the Go collector provided by prometheus package which
+// includes Go-related metrics.
+func initGoMetrics(app *App) {
+	app.prometheusRegistry.MustRegister(prometheus.NewGoCollector())
+}
+
+// initProcessMetrics registers the process collector provided by prometheus
+// package. This is only available on operating systems with a Linux-style proc
+// filesystem and on Microsoft Windows.
+func initProcessMetrics(app *App) {
+	app.prometheusRegistry.MustRegister(
+		prometheus.NewProcessCollector(
+			prometheus.ProcessCollectorOpts{
+				Namespace: "horizon",
+			},
+		),
+	)
+}
+
 // initIngestMetrics registers the metrics for the ingestion into the provided
 // app's metrics registry.
 func initIngestMetrics(app *App) {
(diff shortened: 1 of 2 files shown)


services/horizon/internal/init.go
```diff
@@ -215,6 +215,25 @@ func initDbMetrics(app *App) {
 	app.prometheusRegistry.MustRegister(app.orderBookStream.LatestLedgerGauge)
 }
 
+// initGoMetrics registers the Go collector provided by prometheus package which
+// includes Go-related metrics.
+func initGoMetrics(app *App) {
+	app.prometheusRegistry.MustRegister(prometheus.NewGoCollector())
```
bartekn, line 221: @jacekn this will render metrics without `horizon_` prefix, like: `go_memstats_buck_hash_sys_bytes`. Is it OK?

services/horizon/internal/init.go
```diff
@@ -215,6 +215,25 @@ func initDbMetrics(app *App) {
...
+}
+
+// initProcessMetrics registers the process collector provided by prometheus
+// package. This is only available on operating systems with a Linux-style proc
+// filesystem and on Microsoft Windows.
+func initProcessMetrics(app *App) {
+	app.prometheusRegistry.MustRegister(
+		prometheus.NewProcessCollector(
+			prometheus.ProcessCollectorOpts{
+				Namespace: "horizon",
```
leighmcculloch, line 231: I'm not sure you want to use the namespace on the processor collector.  I think with the webauth and recoverysigner services we landed on only using namespaces for metrics that are truly uniquely defined by that service. General metrics should otherwise have no namespace so that the same metric that means the same thing has the same name.

Maybe this logic doesn't apply to Horizon's use case though.

services/horizon/internal/init.go
leighmcculloch: I think no namespace for the general metrics including the processor ones below is what you should go with. See my comment below.

services/horizon/internal/init.go
bartekn: Yeah, this makes sense. It's just not clear to me why setting a prefix is allowed here and not in `GoCollector`.

services/horizon/internal/init.go
jacekn: @bartekn yeah `go_memstats_buck_hash_sys_bytes` is absolutely fine, that's how most projects I know, including prometheus itself, export go metrics.

jacekn (approved):
LGTM!

PR #3105, opened 2020-10-07
txnbuild: Adds a helper function to determine asset type from string.

### LINKED ISSUE #2513: horizonclient: add a helper to get AssetType based on asset code (and issuer)
Many requests in horizonclient have one or more of `AssetType` fields, which can be derived from the length of asset code.

Chances are we might also need asset issuer as an input to be able to determine whether it is `native` asset or not. (Can we say that an asset is native if the asset code is `XLM` and the asset issuer is empty?)

In another project, I have to implement something like this
```
func assetType(asset string) horizonclient.AssetType {
	if asset == "native" {
		return horizonclient.AssetTypeNative
	}
	assetCode := strings.SplitN(asset, ":", 2)[0]
	if len(assetCode) <= 4 {
		return horizonclient.AssetType4
	}
	return horizonclient.AssetType12
}
```
where asset is in the form of [asset code]:[asset issuer]. It would be helpful if our library has something like this built in.

Shaptic:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Introduces `func ParseAssetString(string) txnbuild.Asset` and closes #2513.

### Why
Essentially, creating an asset from its string representation (and, specific to #2513, determining its type) should be easy and standardized. Refer to #2513 for more context. 

### Known limitations
The function uses `xdr.BuildAssets` internally, which states:
> Technically SEP-11 allows any string up to 12 characters not containing an unescaped colon to represent XLM 
> However, this function only accepts the string "native" to represent XLM.

meaning this technically isn't in full conformance of [SEP-11](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0011.md#asset).

### DIFF
--- a/txnbuild/CHANGELOG.md
+++ b/txnbuild/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this
 file.  This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
-* Expose a new helper function `GetAssetType()`, making it easier to determine an asset type from a string in [canonical form](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0011.md#asset) ([#3105](https://github.com/stellar/go/pull/3105)).
+* Expose a new helper function `ParseAssetString()`, making it easier to build an `Asset` structure from a string in [canonical form](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0011.md#asset) and check its various properties ([#3105](https://github.com/stellar/go/pull/3105)).
 
 ## [v4.0.1](https://github.com/stellar/go/releases/tag/horizonclient-v4.0.1) - 2020-10-02
 
--- a/txnbuild/CHANGELOG.md
+++ b/txnbuild/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this
 file.  This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
-* A new helper function `getAssetType()` has been introduced to make it easier to determine an asset type from an string in [canonical form](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0011.md#asset) ([insert PR number here](https://github.com/stellar/go/pull/xxxx)).
+* Expose a new helper function `GetAssetType()`, making it easier to determine an asset type from a string in [canonical form](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0011.md#asset) ([#3105](https://github.com/stellar/go/pull/3105)).
 
 ## [v4.0.1](https://github.com/stellar/go/releases/tag/horizonclient-v4.0.1) - 2020-10-02
 
--- a/txnbuild/CHANGELOG.md
+++ b/txnbuild/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this
 file.  This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
-* Expose a new helper function `ParseAssetString()`, making it easier to build an `Asset` structure from a string in [canonical form](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0011.md#asset) and check its various properties ([#3105](https://github.com/stellar/go/pull/3105)).
+* Add helper function `ParseAssetString()`, making it easier to build an `Asset` structure from a string in [canonical form](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0011.md#asset) and check its various properties ([#3105](https://github.com/stellar/go/pull/3105)).
 
 ## [v4.0.1](https://github.com/stellar/go/releases/tag/horizonclient-v4.0.1) - 2020-10-02
 
--- a/txnbuild/CHANGELOG.md
+++ b/txnbuild/CHANGELOG.md
@@ -3,6 +3,9 @@
 All notable changes to this project will be documented in this
 file.  This project adheres to [Semantic Versioning](http://semver.org/).
 
+## Unreleased
+* A new helper function `getAssetType()` has been introduced to make it easier to determine an asset type from an string in [canonical form](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0011.md#asset) ([insert PR number here](https://github.com/stellar/go/pull/xxxx)).
+
 ## [v4.0.1](https://github.com/stellar/go/releases/tag/horizonclient-v4.0.1) - 2020-10-02
 
 * Fixed bug in `TransactionFromXDR()` which occurs when parsing transaction XDR envelopes which contain Protocol 14 operations.
--- a/txnbuild/helpers.go
+++ b/txnbuild/helpers.go
@@ -193,31 +193,19 @@ func NewValidationError(field, message string) *ValidationError {
 	}
 }
 
-// Determines the type of an asset string in canonical form (SEP-11).
-//
-// Example:
-//     getAssetType("native") -> txnbuild.AssetTypeNative
-//     getAssetType("ABCD:G0000..[rest of issuer]") -> AssetTypeCreditAlphanum4
-//     getAssetType("ABCD1234EFGH:G0000..[rest of issuer]") -> AssetTypeCreditAlphanum12
-//
-//lint:file-ignore U1000 Created as a helper in response to user feedback.
-func GetAssetType(canonical string) (AssetType, error) {
-	defaultType := AssetTypeNative
+// Parses an asset string in canonical form (SEP-11) into an Asset structure.
+// https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0011.md#asset
+func ParseAssetString(canonical string) (Asset, error) {
 	assets, err := xdr.BuildAssets(canonical)
 	if err != nil {
-		return defaultType, errors.Wrap(err, "failed parsing: is it in canonical (SEP-11) form?")
+		return &NativeAsset{}, errors.Wrap(err, "failed parsing: is it in canonical (SEP-11) form?")
 	}
 
-	// It returns a list, so you'll need to grab the first element.
+	// It returns a list, so we'll need to grab the first element.
 	asset, err := assetFromXDR(assets[0])
 	if err != nil {
-		return defaultType, errors.Wrap(err, "failed to create Asset from XDR")
+		return &NativeAsset{}, errors.Wrap(err, "failed to create Asset from XDR")
 	}
 
-	t, err := asset.GetType()
-	if err != nil {
-		return t, errors.Wrap(err, "failed to determine asset type")
-	}
-
-	return t, nil
+	return asset, nil
 }
--- a/txnbuild/helpers.go
+++ b/txnbuild/helpers.go
@@ -200,7 +200,8 @@ func NewValidationError(field, message string) *ValidationError {
 //     getAssetType("ABCD:G0000..[rest of issuer]") -> AssetTypeCreditAlphanum4
 //     getAssetType("ABCD1234EFGH:G0000..[rest of issuer]") -> AssetTypeCreditAlphanum12
 //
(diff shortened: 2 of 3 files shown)


txnbuild/helpers.go
```diff
@@ -192,3 +192,32 @@ func NewValidationError(field, message string) *ValidationError {
...
+
+// Determines the type of an asset string in canonical form (SEP-11).
+//
+// Example:
+//     getAssetType("native") -> txnbuild.AssetTypeNative
+//     getAssetType("ABCD:G0000..[rest of issuer]") -> AssetTypeCreditAlphanum4
+//     getAssetType("ABCD1234EFGH:G0000..[rest of issuer]") -> AssetTypeCreditAlphanum12
+//
+//lint:file-ignore U1000 Created as a helper in response to user feedback.
+func GetAssetType(canonical string) (AssetType, error) {
```
leighmcculloch, line 204: I think it might be more valuable and round out the API if we have a function ParseAssetString() that returns a type that contains the asset type, asset code, and asset issuer. I believe the function would do much of what this function already does, but it will serve a more basic purpose. What do you think?

Shaptic: You're absolutely right, and judging by my [own comment](https://github.com/stellar/go/issues/2513#issuecomment-689911244) on the issue I had the same line of reasoning. I think I just had tunnel vision developing this.

I will amend to something like
```go
func ParseAssetString(canonical string) (Asset, err) { ...
```

txnbuild/helpers.go
```diff
@@ -200,7 +200,8 @@ func NewValidationError(field, message string) *ValidationError {
 //     getAssetType("ABCD:G0000..[rest of issuer]") -> AssetTypeCreditAlphanum4
 //     getAssetType("ABCD1234EFGH:G0000..[rest of issuer]") -> AssetTypeCreditAlphanum12
 //
-func getAssetType(canonical string) (AssetType, error) {
+//lint:file-ignore U1000 Created as a helper in response to user feedback.
```
abuiles, line 203: @Shaptic why do we have to ignore this helper but not  other functions? what happens if you use this in the demo ?

abuiles (approved):

txnbuild/helpers.go
abuiles: btw, check out `BuildAsset` in xdr/asset.go

txnbuild/helpers.go
Shaptic: Ah, this was `getAssetType` previously which staticcheck didn't like, but we don't need this for exported functions!

txnbuild/helpers.go
```diff
@@ -193,31 +193,19 @@ func NewValidationError(field, message string) *ValidationError {
...
-//lint:file-ignore U1000 Created as a helper in response to user feedback.
-func GetAssetType(canonical string) (AssetType, error) {
-	defaultType := AssetTypeNative
+// Parses an asset string in canonical form (SEP-11) into an Asset structure.
+// https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0011.md#asset
+func ParseAssetString(canonical string) (Asset, error) {
 	assets, err := xdr.BuildAssets(canonical)
 	if err != nil {
-		return defaultType, errors.Wrap(err, "failed parsing: is it in canonical (SEP-11) form?")
+		return &NativeAsset{}, errors.Wrap(err, "failed parsing: is it in canonical (SEP-11) form?")
```
leighmcculloch, line 201: What's the motivation for returning the native asset here? In the error case we could return `nil` I think. Is that possible?

txnbuild/helpers.go
Shaptic: Done!

txnbuild/helpers.go
```diff
@@ -192,3 +192,20 @@ func NewValidationError(field, message string) *ValidationError {
...
+// Parses an asset string in canonical form (SEP-11) into an Asset structure.
+// https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0011.md#asset
+func ParseAssetString(canonical string) (Asset, error) {
+	assets, err := xdr.BuildAssets(canonical)
+	if err != nil {
+		return &NativeAsset{}, errors.Wrap(err, "failed parsing: is it in canonical (SEP-11) form?")
+	}
+
+	// It returns a list, so we'll need to grab the first element.
+	asset, err := assetFromXDR(assets[0])
```
leighmcculloch, line 205: If a list of assets is given to the function is the correct behavior to return only the first, or to error given that this parsing function is intended to deserialize single assets and only single assets? I would have assumed it was the latter.

txnbuild/helpers.go
```diff
@@ -192,3 +192,20 @@ func NewValidationError(field, message string) *ValidationError {
 		Message: message,
 	}
 }
+
+// Parses an asset string in canonical form (SEP-11) into an Asset structure.
+// https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0011.md#asset
+func ParseAssetString(canonical string) (Asset, error) {
+	assets, err := xdr.BuildAssets(canonical)
+	if err != nil {
+		return &NativeAsset{}, errors.Wrap(err, "failed parsing: is it in canonical (SEP-11) form?")
```
leighmcculloch, line 201: I'd keep the error specific to what we know as facts.
```suggestion
		return &NativeAsset{}, errors.Wrap(err, "error parsing asset string")
```

txnbuild/helpers.go
```diff
@@ -192,3 +192,20 @@ func NewValidationError(field, message string) *ValidationError {
...
+	}
+
+	// It returns a list, so we'll need to grab the first element.
+	asset, err := assetFromXDR(assets[0])
+	if err != nil {
+		return &NativeAsset{}, errors.Wrap(err, "failed to create Asset from XDR")
```
leighmcculloch, line 207: This error centers on the issue of XDR, but the caller to this function probably won't be expecting XDR to be involved. That's an implementation detail. We can mention that, but should call out this is happening as part of trying to parse the asset.
```suggestion
		return &NativeAsset{}, errors.Wrap(err, "error parsing asset string via XDR types")
```

leighmcculloch (commented):
Couple suggestions.

txnbuild/helpers.go
Shaptic: There is only `xdr.BuildAssets` when it comes from `canonical string -> Asset` conversion (the `xdr.BuildAsset` does something slightly different). Since that function takes a **list** of strings, it will always return a list, but this new helper only allows a single string. 

We can definitely expect `assets` to contain one and only one asset given `err == nil`.

txnbuild/helpers.go
Shaptic: This is a "still learning Go" snafu; I didn't realize I could return a `nil` type because I thought `txnbuild.Asset` was a struct, not an interface.

txnbuild/helpers.go
leighmcculloch: I think that'd be a good idea and return an error if you get back more than one asset.

txnbuild/helpers.go
Shaptic: @abuiles `BuildAsset` needs a code/issuer/etc. to create an `Asset` whereas we want to only include the canonical asset string to create an Asset which only `BuildAssets` supports.

txnbuild/helpers.go
Shaptic: You're probably right, stranger things have happened. Amended!

txnbuild/helpers.go
```diff
@@ -192,3 +192,20 @@ func NewValidationError(field, message string) *ValidationError {
 		Message: message,
 	}
 }
+
+// Parses an asset string in canonical form (SEP-11) into an Asset structure.
+// https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0011.md#asset
+func ParseAssetString(canonical string) (Asset, error) {
+	assets, err := xdr.BuildAssets(canonical)
+	if err != nil || len(assets) != 1 {
+		return nil, errors.Wrap(err, "error parsing asset string")
```
leighmcculloch, line 201: There's an horrible trap about the errors library we use. If you call `errors.Wrap(err, ...)` and `err` is `nil`, guess what, `errors.Wrap` will return `nil`. You'll need to break these into two if blocks I think and use a non-wrap error for the length check.

If you want to read more about this issue: https://github.com/pkg/errors/issues/174

txnbuild/helpers_test.go
```diff
@@ -352,3 +352,50 @@ func TestValidateOfferManageSellOffer(t *testing.T) {
 	expectedErrMsg := "Field: OfferID, Error: amount can not be negative"
 	require.EqualError(t, err, expectedErrMsg, "valid offerID is required")
 }
+
+func TestAssetStringParsing(t *testing.T) {
+	kp0, kp1 := newKeypair0(), newKeypair1()
```
leighmcculloch, line 357: I'd move the declaration and definition of `kp1` down to the next block. It's not used by this block of code and it'd make the intentions of the code clearer if it was together with the code that uses it.

txnbuild/helpers_test.go
```diff
@@ -352,3 +352,50 @@ func TestValidateOfferManageSellOffer(t *testing.T) {
...
+
+		return expXdr.Equals(actXdr)
+	}
+
+	assert.True(t, compareAssets(native, assets[0]))
+	assert.True(t, compareAssets(cred4, assets[1]))
+	assert.True(t, compareAssets(cred12, assets[2]))
+
+	result, err := ParseAssetString("erroneous:maximus")
+	assert.Error(t, err)
```
leighmcculloch, line 399: You can test the error returned by using `assert.EqualError(t, err, "error string here")`.

txnbuild/helpers_test.go
```diff
@@ -352,3 +352,50 @@ func TestValidateOfferManageSellOffer(t *testing.T) {
...
+		if innerErr != nil {
+			return false
+		}
+
+		return expXdr.Equals(actXdr)
+	}
+
+	assert.True(t, compareAssets(native, assets[0]))
+	assert.True(t, compareAssets(cred4, assets[1]))
+	assert.True(t, compareAssets(cred12, assets[2]))
```
leighmcculloch, line 396: Is the comparison via the xdr necessary? I think you could do `assert.Equal(t, Asset(&native), ParseAssetString(nativeString))`. (Not quite a one liner, you'd need to handle the error of course.)

Shaptic: I originally had a `return a.GetCode() == b.GetCode() && ...` concoction, but then realized comparing the XDR would be more precise. In an ideal world, I'd want to do what you suggested, but can you compare interfaces like this outright in Go? How would it know what to compare?

txnbuild/helpers_test.go
Shaptic: Due to the nested `errors.Wrap`, this gets very hairy very fast. It feels fragile to me to do something like this:

```go

File: txnbuild/helpers.go (hunk 7 above)
Comment (Shaptic) [reply txnbuild/helpers.go:201]: Oh wow :flushed: great catch.

File: txnbuild/CHANGELOG.md [hunk 11]
```diff
@@ -3,6 +3,9 @@
 All notable changes to this project will be documented in this
 file.  This project adheres to [Semantic Versioning](http://semver.org/).
 
+## Unreleased
+* Expose a new helper function `ParseAssetString()`, making it easier to build an `Asset` structure from a string in [canonical form](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0011.md#asset) and check its various properties ([#3105](https://github.com/stellar/go/pull/3105)).
```
leighmcculloch, line 7:
```suggestion
* Add helper function `ParseAssetString()`, making it easier to build an `Asset` structure from a string in [canonical form](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0011.md#asset) and check its various properties ([#3105](https://github.com/stellar/go/pull/3105)).
```

leighmcculloch (approved):
🎉 :shipit:

PR #3107, opened 2020-10-07
services/horizon: Allow using captive core in integration tests

2opremio:
To do so, you need to set the `HORIZON_INTEGRATION_ENABLE_CAPTIVE_CORE` env variable to any value (e.g. `true`).


bartekn (approved):
Cool that it works out of the box! Can we add it to Circle-CI config to start another integrations tests job with captive core enabled?

2opremio:
Yep, I am working on a separate PR for that.

PR #3113, opened 2020-10-08
exp/services/webauth: remove the stellar.toml domain checks

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Remove the stellar.toml domain checks that occur on application start. Essentially revert the changes relating to this specific feature that were introduced in 5e82f16273b7ce97e50866cba207bc52f498e9c7.

### Why
SEP-10 v2.0.0 highlighted an existing feature that the `SIGNING_KEY` field in the `stellar.toml` be the key that the SEP-10 server use for signing. In 5e82f16273b7ce97e50866cba207bc52f498e9c7 we added the feature that the server would verify that the stellar.toml that references the SEP-10 server contains the correct key. This feature is inconvenient because not all SEP-10 services have a stellar.toml. It's also inconvenient when developing. It's also scope creep on what this SEP-10 server implementation is concerned with. It also delays service boot up time which ideally should not block on external services. If we need tools for validating or verifying stellar.toml's we should probably pursue that as a separate tool.

### Known limitations

This is a breaking change because it removes a config parameter, but the services is experimental and we can make breaking changes. For any service using environment variables to provide the config this will not be a breaking change and the additional config will be ignored.

### DIFF
--- a/exp/services/webauth/README.md
+++ b/exp/services/webauth/README.md
@@ -51,7 +51,6 @@ Flags:
       --network-passphrase string          Network passphrase of the Stellar network transactions should be signed for (NETWORK_PASSPHRASE) (default "Test SDF Network ; September 2015")
       --port int                           Port to listen and serve on (PORT) (default 8000)
       --signing-key string                 Stellar signing key(s) used for signing transactions comma separated (first key is used for signing, others used for verifying challenges) (SIGNING_KEY)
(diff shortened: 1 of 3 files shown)


JakeUrban (approved):

JakeUrban:
How different is this change from a simply revert of the original commit that introduced this change?

leighmcculloch:
It should be identical unless I made a mistake. The original change being reverted was bundled with other changes and so I didn't want to revert all of them.

PR #3114, opened 2020-10-08
services/horizon: Run integration tests nightly and include captive core

2opremio:


.circleci/config.yml
```diff
@@ -212,6 +212,22 @@ commands:
...
+        type: boolean
+        default: false
+    steps:
+      - run:
+          name: Run Horizon integration tests <<# parameters.enable-captive-core >>(With captive core)<</ parameters.enable-captive-core >>
+          # Currently all integration tests are in a single directory.
+          # Pulling the image helps with test running time
+          command: |
+            cd ~/go/src/github.com/stellar/go
+            docker pull stellar/quickstart:testing
```
bartekn, line 228: May not be needed after #3106.

bartekn (approved):

.circleci/config.yml
2opremio: Good point. For that we need to decide using Horizon #3106 for validation and not just local development.

PR #3117, opened 2020-10-09
Merge master into 1.10.0 horizon release

tamirms:


bartekn:
Today is the last day of 1.10.0 testing and I'm not confident we should merge it. It would be great to check if refactoring didnt' break anything and it's possible that another one: #3118 will make it into a release branch too. Are there any commits we really want in 1.10.0, https://github.com/stellar/go/pull/3117/commits/aa36914ae047e6a1a526211a17a5d559151f55da I guess? Maybe we can cherry-pick instead?

tamirms:
@bartekn good point! I will close the PR

PR #3122, opened 2020-10-09
txnbuild: Add helper to calculate claimable balance IDs without submitting transactions.

### LINKED ISSUE #3116: Feature Request: Add a function get ID of a CalimableBalance on Transaction object

For wallets built on Stellar, wallets often want to provide/sponsor the base reserve of user accounts so that users could start using the accounts smoothly without having to worry about it. Because of the financial cost of doing the sponsorship, it is important for wallets to keep track of the sponsorships.

In the event of a wallet wants to sponsor a claimable balance entry, it would be handy if the wallet backend is able to know the claimable balance id before the transaction is submitted. Because the sponsorship requires both parties sign off on it, it could be the wallet user who submits the transaction.

A function that returns the balance ID. In principle if you have a transaction and it contains a `CreateClaimableBalanceOp`, then you can deterministically produce the balance ID.

Because the `BalanceID` is calculated based on `OperationID`, `sourceAccount of the transaction`, and `the seqNum of the transaction`, and `the index of this operation in the transaction`, this method would probably have to be of the `Transaction` object.

There are other ways for the wallet backend to know the BalanceID after the transaction is submitted. For example, the response should contain the BalanceID on success, and the wallet can have its users to send this ID to the backend.

Note that even if the wallet backend monitor the ledger by using `claimant` & `sponsor` & `asset` & `amount`, it might be hard to know the exact balance id that the wallet backend signs off because those values are not unique across the ledger.

Shaptic:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Balance IDs are a crucial piece of information when dealing with claimable balances. They are highly predictable by design, and so SDK users shouldn't need to submit the transaction to determine the ID. 

This PR introduces a helper within the txnbuild package:
```go
func (t *Transaction) ClaimableBalanceID(operationIndex int) string {}
```
It returns a claimable balance ID for the particular operation within the transaction, assuming it is of the type `CREATE_CLAIMABLE_BALANCE`.

It also introduces an integration test to confirm this behavior. Since it's appended to an existing test, this only adds a few extra seconds of running time. I believe this also *only* can be integration tested since we need Core to actually run the calculation as well, but I'd love to be wrong.

### Why
Closes #3116.

### Known limitations
Obviously, this only applies to the Go SDK.

(added @jonjove to review to sanity-check the balance ID calculation and @leighmcculloch purely for visibility on the txnbuild changelog)


txnbuild/transaction.go
```diff
@@ -346,6 +346,65 @@ func (t *Transaction) Base64() (string, error) {
...
+		var account Account
+		if creation.SourceAccount == nil {
+			account = &t.sourceAccount
+		} else {
+			account = creation.GetSourceAccount()
+		}
+
+		seq, err := account.GetSequenceNumber()
+		if err != nil {
+			panic(errors.Wrap(err, "failed to query account sequence num"))
```
leighmcculloch, line 369: Can we return these errors instead of gobbling them?

txnbuild/transaction.go
```diff
@@ -346,6 +346,65 @@ func (t *Transaction) Base64() (string, error) {
 	return marshallBase64(t.envelope, t.signatures)
 }
 
+// BalanceIDs returns a list of balance IDs for all possible CreateClaimableBalance
+// operations within this transaction.
+func (t *Transaction) BalanceIDs() []string {
+	var balances []string
+	for opIndex, operation := range t.operations {
+		creation, ok := operation.(*CreateClaimableBalance)
```
leighmcculloch, line 354: I think something that looks inconvenient about this function is that it gives me back a list of balance IDs whose indexes in the returned array won't correlate to the index of the operation on the transaction. I think it might be easier to use this function if:
- The function accepts an operation index, and so only generates an ID for a single operation. The developer can call it for each operation and that would appear to offer no disadvantage and make this function a smaller building block to compose more complex functionality.
- This function is broken down into parts. Part of this function is determining the general "OperationId" for an operation of a transaction. That could be one function by itself that is used by this function.

txnbuild/transaction.go
```diff
@@ -346,6 +346,65 @@ func (t *Transaction) Base64() (string, error) {
 	return marshallBase64(t.envelope, t.signatures)
 }
 
+// BalanceIDs returns a list of balance IDs for all possible CreateClaimableBalance
+// operations within this transaction.
+func (t *Transaction) BalanceIDs() []string {
```
leighmcculloch, line 351: Could we name this function using the full term ClaimableBalanceID? It's simpler if we use consistent exact language everywhere.

txnbuild/transaction.go
```diff
@@ -346,6 +346,65 @@ func (t *Transaction) Base64() (string, error) {
...
+			// TODO: Determine the type programmatically
+			xdr.ClaimableBalanceIdTypeClaimableBalanceIdTypeV0,
+			xdr.Hash(hash))
+		if err != nil {
+			panic(errors.Wrap(err, "unable to parse balance ID as XDR"))
+		}
+
+		balanceIdHex, err := xdr.MarshalHex(balanceIdXdr)
+		if err != nil {
+			panic(errors.Wrap(err, "unable to parse balance ID as hex"))
```
leighmcculloch, line 399: This is encoding to hex, not parsing.
```suggestion
			panic(errors.Wrap(err, "unable to encode balance ID as hex"))
```

leighmcculloch (commented):
Yes! Thanks for doing this. Couple suggestions below.

txnbuild/transaction.go
Shaptic: You're absolutely right; this is exactly the kind of solution to the "I'm not thrilled about the design" problem I was looking for. The code has been amended to be operation-specific!

txnbuild/transaction.go
Shaptic: Addressed!

txnbuild/transaction.go
Shaptic: Good call, fixed! This was originally a product of the fact that I didn't want a half-processed array returned, but with the refactor to be specific to one operation, returning errors fits perfectly into the design.

howardtw (commented):
Thanks for doing this! I left a couple of nitpicks. This PR looks good to me, but I'd defer the approval to @leighmcculloch to make sure it looks good to him too before you merge the PR.

txnbuild/transaction.go
```diff
@@ -346,6 +346,64 @@ func (t *Transaction) Base64() (string, error) {
 	return marshallBase64(t.envelope, t.signatures)
 }
 
+// Returns the (pre-calculated) claimable balance ID for the operation at the
```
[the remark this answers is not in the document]
leighmcculloch: We should add the function name as the first word, for standard Go doc syntax, but keep the rest natural language too.
```suggestion
// ClaimableBalanceID returns the claimable balance ID for the operation at the given index within the transaction.
```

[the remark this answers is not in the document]
leighmcculloch: We should add the function name as the first word, for standard Go doc syntax, but keep the rest natural language too.
```suggestion
// ClaimableBalanceID returns the claimable balance ID for the operation at the given index within the transaction.
```

leighmcculloch (approved):

services/horizon/internal/test/integration/integration.go
```diff
@@ -579,16 +589,20 @@ func (i *Test) SubmitMultiSigOperations(
 
 	tx, err := txnbuild.NewTransaction(txParams)
 	if err != nil {
-		return proto.Transaction{}, err
+		return tx, err
```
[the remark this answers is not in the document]
Shaptic: I spent way too long musing about the implications of a "best effort" return value to the caller before ultimately realizing that it's such a confusing and unnecessary way to approach this API.

PR #3102, opened 2020-10-06
services/horizon: Check state rebuild in state verification integration test

bartekn:
<details>
  <summary>PR Checklist</summary>
  
### PR Structure

* [ ] This PR has reasonably narrow scope (if not, break it down into smaller PRs).
* [ ] This PR avoids mixing refactoring changes with feature changes (split into two PRs
  otherwise).
* [ ] This PR's title starts with name of package that is most changed in the PR, ex.
  `services/friendbot`, or `all` or `doc` if the changes are broad or impact many
  packages.

### Thoroughness

* [ ] This PR adds tests for the most critical parts of the new functionality or fixes.
* [ ] I've updated any docs ([developer docs](https://www.stellar.org/developers/reference/), `.md`
  files, etc... affected by this change). Take a look in the `docs` folder for a given service,
  like [this one](https://github.com/stellar/go/tree/master/services/horizon/internal/docs).

### Release planning

* [ ] I've updated the relevant CHANGELOG ([here](services/horizon/CHANGELOG.md) for Horizon) if
  needed with deprecations, added features, breaking changes, and DB schema changes.
* [ ] I've decided if this PR requires a new major/minor version according to
  [semver](https://semver.org/), or if it's mainly a patch change. The PR is targeted at the next
  release branch if it's not a patch change.
</details>

### What

Check state rebuild in state verification integration test.

### Why

All integration tests start from empty state so we don't test building state from history archives. Extended state verification integration test to trigger state rebuild (`horizon expingest trigger-state-rebuild`) and verify rebuild state again. Such test would catch the bugs like https://github.com/stellar/go/pull/3100 and https://github.com/stellar/go/pull/3096.

### Known limitations

It will make the test 3x slower so let's merge it after moving integration tests to nightly runs.

PR #3100, opened 2020-10-06
services/horizon: Add claimable_balances table to state tables to truncate

bartekn:
<details>
  <summary>PR Checklist</summary>
  
### PR Structure

* [ ] This PR has reasonably narrow scope (if not, break it down into smaller PRs).
* [ ] This PR avoids mixing refactoring changes with feature changes (split into two PRs
  otherwise).
* [ ] This PR's title starts with name of package that is most changed in the PR, ex.
  `services/friendbot`, or `all` or `doc` if the changes are broad or impact many
  packages.

### Thoroughness

* [ ] This PR adds tests for the most critical parts of the new functionality or fixes.
* [ ] I've updated any docs ([developer docs](https://www.stellar.org/developers/reference/), `.md`
  files, etc... affected by this change). Take a look in the `docs` folder for a given service,
  like [this one](https://github.com/stellar/go/tree/master/services/horizon/internal/docs).

### Release planning

* [ ] I've updated the relevant CHANGELOG ([here](services/horizon/CHANGELOG.md) for Horizon) if
  needed with deprecations, added features, breaking changes, and DB schema changes.
* [ ] I've decided if this PR requires a new major/minor version according to
  [semver](https://semver.org/), or if it's mainly a patch change. The PR is targeted at the next
  release branch if it's not a patch change.
</details>

Add `claimable_balances` table to state tables so it's truncated when rebuilding state.

PR #3127, opened 2020-10-14
services/horizon: Check state rebuild in state verification integration test

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

Recreating because https://github.com/stellar/go/pull/3102 was removed because I removed a release branch.

### What

Check state rebuild in state verification integration test.

### Why

All integration tests start from empty state so we don't test building state from history archives. Extended state verification integration test to trigger state rebuild (`horizon expingest trigger-state-rebuild`) and verify rebuild state again. Such test would catch the bugs like https://github.com/stellar/go/pull/3100 and https://github.com/stellar/go/pull/3096.

### Known limitations

It will make the test 3x slower so let's merge it after moving integration tests to nightly runs.


tamirms (approved):

services/horizon/internal/test/integration/integration.go
```diff
@@ -305,6 +305,20 @@ func (i *Test) LedgerIngested(sequence uint32) bool {
...
+// LedgerClosed returns true if the ledger with a given sequence has been
+// closed by Stellar-Core. Panics in case of errors. Note it's different
+// than LedgerIngested because it checks if the ledger was closed, not
+// necessarily ingested (ex. when rebuilding state Horizon does not ingest
+// recent ledgers).
+func (i *Test) LedgerClosed(sequence uint32) bool {
+	root, err := i.Client().Root()
+	if err != nil {
+		panic(err)
+	}
```
Shaptic, line 317: could use `panicIf` here for consistency!

services/horizon/internal/integration/protocol14_state_verifier_test.go
```diff
@@ -100,34 +100,62 @@ func TestProtocol14StateVerifier(t *testing.T) {
...
+	itest.RunHorizonCLICommand([]string{"expingest", "trigger-state-rebuild"})
+
+	// Wait for the second checkpoint ledger and state rebuild
+	for !itest.LedgerClosed(127) {
+		t.Log("First checkpoint ledger (127) not closed yet...")
+		time.Sleep(5 * time.Second)
+	}
+
+	// Wait for the third checkpoint ledger and state verification trigger
+	for !itest.LedgerIngested(191) {
```
Shaptic, line 122: How do we know about these "magic numbers"?

Shaptic (commented):
LGTM but I recommend a second pair of eyes on this! :eyes:

services/horizon/internal/integration/protocol14_state_verifier_test.go
bartekn: Good call!

bartekn:
@Shaptic good feedback. Changes implemented in https://github.com/stellar/go/pull/3127/commits/063ffe2155dfb3a28c2d5689c0c596a0dec84c0b.

services/horizon/internal/integration/protocol14_state_verifier_test.go
Shaptic: Wait lol this was an actual question more than a review comment. Are these documented anywhere or is there a pattern, etc.? Like how did you "know" to use these numbers in particular?

bartekn: But it's worth explaining in the code anyway. The checkpoint ledgers are closed every 64 ledgers (63, 127, 191) and this when we have access to ledger state after a given checkpoint ledger and we can run state verifier. The pattern is `(i+1) mod 64`, for `i = 0, 1, 2...`.

Shaptic: Ah, gotcha! Thanks for explaining :+1:

services/horizon/internal/integration/protocol14_state_verifier_test.go
```diff
@@ -99,35 +105,64 @@ func TestProtocol14StateVerifier(t *testing.T) {
...
+	for !itest.LedgerClosed(secondCheckpoint) {
+		t.Log("Second checkpoint ledger not closed yet...")
+		time.Sleep(5 * time.Second)
+	}
+
+	// Wait for the third checkpoint ledger and state verification trigger
+	for !itest.LedgerIngested(thirdCheckpoint) {
+		t.Log("Third checkpoint ledger not ingested yet...")
+		time.Sleep(5 * time.Second)
+	}
```
Shaptic, line 131: Wait, okay so now that I understand the checkpoint thing a little more, why do we check for `secondCheckpoint`, then `thirdCheckpoint`, and only _then_ do the state verification? It seems the check for the second is redundant, no? Since it has to wait for the third as the next step, and third implies second.

bartekn: Ingestion is running a state machine (`fsm.go` file). During normal Horizon operations with usually stay in `resumeState` all the time. However, `horizon ingest trigger-state-rebuild` command makes the machine jump to `startState`. There will fall into the [case](https://github.com/stellar/go/blob/118e4a01ad4d9b0964ebfc3a5671a062a0a05e6e/services/horizon/internal/ingest/fsm.go#L189) that makes us wait for the next checkpoint. Why? It's because after the first checkpoint and state verification we're at ledger above 63. But the last checkpoint is still 63. So we wait for the network to close the second ledger to be able to catchup. This won't trigger state verification because it's [triggered](https://github.com/stellar/go/blob/118e4a01ad4d9b0964ebfc3a5671a062a0a05e6e/services/horizon/internal/ingest/fsm.go#L496) from `resumeState` only. That's why we need to wait for the third checkpoint.

bartekn:
I merged master with #3167 and reverted https://github.com/stellar/go/pull/3127/commits/e8d89da83f1351e03eb5915b55b96ad979bea941 (integration tests are quick now). Can I have one last review? @2opremio @tamirms @Shaptic

tamirms (approved):

PR #3135, opened 2020-10-16
Update stellartoml client

JakeUrban:
Updating the stellartoml client so that the HTTP response includes all SEP-1 attributes

Also changes `addy` parameter to `addr`.


clients/stellartoml/main.go
```diff
@@ -10,98 +10,98 @@ const StellarTomlMaxSize = 100 * 1024
...
 type Principal struct {
-	Name                  string `toml:"name"`
-	Email                 string `toml:"email"`
-	Keybase               string `toml:"keybase"`
-	Telegram              string `toml:"telegram"`
-	Twitter               string `toml:"twitter"`
-	Github                string `toml:"github"`
-	IdPhotoHash           string `toml:"id_photo_hash"`
-	VerificationPhotoHash string `toml:"verification_photo_hash"`
+    Name                  string `toml:"name"`
```
leighmcculloch, line 38: Go is always tabs. The compiler doesn't care, but `gofmt` will always format for tabs. Don't fight it 😄.

JakeUrban: I was kinda mixing the two while adding these attributes and the formatter yelled at me. I figured there were no tabs allowed but maybe its the opposite? I'm not sure when to use tabs vs spaces now. For example, `string `toml:...` is obviously a space.

JakeUrban: Actually it sounds like indentation should always be tabs, otherwise spaces are fine.

leighmcculloch: I think that is right. My editor does it for me so 🤷.

clients/stellartoml/client_test.go
```diff
@@ -1,10 +1,11 @@
 package stellartoml
 
 import (
-	"net/http"
 	"strings"
 	"testing"
 
+	"net/http"
```
howardtw, line 7: Interesting change. My editor doesn't arrange the imports this way. Do you think we can restore it? Don't bother if you can't.

clients/stellartoml/main.go
```diff
@@ -25,21 +28,88 @@ type Client struct {
...
+type Validator struct {
+	Alias       string `toml:"ALIAS"`
+	DisplayName string `toml:"DISPLAY_NAME"`
+	PublicKey   string `toml:"PUBLIC_KEY"`
+	Host        string `toml:"HOST"`
+	History     string `toml:"HISTORY"`
 }
 
-// Response represents the results of successfully resolving a stellar.toml file
 type Response struct {
```
howardtw, line 80: Can we add a link this SEP-1 here so that client will know which version of SEP-1 this response maps to. https://github.com/stellar/stellar-protocol/blob/f8993e36fa6b5b8bba1254c21c2174d250af4958/ecosystem/sep-0001.md

howardtw (approved):
LGTM with some minor suggestions.

clients/stellartoml/main.go
JakeUrban: What good does that do? That link will always go to the current (only) version right?

howardtw: That link has a commit sha so readers will always know which version that is.

JakeUrban: I see, I can add it, but won't it need to be maintained when this code is updated? Shouldn't we assume the TOML version support coincides with the Go SDK version released?

howardtw: > won't it need to be maintained when this code is updated?

That's right. Whenever we add a new field or drop a new field in SEP-10 and adjust the `Response` struct accordingly, we will have to update the link. IMO it's more clear if the client knows which SEP-1 version that this stellartoml client is compatible with.

> Shouldn't we assume the TOML version supported coincides with the Go SDK version released?

Does the Go SDK use this client and have a specific version of TOML that it uses?

JakeUrban: The single `go` repo confuses me, maybe the Go SDK has nothing to do with this client. I'll add the link 👍

PR #3138, opened 2020-10-16
Use --start-ledger and --start-hash in online captive core.

abuiles:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Use `--start-ledger` and `--start-hash` in captive core.

### Why

This allow us to do a fast catchup without having to call `new-db` and `catchup` before.

### Known limitations

For now we are relying on the history archive to get the ledger header hash. We need to refactor that next.  If we have previous ingested ledgers in Horizon DB, then we should read the hash and ledger sequence from there.

### DIFF
--- a/historyarchive/archive.go
+++ b/historyarchive/archive.go
@@ -20,6 +20,7 @@ import (
 	"sync"
 
 	"github.com/stellar/go/support/errors"
+	"github.com/stellar/go/xdr"
 )
 
 const hexPrefixPat = "/[0-9a-f]{2}/[0-9a-f]{2}/[0-9a-f]{2}/"
@@ -59,6 +60,7 @@ type ArchiveInterface interface {
 	PutPathHAS(path string, has HistoryArchiveState, opts *CommandOptions) error
 	BucketExists(bucket Hash) (bool, error)
 	CategoryCheckpointExists(cat string, chk uint32) (bool, error)
+	GetLedgerHeader(chk uint32) (xdr.LedgerHeaderHistoryEntry, error)
 	GetRootHAS() (HistoryArchiveState, error)
 	GetCheckpointHAS(chk uint32) (HistoryArchiveState, error)
 	PutCheckpointHAS(chk uint32, has HistoryArchiveState, opts *CommandOptions) error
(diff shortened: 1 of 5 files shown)


historyarchive/archive.go
```diff
@@ -149,6 +151,28 @@ func (a *Archive) CategoryCheckpointExists(cat string, chk uint32) (bool, error)
...
+
+	for {
+		err = xdrStream.ReadOne(&ledgerHeader)
+		if err != nil {
+			if err == io.EOF {
+				break
+			}
+			return ledgerHeader, errors.Wrap(err, "error reading from ledger stream")
+		}
+	}
```
bartekn, line 171: Is `for` needed here?

historyarchive/archive.go
```diff
@@ -149,6 +151,28 @@ func (a *Archive) CategoryCheckpointExists(cat string, chk uint32) (bool, error)
 	return a.backend.Exists(CategoryCheckpointPath(cat, chk))
 }
 
+func (a *Archive) GetLedgerHeader(chk uint32) (xdr.LedgerHeaderHistoryEntry, error) {
```
bartekn, line 154: `chk` is cryptic. Why not `sequence`?

ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -246,9 +251,58 @@ func (c *CaptiveStellarCore) openOnlineReplaySubprocess(from uint32) error {
...
+		_, _, err := c.GetLedger(from)
+		return errors.Wrapf(err, "Error fast-forwarding to %d", from)
+	}
+
 	return nil
 }
 
+// runFromParams receives a ledger sequence and calculates the required values to call stellar-core run with --start-ledger and --start-hash
+func (c *CaptiveStellarCore) runFromParams(from uint32) (runFrom uint32, ledgerHash string, nextLedger uint32, err error) {
+	if (from+1)%ledgersPerCheckpoint == 0 {
```
bartekn, line 266: Nit
```suggestion
	if historyarchive.IsCheckpoint(from) {
```

bartekn (approved):

historyarchive/archive.go
abuiles: this is meant to be used only with checkpoint ledgers

historyarchive/archive.go
abuiles: oh, good point, we probably don't need the for.

abuiles: nvm - we do since we are reading from the stream until there is nothing else to read.

historyarchive/archive.go
bartekn: We use it for checkpoints only in our case but you can get any ledger header using the method.

Issue #3037, opened 2020-09-17
Improve developer experience with integration tests


2opremio:
Complementary to #3031 

We have our first set of integration tests! (yay! kudos guys!). But (and there is always a but) the developer experience could be improved:

1. Environment variables need to be set in order to establish:
   a. That we in fact want integration tests to run (`HORIZON_INTEGRATION_TESTS`). We could use [go build tags for that](https://mickey.dev/posts/go-build-tags-testing/), I think.
   b. Where to find the Horizon binary (`HORIZON_BIN_DIR`). This is pretty (particularly in developer machines other than Linux, since it expects a Linux binary) since you need to manually recompile the binary every time you make a Horizon change. As a developer, I would expect that running an integration test automatically propagates my source-code tree.
   c. Whether to re-create the scaffolding (`HORIZON_SKIP_CREATION`). As a developer writing an integration test I want the feedback loop to be as fast as possible without recreating infrastructure and without needing to specify so explicitly.

2. It is stateful. A container is created every time a test is run and sometimes is left behind (if something goes wrong), breaking successive runs.

Additionally I want to insist on the importance of running tests quickly (particularly in the context of #3031). As a developer, I don't want to wait 10 seconds every time I run a test and I don't want to bookkeep the Horizon binary (which takes time).

Ideally, the test creation code would be smart about the work already done in prior runs (e.g. reusing the standalone network). We could make this opt-out (or opt-in) and provide different levels of reusability.

2opremio:
After needing to debug a bug in horizon found through integration tests, I think it would be fantastic if you could run Horizon from the test itself instead of using an opaque binary in the container.

The integration test machinery would spawn Horizon in the host, as part of the test code, making it easy to debug exactly what is happening in Horizon.

leighmcculloch:
Where do the integration tests live? I was trying to find `HORIZON_INTEGRATION_TESTS` in the code and I could not find it.

I used build tags in `pasapeso-backend` for integration tests and the one thing I will say is the downside of doing that is there's no visibility that tests aren't running, and the tests won't be compiled locally when building without the tags. The advantage of using an env var, or adding a custom flag to the go test command, and then checking either in the test itself is you call `t.Skip()` and then there's visibility to what tests are being skipped.

2opremio:
> Where do the integration tests live?

in the `protocol-14` branch until it's released.

Shaptic:
[...]

PR #3143, opened 2020-10-20
services/horizon: Move integration tests away from stellar/quickstart.

### LINKED ISSUE #3031: Use docker-compose workflow in integration tests
Improving upon Bartek's initial implementation of https://github.com/stellar/go/issues/2812 we can use docker compose to spin up a new stand alone network instead of using the stellar quick start docker image.

The advantage of using docker compose is that we can simplify the integration test scaffolding. In the current implementation, we use the golang docker client to programmatically create a docker container from the quickstart image to run the standalone network. Instead, we can simply run the [start script](https://github.com/stellar/go/blob/master/services/horizon/docker/start.sh) as a shell command to spawn the standalone network.  For the tear down process we can run `docker-compose down -v` as a shell command.

Another big advantage of using the docker compose workflow is that we don't need to build a horizon binary from source and copy it over to the quickstart docker container.

We will need to modify the docker-compose workflow slightly in order to support the integration tests workflow.

In particular, we need to make sure that [`horizon ingest init-genesis-state`](https://github.com/stellar/go/pull/2910) is executed when horizon starts running the stand alone network. Otherwise, Horizon ingestion will not start until the first checkpoint ledger is created by stellar core which takes around 5 minutes.

We also will need to make the protocol version configurable in the docker compose workflow. At the moment, the protocol version for the stand alone network is hard coded in the docker-compose yaml file.

### LINKED ISSUE #3121: Move integration tests away from the stellar/quickstart:testing image.
There are too many problems with the developer experience when relying on this all-in-one image that will likely be deprecated in favor of the Docker Compose solution as a "getting started" method soon.

Many issues are covered in #3037; in summary, give me flexibility & speed or give me death.

Shaptic:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
This moves the integration test infrastructure away from creating a stellar/quickstart:testing instance for each test, preferring instead to use the modular Docker Compose method.

### Why
We want more flexibility with our integration tests; see #3037 for more discussion. 

This closes #3031 and closes #3121.

### Known limitations
A couple of kinks:
 - Ctrl+C is unreliable as far as cleanup goes, but per discussion below we may want to address this in a later PR, since it it that failing tests are properly handled on the next run.


go.list
```diff
@@ -125,17 +125,20 @@ github.com/yalp/jsonpath v0.0.0-20150812003900-31a79c7593bb
...
+golang.org/x/net v0.0.0-20200822124328-c89045814202
 golang.org/x/oauth2 v0.0.0-20190226205417-e64efc72b421
-golang.org/x/sync v0.0.0-20190423024810-112230192c58
-golang.org/x/sys v0.0.0-20190606165138-5da285871e9c
+golang.org/x/sync v0.0.0-20200625203802-6e8e738ad208
+golang.org/x/sys v0.0.0-20200323222414-85ca7c5b95cd
 golang.org/x/text v0.3.3
-golang.org/x/tools v0.0.0-20190624180213-70d37148ca0c
+golang.org/x/tools v0.0.0-20201019175715-b894a3290fff
+golang.org/x/xerrors v0.0.0-20200804184101-5ec99f83aff1
```
bartekn, line 141: Can you explain why changes in dependencies were required?

services/horizon/internal/test/integration/integration.go
```diff
@@ -166,99 +170,48 @@ func NewTest(t *testing.T, config Config) *Test {
...
 	config, configOpts := horizon.Flags()
-
 	cmd := &cobra.Command{
 		Use:   "horizon",
 		Short: "client-facing api server for the stellar network",
-		Long:  "client-facing api server for the stellar network. It acts as the interface between Stellar Core and applications that want to access the Stellar network. It allows you to submit transactions to the network, check the status of accounts, subscribe to event streams and more.",
+		Long: `client-facing api server for the stellar network. It acts as the
+interface between Stellar Core and applications that want to access the Stellar
+network. It allows you to submit transactions to the network, check the status
+of accounts, subscribe to event streams and more.`,
```
bartekn, line 183: Nit: fix alignment.
```suggestion
		Long:
`client-facing api server for the stellar network. It acts as the interface
between Stellar Core and applications that want to access the Stellar network.
It allows you to submit transactions to the network, check the status of
accounts, subscribe to event streams and more.`,
```

services/horizon/internal/test/integration/integration.go
```diff
@@ -635,3 +518,9 @@ func panicIf(err error) {
 		panic(err)
 	}
 }
+
+func fatalIf(t *testing.T, err error) {
```
bartekn, line 522: Not in a scope of this PR but can we replace `panicIf` with `fatalIf`. I can be wrong but I think I noticed that `Cleanup` is not called in case of `panic`.

services/horizon/internal/test/integration/integration.go
```diff
@@ -78,86 +78,90 @@ func NewTest(t *testing.T, config Config) *Test {
...
 
-	if !skipCreation {
-		err = createTestContainer(i, image)
-		if err != nil {
-			t.Fatal(errors.Wrap(err, "error creating docker container"))
+	// Walk up the tree until we find "go.mod", which we treat as the root
+	// directory of the project.
+	current, err := os.Getwd()
+	fatalIf(t, err)
+	for !directoryContains(current, "go.mod") {
```
bartekn, line 106: I think in correctly configured environment you should be able to find the project root by concatenating `$GOPATH` env variable and `src/github.com/stellar/go`.

services/horizon/internal/test/integration/integration.go
```diff
@@ -1,22 +1,22 @@
...
+	"path"
+	"path/filepath"
+	"runtime"
 	"strconv"
 	"syscall"
 	"testing"
 	"time"
 
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/container"
```
bartekn, line 19: We should be able to remove all `docker/docker/*` dependencies now. Where do we use it?

services/horizon/docker/docker-compose.standalone.yml
```diff
@@ -14,10 +14,10 @@ services:
     environment:
       - HISTORY_ARCHIVE_URLS=http://host.docker.internal:1570
       - NETWORK_PASSPHRASE=Standalone Network ; February 2017
-  # this container will invoke a request to upgrade stellar core to protocol 13
+  # this container will invoke a request to upgrade stellar core to protocol 14 (by default)
   core-upgrade:
     restart: on-failure
     image: curlimages/curl:7.69.1
-    command: ["-v", "-f", "http://host.docker.internal:11626/upgrades?mode=set&upgradetime=1970-01-01T00:00:00Z&protocolversion=13"]
+    command: ["-v", "-f", "http://host.docker.internal:11626/upgrades?mode=set&upgradetime=1970-01-01T00:00:00Z&protocolversion=${PROTOCOL_VERSION:-14}"]
```
bartekn, line 21: Maybe we can upgrade to the [latest version available](https://github.com/stellar/docker-stellar-core-horizon/blob/14188afdd9d30cb45890b90b217f6e3aabf85d63/start#L284-L287) if it's not set?

2opremio:
@Shaptic could you please make sure that the captive-core integration tests (which are currently disabled until #3144 ) are also ported?

services/horizon/internal/test/integration/integration.go
Shaptic: From what I gather, this would give you the `go get` version on the master branch, but I think we're much likelier to want to use our local branch/repo. Unless, of course, I don't fall into the category of a "correctly configured environment" :sweat_smile:

go.list
Shaptic: Hmm, nope, I can't really explain. This was just a result of running the code; I thought Go just normally auto-updated the dependencies or something. I'll look into it.. hopefully can find a better reason.

services/horizon/internal/test/integration/integration.go
Shaptic: `go fmt` doesn't like this!

services/horizon/internal/test/integration/integration.go
Shaptic: Yep, `IntegrationTest.Close()` was still doing Docker things but we don't call the method anymore.

.circleci/config.yml
```diff
@@ -419,6 +418,9 @@ jobs:
           command: |
             echo "export HORIZON_INTEGRATION_TESTS=true" >> $BASH_ENV
             echo "export HORIZON_BIN_DIR=~/go/src/github.com/stellar/go" >> $BASH_ENV
+      - run:
+          name: Pull latest Stellar Core image
+          command: docker pull stellar/stellar-core
```
2opremio, line 423: I think I would pin this to a specific version

services/horizon/docker/docker-compose.integration-tests.yml
```diff
@@ -0,0 +1,25 @@
+version: '3'
```
2opremio, line 1: Can't we reuse (or at least piggyback on) the standalone configuration file?

services/horizon/docker/stellar-core-integration-tests.cfg
```diff
@@ -0,0 +1,28 @@
+# simple configuration for a standalone test "network"
+# see stellar-core_example.cfg for a description of the configuration parameters
```
2opremio, line 2: Did you consider using a template instead of having multiple configuration files?

services/horizon/docker/docker-compose.integration-tests.yml
tamirms: it doesn't make sense override the standalone condiguration anymore. there are too many differences and we only need two containers (stellar core and stellar core postgres)

services/horizon/internal/test/integration/integration.go
```diff
@@ -85,191 +67,131 @@ func NewTest(t *testing.T, config Config) *Test {
...
-	image := "stellar/quickstart:testing"
-	skipCreation := os.Getenv("HORIZON_SKIP_CREATION") != ""
-
-	if skipCreation {
-		t.Log("Trying to skip container creation...")
-		containers, _ := i.cli.ContainerList(
-			context.Background(),
-			types.ContainerListOptions{All: true, Quiet: true})
+	// Lets you check if a particular directory contains a file.
+	directoryContains := func(root string, needle string) bool {
```
2opremio, line 71:
```suggestion
	directoryContainsFileName := func(dir string, filenameToFind string) bool {
```

services/horizon/internal/test/integration/integration.go
```diff
@@ -85,191 +67,131 @@ func NewTest(t *testing.T, config Config) *Test {
...
 
-	if !skipCreation {
-		err = createTestContainer(i, image)
-		if err != nil {
-			t.Fatal(errors.Wrap(err, "error creating docker container"))
-		}
-	}
+	// Walk up the tree until we find "go.mod", which we treat as the root
+	// directory of the project.
+	current, err := os.Getwd()
```
2opremio, line 86: Doesn't this assume you run the integration tests from the current directory? I am not sure this will always be the case (e.g. from IDEs).

Maybe this will work https://stackoverflow.com/a/18537792/1914440

2opremio:
> * I've only tested on Linux so far (for obvious reasons) which has some quirks relative to other platforms, purely judging from the `start.sh` script we have.
> 
> * Ctrl+C reaaaally does not work well, and I haven't looked at what happens when tests fail, either.
> 
> * It's still TBD whether or not we want to run the `horizon-postgres` container: from a "just let me run the tests" perspective, it might be better/cleaner to keep the DBs as self-contained as possible.

Is this resolved?

services/horizon/docker/stellar-core-integration-tests.cfg
tamirms: could you elaborate on how we could use a template here (e.g. what templating engine would we use, how would we invoke the template to generate the actual configuration we need) ?

2opremio: e.g. https://golang.org/pkg/text/template/ would probably be enough , you would need to use Go (e.g. from the integration tests) to fill it up.

.circleci/config.yml
bartekn: It's worth checking the latest version to confirm it tests work with the latest core version.

services/horizon/internal/integration/protocol14_state_verifier_test.go
```diff
@@ -118,16 +118,17 @@ func TestProtocol14StateVerifier(t *testing.T) {
...
 	// Wait for the second checkpoint ledger and state rebuild
 	// Core will push to history archives *after* checkpoint ledger
-	itest.CloseCoreLedgersUntilSequence(secondCheckpoint + 1)
+	err = itest.CloseCoreLedgersUntilSequence(secondCheckpoint + 1)
 	assert.NoError(t, err)
 
 	// Wait for the third checkpoint ledger and state verification trigger
 	// Core will push to history archives *after* checkpoint ledger
-	itest.CloseCoreLedgersUntilSequence(thirdCheckpoint + 1)
+	err = itest.CloseCoreLedgersUntilSequence(thirdCheckpoint + 1)
```
bartekn, line 131: Thanks!

services/horizon/docker/stellar-core-integration-tests.cfg
tamirms: ok, I think we'll definitely take this approach if we need to generate several stellar core configurations for the different integration test cases. for now it seems that this stellar core configuration will work for all the integration test cases

services/horizon/internal/test/integration/integration.go
bartekn: This is what I mean by [correctly configured environment](https://golang.org/doc/gopath_code.html#Organization) but I should rather say typical environment. AFAIR, since go modules were introduced there were some issues when the code was outside $GOPATH. We can leave existing code if it works, it was more _nit_.

tamirms:
> I've only tested on Linux so far (for obvious reasons) which has some quirks relative to other platforms, purely judging from the start.sh script we have.

I've removed the platform specific quirks so it should work on all operating systems which can run docker-compose. I've run the tests locally on my macbook and on linux via circleci.

> It's still TBD whether or not we want to run the horizon-postgres container: from a "just let me run the tests" perspective, it might be better/cleaner to keep the DBs as self-contained as possible.

Let's not run the horizon-postgres container for now. 

> Ctrl+C reaaaally does not work well, and I haven't looked at what happens when tests fail, either.

When the tests fail, the cleanup hook will destroy all the docker containers created by the integration tests.
Ctrl+C seems to abort instantly for me. However, when I Ctrl+C the docker containers from the integration tests are still running.

services/horizon/internal/integration/protocol14_state_verifier_test.go
2opremio: +1

services/horizon/internal/test/integration/integration.go
Shaptic: The only assumption is that the test is run from _somewhere within the project_. I tried a few methods: your linked one, an approach via [runtime.Caller](https://golang.org/pkg/runtime/#Caller), and the above. This one appeared to be the most reliable; however, I could be wrong (esp. since Goland, Sublime, and VS Code probably all do things slightly differently) and we may need to adjust this.

Shaptic:
I refactored a small bit regarding how we find the docker-compose file in the latest changes 4ede3af69a8e0c79040748def271013b1e912fef.

Unrelated to the above (fails w/ and w/o the changes), but on my local it looks like a single test will fail intermittently with a "still ingesting" error. I've tried but can't track down the cause; it appears to only happen on either `TestProtocol14StateVerifier` or `TestProtocol15Basics`. Not when run individually, mind you, but if all of the tests are run. 

I can definitely see our introduction of concurrency to testing (incl. the captive core stuff) causing hard-to-track-down and [Heisenbug-esque](http://www.catb.org/jargon/html/H/heisenbug.html) test failures. I'm not sure if this is just something mucked up with my environment, but are we ready to merge this?

tamirms:
@Shaptic I thought I fixed the "still ingesting" error earlier in https://github.com/stellar/go/pull/3143/commits/016c91e9da8922aa314f662f24813be09da710ca . But then I saw your comment and was able to reproduce the error in CircleCI. I finally figured out the cause and it's because the root endpoint determines the sequence numbers from a global `ledger.State` variable. This global variable is mutable and persists between test runs. Horizon periodically calls `ledger.SetState()` to update the variable. However, it's possible that the check in https://github.com/stellar/go/pull/3143/commits/016c91e9da8922aa314f662f24813be09da710ca is executed before `ledger.SetState()` is called for the first time by the new Horizon instance.

In https://github.com/stellar/go/pull/3143/commits/027a48c5191625ef43c7d73de4fb11f411bf0c64 , I make sure to clear the ledger state after every test run. This will ensure that `waitForHorizon()` will make sure to wait until Horizon ingestion is ready.

I have run the integration tests on CircleCI 5 times with this latest fix and so far there haven't been any failures.

Shaptic:
Wow, ridiculously subtle, thanks for noticing that! I ran the suite a few times locally, as well: only seeing green :tada:. I think we can push back resolving the Ctrl+C inconsistencies to another PR to get this in and running. 

@stellar/horizon-committers anything we're missing?

services/horizon/internal/test/integration/integration.go
```diff
@@ -85,191 +66,105 @@ func NewTest(t *testing.T, config Config) *Test {
...
-	if os.Getenv("HORIZON_INTEGRATION_ENABLE_CAPTIVE_CORE") == "" {
-		i.startHorizon(containerInfo, coreURL)
+	// Register cleanup handlers (on panic and ctrl+c) so the containers are
+	// stopped even if ingestion or testing fails.
+	cleanup := func() {
+		if i.app != nil {
+			i.app.Close()
+			// Clear the ledger state otherwise the root response
+			// will contain ledger information from the previous test run
+			ledger.SetState(ledger.State{})
```
bartekn, line 102: Good catch @tamirms. Created an issue for this: https://github.com/stellar/go/issues/3195.

services/horizon/internal/test/integration/integration.go
```diff
@@ -2,80 +2,61 @@
...
 )
 
 const (
 	NetworkPassphrase           = "Standalone Network ; February 2017"
-	stellarCorePostgresPassword = "integration-tests-password"
+	stellarCorePostgresPassword = "mysecretpassword"
 	adminPort                   = 6060
+	stellarCorePort             = "11626"
+	stellarCorePostgresPort     = "5641"
+	historyArchivePort          = "1570"
```
bartekn, line 40: Nit: make all ports variables integers for consistency?

bartekn (approved):

PR #3140, opened 2020-10-19
Merge release-horizon-v1.10.1 into master

bartekn:

PR #3146, opened 2020-10-21
services/horizon: Fix Protocol 15 integration test

bartekn:
I broke it when resolving conflicts in https://github.com/stellar/go/pull/3140.

### DIFF
--- a/services/horizon/internal/integration/protocol14_test.go
+++ b/services/horizon/internal/integration/protocol14_test.go
@@ -27,8 +27,8 @@ func TestProtocol15Basics(t *testing.T) {
 	t.Run("Sanity", func(t *testing.T) {
 		root, err := itest.Client().Root()
 		tt.NoError(err)
-		tt.Equal(int32(14), root.CoreSupportedProtocolVersion)
-		tt.Equal(int32(14), root.CurrentProtocolVersion)
+		tt.Equal(int32(15), root.CoreSupportedProtocolVersion)
+		tt.Equal(int32(15), root.CurrentProtocolVersion)
 
 		// Submit a simple tx
 		op := txnbuild.Payment{


services/horizon/internal/integration/protocol14_test.go
```diff
@@ -27,8 +27,8 @@ func TestProtocol15Basics(t *testing.T) {
 	t.Run("Sanity", func(t *testing.T) {
 		root, err := itest.Client().Root()
 		tt.NoError(err)
-		tt.Equal(int32(14), root.CoreSupportedProtocolVersion)
-		tt.Equal(int32(14), root.CurrentProtocolVersion)
+		tt.Equal(int32(15), root.CoreSupportedProtocolVersion)
+		tt.Equal(int32(15), root.CurrentProtocolVersion)
```
Shaptic, line 31: Any thoughts on using `tt.LessEqual` here? It seems odd to have to update these with each new protocol.

Shaptic: Actually, now that I think about it, do we even want a configurable version for integration tests? I can't imagine there's much value in testing for ex. P13 when we're on P16 in the future, for example. Anyway, LGTM but just dropping some thoughts.

Shaptic (approved):

services/horizon/internal/integration/protocol14_test.go
bartekn: I think you're right. In general, when a new protocol version appears for testing we will want to run all existing integration tests to ensure there're no changes in behaviour. We will also write new tests testing new features. So it seems we always want to upgrade to the latest protocol version available in Core.

To sum it up:
* Maybe we should remove sanity test here? I added it when building integration tests engine but it seems it's not needed in the long term.
* Instead of picking the protocol version directly in `integration.Config` maybe we should just upgrade to the latest available version?

Going to merge the PR to fix the tests but I'll create a new issue to discuss this. Thanks for the comment!

PR #3147, opened 2020-10-21
services/horizon: Change protocol version check error

bartekn:
Changed `Protocol version not supported` to `Error checking protocol version is supported` message because an error can be unrelated to protocol version mismatch (ex. error connecting to ledger backend).

### DIFF
--- a/services/horizon/internal/ingest/processor_runner.go
+++ b/services/horizon/internal/ingest/processor_runner.go
@@ -206,7 +206,7 @@ func (s *ProcessorRunner) RunHistoryArchiveIngestion(checkpointLedger uint32) (i
 		}
 	} else {
 		if err = s.checkIfProtocolVersionSupported(checkpointLedger); err != nil {
-			return changeStats.GetResults(), errors.Wrap(err, "Error checking protocol version is supported")
+			return changeStats.GetResults(), errors.Wrap(err, "Error while checking for supported protocol version")
 		}
 
 		if err = s.validateBucketList(checkpointLedger); err != nil {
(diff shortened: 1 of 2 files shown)


services/horizon/internal/ingest/processor_runner.go
```diff
@@ -206,7 +206,7 @@ func (s *ProcessorRunner) RunHistoryArchiveIngestion(checkpointLedger uint32) (i
 		}
 	} else {
 		if err = s.checkIfProtocolVersionSupported(checkpointLedger); err != nil {
-			return changeStats.GetResults(), errors.Wrap(err, "Protocol version not supported")
+			return changeStats.GetResults(), errors.Wrap(err, "Error checking protocol version is supported")
```
ire-and-curses, line 209: The change makes sense, but maybe we can word it more clearly?
```suggestion
Error while checking for supported protocol version (<= 15)
```
Where the reported number is a variable

abuiles (approved):

services/horizon/internal/ingest/processor_runner.go
```diff
@@ -279,7 +279,7 @@ func (s *ProcessorRunner) RunTransactionProcessorsOnLedger(ledger uint32) (io.St
 	}
 
 	if err = s.checkIfProtocolVersionSupported(ledger); err != nil {
-		return ledgerTransactionStats.GetResults(), errors.Wrap(err, "Protocol version not supported")
+		return ledgerTransactionStats.GetResults(), errors.Wrap(err, "Error checking protocol version is supported")
```
2opremio, line 282:
```suggestion
		return ledgerTransactionStats.GetResults(), errors.Wrap(err, "Error checking whether protocol version is supported")
```

2opremio: And the same for the rest

2opremio:
Alternative suggestion for the text. Otherwise LGTM

2opremio (approved):

2opremio:
@bartekn can you also correct the other instances of the text?

2opremio:
In fact, it may be worth factoring it out in a constant (or error function)

PR #3154, opened 2020-10-22
services/horizon: Fix captive-core's reset of its tmp dir

### LINKED ISSUE #3145: Captive Core's temporary db directory disappears breaking execution
While working on #3144 I encountered the following error in captive core:

```
time="2020-10-21T17:22:50.833Z" level=error msg="Error in ingestion state machine" current_state="resume(latestSuccessfullyProcessedLedger=1)" error="error preparing range: opening subprocess: error running stellar-core: error waiting for `stellar-core new-db` subprocess: could not start `stellar-core [new-db]` cmd: chdir /tmp/captive-stellar-core-9f8cf1da5cf913b7: no such file or directory" next_state=start pid=195 service=ingest
```
which happens in a loop

2opremio:
### What

Fix how the Captive Core backend resets its temporary directory. Previously:

* it removed its contents, but didn't reset its reference to the path.
* it didn't re-create the temporary directory after `close()`

### Why

It made Captive core enter an error loop after a reset was attempted (e.g. after encountering an unexpected state).

Fixes #3145 

### Known limitations

It may be cleaner to just create another runner instead of just expecting `stellarCoreRunner` to work after a call to `close()` (which is how it is today).

### DIFF
--- a/ingest/ledgerbackend/stellar_core_runner.go
+++ b/ingest/ledgerbackend/stellar_core_runner.go
@@ -51,13 +51,6 @@ func newStellarCoreRunner(executablePath, configPath, networkPassphrase string, 
 		nonce:             fmt.Sprintf("captive-stellar-core-%x", r.Uint64()),
 	}
 
-	// Create temp dir
-	dir := runner.getTmpDir()
-	err := os.MkdirAll(dir, 0755)
-	if err != nil {
-		return nil, errors.Wrap(err, "error creating subprocess tmpdir")
-	}
-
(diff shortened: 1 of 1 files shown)


bartekn (approved):

2opremio:
> It may be cleaner to just create another runner instead of just expecting `stellarCoreRunner` to work after a call to `close()` (which is how it is today).

It is much cleaner (and there was a bug preventing this from happening). 

I created #3162 , which overrides this PR (sorry @bartekn )

PR #3166, opened 2020-10-28
xdr: Adds a helper to extract balance IDs from transaction results.

### LINKED ISSUE #3126: txnbuild: add helpers to obtain Claimable Balance IDs
CAP 23 (as part of Stellar Protocol 14 and released with Horizon 1.9.0) , allows creating and claiming Claimable Balances.

Those balances are queried and claimed using their ID which can be obtained:

1. Before submitting the transaction creating the claim. The Claim ID is internally just a Hash which can be inferred even before Core accepts the transaction. A helper method/function would be useful in order to obtain this.
2. After the transaction creating the claim is submitted (i.e. in the transaction result). However, the claimable ID is pretty buried withing the result XDR, making it hard to obtain through the raw API result. A helper would come in handy.

(1) is already addressed by #3122 , (2) still needs to be implemented

Shaptic:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
See title.

### Why
Closes #3126.

### DIFF
--- a/txnbuild/helpers.go
+++ b/txnbuild/helpers.go
@@ -2,6 +2,7 @@ package txnbuild
 
 import (
 	"fmt"
+
 	"github.com/stellar/go/amount"
 	"github.com/stellar/go/strkey"
 	"github.com/stellar/go/support/errors"
@@ -212,4 +213,28 @@ func ParseAssetString(canonical string) (Asset, error) {
 	}
 
 	return asset, nil
+}
+
+// ExtractBalanceId will parse the operation result at `opIndex` within the
+// given `txResult`.
+//
+// If the specified operation index does not point to a successful
+// `CreateClaimableBalance` operation result, this function panics.
+func ExtractBalanceId(txResult xdr.TransactionResult, opIndex int) (string, error) {
+	opResults, ok := txResult.OperationResults()
+	if !ok {
+		return "", errors.New("Failed to retrieve transaction's operation results")
+	}
+
+	if opIndex < 0 || opIndex >= len(opResults) {
+		return "", errors.New("Invalid operation index")
+	}
+
+	result := opResults[opIndex]
+	balanceId, err := xdr.MarshalHex(result.MustTr().MustCreateClaimableBalanceResult())
+	if err != nil {
+		return "", errors.Wrap(err, "Failed to determine balance ID")
+	}
+
+	return balanceId, nil
 }
--- a/txnbuild/helpers.go
+++ b/txnbuild/helpers.go
@@ -193,7 +193,7 @@ func NewValidationError(field, message string) *ValidationError {
 	}
 }
 
-// Parses an asset string in canonical form (SEP-11) into an Asset structure.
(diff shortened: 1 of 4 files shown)


Zagan202 (commented):
LGTM

txnbuild/helpers.go
```diff
@@ -213,3 +214,27 @@ func ParseAssetString(canonical string) (Asset, error) {
 
 	return asset, nil
 }
+
+// ExtractBalanceId will parse the operation result at `opIndex` within the
+// given `txResult`.
+//
+// If the specified operation index does not point to a successful
+// `CreateClaimableBalance` operation result, this function panics.
+func ExtractBalanceId(txResult xdr.TransactionResult, opIndex int) (string, error) {
```
bartekn, line 223: For consistency.
```suggestion
func ExtractBalanceID(txResult xdr.TransactionResult, opIndex int) (string, error) {
```

bartekn (commented):
I think that `ExtractBalanceID` should be a method on `xdr.TransactionResult`. This would allow getting balance ID without having to import `txnbuild`.

txnbuild/helpers.go
Shaptic: I agree! Done in 1ee820fe94f6be62f6e3951d77498803bc24e782. For some reason, I thought the only place I could've put it is within `xdr_generated.go` (which would've been a bad idea obviously), but I figured out where to put it.

Shaptic:
@bartekn I agree! Done in 1ee820f. For some reason, I thought the only place I could've put it is within `xdr_generated.go` (which would've been a bad idea obviously), but I figured out where to put it.

xdr/transaction_result.go
```diff
@@ -30,3 +32,27 @@ func (r TransactionResultPair) OperationResults() ([]OperationResult, bool) {
 func (r TransactionResultPair) InnerHash() Hash {
 	return r.Result.Result.MustInnerResultPair().TransactionHash
 }
+
+// ExtractBalanceId will parse the operation result at `opIndex` within the
+// given `txResult`.
+//
+// If the specified operation index does not point to a successful
+// `CreateClaimableBalance` operation result, this function panics.
+func (r TransactionResult) ExtractBalanceID(opIndex int) (string, error) {
```
bartekn, line 41: LGTM! One suggestion I have is to return `ClaimableBalanceId` and create a helper method that returns hex (`ExtractBalanceIDHex`). This would be consistent with other methods (like `Marshal` -> `MarshalHex`).
```suggestion
func (r TransactionResult) ExtractBalanceID(opIndex int) (ClaimableBalanceId, error) {
```

bartekn (approved):

xdr/transaction_result.go
Shaptic: Hmm, though I'm having trouble envisioning a use case for the raw XDR structure, I've added it as suggested for consistency in 41537b4927d3cf510a2708002da3a1f065edd552!

Issue #3153, opened 2020-10-22
Make integration tests work with Captive core


2opremio:
There was an initial effort at #3114  (which didn't work because of #3137). The current effort is at #3144

2opremio:
This is blocked by #3157

2opremio:
And now it's blocked by https://github.com/stellar/stellar-core/issues/2778

tamirms:
fixed by https://github.com/stellar/go/pull/3291

PR #3167, opened 2020-10-28
services/horizon: Update integration tests to use MANUAL_CLOSE

2opremio:
### What

Use Core's `MANUAL_CLOSE` mode in integration tests  so that we don't have to wait for ledgers to be closed (we close them explicitly).

### Why

It considerably reduces the execution time of integration tests (From ~13min to ~5min).

Also, it paves the way for enabling integration tests for Captive Core (https://github.com/stellar/go/issues/3153) and contributes towards #3037 .

### Known limitations

We are changing the timing in which the production network operates, which makes the tests deviate from a production scenario.

### DIFF
--- a/services/horizon/internal/test/integration/integration.go
+++ b/services/horizon/internal/test/integration/integration.go
@@ -22,11 +22,13 @@ import (
 	"github.com/spf13/cobra"
 
 	sdk "github.com/stellar/go/clients/horizonclient"
+	"github.com/stellar/go/clients/stellarcore"
 	"github.com/stellar/go/keypair"
 	proto "github.com/stellar/go/protocols/horizon"
 	horizon "github.com/stellar/go/services/horizon/internal"
 	"github.com/stellar/go/support/db/dbtest"
 	"github.com/stellar/go/support/errors"
+	"github.com/stellar/go/support/log"
 	"github.com/stellar/go/txnbuild"
 	"github.com/stellar/go/xdr"
 	"github.com/stretchr/testify/assert"
@@ -60,6 +62,7 @@ type Test struct {
 	config    Config
 	cli       client.APIClient
 	hclient   *sdk.Client
+	cclient   *stellarcore.Client
 	container container.ContainerCreateCreatedBody
 	app       *horizon.App
 }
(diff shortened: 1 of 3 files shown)


services/horizon/internal/test/integration/integration.go
```diff
@@ -602,11 +638,32 @@ func (i *Test) CreateSignedTransaction(
...
+	return i.SubmitTransactionXDR(txb64)
+}
+
+func (i *Test) SubmitTransactionXDR(txb64 string) (proto.Transaction, error) {
+	// Core runs in manual-close mode, so we need to close ledgers explicitly
+	// We need to close the ledger in parallel because Horizon's submission endpoint
+	// blocks until the transaction is in a closed ledger
+	go func() {
+		// This sleep is ugly, but a better approach would probably require
+		// instrumenting Horizon to tell us when the transaction was sent to core.
```
bartekn, line 661: Nit: we can start 2 go routines:
* First one with `SubmitTransactionXDR` that will write to unbuffered channel when done.
* Second that closes core ledgers in an infinite loop with a `select` statement with 2 cases: a) time.Sleep, b) on a unbuffered channel from previous point.

This way when submission is done, second go routine will escape (unbuffered channel), but if not it will close another ledger after some time.

services/horizon/internal/test/integration/integration.go
```diff
@@ -602,11 +638,32 @@ func (i *Test) CreateSignedTransaction(
 	return tx, nil
 }
 
+func (i *Test) CloseCoreLedger() error {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	defer cancel()
+	return i.cclient.ManualClose(ctx)
```
bartekn, line 644: Logging can be helpful with debugging.
```suggestion
	i.t.Log("Closing ledger manually...")
	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
	defer cancel()
	return i.cclient.ManualClose(ctx)
```

bartekn (approved):
Awesome!

services/horizon/internal/test/integration/integration.go
2opremio: Ops, I saw this after merging. I will incorporate it later, thanks.

services/horizon/internal/test/integration/integration.go
2opremio: Note taken, thanks.

PR #3171, opened 2020-10-29
services/horizon: Add some manual-close improvements to integration tests

2opremio:
* Poll Core to close ledgers synchronously
* Wait for manual-closing goroutine when submitting transactions

### DIFF
--- a/services/horizon/internal/test/integration/integration.go
+++ b/services/horizon/internal/test/integration/integration.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"os/signal"
 	"strconv"
+	"sync"
 	"syscall"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


services/horizon/internal/test/integration/integration.go
```diff
@@ -703,19 +720,26 @@ func (i *Test) SubmitTransaction(tx *txnbuild.Transaction) (proto.Transaction, e
...
+		// instrumenting Horizon to tell us when the submission is done.
 		time.Sleep(time.Millisecond * 100)
 		if err := i.CloseCoreLedger(); err != nil {
 			log.Fatalf("failed to CloseCoreLedger(): %s", err)
 		}
 	}()
-
-	return i.Client().SubmitTransactionXDR(txb64)
+	tx, err := i.Client().SubmitTransactionXDR(txb64)
+	wg.Wait()
```
bartekn, line 741: I think it won't work if go routine closes the ledger before we actually submit transaction to core. I think we need to do it as explained in: https://github.com/stellar/go/pull/3167#discussion_r513376051.

2opremio: I think that it's enough to manual-close in a loop until `SubmitTransactionXDR` returns.

bartekn: This would work but there is no loop in go routine above.

2opremio: Yep, I will add it.

2opremio:
@bartekn PTAL

services/horizon/internal/test/integration/integration.go
2opremio: Done

bartekn (approved):

PR #3179, opened 2020-10-31
services/horizon: Temporary fix for claimable_balances predicates JSON marshaling

bartekn:
Do not merge. This is not a final fix.

### DIFF
--- a/xdr/json.go
+++ b/xdr/json.go
@@ -99,7 +99,11 @@ func (c ClaimPredicate) toJSON() (claimPredicateJSON, error) {
 		*payload.Not, err = c.MustNotPredicate().toJSON()
 	case ClaimPredicateTypeClaimPredicateBeforeAbsoluteTime:
 		payload.AbsBefore = new(time.Time)
-		*payload.AbsBefore = time.Unix(int64(c.MustAbsBefore()), 0).UTC()
+		absBeforeTime := time.Unix(int64(c.MustAbsBefore()), 0)
+		if absBeforeTime.Year() > 9999 {
+			absBeforeTime = time.Date(9999, time.December, 31, 23, 59, 59, 0, time.UTC)
+		}
+		*payload.AbsBefore = absBeforeTime.UTC()
 	case ClaimPredicateTypeClaimPredicateBeforeRelativeTime:
 		payload.RelBefore = new(int64)
 		*payload.RelBefore = int64(c.MustRelBefore())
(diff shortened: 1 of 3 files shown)


ire-and-curses (approved):

xdr/json.go
```diff
@@ -99,7 +99,11 @@ func (c ClaimPredicate) toJSON() (claimPredicateJSON, error) {
 		*payload.Not, err = c.MustNotPredicate().toJSON()
 	case ClaimPredicateTypeClaimPredicateBeforeAbsoluteTime:
 		payload.AbsBefore = new(time.Time)
-		*payload.AbsBefore = time.Unix(int64(c.MustAbsBefore()), 0).UTC()
```
nyetwurk, line 102: Normally this is because int64 is one of `MaxInt` or `MinInt`.. may want to check that `MustAbsBefore()` doesn't return those.

xdr/json.go
```diff
@@ -99,7 +99,11 @@ func (c ClaimPredicate) toJSON() (claimPredicateJSON, error) {
...
+		absBeforeTime := time.Unix(int64(c.MustAbsBefore()), 0)
+		if absBeforeTime.Year() > 9999 {
```
leighmcculloch, line 103: I'm not sure if we've discussed a future fix for this yet, but I think we could wrap the time.Time type in the object being marshalled and write our own MarshalJSON for the type that uses the Go RFC3339 formatter without checking if the year is above 9999 and conform to the more general ISO8601, which would preserve the original data. Go's erroring here to conform strictly to RFC3339 when marshalling but the Go RFC3339 formatter supports it outside of the time.Time marshaller, and larger than 4 digit years are supported by ISO8601 as long as it is prefixed with + for any year above 9999, which is probably the more important standard we conform to as that's what most programming languages support. RFC3339 is a specific profile of ISO8601 and so more strict but we probably don't need to bind ourselves to that unless we state it in documentation?

leighmcculloch (commented):
Nice fix. Question below about the future longer term fix.

xdr/json.go
bartekn: Thanks for the comment! Yes, I think we're going to write our own `MarshalJSON` but I didn't do that because I wasn't aware that `time.Time` implements ISO8601 outside `MarshalJSON`. Thus I though this is too complicated and would take much more time than the fix in this PR.

bartekn:
Closing because it was just a quick solution to fix SDFs testnet Horizon. The final fix is coming this Monday.

xdr/json.go
bartekn: Hmm, I wrote a small script to check this out and while [one-liner `MarshalJSON`](https://play.golang.com/p/qSrqwZ2LVhh) works like a charm, `UnmarshalJSON` is not easy because [`time.Parse`](https://golang.org/pkg/time/#Parse) actually expects the year to be 0000-9999:
> Years must be in the range 0000..9999

PR #3181, opened 2020-11-02
support/fuzz: Add xdr.ClaimPredicate JSON fuzzing

bartekn:
### What

Add `support/fuzz` package and the first fuzzer for `xdr.ClaimPredicate` JSON methods.

### Why

A bug in `xdr.ClaimPredicate.MarshalJSON` (https://github.com/stellar/go/pull/3179) stopped ingestion in testnet. I wrote a fuzzer to ensure there are no more bugs. Added files to separate `support/fuzz` package to add more fuzzers in the future.

### Known limitations

Do we want to add fuzzers here or to separate repo?


tamirms:
how did you generate the corpus?

bartekn:
I created just one file (`0`) from unit tests. The rest was found by go-fuzz.

support/fuzz/xdr/jsonclaimpredicate/fuzz.go
```diff
@@ -0,0 +1,81 @@
...
+			return false
+		}
+		return validate(**p.NotPredicate)
+	case xdr.ClaimPredicateTypeClaimPredicateBeforeAbsoluteTime:
+		return *p.AbsBefore >= 0
+	case xdr.ClaimPredicateTypeClaimPredicateBeforeRelativeTime:
+		return *p.RelBefore >= 0
+	}
+
+	panic("Invalid type")
```
leighmcculloch, line 80: ❓ What's the motivation for panicing here? Should this return false rather than panicing given that it is not valid?

support/fuzz/xdr/jsonclaimpredicate/fuzz.go
```diff
@@ -0,0 +1,81 @@
...
+	"encoding/json"
+
+	"github.com/stellar/go/xdr"
+)
+
+// Fuzz is go-fuzz function for fuzzing xdr.ClaimPredicate JSON
+// marshaller and unmarshaller. To run:
+//   1. Install go-fuzz: https://github.com/dvyukov/go-fuzz
+//   2. go-fuzz-build
+//   3. go-fuzz
```
leighmcculloch, line 16: Can you place these commands in a `gofuzz.sh` file in the repo root in a similar fashion as `govet.sh`? The shell script can install the go-fuzz tool at a specific version, then run the commands. It'll mean easy for others to get going without having to stumble on this command, and we can include the fuzz test in a CI run, which I think will be important because the build tag means we don't verify this code compiles anywhere in tests.

leighmcculloch (approved):
💯  adding some fuzzing. One suggestion (💡 ), and one question (❓ ) are inline. Regardless 👍 .

In regards to where this code lives I think it should live inside this monorepo. However, I'm not sure about `support/fuzz` being the place. It seems akin to putting tests in a separate package. Like tests I think fuzz tests should live close to the code they are testing. The README for go-fuzz [points out](https://github.com/dvyukov/go-fuzz#random-notes) it is designed for putting in the package it is testing.

support/fuzz/xdr/jsonclaimpredicate/fuzz.go
bartekn: This line should never be executed because `Unmarshal` called before `validate` should error on non-existent union arm and we check all available options in the `switch` above.

support/fuzz/xdr/jsonclaimpredicate/fuzz.go
bartekn: This is a great idea but I'd like to do it in a separate PR. We won't run fuzzer in Circle-CI, we probably need to use our internal infra for this. There are multiple reasons: `go-fuzz` runs indefinitely, we need to update the corpus when new files are generated and it's better to run fuzzing on a powerful machine. I'll create a script but when we agree when we want to run fuzzing.

bartekn:
I moved the file to `xdr/fuzz/jsonclaimpredicate`. The reason for this location is that in case there are multiple fuzzing function per package they will have different `corpus` dirs. It will be much easier to run this way (using default values for CLI params so with `go-fuzz`).

PR #3183, opened 2020-11-02
xdr: Add a custom marshaller for claim predicate timestamp

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add a custom marshaller for representing the `AbsBefore` timestamp.

### Why

Currently, claimable balances which have a `AbsBefore` timestamp beyond year 9999 causes Horizon ingestion to crash. The crash is due to the fact that the `time.Parse()` implementation does not support parsing strings that far into the future.

### Known limitations

* Using a regexp to get around the fact that  `time.Parse()` doesn't accept `time.Time` instances with years outside the range of 0000..9999 range is a bit hacky
* Perhaps it would be cleaner to represent `AbsBefore` as a unix timestamp integer. However, doing such a change now would break the API and it would require changes downstream in the SDK implementations.

### DIFF
--- a/xdr/json.go
+++ b/xdr/json.go
@@ -26,11 +26,16 @@ func (t ISO8601Time) MarshalJSON() ([]byte, error) {
 
 // UnmarshalJSON parses a JSON string into a ISO8601Time instance.
 func (t *ISO8601Time) UnmarshalJSON(b []byte) error {
+	if len(b) < 2 {
+		return fmt.Errorf("%s is too short", string(b))
+	}
+
 	if string(b) == "null" {
 		return nil
 	}
+
 	if b[0] != '"' || b[len(b)-1] != '"' {
-		return fmt.Errorf("%s does not begin and end with double quotes", b)
+		return fmt.Errorf("%s does not begin and end with double quotes", string(b))
 	}
 	trimmed := string(b[1 : len(b)-1])
 	m := reISO8601.FindStringSubmatch(trimmed)
--- a/xdr/json.go
+++ b/xdr/json.go
@@ -10,8 +10,8 @@ import (
 	"github.com/stellar/go/support/errors"
 )
 
-// ISO8601Time is a timestamp which supports parsing dates which have a year outside the 0000..9999 range
-type ISO8601Time struct {
+// iso8601Time is a timestamp which supports parsing dates which have a year outside the 0000..9999 range
+type iso8601Time struct {
 	time.Time
 }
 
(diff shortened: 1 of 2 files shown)


xdr/json.go
```diff
@@ -3,15 +3,62 @@ package xdr
...
+func (t ISO8601Time) MarshalJSON() ([]byte, error) {
+	return []byte(fmt.Sprintf(`"%s"`, t.Format(time.RFC3339))), nil
+}
+
+// UnmarshalJSON parses a JSON string into a ISO8601Time instance.
+func (t *ISO8601Time) UnmarshalJSON(b []byte) error {
+	if string(b) == "null" {
+		return nil
+	}
+	if b[0] != '"' || b[len(b)-1] != '"' {
```
bartekn, line 32: We should check `len(b)`, otherwise it will panic for `len(b) < 2`.

bartekn (approved):

xdr/json.go
```diff
@@ -3,15 +3,62 @@ package xdr
...
 	"fmt"
+	"regexp"
+	"strconv"
 	"time"
+
+	"github.com/stellar/go/support/errors"
 )
 
+// ISO8601Time is a timestamp which supports parsing dates which have a year outside the 0000..9999 range
+type ISO8601Time struct {
```
leighmcculloch, line 14: ❗ By exporting this type we are expanding the API of the xdr package. Given that this type is only used by the not-exported `claimPredicateJSON` type, could it also be not-exported?

xdr/json.go
```diff
@@ -3,15 +3,62 @@ package xdr
...
+	time.Time
+}
+
+// reISO8601 is the regular expression used to parse date strings in the
+// ISO 8601 extended format, with or without an expanded year representation.
+var reISO8601 = regexp.MustCompile(`^([-+]?\d{4,})-(\d{2})-(\d{2})`)
+
+// MarshalJSON serializes the timestamp to a string
+func (t ISO8601Time) MarshalJSON() ([]byte, error) {
+	return []byte(fmt.Sprintf(`"%s"`, t.Format(time.RFC3339))), nil
```
leighmcculloch, line 24: ❗ Do you think we need to include the `+` prefix for years that are over `9999`? Technically this is not ISO88601 without it.

leighmcculloch, line 24: ❗  To safely convert the string to JSON we should probably use `json.Marshal`. It might be safe as is, but I don't personally know all the character combinations that could end up in the string to know if it is safe. Given the json package is an expert in this we could rely on it for that part.

```suggestion
	return json.Marshal(t.Format(time.RFC3339))
```

xdr/json.go
```diff
@@ -3,15 +3,62 @@ package xdr
...
+	month, _ := strconv.Atoi(m[2])
+	day, _ := strconv.Atoi(m[3])
+
+	ts, err := time.Parse(time.RFC3339, "2006-01-02"+trimmed[len(m[0]):])
+	if err != nil {
+		return errors.Wrap(err, "Could not extract time")
+	}
+
+	t.Time = time.Date(year, time.Month(month), day, ts.Hour(), ts.Minute(), ts.Second(), ts.Nanosecond(), ts.Location())
+	return nil
```
leighmcculloch: I think we're doing more lifting than we need to do with the parsing we're doing here. Checking null, and looking for quotes is all work that the `json` package will do for us. We also don't need to parse out the full date with a regex manually given that the time library will do that for us. I think we should leverage `json.Unmarshal` if we can.

This is one approach:
```suggestion
	var s *string
	err := json.Unmarshal(data, &s)
	if err != nil {
		return err
	}
	if s == nil {
		return nil
	}
	parts := strings.SplitN(*s, "-", 2)
	year, err := strconv.ParseInt(parts[0], 10, 64)
	if err != nil {
		return err
	}
	rest, err := time.Parse(time.RFC3339, "0000-"+parts[1])
	if err != nil {
		return err
	}
	t.Time = rest.AddDate(int(year), 0, 0)
	return nil
```

bartekn (approved):

xdr/json.go
tamirms: unfortunately splitting on the first "-" doesn't work on timestamps which have a negative year

PR #3189, opened 2020-11-03
ingest: Render captive core logs as Horizon logs

### LINKED ISSUE #2669: Improve captive-core error reporting
If the captive core fails to start for whatever reason, the relevant error isn't printed and a parsing error is printed instead:

```
2020/06/04 19:55:15 error preparing range: opening getting ledger `from-1`: unmarshalling framed LedgerCloseMeta: unmarshalling XDR frame header: xdr:DecodeUint: EOF while decoding 4 bytes - read: '[]'
```

In this case, the relevant error was:
```
/usr/local/bin/stellar-core: error while loading shared libraries: libpq.so.5: cannot open shared object file: No such file or directory
```

which wasn't printed out until enabling logging (uncommenting the following lines):

https://github.com/stellar/go/blob/05bc2caec65bc795559b70ee1c25afe398148bc1/exp/ingest/ledgerbackend/stellar_core_runner.go#L62-L76

We should also print out errors with proper formatting (i.e. the same logging format as the rest of horizon messages)

Shaptic:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Adds an optional `*log.Entry` instance to the `stellarCoreRunner` and `CaptiveStellarCore` types so that we can process Captive Core's logs and redirect them to the Horizon log with the appropriate formatting / level / etc. The runner appends a `subservice=captive-core` to the entries as outlined in #2669, making filtering easier.

I've mapped `Error` and `Fatal` log levels to `Error`, while everything else maps to its equivalent. This is obviously trivial to change if we want to prioritize differently.

Specifically, a log that used to look like this:
```
GATBN [History WARNING] No writable archives configured, history will not be written.
```
now looks like
```
WARN[2020-11-03T15:40:28.421-08:00] History: No writable archives configured, history will not be written.  pid=59341 service=ingest subservice=stellar-core
```

### Why
This closes #2669, which was re-opened to treat Core output like Horizon logs.

### Known Limitations 
Obviously, if Core's output format changes, the regex may need to be updated.


ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -106,7 +111,7 @@ func (r *stellarCoreRunner) getConfFileName() string {
 	return filepath.Join(r.tempDir, "stellar-core.conf")
 }
 
-func (*stellarCoreRunner) getLogLineWriter() io.Writer {
+func (runner *stellarCoreRunner) getLogLineWriter() io.Writer {
```
tamirms, line 114: typically we try to make the function receiver parameter have the same name across all the function definitions. I think it would be better to call the parameter `r` and rename the pipe reader variable to something else

ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -221,6 +261,10 @@ func (r *stellarCoreRunner) getProcessExitChan() <-chan error {
 	return r.processExit
 }
 
+func (c *stellarCoreRunner) setLogger(logger *log.Entry) {
```
tamirms, line 264: `func (r *stellarCoreRunner) setLogger(logger *log.Entry) {` to be consistent with the other function defiitions

ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -175,6 +182,7 @@ func (c *CaptiveStellarCore) openOfflineReplaySubprocess(from, to uint32) error
 		if err != nil {
 			return errors.Wrap(err, "error creating stellar-core runner")
 		}
+		c.stellarCoreRunner.setLogger(c.log)
```
tamirms, line 185: I think it would be cleaner if you moved this into `stellarCoreRunnerFactory()` because we wouldn't have to remember to call `stellarCoreRunner.setLogger()` when constructing a new `stellarCoreRunner` instance

Shaptic:
All good points; I believe they're resolved now @tamirms.

ingest/ledgerbackend/captive_core_backend.go
Shaptic: Good point, moved to `CaptiveStellarCore.SetLogger()` to wrap the factory method.

ingest/ledgerbackend/stellar_core_runner.go
Shaptic: Definitely an oversight + laziness on my part; resolved!

ingest/ledgerbackend/stellar_core_runner.go
Shaptic: Absolutely, not even sure why it's an `r`.

ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -139,6 +139,20 @@ func NewCaptive(executablePath, configPath, networkPassphrase string, historyURL
 	}, nil
 }
 
+func (c *CaptiveStellarCore) SetLogger(logger *log.Entry) {
```
tamirms, line 142: I think if you call `SetLogger()` `n` times it will call `core.setLogger()` `n-1` times in each invocation of `previousFactory()`

I think you could avoid this by keeping the `log *log.Entry` instance variable as you did before while also modifying the constructor:
```golang
	c := &CaptiveStellarCore{
		archive:           archive,
		executablePath:    executablePath,
		configPath:        configPath,
		historyURLs:       historyURLs,
		networkPassphrase: networkPassphrase,
		waitIntervalPrepareRange: time.Second,
	}
        c.stellarCoreRunnerFactory = func(configPath2 string) (stellarCoreRunnerInterface, error) {
		runner := newStellarCoreRunner(executablePath, configPath2, networkPassphrase, historyURLs)
		runner. setLogger(c.log)
		return runner
	}
	return c, nil
```

ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -42,6 +44,9 @@ type stellarCoreRunner struct {
 	metaPipe    io.Reader
 	tempDir     string
 	nonce       string
+
+	// Optionally, logging can be done to something other than stdout.
+	Log *log.Entry
```
bartekn, line 49: Let's make it private as other fields.
```suggestion
	log *log.Entry
```

ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -118,10 +124,45 @@ func (*stellarCoreRunner) getLogLineWriter() io.Writer {
...
-			fmt.Print(line)
+
+			// If there's a logger, we attempt to extract metadata about the log
+			// entry, then redirect it to the logger. Otherwise, we just use stdout.
+			if r.Log == nil {
+				fmt.Print(line)
+				continue
+			}
+
+			levelRx := regexp.MustCompile(`\[(\w+) ([A-Z]+)\]`)
```
bartekn, line 135: Nit, `Compile` functions are expensive. If you move it outside this loop it will be faster (you need to compile regexp just once).

services/horizon/internal/ingest/main.go
```diff
@@ -182,6 +182,8 @@ func NewSystem(config Config) (System, error) {
 				cancel()
 				return nil, errors.Wrap(err, "error creating captive core backend")
 			}
+			captiveCoreBackend.SetLogger(log)
+			ledgerBackend = captiveCoreBackend
```
bartekn, line 186: Nit: you can leave the old code and write:
```go
ledgerBackend.(*ledgerbackend.CaptiveStellarCore).SetLogger(log)
```

ingest/ledgerbackend/captive_core_backend.go
bartekn, line 142: I think this name better describes what will be logged.
```suggestion
func (c *CaptiveStellarCore) SetStellarCoreLogger(logger *log.Entry) {
```

ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -118,10 +124,45 @@ func (*stellarCoreRunner) getLogLineWriter() io.Writer {
...
+			indices := levelRx.FindStringSubmatchIndex(line)
```
bartekn, line 136: This is too complicated. If you use `FindSubmatch` and rewrite the regexp like in https://play.golang.com/p/mYf-9iMkOCg you won't have to deal with cutting the string yourself.

ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -221,6 +262,14 @@ func (r *stellarCoreRunner) getProcessExitChan() <-chan error {
 	return r.processExit
 }
 
+func (r *stellarCoreRunner) setLogger(logger *log.Entry) {
+	if logger != nil {
+		r.Log = logger.WithField("subservice", "stellar-core")
```
bartekn, line 267: I think we should move expanding the logger with `WithField` to the caller/user of the package. The `subservice` keyword makes sense in Horizon but it's possible to use `ingest` package outside Horizon context.

bartekn (commented):
Looks good! Minor comments and improvements ideas.

ingest/ledgerbackend/captive_core_backend.go
Shaptic: You're absolutely right. I didn't think the `n-1` thing was a very big deal (how many times can a guy call `SetLogger`, anyway? heh) but your design is far cleaner. I'm not sure how I overlooked the fact that I could modify `runner` within the factory method itself. 

Addressed in 675b09b572e842d98053ce2a7c1579250648343a.

ingest/ledgerbackend/stellar_core_runner.go
Shaptic: I initially did this, but the problem is that I still needed the index to trim the line at the end. To do the above, I'd unfortunately have to match twice; do we want to do this? The code is definitely cleaner but slightly less efficient; it's done in 71873746, what do you think?

ingest/ledgerbackend/stellar_core_runner.go
Shaptic: I don't have enough context on this to comment, honestly, so I'm taking your word that this is a better approach! Done in a7908581859bcd4eb55f18d6a949f564f39daf51.

services/horizon/internal/ingest/main.go
Shaptic: Though less lines, is it cleaner? As someone newish to Go, it's quite a bit harder to read, especially with the new change about name + subservice, it becomes: 
```go
ledgerBackend.(*ledgerbackend.CaptiveStellarCore).SetStellarCoreLogger(log.WithField("subservice", "stellar-core"))
```
:open_mouth:

ingest/ledgerbackend/stellar_core_runner.go
bartekn: > but the problem is that I still needed the index to trim the line at the end

You don't have to if you change the regexp to `G[A-Z]{4} \[(\w+) ([A-Z]+)\] (.*)` (as in play.golang example above). The line remaining is also a group match so it's in `[3]`.

ingest/ledgerbackend/stellar_core_runner.go
bartekn: Let me explain: everything in `ingest` package so also `CaptiveStellarCore` backend can be used outside Horizon. It's built this way so devs can develop more complicated apps/ingestion pipelines that can't be developed using Horizon. Because of that they may want to add another field (different than `subservice=stellar-core`) so we should allow clients to set it to the value they want.

services/horizon/internal/ingest/main.go
bartekn: You're right!

tamirms (approved):

ingest/ledgerbackend/stellar_core_runner.go
Shaptic: :facepalm: of course; I didn't notice the last match.

Issue #3191, opened 2020-11-04
Parsing CLI params in `db migrate` command not working


bartekn:
>`2020/11/03 10:49:15 Invalid config: db-url is blank. Please specify --db-url on the command line or set the DATABASE_URL environment variable.`
>
>after set `DATABASE_URL` as environment variable and use `db init`
>
>`2020/11/03 10:50:24 pq: relation "offers" already exists handling 19_offers.sql`

_Originally posted by @tundak in https://github.com/stellar/go/issues/3180#issuecomment-721043334_

I git bisected it to this commit: https://github.com/stellar/go/commit/b0ec1a1b2a7c048def0600e4b62036ac1a378861

PR #3192, opened 2020-11-04
services/horizon/cmd: Fix db-url flag parsing in horizon db commands

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Close https://github.com/stellar/go/issues/3191

The `horizon db migrate` and `horizon db init` commands share the `db-url` command line parameter with the horizon service flags defined in services/horizon/cmd/root.go. The horizon service flags are also initialized in root.go:

```golang
func init() {
	err := flags.Init(rootCmd)
	if err != nil {
		stdLog.Fatal(err.Error())
	}
}
```

To parse the `db-url` param in the migrate and init commands, we need to call `flag.Require()` and `flag.SetValue()`.

The reason why this code wasn't working before is because we were calling on `Require()` and `SetValue()` on `dbURLConfigOption` which was never initialized against any of the db commands or the root command.

### Known limitations

[N/A]

### DIFF
--- a/services/horizon/cmd/db.go
+++ b/services/horizon/cmd/db.go
@@ -25,13 +25,15 @@ var dbCmd = &cobra.Command{
 	Short: "commands to manage horizon's postgres db",
 }
 
-var dbURLConfigOption = support.ConfigOption{
-	Name:      "db-url",
-	EnvVar:    "DATABASE_URL",
-	ConfigKey: &config.DatabaseURL,
-	OptType:   types.String,
-	Required:  true,
-	Usage:     "horizon postgres database to connect with",
+func requireAndSetFlag(name string) {
+	for _, flag := range flags {
+		if flag.Name == name {
+			flag.Require()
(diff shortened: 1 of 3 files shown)


services/horizon/cmd/db.go
```diff
@@ -25,24 +25,25 @@ var dbCmd = &cobra.Command{
...
-	Usage:     "horizon postgres database to connect with",
+func requireAndSetFlag(name string) {
+	for _, flag := range flags {
+		if flag.Name == name {
+			flag.Require()
+			flag.SetValue()
+			return
+		}
+	}
+	log.Fatal("could not find db-url flag")
```
bartekn, line 36:
```suggestion
	log.Fatalf("could not find %s flag", name)
```

bartekn (approved):
Please add CHANGELOG. This will be 1.11.1.

PR #3201, opened 2020-11-09
services/horizon: Avoid waiting for the first checkpoint in captive-core online mode

2opremio:
### What

As discussed with @rokopt and @graydon , there is no need to provide a ledger hash when streaming from ledger before the first checkpoint.

### Why

It allows Horizon to start ingesting before the first checkpoint is reached (which incurs in a ~5min wait when ingesting from the beginning).

### Known limitations
N/A

### DIFF
--- a/ingest/ledgerbackend/captive_core_backend.go
+++ b/ingest/ledgerbackend/captive_core_backend.go
@@ -269,28 +269,30 @@ func (c *CaptiveStellarCore) runFromParams(from uint32) (runFrom uint32, ledgerH
 	}
 
 	if from <= 63 {
-		// For ledgers before (and including) first checkpoint, get/wait the first
-		// checkpoint to get the ledger header. It will always start streaming
-		// from ledger 2.
+		// For ledgers before (and including) first checkpoint, we start streaming
+		// without providing a hash, to avoid waiting for the checkpoint.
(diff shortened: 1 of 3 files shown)


ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -284,28 +284,29 @@ func (c *CaptiveStellarCore) runFromParams(from uint32) (runFrom uint32, ledgerH
 	}
 
 	if from <= 63 {
```
2opremio, line 286: Maybe we should provide the hash when from `from == 63`?

bartekn (approved):

tamirms:
@2opremio I thought this depends on https://github.com/rokopt/stellar-core/commit/3cea6ee9ec5bc08495347e0aebf14f9dd6acb9d4 which, as far as I know, has not been released in stellar core?

2opremio:
> @2opremio I thought this depends on [rokopt/stellar-core@3cea6ee](https://github.com/rokopt/stellar-core/commit/3cea6ee9ec5bc08495347e0aebf14f9dd6acb9d4) which, as far as I know, has not been released in stellar core?

AFAIU, that change is to allow providing `--start-at-ledger 1` (which we still work around)

Issue #3172, opened 2020-10-29
Remove trust on history archives in captive core


bartekn:
Since _run-from_ feature was merged (https://github.com/stellar/go/pull/3163) ingestion system trusts history archives completely in the init stage of `buildState` and `resumeState` when Stellar-Core subprocess is started. The reason is that Stellar-Core requires both the sequence and the hash of the previous ledger we want to start from. Currently we get these values from history archives but we shouldn't trust them.

There are two issue we need to fix:
* In `buildState` we usually don't have any information about the previous ledgers (we do when rebuilding state). In such case a recommended way to obtain the ledger hash is to sync Stellar-Core (`stellar-core run`) and wait for the latest ledger close.
* In `resumeState` we usually have hash of the previously closed ledger. In such case we should allow passing `LedgerHashStore` interface implementation that will allow CaptiveCore get info about known ledger hashes (as suggested by @tamirms). The interface can be very simple:
  ```go
  type LedgerHashesStore interface {
  	// First argument indicates if the hash for the given sequence is available.
  	GetLedgerHash(sequence uint32) (bool, xdr.Hash, error)
  }
  ```

The first issue seems quite complicated, so let's focus on the second case first.

tamirms:
**before** stellar core starts streaming tx meta it will validate the ledger chain from consensus all the way back to the start ledger hash. In other words, `stellar core run` should never emit untrusted ledgers no matter what start ledger hash you provide. So if the ledger hash provided by the history archives is corrupt, the worst impact it will have on horizon is that ingestion will be blocked because the `stellar core run` command will exit with an error before streaming any ledgers.

In the resume state we can prevent this scenario where ingestion is blocked by always using the horizon db to determine ledger hashes. In the build state it is possible that the horizon db is empty so we may not find the ledger hash in the horizon db. But, even if we did have the correct ledger hash, we still need to process the history archives in the build state. So, no matter what, ingestion will be blocked in the build state until the history archives are fixed. Because of this fact, I think we should leave the build state as is.

PR #3203, opened 2020-11-10
ingest/ledgerbackend: Implement db backed ledger store for the captive core backend

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

When running `stellar-core run --start-at-ledger <sequence>` you need to provide an additional  `--start-at-hash` parameter. We were only able to use `--start-at-ledger` values which are checkpoint sequences because  we can obtain the ledger hash for any checkpoint from the history archives.

However, we do have ledger hashes for the ledgers Horizon has previously ingested. This PR uses the horizon database of ingested ledgers as an additional source for obtaining ledger hashes.

### Why

The history archives cannot be fully trusted because there's always the possibility that someone can compromise the history archives either by compromising the S3 bucket which stores the archives or compromising one of the layers above the s3 bucket. When we provide a corrupt ledger hash to stellar core as `--start-at-hash` parameter, stellar core will download the ledger chain and validate all the hashes from consensus back to the start hash. If the start hash is not valid, stellar-core will exit with an error before streaming any ledgers. This means that ingestion will be blocked until the history archives are fixed.

Since stellar core will always verify the ledger chain before emitting any ledgers, we can assume that any ledger hashes ingested into the horizon database are correct. So, in the scenario where the history archives are corrupt, we can avoid blocking ingestion by using ledger hashes found in the Horizon database.

Close https://github.com/stellar/go/issues/3172


ingest/ledgerbackend/ledger_store.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+// Ledger contains information about a ledger (sequence number and hash)
+type Ledger struct {
+	Sequence uint32 `db:"sequence"`
+	Hash     string `db:"ledger_hash"`
+}
+
+// LedgerStore is used to query ledger data from the Horizon DB
+type LedgerStore interface {
```
bartekn, line 15: Nit: emphasise that trust is required.
```suggestion
// TrustedLedgerStore is used to query ledger data from a trusted source.
// The store should contain ledgers verified by Stellar-Core, do not use untrusted
// source like history archives.
type TrustedLedgerStore interface {
```

ingest/ledgerbackend/ledger_store.go
```diff
@@ -0,0 +1,54 @@
...
+	session *db.Session
+}
+
+// NewDBLedgerStore constructs a new DBLedgerStore
+func NewDBLedgerStore(session *db.Session) LedgerStore {
+	return DBLedgerStore{session: session}
+}
+
+// LastLedger returns the highest ledger which is less than `seq` if there exists such a ledger
+func (l DBLedgerStore) LastLedger(seq uint32) (Ledger, bool, error) {
```
bartekn, line 39: I think it belongs to `services/horizon`. Or rename it to `HorizonDBLedgerStore`.

ingest/ledgerbackend/ledger_store.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+// LedgerStore is used to query ledger data from the Horizon DB
+type LedgerStore interface {
+	// LastLedger returns the highest ledger which is less than `seq` if there exists such a ledger
+	LastLedger(seq uint32) (Ledger, bool, error)
+}
+
+// EmptyLedgerStore is a ledger store which is empty
+type EmptyLedgerStore struct{}
```
bartekn, line 21: Do we really need this? I'm wondering if we could just allow passing `nil` `LedgerStore` and then if it's need have a special case to skip checking `LedgerStore` or make `EmptyLedgerStore` private and set it if `LedgerStore=nil`.

This will be especially useful when we decide to use config struct in `CaptiveStellarCore`.

ingest/ledgerbackend/ledger_store.go
```diff
@@ -0,0 +1,54 @@
...
+// Ledger contains information about a ledger (sequence number and hash)
+type Ledger struct {
+	Sequence uint32 `db:"sequence"`
+	Hash     string `db:"ledger_hash"`
+}
+
+// LedgerStore is used to query ledger data from the Horizon DB
+type LedgerStore interface {
+	// LastLedger returns the highest ledger which is less than `seq` if there exists such a ledger
+	LastLedger(seq uint32) (Ledger, bool, error)
```
bartekn, line 17: Still not sure if my suggestion is good but `LastLedger` is a bit confusing.
```suggestion
	LatestLedgerBefore(seq uint32) (Ledger, bool, error)
```

ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -283,6 +295,21 @@ func (c *CaptiveStellarCore) runFromParams(from uint32) (runFrom uint32, ledgerH
...
+			// if we run from a checkpoint sequence
+			// the first ledger we will stream is the next ledger
+			nextLedger = ledger.Sequence + 1
+		} else {
+			// if we run from a non-checkpoint sequence
+			// the first ledger we will stream is the first ledger in the checkpoint range which spans ledger.Sequence
+			nextLedger = historyarchive.PrevCheckpoint(ledger.Sequence) + 1
+		}
+		return
+	}
```
bartekn, line 311: We can make this faster for the case when `from` is after the latest checkpoint if we change the order. Here is how:

Non-checkpoint ledger header will only be available in history archives when next checkpoint is published. This means that if `from` is before the latest checkpoint we'll have to wait up to 5 mins for the checkpoint to be published. To prevent this, we have this special condition here:
```go
	if !historyarchive.IsCheckpoint(from) {
		from = historyarchive.PrevCheckpoint(from)
	}
```
that starts from the previous checkpoint (we know that it exists) and then fast-forward's (in `GetLedger`) to the one we want.

So if we change the order of sections of this function to:
1. Before first checkpoint case.
2. Change `from` to previous checkpoint (if not a checkpoint ledger).
3. Get data from `ledgerStore`.
4. Use ledgerStore or history archive.

We will support this case.

tamirms:

bartekn (approved):
LGTM!

Issue #3200, opened 2020-11-09
Helper for more complicated shutdown dependencies


bartekn:
@tamirms:
> I wonder if we can simplify the shutdown logic further by using contexts and passing them down to each component down to the `stellarCoreRunner` instance. Within `stellarCoreRunner` we can use https://golang.org/pkg/os/exec/#CommandContext to execute stellar core. Perhaps we can make an issue separate from this PR to investigate if using contexts would help.

_Originally posted by @tamirms in https://github.com/stellar/go/pull/3187#discussion_r517981515_

@bartekn

> I looked into using `CommandContext` and I think it won't help much. We still need `close()` method to do some cleanup and we can't do it with context alone. But we can definitely investigate how to write a shutdown code for more complicated object connections. For example, to figure out how to close `CaptiveCore`, `bufferedLedgerMetaReader` and `stellarCoreRunner` I drew a graph to understand the dependencies.

To elaborate, I think a helper that maintains a dependency graph of services that can be shutdown (and in which order) will be helpful in Horizon.

bartekn:
I checked if any of existing packages implement something like this and I came across [`tomb`](https://godoc.org/gopkg.in/tomb.v2). It's much simpler that what I was thinking about, for example no dependency graph but maybe it's actually better - we can create a graph implicitly (by using `tomb` within struct that starts go routines) rather than explicitly (by creating a graph struct with all dependencies).

I will try to create a quick experiment in `CaptiveStellarCore` to check if we can use it.

bartekn:
We just merged another approach here: https://github.com/stellar/go/pull/3278. Closing for now, will reopen if/when we want to revise the solution.

PR #3213, opened 2020-11-10
ingest/ledgerbackend: Fix flaky Captive Core tests

bartekn:
This commit fixes two issues that cause Captive Core tests flaky:
* It was possible that `bufferedLedgerMetaReader` go routine was still running after calling `CaptiveStellarCore.Close()` because it returned only when getting an exit signal from `stellarCoreRunner` (could happen after `CaptiveStellarCore.Close()`). To fix it a new `WaitForClose()` method was created that blocks until go routine returns.
* `bufferedLedgerMetaReader.readLedgerMetaFromPipe()` was retuning error on Stellar-Core process graceful exit. Now it returns an error only when Stellar-Core was closed with an error.

I'm not very satisfied with a solutions here. I think we should research https://github.com/stellar/go/issues/3200 and automate this.

### DIFF
--- a/ingest/ledgerbackend/buffered_meta_pipe_reader.go
+++ b/ingest/ledgerbackend/buffered_meta_pipe_reader.go
@@ -1,0 +1,162 @@
+package ledgerbackend
+
+import (
+	"bufio"
+	"io"
+	"time"
+
+	"github.com/pkg/errors"
+	"github.com/stellar/go/support/log"
+	"github.com/stellar/go/xdr"
+)
+
+const (
+	// The constants below define sizes of metaPipeBufferSize (binary) and
+	// ledgerReadAheadBufferSize (in ledgers). In general:
+	//
+	//   metaPipeBufferSize >=
+	//    ledgerReadAheadBufferSize * max over networks (average ledger size in bytes)
+	//
+	// so that meta pipe buffer always have binary data that can be unmarshaled into
+	// ledger buffer.
+	// After checking a few latest ledgers in pubnet and testnet the average size
+	// is: 100,000 and 5,000 bytes respectively.
+
+	// metaPipeBufferSize defines the meta pipe buffer size. We need at least
+	// a couple MB to ensure there are at least a few ledgers captive core can
+	// unmarshal into read-ahead buffer while waiting for client to finish
+	// processing previous ledgers.
+	metaPipeBufferSize = 10 * 1024 * 1024
+	// ledgerReadAheadBufferSize defines the size (in ledgers) of read ahead
+	// buffer that stores unmarshalled ledgers. This is especially important in
+	// an online mode when GetLedger calls are not blocking. In such case, clients
+	// usually wait for a specific time duration before checking if the ledger is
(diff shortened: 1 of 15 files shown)


ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -502,7 +488,22 @@ func (c *CaptiveStellarCore) Close() error {
...
+
 	return nil
 }
+
+func wrapStellarCoreRunnerError(r stellarCoreRunnerInterface) error {
+	processErr := r.getProcessExitError()
+	if processErr != nil {
+		return errors.Wrap(processErr, "stellar-core process exited with an error")
+	}
+	return errors.New("stellar-core process exited unexpectedly without an error")
```
tamirms, line 508: in the case of an offline replay, after streaming all the ledgers in the catchup range, we expect stellar-core to exit without error, right?

bartekn: Good catch! I fixed it in https://github.com/stellar/go/pull/3213/commits/873797ed441a05937e4fda98e6606da6ac73b1b5 and I was able to simplify `GetLedger`: it no longer needs to handle `getProcessExitChan()` because if Stellar-Core exits, `bufferedLedgerMetaReader` will send an error to the channel.

ingest/ledgerbackend/buffered_meta_pipe_reader.go
```diff
@@ -112,24 +114,46 @@ func (b *bufferedLedgerMetaReader) readLedgerMetaFromPipe() (*xdr.LedgerCloseMet
...
 }
 
-func (b *bufferedLedgerMetaReader) Start(untilSequence uint32) {
+func (b *bufferedLedgerMetaReader) waitForClose() {
+	// If buffer is full, keep reading to make sure it receives
+	// a shutdown signal from stellarCoreRunner.
+loop:
+	for {
+		select {
+		case <-b.c:
```
tamirms, line 135: can this function be simplified to just `<-b.closed` ?

bartekn: Unfortunately no and the reason is in the comment above:
```go
	// If buffer is full, keep reading to make sure it receives
	// a shutdown signal from stellarCoreRunner.
```
To make it more clear `start` go routine closes `b.close` only on error or after receiving process exit signal from Stellar-Core. However, the channel is buffered (up to 20 messages now) so in case it's full we need to read one message to unblock it so `start` can exit.

ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -282,7 +282,6 @@ func (r *stellarCoreRunner) close() error {
 	if r.started {
 		close(r.shutdown)
 		r.wg.Wait()
-		close(r.processExit)
```
bartekn, line 285: This was causing `close of closed channel` error because `r.processExit` is closed right after `cmd.Wait()`.

tamirms (approved):

Issue #3195, opened 2020-11-05
services/horizon/ledger.State package in Horizon is a global variable


bartekn:
`horizon/ledger.State` is a global variable. This is a bad pattern and caused some issues (see https://github.com/stellar/go/pull/3143/#issuecomment-721128315). We should make the state local to `horizon.App` and pass a pointer to it to other parts of the system that need it.

Issue #2674, opened 2020-06-05
Improve a way `db.Session` is passed to actions


bartekn:
The way we pass `db.Session` to actions is currently broken.
* We do so in multiple places instead of one,
* Because of above it's easy to forgot to prepare a session properly (ex. `Clone()` it like in https://github.com/stellar/go/pull/2652).
* We use `context.Context` to pass it to actions.

We should talk about a new design (possibly something like in https://github.com/stellar/go/pull/1861).

2opremio:
@bartekn do you think this still applies, I recall the actions having a helper for this.

PR #3216, opened 2020-11-11
services/horizon: Remove global ledger state

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Close https://github.com/stellar/go/issues/3195

> horizon/ledger.State is a global variable. This is a bad pattern and caused some issues (see #3143). We should make the state local to horizon.App and pass a pointer to it to other parts of the system that need it.

### DIFF
--- a/services/horizon/internal/actions/ledger.go
+++ b/services/horizon/internal/actions/ledger.go
@@ -13,16 +13,16 @@ import (
 )
 
 type GetLedgersHandler struct {
-	LedgerCache *ledger.Cache
+	LedgerState *ledger.State
 }
 
 func (handler GetLedgersHandler) GetResourcePage(w HeaderWriter, r *http.Request) ([]hal.Pageable, error) {
-	pq, err := GetPageQuery(handler.LedgerCache, r)
+	pq, err := GetPageQuery(handler.LedgerState, r)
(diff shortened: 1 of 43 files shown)


services/horizon/internal/actions/ledger.go
```diff
@@ -12,15 +12,17 @@ import (
...
 func (handler GetLedgersHandler) GetResourcePage(w HeaderWriter, r *http.Request) ([]hal.Pageable, error) {
-	pq, err := GetPageQuery(r)
+	pq, err := GetPageQuery(handler.LedgerState, r)
 	if err != nil {
 		return nil, err
 	}
 
-	err = validateCursorWithinHistory(pq)
+	err = validateCursorWithinHistory(handler.LedgerState, pq)
 	if err != nil {
```
bartekn, line 26: I don't think it's an issue now but it's possible that `LedgerState` will change between the first and second call to `CurrentState()`. It can be safer to read the state once and pass it to the functions that need it.

In the future, when https://github.com/stellar/go/issues/2674 is done, we can pass the actual state to the handler so it's not possible to misuse it.

bartekn (approved):

Issue #3235, opened 2020-11-20
clients/stellartoml: parser has become brittle


leighmcculloch:
### What version are you using?

15de5ffd450223e1bc5abf80575d906dc2eabb6e

### What did you do?

I used the Stellar TOML client to retrieve stellar.toml's from a handful of domains.

### What did you expect to see?

I expected to see parsing to succeed.

### What did you see instead?

Some of the responses failed parsing because some of the bool fields are strings in anchor stellar.tomls. These two fields seemed to be the ones most commonly the wrong type.

https://github.com/leighmcculloch/stellar--go/blob/c6e9fc962c928ecea522e24b29b71faf99f3f71e/clients/stellartoml/main.go#L60-L61

These used to parse fine, but since we've added more fields to the client code in 70596c57feb6b51584df13c91f18d53cc4e5dbf4 this now doesn't pass for all stellar.toml hosts.

Clearly domains hosting stellar.toml files need to host valid toml files, but maybe we should be more tolerant to `'true'` vs `true` values. This issue is intended to be space to discuss this.

leighmcculloch:
cc @stellar/integration @stellar/nebula

PR #3236, opened 2020-11-20
tools/sep10-homedomaincheck: add tool to verify SEP-10 v3 home domains

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Short script/tool to validate if a domain is serving a SEP-10 implementation with a home domain set appropriately.

### Why

I wanted to see which domains had adopted the new pattern for including a home domain instead of an anchor name in SEP-10 transactions.

### Known limitations

This isn't in a state that is really intended to be used by anyone else. It is also probably not long term what we need as we normally run our checks as javascript web apps.

cc @JakeUrban You're welcome to take this, modify it, or absorb its functionality into other testing products you're building.


clients/stellartoml/main.go
```diff
@@ -57,8 +57,6 @@ type Currency struct {
 	Image                       string   `toml:"image"`
 	FixedNumber                 int      `toml:"fixed_number"`
 	MaxNumber                   int      `toml:"max_number"`
-	IsUnlimited                 bool     `toml:"is_unlimited"`
-	IsAssetAnchored             bool     `toml:"is_asset_anchored"`
```
leighmcculloch, line 61: I had issues with some domains serving stellar.toml files that used strings for these values instead of bools. I opened a separate issue regarding that: #3235.

tools/sep10-homedomaincheck/main.go
```diff
@@ -0,0 +1,79 @@
...
+func check(homeDomain string) (string, error) {
+	tomlResp, err := stellartoml.DefaultClient.GetStellarToml(homeDomain)
+	if err != nil {
+		return "", err
+	}
+	if tomlResp.WebAuthEndpoint == "" {
+		return "", errors.New("sep-10 not supported")
+	}
+	queryParams := url.Values{}
+	queryParams.Set("account", "GAAZI4TCR3TY5OJHCTJC2A4QSY6CJWJH5IAJTGKIN2ER7LBNVKOCCWN7")
```
leighmcculloch, line 61: This can be any account on the pubnet network. I just hard coded a value which is the first account on the network.

JakeUrban: Probably makes sense to just use something like the Go SDK's equivalent of `Keypair.random()`

JakeUrban (approved):
This is great, thanks. I can improve upon this script in another PR or I can just push to your branch, either way.

JakeUrban:
Looks like this PR is unmergeable until it has a CHANGELOG.md

tools/sep10-homedomaincheck/main.go
leighmcculloch: Maybe. Some SEP-10 servers may only allow accounts that exist, it is technically permitted to do that. This account is fine to hardcode, it is the first account that was ever created on the network and it still exists and it can't go anywhere because its signers are locked.

leighmcculloch:
>I can improve upon this script in another PR or I can just push to your branch, either way.

Feel free to push directly to this branch. If we feel like we want to merge this in the end we can, although I was mostly just sharing it with you conveniently.

leighmcculloch:
@JakeUrban Do you think we need this long term? I think long term it will be much better if this problem is solved by the existing anchor server tests. Which means ultimately closing this. What do you think?

JakeUrban:
Thats fine by me, go ahead.

PR #3237, opened 2020-11-20
services/horizon: Trivial improvements to Stellar Core log parsing

Shaptic:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Improves log output:
 - `Infof` -> `Info` for non-formatted strings
 - Lines w/o pubkey prefix are parsed correctly

### Why
When the line from Stellar Core's logs contains `%`, it's interpreted as a formatting directive by the output function (e.g. `Infof`), causing extraneous output. It should just use `Info` to avoid this.

Not all Core lines begin with a public key prefix `G[AB12]` (e.g. ones that begin with `<startup>` instead), so we shouldn't enforce this in the regex.

### DIFF
--- a/ingest/ledgerbackend/stellar_core_runner.go
+++ b/ingest/ledgerbackend/stellar_core_runner.go
@@ -162,7 +162,7 @@ func (r *stellarCoreRunner) getLogLineWriter() io.Writer {
 				} else {
 					r.Log.Info(line)
 				}
-			} else if len(line) > 1 { // non-empty
+			} else {
 				r.Log.Info(line)
 			}
 		}
--- a/ingest/ledgerbackend/stellar_core_runner.go
+++ b/ingest/ledgerbackend/stellar_core_runner.go
@@ -160,10 +160,10 @@ func (r *stellarCoreRunner) getLogLineWriter() io.Writer {
 				if writer, ok := levelMapping[strings.ToUpper(level)]; ok {
 					writer("%s: %s", category, line)
 				} else {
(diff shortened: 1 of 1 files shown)


ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -160,10 +160,10 @@ func (r *stellarCoreRunner) getLogLineWriter() io.Writer {
 				if writer, ok := levelMapping[strings.ToUpper(level)]; ok {
 					writer("%s: %s", category, line)
 				} else {
-					r.Log.Infof(line)
+					r.Log.Info(line)
 				}
-			} else {
-				r.Log.Infof(line)
+			} else if len(line) > 1 { // non-empty
```
tamirms, line 165: is this check necessary? a few [lines](https://github.com/stellar/go/pull/3237/files#diff-a64b634d433438f2ef9fde13a2a782d626d326767fab4c87bb8f60c4f546e1a9R136-R138) above there is an empty string check:

```golang
			if line == "" {
				continue
			}
```

Shaptic: Because we do `.ReadString("\n")`, the line will always be non-empty. This sometimes results in log lines like:
```
time="2020-etc." level=info msg="\n" pid=276827 service=ingest subservice=stellar-core
```
when writing to a file, which clutters more-meaningful lines.

tamirms (approved):

PR #3243, opened 2020-11-23
horizon/ingest: Add a writeup on how the FSM works.

Shaptic:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
As I was working through digesting the finite state machine involved in ingestion, I took notes to keep it all straight in my head. It's almost done, with just a few minor additions pending. 

A proofread to make sure that **I** properly understand it would be amazing.

### Why
Maybe it'll be valuable to the next person who has to ingest (heh) the FSM for the first time.

The rendered doc can be viewed by [clicking here directly](https://github.com/stellar/go/blob/e10b10c288c235237ec9d1e944a93d628a5a2c55/services/horizon/internal/ingest/docs.md).


services/horizon/internal/ingest/docs.md
```diff
@@ -0,0 +1,130 @@
...
+
+With the range prepared, only one other primary state transition is possible. If the last-known ledger of the Core backend is outdated relative to the above `ingestLedger`, we'll block until the requisite ledger is seen and processed by Core. **Next state**: [`resume`](#resume-state) again, with the last-processed ledger set to whatever is last-known to Core.
+
+Otherwise, we can actually turn the ledger into time-series data: this is exactly the responsibility of `RunAllProcessorsOnLedger` and all of its subsequent friends. The deltas for the ledger(s) are ingested into the time-series db, then verified.
+
+**Next state**: [`resume`](#resume-state) again, except now targeting the *next* ledger.
+
+
+## `historyRange` state
+The purpose of this state is to ingest a particular ledger range into the time-series database.
```
bartekn, line 105: One important difference between `historyRange` and `resume` other than ledger range vs. single ledger is that history skips tx meta updates (which make it faster). We can skip meta updates because the following transition will always be `start -> build` so the state will be rebuilt anyway.

services/horizon/internal/ingest/docs.md
```diff
@@ -0,0 +1,130 @@
...
+  - `historyRangeState`
+  - `reingestHistoryRange`
+  - `waitForCheckpoint`
+
+#### Definitions
+There are some important terms that need to be defined for clarity as they're used extensively in both the codebase and this breakdown:
+
+  - the `historyQ` field corresponds to **historical *time-series* data**, *not* to the history *archives* (which only contain *cumulative* data); I prefer to refer to it as the `TimeSeriesDB`.
+  - the `lastIngestedLedger` thus corresponds to the last ledger that Horizon  ingested into the time-series tables
+  - the `lastHistoryLedger`, however, corresponds to the last ledger that Core+Horizon are *aware* of, not necessarily *ingested*; I refer to it as the `lastKnownLedger`
```
bartekn, line 18: Just wanted to clarify a difference between these two values. `lastHistoryLedger` is the last ledger sequence from `history_ledgers` table, `lastIngestedLedger ` is the value from `key_value_store`. Here are the cases when these two won't be equal:
* Before version 1.0.0 Horizon did not track state/tx meta. `lastIngestedLedger` was introduced to better track when history/state data are in sync.
* `historyRange` state ingests ledgers into `history_ledgers` but does not update `key_value_store`. Because we have two values we can detect when state data is behind history data in a DB.

One of the most important job of FSM is to make sure these two values are equal.

services/horizon/internal/ingest/docs.md
```diff
@@ -0,0 +1,130 @@
...
+There are a few possible branches in this state.
+
+##### DB upgrade or fresh start
+The "happiest path" is the following: either the ingestion database is empty, so we can start purely from scratch, or the ingestion database is outdated, meaning it needs to be upgraded and so can effectively be started from scratch after catch-up.
+
+This branches differently depending on the last known ledger:
+
+  - If it's newer than the last checkpoint, we need to wait for a new checkpoint to get the latest cumulative data. Note that though we probably *could* make incremental changes from block to block to the cumulative data, that would be more effort than it's worth relative to just waiting on the next history archive to get dumped. **Next state**: [`waitForCheckpoint`](#waitforcheckpoint-state).
+
+  - If it's older, however, then we can just grok the missing gap (i.e. until the next checkpoint) and build up (only) the time-series data. **Next state**: [`historyRange`](#historyrange-state).
```
bartekn, line 35: Small fix: we _catchup_ up to the latest checkpoint, not the next checkpoint. Next checkpoint would work too but it'd be slower because `historyRange` don't update tx meta.
```suggestion
  - If it's older, however, then we can just grok the missing gap (i.e. until the latest checkpoint) and build up (only) the time-series data. **Next state**: [`historyRange`](#historyrange-state).
```

services/horizon/internal/ingest/docs.md
```diff
@@ -0,0 +1,130 @@
...
+
+In the other cases (matching last-known and last-checkpoint ledger, or no last-known), **next state**: [`build`](#build-state).
+
+##### Otherwise
+If we can't have a clean slate to work with, we need to fix partial state. Specifically,
+
+  - If the last-known ledger is ahead of the last-ingested ledger, then Horizon is aware of much more data than it has processed. Here, we'll reset the time-series database and start over. **Next state**: [`start`](#start-state), with `lastIngestedLedger == 0`.
```
bartekn, line 42:
```suggestion
  - If the last-known ledger is ahead of the last-ingested ledger, then state data is behind history data in a DB. Here, we'll reset the time-series database and start over. **Next state**: [`start`](#start-state), with `lastIngestedLedger == 0`.
```

services/horizon/internal/ingest/docs.md
```diff
@@ -0,0 +1,130 @@
...
+
+
+### Properties
+This state tracks the:
+
+  - `checkpointLedger`, which is Horizon's last-known (though possibly-unprocessed) checkpoint ledger, and 
+  - `stop`, which optionally (though *universally*) transitions to the [`stop`](#stop-state) after this state is complete.
+
+### Process
+There's an important bit about this state: only one node should ever build, globally. Hence, there are a few checks at the start to ensure this.
```
bartekn, line 60: In general: only one ingesting node should ever write to a DB at the same time. It's not only about `build`.

bartekn (commented):
Looks good, added some clarifications. I wonder if we should put this in `fsm.go` godoc rather than a separate file.

services/horizon/internal/ingest/docs.md
Shaptic: Can it be more precise than "a" DB? I definitely want to know (or be reminded of) *which* DB here matters (more accurately a table rather than a DB, right?). Would saying something like
> then, the data in the ledger state table is lagging behind the ingested data in the time-series table

be precise / accurate?

services/horizon/internal/ingest/docs.md
Shaptic: What do you mean by tx meta being skipped? I see the following loop:

```go
for cur := h.fromLedger; cur <= h.toLedger; cur++ {
	if err = runTransactionProcessorsOnLedger(s, cur); err != nil {
		return start(), err
	}
}
```
Isn't this how txmeta for a particular ledger gets processed?

services/horizon/internal/ingest/docs.md
Shaptic: Great caveat, integrated!

services/horizon/internal/ingest/docs.md
Shaptic: Awesome details, I added them in!

Shaptic:
By keeping this in an `.md` file, @bartekn, you can actually click states to jump from one to another to follow the FSM around. There's a good amount of inline code comments in `fsm.go` itself to supplement it, I think?

services/horizon/internal/ingest/docs.md
bartekn: I think naming is far from ideal (but we couldn't figure out better names) so can be confusing.

Here's a quick summary of processor types: https://godoc.org/github.com/stellar/go/ingest#hdr-Processors. In Horizon we have two pipelines: [transaction](https://github.com/stellar/go/blob/423317f2950fed914917519117762a491314e5f0/services/horizon/internal/ingest/processor_runner.go#L122) pipeline that inserts tx, ops, effects data but (mostly) ignores txmeta and [change](https://github.com/stellar/go/blob/423317f2950fed914917519117762a491314e5f0/services/horizon/internal/ingest/processor_runner.go#L100) pipeline that update states tables (so use txmeta only - operates on [`io.Change`](https://godoc.org/github.com/stellar/go/ingest/io#Change) objects).

services/horizon/internal/ingest/docs.md
bartekn: I mean Horizon DB. It looks like naming in `ingest` package is not clear (maybe we should improve it?). To clarify:
* State tables - all tables that contain current state like `accounts`, `offers`.
* Change processors - processors ingesting tx meta that update state tables (accept `io.Change` - not related to a particular transaction, it's a transition of a ledger entry from one state to another).
* History tables - all tables that contain historical data like `history_transactions`, `history_operations`.
* Transaction processors - processors ingesting history table (accept `io.LedgerTransaction`).

Shaptic: Superbly clear & integrated into the doc, ty. :)

services/horizon/internal/ingest/docs.md
Shaptic: Another excellent clarification that ties into the below one, too, thank you!

services/horizon/internal/ingest/docs.md
```diff
@@ -0,0 +1,147 @@
...
+  - `historyRange`
+  - `reingestHistoryRange`
+  - `waitForCheckpoint`
+
+There is also the `stressTest` state, but that exists in its own world and is used only for testing ingestion, as its name implies.
+
+#### Definitions
+There are some important terms that need to be defined for clarity as they're used extensively in both the codebase and this breakdown:
+
+  - the `historyQ` field corresponds to **historical *time-series* data**, *not* to the history *archives* (which only contain *cumulative* data); I prefer to refer to it as the `TimeSeriesDB`.
```
tamirms, line 17: `historyQ` is the interface for all database operations. we use `historyQ` to update / query both state tables and history tables

services/horizon/internal/ingest/docs.md
```diff
@@ -0,0 +1,147 @@
...
+  - the `lastIngestedLedger` thus corresponds to the last ledger that Horizon  ingested into the time-series tables (coming from the `key_value_store` table)
```
tamirms, line 18: all the data in the horizon state tables are reflect the state of the stellar network at ledger `lastIngestedLedger`

services/horizon/internal/ingest/docs.md
```diff
@@ -0,0 +1,147 @@
...
+  - the `lastHistoryLedger`, however, corresponds to the last ledger that Core+Horizon are *aware* of, not necessarily *ingested* (coming from `history_ledgers` table); I'll usually refer to it as the `lastKnownLedger`
```
tamirms, line 19: the history tables consist of all transactions, operations, effects, etc which occur at a given ledger. this data can be called described as following time series where the ledger sequence is the unit of time. the history data can be ingested independently of the state tables (which are cumulative aggregations) using the `horizon db reingest range` command. `lastHistoryLedger` refers to the last ledger which was ingested into the history tables. Because history tables can be ingested independently it is possible that `lastHistoryLedger` does not equal `lastIngestedLedger`.

services/horizon/internal/ingest/docs.md
```diff
@@ -0,0 +1,147 @@
...
+
+#### Tables
+Within the Horizon database, there are a number of tables touched by ingestion that are worth differentiating explicitly. With these in mind, the subsequently-described states and their respective operations should be much clearer.
+
+The database contains:
+
+  - **History tables** -- all tables that contain historical time-series data, such as `history_transactions`, `history_operations`, etc.
+  - **Transaction processors** -- processors that ingest the history tables (described by the `io.LedgerTransaction` interface).
+  - **State tables** -- all tables that contain the current cumulative state, such as accounts, offers, etc.
+  - **Change processors** -- processors ingesting tx meta (time-series data) that update state tables. These aren't related to a particular *transaction*, but rather describe a *transition* of a ledger entry from one state to another (described by the `io.Change` interface).
```
tamirms, line 39: change processors can also ingest from the history archives, not just tx meta. in other words, change processors are fed `io.Change` objects. It is possible to extract changes from tx meta. It is also possible to extract changes from the history archives. a change extracted from tx meta represents the delta between one ledger and it's successor (e.g. going from `n` to `n+1`). a change extracted from the history archive represents the delta between ledger 0 (the empty state) and the checkpoint ledger.

bartekn:
@Shaptic just wanted to confirm: do you want to merge it when all the states are described or sooner? Not sure if it's urgent to review.

services/horizon/internal/ingest/docs.md
Shaptic: Awesome, makes total sense: `io.Change` is for any deltas. Updated :heavy_check_mark:

services/horizon/internal/ingest/docs.md
Shaptic: Big oversight on my part, thank you for noting this. Gotta say, though, it's not great naming :sweat_smile:

services/horizon/internal/ingest/docs.md
Shaptic: Superb, noted.

services/horizon/internal/ingest/docs.md
Shaptic: Oh okay, that's subtle but makes it much clearer. Good clarification. :)

Shaptic:
@bartekn I think all of the states *are* described, no? I would like to fill out the *Ingestion* and *Range Preparation* sections eventually, but it would be best to merge the rest now as I need deeper understanding (or someone else could fill them in, heh).

ire-and-curses:
Yeah this looks good to me - let's go ahead and merge what we've got.

ire-and-curses (approved):
Yeah this looks good to me - let's go ahead and merge what we've got.

PR #3258, opened 2020-11-27
ingest/ledgerbackend: Handle user initiated shutdown during catchup

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Update shutdown code to handle user initiated shutdown. Update shutdown code and go routines to use [`tomb`](https://godoc.org/gopkg.in/tomb.v1).

### Why

Shutdown code was getting too complicated. We were missing some obvious cases (like user initiated shutdown), run multiple times into problems connected to shutdown (ex. closing a closed channel) and in general code was messy. In https://github.com/stellar/go/issues/3200 I found [`tomb`](https://godoc.org/gopkg.in/tomb.v1) and decided to give it a try. It has simple yet powerful API that covers sending kill signals, wait for a go routine to return and handling errors. The new code is much clearer than my first attempt in https://github.com/stellar/go/pull/3258/commits/f876d60e8971181f0b04f4d424733d807119b5f1 that requires three extra methods on `stellarCoreRunner`. It makes methods like `bufferedLedgerMetaReader.waitForClose` (renamed to `Close` as it now allows user initiated close too) easier to understand. Finally, tests are also easier to write.

### Known limitations

Let me know if you like `tomb`. If it's OK it would be great to use it in Horizon.


tamirms:
@bartekn looks like there's a v2 of tomb https://github.com/go-tomb/tomb/tree/v2

bartekn:
>@bartekn looks like there's a v2 of tomb https://github.com/go-tomb/tomb/tree/v2

Yes, I saw that but we already have `v1` in our deps and `v2` doesn't add anything that we need here. We may upgrade in the future if necessary.

ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -387,11 +390,24 @@ func (c *CaptiveStellarCore) PrepareRange(ledgerRange Range) error {
 
 	for {
 		select {
-		case <-c.stellarCoreRunner.getProcessExitChan():
-			// Return only in case of Stellar-Core process error. Normal exit
-			// is expected in catchup when all ledgers sent to a buffer.
-			processErr := c.stellarCoreRunner.getProcessExitError()
-			if processErr != nil {
+		case <-c.stellarCoreRunner.getTomb().Dead():
```
tamirms, line 393: @bartekn I forgot why we need to wait for the first ledger. Could you explain why this loop is necessary?

bartekn: We want to block `PrepareRange` until there's something we can stream to the user. This is especially useful for unbounded ranges that in which `GetLedger` is blocking to make the process faster. Also allows creates some metrics connected to init time.

tamirms: ok, can we call `GetLedger(from)` in `openOfflineReplaySubprocess()` and `openOnlineReplaySubprocess()`? we do this in `openOnlineReplaySubprocess()` already in the case we need to fast forward but we might as well do it in all cases if we want `PrepareRange` to block until the first ledger is available.

i think if we call `GetLedger(from)` in those functions we can remove this loop

bartekn: Good call but instead maybe we can remove it from `openOnlineReplaySubprocess` and move it to `PrepareRange`. I missed that we do it in `openOnlineReplaySubprocess` but we don't check if core process is dead so shutdown is broken in this case. Going to do that in the following commit.

tamirms:
@bartekn I think there are potentially thread safety issues if Close() is called concurrently with GetLedger() or PrepareRange(). If we allow user initiated shutdown, Close() can be called at any time regardless of the state of the ingestion system

bartekn:
> @bartekn I think there are potentially thread safety issues if Close() is called concurrently with GetLedger() or PrepareRange(). If we allow user initiated shutdown, Close() can be called at any time regardless of the state of the ingestion system

I had this in mind when working on this PR. If you see any issues connected to calling `Close()` when `PrepareRange()` or `GetLedger()` is running can you add comments inline?

ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -474,7 +490,7 @@ loop:
 			return false, xdr.LedgerCloseMeta{}, nil
```
tamirms, line 490: If Close() is called concurrently with this function is there a possible race condition where the ledger buffer is set to nil right before it is accessed here?

bartekn: Good catch, I focused on shutdown code and forgot about the obvious things. Fixed in https://github.com/stellar/go/pull/3258/commits/648ebf501f9d90799441108bd94ee8e77b6795fa.

bartekn:
I run the code again after recent changes to ensure it's working ok but I noticed a very slow catchup stage. I fixed it in https://github.com/stellar/go/pull/3258/commits/6acf8f84add68478c424c550b5187f58ea754f6e. The problem was connected to the fact that only one ledger was fetched from the buffer in `PrepareRange`, I changed it to empty the buffer and then sleep (also, decreased sleep time from 1s to 100ms).

tamirms:
@bartekn I think the code is thread-safe now. the only issue I see is that calling `Close()` may take a long time because it will block on acquiring the lock until `PrepareRange()` completes

bartekn:
> the only issue I see is that calling Close() may take a long time because it will block on acquiring the lock until PrepareRange() completes

Good call again, fixed in https://github.com/stellar/go/pull/3258/commits/0505023ad924564f9fb883cc63f08e01126ac8dc.

tamirms:
@bartekn I think there is another scenario which is not covered. Consider if the ingestion system's shutdown method is called right before the state machine go routine calls `PrepareRange()`. I think there's a possible race condition where the ingestion system shutdown method calls `Close()` right before the state machine go routine calls `PrepareRange()`.  While in `PrepareRange()` the function continues to execute as normal and it would be unaware that the system is shutting down.

bartekn:
@tamirms you are right but except an obvious idea which is disallow reusing `CaptiveStellarCore` after closing that would require a larger refactor I'm not sure how to solve this. I'd vote for creating an issue for this and other issues connected to shutdown and fix them after beta release. What do you think?

tamirms:
@bartekn sounds good, I think I have an idea on how to tackle it but I think I'd need a day or two to verify the idea. Fixing it in the beta release makes sense

ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -355,24 +356,51 @@ func (c *CaptiveStellarCore) PrepareRange(ledgerRange Range) error {
...
+				return errors.Wrapf(err, "Error fast-forwarding to %d", ledgerRange.from)
+			}
+
+			// If nextLedger > ledgerRange.from then ledgerRange.from is cached.
+			if c.nextLedger > ledgerRange.from {
+				break waitloop
+			}
 		}
+
 		time.Sleep(c.waitIntervalPrepareRange)
```
2opremio, line 404: Do we really need an explicit sleep here?

I would (at the very least) use a ticker and intertwine the check with the other channels in the `select` statement.

Even better, It may even be possible to remove the wait and condition by using an asynchronous go-routine for fast-forwarding. 

It may seem outside of the scope of this PR, but it affects the shutdown wait time.

ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -355,24 +356,51 @@ func (c *CaptiveStellarCore) PrepareRange(ledgerRange Range) error {
...
+			default:
 				return errors.Wrap(processErr, "stellar-core process exited with an error")
 			}
 		default:
 		}
-		// Wait for the first ledger or an error
-		if len(c.ledgerBuffer.getChannel()) > 0 {
-			break
+		// Wait/fast-forward to the expected ledger or an error. We need to check
+		// buffer length because `GetLedger` may be blocking.
```
2opremio, line 391: I think it would make sense to add a context to `GetLedger()` which we can cancel on shutdown.

ingest/ledgerbackend/captive_core_backend.go
bartekn: We can try it out for sure! Can you create a new issue? I'm not sure if we'll be able to test it before beta.

ingest/ledgerbackend/captive_core_backend.go
bartekn: I really wanted to avoid using `context.WithCancel` here because it doesn't provide two things:
1. wait until cancel request has been handled so the caller can be sure that it's safe to exit (like: core process exit, files removed, or go routine in a buffer returns),
2. easy way to pass a final error value if any.

`tomb` provides mechanism for both and because of this: simplifies entire process. I'm open for discussion, can we talk in a new issue?

bartekn (commented):
@Shaptic you asked for some comments explaining how this code works. I recommend reading [this article](https://blog.labix.org/2011/10/09/death-of-goroutines-under-control) (I know you did) but as requested added a few comments. As I mentioned in the other comment, the main advantage of `tomb` is that it allows sending cancellation signal but also allow propagating errors back to the caller (and original error is not overwritten, this is useful when there is user-initiated shutdown so you can send `Cancelled` error to detect it) and allows waiting for the go routine to actually return. Obviously it's possible to built similar solution using channels, wait groups but `tomb` gives this out of the box. Read on...

ingest/ledgerbackend/captive_core_backend.go
2opremio: Sure. I will open a new issue.

ingest/ledgerbackend/captive_core_backend.go
2opremio: Sure

ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -291,36 +294,38 @@ func (r *stellarCoreRunner) getMetaPipe() io.Reader {
...
+	return r.tomb
 }
 
 func (r *stellarCoreRunner) close() error {
 	var err1, err2 error
 
+	if r.tomb != nil {
+		// Kill tomb with context.Canceled. Kill will be called again in start()
+		// when process exit is handled but the error value will not be overwritten.
+		r.tomb.Kill(context.Canceled)
```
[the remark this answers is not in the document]
Shaptic: The answer to this is probably painfully-obvious, but by "user-initiated shutdown," we generally mean something akin to Ctrl+C and/or via a code path (e.g. calling `app.Close()`) right?

ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -340,11 +333,19 @@ func (c *CaptiveStellarCore) runFromParams(from uint32) (runFrom uint32, ledgerH
...
+// Returns `context.Cancelled` if `Close` was called from another Go routine.
 func (c *CaptiveStellarCore) PrepareRange(ledgerRange Range) error {
+	// c.mutex is unlocked on error and before waitloop to allow calling Close()
+	// to intrerupt the process.
+	c.mutex.Lock()
+
 	// Range already prepared
-	if prepared, err := c.IsPrepared(ledgerRange); err != nil {
+	if prepared, err := c.isPrepared(ledgerRange); err != nil {
+		c.mutex.Unlock()
```
Shaptic, line 345: Would it be safer / cleaner to do something like

```go
defer func() {
  if !unlocked { 
    c.mutex.Unlock()
  }
}
```
instead (and set `unlocked=true` below before the `waitloop`)? Or, alternatively, moving the waitloop to its own function would let us just `defer c.mutex.Unlock()` directly here.

It just seems like it could be really easy for someone (like me lol) to edit this code in the future, then forget to unlock the mutex and be in a dangerous state.

bartekn:
Closing in favour of: https://github.com/stellar/go/pull/3278.

PR #3259, opened 2020-12-01
services/horizon/internal: Determine if Stellar Core is synced to network

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

### Why

### Known limitations

[TODO or N/A]


services/horizon/internal/ledger/main.go
```diff
@@ -12,10 +12,9 @@ import (
 // Status represents a snapshot of both horizon's and stellar-core's view of the
 // ledger.
 type Status struct {
-	CoreLatest       int32  `db:"core_latest"`
-	HistoryLatest    int32  `db:"history_latest"`
-	HistoryElder     int32  `db:"history_elder"`
-	ExpHistoryLatest uint32 `db:"exp_history_latest"`
+	HistoryLatest    int32
+	HistoryElder     int32
+	ExpHistoryLatest uint32
```
2opremio, line 17: Are we sure this won't break DB queries?

tamirms: those tags were only used in testing code: [services/horizon/internal/test/t.go](https://github.com/stellar/go/pull/3259/files/b68a74dd393007616b78d3fbed62c2d7c785007b#diff-6bc3c32ff5251e1fc7b8f32f4b34c90cf186ace2e7d9b5ce20c96a226325c5d8L150) . I couldn't find any application code which uses the db tags

services/horizon/internal/app.go
```diff
@@ -357,28 +368,35 @@ func (a *App) UpdateStellarCoreInfo() {
...
 	}
-
-	resp, err := core.Info(context.Background())
 	if err != nil {
-		log.Warnf("could not load stellar-core info: %s", err)
+		log.WithField("err", err.Error()).Warnf(
+			"failed to obtain stellar-core info after %v retries",
+			maxUpdateLedgerStateAttempts,
+		)
+		a.coreSettings.setNotSynced()
```
Shaptic, line 384: Should we assume "not synced" and unset it if this succeeds rather than assuming "synced" and setting if it fails?

tamirms: if the request succeeds we will set the synced field based on the response structure of the response. this happens a few lines below when we call `	a.coreSettings.set(coreInfo)`.

services/horizon/internal/app.go
```diff
@@ -489,7 +507,7 @@ func (a *App) init() error {
 			KeyPath:  a.config.TLSKey,
 		}
 	}
-	a.webServer, err = httpx.NewServer(config, routerConfig, a.ledgerState)
+	a.webServer, err = httpx.NewServer(config, routerConfig, a, a.ledgerState)
```
2opremio, line 510: If you pass the full application then you probably don't need to explicitly pass the ledger state.

However, I think it would be better if you narrow down what you pass instead of passing the full app.

I recall intentionally removing the app dependency from the server for clarity.

services/horizon/internal/app.go
```diff
@@ -30,17 +30,30 @@ import (
 	"github.com/stellar/go/support/log"
 )
 
+const (
+	infoRequestTimeout           = time.Millisecond * 500
+	maxUpdateLedgerStateAttempts = 3
```
2opremio, line 35: This is a bit too implicit (e.g. why not 4 or 5?)

I don't think we should make it configurable either (i.e. `ERR_TOO_MANY_KNOBS`), but maybe we should continuously retry in the background (adding a backoff) instead of setting a _magical_ number of retries.

2opremio (approved):
Comment about retries (but not a showstopper). I am missing some context as to why this is needed though.

PR #3260, opened 2020-12-01
ingest/ledgerbackend: Remove returning error on Stellar-Core process exit during catchup

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

Fixes a bug introduced in https://github.com/stellar/go/commit/a10c00015768d34d017b454254ab45377187848f because of which `PrepareRange` and `GetLedger` methods could return an error after Stellar-Core process exit but before all ledgers are read from the buffer. To fix it we now handle process exit in `bufferedLedgerMetaReader` only and only in case of errors. In `PrepareRange` we return error only on Stellar-Core exit with error. This won't work with user initiated shutdown, it will be fixed in #3258.

Thanks for the report, @Isaiah-Turner!

### DIFF
--- a/ingest/ledgerbackend/buffered_meta_pipe_reader.go
+++ b/ingest/ledgerbackend/buffered_meta_pipe_reader.go
@@ -100,6 +100,13 @@ func (b *bufferedLedgerMetaReader) readLedgerMetaFromPipe(untilSequence uint32) 
 		select {
 		case <-b.runner.getProcessExitChan():
 			if untilSequence != 0 {
+				// If untilSequence != 0 it's possible that Stellar-Core process
+				// exits but there are still ledgers in a buffer (catchup). In such
+				// case we ignore cases when Stellar-Core exited with no errors.
+				processErr := b.runner.getProcessExitError()
+				if processErr != nil {
+					return nil, errors.Wrap(processErr, "stellar-core process exited with an error")
+				}
 				time.Sleep(time.Second)
 				continue
 			}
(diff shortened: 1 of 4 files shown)


ingest/ledgerbackend/buffered_meta_pipe_reader.go
```diff
@@ -99,6 +99,10 @@ func (b *bufferedLedgerMetaReader) readLedgerMetaFromPipe() (*xdr.LedgerCloseMet
 		// Wait for LedgerCloseMeta buffer to be cleared to minimize memory usage.
 		select {
 		case <-b.runner.getProcessExitChan():
+			if untilSequence != 0 {
```
tamirms, line 102: we also handle `<-b.runner.getProcessExitChan` a few lines below. Should we also have a `untilSequence != 0 ` check in that block of code?

tamirms: looks like `getProcessExitChan()` is handled a few lines above as well

bartekn: We only handle the other instances of `getProcessExitChan` if an error occurs (so it ignores cases when core is done but there are still ledgers in a buffer). However, I think there's a bug. In `for frameLength > metaPipeBufferSize && len(b.c) > 0` we should return an error if core was closed with an error. Going to add it in a new commit.

bartekn: OK fixed in aecfd759c1.

tamirms: do we still need to worry about the case where `len(b.c) > 0` but `frameLength <= metaPipeBufferSize` ?

bartekn: In such case (and on Core exit) we won't enter into a `for` loop but the `xdr.Unmarshal` below will error and we will handle the error there.

tamirms (approved):

PR #3262, opened 2020-12-01
services/horizon: Refactor Captive Core configuration flags

### LINKED ISSUE #3211: Research generating stellar-core.cfg in Captive-Core online mode
It may be possible to generate Stellar-Core config as we [do in the offline mode](https://github.com/stellar/go/blob/507901ae6f7069a17f23f9c18d557a918d5e5732/ingest/ledgerbackend/stellar_core_runner.go#L90) since we Captive Core mode runs in memory. We need to ensure if there aren't any other config values users may want to change. At the same time we should require providing quorum slices for the config (may use the same toml format, quorum info will be appended to generated config).

2opremio:
* Rename `--stellar-core-config-path` to `--stellar-captive-core-quorum-path`.
  The core config path is automatically generated in captive-core online mode.
  `--stellar-captive-core-quorum-path` should point to a toml config file containing
  `[QUORUM_SET]` entries, with the same semantics as in the stellar core configuration.

* Add `--stellar-captive-core-http-port` to indicate what port (if any) captive core
  shoulf listen on. If `--stellar-core-url` is unset and the local Captive core is enabled,
  `--stellar-core-url` will be implicitly use `http://localhost:<stellar_captive_core_http_port>`

* Remove unused entries in `CaptiveStellarCore`

Closes #3211

### DIFF
--- a/exp/services/captivecore/main.go
+++ b/exp/services/captivecore/main.go
@@ -20,7 +20,7 @@ import (
 
 func main() {
 	var port int
-	var networkPassphrase, binaryPath, quorumConfigPath, dbURL string
+	var networkPassphrase, binaryPath, configAddendumPath, dbURL string
 	var historyArchiveURLs []string
 	var stellarCoreHTTPPort uint16
 	var logLevel logrus.Level
@@ -52,12 +52,12 @@ func main() {
 			ConfigKey:   &binaryPath,
 		},
 		&config.ConfigOption{
-			Name:        "stellar-captive-core-quorum-path",
+			Name:        "captive-core-addendum-path",
 			OptType:     types.String,
 			FlagDefault: "",
 			Required:    false,
-			Usage:       "path to stellar core quorum config file (i.e. a stellar core config file with [QUORUM_SET] entries)",
-			ConfigKey:   &quorumConfigPath,
+			Usage:       "path to an addendum for the Stellar Core configuration file used by captive core. It must, at least, include enough details to define a quorum set",
+			ConfigKey:   &configAddendumPath,
 		},
 		&config.ConfigOption{
 			Name:        "history-archive-urls",
@@ -113,11 +113,11 @@ func main() {
 			logger.SetLevel(logLevel)
 
 			captiveConfig := ledgerbackend.CaptiveCoreConfig{
-				StellarCoreBinaryPath: binaryPath,
-				QuorumConfigPath:      quorumConfigPath,
-				NetworkPassphrase:     networkPassphrase,
-				HistoryArchiveURLs:    historyArchiveURLs,
-				HTTPPort:              stellarCoreHTTPPort,
+				StellarCoreBinaryPath:  binaryPath,
+				CoreConfigAddendumPath: configAddendumPath,
+				NetworkPassphrase:      networkPassphrase,
+				HistoryArchiveURLs:     historyArchiveURLs,
+				HTTPPort:               stellarCoreHTTPPort,
 			}
 
 			var dbConn *db.Session
--- a/exp/services/captivecore/main.go
+++ b/exp/services/captivecore/main.go
@@ -113,11 +113,11 @@ func main() {
 			logger.SetLevel(logLevel)
 
 			captiveConfig := ledgerbackend.CaptiveCoreConfig{
-				StellarCoreBinaryPath:  binaryPath,
-				CoreConfigAddendumPath: configAddendumPath,
-				NetworkPassphrase:      networkPassphrase,
-				HistoryArchiveURLs:     historyArchiveURLs,
-				HTTPPort:               stellarCoreHTTPPort,
+				BinaryPath:         binaryPath,
+				AddendumPath:       configAddendumPath,
+				NetworkPassphrase:  networkPassphrase,
+				HistoryArchiveURLs: historyArchiveURLs,
+				HTTPPort:           stellarCoreHTTPPort,
 			}
 
 			var dbConn *db.Session
--- a/exp/services/captivecore/main.go
+++ b/exp/services/captivecore/main.go
@@ -22,7 +22,7 @@ func main() {
 	var port int
 	var networkPassphrase, binaryPath, configAddendumPath, dbURL string
 	var historyArchiveURLs []string
-	var stellarCoreHTTPPort uint16
+	var stellarCoreHTTPPort uint
 	var logLevel logrus.Level
 	logger := supportlog.New()
 
@@ -98,8 +98,8 @@ func main() {
 		&config.ConfigOption{
 			Name:        "stellar-captive-core-http-port",
 			ConfigKey:   &stellarCoreHTTPPort,
-			OptType:     types.Uint16,
-			FlagDefault: uint16(11626),
+			OptType:     types.Uint,
+			FlagDefault: uint(11626),
 			Required:    false,
 			Usage:       "HTTP port for captive core to listen on (0 disables the HTTP server)",
 		},
--- a/exp/services/captivecore/main.go
+++ b/exp/services/captivecore/main.go
@@ -20,7 +20,7 @@ import (
 
 func main() {
 	var port int
-	var networkPassphrase, binaryPath, configAddendumPath, dbURL string
+	var networkPassphrase, binaryPath, configAppendPath, dbURL string
 	var historyArchiveURLs []string
 	var stellarCoreHTTPPort uint
 	var logLevel logrus.Level
@@ -52,12 +52,12 @@ func main() {
 			ConfigKey:   &binaryPath,
 		},
 		&config.ConfigOption{
-			Name:        "captive-core-addendum-path",
+			Name:        "captive-core-config-append-path",
 			OptType:     types.String,
 			FlagDefault: "",
 			Required:    false,
-			Usage:       "path to an addendum for the Stellar Core configuration file used by captive core. It must, at least, include enough details to define a quorum set",
-			ConfigKey:   &configAddendumPath,
+			Usage:       "path to additional configuration for the Stellar Core configuration file used by captive core. It must, at least, include enough details to define a quorum set",
+			ConfigKey:   &configAppendPath,
 		},
 		&config.ConfigOption{
 			Name:        "history-archive-urls",
@@ -114,7 +114,7 @@ func main() {
 
 			captiveConfig := ledgerbackend.CaptiveCoreConfig{
 				BinaryPath:         binaryPath,
-				AddendumPath:       configAddendumPath,
+				ConfigAppendPath:   configAppendPath,
 				NetworkPassphrase:  networkPassphrase,
 				HistoryArchiveURLs: historyArchiveURLs,
 				HTTPPort:           stellarCoreHTTPPort,
--- a/exp/services/captivecore/main.go
+++ b/exp/services/captivecore/main.go
@@ -101,7 +101,7 @@ func main() {
 			OptType:     types.Uint16,
 			FlagDefault: uint16(11626),
 			Required:    false,
-			Usage:       "HTTP port for captive core to listen on (0 disables the http server)",
+			Usage:       "HTTP port for captive core to listen on (0 disables the HTTP server)",
 		},
 	}
 	cmd := &cobra.Command{
--- a/exp/services/captivecore/main.go
+++ b/exp/services/captivecore/main.go
@@ -20,8 +20,9 @@ import (
 
 func main() {
 	var port int
-	var networkPassphrase, binaryPath, configPath, dbURL string
+	var networkPassphrase, binaryPath, quorumConfigPath, dbURL string
 	var historyArchiveURLs []string
+	var stellarCoreHTTPPort uint16
 	var logLevel logrus.Level
 	logger := supportlog.New()
 
@@ -51,12 +52,12 @@ func main() {
 			ConfigKey:   &binaryPath,
 		},
 		&config.ConfigOption{
-			Name:        "stellar-core-config-path",
+			Name:        "stellar-captive-core-quorum-path",
 			OptType:     types.String,
 			FlagDefault: "",
 			Required:    false,
-			Usage:       "path to stellar core config file",
-			ConfigKey:   &configPath,
+			Usage:       "path to stellar core quorum config file (i.e. a stellar core config file with [QUORUM_SET] entries)",
+			ConfigKey:   &quorumConfigPath,
 		},
 		&config.ConfigOption{
 			Name:        "history-archive-urls",
@@ -94,6 +95,14 @@ func main() {
 			Required:  false,
 			Usage:     "horizon postgres database to connect with",
 		},
+		&config.ConfigOption{
+			Name:        "stellar-captive-core-http-port",
+			ConfigKey:   &stellarCoreHTTPPort,
+			OptType:     types.Uint16,
+			FlagDefault: uint16(11626),
+			Required:    false,
+			Usage:       "HTTP port for captive core to listen on (0 disables the http server)",
+		},
 	}
 	cmd := &cobra.Command{
 		Use:   "captivecore",
@@ -105,9 +114,10 @@ func main() {
 
 			captiveConfig := ledgerbackend.CaptiveCoreConfig{
 				StellarCoreBinaryPath: binaryPath,
-				StellarCoreConfigPath: configPath,
+				QuorumConfigPath:      quorumConfigPath,
 				NetworkPassphrase:     networkPassphrase,
 				HistoryArchiveURLs:    historyArchiveURLs,
+				HTTPPort:              stellarCoreHTTPPort,
 			}
 
 			var dbConn *db.Session
--- a/ingest/ledgerbackend/captive_core_backend.go
+++ b/ingest/ledgerbackend/captive_core_backend.go
@@ -76,7 +76,7 @@ type CaptiveStellarCore struct {
 	stellarCoreRunner stellarCoreRunnerInterface
 
 	// For testing
-	stellarCoreRunnerFactory func(captiveCoreAddendumPath string) (stellarCoreRunnerInterface, error)
+	stellarCoreRunnerFactory func(mode stellarCoreRunnerMode, captiveCoreAddendumPath string) (stellarCoreRunnerInterface, error)
 
 	// Defines if the blocking mode (off by default) is on or off. In blocking mode,
 	// calling GetLedger blocks until the requested ledger is available. This is useful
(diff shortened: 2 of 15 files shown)


ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -96,24 +95,34 @@ func (r *stellarCoreRunner) generateConfig() string {
...
+		// Add a fictional quorum -- necessary to convince core to start up;
+		// but not used at all for our purposes. Pubkey here is just random.
+		lines = append(lines,
+			"[QUORUM_SET]",
+			"THRESHOLD_PERCENT=100",
+			`VALIDATORS=["GCZBOIAY4HLKAJVNJORXZOZRAY2BJDBZHKPBHZCRAIUR5IHC2UHBGCQR"]`)
+	}
+	result := strings.ReplaceAll(strings.Join(lines, "\n"), "\\", "\\\\")
+	if r.quorumConfigPath != "" {
+		// TODO: parse toml file and ensure it is a correct quorum config file
```
bartekn, line 115: I think it'd fine if we just check if the first line is `[QUORUM_SET]` so users don't add any extra config values before it. If toml is invalid, Core will just crash and user will see the output so we probably don't need o parse the file.

bartekn (commented):
LGTM. I remember you wanted to add some improvements so will approve later.

exp/services/captivecore/main.go
```diff
@@ -51,12 +52,12 @@ func main() {
...
 		},
 		&config.ConfigOption{
-			Name:        "stellar-core-config-path",
+			Name:        "stellar-captive-core-quorum-path",
 			OptType:     types.String,
 			FlagDefault: "",
 			Required:    false,
-			Usage:       "path to stellar core config file",
-			ConfigKey:   &configPath,
+			Usage:       "path to stellar core quorum config file (i.e. a stellar core config file with [QUORUM_SET] entries)",
```
tamirms, line 59: it's worth mentioning that the config file contains only `QUORUM_SET` entries and nothing else

2opremio:
**UPDATE**

It turns out that `[QUORUM_SET]` entries were not enough to specify how to connect to a remote network (they don't include host names). Additionally `[QUORUM_SET]` is [deprecated by Stellar Core](https://github.com/stellar/stellar-core/blob/master/docs/stellar-core_example.cfg#L555-L558)

In reality we need a combination of `[[HOME_DOMAINS]]` and `[[VALIDATORS]]`.

Thus, I have renamed `--stellar-captive-core-quorum-path` to `captive-core-addendum-path`, and I have documented it at https://github.com/stellar/go/pull/3262/files#diff-dcbfcf6a4881548aa52f3762303b0f1cf6ad6441a011cfee5273961012619e5e

PTAL

exp/services/captivecore/main.go
2opremio: It doesn't apply anymore, read below

ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -62,12 +62,9 @@ func roundDownToFirstReplayAfterCheckpointStart(ledger uint32) uint32 {
...
 type CaptiveStellarCore struct {
-	executablePath    string
-	configPath        string
-	networkPassphrase string
-	historyURLs       []string
-	archive           historyarchive.ArchiveInterface
-	ledgerHashStore   TrustedLedgerHashStore
+	coreConfigAddendumPath string
+	archive                historyarchive.ArchiveInterface
+	ledgerHashStore        TrustedLedgerHashStore
```
bartekn, line 67: Thanks for the cleanup here!

ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -79,7 +76,7 @@ type CaptiveStellarCore struct {
 	stellarCoreRunner stellarCoreRunnerInterface
 
 	// For testing
-	stellarCoreRunnerFactory func(configPath string) (stellarCoreRunnerInterface, error)
+	stellarCoreRunnerFactory func(quorumConfigPath string) (stellarCoreRunnerInterface, error)
```
bartekn, line 79: Still mentions `quorum`.

ingest/ledgerbackend/stellar_core_runner_test.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"io/ioutil"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGenerateConfig(t *testing.T) {
+	tmpFile, err := ioutil.TempFile("", "test-generate-config")
```
bartekn, line 12: Nit: I'm afraid it can be a problem in the future. Unit tests should generate file, make connections etc. An alternative can be to create and implement `Ioutil` interface with the methods we need and then mock it. Up to you if you want to do it.

services/horizon/internal/docs/captive_core.md
```diff
@@ -13,7 +13,32 @@ Captive Stellar-Core can be used in both reingestion and normal Horizon operatio
 To enable captive mode three feature config variables are required:
 * `ENABLE_CAPTIVE_CORE_INGESTION=true`,
 * `STELLAR_CORE_BINARY_PATH` - defines a path to the `stellar-core` binary,
-* `STELLAR_CORE_CONFIG_PATH` - defines a path to the `stellar-core.cfg` file (not required when reingesting).
+* `CAPTIVE_CORE_ADDENDUM_PATH` - (not required when reingesting) defines a path to an addendum for the Stellar Core configuration file used by captive core. It must, at least, include enough details to define a quorum set. For instance, to connect to teh Stellar testnet through `core-testnet1.stellar.org`:
```
bartekn, line 16: I have mixed feelings about `addendum` name. Maybe `append` or `footer` would be better? @ire-and-curses opinions from native-English?

services/horizon/internal/ingest/main.go
```diff
@@ -177,11 +178,12 @@ func NewSystem(config Config) (System, error) {
...
 			captiveCoreBackend, err = ledgerbackend.NewCaptive(
 				ledgerbackend.CaptiveCoreConfig{
-					StellarCoreBinaryPath: config.StellarCoreBinaryPath,
-					StellarCoreConfigPath: config.StellarCoreConfigPath,
-					NetworkPassphrase:     config.NetworkPassphrase,
-					HistoryArchiveURLs:    []string{config.HistoryArchiveURL},
-					LedgerHashStore:       ledgerbackend.NewHorizonDBLedgerHashStore(config.HistorySession),
+					StellarCoreBinaryPath:  config.StellarCoreBinaryPath,
+					CoreConfigAddendumPath: config.CaptiveCoreQuorumConfigPath,
+					HTTPPort:               config.CaptiveCoreHTTPPort,
```
bartekn, line 183: Can we make these three consistent. First one is `StellarCore`, second `Core` and the last one doesn't mention Stellar-Core but it's Stellar-Core's port.

services/horizon/internal/config.go
```diff
@@ -16,12 +16,13 @@ type Config struct {
...
-	StellarCoreDatabaseURL     string
-	StellarCoreURL             string
-	RemoteCaptiveCoreURL       string
+	EnableCaptiveCoreIngestion  bool
+	StellarCoreBinaryPath       string
+	CaptiveCoreQuorumConfigPath string
+	StellarCoreDatabaseURL      string
+	StellarCoreURL              string
+	RemoteCaptiveCoreURL        string
+	CaptiveCoreHTTPPort         uint
```
bartekn, line 25: As in previous comment, can we stick to `StellarCore` or `CaptiveCore`?

services/horizon/internal/docs/captive_core.md
bartekn, line 16:
```suggestion
* `CAPTIVE_CORE_ADDENDUM_PATH` - (not required when reingesting) defines a path to an addendum for the Stellar Core configuration file used by captive core. It must, at least, include enough details to define a quorum set. For instance, to connect to the Stellar testnet through `core-testnet1.stellar.org`:
```

ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -96,24 +96,33 @@ func (r *stellarCoreRunner) generateConfig() string {
...
 		lines = append(lines, fmt.Sprintf(`get="curl -sf %s/{0} -o {1}"`, val))
 	}
-	// Add a fictional quorum -- necessary to convince core to start up;
-	// but not used at all for our purposes. Pubkey here is just random.
-	lines = append(lines,
-		"[QUORUM_SET]",
-		"THRESHOLD_PERCENT=100",
-		`VALIDATORS=["GCZBOIAY4HLKAJVNJORXZOZRAY2BJDBZHKPBHZCRAIUR5IHC2UHBGCQR"]`)
-	return strings.ReplaceAll(strings.Join(lines, "\n"), "\\", "\\\\")
+	if r.coreConfigAddendumPath == "" {
```
bartekn, line 105: Can we return an error here if `generateConfig` was called from `runFrom` (so we expect to have quorum slice set by user)? I think that right now it will just create _catchup_ config.

2opremio: The caller ensures that the path isn't empty in online mode.

bartekn: OK, I can see it now but I think `stellarCoreRunner` should do it because it's their responsibility. It's possible that it will be used outside of `CaptiveStellarCore` in the future and this will generate invalid config files.

2opremio: Fair enough. Then we should pass online/offline as a parameter. Are you OK with that?

services/horizon/internal/docs/captive_core.md
```diff
@@ -13,7 +13,34 @@ Captive Stellar-Core can be used in both reingestion and normal Horizon operatio
 To enable captive mode three feature config variables are required:
 * `ENABLE_CAPTIVE_CORE_INGESTION=true`,
 * `STELLAR_CORE_BINARY_PATH` - defines a path to the `stellar-core` binary,
-* `STELLAR_CORE_CONFIG_PATH` - defines a path to the `stellar-core.cfg` file (not required when reingesting).
+* `CAPTIVE_CORE_ADDENDUM_PATH` - (not required when reingesting) defines a path to an addendum for the Stellar Core configuration file used by captive core. It must, at least, include enough details to define a quorum set. For instance, to connect to the Stellar testnet through `core-testnet1.stellar.org`:
```
tamirms, line 16: I think it's worth being more descriptive about reingesting. maybe replace "(not required when reingesting)" to "(not not required when running the reingestion command:  `stellar-horizon db reingest range`)"

tamirms: or you could make "reingesting" link to the `##### Reingestion` section of the markdown file

ingest/ledgerbackend/stellar_core_runner_test.go
2opremio: I think that mocking the file access only for this is overkill (at this point).

ingest/ledgerbackend/stellar_core_runner.go
bartekn: Sounds good.

services/horizon/internal/docs/captive_core.md
```diff
@@ -13,7 +13,34 @@ Captive Stellar-Core can be used in both reingestion and normal Horizon operatio
...
+  ADDRESS="core-testnet1.stellar.org"
+  ```
+  
+  The base config file the addendum will be appended to is:
+  ```
+  # Generated file -- do not edit"
+  RUN_STANDALONE=true
+  NODE_IS_VALIDATOR=false
+  DISABLE_XDR_FSYNC=true
+  UNSAFE_QUORUM=true
```
tamirms, line 35: are we sure want to default to `UNSAFE_QUORUM=true` ? 

https://github.com/stellar/stellar-core/blob/master/docs/stellar-core_example.cfg#L237-L248

bartekn: I don't think we should include base file here. It can be different for catchup/run from. Ex. in catchup `UNSAFE_QUORUM` is fine because Stellar-Core doesn't connect to the network.

2opremio: I am just reusing the config of offline mode. My guess is that this is enough for a watcher node, but I am not sure.

2opremio: Ah, I see.

I will remove the `UNSAFE_QUORUM` when in online mode and remove the base config from the docs. Should I also remove `DISABLE_XDR_FSYNC=true`

But then .. we should probably provide a way for the user to print the config file.

I would output it in the logs, but maybe it's overkill. Maybe we can print it in debug mode?

services/horizon/internal/docs/captive_core.md
ire-and-curses: how about CAPTIVE_CORE_CONFIG_PATH? From the user's perspective, this is the only way captive core can be configured.

2opremio: I think that would be hiding the fact that we are adding lines to a config file we cook ourselves.

But we could also clarify that in the flag description ... so, I don't know @bartekn thoughts?

services/horizon/internal/docs/captive_core.md
2opremio: I think I've address this, PTAL.

services/horizon/internal/docs/captive_core.md
ire-and-curses: CAPTIVE_CORE_APPEND_PATH is also clear, so we can just go with that.

ire-and-curses: I agree that addendum is pretty obscure for non-native speakers, and usually it has a connection to something legal.

2opremio: Other suggestions are:
* appendix
* suffix

But I can go with append if you agree.

tamirms: how about `CAPTIVE_CORE_CONFIG_OVERRIDE` ?

* `CAPTIVE_CORE_CONFIG_OVERRIDE` - (not required when running `horizon db reingest range`) defines a path to a Stellar Core configuration file which will be merged with Horizon's default configuration for running Stellar Core in captive core mode.  The configuration file must, at least, include enough details to define a quorum set. For instance, to connect to the Stellar testnet through `core-testnet1.stellar.org`:

  ```
  [[HOME_DOMAINS]]
  HOME_DOMAIN="testnet.stellar.org"
  QUALITY="MEDIUM"
  [[VALIDATORS]]
  NAME="sdf_testnet_1"
  HOME_DOMAIN="testnet.stellar.org"
  PUBLIC_KEY="GDKXE2OZMJIPOSLNA6N6F2BVCI3O777I2OOC4BV7VOYUEHYX7RTRYA7Y"
  ADDRESS="core-testnet1.stellar.org"
  ```

  The full configuration to be used will be printed out by Horizon when running horizon with `--log-level debug`
```

File: ingest/ledgerbackend/stellar_core_runner.go [hunk 12]
```diff
@@ -66,54 +76,68 @@ func newStellarCoreRunner(executablePath, configPath, networkPassphrase string,
...
 }
 
-func (r *stellarCoreRunner) generateConfig() string {
+func (r *stellarCoreRunner) generateConfig() (string, error) {
+	if r.mode == stellarCoreRunnerModeOnline && r.coreConfigAddendumPath == "" {
+		return "", errors.New("stellar-core addendum config file path cannot be empty in online mode")
+	}
 	lines := []string{
 		"# Generated file -- do not edit",
 		"RUN_STANDALONE=true",
```
tamirms, line 105: sorry I missed this during my earlier review but, I think `"RUN_STANDALONE=true"` should only be enabled in offline mode

services/horizon/internal/docs/captive_core.md
bartekn: Isn't _override_ is a synonym of _replace_? I think that _append_ is great. So maybe `CAPTIVE_CORE_CONFIG_APPEND_PATH`?

tamirms: Currently we are only augmenting the default config with validations. But we might want to allow properties that we configure as default to be overridden in the future if the use case makes sense.

Anyways, I’m not opposed to using append in the name. Just wanted to offer up an alternative

tamirms: For example we might need to enable unsafe quorum for the captive core integration tests

2opremio: I have renamed it to `CAPTIVE_CORE_CONFIG_APPEND_PATH`  but .... maybe it's better to use `CAPTIVE_ADDITIONAL_CONFIG_PATH` ? 

It's longer though.

Happy to change the renaming again, but let's settle it please.

ingest/ledgerbackend/stellar_core_runner.go
2opremio: I changed it it ... but this enables peer connections, I think. Are we sure about this?

tamirms: according to the stellar-core [doc](https://github.com/stellar/stellar-core/blob/master/docs/stellar-core_example.cfg#L301-L304) when you set it to true it prevents core from connecting to other peers

tamirms: I think in online mode we need to be able to connect to peers to sync with consensus

tamirms (approved):

services/horizon/internal/docs/captive_core.md
```diff
@@ -13,7 +13,22 @@ Captive Stellar-Core can be used in both reingestion and normal Horizon operatio
...
+  QUALITY="MEDIUM"
+
+  [[VALIDATORS]]
+  NAME="sdf_testnet_1"
+  HOME_DOMAIN="testnet.stellar.org"
+  PUBLIC_KEY="GDKXE2OZMJIPOSLNA6N6F2BVCI3O777I2OOC4BV7VOYUEHYX7RTRYA7Y"
+  ADDRESS="core-testnet1.stellar.org"
+  ```
+  
+  The full configuration to be used will be printed out by Horizon when runnign horizon with `--log-level debug`
```
ire-and-curses, line 29:
```suggestion
  The full configuration to be used will be printed out by Horizon when running horizon with `--log-level debug`,
```

ingest/ledgerbackend/stellar_core_runner_test.go
```diff
@@ -0,0 +1,33 @@
...
+
+	tmpFile.WriteString(`[[HOME_DOMAINS]]
+HOME_DOMAIN="testnet.stellar.org"
+QUALITY="HIGH"
+
+[[VALIDATORS]]
+NAME="sdf_testnet_1"
+HOME_DOMAIN="testnet.stellar.org"
+PUBLIC_KEY="GDKXE2OZMJIPOSLNA6N6F2BVCI3O777I2OOC4BV7VOYUEHYX7RTRYA7Y"
+ADDRESS="core-testnet1.stellar.org"
```
Shaptic, line 29: Worth noting that this is not a valid configuration stub. From the Horizon Captive Core logs:

```
Got an exception: Failed to parse '/tmp/captive-stellar-core-3bdb97927775752a/stellar-core.conf' :Critical and High quality validators sdf_testnet_1 must have redundancy of at least 3
```

2opremio: Yes, it has to be "MEDIUM", I fixed it in the [docs](https://github.com/stellar/go/blob/master/services/horizon/internal/docs/captive_core.md#configuration), but not here.

PR #3264, opened 2020-12-03
services/horizon/internal/test/integration: Use ARTIFICIALLY_ACCELERATE_TIME_FOR_TESTING in integration tests

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Use `ARTIFICIALLY_ACCELERATE_TIME_FOR_TESTING=true` in integration tests instead of manual close.

### Why

Running the horizon integration tests with captive core enabled has turned out to be very difficult due to issues with running Stellar Core in manual close mode (see https://github.com/stellar/stellar-core/issues/2778 for more details).

We use manual close mode to speed up our integration tests. Otherwise, we'd need to wait 5 seconds for every ledger close and that would result in our integration tests taking over 20 minutes to complete.

The Stellar Core team has suggested we use `ARTIFICIALLY_ACCELERATE_TIME_FOR_TESTING=true` instead of manual close in our integration tests because it would fix many of the issues we encountered when attempting to run the integration tests with captive core enabled.

I have verified that the execution time of our integration tests with `ARTIFICIALLY_ACCELERATE_TIME_FOR_TESTING=true` are still on par with running Stellar Core in manual close mode.

### Known limitations

ARTIFICIALLY_ACCELERATE_TIME_FOR_TESTING drops the time between ledger closes from ~5s to ~1s, and the number of ledgers between checkpoints from 64 to 8. Horizon assumes that each checkpoint spans 64 ledgers. In a follow up commit we should make the size of a checkpoint configurable because there is code exercised by captive core ingestion which relies on the assumption that a checkpoint spans 64 ledgers.

### DIFF
--- a/services/horizon/docker/stellar-core-integration-tests.cfg
+++ b/services/horizon/docker/stellar-core-integration-tests.cfg
@@ -2,7 +2,7 @@
 # see stellar-core_example.cfg for a description of the configuration parameters
 
 RUN_STANDALONE=false
-MANUAL_CLOSE=true
+ARTIFICIALLY_ACCELERATE_TIME_FOR_TESTING=true
 
 NETWORK_PASSPHRASE="Standalone Network ; February 2017"
 
--- a/services/horizon/internal/integration/protocol14_sponsorship_ops_test.go
+++ b/services/horizon/internal/integration/protocol14_sponsorship_ops_test.go
@@ -363,7 +363,7 @@ func TestSponsorships(t *testing.T) {
 		// Submit the preauthorized transaction
 		var txResult xdr.TransactionResult
 		tt.NoError(err)
-		txResp, err = itest.SubmitTransactionXDR(preAuthTxB64)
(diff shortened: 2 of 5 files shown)


services/horizon/internal/integration/protocol14_claimable_balance_ops_test.go
```diff
@@ -26,7 +26,7 @@ func TestClaimableBalanceCreationOperationsAndEffects(t *testing.T) {
 			Amount: "10",
 			Asset:  txnbuild.NativeAsset{},
 		}
-
+
```
2opremio, line 29: whitespace before EOL

tamirms: @2opremio thanks, this should be ready for review now

bartekn (approved):

PR #3265, opened 2020-12-04
ingest/ledgerbackend: Fix captive core bug where --start-at-hash parameter is omitted

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

In https://github.com/stellar/go/pull/3201 we assumed it was possible to omit the `--start-at-hash` parameter when starting from the genesis ledger. However, Stellar Core actually fails when attempting that command. The `--start-at-hash` parameter is always required when using `--start-at-ledger`.

This means that if we want to stream from the genesis ledger we need to determine the either the horizon db or the history archives. 

~~In the case that the first checkpoint has not been published we attempt to invoke `stellar-core run --in-memory` without any start parameters. The expected behavior from stellar core is that it will join the network and stream from the nearest checkpoint boundary, which in this case would be the genesis ledger since the first checkpoint has not been published yet.~~

[edit:] I tested the behavior of `stellar-core run --in-memory` on a stand alone network and unfortunately stellar core will keep buffering ledgers until the first checkpoint is published before it starts streaming any ledgers. Therefore it does not make any sense to use this work around.

### Known limitations

N / A

### DIFF
--- a/ingest/ledgerbackend/captive_core_backend.go
+++ b/ingest/ledgerbackend/captive_core_backend.go
@@ -310,30 +310,28 @@ func (c *CaptiveStellarCore) runFromParams(from uint32) (runFrom uint32, ledgerH
 	}
 
 	if from <= 63 {
-		// For ledgers before (and including) first checkpoint, we start streaming
-		// without providing a hash, to avoid waiting for the checkpoint.
-		// It will always start streaming from ledger 2.
+		// For ledgers before (and including) first checkpoint, get/wait the first
+		// checkpoint to get the ledger header. It will always start streaming
+		// from ledger 2.
 		nextLedger = 2
-		runFrom = 2
 		// The line below is to support a special case for streaming ledger 2
 		// that works for all other ledgers <= 63 (fast-forward).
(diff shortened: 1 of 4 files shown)


2opremio (approved):

ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -254,12 +254,26 @@ func (c *CaptiveStellarCore) openOnlineReplaySubprocess(from uint32) error {
...
 
-	runFrom, ledgerHash, nextLedger, err := c.runFromParams(from)
-	if err != nil {
-		return errors.Wrap(err, "error calculating ledger and hash for stelar-core run")
-	}
+	var nextLedger uint32
+	if latestCheckpointSequence == 0 {
+		if c.log != nil {
+			c.log.Info("checkpoint ledger has not been published, attempting to run from genesis ledger")
+		}
```
bartekn, line 261: Let's remove this. `s.log` is meant to be used for stellar-core log only (set via `SetStellarCoreLogger`).
```suggestion
```

bartekn (approved):

Issue #3248, opened 2020-11-24
Add testnet/pubnet sync, catchup (BoundedRange) and reingest integration tests


bartekn:
Add integration tests that sync with the network: build state and ingest a few ledgers. This will help ensure that Horizon is able to sync and catch bugs like https://github.com/stellar/go/pull/3231 before release testing.

bartekn:
To clarify, I think we should test 3 things:
1. starting from scratch to check if horizon can build state without a problems + a few ledgers,
2. starting from existing DB (`start` -> `resume` transition in FSM scenario). this would catch https://github.com/stellar/go/pull/3231,
3. `horizon db reingest` so we are sure that reingesting works (I believe it doesn’t in the latest version because of https://github.com/stellar/go/pull/3260).

bartekn:
@2opremio @tamirms you asked why we should do testnet/pubnet tests and I answered that I was probably wrong and standalone should work too. That's true but what is the status of captive core tests in standalone? I'm looking at #3267 and realized that it wouldn't catch #3231 or #3260 because both issues were connected to captive core.

2opremio:
Note that as soon as the integration tests can be run with captive core (which is almost there, I think) the tests at #3267 will be run with captive core too.

bartekn:
OK! I just wanted to confirm we'll be able to have them before beta. If not, it would be great to have testnet/pubnet tests.

PR #3266, opened 2020-12-04
services/horizon/ingest: Close LedgerBackend in Shutdown

bartekn:
<details>
  <summary>PR Checklist</summary>
  
### PR Structure

* [ ] This PR has reasonably narrow scope (if not, break it down into smaller PRs).
* [ ] This PR avoids mixing refactoring changes with feature changes (split into two PRs
  otherwise).
* [ ] This PR's title starts with name of package that is most changed in the PR, ex.
  `services/friendbot`, or `all` or `doc` if the changes are broad or impact many
  packages.

### Thoroughness

* [ ] This PR adds tests for the most critical parts of the new functionality or fixes.
* [ ] I've updated any docs ([developer docs](https://www.stellar.org/developers/reference/), `.md`
  files, etc... affected by this change). Take a look in the `docs` folder for a given service,
  like [this one](https://github.com/stellar/go/tree/master/services/horizon/internal/docs).

### Release planning

* [ ] I've updated the relevant CHANGELOG ([here](services/horizon/CHANGELOG.md) for Horizon) if
  needed with deprecations, added features, breaking changes, and DB schema changes.
* [ ] I've decided if this PR requires a new major/minor version according to
  [semver](https://semver.org/), or if it's mainly a patch change. The PR is targeted at the next
  release branch if it's not a patch change.
</details>

### What

TODO check if it works correctly with https://github.com/stellar/go/pull/3258.

### Why

[TODO: Why this change is being made. Include any context required to understand the why.]

### Known limitations

[TODO or N/A]

PR #3267, opened 2020-12-07
services/horizon: Add re-ingestion and resume integration tests

2opremio:
They required:

1. Adding new integration test infrastructure to:
   * Shutdown the integration test early (needed for resuming a Horizon session)
   * Obtain the Horizon configuration (in order to reuse it with the re-ingest command)

2. Modifying the `db reingest range` command (in order to make it testeable). The flag-parsing
   part and intialization logic have been split out and the latter is used by the integration test.

Fixes https://github.com/stellar/go/issues/3248

### DIFF
--- a/services/horizon/internal/integration/db_test.go
+++ b/services/horizon/internal/integration/db_test.go
@@ -1,0 +1,83 @@
+package integration
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+
+	horizoncmd "github.com/stellar/go/services/horizon/cmd"
+	"github.com/stellar/go/services/horizon/internal/db2/schema"
+	"github.com/stellar/go/services/horizon/internal/test/integration"
+	"github.com/stellar/go/support/db"
+	"github.com/stellar/go/support/db/dbtest"
+	"github.com/stellar/go/txnbuild"
+)
+
+func initializeDBIntegrationTest(t *testing.T) (itest *integration.Test, reachedLedger int32) {
+	itest = integration.NewTest(t, protocol15Config)
+	master := itest.Master()
+	tt := assert.New(t)
+
+	// Initialize the database with some ledgers including some transactions we submit
+	op := txnbuild.Payment{
+		Destination: master.Address(),
+		Amount:      "10",
+		Asset:       txnbuild.NativeAsset{},
+	}
+	// TODO: should we enforce certain number of ledgers to be ingested?
+	for i := 0; i < 8; i++ {
+		txResp := itest.MustSubmitOperations(itest.MasterAccount(), master, &op)
+		reachedLedger = txResp.Ledger
+	}
+
+	root, err := itest.Client().Root()
+	tt.NoError(err)
+	tt.LessOrEqual(reachedLedger, root.HorizonSequence)
+
+	return
+}
+
+func TestReingestDB(t *testing.T) {
+	itest, reachedLedger := initializeDBIntegrationTest(t)
+	tt := assert.New(t)
+
+	// Create a fresh Horizon database
+	newDB := dbtest.Postgres(t)
+	// TODO: Unfortunately Horizon's ingestion System leaves open sessions behind,leading to
+	//       a "database  is being accessed by other users" error when trying to drop it
+	// defer newDB.Close()
+	freshHorizonPostgresURL := newDB.DSN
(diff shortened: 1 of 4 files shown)


services/horizon/internal/integration/db_test.go
```diff
@@ -0,0 +1,83 @@
...
+
+func TestReingestDB(t *testing.T) {
+	itest, reachedLedger := initializeDBIntegrationTest(t)
+	tt := assert.New(t)
+
+	// Create a fresh Horizon database
+	newDB := dbtest.Postgres(t)
+	// TODO: Unfortunately Horizon's ingestion System leaves open sessions behind,leading to
+	//       a "database  is being accessed by other users" error when trying to drop it
+	// defer newDB.Close()
```
bartekn, line 49: It may be because we don't close `ledgerbackend` (so in case of `DatabaseBackend` we don't [close DB sessions](https://github.com/stellar/go/blob/5ed04d24078ba1e72fad169b77b7b3b5c1e397cc/ingest/ledgerbackend/database_backend.go#L222)), I fixed it in https://github.com/stellar/go/pull/3266.

services/horizon/internal/test/integration/integration.go
```diff
@@ -78,41 +83,66 @@ func NewTest(t *testing.T, config Config) *Test {
...
 
 	i.cclient = &stellarcore.Client{URL: "http://localhost:" + strconv.Itoa(stellarCorePort)}
 	i.waitForCore()
 
+	i.shutdownCalls = append(i.shutdownCalls, func() {
+		if i.app != nil {
+			i.app.Close()
+		}
+		runComposeCommand("down", "-v", "--remove-orphans")
+	})
```
bartekn, line 98: Do we need this? You can actually register multiple functions using [`T.Cleanup`](https://golang.org/pkg/testing/#T.Cleanup):
> Cleanup registers a function to be called when the test and all its subtests complete. Cleanup functions will be called in last added, first called order.

services/horizon/internal/integration/db_test.go
```diff
@@ -0,0 +1,83 @@
...
+
+func TestResumeFromInitializedDB(t *testing.T) {
+	itest, reachedLedger := initializeDBIntegrationTest(t)
+	tt := assert.New(t)
+
+	// Stop the integration test, and restart it with the same database
+	oldDBURL := itest.GetHorizonConfig().DatabaseURL
+	itestConfig := protocol15Config
+	itestConfig.PostgresURL = oldDBURL
+	itest.Shutdown()
```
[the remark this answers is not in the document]
2opremio: I don't see how you can reuse itest if the idea is to stop Horizon and resume it. Can you suggest how?

services/horizon/internal/test/integration/integration.go
2opremio: We need this to stop Horizon and restarted within the context of a single test, AFAIU `T.Cleanup` will only run the cleanup once the test is done.

services/horizon/internal/integration/db_test.go
2opremio: Thanks. I will check once #3266 is merged.

services/horizon/internal/integration/db_test.go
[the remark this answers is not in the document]
bartekn: OK, I understand now. I didn't realized it's happening here. Time to bed.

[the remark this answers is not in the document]
2opremio: No worries, sleep tight!

services/horizon/internal/integration/db_test.go
```diff
@@ -0,0 +1,83 @@
...
+
+	itest = integration.NewTest(t, itestConfig)
+
+	successfullyResumed := func() bool {
+		root, err := itest.Client().Root()
+		tt.NoError(err)
+		return root.HorizonSequence >= reachedLedger
```
tamirms, line 79: I think it's worth checking that horizon eventually advances to the ledger after `reachedLedger`, e.g. `root.HorizonSequence > reachedLedger`

tamirms (approved):

services/horizon/internal/integration/db_test.go
2opremio: Good point

PR #3269, opened 2020-12-08
ingest/ledgerbackend: Fix possible nil pointer when accessing captive core logger

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

There is a `log` field in the `stellarCoreRunner` struct. The field is considered optional and the intended behavior is that if `log` is nil then instead of using `log` to print a message we will print it to stdout using `fmt.Println()`.

However, when we generate the stellar core configuration file for captive core we print the file using `log.Debugf()`. This causes the code to panic because `log` can be nil.

To fix this bug I have refactored the code so that `log` is always non-nil. A custom logger can be provided to `stellarCoreRunner` via its constructor. And if the custom logger is omitted, the constructor will create a new logrus instance which is configured to print to standard out.

### Known limitations

[N/A]

### DIFF
--- a/ingest/ledgerbackend/stellar_core_runner.go
+++ b/ingest/ledgerbackend/stellar_core_runner.go
@@ -15,6 +15,7 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 
 	"github.com/stellar/go/support/log"
 )
@@ -27,7 +28,6 @@ type stellarCoreRunnerInterface interface {
 	getProcessExitChan() <-chan struct{}
 	// getProcessExitError returns an exit error of the process, can be nil
 	getProcessExitError() error
-	setLogger(*log.Entry)
 	close() error
 }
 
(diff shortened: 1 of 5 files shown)


2opremio (approved):

ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -75,18 +73,26 @@ func newStellarCoreRunner(executablePath, coreConfigAppendPath, networkPassphras
 		return nil, errors.Wrap(err, "error creating subprocess tmpdir")
 	}
 
+	coreLogger := config.Log
+	if coreLogger == nil {
+		coreLogger = log.New()
+		coreLogger.Logger.SetOutput(os.Stdout)
+		coreLogger.SetLevel(logrus.InfoLevel)
+	}
```
bartekn, line 81: I think allowing `nil` was useful for cases in which devs don't really care about Stellar-Core output. It's not recommended but I guess could be useful in simple apps/tools.

ire-and-curses: Have we ever needed that? Also seems like they have options in that case:
- they just add a custom logger that does nothing
- they could just redirect stdout/stderr to /dev/null

PR #3273, opened 2020-12-08
all: Parameterize the checkpoint frequency

2opremio:
It used to be hardcoded to 64 (which is what _normal_ Stellar Core).

However, when using ARTIFICIALLY_ACCELERATE_TIME_FOR_TESTING in integration tests, the frequency is reduced to 8.

Followup to #3264

### DIFF
--- a/exp/services/captivecore/main.go
+++ b/exp/services/captivecore/main.go
@@ -110,7 +110,7 @@ func main() {
 			OptType:     types.Uint32,
 			FlagDefault: uint32(64),
 			Required:    false,
-			Usage:       "establishes how many ledgers exist between checkpoints, do NOT change this if you really know what you are doing",
+			Usage:       "establishes how many ledgers exist between checkpoints, do NOT change this unless you really know what you are doing",
 		},
 	}
 	cmd := &cobra.Command{
--- a/exp/services/captivecore/main.go
+++ b/exp/services/captivecore/main.go
@@ -23,6 +23,7 @@ func main() {
 	var networkPassphrase, binaryPath, configAppendPath, dbURL string
 	var historyArchiveURLs []string
 	var stellarCoreHTTPPort uint
+	var checkpointFrequency uint32
 	var logLevel logrus.Level
 	logger := supportlog.New()
 
@@ -103,6 +104,14 @@ func main() {
 			Required:    false,
 			Usage:       "HTTP port for captive core to listen on (0 disables the HTTP server)",
 		},
+		&config.ConfigOption{
+			Name:        "checkpoint-frequency",
+			ConfigKey:   &checkpointFrequency,
+			OptType:     types.Uint32,
+			FlagDefault: uint32(64),
+			Required:    false,
+			Usage:       "establishes how many ledgers exist between checkpoints, do NOT change this if you really know what you are doing",
+		},
 	}
 	cmd := &cobra.Command{
 		Use:   "captivecore",
@@ -113,12 +122,13 @@ func main() {
 			logger.SetLevel(logLevel)
 
 			captiveConfig := ledgerbackend.CaptiveCoreConfig{
-				BinaryPath:         binaryPath,
-				ConfigAppendPath:   configAppendPath,
-				NetworkPassphrase:  networkPassphrase,
-				HistoryArchiveURLs: historyArchiveURLs,
-				HTTPPort:           stellarCoreHTTPPort,
-				Log:                logger.WithField("subservice", "stellar-core"),
+				BinaryPath:          binaryPath,
+				ConfigAppendPath:    configAppendPath,
+				NetworkPassphrase:   networkPassphrase,
+				HistoryArchiveURLs:  historyArchiveURLs,
+				CheckpointFrequency: checkpointFrequency,
+				HTTPPort:            stellarCoreHTTPPort,
+				Log:                 logger.WithField("subservice", "stellar-core"),
 			}
 
 			var dbConn *db.Session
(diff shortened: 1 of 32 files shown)


historyarchive/range.go
```diff
@@ -10,30 +10,36 @@ import (
...
+func (c CheckpointManager) IsCheckpoint(i uint32) bool {
+	return (i+1)%c.checkpointFreq == 0
+}
+
+func (c CheckpointManager) PrevCheckpoint(i uint32) uint32 {
+	freq := c.checkpointFreq
 	if i < freq {
 		return freq - 1
 	}
 	return (((i + 1) / freq) * freq) - 1
```
Shaptic, line 35: Unrelated to this change, but what the heck is this math? Don't the `/ freq * freq` cancel out, then the `+ 1 - 1` cancel, so this is just `i`? Happens again below on [line 44](https://github.com/stellar/go/pull/3273/files#diff-fff8b6e0a8f6c057d7cd0cb2654f01d15f1d3cd9e59bf4e596a61df1616d0eb2R44).

2opremio: It doesn't cancel out because it's integer division.

`(a / b)  * b` is a (cheaper I think) alternative to `a - (a % b)` (e.g. `(67 / 64) * 64 = 1 * 64 = 64` )

The `+1` a and `-1` are due to checkpoint ranges being zero-based (e.g. 1 to 63, 64 to 127 ...) and the  checkpoint being the last ledger in the range (e.g. 63, 127 ...).

bartekn: It's [quotient in integers](https://golang.org/ref/spec#Arithmetic_operators) so `i` won't work because `(i+1)/freq` is truncated towards zero. Here's an [example](https://play.golang.com/p/NEKmYmyFbwT) with i=64.

bartekn: LOL, sorry, @2opremio I didn't see your reply when I was writing my comment. Your answer is much better.

exp/services/captivecore/main.go
```diff
@@ -103,6 +104,14 @@ func main() {
 			Required:    false,
 			Usage:       "HTTP port for captive core to listen on (0 disables the HTTP server)",
 		},
+		&config.ConfigOption{
+			Name:        "checkpoint-frequency",
+			ConfigKey:   &checkpointFrequency,
+			OptType:     types.Uint32,
+			FlagDefault: uint32(64),
+			Required:    false,
+			Usage:       "establishes how many ledgers exist between checkpoints, do NOT change this if you really know what you are doing",
```
tamirms, line 113: do NOT change this **unless** you really know what you are doing

ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -107,6 +108,8 @@ type CaptiveCoreConfig struct {
 	NetworkPassphrase string
 	// HistoryArchiveURLs are a list of history archive urls
 	HistoryArchiveURLs []string
+	// CheckpointFrequency is the number of ledgers between checkpoints
+	CheckpointFrequency uint32
```
tamirms, line 112: if `CheckpointFrequency` is 0 can we make the constructor set it to 64 by default, or is that too much of a hack? I think it would be nice if we could somehow make 64 the default without requiring the consumer of the library to set it explicitly

exp/services/captivecore/main.go
2opremio: Good catch

ingest/ledgerbackend/captive_core_backend.go
2opremio: Uhm, I considered but I considered it a bit of a hack (I didn't want to open the door to having magic constants in the code to enforce coder use the checkpointManager). I would also need to do the same in a couple of other places.

On the other hand, the value not being 64 is very unusual, so, I am on the fence about this. @Shaptic @bartekn thoughts?

bartekn: I'd like to propose an alternative. Because checkpoint frequency is actually an archive property (normal archives will have checkpoints every 64 ledgers, others 8 ledger) can we make it a field on `historyarchive.Archive` (and I guess on `ConnectOptions`) that defaults to 64? The problem I can see now is that you can `archive` with 64 ledger frequency but `checkpointManager` with 8.

Also, I think we shouldn't ask for it specifically: we either hide it completely or set defaults automatically (like @tamirms suggested). This will be super confusing for people who don't know about `ARTIFICIALLY_ACCELERATE_TIME_FOR_TESTING`. What's more this won't be useful for 99% of users who don't need many complicated integrations tests like we do. 

As for magic constants, I'm not sure if I understand but we can create something like:
```go
const (
    normalCheckpointFrequency = 64
    acceleratedCheckpointFrequency = 8
)
```
+ add some comments so it should be clear.

historyarchive/range.go
Shaptic: Ah of course, didn't realize it was integer math :facepalm:.

ingest/ledgerbackend/captive_core_backend.go
Shaptic: From context, it seems like if we can avoid "requiring the consumer of the library to set it explicitly" when 99.9% of the time it'll just be set to 64, that'd be a Good Thing. So I don't think there's anything wrong with hardcoding 64 as the default as long as it's well-documented (like above). Constants are only "magical" when nothing explains them. :)

2opremio: OK, I will make the fields default to 64. Thanks for your input!

2opremio:
PTAL

tamirms (approved):

PR #3277, opened 2020-12-09
support/http: add option for adding tcp keep alive to server

acharb:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Adding the field `TCPKeepAlive` to support/http Config

### Why

Gives the option to specify a tcp keep alive timeout for the graceful server. Otherwise it always defaults to 0s

### Known limitations

n/a

### DIFF
--- a/support/http/main_test.go
+++ b/support/http/main_test.go
@@ -27,4 +27,5 @@ func TestRun_setup(t *testing.T) {
 	assert.Equal(t, time.Duration(0), srv.WriteTimeout)
 	assert.Equal(t, time.Duration(0), srv.IdleTimeout)
 	assert.Equal(t, defaultListenAddr, srv.Server.Addr)
+	assert.Equal(t, time.Duration(0), srv.TCPKeepAlive)
 }
--- a/support/http/main_test.go
+++ b/support/http/main_test.go
@@ -1,6 +1,7 @@
 package http
 
 import (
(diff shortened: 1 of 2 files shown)


support/http/main_test.go
```diff
@@ -27,4 +27,5 @@ func TestRun_setup(t *testing.T) {
 	assert.Equal(t, time.Duration(0), srv.WriteTimeout)
 	assert.Equal(t, time.Duration(0), srv.IdleTimeout)
 	assert.Equal(t, defaultListenAddr, srv.Server.Addr)
+	assert.Equal(t, time.Duration(0), srv.TCPKeepAlive)
 }
```
leighmcculloch, line 31: 💡 Thanks for adding assertions to the existing tests. There's no test here that shows what happens if a non-default TCPKeepAlive is provided, or any of the other parameters here. When changing an area of code we should consider if it is worth making the tests better, or if the tests that already exist sufficiently test the code we're about to change. This change is minimal, but we could add another test that tests non-default setup.

leighmcculloch (approved):
One suggestion (💡 ) but this looks good to me regardless.

howardtw (approved):
LGTM. Per @leighmcculloch 's comment, we can add an additional test to test the configs are properly propagated.

Issue #3228, opened 2020-11-17
Reliability issues in remote-core http server


2opremio:
While chaos-monkey-testing captive-core in the remote http configuration I killed the core child process to simulate a crash.

Instead of respawning the core process, the captive core http server:

1. Panicked due to the channel double-close (See backlog at the end of the description)
2. Got stuck, which is more worrisome. Instead of dying, the captive core server process lingered, not giving an opportunity to the supervisor (Kubernetes in this case) to re-spawn it:
```
root@horizon-with-remote-core-6bddf785b9-r9vlg:/# killall stellar-core
root@horizon-with-remote-core-6bddf785b9-r9vlg:/# ps aux
USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root         1 20.5  1.5 1539976 32040 ?       Ssl  18:13  12:26 /captivecore --port=8080
root      7161  0.0  0.1  18504  3372 pts/0    Ss   19:11   0:00 bash
root      7220  0.0  0.1  34400  2792 pts/0    R+   19:14   0:00 ps aux
root@horizon-with-remote-core-6bddf785b9-r9vlg:/# ps aux
USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root         1 18.6  1.3 1539976 28016 ?       Ssl  18:13  12:27 /captivecore --port=8080
root      7161  0.0  0.1  18504  3372 pts/0    Ss   19:11   0:00 bash
root      7221  0.0  0.1  34400  2748 pts/0    R+   19:20   0:00 ps aux
root@horizon-with-remote-core-6bddf785b9-r9vlg:/#
```

Panic log:

```
Panic: close of closed channel
goroutine 418366 [running]:
runtime/debug.Stack(0x1f, 0x0, 0x0)
	/usr/local/go/src/runtime/debug/stack.go:24 +0x9d
runtime/debug.PrintStack()
	/usr/local/go/src/runtime/debug/stack.go:16 +0x22
github.com/go-chi/chi/middleware.Recoverer.func1.1(0xc001431600, 0xf89860, 0xc001450380)
	/go/pkg/mod/github.com/go-chi/chi@v4.0.3+incompatible/middleware/recoverer.go:28 +0x1e3
panic(0xc70f60, 0xf61e60)
	/usr/local/go/src/runtime/panic.go:969 +0x166
[...]

PR #3278, opened 2020-12-10
ingest/ledgerbackend: Use context to handle termination and cleanup of captive core

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This PR implements robust shutdown behavior in `CaptiveStellarCore` by embedding a context in `stellarCoreRunner`.

`CaptiveStellarCore` exposes the following functions:

* `PrepareRange()`
* `IsPrepared()`
* `GetLedger()`
* `GetLatestLedgerSequence()`
* `Close()`

`PrepareRange()`, `IsPrepared()`, `GetLedger()`, and `GetLatestLedgerSequence()` cannot be called concurrently. Only one go routine should be calling those functions on a `CaptiveStellarCore` instance. If any of those functions are accessed concurrently then the behavior of `CaptiveStellarCore` will be undefined.

However, `Close()` can be called concurrently with any of the functions listed above. In other words,  it is safe for `Close()` to interrupt any pending `PrepareRange()`, `IsPrepared()`, `GetLedger()`, or `GetLatestLedgerSequence()` operation. Once `Close()` is called, any pending operation (e.g. `PrepareRange()` / `GetLedger()`) is expected to terminate quickly. The `Close()` operation should also complete in a timely manner.

After `Close()` is called, the captive core subprocess should be reaped, the captive core named pipe should be closed, the temporary directory used by captive core should be deleted, and all go routines related to buffering captive core logs or streaming ledgers from the named pipe should be terminated.

<del>Note that you can still use a `CaptiveStellarCore` instance after calling `Close()`. If you call `PrepareRange()` after calling `Close()` then `CaptiveStellarCore` will create a new captive core subprocess (along with the named pipe, buffering go routines, etc) and begin a new session.</del> [Edit: the behavior of `Close()` is now terminal. Once it is called the `CaptiveStellarCore` instance will fail on all subsequent operations]

The `CaptiveCoreConfig` struct has been extended to include an optional `Context` field. The context in `CaptiveCoreConfig` acts as a parent context for all `PrepareRange()` sessions. In other words, if the parent context is cancelled then, any pending `PrepareRange()` / `GetLedger()` operation will be terminated even though `Close()` was not called. Furthermore, once the parent context is terminated the `CaptiveStellarCore` instance will effectively be dead (e.g. it will not be possible start any more `PrepareRange()` sessions).

The behavior described above is implemented by embedding a context in the `stellarCoreRunner` struct. `stellarCoreRunner` will pass that context into `exec.CommandContext()` when spawning the captive core process. When we want to abort the running `stellarCoreRunner` instance we cancel the context associated with `stellarCoreRunner` which will kill the captive core subprocess (by calling os.Process.Kill) if the stellar core has not already exited on its own (see https://golang.org/pkg/os/exec/#CommandContext ). By using `exec.CommandContext()` we are able to remove some of the cleanup code from  `stellarCoreRunner` including the unix and windows implementations of `processIsAlive()`.

Previously, I mentioned that `CaptiveStellarCore.Close()` is able to safely interrupt any pending `CaptiveStellarCore` operation. This invariant is guaranteed because `CaptiveStellarCore` uses a read write lock to protect access to `stellarCoreRunner`. While the read lock is held, `CaptiveStellarCore` cannot assign its `stellarCoreRunner` field to a new value. `stellarCoreRunner.close()` is also thread safe and idempotent (so it doesn't matter if `CaptiveStellarCore.Close()` is called multiple times).

Once `stellarCoreRunner.close()` is invoked it will cancel the `stellarCoreRunner` context and it will also close the channel used by `CaptiveStellarCore` to read ledgers from captive core. At this point any blocking calls to `GetLedger()` will be able to terminate quickly.

The code which determines how `GetLedger()` terminates on error is encapsulated in the `checkMetaPipeResult()` function.
There are basically 3 types of errors which can arise in `GetLedger()`:

1. User initiated shutdown by canceling the parent context or calling Close().
2. The stellar core process exited unexpectedly.
3.  Some error was encountered while consuming the ledgers emitted by captive core (e.g. parsing invalid xdr)

All these cases are handled in `checkMetaPipeResult()`.  

Some other points to note while reading the PR:

* the implementation of `bufferedLedgerMetaReader` is now more decoupled from `CaptiveStellarCore` and `stellarCoreRunner`
* `bufferedLedgerMetaReader` is now created and owned by `stellarCoreRunner`. when `stellarCoreRunner` terminates, it will dispose of its `bufferedLedgerMetaReader` instance
*  `CaptiveStellarCore` does not have to listen to any exit channels. `CaptiveStellarCore` only needs to handle the case when the `stellarCoreRunner` meta pipe is exhausted.
* In the previous code the go routine created by `stellarCoreRunner` to consume stdout and stderr from captive core was never terminated. This PR fixes the go routine leak.
* The named pipe is now properly closed by `stellarCoreRunner`

### Known limitations

I need to fix the tests. I plan to do so once I receive approval on the overall approach of the PR.

### DIFF
--- a/ingest/ledgerbackend/buffered_meta_pipe_reader.go
+++ b/ingest/ledgerbackend/buffered_meta_pipe_reader.go
@@ -2,6 +2,7 @@ package ledgerbackend
 
 import (
 	"bufio"
+	"context"
 	"io"
 	"time"
 
@@ -62,20 +63,18 @@ type metaResult struct {
 // until xdr.LedgerCloseMeta objects channel is empty. This prevents memory
 // exhaustion when network closes a series a large ledgers.
 type bufferedLedgerMetaReader struct {
-	r      *bufio.Reader
-	c      chan metaResult
-	runner stellarCoreRunnerInterface
-	closed chan struct{}
+	r   *bufio.Reader
+	c   chan metaResult
+	ctx context.Context
 }
 
 // newBufferedLedgerMetaReader creates a new meta reader that will shutdown
 // when stellar-core terminates.
-func newBufferedLedgerMetaReader(runner stellarCoreRunnerInterface) *bufferedLedgerMetaReader {
+func newBufferedLedgerMetaReader(ctx context.Context, reader io.ReadCloser) *bufferedLedgerMetaReader {
 	return &bufferedLedgerMetaReader{
-		c:      make(chan metaResult, ledgerReadAheadBufferSize),
-		r:      bufio.NewReaderSize(runner.getMetaPipe(), metaPipeBufferSize),
-		runner: runner,
-		closed: make(chan struct{}),
+		c:   make(chan metaResult, ledgerReadAheadBufferSize),
+		r:   bufio.NewReaderSize(reader, metaPipeBufferSize),
+		ctx: ctx,
 	}
 }
 
@@ -84,33 +83,17 @@ func newBufferedLedgerMetaReader(runner stellarCoreRunnerInterface) *bufferedLed
 //   * Meta pipe buffer is full so it will wait until it refills.
 //   * The next ledger available in the buffer exceeds the meta pipe buffer size.
 //     In such case the method will block until LedgerCloseMeta buffer is empty.
-func (b *bufferedLedgerMetaReader) readLedgerMetaFromPipe(untilSequence uint32) (*xdr.LedgerCloseMeta, error) {
+func (b *bufferedLedgerMetaReader) readLedgerMetaFromPipe() (*xdr.LedgerCloseMeta, error) {
 	frameLength, err := xdr.ReadFrameLength(b.r)
 	if err != nil {
-		select {
-		case <-b.runner.getProcessExitChan():
-			return nil, wrapStellarCoreRunnerError(b.runner)
-		default:
-			return nil, errors.Wrap(err, "error reading frame length")
-		}
+		return nil, errors.Wrap(err, "error reading frame length")
 	}
 
 	for frameLength > metaPipeBufferSize && len(b.c) > 0 {
 		// Wait for LedgerCloseMeta buffer to be cleared to minimize memory usage.
 		select {
-		case <-b.runner.getProcessExitChan():
-			if untilSequence != 0 {
-				// If untilSequence != 0 it's possible that Stellar-Core process
-				// exits but there are still ledgers in a buffer (catchup). In such
-				// case we ignore cases when Stellar-Core exited with no errors.
-				processErr := b.runner.getProcessExitError()
-				if processErr != nil {
-					return nil, errors.Wrap(processErr, "stellar-core process exited with an error")
-				}
-				time.Sleep(time.Second)
-				continue
-			}
-			return nil, wrapStellarCoreRunnerError(b.runner)
+		case <-b.ctx.Done():
+			return nil, b.ctx.Err()
 		case <-time.After(time.Second):
 		}
 	}
@@ -118,17 +101,7 @@ func (b *bufferedLedgerMetaReader) readLedgerMetaFromPipe(untilSequence uint32) 
 	var xlcm xdr.LedgerCloseMeta
 	_, err = xdr.Unmarshal(b.r, &xlcm)
 	if err != nil {
-		if err == io.EOF {
-			err = errors.Wrap(err, "got EOF from subprocess")
-		}
-		err = errors.Wrap(err, "unmarshalling framed LedgerCloseMeta")
-
-		select {
-		case <-b.runner.getProcessExitChan():
-			return nil, wrapStellarCoreRunnerError(b.runner)
-		default:
-			return nil, err
-		}
+		return nil, errors.Wrap(err, "unmarshalling framed LedgerCloseMeta")
 	}
 	return &xlcm, nil
 }
@@ -137,52 +110,29 @@ func (b *bufferedLedgerMetaReader) getChannel() <-chan metaResult {
 	return b.c
 }
 
-func (b *bufferedLedgerMetaReader) waitForClose() {
-	// If buffer is full, keep reading to make sure it receives
-	// a shutdown signal from stellarCoreRunner.
-loop:
-	for {
-		select {
-		case <-b.c:
-		case <-b.closed:
-			break loop
-		}
-	}
-}
-
 // Start starts an internal go routine that reads binary ledger data into
-// internal buffers. The go routine returns when Stellar-Core process exits
-// however it won't happen instantly when data is read. A blocking method:
-// waitForClose() can be used to block until go routine returns.
-func (b *bufferedLedgerMetaReader) start(untilSequence uint32) {
+// internal buffers. The go routine returns when it encounters an error (including io.EOF)
+// or its context is terminated.
+func (b *bufferedLedgerMetaReader) start() {
 	printBufferOccupation := time.NewTicker(5 * time.Second)
-	defer func() {
-		printBufferOccupation.Stop()
-		close(b.closed)
-	}()
+	defer printBufferOccupation.Stop()
+	defer close(b.c)
 
 	for {
 		select {
 		case <-printBufferOccupation.C:
 			log.Debug("captive core read-ahead buffer occupation:", len(b.c))
+		case <-b.ctx.Done():
+			return
 		default:
 		}
 
-		meta, err := b.readLedgerMetaFromPipe(untilSequence)
+		meta, err := b.readLedgerMetaFromPipe()
 		if err != nil {
-			// When `GetLedger` sees the error it will close the backend. We shouldn't
-			// close it now because there may be some ledgers in a buffer.
 			b.c <- metaResult{nil, err}
 			return
 		}
 
 		b.c <- metaResult{meta, nil}
-
-		if untilSequence != 0 {
-			if meta.LedgerSequence() >= untilSequence {
-				// we are done
-				return
-			}
-		}
 	}
 }
--- a/ingest/ledgerbackend/buffered_meta_pipe_reader.go
+++ b/ingest/ledgerbackend/buffered_meta_pipe_reader.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
+
 	"github.com/stellar/go/support/log"
 	"github.com/stellar/go/xdr"
 )
@@ -70,7 +71,7 @@ type bufferedLedgerMetaReader struct {
 
 // newBufferedLedgerMetaReader creates a new meta reader that will shutdown
 // when stellar-core terminates.
-func newBufferedLedgerMetaReader(ctx context.Context, reader io.ReadCloser) *bufferedLedgerMetaReader {
+func newBufferedLedgerMetaReader(ctx context.Context, reader io.Reader) *bufferedLedgerMetaReader {
 	return &bufferedLedgerMetaReader{
 		c:   make(chan metaResult, ledgerReadAheadBufferSize),
 		r:   bufio.NewReaderSize(reader, metaPipeBufferSize),
@@ -110,9 +111,8 @@ func (b *bufferedLedgerMetaReader) getChannel() <-chan metaResult {
 	return b.c
 }
 
-// Start starts an internal go routine that reads binary ledger data into
-// internal buffers. The go routine returns when it encounters an error (including io.EOF)
-// or its context is terminated.
+// Start starts a loop that reads binary ledger data into internal buffers.
+// The function returns when it encounters an error (including io.EOF) or its context is terminated.
 func (b *bufferedLedgerMetaReader) start() {
 	printBufferOccupation := time.NewTicker(5 * time.Second)
 	defer printBufferOccupation.Stop()
--- a/ingest/ledgerbackend/buffered_meta_pipe_reader.go
+++ b/ingest/ledgerbackend/buffered_meta_pipe_reader.go
@@ -89,9 +89,7 @@ func (b *bufferedLedgerMetaReader) readLedgerMetaFromPipe() (*xdr.LedgerCloseMet
 
 	for frameLength > metaPipeBufferSize && len(b.c) > 0 {
 		// Wait for LedgerCloseMeta buffer to be cleared to minimize memory usage.
-		select {
-		case <-time.After(time.Second):
-		}
+		<-time.After(time.Second)
 	}
 
 	var xlcm xdr.LedgerCloseMeta
--- a/ingest/ledgerbackend/buffered_meta_pipe_reader.go
+++ b/ingest/ledgerbackend/buffered_meta_pipe_reader.go
@@ -2,7 +2,6 @@ package ledgerbackend
 
 import (
 	"bufio"
-	"context"
 	"io"
 	"time"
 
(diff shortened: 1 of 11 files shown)


bartekn (commented):
Thanks for this proposal. Here's a list of what I like and don't like. Also added some inline comments. Please note that I haven't run the code so I can be wrong below.

Like:
* `bufferedMetaPipeReader` used in `stellarCoreRunner` directly. I didn't do that in the PR that introduced `bufferedMetaPipeReader` because I was thinking that we should keep it separate to give access to _raw_ meta stream because maybe we'll use it in the future. But we can split it again when needed.

Don't like:
* Looks like in order to properly close the backend you need to call `Close()` and `cancel()` the `Context`. This can be confusing because `Context` is not part of the `LedgerBackend` interface so it will be hard to reuse the code with different backend. It will be a problem even in simplest apps that want to close the backend on CMD+C.
* I think that two shutdown signals, in general, create all kinds of new concurrency issues. One example: if we cancel context and call `Close()` it's possible that process exit error returned by `r.cmd.Wait()` will be `killed` instead of `ErrCanceled` because `Close()` code executed first (I added an inline comment with example when this can happen). We can add a bunch of code to enforce the order we want but `tomb` has a nice property of `Kill`: it doesn't overwrite the error on following calls to `Kill` so we'll still know the original error.
* I added an inline comment to `checkMetaPipeResult` but wanted to discuss it here too. This is quite complicated because it has to check 3 different errors (`stellarCoreRunner.context().Err()`, `result.err` and `getProcessExitError()`) and `ok` value. Developer not only need to know the internals of `stellarCoreRunner` and `bufferedMetaPipeReader` but also the order in which events execute (context cancellation, channel close, etc.). This was really difficult for me to understand. In `tomb` you receive a single error (in our case it's `nil`, `context.Cancelled` or actual error) and using a single method (so no issues with potential concurrency issues I explained in the inline comment).
* In general I don't like we're back to using using wait groups and closing channels which can be easily misused and were misued in the past: https://github.com/stellar/go/issues/3228 https://github.com/stellar/go/issues/3159 but [not only](https://github.com/issues?q=%22close+of+closed+channel%22) in our project. `tomb` can protect us from this pretty well.

One extra note, when I was searching for solutions to this problem I came across: https://dave.cheney.net/2017/08/20/context-isnt-for-cancellation which I think it's worth checking. I don't fully agree with this statement because I think that context cancellation (by timeout/deadline/cancel) is a great way to handle HTTP request cancellation. But in cases when ack is needed and the final error can be overwritten along the way I think context is really hard to maintain.

ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -462,65 +451,87 @@ func (c *CaptiveStellarCore) GetLedger(sequence uint32) (bool, xdr.LedgerCloseMe
...
 	}
 	// All paths above that break out of the loop (instead of return)
 	// set e to non-nil: there was an error and we should close and
 	// reset state before retuning an error to our caller.
-	c.Close()
+	c.stellarCoreRunner.close()
 	return false, xdr.LedgerCloseMeta{}, errOut
 }
 
+func (c *CaptiveStellarCore) checkMetaPipeResult(result metaResult, ok bool) error {
```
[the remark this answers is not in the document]
tamirms: yes, that is a valid case. what's wrong with that scenario?

ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -278,49 +302,87 @@ func (r *stellarCoreRunner) runFrom(from uint32, hash string) error {
...
+	r.cancel()
+	r.lock.Unlock()
+
+	// wait for the stellar core process to terminate
+	r.wg.Wait()
+
+	// drain meta pipe channel to make sure the ledger buffer goroutine exits
+	for range r.getMetaPipe() {
+
 	}
```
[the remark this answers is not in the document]
tamirms: I don't see how this is an issue. If we are Closing the captive core instance we don't care about reading the ledgers and we want to terminate as soon as possible.

tamirms:
@bartekn 

> Looks like in order to properly close the backend you need to call Close() and cancel() the Context. This can be confusing because Context is not part of the LedgerBackend interface so it will be hard to reuse the code with different backend. It will be a problem even in simplest apps that want to close the backend on CMD+C.

This isn't quite right. If you call `Close()` on a captive core instance it will terminate any pending PrepareRange / GetLedger operations. That is guaranteed. You can still create a new session but in order to do so you must call  `PrepareRange()` **after** calling `Close()`. The benefit of the parent context is that if you cancel it, the captive core instance becomes completely disabled making it impossible to start any new sessions. This solves the issue I mentioned here https://github.com/stellar/go/pull/3258#issuecomment-741683416 .

> I think that two shutdown signals, in general, create all kinds of new concurrency issues. One example: if we cancel context and call Close() it's possible that process exit error returned by r.cmd.Wait() will be killed instead of ErrCanceled because Close() code executed first (I added an inline comment with example when this can happen). We can add a bunch of code to enforce the order we want but tomb has a nice property of Kill: it doesn't overwrite the error on following calls to Kill so we'll still know the original error.

I think you misunderstood the semantics of `processExitError`. That variable is supposed to contain the process exit error as determined by `cmd.Wait()`. If you want to know if a captive core instance was terminated by `Close()` you can check the context error. In other words, `context().Err() != nil` if and only if `Close()` was called.


> I added an inline comment to checkMetaPipeResult but wanted to discuss it here too. This is quite complicated because it has to check 3 different errors (stellarCoreRunner.context().Err(), result.err and getProcessExitError()) and ok value. Developer not only need to know the internals of stellarCoreRunner and bufferedMetaPipeReader but also the order in which events execute (context cancellation, channel close, etc.). This was really difficult for me to understand. In tomb you receive a single error (in our case it's nil, context.Cancelled or actual error) and using a single method (so no issues with potential concurrency issues I explained in the inline comment).

I think the comment about the complexity is a fair point. However, I think it's acceptable because this complexity is completely encapsulated within 1 function `checkMetaPipeResult()`. Also, `stellarCoreRunner` is not an exported type, it is coupled to `CaptiveStellarCore`. I view it as an internal type and part of the implementation details of `CaptiveStellarCore`. I can add some comments to explain the function. I don't agree with your comment about concurrency issues though. 

> In general I don't like we're back to using using wait groups and closing channels which can be easily misused and were misued in the past: #3228 #3159 but not only in our project. tomb can protect us from this pretty well.

I believe the issues we had in the past could be attributed to the difficulty of working with concurrency in general more so than the specific use of waitgroups or channels. In your PR where you introduced tomb we were still able to find a bunch of concurrency related bugs.

ingest/ledgerbackend/captive_core_backend.go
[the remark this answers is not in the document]
bartekn: You don't want `GetLedger` or `PrepareRange` to return errors in case of cancellation (except error that specifies this was cancelled). `killed` error can confuse users that something went wrong and I guess we'll see it every once in a while when shutting down Horizon (we do handler `context.Cancelled` though and don't log it with `error` level).

ingest/ledgerbackend/stellar_core_runner.go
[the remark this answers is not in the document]
bartekn: Note that this is done it two different routines. So if `GetLedger` will read ledgers out of order it will error.

ingest/ledgerbackend/captive_core_backend.go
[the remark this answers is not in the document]
tamirms: If `Close()` is called on a `CaptiveStellarCore` which is running `GetLedger()` or `PrepareRange()` the only possible error `GetLedger()` or `PrepareRange()`  can return is `context.Cancelled`. There is no scenario where a `killed` error will be returned

ingest/ledgerbackend/stellar_core_runner.go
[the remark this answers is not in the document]
tamirms: > Note that this is done it two different routines. So if GetLedger will read ledgers out of order it will error.

This will only occur if we're `Closing()` captive core and in that scenario `GetLedger()` doesn't care about the ledgers and needs to exit asap

bartekn:
> This isn't quite right. If you call Close() on a captive core instance it will terminate any pending PrepareRange / GetLedger operations. That is guaranteed. You can still create a new session but in order to do so you must call PrepareRange() after calling Close(). The benefit of the parent context is that if you cancel it, the captive core instance becomes completely disabled making it impossible to start any new sessions. This solves the issue I mentioned here #3258 (comment) .

OK, I understand now. I was also thinking about it after your comment in #3258 and I think what we could do is to make `Close()` actually close the ledger backend (any backend) so it's not usable anymore and you need to create a new one. I checked how we use it in Horizon and we don't really need to close it other than when closing the app and I believe this will be the same for other apps. I think this will make the interface much more clear and an extra `Context` won't be required.

> I think you misunderstood the semantics of processExitError. That variable is supposed to contain the process exit error as determined by cmd.Wait(). If you want to know if a captive core instance was terminated by Close() you can check the context error. In other words, context().Err() != nil if and only if Close() was called.

OK, I can see it now after you explained what `CaptiveStellarCore.Context` is used for.

> I don't agree with your comment about concurrency issues though.

Again, it's clear after `CaptiveStellarCore.Context` explanations. But it's more complicated compare to `tomb` in my opinion.

> I believe the issues we had in the past could be attributed to the difficulty of working with concurrency in general more so than the specific use of waitgroups or channels. In your PR where you introduced tomb we were still able to find a bunch of concurrency related bugs.

Agreed! I meant: in my opinion it's much easier to make mistakes like closing channel twice when operating on channels directly compared to when using `tomb` because it's API protects from something like this. I'm not saying it's the case here but it may happen when we refactor the code in the future.

tamirms:
@bartekn I have updated the PR to make the context in `CaptiveCoreConfig` optional. Also, I have changed the behavior of `CaptiveStellarCore.Close()` to make it terminal. Once `CaptiveStellarCore.Close()` is invoked the `CaptiveStellarCore` instance will no longer be functional

bartekn (approved):

bartekn (approved):

ingest/ledgerbackend/buffered_meta_pipe_reader.go
```diff
@@ -62,20 +64,18 @@ type metaResult struct {
...
-func newBufferedLedgerMetaReader(runner stellarCoreRunnerInterface) *bufferedLedgerMetaReader {
+func newBufferedLedgerMetaReader(ctx context.Context, reader io.Reader) *bufferedLedgerMetaReader {
 	return &bufferedLedgerMetaReader{
-		c:      make(chan metaResult, ledgerReadAheadBufferSize),
-		r:      bufio.NewReaderSize(runner.getMetaPipe(), metaPipeBufferSize),
-		runner: runner,
-		closed: make(chan struct{}),
+		c:   make(chan metaResult, ledgerReadAheadBufferSize),
+		r:   bufio.NewReaderSize(reader, metaPipeBufferSize),
+		ctx: ctx,
```
2opremio, line 78: In general, it's not considered a good practice to store the context unless it's an ephemeral object (e.g. a request).

Normally, the context is passed to each operation so that it can be configured per operation.

ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -118,31 +124,48 @@ type CaptiveCoreConfig struct {
 	// Log is an (optional) custom logger which will capture any output from the Stellar Core process.
 	// If Log is omitted then all output will be printed to stdout.
 	Log *log.Entry
+	// Context is the (optional) context which controls the lifetime of a CaptiveStellarCore instance. Once the context is done
+	// the CaptiveStellarCore instance will not be able to stream ledgers from Stellar Core or spawn new
+	// instances of Stellar Core. If Context is omitted CaptiveStellarCore will default to using context.Background.
+	Context context.Context
```
2opremio, line 130: same here

ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -194,25 +218,12 @@ func (c *CaptiveStellarCore) openOfflineReplaySubprocess(from, to uint32) error
...
-	if c.lastLedger == nil && c.nextLedger != 0 && c.nextLedger <= from {
-		// Use existing session, GetLedger will fast-forward
-		return nil
-	}
-
-	err := c.Close()
-	if err != nil {
-		return errors.Wrap(err, "error closing existing session")
-	}
-
```
2opremio, line 215: NIIIIIICE!

ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -348,46 +376,37 @@ func (c *CaptiveStellarCore) runFromParams(from uint32) (runFrom uint32, ledgerH
...
-			if processErr != nil {
-				return errors.Wrap(processErr, "stellar-core process exited with an error")
-			}
-		default:
-		}
-		// Wait for the first ledger or an error
-		if len(c.ledgerBuffer.getChannel()) > 0 {
-			break
-		}
-		time.Sleep(c.waitIntervalPrepareRange)
```
2opremio, line 383: Much cleaner without this. Great!

ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -38,6 +35,11 @@ const (
 	stellarCoreRunnerModeOffline
 )
 
+type pipe struct {
+	Reader io.ReadCloser
+	Writer io.Closer
+}
```
2opremio, line 41: Why can't you use `io.Pipe`? Also, how come a Writer is only a `Closer`, don't you need to write to it? In that case I would probably rewire/rename the abstraction.

2opremio:
I think this is a great step forward. The code is way cleaner and (IMO) easier to understand.

I am weary of bugs being introduced due to the changes being considerable ... but that's what a beta is for.

I think we should try to do some release testing specifically for this PR.

ingest/ledgerbackend/stellar_core_runner.go
tamirms: the windows implementation doesn't fit into the `io.Pipe` abstraction. if you look at stellar_core_runner_windows.go you'll find that the named pipe is modeled using a client / server abstraction.

ingest/ledgerbackend/buffered_meta_pipe_reader.go
tamirms: the context isn't necessary but there is a special case in `readLedgerMetaFromPipe()` where the context is used for early termination:

```golang
	for frameLength > metaPipeBufferSize && len(b.c) > 0 {
		// Wait for LedgerCloseMeta buffer to be cleared to minimize memory usage.
		select {
		case <-b.ctx.Done():
			return nil, b.ctx.Err()
		case <-time.After(time.Second):
		}
	}
```

without the context, there would be at most a 1 second delay to terminating the `bufferedLedgerMetaReader.start()` go routine after stellar core exits.

However, in practice, I'd expect the specific scenario which triggers the 1 second delay to be rare because it only occurs if we are in the middle of processing a 10 mb ledger and there are still ledgers in the buffer which have not been ingested. Maybe it makes sense to remove the context to simplify the code? @2opremio what do you think?

bartekn:
> I think we should try to do some release testing specifically for this PR.

Agree, the plan is to test the last commit when all the remaining PRs are merged (and then master is merged into the release branch).

ingest/ledgerbackend/captive_core_backend.go
tamirms: we could consider moving the context so that it's a parameter to `PrepareRange()` instead. but i'd prefer implementing that change in a follow up PR if necessary

ingest/ledgerbackend/stellar_core_runner.go
tamirms: I tried to address this in https://github.com/stellar/go/pull/3278/commits/b2026be40e9c7e868ffb2044dd90f1a901b20d81

ingest/ledgerbackend/buffered_meta_pipe_reader.go
tamirms: removed in https://github.com/stellar/go/pull/3278/commits/9e22e1984c70540432e28f7346ec94534e2b1e2c

ingest/ledgerbackend/captive_core_backend.go
2opremio: Sure, please do.

2opremio (approved):

PR #3283, opened 2020-12-16
services/horizon/internal/docs: Add Captive Core description and expand flags section

2opremio:

Issue #3205, opened 2020-11-10
Create captive-core docs/explainers


bartekn:

Shaptic:
Closed with #3285. If we want more doc improvements we should open new, more-specific issues.

PR #3285, opened 2020-12-17
services/horizon/docs: Adds migration/deployment guide for Captive Core.

### LINKED ISSUE #3204: Create Horizon Captive-Core migration guide

Shaptic:
### What
This should walk a user to transitioning from Horizon v1.x -> v2.0-beta. There are two scenarios covered:
  - everything on a single machine, so essentially the user can drop Core and replace it w/ Captive Core
  - three machines: two Horizons (with only one of them ingesting, just for diversity) and one (remote) Captive Core. It should be relatively self-explanatory to readers that Captive Core just replaces their existing Core instance. Though I did test all of these on the same machine (just varied ports, etc.), it should still be accurate.

Here's a [direct link](https://github.com/stellar/go/blob/66f3bab24f30f9a7ece39ecb2ceb587b15302d45/services/horizon/internal/docs/captive_core_migration.md) to the rendered Markdown. This should be merged into `captive_core.md` once @2opremio's #3283 is merged. Then, that closes #3204 and #3205.

### DIFF
--- a/services/horizon/internal/docs/captive_core_migration.md
+++ b/services/horizon/internal/docs/captive_core_migration.md
@@ -1,0 +1,99 @@
+**TODO:** This should be merged into `captive_core.md` once @fons and I wrap up our individual portions (or decide how break into separate files; maybe we need a `captive_core/` doc subdirectory).
+
+# Migration 
+In this section, we'll discuss migrating existing systems running the [latest](https://github.com/stellar/go/releases/latest) stable version of Horizon ([1.13](https://github.com/stellar/go/releases/tag/horizon-v1.13.0) as of this writing) to the new 2.0 beta. 
+
+**Environment assumptions**:
+
+  - For simplicity, we assume a single-machine Ubuntu setup running both Horizon and Core with a single local PostgreSQL server. The URI in the configs outlined later give further insight into the setup. Loosening this assumption is covered briefly in a [later section](#multi-machine-setup).
+
+  - We assume your machine has enough memory to hold Captive Core's in-memory database (~3GiB), which is a larger memory requirement than a traditional Core setup (which would have an on-disk database).
+
+  - We assume that your node joined the Stellar network recently (that is, it has a few thousand ledgers synced / ingeste), though this doesn't really matter. This assumption can be easily relaxed to any range, but we clarify it here so the chosen ledger ranges below make sense.
+
+
+## Installing
+The process for upgrading both Stellar Core and Horizon are covered [here](https://github.com/stellar/packages/blob/master/docs/upgrading.md#upgrading); the only difference is that since we're migrating to Horizon v2.0-beta, we need to first add the unstable repository. This is described [here](https://github.com/stellar/packages/blob/master/docs/adding-the-sdf-stable-repository-to-your-system.md#adding-the-bleeding-edge-unstable-repository), but in brief, you just need to add the URL:
+
+```bash
+echo "deb https://apt.stellar.org xenial unstable" | sudo tee -a /etc/apt/sources.list.d/SDF-unstable.list
+```
+
+Then, you can install the Captive Core packages:
+
+```bash
+sudo apt install stellar-captive-core
+```
+
+And you're ready to upgrade.
+
+
+## Upgrading
+At this point, all that is left to do is to:
+
+ - modify the Horizon configuration to enable Captive Core (we will assume it lives in `/etc/default/stellar-horizon`, the default)
+ - create a Captive Core configuration stub
+ - stop the existing Stellar Core instance
+ - restart Horizon
+
+
+### Configure Horizon
+First, add the following lines to the Horizon configuration to enable a Captive Core subprocess:
+
+```bash
+echo "STELLAR_CORE_BINARY_PATH=$(which stellar-core)
+CAPTIVE_CORE_CONFIG_APPEND_PATH=/etc/default/stellar-captive-core.cfg
+ENABLE_CAPTIVE_CORE_INGESTION=1" | sudo tee -a /etc/default/stellar-horizon
+```
+
+### Configure Captive Core
+Captive Core runs with a trimmed down configuration "stub": at minimum, it must contain enough info to set up a quorum. For example,
+
+```toml
+[[HOME_DOMAINS]]
+HOME_DOMAIN="testnet.stellar.org"
+QUALITY="HIGH"
+
+[[VALIDATORS]]
+NAME="sdf_testnet_1"
+HOME_DOMAIN="testnet.stellar.org"
+PUBLIC_KEY="GDKXE2OZMJIPOSLNA6N6F2BVCI3O777I2OOC4BV7VOYUEHYX7RTRYA7Y"
+ADDRESS="core-testnet1.stellar.org"
+HISTORY="curl -sf http://history.stellar.org/prd/core-testnet/core_testnet_001/{0} -o {1}"
+
+[[VALIDATORS]]
+NAME="sdf_testnet_2"
+HOME_DOMAIN="testnet.stellar.org"
+PUBLIC_KEY="GCUCJTIYXSOXKBSNFGNFWW5MUQ54HKRPGJUTQFJ5RQXZXNOLNXYDHRAP"
+ADDRESS="core-testnet2.stellar.org"
+HISTORY="curl -sf http://history.stellar.org/prd/core-testnet/core_testnet_002/{0} -o {1}"
+
+[[VALIDATORS]]
+NAME="sdf_testnet_3"
+HOME_DOMAIN="testnet.stellar.org"
+PUBLIC_KEY="GC2V2EFSXN6SQTWVYA5EPJPBWWIMSD2XQNKUOHGEKB535AQE2I6IXV2Z"
+ADDRESS="core-testnet3.stellar.org"
+HISTORY="curl -sf http://history.stellar.org/prd/core-testnet/core_testnet_003/{0} -o {1}"
+
+```
+
+The rest of the configuration will be generated automagically at runtime.
+
+**Note:** Using your existing Stellar Core configuration will not work (**why not???**). Running Horizon will fail with the following error, or errors like it:
+
+    default: Config from /tmp/captive-stellar-core-38cff455ad3469ec/stellar-core.conf
+    default: Got an exception: Failed to parse '/tmp/captive-stellar-core-38cff455ad3469ec/stellar-core.conf' :Key HTTP_PORT already present at line 10 [CommandLine.cpp:1064]
+
+
+### Restarting Services
+Now, we can stop Core (which hopefully doesn't need an explanation) and restart Horizon:
+
+```bash
+stellar-horizon-cmd serve
+```
+
+The logs should show Captive Core running successfully as a subprocess, and eventually Horizon will be running as usual, except with Captive Core rapidly generating transaction metadata in-memory!
+
+
+## Multi-Machine Setup
+If you plan on running Horizon and Captive Core on separate machines, you'll need to change only a few things. Namely, rather than configuring the `STELLAR_CORE_BINARY` variable, you'll need to point Horizon at the Remote Captive Core instance via `REMOTE_CAPTIVE_CORE_URL`.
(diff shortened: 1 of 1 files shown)


2opremio:
I think it would make sense to put everything into the same file.

services/horizon/internal/docs/captive_core_migration.md
```diff
@@ -0,0 +1,183 @@
+# Migration 
+In this section, we'll discuss migrating existing systems running the [latest](https://github.com/stellar/go/releases/latest) stable version of Horizon ([1.13](https://github.com/stellar/go/releases/tag/horizon-v1.13.0) as of this writing) to the new 2.0 beta.
```
2opremio, line 2: I think we should be less specific when it comes 1.13 so that it doesn't bitrot so quickly.

Mentioning something in the lines of _migrating pre-2.0 Horizon to Horizon with Captive Core (>=2.0)_

services/horizon/internal/docs/captive_core_migration.md
```diff
@@ -0,0 +1,183 @@
+# Migration 
+In this section, we'll discuss migrating existing systems running the [latest](https://github.com/stellar/go/releases/latest) stable version of Horizon ([1.13](https://github.com/stellar/go/releases/tag/horizon-v1.13.0) as of this writing) to the new 2.0 beta. 
+
+**Environment assumptions**:
+
+  - We assume that the **PostgreSQL server** lives at the `db.local` hostname, and has a `horizon` database accessible by the `postgres:secret` username-password combo.
+
+  - We assume your machine has **enough memory** to hold Captive Core's in-memory database (~3GiB), which is a larger memory requirement than a traditional Core setup (which would have an on-disk database).
```
2opremio, line 8:
```suggestion
  - We assume your machine has **enough memory** to hold Captive Core's in-memory database (~3GiB of RAM), which is a larger memory requirement than a traditional Core setup (which would have an on-disk database).
```

services/horizon/internal/docs/captive_core_migration.md
```diff
@@ -0,0 +1,183 @@
...
+## Upgrading
+At this point, all that is left to do is to:
+
+ - modify the Horizon configuration to enable Captive Core (we will assume it lives in `/etc/default/stellar-horizon`, the default)
+ - create a Captive Core configuration stub
+ - stop the existing Stellar Core instance
+ - restart Horizon
+
+### Configure Captive Core
+Captive Core runs with a trimmed down configuration "stub": at minimum, it must contain enough info to set up a quorum (see [above](#todo-fons-section-link)). For example, if relying exclusively on SDF's validators:
```
2opremio, line 48: I guess `#todo-fons-section-link` refers to the intro, right?

services/horizon/internal/docs/captive_core_migration.md
```diff
@@ -0,0 +1,183 @@
...
+
+## Multi-Machine Setup
+If you plan on running Horizon and Captive Core on separate machines, you'll need to change only a few things. Namely, rather than configuring the `STELLAR_CORE_BINARY` variable, you'll need to point Horizon at the Remote Captive Core instance via `REMOTE_CAPTIVE_CORE_URL`.
+
+In this section, we'll work through a hypothetical architecture with two Horizon instances, one of which is an ingestion instance, and a single Captive Core instance.
+
+### Remote Captive Core
+First, we need to start running the Captive Core server.
+
+This must be installed from source, as it's not a published package yet (TODO: right?). These instructions presume a "sane" Golang environment (namely, one with `$GOPATH` defined):
```
2opremio, line 116: I don't think it is. @bartekn , do you know?

services/horizon/internal/docs/captive_core_migration.md
```diff
@@ -0,0 +1,183 @@
...
+Now, let's run a Captive Core instance:
+
+```bash
+stellar-captive-core \
+  --network-passphrase='Test SDF Network ; September 2015' \
+  --history-archive-urls='https://history.stellar.org/prd/core-testnet/core_testnet_001' \
+  --db-url='postgres://postgres:secret@db.local:5432/horizon?sslmode=disable' \
+  --port=8080 \
+  --stellar-core-binary-path=$(which stellar-core) \
+  --captive-core-config-append-path=/etc/default/stellar-captive-core.cfg
```
2opremio, line 133: nitpick: Given the amount of flags, I think it would be simpler to place them as env variables in a file and `source` it.

I _think_ that's what the rest of the documentation does (I am not 100% sure though).

services/horizon/internal/docs/captive_core_migration.md
```diff
@@ -0,0 +1,183 @@
...
+  --remote-captive-core-url=http://captive-core.local:8080 \
+  --stellar-core-url=http://captive-core.local:11626 \
+  --port=8000
+```
+
+At this point, you should be able to hit port 8000 on the above instance and watch the `ingest_latest_ledger` value grow.
+
+
+# Reingestion
+If you need to manually reingest some ledgers (for example, you want history for some ledgers that closed before your asset got issued), you can still do this with Captive Core.
```
2opremio, line 175: I think it's worth referring to the section in which the full history is reingested.

services/horizon/internal/docs/captive_core_migration.md
2opremio: Whatever approach we use, it should be consistent with https://github.com/stellar/go/blob/master/services/horizon/internal/docs/captive_core.md#using-captive-core-to-reingest-the-full-public-network-history

services/horizon/internal/docs/captive_core_migration.md
```diff
@@ -0,0 +1,183 @@
...
+
+To start off simply, we assume a **single-machine Ubuntu setup** running both Horizon and Core with a single local PostgreSQL server. Then, this assumption is loosened in a [later section](#multi-machine-setup).
+
+
+## Installing
+The process for upgrading both Stellar Core and Horizon are covered [here](https://github.com/stellar/packages/blob/master/docs/upgrading.md#upgrading); the only difference is that since we're migrating to Horizon v2.0-beta, we need to first add the unstable repository. This is described [here](https://github.com/stellar/packages/blob/master/docs/adding-the-sdf-stable-repository-to-your-system.md#adding-the-bleeding-edge-unstable-repository), but in brief, you just need to add the URL:
+
+```bash
+echo "deb https://apt.stellar.org xenial unstable" | sudo tee -a /etc/apt/sources.list.d/SDF-unstable.list
+```
```
bartekn, line 20: I think that link to stellar/packages is enough. It's possible that something will change there and we'll forget to update this doc.
```suggestion
```

services/horizon/internal/docs/captive_core_migration.md
```diff
@@ -0,0 +1,183 @@
...
+
+Then, you can install the Captive Core packages:
+
+```bash
+sudo apt install stellar-captive-core
+```
+
+And you're ready to upgrade.
+
+**Note**: Until v2.0-beta is tagged & released, you'll need to build Horizon from master for this to actually render the latest binaries. That's done pretty simply, given a valid Go environment:
```
bartekn, line 30:
```suggestion
**Note**: Until v2.0-beta is tagged & released, you'll need to build Horizon from `release-horizon-v2.0.0-beta` for this to actually render the latest binaries. That's done pretty simply, given a valid Go environment:
```

services/horizon/internal/docs/captive_core_migration.md
```diff
@@ -0,0 +1,183 @@
...
+NAME="sdf_testnet_3"
+HOME_DOMAIN="testnet.stellar.org"
+PUBLIC_KEY="GC2V2EFSXN6SQTWVYA5EPJPBWWIMSD2XQNKUOHGEKB535AQE2I6IXV2Z"
+ADDRESS="core-testnet3.stellar.org"
+HISTORY="curl -sf http://history.stellar.org/prd/core-testnet/core_testnet_003/{0} -o {1}"
+```
+
+(We'll assume this stub lives at `/etc/default/stellar-captive-core.toml`.) The rest of the configuration will be generated automagically at runtime.
+
+**Note:** Using your existing Stellar Core configuration will not work (**why not???**). Running Horizon will fail with the following error, or errors like it:
```
bartekn, line 79: I think we shouldn't add too much details here.
```suggestion
**Note:** Using your existing Stellar Core configuration will not work because Horizon require a specific settings. Running Horizon will fail with the following error, or errors like it:
```

services/horizon/internal/docs/captive_core_migration.md
```diff
@@ -0,0 +1,183 @@
...
+
+### Configure Horizon
+First, add the following lines to the Horizon configuration to enable a Captive Core subprocess:
+
+```bash
+echo "STELLAR_CORE_BINARY_PATH=$(which stellar-core)
+CAPTIVE_CORE_CONFIG_APPEND_PATH=/etc/default/stellar-captive-core.toml" | sudo tee -a /etc/default/stellar-horizon
+```
+
+(Note that setting `ENABLE_CAPTIVE_CORE_INGESTION=1` is not necessary as of Horizon 2.0 because it's the default. TODO: This isn't true on master, so gotta make sure.)
```
bartekn, line 92:
```suggestion
```
Beta code lives in `release-horizon-v2.0.0-beta`.

services/horizon/internal/docs/captive_core_migration.md
bartekn: I think it's actually part of `stellar-captive-core` -> `stellar-captive-core-api` binary.

bartekn (commented):
I agree with @2opremio that maybe we should have a single doc with all the info about captive core testing. So worth merging it with https://github.com/stellar/go/pull/3283 when done.

services/horizon/internal/docs/captive_core_migration.md
Shaptic: Yeah, since you cover the configuration stub. Once we merge the two, I think we can drop the actual snippet from the intro and leave it here (since it's an example, just like this guide), but still leave the description etc. and link to it from here as a reminder to the reader.

services/horizon/internal/docs/captive_core_migration.md
Shaptic: Hm yeah, fair enough. Since we're generally talking about environments set up via our official packages, we should standardize to those configurations: the config lives as key-value pairs in `/etc/default/stellar-horizon` and the `stellar-horizon-cmd` binary wraps that up. That means we should probably update your linked doc to this model as well.

services/horizon/internal/docs/captive_core_migration.md
Shaptic: Agreed, will do!

services/horizon/internal/docs/captive_core_migration.md
Shaptic: I'll fix the docs, but I'm not really a fan of this being the case: as a dev, I'd want to either (a) do nothing, config-wise, or (b) use the same one as I always have and let Horizon fill in the details. I wish I'd realized the implications while the PR was out! 

In the future (imo ofc), maybe we should aim to add only whatever is necessary when generating the config?

services/horizon/internal/docs/captive_core_migration.md
Shaptic: Yep, good point!

services/horizon/internal/docs/captive_core_migration.md
Shaptic: 3GiB refers to the size of the database, and RAM is already implied by "in-memory", right? 

I've rephrased it slightly to emphasize this; anyway, this section will probably be dropped after merging with yours!

services/horizon/internal/docs/captive_core_migration.md
Shaptic: Yep, fixed this to match your header for a simpler merge.

services/horizon/internal/docs/captive_core_migration.md
Shaptic: Yep! It's an older version, though, so I've included both ways of installing.

services/horizon/internal/docs/captive_core_migration.md
Shaptic: Done!

services/horizon/internal/docs/captive_core_migration.md
bartekn: The problem is we can't append to existing config files because you can't have two same sections in toml file. We also can't overwrite the file because there are too many config variables in Stellar-Core we have not tested that can potentially break the integration. The other issue with overwriting is that it can confuse users if they set a specific setting and we change it later.

I think what we can include in this doc is that users should remove everything above `[[HOME_DOMAINS]]` because it won't be required in captive core.

bartekn:
Going to merge the PR and then merge the doc with https://github.com/stellar/go/pull/3283 to provide users with all-in-one doc about Captive Core. If you want to add any updates please update the final doc.

PR #3291, opened 2020-12-18
services/horizon/internal/test/integration: Run captive core integration tests

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This commit adds a CircleCI job so that the integration tests are run twice (once with captive core enabled and the other time with captive core disabled) for every PR. 

The integration tests with captive core enabled are about 1.5 minutes slower than running the tests with captive core disabled.

### Known limitations

[N/A]


services/horizon/docker/captive-core-reingest-range-integration-tests.cfg
```diff
@@ -0,0 +1,9 @@
+ARTIFICIALLY_ACCELERATE_TIME_FOR_TESTING=true
+
+[[VALIDATORS]]
+NAME="local_core"
+HOME_DOMAIN="core.local"
+# From "SACJC372QBSSKJYTV5A7LWT4NXWHTQO6GHG4QDAVC2XDPX6CNNXFZ4JK"
+PUBLIC_KEY="GD5KD2KEZJIGTC63IGW6UMUSMVUVG5IHG64HUTFWCHVZH2N2IBOQN7PS"
+ADDRESS="localhost"
+QUALITY="MEDIUM"
```
bartekn, line 9: Do we need a config file for reingestion?

services/horizon/internal/test/integration/integration.go
```diff
@@ -115,6 +124,21 @@ func NewTest(t *testing.T, config Config) *Test {
 	return i
 }
 
+func (i *Test) RestartHorizon() {
+	i.app.Close()
+
+	// wait for horizon to shut down completely
+	time.Sleep(time.Second)
```
bartekn, line 131: Nit: alternatively we can add a channel that will be closed when `Serve()` is done in `startHorizon()`.

services/horizon/internal/test/integration/integration.go
```diff
@@ -131,8 +155,9 @@ func (i *Test) Shutdown() {
 	})
 }
 
-func (i *Test) startHorizon() {
-	horizonPostgresURL := i.config.PostgresURL
+func (i *Test) startHorizon(
+	captiveCoreBinaryPath, captiveCoreConfigPath, horizonPostgresURL string, buildGenesisBlock bool,
```
bartekn, line 159: I think we should rename `buildGenesisBlock` to `buildGenesisState` or `buildGenesisCheckpoint`. I wasn't sure if by block you mean state at ledger 1 or txs/ops in ledger 1.

services/horizon/internal/test/integration/integration.go
```diff
@@ -176,14 +200,29 @@ of accounts, subscribe to event streams and more.`,
...
 		"--admin-port",
 		strconv.Itoa(i.AdminPort()),
 		"--checkpoint-frequency",
 		"8", // due to ARTIFICIALLY_ACCELERATE_TIME_FOR_TESTING
-	})
+	}
+
+	if len(captiveCoreBinaryPath) > 0 {
+		args = append(args, "--enable-captive-core-ingestion")
+	}
```
bartekn, line 222: Can we set this explicitly to `true`/`false`? Won't work in 2.0.0-beta branch where it's set to `true` by default.

bartekn (approved):

ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -19,7 +19,7 @@ var _ LedgerBackend = (*CaptiveStellarCore)(nil)
 
 func (c *CaptiveStellarCore) roundDownToFirstReplayAfterCheckpointStart(ledger uint32) uint32 {
 	r := c.checkpointManager.GetCheckpointRange(ledger)
-	if r.Low == 0 {
+	if r.Low <= 1 {
```
2opremio, line 22: Ops!

services/horizon/docker/captive-core-integration-tests.cfg
```diff
@@ -0,0 +1,13 @@
...
+UNSAFE_QUORUM=true
+FAILURE_SAFETY=0
+
+[[VALIDATORS]]
+NAME="local_core"
+HOME_DOMAIN="core.local"
+# From "SACJC372QBSSKJYTV5A7LWT4NXWHTQO6GHG4QDAVC2XDPX6CNNXFZ4JK"
+PUBLIC_KEY="GD5KD2KEZJIGTC63IGW6UMUSMVUVG5IHG64HUTFWCHVZH2N2IBOQN7PS"
+ADDRESS="localhost"
+QUALITY="MEDIUM"
```
2opremio, line 13: +EOL

services/horizon/docker/captive-core-reingest-range-integration-tests.cfg
2opremio, line 9: Same

services/horizon/internal/integration/db_test.go
```diff
@@ -56,8 +58,29 @@ func TestReingestDB(t *testing.T) {
 	_, err = schema.Migrate(dbConn.DB.DB, schema.MigrateUp, 0)
 	tt.NoError(err)
 
+	// cap reachedLedger to the nearest checkpoint ledger because reingest range cannot ingest past the most
+	// recent checkpoint ledger when using captive core
```
2opremio, line 62: Then we should probably make sure a full checkpoint was ingested.

2opremio: And ... shouldn't captive core be able to do the ingestion as long as the end ledger exists? (even if it's not covered by a checkpoint).

In that case I think Horizon should just wait for the checkpoint (if it needs to) and warn about it.

services/horizon/internal/test/integration/integration.go
```diff
@@ -115,6 +124,21 @@ func NewTest(t *testing.T, config Config) *Test {
...
+	time.Sleep(time.Second)
+
+	i.startHorizon(
+		i.horizonConfig.CaptiveCoreBinaryPath,
+		i.horizonConfig.CaptiveCoreConfigAppendPath,
+		i.horizonConfig.DatabaseURL,
+		false,
+	)
+	i.waitForHorizon()
+}
```
2opremio, line 140: Nice, thanks.

services/horizon/internal/test/integration/integration.go
```diff
@@ -176,14 +200,29 @@ of accounts, subscribe to event streams and more.`,
 			hostname,
 			stellarCorePostgresPort,
 		),
+
+		"--stellar-core-binary-path",
+		captiveCoreBinaryPath,
+		"--captive-core-config-append-path",
+		captiveCoreConfigPath,
+		"--captive-core-http-port",
+		"0",
```
2opremio, line 209: If this is because it clashes with the http port of the other core instance, I think it's worth commenting it.

2opremio (approved):
Only nitpicks.

The comment about requiring a checkpointed ledger as the end range of reingestion 
(if confirmed)  is unrelated to this PR but I think we should create an issue about it.

2opremio:
BTW, huge thanks for getting this done 🥳 

I think it's a big milestone for testing captive core!

services/horizon/internal/integration/db_test.go
bartekn: I remember discussing this issue with the team or just one of you. I think what we agreed to is that if `to` is in front of the last checkpoint ledger we should error instead of changing `to` or waiting.

services/horizon/internal/test/integration/integration.go
bartekn: Can we run two cores at the same time in integration tests? We shouldn't, right?

services/horizon/internal/integration/db_test.go
2opremio: I don't recall, but after thinking about it, silently waiting for 5 minutes may not be the best approach. It sucks that we have to wait for a waitpoint though.

services/horizon/docker/captive-core-reingest-range-integration-tests.cfg
tamirms: we need to set `ARTIFICIALLY_ACCELERATE_TIME_FOR_TESTING=true` to be compatible with the stellar core container which is running the stand alone network.

The quorum set has no significance since we'll be doing reingestion which relies only on the history archives. But stellar-core still requires a non empty quorum set for the configuration file to be valid so I cannot leave the section empty.

bartekn: I wonder if we can add another `stellarCoreRunnerMode`: `stellarCoreRunnerModeOfflineIntegrationTests` that would generate config with the value set. Not urgent and maybe not needed so just sharing an idea.

tamirms: @bartekn another possibility is to have stellarCoreRunner set `ARTIFICIALLY_ACCELERATE_TIME_FOR_TESTING=true` in the config if the `CheckpointFrequency` in the captive core config is equal to 8.

services/horizon/internal/integration/db_test.go
tamirms: @bartekn @2opremio I created an issue to discuss the best approach here https://github.com/stellar/go/issues/3294

PR #3292, opened 2020-12-18
ingest/ledgerbackend: Dont automatically close after reading last ledger in range

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Remove the behavior in `CaptiveStellarCore` where the instance is automatically closed after reading the last ledger in the range.

### Why

Once `CaptiveStellarCore` is closed functions like `GetLatestLedgerSequence()` return errors like "session is closed, call PrepareRange first." 

So if `GetLatestLedgerSequence()` is called  after `PrepareRange(ledgerbackend.BoundedRange(1000, 1000))` it will return a "session is closed" error, which is bizarre because `CaptiveStellarCore` was never explicitly closed.

### Known limitations

[N/A]

### DIFF
--- a/ingest/ledgerbackend/captive_core_backend_test.go
+++ b/ingest/ledgerbackend/captive_core_backend_test.go
@@ -609,6 +609,45 @@ func TestCaptiveGetLedger(t *testing.T) {
 	mockRunner.AssertExpectations(t)
 }
 
+func TestCaptiveStellarCore_PrepareRangeAfterClose(t *testing.T) {
+	executablePath := "/etc/stellar-core"
+	networkPassphrase := network.PublicNetworkPassphrase
+	historyURLs := []string{"http://localhost"}
+
+	captiveStellarCore, err := NewCaptive(
+		CaptiveCoreConfig{
+			BinaryPath:         executablePath,
+			NetworkPassphrase:  networkPassphrase,
+			HistoryArchiveURLs: historyURLs,
+		},
+	)
+	assert.NoError(t, err)
+
+	assert.NoError(t, captiveStellarCore.Close())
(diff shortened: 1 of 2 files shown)


2opremio (approved):
Nice and neat!

ingest/ledgerbackend/captive_core_backend_test.go
```diff
@@ -747,13 +745,25 @@ func TestCaptiveGetLedger_ErrClosingAfterLastLedger(t *testing.T) {
...
-	err := captiveBackend.PrepareRange(BoundedRange(65, 66))
+	boundedRange := BoundedRange(65, 66)
+	err := captiveBackend.PrepareRange(boundedRange)
 	assert.NoError(t, err)
 
-	_, _, err = captiveBackend.GetLedger(66)
-	tt.EqualError(err, "error closing session: transient error")
+	assert.NoError(t, captiveBackend.Close())
+
+	_, _, err = captiveBackend.GetLedger(boundedRange.to)
```
bartekn, line 757: Can we test `PrepareRange` here too (after backend is closed)? We will have tests for this case for all methods.

bartekn (approved):

ingest/ledgerbackend/captive_core_backend_test.go
tamirms: ok, this scenario is hard to test with a `stellarCoreRunnerMock` but I can add a separate test case for it.

bartekn: OK, thanks!

PR #3304, opened 2020-12-22
services/horizon: Allow --stellar-core-binary-path to be prefilled from the PATH

### LINKED ISSUE #3290: services/horizon: Default --stellar-core-binary-path=stellar-core if it exists in $PATH.
Filling out parameters is :sleeping: :zzz: 

Sane defaults! It should be trivial to infer the stellar-core binary path if it exists in the environment `PATH`. 

I already do `--stellar-core-binary-path=$(which stellar-core)` sometimes and shouldn't need to.

Shaptic:
Closes #3290.

### DIFF
--- a/services/horizon/internal/flags.go
+++ b/services/horizon/internal/flags.go
@@ -5,6 +5,7 @@ import (
 	"go/types"
 	stdLog "log"
 	"os"
+	"os/exec"
 	"strings"
 
 	"github.com/sirupsen/logrus"
@@ -95,7 +96,7 @@ func Flags() (*Config, support.ConfigOptions) {
 			OptType:     types.String,
 			FlagDefault: "",
 			Required:    false,
-			Usage:       "path to stellar core binary (--remote-captive-core-url has higher precedence)",
+			Usage:       "path to stellar core binary (--remote-captive-core-url has higher precedence). If captive core is enabled, look for the stellar-core binary in $PATH by default.",
 			ConfigKey:   &config.CaptiveCoreBinaryPath,
 		},
 		&support.ConfigOption{
@@ -416,16 +417,34 @@ func ApplyFlags(config *Config, flags support.ConfigOptions) {
 
 	if config.EnableCaptiveCoreIngestion {
 		binaryPath := viper.GetString(StellarCoreBinaryPathName)
+
+		// If the user didn't specify a Stellar Core binary, we can check the
+		// $PATH and possibly fill it in for them.
+		if binaryPath == "" {
+			if result, err := exec.LookPath("stellar-core"); err == nil {
+				binaryPath = result
+				viper.Set(StellarCoreBinaryPathName, binaryPath)
+				config.CaptiveCoreBinaryPath = binaryPath
+			}
+		}
+
 		remoteURL := viper.GetString("remote-captive-core-url")
+
+		// NOTE: If both of these are set (regardless of user- or PATH-supplied
+		//       defaults for the binary path), the Remote Captive Core URL
+		//       takes precedence.
 		if binaryPath == "" && remoteURL == "" {
 			stdLog.Fatalf("Invalid config: captive core requires that either --stellar-core-binary-path or --remote-captive-core-url is set")
 		}
+
+		// If we don't supply an explicit core URL and we are running a local
+		// captive core process with the http port enabled, point to it.
 		if config.StellarCoreURL == "" && config.RemoteCaptiveCoreURL == "" && config.CaptiveCoreHTTPPort != 0 {
-			// If we don't supply an explicit core URL and we are running a local captive core process
-			// with the http port enabled, point to it.
 			config.StellarCoreURL = fmt.Sprintf("http://localhost:%d", config.CaptiveCoreHTTPPort)
+			viper.Set(StellarCoreURLFlagName, config.StellarCoreURL)
 		}
 	}
+
 	if config.Ingest {
 		// When running live ingestion a config file is required too
 		validateBothOrNeither(StellarCoreBinaryPathName, CaptiveCoreConfigAppendPathName)


bartekn (commented):
I think changing default of the flag to `stellar-core` would be equivalent solution but we won't need to exec commands.
```go
		&support.ConfigOption{
			Name:        StellarCoreBinaryPathName,
			OptType:     types.String,
			FlagDefault: "stellar-core",
			Required:    false,
			Usage:       "path to stellar core binary (--remote-captive-core-url has higher precedence)",
			ConfigKey:   &config.CaptiveCoreBinaryPath,
		},
```

services/horizon/internal/flags.go
```diff
@@ -404,6 +405,21 @@ func ApplyFlags(config *Config, flags support.ConfigOptions) {
 
 	if config.EnableCaptiveCoreIngestion {
 		binaryPath := viper.GetString(StellarCoreBinaryPathName)
+
+		// If the user didn't specify a Stellar Core binary, we can check the
+		// $PATH and possibly fill it in for them.
+		if binaryPath == "" {
+			cmd := exec.Command("which", "stellar-core")
```
2opremio, line 412: Please use [`exec.LookPath`](https://golang.org/pkg/os/exec/#LookPath) instead.

Shaptic: Brilliant, thanks for the link!

Shaptic:
@bartekn setting the default directly like that is more dangerous, I think? An explicit lookup (e.g. `LookPath` per @2opremio resulting in `/usr/bin/stellar-core`, as done in 9f43086d) gives us a pretty strong guarantee that the default is actually a _valid value_, and we also will only execute it in the case of `if config.EnableCaptiveCoreIngestion` which is the only safe path to set it as a default.

Otherwise, as you noted re: append-path, we would now need to do change the config parsing code to work around the fact that `viper.GetString(StellarCoreBinaryPathName) == ""` will **never** be true anymore. This could also lead to weird regressions since there are so many combinations of parameters.

I could be wrong, of course; the happy path definitely works with the `FlagDefault` set. I'm just worried about weird environment edge cases.

services/horizon/internal/flags.go
```diff
@@ -404,16 +405,32 @@ func ApplyFlags(config *Config, flags support.ConfigOptions) {
 
 	if config.EnableCaptiveCoreIngestion {
 		binaryPath := viper.GetString(StellarCoreBinaryPathName)
+
+		// If the user didn't specify a Stellar Core binary, we can check the
+		// $PATH and possibly fill it in for them.
+		if binaryPath == "" {
+			if result, err := exec.LookPath("stellar-core"); err == nil {
+				binaryPath = result
+				stdLog.Println("Found Stellar Core in $PATH:", binaryPath)
```
2opremio, line 414: I think we only use `stdLog` for fatal errors.

services/horizon/internal/flags.go
```diff
@@ -404,16 +405,32 @@ func ApplyFlags(config *Config, flags support.ConfigOptions) {
...
+				viper.Set(StellarCoreBinaryPathName, binaryPath)
```
2opremio, line 415: Is this necessary? (just asking). The line

```go
config.StellarCoreURL = fmt.Sprintf("http://localhost:%d", config.CaptiveCoreHTTPPort)
```

below, doesn't do it.

Would you mind making it consistent?

services/horizon/internal/flags.go
```diff
@@ -404,16 +405,32 @@ func ApplyFlags(config *Config, flags support.ConfigOptions) {
...
+				config.CaptiveCoreBinaryPath = binaryPath
+			}
+		}
+
 		remoteURL := viper.GetString("remote-captive-core-url")
+
+		// NOTE: If both of these are set, Remote Captive Core takes precedence.
 		if binaryPath == "" && remoteURL == "" {
 			stdLog.Fatalf("Invalid config: captive core requires that either --stellar-core-binary-path or --remote-captive-core-url is set")
```
2opremio, line 424: I would rephrase this to make it more precise after the changes in this PR.

services/horizon/internal/flags.go
```diff
@@ -95,7 +96,7 @@ func Flags() (*Config, support.ConfigOptions) {
 			OptType:     types.String,
 			FlagDefault: "",
 			Required:    false,
-			Usage:       "path to stellar core binary (--remote-captive-core-url has higher precedence)",
+			Usage:       "path to stellar core binary (--remote-captive-core-url has higher precedence). If captive core is enabled, look in $PATH by default.",
```
2opremio, line 99:
```suggestion
			Usage:       "path to stellar core binary (--remote-captive-core-url has higher precedence). If captive core is enabled, look for stellar-core in $PATH by default.",
```

services/horizon/internal/flags.go
```diff
@@ -95,7 +96,7 @@ func Flags() (*Config, support.ConfigOptions) {
 			OptType:     types.String,
 			FlagDefault: "",
```
2opremio, line 97: I am thinking it may be cleaner to simply set the default (based in the path search) here.

It would remove the need to clarify the fact that we look for the binary in PATH by default and would make the code more readable.

services/horizon/internal/flags.go
Shaptic: It is, actually. Due to the later line:

```go
if config.Ingest {
	// When running live ingestion a config file is required too
	validateBothOrNeither(StellarCoreBinaryPathName, CaptiveCoreConfigAppendPathName)
}
```

will check `viper.Get(StellarCoreBinaryPathName)` internally, so if the binary _isn't_ set by the user and _is_ found in the PATH and ingestion _is_ enabled, then this line will error out because it won't think there's a binary.

I made the latter line consistent in c175f071.

services/horizon/internal/flags.go
Shaptic: I had initially borrowed this from the docstring, but reworded it in 267c0e81 because you're right.

services/horizon/internal/flags.go
Shaptic: Have you seen [my comment above](https://github.com/stellar/go/pull/3304#issuecomment-749764667)? I'm not sure if setting the default "absolutely" will cause weird configuration bugs. 

I'd be happy to change it to just set the `FlagDefault` because I completely agree that it's clearer, etc., but I just want to be sure it doesn't cause weirder problems by essentially causing the parameter to always be valid.

PR #3308, opened 2021-01-04
strkey: Fix panic on invalid length decoded string

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Change the behaviour of `strkey.Decode` to return an error instead of 'panic' on invalid length decoded strings.

### Why

The behaviour is not consistent because `strkey.Decode` returns errors on other types of invalid inputs. It can also crash applications that do not handle `panic` properly.

### Known limitations

N/A


strkey/main.go
```diff
@@ -74,6 +74,11 @@ func Decode(expected VersionByte, src string) ([]byte, error) {
 		return nil, err
 	}
 
+	// check length
+	if len(raw) < 3 {
+		return nil, errors.New("decoded string value length invalid")
```
2opremio, line 79: This message is a bit convoluted how about `decoded string is too short`?

2opremio (approved):
Minor nitpick

strkey/main.go
bartekn: Changed!

PR #3311, opened 2021-01-06
Remove captive core how-it-works section

2opremio:
It's covered by the rest of the document

Leftover from #3283 (it was supposed to go in as a review commit but I forgot to push it)


2opremio:
Actually, now that I see it, it seems that this was succesfully updated in later PRs, so I think I will close this and leave it as is.

PR #3312, opened 2021-01-06
all: Use stdlib crypto/ed25519

bartekn:
`crypto/ed22519` (previously `golang.org/x/crypto/ed25519`) is part of stdlib since Go 1.13.


go.mod
```diff
@@ -79,7 +79,7 @@ require (
 	github.com/yudai/golcs v0.0.0-20150405163532-d1c525dea8ce // indirect
 	github.com/yudai/pp v2.0.1+incompatible // indirect
 	github.com/ziutek/mymysql v1.5.4 // indirect
-	golang.org/x/crypto v0.0.0-20191112222119-e1110fd1c708
+	golang.org/x/crypto v0.0.0-20191112222119-e1110fd1c708 // indirect
```
leighmcculloch, line 82: Which dependency is still using this? `go mod why` should tell us. This version looks pretty old. We should use the latest version of this: `v0.0.0-20201221181555-eec23a3978ad`.

leighmcculloch (approved):

go.mod
bartekn: TIL `go mod why`! It's:
```
# golang.org/x/crypto/ed25519
github.com/stellar/go/exp/services/recoverysigner/internal/serve
gopkg.in/square/go-jose.v2
golang.org/x/crypto/ed2551
```

leighmcculloch: Ah, there's probably not much we can do about that. Thanks for checking.

leighmcculloch: We could advocate for `square/go-jose` to upgrade to a newer version of Go and use the non-x package, but it depends how strictly they take their breaking changes since it would mean the package can't work on Go 1.12 or earlier.

PR #3302, opened 2020-12-21
horizon/docs: Minor aesthetic, phrasing, and grammatical improvements

Shaptic:
Mostly created so that we can use the PR UI to comment on places where we can make improvements.

Summarized list of changes / their rationale:

  - Most Markdown renderers prefer paragraphs to be all on one line (let your
    editor do the wrapping) as opposed to with manual breaks. This can be
    especially problematic with additional formatting e.g. `[links broken over
    a newline](#)`, etc. 

    GitHub's renderer is solid, but better safe than sorry for portability.
    Dropping whitespace changes from the diff should help w/ viewing. 

  - Unifying usage of "Captive Core" or "Captive Stellar-Core" (consistent
    capitalization, hyphenation, etc.).

  - Drop the example in the `CAPTIVE_CORE_CONFIG_APPEND_PATH` and link to the
    below one, instead, to avoid them diverging or otherwise getting invalid.

  - Fixed a few comma splices and other punctuation pedantry

  - Prefer the `systemctl` method of managing Horizon

  - Specify syntax highlighting methods where possible (e.g. the Bash script)

PR #3313, opened 2021-01-06
services/horizon: Restore old fixes to the Captive Core documentation

Shaptic:
In #3310, the Captive Core documentation merge undid some changes, likely due to the fact that I merged #3302 into master right before this merge.

This is pretty apparent when you look at [the diff of said file](https://github.com/stellar/go/pull/3310/files#diff-dcbfcf6a4881548aa52f3762303b0f1cf6ad6441a011cfee5273961012619e5eL50-R66) (e.g. capitilization fixes are undone, I dropped the TOML to keep them in one place, etc.), so this should be an easy approve :heavy_check_mark:.


bartekn (approved):
Thanks for the fix and sorry about that!

Shaptic:
np :+1:

PR #3314, opened 2021-01-06
services/friendbot: reduce friendbot minion creation batch to 50, allow for retries, add tests

acharb:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This PR is reducing the friendbot minion (channel account) creation batch from 100 to 50, and allows for retries.

Also, it adds tests for the `createMinionAccounts` method  

closes https://github.com/stellar/nebula/issues/84

### Why

When network congestion was high on testnet 100 account creations/transaction threw a timeout error. See [slack alert here](https://stellarfoundation.slack.com/archives/CANJGG94K/p1609949425128800)

Reducing the # of accounts created per transaction + allowing for some retries should reduce the chances of the timeout error in case this method is run during high congestion in the future

### Known limitations

n/a

### DIFF
--- a/services/friendbot/init_friendbot.go
+++ b/services/friendbot/init_friendbot.go
@@ -130,12 +130,12 @@ func createMinionAccounts(botAccount internal.Account, botKeypair *keypair.Full,
 
 		resp, err := hclient.SubmitTransactionXDR(txe)
 		if err != nil {
-			log.Println(resp)
+			log.Printf("%+v\n", resp)
 			switch e := err.(type) {
 			case *horizonclient.Error:
 				problemString := fmt.Sprintf("Problem[Type=%s, Title=%s, Status=%d, Detail=%s, Extras=%v]", e.Problem.Type, e.Problem.Title, e.Problem.Status, e.Problem.Detail, e.Problem.Extras)
 				// If we hit an error here due to network congestion, try again until we hit max # of retries allowed
-				if e.Problem.Type == "timeout" {
+				if e.Problem.Status == http.StatusGatewayTimeout {
 					err = errors.Wrap(errors.Wrap(e, problemString), "submitting create accounts tx")
 					if currentsubmitTxRetry >= submitTxRetriesAllowed {
 						return minions, errors.Wrap(err, fmt.Sprintf("after retrying %d times", currentsubmitTxRetry))
--- a/services/friendbot/init_friendbot.go
+++ b/services/friendbot/init_friendbot.go
@@ -45,9 +45,16 @@ func initFriendbot(
 	// already confirmed that friendbotSecret is a seed.
 	botKeypair := botKP.(*keypair.Full)
 	botAccount := internal.Account{AccountID: botKeypair.Address()}
+	// set default values
 	minionBalance := "101.00"
 	if numMinions == 0 {
 		numMinions = 1000
+	}
+	if minionBatchSize == 0 {
+		minionBatchSize = 50
+	}
+	if submitTxRetriesAllowed == 0 {
+		submitTxRetriesAllowed = 5
 	}
 	log.Printf("Found all valid params, now creating %d minions", numMinions)
 	minions, err := createMinionAccounts(botAccount, botKeypair, networkPassphrase, startingBalance, minionBalance, numMinions, minionBatchSize, submitTxRetriesAllowed, baseFee, hclient)
--- a/services/friendbot/init_friendbot.go
+++ b/services/friendbot/init_friendbot.go
@@ -20,6 +20,8 @@ func initFriendbot(
 	startingBalance string,
 	numMinions int,
 	baseFee int64,
+	minionBatchSize int,
+	submitTxRetriesAllowed int,
 ) (*internal.Bot, error) {
 	if friendbotSecret == "" || networkPassphrase == "" || horizonURL == "" || startingBalance == "" || numMinions < 0 {
 		return nil, errors.New("invalid input param(s)")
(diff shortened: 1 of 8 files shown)


services/friendbot/internal/account.go
```diff
@@ -30,7 +30,7 @@ func (a Account) GetSequenceNumber() (int64, error) {
 }
 
 // RefreshSequenceNumber gets an Account's correct in-memory sequence number from Horizon.
-func (a *Account) RefreshSequenceNumber(hclient *horizonclient.Client) error {
+func (a *Account) RefreshSequenceNumber(hclient horizonclient.ClientInterface) error {
```
acharb, line 33: changed these to `ClientInterface` to be able to test and mock horizon response

services/friendbot/init_friendbot.go
```diff
@@ -56,10 +56,14 @@ func initFriendbot(
 	return &internal.Bot{Minions: minions}, nil
 }
 
-func createMinionAccounts(botAccount internal.Account, botKeypair *keypair.Full, networkPassphrase, newAccountBalance, minionBalance string, numMinions int, baseFee int64, hclient *horizonclient.Client) ([]internal.Minion, error) {
+func createMinionAccounts(botAccount internal.Account, botKeypair *keypair.Full, networkPassphrase, newAccountBalance, minionBalance string, numMinions int, baseFee int64, hclient horizonclient.ClientInterface) ([]internal.Minion, error) {
 	var minions []internal.Minion
 	numRemainingMinions := numMinions
-	minionBatchSize := 100
+	minionBatchSize := 50
```
accordeiro, line 62: Should we make this a config on `friendbot.cfg` so that we can tweak these numbers without having to update the codebase?

services/friendbot/init_friendbot.go
```diff
@@ -56,10 +56,14 @@ func initFriendbot(
...
+	// Allow retries to account for testnet congestion
+	submitTxRetriesAllowed := 5
```
accordeiro, line 64: Same as above, should we make this configurable?

services/friendbot/init_friendbot.go
```diff
@@ -126,14 +130,26 @@ func createMinionAccounts(botAccount internal.Account, botKeypair *keypair.Full,
...
+				// If we hit an error here due to network congestion, try again until we hit max # of retries allowed
+				if e.Problem.Status == http.StatusGatewayTimeout {
+					err = errors.Wrap(errors.Wrap(e, problemString), "submitting create accounts tx")
+					if currentsubmitTxRetry >= submitTxRetriesAllowed {
+						return minions, errors.Wrap(err, fmt.Sprintf("after retrying %d times", currentsubmitTxRetry))
+					}
+					log.Println(err)
+					log.Println("trying again to submit create accounts tx")
+					currentsubmitTxRetry += 1
+					continue
```
accordeiro, line 146: Have you considered adding some kind of exponential backoff here? I'm not sure waiting would make any difference, though – but thought worth mentioning.

accordeiro (approved):
Looks great! Left a few suggestions inline 👇

services/friendbot/init_friendbot.go
acharb: That's a good idea I'll add that 👍 

Just to be sure does the value have to be added to another config file for deployment? Eg. the `friendbot_secret` in `friendbot.cfg` obviously is not the one used in deployment, but I don't see where that's being injected from

services/friendbot/init_friendbot.go
acharb: By "exponential backoff" do you mean adding a delay? I thought about adding a delay in the loop. Though I didn't think it was worth the complexity. My thinking is because a later batch isn't affected by the congestion created by previous batches (afaik), so each batch was in effect independent and later batches wouldn't need a delay

Unless there are other services that similarly have "submit retries" like this one submitting transactions at the same time, but I don't think that's very likely

services/friendbot/init_friendbot.go
acharb: https://github.com/stellar/go/pull/3314/commits/83d5bc0ed4bcf6aefa9a4372baac0ecf62ba6d0c

services/friendbot/init_friendbot_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+	botKeypair := botKP.(*keypair.Full)
+	botAccountID := botKeypair.Address()
+	botAccountMock := horizon.Account{
+		AccountID: botAccountID,
+		Sequence:  "1",
+	}
+	botAccount := internal.Account{AccountID: botAccountID, Sequence: 1}
+
+	horizonClientMock := horizonclient.MockClient{}
```
marcelosalloum, line 30: I ❤️  seeing people using `horizonclient.MockClient{}` 🙌

marcelosalloum (approved):
LGTM!

Aside from the change you're proposing here, could you additionally log the IP of the client performing the http request? I have this theory that some people use Friendbot to fund multiple accounts at once (I do that) because Friendbot doesn't support batch creation and I would like to confirm if there is some truth in this belief. I'm considering this wouldn't take more than a couple lines of code, but if it does you can ignore my comment 😉 .

services/friendbot/init_friendbot.go
accordeiro: Yeah, I think that sounds right. We can add this later if necessary, but I agree with keeping things simple for now.

services/friendbot/init_friendbot_test.go
acharb: I think I'd like to keep this PR independent of new feature adds, but I created a separate ticket for that [here](https://github.com/stellar/go/issues/3319)

`horizonclient.MockClient{}` is great :)

services/friendbot/init_friendbot.go
acharb: spoke with @jacekn [here](https://stellarfoundation.slack.com/archives/CTGJ17JCA/p1610008938307500) that he'll be adding these variables to the secret file

Issue #3110, opened 2020-10-08
Ingestion-as-a-library


2opremio:
We should allow users to create their own core-ingestion pipelines.

High level tasks:

- [x] Make exp/ingest public
- [x] Keep internal/expingest private, but audit it:
   - [x] https://github.com/stellar/go/issues/2851 (`expingest -> ingest` rename issue here - bartek) 
- [ ] Make sure developer experience is good:
   - [x] https://github.com/stellar/go/pull/3318 Rename `SingleLedgerStateReader`
   - [x] https://github.com/stellar/go/pull/3324 Get rid of `io` subpackage and move the code to the top level `ingest` package
   - [x] https://github.com/stellar/go/pull/3330 Get rid of `errors` subpackage and move `StateError` to top level `ingest` package. Also consider removing `ingest.EOF` alias.
   - [x] #3332 Rename `LedgerEntryChangeCache` to something which is more indicative of the operation it's performing (compating / coalescing changes) (@2opremio )
   - [x] ~either make mock implementations private or move them to separate testing package~ (skipped, see [comment below](https://github.com/stellar/go/issues/3110#issuecomment-759665032))
   - [x] #3329 conform implementations to their respective interfaces
   - [x] #3328 move `Range` to its own file
   - [x] #3333 consider moving `StateVerifier` to horizon/internal ingestion package (@2opremio)
   - [x]  https://github.com/stellar/go/pull/3335 Investigate removing `GenesisLedgerStateReader` (@tamirms)
   - [x] #3337 consider moving processor interface to horizon/internal ingestion package (@2opremio)
   - [x] https://github.com/stellar/go/pull/3339 Consider removing `HistoryArchiveBackend`
   - [x] Check documentation
   - [x] #3320 Write examples and a tutorial (piggyback on bartek's ingestion code for the cloudwatch dashboard?) (@Shaptic)
   - [x] ~Look into what the pipeline from the Infra team, dumping info into BigQuery (the library should make their use case possible and simple).~ @tamirms Update https://github.com/stellar/stellar-etl to use the newest `github.com/go/ingest` (after discussion with @nikhilsaraf)

Shaptic:
With regards to the following item:

> either make mock implementations private or move them to separate testing package

This may not be possible. Enumerating the approaches:

 - **making mocked objects private**: This involves hiding `MockChangeProcessor` and `MockChangeReader`. However, doing this results in the `horizon/internal/ingest` package breaking, since it also uses these mocked objects within its tests.
[...]

PR #3320, opened 2021-01-08
ingest: Add a tutorial for the ingestion libary.

Shaptic:
### What
Adds a short tutorial showcasing what you can do with the new, standalone `ingest` package.

It adds a handful of examples:
 - a _very_ basic "hello world"-esque example on processing a single ledger,
 - a more-comprehensive example demonstrating how `LedgerBackend`s and `LedgerTransactionReader`s can interact, demoing how you can read transactions/operations to track statistics about them (based loosely on @bartekn's code).
 - another example highlighting the differences in `ChangeReader` with the above and how they can integrate with history archives, demoing how to track how certain types of changes are made between checkpoint ledgers.

You can view the rendered Markdown directly [here](https://github.com/stellar/go/blob/efe0bf8f09f2b1a35f36a877407998d303879f89/ingest/tutorial/getting-started.md).

### Why
We'd like the ingestion library to be easy to use. See #3110.

### Known limitations
I'd like to add more to the _Snippets_ section, and filling out the _Ingesting a genesis ledger_ section is a TODO.


ingest/tutorial/example_statistics.go
```diff
@@ -0,0 +1,96 @@
...
+      "https://history.stellar.org/prd/core-testnet/core_testnet_001",
+    },
+  }
+)
+
+func main() {
+  lg := log.New()
+  lg.SetLevel(logrus.InfoLevel)
+  lg.Entry.Logger.Out = ioutil.Discard
+  config.Log = lg
```
tamirms, line 30: I think the log field is optional so perhaps it's better to omit the field to simplify the example

ingest/tutorial/example_statistics.go
```diff
@@ -0,0 +1,96 @@
...
+        successfulTransactions++
+        operationsInSuccessful += operationCount
+      } else {
+        failedTransactions++
+        operationsInFailed += operationCount
+      }
+    }
+
+    if delta < ledgersToRead {
+      fmt.Printf("\r")
```
tamirms, line 79: should this be `\n`?

ingest/tutorial/getting-started.md
```diff
@@ -0,0 +1,306 @@
...
+TODO
+
+
+
+# Architecture
+This is described in a little more detail in `doc.go`, its accompanying examples, and but from a high level, the ingestion library is broken down into a few modular components:
+
+
+```txt
+                 [ Processors ]
```
tamirms, line 11: I think you should also cover the `CheckpointChangeReader` (formerly called `SingleLedgerStateReader` https://github.com/stellar/go/pull/3318) and explain how it differs from using a LedgerBackend

Shaptic: Absolutely! This section needs a lot of expansion.

ingest/tutorial/example_statistics.go
Shaptic: Nah, the output in the is "progress bar"-esque, so the output is like:

```
Processed 1/100 ledgers (1%)... \r
Processed 2/100 ledgers (2%)... \r
```

etc. staying on a single line.

ingest/tutorial/example_statistics.go
Shaptic: It's optional but there's a ton of output by default. I think including this in the "full" example has the side benefit of incorporating a snippet (see [Snippets](https://github.com/stellar/go/blob/93667672813207994f1e705013c3fa4916dcd95c/ingest/tutorial/getting-started.md#snippets), which will expand more I hope); it's not discussed in the Markdown tutorial portion. However I do see the merit in keeping it simple. Once there are more examples I think it'll be safe to drop it from this one, so good point!

ingest/tutorial/example_statistics.go
```diff
@@ -0,0 +1,96 @@
...
+		HistoryArchiveURLs: []string{
+			"https://history.stellar.org/prd/core-testnet/core_testnet_001",
+		},
+	}
+)
+
+func main() {
+	lg := log.New()
+	lg.SetLevel(logrus.InfoLevel)
+	lg.Entry.Logger.Out = ioutil.Discard
```
bartekn, line 29: Does it mean the log will be discarded? Can we just pass `nil` log?

ingest/tutorial/example_statistics.go
```diff
@@ -0,0 +1,96 @@
...
+	err = backend.PrepareRange(ledgerRange)
+	panicIf(err)
+	fmt.Println("done.")
+
+	// These are the statistics that we're tracking.
+	var successfulTransactions, failedTransactions int
+	var operationsInSuccessful, operationsInFailed int
+
+	var delta uint32 = 0
+	for ; delta <= ledgersToRead; delta++ {
```
bartekn, line 51: I think it would be more clear that way. Then it's easier to understand what needs to be passed to `NewLedgerTransactionReader`.
```suggestion
	for seq := startingSeq; seq <= startingSeq+ledgersToRead; seq++ {
```

ingest/tutorial/example_statistics.go
```diff
@@ -0,0 +1,96 @@
...
+		fmt.Printf("Processed %d/%d ledgers (%0.1f%%)... ",
+			delta, ledgersToRead, 100*float32(delta)/float32(ledgersToRead))
```
bartekn, line 53: I think it's overcomplicated for an example and distract reader from the essence.
```suggestion
		fmt.Printf("Processing %d... ", seq)
```

ingest/tutorial/example_statistics.go
```diff
@@ -0,0 +1,96 @@
...
+				operationsInSuccessful += operationCount
+			} else {
+				failedTransactions++
+				operationsInFailed += operationCount
+			}
+		}
+
+		if delta < ledgersToRead {
+			fmt.Printf("\r")
+		}
```
bartekn, line 80: Let's focus on the essence.
```suggestion
```

ingest/tutorial/getting-started.md
```diff
@@ -0,0 +1,306 @@
...
+     |        |         |           |          |
+  Captive  Database   History    Remote       etc.
+  Core                Archive    Captive
+                                 Core 
+```
+
+
+
+# Hello, World!
+As is tradition, we'll start with a simplistic example that ingests a single ledger from the network. We're immediately faced with a decision, though: _What's the backend?_ We'll use a Captive Stellar-Core backend in this example because it requires no setup, but there are number of alternatives available. You could also use:
```
bartekn, line 31: > What's the backend?

We don't actually answer the question here.

ingest/tutorial/getting-started.md
```diff
@@ -0,0 +1,306 @@
...
+    fmt.Errorf("The ledger doesn't exist on the backend.\n")
+    return
+  }
+
+  panicIf(err)
+
+  // Now `ledger` is a raw `xdr.LedgerCloseMeta` object containing the  
+  // transactions contained within this ledger.
+  fmt.Printf("\nHello, Protocol %d.\n",
+    ledger.V0.LedgerHeader.Header.LedgerVersion)
```
bartekn, line 75: Maybe we can use a value that we already have a helper for? This one is scary ;).
```suggestion
  fmt.Printf("\nHello, Sequence %d.\n", ledger.LedgerSequence())
```

bartekn (commented):
LGTM! Added some minor comments. As @tamirms mentioned it would be great to have a second example with `CheckpointChangeReader` to ex. print number of accounts, trust lines, etc. It will help users understand the difference between the two.

ingest/tutorial/example_statistics.go
Shaptic: Yeah, just to clean up output, since demo is of `LedgerTransactionReader` more than backends. A `nil` log defaults to stdout by default.

ingest/tutorial/getting-started.md
Shaptic: Great point!

ingest/tutorial/getting-started.md
Shaptic: Don't we?

ingest/tutorial/getting-started.md
Shaptic: Done in latest push!

ingest/doc.go
```diff
@@ -30,8 +31,9 @@ Currently there are three types of readers (all in ingest/io package):
   * LedgerChangeReader reads all changes to ledger entries created as a result
     of transactions (fees and meta) and protocol upgrades in a given ledger.
 
-Warning: Please note that readers stream both successful and failed transactions.
```
tamirms, line 33: looks like there's still a reference to `SingleLedgerStateReader` up above. Can you update it to `CheckpointChangeReader`?

ingest/doc.go
```diff
@@ -48,8 +50,8 @@ There are two types of processors (ingest/io package):
     (io.LedgerTransaction).
 
 For an object to be a processor, it needs to implement a single method:
-ProcessChange or ProcessTransaction. This is a very simple yet powerful interface
-that allows building many kinds of features.
+ProcessChange or ProcessTransaction. This is a very simple yet powerful
```
tamirms, line 53: we have removed Processors from the ingest package so you can remove the entire Processors section of the documentation

ingest/tutorial/example_claimables.go
```diff
@@ -0,0 +1,70 @@
...
+	panicIf(err)
+
+	// First, we need to establish a safe fallback in case of any problems, so
+	// we'll set a 10-second timeout.
+	ctx, canceller := context.WithTimeout(context.Background(), 5*time.Second)
+	defer canceller()
+
+	// We pass 123455 because given a checkpoint frequency of 64 ledgers (the
+	// default in `ConnectOptions`, above), 123455+1 mod 64 == 0. Incompatible
+	// sequence numbers will likely result in 404 errors.
```
tamirms, line 33: I think it will result in an invalid checkpoint sequence error but you can try it out to see what happens

ingest/tutorial/example_claimables.go
```diff
@@ -0,0 +1,70 @@
...
+	entries, newTotal, newCBs := 0, 0, 0
+	for {
+		entry, err := reader.Read()
+		if err == io.EOF {
+			break
+		}
+		panicIf(err)
+
+		entries++
+		isNewEntry := entry.LedgerEntryChangeType() == xdr.LedgerEntryChangeTypeLedgerEntryCreated
```
tamirms, line 46: the change type will always be `xdr.LedgerEntryChangeTypeLedgerEntryCreated` when reading from a checkpoint change reader. A checkpoint change reader will produce the entire state of the stellar network at that particular checkpoint sequence. This is in contrast to the `LedgerChangeReader` which returns changes occurring since the preceding ledger.

Imagine we have a new stellar network. At ledger 1 the root account is created. At ledger 5 the root account creates another account. If you were to create a checkpoint change reader for the first checkpoint ledger, you would get back a series of changes reflecting the stellar network at that ledger. Specifically, you would get a  LedgerEntryChangeTypeLedgerEntryCreated change for every ledger entry that exists at that sequence number. In this example, you would get two account ledger entries (one for the root account and another for the new account created at ledger 5).

If you created a LedgerChangeReader for ledger 2, it would not return any changes because there were no changes produced when applying ledger 2. But if you created a LedgerChangeReader for ledger 5 you would get 2 changes:

one change would be of type xdr.LedgerEntryChangeTypeLedgerEntryCreated which  represents the new account that was created by the root account. 

The other change would be of type xdr.LedgerEntryChangeTypeLedgerEntryUpdated corresponding to the change in the root account (the root account's balance is lower because it created another account, also its sequence number should have incremented).

ingest/tutorial/getting-started.md
```diff
@@ -0,0 +1,393 @@
...
+```
+
+This is described in a little more detail in [`doc.go`](../doc.go), its accompanying examples, the documentation within this package, and the rest of this tutorial.
+
+
+
+# Hello, World!
+As is tradition, we'll start with a simplistic example that ingests a single ledger from the network. We're immediately faced with a decision, though: _What's the backend?_ We'll use a **Captive Stellar-Core backend** in this example because it requires (little-to-)no setup, but there are number of alternatives available. You could also use:
+
+  - a **database** (via `NewDatabaseBackend()`), which would ingest ledgers stored in a database (like Core's, for example);
```
tamirms, line 42: the database backend is specifically coupled to stellar core's db. it cannot ingest ledgers from an arbitrary database because it uses sql queries tailored to stellar core's schema

ingest/tutorial/getting-started.md
```diff
@@ -0,0 +1,393 @@
...
+
+  - a **remote Captive Core** instance (via `NewRemoteCaptive()`), which works much like Captive Core, but points to an instance that isn't (necessarily) running locally;
+
+...or even a completely customized backend that fits your architectural needs, as long as it fulfills the interface requirements of `LedgerBackend`.
```
tamirms, line 46: I think it's best to omit this line because it's not clear how to create a customized backend. and it's not thing we would recommend doing at this point.

ingest/tutorial/getting-started.md
```diff
@@ -0,0 +1,393 @@
...
+
+With that in mind, here's a minimalist example of the ingestion library:
+
+```go
+package main
+
+import (
+  "fmt"
+
+  backends "github.com/stellar/go/ingest/ledgerbackend"
```
tamirms, line 56: I don't think it's necessary to have the `backends` alias

ingest/tutorial/getting-started.md
```diff
@@ -0,0 +1,393 @@
...
+
+Hello, Ledger #123456.
+```
+
+There's obviously much, *much* more we can do with the ingestion library. Let's work through some more comprehensive examples.
+
+
+
+# **Example**: Tracking Ledger Statistics
+In this section, we'll demonstrate how to use backends and processors to actually learn something meaningful about the Stellar network. Again, we'll use a specific backend here (Captive Core, again), but the processing can be done with any of them.
```
tamirms, line 152: there is no use of processors anymore so you can remove the reference to that in this tutorial

ingest/tutorial/getting-started.md
```diff
@@ -0,0 +1,393 @@
...
+			UnsignedRequests:  false,
+		},
+	)
+	panicIf(err)
+```
+
+## Tracking Changes
+Now we can use the history archive to actually read in all of the changes that have accumulated at a particular checkpoint. 
+
+TODO: Is this the changes between one checkpoint and the previous checkpoint? Or is it the changes since the checkpoint and the genesis ledger? Gut says former but not sure.
```
tamirms, line 294: it is the changes since the network was created

ingest/tutorial/getting-started.md
```diff
@@ -0,0 +1,393 @@
...
+```
+
+
+
+# Snippets
+This section outlines a brief collection of common things you may want to do with the library. We assume a very generic `backend` variable where necessary that is one of the aforementioned `LedgerBackend` instances to avoid boilerplate.
+
+
+### Disabling output logs
+Certain backends (like Captive Core) can be very noisy. The backends will log to standard output by default; you can disable it by redirecting to `ioutil.Discard`, instead:
```
tamirms, line 351: instead of discarding all the log output, I would recommend setting the log level to warn which should reduce the log volume significantly (if there are no warnings / errors then it will not even produce any logs)

ingest/tutorial/getting-started.md
bartekn: Long answer: you can actually get both: changes since the network was created and changes between any two checkpoints. The latter is rather complicated but possible. The data structure used in buckets is explained here: https://github.com/stellar/stellar-core/blob/master/src/bucket/BucketList.h

ingest/tutorial/getting-started.md
```diff
@@ -0,0 +1,393 @@
...
+## Tracking Changes
+Now we can use the history archive to actually read in all of the changes that have accumulated at a particular checkpoint. 
+
+TODO: Is this the changes between one checkpoint and the previous checkpoint? Or is it the changes since the checkpoint and the genesis ledger? Gut says former but not sure.
+
+```go
+	// First, we need to establish a safe fallback in case of any problems, so 
+	// we'll set a 10-second timeout.
+	ctx, canceller := context.WithTimeout(context.Background(), 10*time.Second)
+	defer canceller()
```
bartekn, line 300:
```suggestion
	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
	defer cancel()
```

ingest/tutorial/getting-started.md
```diff
@@ -0,0 +1,393 @@
...
+
+	reader, err := ingest.NewCheckpointChangeReader(ctx, historyArchive, 123455)
+	panicIf(err)
+```
+
+Notice the importance of the sequence number: 123456 is a checkpoint ledger and thus exists in the history archive. This is based on the default checkpoint frequency of 64 ledgers (see `ingest.ConnectOptions`): `123456 mod 64 == 0`.  (Incompatible sequence numbers will likely result in 404 errors.)
```
bartekn, line 306: 123456 is not a checkpoint ledger. See my other comment above.

ingest/tutorial/example_claimables.go
Shaptic: Awesome explanation, thank you so much! With this in mind, I think that means there's a lot of value in creating (yet another) example demonstrating how `LedgerChangeReader` is used, to differentiate this exact distinction.

ingest/tutorial/example_claimables.go
Shaptic: Hmm, nope, see #3346?

ingest/doc.go
Shaptic: Done!

ingest/doc.go
Shaptic: :heavy_check_mark:

ingest/tutorial/getting-started.md
Shaptic: good call, clarified!

ingest/tutorial/getting-started.md
Shaptic: cool, done!

ingest/tutorial/getting-started.md
Shaptic: Thanks for the clarity! Might be value in doing the latter as a demo later.

ingest/checkpoint_change_reader.go
```diff
@@ -71,6 +71,11 @@ const (
 )
 
 // NewCheckpointChangeReader constructs a new CheckpointChangeReader instance.
+//
+// The ledger sequence must precede a checkpoint ledger. By default (see
```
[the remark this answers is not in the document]
Shaptic: Ah yes, subtle. Clarified!

ingest/tutorial/example_claimables.go
```diff
@@ -0,0 +1,64 @@
...
+			NetworkPassphrase: config.NetworkPassphrase,
+			S3Region:          "eu-west-1",
+			UnsignedRequests:  false,
+		},
+	)
+	panicIf(err)
+
+	// First, we need to establish a safe fallback in case of any problems
+	// during history archive processing, so we'll set a 30-second timeout.
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
```
tamirms, line 28: I would prefer to use `context.Background()` because the history archives for pubnet are huge. It takes horizon around 15-20 minutes to process the pubnet history archives.

ingest/doc.go
```diff
@@ -1,55 +1,37 @@
 /*
 
-Package ingest provides primitives for building custom ingestion engines.
+The "ingest" package provides primitives for building custom ingestion engines.
```
bartekn, line 3: I think it's a convention to start  `doc.go` with `Package x`. At least `go-lint` throws warning.

ingest/tutorial/example_hello.go
```diff
@@ -0,0 +1,28 @@
...
+	backends "github.com/stellar/go/ingest/ledgerbackend"
+)
+
+func helloworld() {
+	backend, err := backends.NewCaptive(config)
+	panicIf(err)
+	defer backend.Close()
+
+	// Prepare a single ledger to be ingested,
+	backend.PrepareRange(backends.BoundedRange(123456, 123456))
```
bartekn, line 15:
```suggestion
	err = backend.PrepareRange(backends.BoundedRange(123456, 123456))
	panicIf(err)
```

ingest/tutorial/getting-started.md
```diff
@@ -0,0 +1,387 @@
...
+  "github.com/stellar/go/ingest/ledgerbackend"
+)
+
+func main() {
+  backend, err := ledgerbackend.NewCaptive(config)
+  panicIf(err)
+  defer backend.Close()
+
+  // Prepare a single ledger to be ingested,
+  backend.PrepareRange(ledgerbackend.BoundedRange(123456, 123456))
```
bartekn, line 63: Handle `err`.

ingest/tutorial/example_common.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+var (
+	config = ledgerbackend.CaptiveCoreConfig{
+		// Change these based on your environment:
+		BinaryPath:        "/usr/local/bin/stellar-core",
+		ConfigAppendPath:  "stellar-core-stub.toml",
+		NetworkPassphrase: "Test SDF Network ; September 2015",
+		HistoryArchiveURLs: []string{
+			"https://history.stellar.org/prd/core-testnet/core_testnet_001",
```
bartekn, line 16: While we support only one archive now I think it's worth using a config that will be used in the future.
```suggestion
			"https://history.stellar.org/prd/core-testnet/core_testnet_001",
			"https://history.stellar.org/prd/core-testnet/core_testnet_002",
			"https://history.stellar.org/prd/core-testnet/core_testnet_003",
```

ingest/tutorial/getting-started.md
```diff
@@ -0,0 +1,387 @@
...
+## History Archive Connections
+First thing's first: we need to establish a connection to a history archive.
+
+```go
+	// Open a history archive using our existing configuration details.
+	historyArchive, err := historyarchive.Connect(
+		config.HistoryArchiveURLs[0],   // assumes a CaptiveCoreConfig
+		historyarchive.ConnectOptions{
+			NetworkPassphrase: config.NetworkPassphrase,
+			UnsignedRequests:  false,
```
bartekn, line 278: `false` is default for `bool` so can be omitted in the example.
```suggestion
```

ingest/tutorial/getting-started.md
```diff
@@ -0,0 +1,387 @@
...
+	// First, we need to establish a safe fallback in case of any problems
+	// during history archive processing, so we'll set a 30-second timeout.
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+
+	reader, err := ingest.NewCheckpointChangeReader(ctx, historyArchive, 123455)
+	panicIf(err)
+```
+
+By default, checkpoints occur every 64 ledgers (see `historyarchive.ConnectOptions` for changing this). More specifically, given ledger `n`, if `n+1 mod 64 == 0`, then `n` is a checkpoint ledger. This is true above for `n == 123455`.
```
bartekn, line 299: Nit: Alternatively, `n*64 - 1 where n = 1, 2, 3, ...` can be easier to understand.

bartekn (approved):
LGTM! Please take a look at the comments.

ingest/tutorial/example_claimables.go
Shaptic: Added a note regarding this!

PR #3324, opened 2021-01-09
ingest: Remove io package and move contents to top level ingest package

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Move the contents of the `ingest/io` package to the top level `ingest` package. Naming a package `io` is not ideal because it clashes with the `io` package of the golang standard library.

Ingestion implies that there will be some sort of i/o operations because we will be taking data from an external source, transforming it, and then transferring it to an internal source. So, I think there is no loss of meaning if we get rid of `ingest/io` and instead place the contents in the `ingest` package.
 

### Why

This PR is part of #3110 which is aimed at making the ingestion library ready for public use.

### Known limitations

[N/A]


ingest/errors.go
```diff
@@ -1,4 +1,4 @@
-package io
```
bartekn, line 1: We have `ingest/errors` package. Maybe we should move it there?

bartekn (approved):

ingest/errors.go
tamirms: I actually wanted to move `ingest/errors` into `ingest` (will do in another PR) because `errors` clashes with the golang standard library `errors` package

PR #3332, opened 2021-01-11
services/horizon: Rename LedgerEntryChangeCache to ChangeCompactor

2opremio:
### What

Rename `LedgerEntryChangeCache` to `ChangeCompactor`

### Why

The name is more descriptive. It's part of #3110 

### Known limitations

It may break 3rd party code using this, but it's highly unlikely.


tamirms (approved):

bartekn:
LGTM but can we remove `Cache` from the name. I think it's confusing and I've been planning to remove it but there was always something else ;).

2opremio:
How about ChangeCompactor?

PR #3333, opened 2021-01-11
services/horizon: Move ingest/verify to services/horizon/internal/verify

2opremio:
### What

Move package `ingest/verify` to `services/horizon/internal/verify`.

### Why

It's a a usecase which we (at least @tamirms and I) feel like it will be confusing for the end user.

If there is a demand for it, we can always move it back.

It's part of #3110


bartekn:
I think it can be valuable for `ingest` users who update internal state using `LedgerEntryChanges`. If we move it to horizon internal package nobody will ever find it. I'm OK with merging it but please reconsider.

2opremio:
I discussed it with Tamir and we thought that the abstraction reflected a specific use case which is hard to understand (and probably not common enough to justify exposing in the public ingestion API).

I am happy to be proved wrong by end users though.

tamirms (approved):

PR #3337, opened 2021-01-12
Move ingest/ChangeProcessor interface to the horizon/internal/ingest

2opremio:
### What

Move the `ChangeProcessor` interface to the internal ingestion package

### Why

Due to #3110 , I had a conversation with Tamir in which we saw the processor as a glorified for loop.

### Known limitations

The `StatsChangeProcessor` and `StatsLedgerTransactionProcessor` stop making sense after the `ChangeProcessor` is moved but shouldn't be simply moved because the `dump-ledger-state` utility uses them. (Thus, I will leave this PR as a draft).


services/horizon/internal/ingest/processors/change_processors.go
```diff
@@ -1,22 +1,23 @@
...
+package processors
 
 import (
 	"io"
 
+	"github.com/stellar/go/ingest"
 	"github.com/stellar/go/support/errors"
 )
 
 type ChangeProcessor interface {
```
tamirms, line 10: I think it would be better to move these interfaces into services/horizon/internal/ingest/processor_runner.go and merge them with the horizonChangeProcessor and horizonTransactionProcessor interfaces

2opremio: We can't (for `StatsChangeProcessor` at least) since it's used publicly

tamirms (approved):

PR #3346, opened 2021-01-15
ingest: Introduce a new error for invalid checkpoint sequences

Shaptic:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Introduces a new error for when users pass a sequence number that doesn't correspond to a checkpoint ledger.

### Why
This may end up being a common issue for consumers of the ingestion library since checkpoint frequency is a subtle point mostly documented within the code itself. Right now, when passing a non-checkpoint sequence number to `ingest.NewCheckpointChangeReader`, the following error occurs:

```
unable to get checkpoint HAS at ledger sequence 123450: Bad HTTP response '404 Not Found' for GET 'https://history.stellar.org/prd/core-testnet/core_testnet_001/history/00/01/e2/history-0001e23a.json'
```

This is pretty vague. The new error would be:

```
123456 is not a checkpoint ledger, try 123455 or 123519 (in general, try n where n+1 mod 64 == 0)
```

### Known limitations
n/a

### DIFF
--- a/ingest/checkpoint_change_reader.go
+++ b/ingest/checkpoint_change_reader.go
@@ -81,8 +81,14 @@ func NewCheckpointChangeReader(
 	archive historyarchive.ArchiveInterface,
 	sequence uint32,
 ) (*CheckpointChangeReader, error) {
-	if !archive.GetCheckpointManager().IsCheckpoint(sequence) {
-		return nil, NewCheckpointError(sequence, archive.GetCheckpointManager())
+	manager := archive.GetCheckpointManager()
+
+	// The nth ledger is a checkpoint ledger iff: n+1 mod f == 0, where f is the
+	// checkpoint frequency (64 by default).
+	if !manager.IsCheckpoint(sequence) {
+		return nil, errors.Errorf("%d is not a checkpoint ledger, try %d or %d",
+			sequence, manager.PrevCheckpoint(sequence),
+			manager.NextCheckpoint(sequence))
 	}
 
 	has, err := archive.GetCheckpointHAS(sequence)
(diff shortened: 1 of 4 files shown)


ingest/errors.go
```diff
@@ -1,12 +1,37 @@
...
+var ErrNotCheckpoint = errors.New("sequence number is not a checkpoint ledger")
+
+type CheckpointError struct {
+	error
+	precedingCheckpoint uint32
+	followingCheckpoint uint32
+}
+
+func NewCheckpointError(sequence uint32, checkpointMgr historyarchive.CheckpointManager) CheckpointError {
+	return CheckpointError{
```
tamirms, line 28: I think we can just return ErrNotCheckpoint here because the caller of `NewCheckpointChangeReader` will not know the specific type of error returned unless they know in advance to check for the CheckpointError type.

Shaptic: Alright, seems good. I'd really like to make it easy for a developer to fix the error, though. Do you think there's merit to a better error message, such as the one I've added in a024ce0cc364fa2b8c810fea5e50accd1badf05b?

tamirms (approved):

ingest/errors.go
tamirms: yeah, I think that looks good!

ingest/checkpoint_change_reader.go
```diff
@@ -76,6 +76,16 @@ func NewCheckpointChangeReader(
 	archive historyarchive.ArchiveInterface,
 	sequence uint32,
 ) (*CheckpointChangeReader, error) {
+	manager := archive.GetCheckpointManager()
+
+	// The nth ledger is a checkpoint ledger iff: n+1 mod f == 0, where f is the
+	// checkpoint frequency (64 by default).
+	if !manager.IsCheckpoint(sequence) {
+		return nil, errors.Errorf("%d is not a checkpoint ledger, try %d or %d",
```
bartekn, line 84: Maybe we should add `n+1 mod f == 0` to the message? It will make it clear for users who don't know what are these values.

bartekn (approved):

ingest/checkpoint_change_reader.go
Shaptic: Done in 9177540aa7a4d4c039cecadb24bc660f8032bec8; thoughts?

PR #3288, opened 2020-12-17
all: update scripts to test all code with go 1.16beta1

leighmcculloch:
<details>
  <summary>PR Checklist</summary>
  
### PR Structure

* [x] This PR has reasonably narrow scope (if not, break it down into smaller PRs).
* [x] This PR avoids mixing refactoring changes with feature changes (split into two PRs
  otherwise).
* [x] This PR's title starts with name of package that is most changed in the PR, ex.
  `services/friendbot`, or `all` or `doc` if the changes are broad or impact many
  packages.

### Thoroughness

* [x] This PR adds tests for the most critical parts of the new functionality or fixes.
* [x] I've updated any docs ([developer docs](https://www.stellar.org/developers/reference/), `.md`
  files, etc... affected by this change). Take a look in the `docs` folder for a given service,
  like [this one](https://github.com/stellar/go/tree/master/services/horizon/internal/docs).

### Release planning

* [x] I've updated the relevant CHANGELOG ([here](services/horizon/CHANGELOG.md) for Horizon) if
  needed with deprecations, added features, breaking changes, and DB schema changes.
* [x] I've decided if this PR requires a new major/minor version according to
  [semver](https://semver.org/), or if it's mainly a patch change. The PR is targeted at the next
  release branch if it's not a patch change.
</details>

### What

Update builds to test all code with Go 1.16beta1.

### Why

To prepare for when Go 1.16 is released.

### Merging

This change is intended to be merged to `master` after #3347 and contains changes from that PR.

### Known limitations

This PR will not utilize any new functionality or features of Go 1.16 because we will support Go 1.15 until Go 1.17 is released.

This PR will not deprecate our use of Go 1.13 until a stable non-beta version of Go 1.16.

The builds will test using Debian Buster base images because Stretch images are not available in the official Docker Hub Golang images since Go 1.15.

PR #3347, opened 2021-01-15
all: replace all circleci images with official images

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Replace the CircleCI convenience images `circleci:golang` and `circleci:buildpack-deps` with the official images `golang` and `buildpack-deps` available on DockerHub.

### Why

The CircleCI golang and buildpack-deps convenience images are based on the official images on DockerHub and add things like NodeJS, and other useful tools like curl, zip, etc, and most notably setup a non-root user to run the tests.

The convenience images we are using are the current generation images, but in 2020 CircleCI [released](https://circleci.com/blog/announcing-our-next-generation-convenience-images-smaller-faster-more-deterministic/) next generation images and so at some point we will have to update the images we're using and the new images are not identical.

The convenience images also come with constraints:
- CircleCI doesn't release images for every upstream tag, and so we don't get Go beta releases on the current generation, and we don't get release-candidates on the new generation images. I hit this limitation every Go release when I run our tests against the beta releases to see if we find any issues. e.g. https://github.com/stellar/go/pull/3288. There is some discussion about that at https://github.com/circleci/circleci-images/issues/531#issuecomment-759888822.
- The images are uniquely designed for CircleCI so they make it harder for us to experiment with other CI products.

We rely on very little in the convenience images. From the testing I've done the only thing we are dependent on is the install of the `zip` binary, which is very easy for us to install as part of our build.

Because the convenience images are built on the upstream official images that this change switches to, there is no change in operating system or operating system version, and there should be no changes to artifacts built.

### Known limitations

N/A


.circleci/config.yml
```diff
@@ -183,7 +183,7 @@ commands:
     steps:
       - run:
           name: Install dockerize
-          command: wget https://github.com/jwilder/dockerize/releases/download/$DOCKERIZE_VERSION/dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz && sudo tar -C /usr/local/bin -xzvf dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz && rm dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz
+          command: wget https://github.com/jwilder/dockerize/releases/download/$DOCKERIZE_VERSION/dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz && tar -C /usr/local/bin -xzvf dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz && rm dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz
```
marcelosalloum, line 186: ❓  Why add `sudo` here?

marcelosalloum (approved):
LGTM! I just have a question about `sudo` ⬇️ :

.circleci/config.yml
leighmcculloch: The `sudo` is being removed in this change. It existed here because the CircleCI convenience images setup a non-root user within the image that has the ability to run commands with root permissions using sudo. Sudo is required to write to the `/usr/local/bin` path. It isn't required once the change to the official images is made because the official images contain no additional users and so everything runs as root.

marcelosalloum: Cool, thanks for explaining.

bartekn (approved):

Issue #3348, opened 2021-01-16
Add support for Asset Clawbacks (CAP 35)


2opremio:
https://github.com/stellar/stellar-protocol/blob/master/core/cap-0035.md

High priority tasks:
- [x] #3349 Update XDR definitions
- [x] #3350 Add support for `ClawbackOp`
- [x] #3351 Add support for account flag `AUTH_CLAWBACK_ENABLED_FLAG`
- [x] #3351 Add support for asset flag `CLAWBACK_ENABLED_FLAG`
- [x] Add support for the above in https://github.com/elucidsoft/dotnet-stellar-sdk ( @leighmcculloch ). WIP at https://github.com/stellar/dotnet-stellar-sdk/tree/cap35
- [x] #3368 Add integration test for the `Clawback` operation (needs a Core build with CAP35 support)
        - Replace the temporary stellar-core version and docker image by the definitive one once ready

Lower priority tasks:
- [x] #3378 Add support for the`SetTrustlineFlags` operation
       - It requires updating the XDR definitions with the changes at https://github.com/stellar/stellar-protocol/pull/838
- [x] https://github.com/stellar/go/issues/3448 Update XDR with the changes from https://github.com/stellar/stellar-protocol/pull/878#event-4409243096
- [x] #3356 Add support for `ClaimableBalanceEntryExtensionV1`
- [x] #3357 Add support for new error `SET_OPTIONS_AUTH_REVOCABLE_REQUIRED` in `txsub`
- [x] https://github.com/stellar/go/pull/3376 Add result codes for new operations in `txsub`
- [x] #3350 Add ingestion support for `ClawbackClaimableBalanceOp`
- [x] #3358 Add effects for `ClawbackOp` and `ClawbackClaimableBalanceOp`
- [x] Add `AccountCredited` effect for `ClawbackClaimableBalanceOp` ( leftover from https://github.com/stellar/go/pull/3358#discussion_r561784015 until we understand Core's behavior)
- [x]  #3379 Make sure there is support for all the above in the Go SDK
       - Add txnbuild support for all the new operations
       - Improve testing for the clawback operation https://github.com/stellar/go/pull/3364#discussion_r562759282
- [ ]  https://github.com/stellar/go/issues/3447 Make sure there are integration tests for all the operations (Already done some of it at #3422 #3424)
- [ ] Update documentation at https://github.com/stellar/new-docs (@Shaptic, #3449)
         - Add new operations
         - Add new effects
[...]

PR #3349, opened 2021-01-16
[CAP 35] Update XDR defintions and xdr package for CAP35

2opremio:
Using the patch from https://github.com/stellar/stellar-protocol/blob/master/core/cap-0035.md#xdr-changes

Part of #3348


xdr/asset.go
```diff
@@ -45,8 +45,8 @@ func MustNewCreditAsset(code string, issuer string) Asset {
 }
 
 // NewAllowTrustAsset returns a new allow trust asset, panicking if it can't.
-func NewAllowTrustAsset(code string) (AllowTrustOpAsset, error) {
-	a := AllowTrustOpAsset{}
+func NewAllowTrustAsset(code string) (AssetCode, error) {
```
leighmcculloch, line 48: I think this should be renamed to `NewAssetCode` to continue to match the name of the type it is creating? We can keep the old function too for backwards compatibility if that is useful.

xdr/asset.go
```diff
@@ -67,7 +67,7 @@ func NewAllowTrustAsset(code string) (AllowTrustOpAsset, error) {
 }
 
 // MustNewAllowTrustAsset returns a new allow trust asset, panicking if it can't.
-func MustNewAllowTrustAsset(code string) AllowTrustOpAsset {
+func MustNewAllowTrustAsset(code string) AssetCode {
```
leighmcculloch, line 70: I think this should be renamed to `MustNewAssetCode` to continue to match the name of the type it is creating?We can keep the old function too for backwards compatibility if that is useful.

xdr/asset.go
```diff
@@ -222,22 +222,22 @@ func (a *Asset) SetNative() error {
 
 // ToAllowTrustOpAsset for Asset converts the Asset to a corresponding XDR
 // "allow trust" asset, used by the XDR allow trust operation.
-func (a *Asset) ToAllowTrustOpAsset(code string) (AllowTrustOpAsset, error) {
+func (a *Asset) ToAllowTrustOpAsset(code string) (AssetCode, error) {
 	length := len(code)
```
leighmcculloch, line 226: I think this should be renamed to `ToAssetCode` to continue to match the name of the type it is creating?We can keep the old function too for backwards compatibility if that is useful.

xdr/xdr_generated.go
```diff
@@ -8656,159 +8994,20 @@ var (
...
-//        case ASSET_TYPE_CREDIT_ALPHANUM4:
-//            AssetCode4 assetCode4;
-//
-//        case ASSET_TYPE_CREDIT_ALPHANUM12:
-//            AssetCode12 assetCode12;
-//
-//            // add other asset types here in the future
-//        }
-//
-type AllowTrustOpAsset struct {
```
leighmcculloch, line 8673: We can use Go type aliases to define that AllowTrustOpAsset is the same as AssetCode for backwards compatibility, if that is useful. If that is not useful then I wouldn't worry about it.

leighmcculloch (approved):
🎉 Looks great.

I left some suggestions (💡) as comments inline, but I defer to you for all of them to you for whether they're the right call or not as I'm not sure if backwards compatibility is important for this change, and not sure what other plans you might have.

FYI the XDR is likely to undergo at least one more change that isn't yet captured in the diff in the CAP-35 document. That change is discussed on the mailing list as item 4 in this message: https://groups.google.com/g/stellar-dev/c/hPhkXhrl5-Y/m/Hi9XaWhkCwAJ.

xdr/asset.go
2opremio: Makes sense, will do

xdr/asset.go
2opremio: good point

xdr/xdr_generated.go
2opremio: Also a good point. I am not sure it's worth maintaining backwards compatibility here. @stellar/horizon-committers thoughts?

leighmcculloch: I'm not convinced it's worth it either. I think I've seen us do major version releases for new protocols before?

xdr/asset.go
2opremio: I had to use `NewAssetCodeFromString` because xdrgen already generated `NewAssetCode`

xdr/asset.go
2opremio: I used `MustNewAssetCodeFromString` for consistency with the change above

xdr/asset.go
2opremio: Done

xdr/xdr_generated.go
2opremio: I think I will merge now. @stellar/horizon-committers please let me know if you think I should change this in terms of backwards compatibility.

I will also add a changelog entry indicating the changes.

PR #3352, opened 2021-01-17
Clarify that MinBaseFee is in stroops

billallen256:
### What

Clarifies that MinBaseFee is in stroops.  No functional change.

### Why

The units of this constant was not documented.

### Known limitations

N/A


leighmcculloch (approved):
Thanks @gershwinlabs for submitting this improvement. 💯

PR #3356, opened 2021-01-19
[CAP 35] Add support for new Claimable Balance clawback flag

2opremio:
* ALTER the claimable_balances SQL table to add a flags column
* Add ingestion support for for the new flag (including verification)
* Add support to the Horizon Go client
* Extend Claimable balance effects to include the flag

Part of #3348

### DIFF
--- a/protocols/horizon/main.go
+++ b/protocols/horizon/main.go
@@ -660,7 +660,7 @@ type PathsPage struct {
 
 // ClaimableBalanceFlags represents the state of a claimable balance's flags
 type ClaimableBalanceFlags struct {
-	ClaimableBalanceClawbackEnabled bool `json:"claimable_balance_clawback_enabled"`
+	ClawbackEnabled bool `json:"clawback_enabled"`
 }
 
 // ClaimableBalance represents a claimable balance
--- a/protocols/horizon/main.go
+++ b/protocols/horizon/main.go
@@ -658,20 +658,26 @@ type PathsPage struct {
 	} `json:"_embedded"`
 }
 
+// ClaimableBalanceFlags represents the state of a claimable balance's flags
+type ClaimableBalanceFlags struct {
+	ClaimableBalanceClawbackEnabled bool `json:"claimable_balance_clawback_enabled"`
+}
+
 // ClaimableBalance represents a claimable balance
 type ClaimableBalance struct {
 	Links struct {
 		Self hal.Link `json:"self"`
 	} `json:"_links"`
 
-	BalanceID          string     `json:"id"`
-	Asset              string     `json:"asset"`
-	Amount             string     `json:"amount"`
-	Sponsor            string     `json:"sponsor,omitempty"`
-	LastModifiedLedger uint32     `json:"last_modified_ledger"`
-	LastModifiedTime   *time.Time `json:"last_modified_time"`
-	Claimants          []Claimant `json:"claimants"`
-	PT                 string     `json:"paging_token"`
+	BalanceID          string                `json:"id"`
+	Asset              string                `json:"asset"`
+	Amount             string                `json:"amount"`
+	Sponsor            string                `json:"sponsor,omitempty"`
+	LastModifiedLedger uint32                `json:"last_modified_ledger"`
+	LastModifiedTime   *time.Time            `json:"last_modified_time"`
+	Claimants          []Claimant            `json:"claimants"`
+	Flags              ClaimableBalanceFlags `json:"flags"`
+	PT                 string                `json:"paging_token"`
 }
 
 type ClaimableBalances struct {
--- a/services/horizon/internal/ingest/processors/effects_processor.go
+++ b/services/horizon/internal/ingest/processors/effects_processor.go
@@ -788,15 +788,14 @@ func (e *effectsWrapper) addBumpSequenceEffects() error {
 			details := map[string]interface{}{"new_seq": afterAccount.SeqNum}
 			e.add(source.Address(), history.EffectSequenceBumped, details)
 		}
-
 		break
 	}
 
 	return nil
 }
 
-func setClaimableBalanceFlagDetails(details map[string]interface{}, flags xdr.Uint32) {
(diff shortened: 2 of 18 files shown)


2opremio:
Question: should I add the flag as a detail of the `ClaimableBalance`/`ClaimClaimableBalance` operations? 

I decided not to since it's simply a sideeffect of the clawback flag from account used to create the claimable balance. However, I could add it if @stellar/horizon-committers think it makes sense.

services/horizon/internal/ingest/verify.go
```diff
@@ -577,11 +577,19 @@ func addClaimableBalanceToStateVerifier(
...
-			Amount:    xdr.Int64(row.Amount),
+			Amount:    row.Amount,
+		}
+		if row.Flags != 0 {
+			cBalance.Ext = xdr.ClaimableBalanceEntryExt{
+				V: 1,
+				V1: &xdr.ClaimableBalanceEntryExtensionV1{
+					Flags: xdr.Uint32(row.Flags),
+				},
+			}
```
bartekn, line 592: This will fail if history archive entry is at V1 and DB entry Flags is 0. We should transform claimable balance in `transformEntry` (it contains some examples already so should be easy to add a new type).

bartekn (approved):
LGTM added one minor, one major comments. Plus what @leighmcculloch said about adding flag methods to `xdr` package.

protocols/horizon/main.go
```diff
@@ -658,20 +658,26 @@ type PathsPage struct {
 	} `json:"_embedded"`
 }
 
+// ClaimableBalanceFlags represents the state of a claimable balance's flags
+type ClaimableBalanceFlags struct {
+	ClaimableBalanceClawbackEnabled bool `json:"claimable_balance_clawback_enabled"`
```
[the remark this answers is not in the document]
2opremio: I was trying to be consistent with the `Auth` flags, but sure!

services/horizon/internal/ingest/verify.go
2opremio: Sure thing, will do

xdr/claimable_balance_entry.go
```diff
@@ -0,0 +1,9 @@
+package xdr
+
+func (entry *ClaimableBalanceEntry) Flags() Uint32 {
```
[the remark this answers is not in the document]
2opremio: Sure

services/horizon/internal/resourceadapter/claimable_balances.go
```diff
@@ -41,6 +41,10 @@ func PopulateClaimableBalance(
 		dest.LastModifiedTime = &ledger.ClosedAt
 	}
 
+	if xdr.ClaimableBalanceFlags(claimableBalance.Flags)&xdr.ClaimableBalanceFlagsClaimableBalanceClawbackEnabledFlag != 0 {
```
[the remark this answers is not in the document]
2opremio: I will implement option 2

services/horizon/internal/ingest/processors/effects_processor.go
```diff
@@ -868,6 +878,7 @@ func (e *effectsWrapper) addClaimClaimableBalanceEffects() error {
 			if err != nil {
 				return fmt.Errorf("Invalid balanceId in meta changes for op: %d", e.operation.index)
 			}
+			flags = cBalance.Flags()
```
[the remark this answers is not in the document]
2opremio: You are totally right, good catch

services/horizon/internal/ingest/verify.go
2opremio: I think I've addressed it correctly but please take a look. What would be the best way to test this?

2opremio:
PTAL, I think everything is addressed.

I would like to add a test for the state verify problem @bartekn pointed out but I am not sure what's the best way forward with that.

services/horizon/internal/resourceadapter/claimable_balances.go
```diff
@@ -41,6 +41,10 @@ func PopulateClaimableBalance(
 		dest.LastModifiedTime = &ledger.ClosedAt
 	}
 
+	if xdr.ClaimableBalanceFlags(claimableBalance.Flags).IsClawbackEnabled() {
```
tamirms, line 44: `dest.Flags.ClawbackEnabled = xdr.ClaimableBalanceFlags(claimableBalance.Flags).IsClawbackEnabled()`

tamirms:
> I would like to add a test for the state verify problem @bartekn pointed out but I am not sure what's the best way forward with that.

I think you can add a unit test on the `transformEntry()` function which calls `transformEntry()` on a v0 Claimable Balance and asserts the result is equal to calling `transformEntry()` on the equivalent v1 Claimable Balance.

services/horizon/internal/resourceadapter/claimable_balances.go
2opremio: I am not sure it makes a difference but I will change it

tamirms: yeah, it was a style nitpick. they are both equivalent

2opremio:
OK, I think everything is addressed. PTAL

services/horizon/internal/ingest/verify.go
2opremio: I added a test as suggested by @tamirms

tamirms (approved):

services/horizon/internal/ingest/verify.go
bartekn: Sorry for after-merge comment. I think `NormalizeClaimableBalanceExtension` is a good idea but for consistency I'd do that in `transformEntry` for now.  Then, in another PR we can create `NormalizeEntryExtension` for all `LedgerEntry`ies and remove `transformEntry`. I can do it if you and @tamirms agree.

2opremio: seems like a good idea

Issue #3069, opened 2020-09-29
Remove duplication of effects code


2opremio:
Effects are parsed and defined in multiple places.

Effect types and type names are identically defined at`history/main.go` and `effects/main.go`

They are parsed almost identically at `effects/main.go.UnmarshallEffect()`, `resourceadapter/effects.go.NewEffect()`

This makes effects hard to maintain and error-prone. For instance, see #3068 which was found by @abuiles after yet another problem with effects at #3060

2opremio:
In fact, I think much of the effects code could be autogenerated.

PR #3358, opened 2021-01-19
[CAP 35] Add effects for Clawback{ClaimableBalance} operations

2opremio:
New effects are:
* ~`AssetClawedBack` (as a result of a Clawback operation)~ (the `Account{Debited,Credited}` effects have been reused`)
* `ClaimableBalanceClawedBack` (as a result of a ClawbackClaimableBalance operation)

Also, rename the `claimable_balance_id` of the `ClawbackClaimableBalance` operation to simply `balance_id`, for consistency with its XDR field name.

Part of #3348


protocols/horizon/effects/main.go
```diff
@@ -222,6 +229,8 @@ var EffectTypeNames = map[EffectType]string{
 	EffectSignerSponsorshipCreated:                 "signer_sponsorship_created",
 	EffectSignerSponsorshipUpdated:                 "signer_sponsorship_updated",
 	EffectSignerSponsorshipRemoved:                 "signer_sponsorship_removed",
+	EffectAssetClawedBack:                          "asset_clawed_back",
```
leighmcculloch, line 232: ❔ Does it make sense for this to be `trustline_clawed_back` given that the other effects appear to specify the type of object that was operated on in their name, and the asset is not being operated on, the balance of an accounts trustline has been changed.

I don't feel strongly, and definitely defer to you Horizon folks about this, something just feels inconsistent with the name starting with `asset_`.

protocols/horizon/operations/main.go
```diff
@@ -278,7 +278,7 @@ type Clawback struct {
 // Clawback.
 type ClawbackClaimableBalance struct {
 	Base
-	ClaimableBalanceID *string `json:"claimable_balance_id,omitempty"`
+	ClaimableBalanceID *string `json:"balance_id,omitempty"`
```
leighmcculloch, line 281: ❗ This and the effects location where the id is referenced should be `claimable_balance_id` because using an incomplete name for the type can be confusing. The formal name is claimable balance.

leighmcculloch (approved):
🎉 One question (❔) ~and one naming concern (❗)~.

protocols/horizon/operations/main.go
leighmcculloch: Oh, I see we already are using the abbreviated balance_id term throughout the API. Doh. Please ignore my comment.

protocols/horizon/effects/main.go
2opremio: I considered it, but is the Asset which is clawed back, not the full trustline.

2opremio: @stellar/horizon-committers would you agree?

tamirms (approved):

protocols/horizon/effects/main.go
leighmcculloch: What about `clawed_back`? They are both 1-1 with the ops then in past tense. I agree `trustline_clawed_back` is also awkward.

leighmcculloch: What about `clawed_back`? Then they are both 1-1 with the ops then in past tense. I agree `trustline_clawed_back` is also awkward.

protocols/horizon/effects/main.go
```diff
@@ -222,6 +229,8 @@ var EffectTypeNames = map[EffectType]string{
...
+	EffectClaimableBalanceClawedBack:               "claimable_balance_clawed_back",
```
leighmcculloch, line 233: Are these two new effects on the issuer account? or are they on the account and claimable balance that is losing the funds or being destroyed? or are they visible on both?

protocols/horizon/effects/main.go
2opremio: OK, I will rename it to simply `clawed_back`

protocols/horizon/effects/main.go
```diff
@@ -171,6 +171,13 @@ const (
 
 	// EffectSignerSponsorshipRemoved occurs when the sponsorship of a signer is removed
 	EffectSignerSponsorshipRemoved EffectType = 74 // from revoke_sponsorship
+
+	// EffectAssetClawedBack occurs when an asset is clawed back
+	EffectAssetClawedBack EffectType = 75 // from clawback
+
+	// EffectClaimableBalanceClawedBack occurs when a claimable balance is clawed back
+	EffectClaimableBalanceClawedBack EffectType = 76 // from clawback_claimable_balance
```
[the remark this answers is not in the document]
2opremio: Sure

services/horizon/internal/db2/history/main.go
```diff
@@ -177,6 +177,12 @@ const (
 
 	// EffectSignerSponsorshipRemoved occurs when the sponsorship of a signer is removed
 	EffectSignerSponsorshipRemoved EffectType = 74 // from revoke_sponsorship
+
+	// EffectAssetClawedBack occurs when an asset is clawed back
+	EffectAssetClawedBack EffectType = 75 // from clawback
+
+	// EffectClaimableBalanceClawedBack occurs when a claimable balance is clawed back
+	EffectClaimableBalanceClawedBack EffectType = 76 // from clawback_claimable_balance
```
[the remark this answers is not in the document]
2opremio: Yes, I agree, but we should fix it systematically in a separate PR. There is a ticket for it already https://github.com/stellar/go/issues/3069

protocols/horizon/effects/main.go
2opremio: Really good point.

Based on the participants code from a prior PR ( #3350 ) the `ClawedBack` effect will be in both accounts.

However, for `ClaimableBalanceClawedBack`it will only be in the account running the operation. I _think_ the account whose claimable balance is removed may receive a `EffectClaimableBalanceClaimed` which isn't totally accurate, I will see if I can refine this.

leighmcculloch: 🤔 The exact same effect appearing on two different accounts doesn't seem quite right, at least it seems inconsistent with other effects that I'm aware of, and it seems like there is some dual semantics that could be confusing for consumers.

If someone does a payment op to me, my account gets an `account_credited` effect and their account an `account_debited` effect. The symmetry is appropriate and makes building systems that consume the events straightforward and clear, both accounts are effected. Here's an example I just ran of an issuer issuing an asset to an account and receiving the return of the asset. Will we see `account_credited` and `account_debited` effects? I think we need effects like `account_credited_via_clawback` and `account_debited_via_clawback` to shift the effects to describe the actual effect on the account rather than echo the operation. This mirrors issuing and burning:

Example of issuing and burning:

Account: `GCFMEPHT2YDZICNQ4XAXUDSBAW4CC2GTPH6G5HLGTX7GCMNQXWY6HJOO`
Issuer: `GBSUVAJUG2IVIWB22O7BHMPV4YPPXLOT77NKJRDI3LOXUWXHC74AA6PG`
Transactions:
1. Account trusts the Issuer:
   ```
   AAAAAgAAAACKwjzz1geUCbDlwXoOQQW4IWjTefxunWad/mExsL2x4wAAAGQAFSYeAAAAAQAAAAEAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAEAAAAAAAAABgAAAAFURVNUAAAAAGVKgTQ2kVRYOtO+E7H15h77rdP/2qTEaNrdelrnF/gAf/////////8AAAAAAAAAAbC9seMAAABAl5c8ZCYLNl4zpcf15tW6KJoMSCZU7wJBdPxCAlwRmnIRgNTTHPfbs7M6oQujjr2GUm5ZldsHF1RcgiBljbfFDg==
   ```
   https://horizon-testnet.stellar.org/transactions/194344c631656e35c15f39a28db63dab963761fd716fce4bc2da17cb042f7b50
   - trustline_created on account
   - no effects on issuer

2. Issuer issues to Account:
   ```
   AAAAAgAAAABlSoE0NpFUWDrTvhOx9eYe+63T/9qkxGja3Xpa5xf4AAAAAGQAFSYgAAAAAQAAAAEAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAEAAAAAAAAAAQAAAACKwjzz1geUCbDlwXoOQQW4IWjTefxunWad/mExsL2x4wAAAAFURVNUAAAAAGVKgTQ2kVRYOtO+E7H15h77rdP/2qTEaNrdelrnF/gAAAAAADuaygAAAAAAAAAAAecX+AAAAABA6PesyHUaMA6+NuAzmNwiSDQ1b61Zgdax5Wm2B1KjE/pvZ5q3KmXh8w5oaSLoRJt97jZLREeRCgmcTJU5j7EgAg==
   ```
   https://horizon-testnet.stellar.org/transactions/393786c52e82ce7b2db9703c3c5a48768499f8f92ac7fe3d7e898e72a7e73e12
   - account_debited on issuer
   - account_credited on account

3. Account burns to Issuer:
   ```
   AAAAAgAAAACKwjzz1geUCbDlwXoOQQW4IWjTefxunWad/mExsL2x4wAAAGQAFSYeAAAAAgAAAAEAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAEAAAAAAAAAAQAAAABlSoE0NpFUWDrTvhOx9eYe+63T/9qkxGja3Xpa5xf4AAAAAAFURVNUAAAAAGVKgTQ2kVRYOtO+E7H15h77rdP/2qTEaNrdelrnF/gAAAAAADuaygAAAAAAAAAAAecX+AAAAABAz/eAiRZorQKTWERT9nHncWGdCHwfdeWdCVQbtVnnSn334fih2mhcgOq4J8hRIYS69Fuj9F9q/K8QyNtlQDiDAQ==
   ```
   https://horizon-testnet.stellar.org/transactions/a5097f5dcc724f3c33df3f3cd4a7fa38503a262f95c2561f72b1088e61a48ac8
   - account_debited on account
   - account_credited on issuer

leighmcculloch: Using this ☝🏻 model the claimable balance could still be something like `claimable_balance_clawed_back`, but the issuing account should still get a `account_credited_via_claimable_balance_clawback`.

2opremio: I think effects have always appeared in all the participants of an operation, so it's nothing new ( @bartekn correct me if I am wrong)

2opremio: >  I _think_ the account whose claimable balance is removed may receive a `EffectClaimableBalanceClaimed` which isn't totally accurate, I will see if I can refine this.

After further thinking, this doesn't really make sense, since a claimable balance is not really associated to an account after creation, it's an entity of its own.

2opremio: > I think we need effects like `account_credited_via_clawback` and `account_debited_via_clawback`

- Regarding `account_credited_via_clawback`, don't the `clawedback` and `claimable_balance` effect provide exactly that?

- Regarding `account_debited_via_clawback`, doesn't the `clawedback` effect tell you that? (you will see that the `from` field matches the account).


I defer to @bartekn and @tamirms on this one though, since they understand effects better.

2opremio:
@bartekn @tamirms PTAL

protocols/horizon/effects/main.go
leighmcculloch: > * Regarding `account_credited_via_clawback`, don't the `clawedback` and `claimable_balance` effect provide exactly that?

I don't think there is a `claimable_balance` effect, at least I don't see it here in the docs: https://developers.stellar.org/api/resources/effects/types/. The effects for claimable balances are `Claimable Balance Created`, `Claimable Balance Claimant Created`, `Claimable Balance Claimed`.

---

The reason I suggest orienting the effects around debiting and crediting for clawbacks is because for financial applications the debit and credit is what matters most. Applications need to surface these effects as changes in balances. The change in balance is the ultimate effect of clawback.

The reason I suggest distinguishing them from the existing `account_debited`/`account_credited` effects is because some applications will want to listen for clawbacks specifically. We could potentially reuse the existing `account_debited` and `account_credited` and insert a boolean into the existing effect to indicate if the effect is related to a clawback. Or I guess the client can always look at the operation to see that it is a `ClawbackOp`, so maybe we can reuse the existing effects with zero changes to them. That would probably be most backwards compatible with existing applications.

For claimable balances I don't think we should reuse `claimable_balance_claimed`, because it is not being claimed by a claimant, therefore `claimable_balance_clawed_back`.

So, I'm changing my proposal a bit to:
- Reuse `account_credited` (issuer) and `account_debited` (account) for ClawbackOp. No changes to effects. Clients can look at the operation to determine the balance change is due to a clawback. A credit should be created on the issuer.
- Reuse `account_credited` for `ClawbackClaimableBalanceOp` for the issuer. Add `claimable_balance_clawed_back` for the claimable balance itself.

tamirms: I'm in favor of @leighmcculloch's last proposal:

> So, I'm changing my proposal a bit to:

> Reuse account_credited (issuer) and account_debited (account) for ClawbackOp. No changes to effects. Clients can look at the operation to determine the balance change is due to a clawback. A credit should be created on the issuer.

>Reuse account_credited for ClawbackClaimableBalanceOp for the issuer. Add claimable_balance_clawed_back for the claimable balance itself.

This behavior seems to be consistent with other operations which affect account balances (See `addAccountMergeEffects()`, `addAccountCreatedEffects()`, `addCreateClaimableBalanceEffects()`, etc).

However, it looks like offer related operations do not create `account_credited` or `account_debited` effects when a trade is executed. Instead there is a `trade` effect which details the amounts which were bought and sold. I find the fact that there is no `account_credited` / `account_debited` effect surprising. Perhaps the `trade` effect could be used as precedence for allowing other effects such as the clawback effects to also imply account debit / credit. But my preference is to make the clawback operation also create account debit / credit effects.

2opremio:
> I don't think there is a `claimable_balance` effect

I meant `Claimable Balance Claimed`, sorry

protocols/horizon/effects/main.go
2opremio: > I don't think there is a `claimable_balance` effect

I meant `Claimable Balance Claimed`, sorry

2opremio: Alright, I will implement the proposal

2opremio: However, I think I will maintain the Clawedback effect

2opremio: It's now implemented. PTAL

services/horizon/internal/ingest/processors/effects_processor.go
```diff
@@ -926,6 +930,47 @@ func (e *effectsWrapper) addIngestTradeEffects(buyer xdr.AccountId, claims []xdr
...
+		from.Address(),
+		history.EffectAccountDebited,
+		details,
+	)
+
+	return nil
+}
+
+func (e *effectsWrapper) addClawbackClaimableBalanceEffects() error {
+	op := e.operation.operation.Body.MustClawbackClaimableBalanceOp()
```
tamirms, line 958: should there also be an `account_credited` effect to the issuer here?

protocols/horizon/effects/main.go
```diff
@@ -253,13 +258,15 @@ type AccountCreated struct {
 type AccountCredited struct {
 	Base
 	base.Asset
-	Amount string `json:"amount"`
+	Amount     string `json:"amount"`
+	IsClawback *bool  `json:"is_clawback,omitempty"`
```
tamirms, line 262: I don't think `is_clawback` is necessary because it can be inferred from the source operation

2opremio: OK, I will remove it

services/horizon/internal/ingest/processors/effects_processor.go
2opremio: Uhm, maybe yeah

2opremio: I think it's tricky to obtain by looking at the operation.

I see two possibilities:

* make a DB query to obtain the original claimable balance from the claimable balance ID
* infer the balance change from the operation changes

both of them may be overkill. Thoughts?

2opremio: I talked to @tamirms and we agreed to use the operation changes.

However, I will postpone it (I will add an entry to #3348 ) until I write an integration test and understand Core's behavior better.

protocols/horizon/effects/main.go
2opremio: It turns out that emitting a `account_credited` is wrong, because the `clawback` issuer doesn't get the asset back. The asset is just burned, I will create a PR removing it.

leighmcculloch: All credits to the issuer burn an asset because the issuer doesn't hold a balance of the assets. Credits are still appropriate though. That's what payments operations do. See the example here: https://github.com/stellar/go/pull/3358#discussion_r561139271.

PR #3359, opened 2021-01-19
txnbuild: impl SEP-10 v3.1 add web_auth_domain support

### LINKED ISSUE #3334: SEP-10 v3.1 Changes (from the SDF)
### TL;DR

Implement the changes described the [SEP-10 3.1 commit](https://github.com/stellar/stellar-protocol/commit/6c8c9cf6685c85509835188a136ffb8cd6b9c11c).

This likely involves adding an optional argument to the SEP-10 utility functions for adding an additional Manage Data operation containing the domain of the SEP-10 server (`web_auth_domain`).

### Background & Description

Thank you for implementing [SEP-10 v3.0](9d121f9), which reintroduced verification of the home domain of the service requiring SEP-10 authentication, ensuring that signed challenges cannot be used for any other service.

SEP-10 v3.1 adds an optional Manage Data operation that includes the home domain of the service issuing SEP-10 challenges. This allows clients and servers to verify the issuer of the challenge.

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Implement SEP-10 v3.1, adding support for building and verifying SEP-10 challenge transactions with the web_auth_domain manage data operation.

### Why

To maintain compatibility with SEP-10.

Close #3334.

### Compatibility

This SEP-10 change is completely backwards compatible with any server or client configuration using SEP-10 v2.1 or greater. If a server is on v3.1 and a client is on v2.1, they are compatible. If a server is on v2.1 and a client is on v3.1, they are compatible.

The SDK changes are however not backwards compatible because of changes to exported function signatures in the `txnbuild` package. Because of this, this change needs releasing in a major release of txnbuild.

The webauth changes are not backwards compatible because the server requires the `--domain` parameter to be specified for successful startup. This does not affect how webauth gets released because it is experimental. This will require updates to any deployments using webauth.

### Merging

This change should be merged when ready to `release-txnbuild-v6.0.0`.

### Known limitations

N/A


txnbuild/transaction_test.go
```diff
@@ -1681,9 +1711,7 @@ func TestReadChallengeTx_invalidSeqNoNotZero(t *testing.T) {
 	assert.NoError(t, err)
 	tx64, err := tx.Base64()
 	require.NoError(t, err)
-	readTx, readClientAccountID, _, err := ReadChallengeTx(tx64, serverKP.Address(), network.TestNetworkPassphrase, []string{"testanchor.stellar.org"})
-	assert.Equal(t, tx, readTx)
-	assert.Equal(t, "", readClientAccountID)
+	_, _, _, err = ReadChallengeTx(tx64, serverKP.Address(), network.TestNetworkPassphrase, "testwebauth.stellar.org", []string{"testanchor.stellar.org"})
```
leighmcculloch, line 1714: ℹ️ These expectations are unrealistic and only passed in this test because of the specific setup of the test. The Go SDK does not support stable round tripping an object through XDR serialization and deserialization, which is what these assertions assume. The transaction object contains a reference to a SimpleAccount which includes the current sequence number of the account, and that information isn't stored in the XDR and therefore the deserialized objects will have inconsistent information. This becomes a problem in this PR because the simple account with the non-zero sequence number is included in the `webAuthDomainOp`.

leighmcculloch:
Status Update:
 - [x] Add new field to txnbuild functions.
 - [x] Update txnbuild tests to provide new field.
 - [ ] Update dependencies that live in the monorepo (i.e. webauth).
 - [ ] Add new tests and new logic to txnbuild.

txnbuild/transaction_test.go
leighmcculloch: I addressed this another way in 85934424.

txnbuild/transaction.go
```diff
@@ -1040,8 +1049,19 @@ func ReadChallengeTx(challengeTx, serverAccountID, network string, homeDomains [
 		if op.SourceAccount == nil {
 			return tx, clientAccountID, matchedHomeDomain, errors.New("operation should have a source account")
 		}
-		if op.SourceAccount.GetAccountID() != serverAccountID {
-			return tx, clientAccountID, matchedHomeDomain, errors.New("subsequent operations are unrecognized")
+		switch op.Name {
+		case "web_auth_domain":
+			if op.SourceAccount.GetAccountID() != serverAccountID {
+				return tx, clientAccountID, matchedHomeDomain, errors.New("web auth domain must have serve source account")
```
JakeUrban, line 1055: nit / typo
```suggestion
				return tx, clientAccountID, matchedHomeDomain, errors.New("web auth domain operation must have server source account")
```

txnbuild/transaction.go
```diff
@@ -1040,8 +1049,19 @@ func ReadChallengeTx(challengeTx, serverAccountID, network string, homeDomains [
...
+			}
+			if !bytes.Equal(op.Value, []byte(webAuthDomain)) {
+				return tx, clientAccountID, matchedHomeDomain, errors.Errorf("web auth domain is %q but require %q", string(op.Value), webAuthDomain)
```
JakeUrban, line 1058: nit
```suggestion
				return tx, clientAccountID, matchedHomeDomain, errors.Errorf("web auth domain operation value is %q but expected %q", string(op.Value), webAuthDomain)
```

txnbuild/transaction.go
JakeUrban: This condition is checked in both `case` entries, why not just keep it where it was? I see the error messages are different, but this one isn't tested.

JakeUrban (commented):
Nothing blocking, just a couple nit picks. Otherwise LGTM 👍

txnbuild/transaction.go
leighmcculloch: 208ead0f

txnbuild/transaction.go
leighmcculloch: Thanks, those errors are better. I updated the error and added the missing tests in 208ead0f.

I think the value in doing the check separately is that it communicates the intent. The intent isn't that all subsequent operations be signed by the server but that separately defined the web auth domain is to be signed by the server, and unrecognized operations are too.

leighmcculloch (commented):
Thanks @JakeUrban. I addressed your comments.

JakeUrban (approved):

tamirms (approved):

PR #3360, opened 2021-01-21
services/horizon: Add Horizon 1.14.0 changelog

tamirms:
Add change log entry for the horizon 1.14.0 release


bartekn (approved):
LGTM. I think we can move `--checkpoint-frequency` to the bottom. It's really a dev feature.

PR #3361, opened 2021-01-21
xdr: Add LedgerEntry.Normalize() method

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add `LedgerEntry.Normalize()` method. Remove `transformEntry` in Horizon.

### Why

It's possible that ledger entry fetched from History Archives and Horizon DB are equal but are not binary the same due to ledger entry extensions. Previously normalization was done in `transformEntry` function but in #3356 a new function `NormalizeClaimableBalance` was added. To keep the code consistent, I updated it to support all ledger entry types and removed `transformEntry`.

### Known limitations

`Normalize()` is changing the internal values of the ledger entry. Ideally it should return a new, normalized object. However, due to extensive number of pointers it's possible to actually modify the original entry. Because of this we always call `Normalize()` before passing it to `verifier.Write()`. Alternatively, we could do that inside `verifier.Write()` but this would be a side-effect and I'm not sure if this is a good idea (since it's not part of the global `ingest` maybe it's fine).


2opremio (approved):

services/horizon/internal/ingest/verify/main.go
```diff
@@ -96,6 +97,7 @@ func (v *StateVerifier) GetLedgerKeys(count int) ([]xdr.LedgerKey, error) {
 // Write compares the entry with entries in the latest batch of entries fetched
 // using `GetEntries`. Entries don't need to follow the order in entries returned
 // by `GetEntries`.
+// Write expects entry to be normalized (call Normalize()).
```
Shaptic, line 100: Is there any way we can enforce this in code (e.g. checking a `.Normalize` property, or something?) unless it's already obvious that not doing this will blow things up. I'm just thinking it will be pretty easy to overlook this when we write new code!

services/horizon/internal/ingest/verify/main_test.go
```diff
@@ -289,15 +290,19 @@ func makeAccountLedgerEntry() xdr.LedgerEntry {
...
+	entry := xdr.LedgerEntry{
 		Data: xdr.LedgerEntryData{
 			Type: xdr.LedgerEntryTypeOffer,
 			Offer: &xdr.OfferEntry{
 				SellerId: xdr.MustAddress("GC3C4AKRBQLHOJ45U4XG35ESVWRDECWO5XLDGYADO6DPR3L7KIDVUMML"),
 			},
 		},
 	}
+	entry.Normalize()
+	return entry
```
Shaptic, line 307: This might be overkill, but `LedgerEntry.Normalize()` was modified to return "self":

```go
func (entry *LedgerEntry) Normalize() *LedgerEntry {
	// etc...
	return entry
}
```

then we can keep the simpler `return xdr.LedgerEntry{ ... }.Normalize()` paradigm?

Shaptic (commented):
Don't feel comfortable :heavy_check_mark:'ing, but left some thoughts!

services/horizon/internal/ingest/verify/main.go
bartekn: You're right about missing this in new code. Checking if entry is normalized is quite easy but it may blow up too if not tested properly. I think I'll spend more time on `Normalize` method and update it to return a normalized copy. Then we can use it inside `Write()`. One thing I'm worried about is memory pressure/consumption because for each entry in history archive and DB will create two objects instead of one.

Shaptic: Does it have to be a copy? I'm assuming by "inside `Write()`" you mean via something like `verifier.Write(entry.Normalize())`; this can still be done in-place I think.

bartekn: OK, I decided not to make a copy: I was worried about memory implications (for each ledger entry we'd need to make a normalized copy). I moved `Normalize()` call to inside `Write()` because the API is not public. Updated comments.

PR #3364, opened 2021-01-21
[CAP 35] txnbuild: Add txnbuild support for the Clawback operation

2opremio:
Part of #3348

### DIFF
--- a/txnbuild/allow_trust.go
+++ b/txnbuild/allow_trust.go
@@ -2,6 +2,7 @@ package txnbuild
 
 import (
 	"bytes"
+	"fmt"
 
 	"github.com/stellar/go/support/errors"
 	"github.com/stellar/go/xdr"
@@ -61,6 +62,20 @@ func (at *AllowTrust) BuildXDR() (xdr.Operation, error) {
 	return op, nil
 }
 
+func assetCodeToCreditAsset(assetCode xdr.AssetCode) (CreditAsset, error) {
+	switch assetCode.Type {
+	case xdr.AssetTypeAssetTypeCreditAlphanum4:
+		code := bytes.Trim(assetCode.AssetCode4[:], "\x00")
+		return CreditAsset{Code: string(code[:])}, nil
+	case xdr.AssetTypeAssetTypeCreditAlphanum12:
+		code := bytes.Trim(assetCode.AssetCode12[:], "\x00")
+		return CreditAsset{Code: string(code[:])}, nil
+	default:
+		return CreditAsset{}, fmt.Errorf("unknown asset type: %d", assetCode.Type)
+	}
+
+}
+
 // FromXDR for AllowTrust initialises the txnbuild struct from the corresponding xdr Operation.
 func (at *AllowTrust) FromXDR(xdrOp xdr.Operation) error {
 	result, ok := xdrOp.Body.GetAllowTrustOp()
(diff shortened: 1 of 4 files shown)


txnbuild/clawback.go
```diff
@@ -0,0 +1,109 @@
...
+	if cb.Asset.IsNative() {
+		return xdr.Operation{}, errors.New("clawbacks don't apply to native (XLM) asset")
+	}
+
+	xdrAmount, err := amount.Parse(cb.Amount)
+	if err != nil {
+		return xdr.Operation{}, errors.Wrap(err, "failed to parse amount")
+	}
+
+	if cb.Asset == nil {
```
tamirms, line 38: this check should be done before `if cb.Asset.IsNative`

bartekn: This is also checked in `Validate` which isn't used anywhere (?).

txnbuild/allow_trust.go
```diff
@@ -61,6 +62,20 @@ func (at *AllowTrust) BuildXDR() (xdr.Operation, error) {
...
 }
 
+func assetCodeToCreditAsset(assetCode xdr.AssetCode) (CreditAsset, error) {
+	switch assetCode.Type {
+	case xdr.AssetTypeAssetTypeCreditAlphanum4:
+		code := bytes.Trim(assetCode.AssetCode4[:], "\x00")
+		return CreditAsset{Code: string(code[:])}, nil
+	case xdr.AssetTypeAssetTypeCreditAlphanum12:
+		code := bytes.Trim(assetCode.AssetCode12[:], "\x00")
+		return CreditAsset{Code: string(code[:])}, nil
```
leighmcculloch, line 72: 🐛  Is the intention to trim `\x00` from both the left and right sides of the byte slice, or only the right? I expect we only need to discard trailing `\x00`s and therefore should use `bytes.TrimRight` because when I've seen padding discussed in documentation (e.g. [SEP-11](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0011.md#asset)) it is only used to pad the tail.
```suggestion
	case xdr.AssetTypeAssetTypeCreditAlphanum4:
		code := bytes.TrimRight(assetCode.AssetCode4[:], "\x00")
		return CreditAsset{Code: string(code[:])}, nil
	case xdr.AssetTypeAssetTypeCreditAlphanum12:
		code := bytes.TrimRight(assetCode.AssetCode12[:], "\x00")
		return CreditAsset{Code: string(code[:])}, nil
```
I realize you're just moving existing code, but I wonder if this is a bug we need to address separately to this PR.

txnbuild/clawback.go
```diff
@@ -0,0 +1,109 @@
...
+	var fromMuxedAccount xdr.MuxedAccount
+
+	err := fromMuxedAccount.SetAddress(cb.From)
+	if err != nil {
+		return xdr.Operation{}, errors.Wrap(err, "failed to set from address")
+	}
+
+	// Validate this is an issued asset
+	if cb.Asset.IsNative() {
+		return xdr.Operation{}, errors.New("clawbacks don't apply to native (XLM) asset")
```
leighmcculloch, line 30: 💡 It may be a little clearer to use the term support here since product developers are more likely to think about features as being supported or not, rather than if operations will be able to apply successfully.
```suggestion
		return xdr.Operation{}, errors.New("clawbacks don't support the native (XLM) asset")
```

txnbuild/clawback_test.go
```diff
@@ -0,0 +1,82 @@
...
+			Operations:           []Operation{&payment},
+			BaseFee:              MinBaseFee,
+			Timebounds:           NewInfiniteTimeout(),
+		},
+	)
+	if assert.Error(t, err) {
+		expected := "validation failed for *txnbuild.Clawback operation: Field: Asset, Error: asset code length must be between 1 and 12 characters"
+		assert.Contains(t, err.Error(), expected)
+	}
+}
```
leighmcculloch, line 82: 💡 I think we could also have tests for:
- validation success
- round trip conversion between the xdr to the op and back result in same obj
- round trip conversion between the op to the xdr and back result in same obj
- test for build xdr
- test for from xdr

leighmcculloch (commented):
+1 to @tamirms's comment, otherwise looks great.

I also left some suggestions (💡) inline, but defer to you on them.

I also mention what I think is a bug (🐛) in existing code being moved. I think the bug should be fixed separately though on `master`, not in this PR.

txnbuild/allow_trust.go
2opremio: Would you mind creating a ticket?

txnbuild/clawback.go
2opremio: @bartekn what do you mean? it implements an interface which is used in `NewTransaction()`

```
	for _, op := range tx.operations {
		if verr := op.Validate(); verr != nil {
			return nil, errors.Wrap(verr, fmt.Sprintf("validation failed for %T operation", op))
		}
```

txnbuild/allow_trust.go
2opremio: (I think it's out of the scope of this PR)

txnbuild/clawback_test.go
2opremio: Good point,  I will add this to #3348

txnbuild/allow_trust.go
leighmcculloch: 👍🏻 https://github.com/stellar/go/issues/3369

PR #3365, opened 2021-01-21
[CAP 35] Add operations stats for new clawback operations

2opremio:
Part of #3348


leighmcculloch (approved):
🎉 Looks good. Looks like there are existing tests for this file that could be updated too to include these new stats.

bartekn (approved):

PR #3366, opened 2021-01-22
[CAP 35] Stop emitting an account_credited effect for Clawback

2opremio:
The clawed-back amount is just burned. No account is credited.


leighmcculloch (changes requested):
I don't think we should make this change.

I think we should continue to generate an `account_credited` effect for the issuer. That is the existing behavior of payments and other operations that send an amount to the issuer that burn the asset. An issuer never has a balance and any amount of its assets credited to it burn.

PR #3375, opened 2021-01-27
ingest: Use a *random* history archive from the Captive Core configuration list

Shaptic:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Introduces `historyarchive.ConnectAny`, which chooses a random history archive from a list of URLs until succeeding.

### Why
Previously, listing more than one history archive in the Captive Core configuration had no impact since the first was always used. This is (more or less) a preliminary effort into being smarter about the way we use history archives.

Moving forward, the goal may be to refactor the `historyarchive` library to use a random archive for every request for redundancy purposes, but this can be done in future PRs.

### Known limitations
I'd like to add a test.

### DIFF
--- a/historyarchive/archive.go
+++ b/historyarchive/archive.go
@@ -420,12 +420,25 @@ func Connect(u string, opts ConnectOptions) (*Archive, error) {
 	return &arch, err
 }
 
+func MustConnect(u string, opts ConnectOptions) *Archive {
+	arch, err := Connect(u, opts)
+	if err != nil {
+		log.Fatal(err)
+	}
+	return arch
+}
+
 func ConnectAny(urls []string, opts ConnectOptions) (*Archive, error) {
+	if len(urls) == 0 {
+		return nil, errors.New("No history archives provided")
+	}
+
 	var outerErr error
 	for len(urls) > 0 {
 		i := rand.Intn(len(urls))
 		archive, err := Connect(urls[i], opts)
 		if err != nil {
+			// Drop the failed archive from the list and try again
 			urls = append(urls[:i], urls[i+1:]...)
 			outerErr = err
 			continue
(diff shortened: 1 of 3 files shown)


historyarchive/archive.go
```diff
@@ -335,6 +336,23 @@ func Connect(u string, opts ConnectOptions) (*Archive, error) {
 	return &arch, err
 }
 
+func ConnectAny(urls []string, opts ConnectOptions) (*Archive, error) {
```
bartekn, line 339: This function will return `nil, nil` on `urls` with no elements. I think we should handle this error.

bartekn (approved):

historyarchive/archive.go
Shaptic: Good catch!

historyarchive/archive.go
```diff
@@ -342,3 +343,25 @@ func MustConnect(u string, opts ConnectOptions) *Archive {
...
+		return nil, errors.New("No history archives provided")
+	}
+
+	var outerErr error
+	for len(urls) > 0 {
+		i := rand.Intn(len(urls))
+		archive, err := Connect(urls[i], opts)
+		if err != nil {
+			// Drop the failed archive from the list and try again
+			urls = append(urls[:i], urls[i+1:]...)
```
bartekn, line 358: Wanted to merge it to include in the next Horizon release but realized that we probably shouldn't modify `urls` because it's a side effect for a caller. Can we copy `urls` before modifying it?

Shaptic: Wow nice catch, good thing I didn't merge this then. Fixed in f2a9c279.

Shaptic:
Closing in lieu of #3402.

PR #3377, opened 2021-01-28
txnbuild: Remove TxEnvelope() functions from txnbuild

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Remove `TxEnvelope()` from `Transaction` and `FeeBumpTransaction`. Now `ToXDR()` is the only function which produces XDR transaction envelopes.

### Why

It simplifies the API to just have one function which produces an XDR transaction envelope. Although it would be nice to have immutable transaction envelopes, the inconvenience of checking the error result might outweigh the benefit. Given that the `Signatures()` and `Operations()` functions assume the caller will not modify the returned slice, I think it is ok for us to be consistent and hold the same assumption in `ToXDR()`.

### Known limitations

[N/A]

### DIFF
--- a/txnbuild/transaction.go
+++ b/txnbuild/transaction.go
@@ -195,7 +195,6 @@ type Transaction struct {
 	operations    []Operation
 	memo          Memo
 	timebounds    Timebounds
-	signatures    []xdr.DecoratedSignature
 }
 
 // BaseFee returns the per operation fee for this transaction.
@@ -232,7 +231,7 @@ func (t *Transaction) Operations() []Operation {
 // Signatures returns the list of signatures attached to this transaction.
 // The contents of the returned slice should not be modified.
 func (t *Transaction) Signatures() []xdr.DecoratedSignature {
-	return t.signatures
+	return t.envelope.Signatures()
 }
 
 // Hash returns the network specific hash of this transaction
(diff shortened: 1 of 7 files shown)


txnbuild/transaction.go
```diff
@@ -313,21 +300,19 @@ func (t *Transaction) AddSignatureBase64(network, publicKey, signature string) (
...
 func (t *Transaction) ToXDR() xdr.TransactionEnvelope {
 	env := t.envelope
 	switch env.Type {
 	case xdr.EnvelopeTypeEnvelopeTypeTx:
+		env.V1 = new(xdr.TransactionV1Envelope)
+		*env.V1 = *t.envelope.V1
 		env.V1.Signatures = t.signatures
 	case xdr.EnvelopeTypeEnvelopeTypeTxV0:
+		env.V0 = new(xdr.TransactionV0Envelope)
+		*env.V0 = *t.envelope.V0
```
leighmcculloch, line 315: ❓ Why does this change include copying the V0 and V1 to itself? The signatures aren't copied, or the original envelope object. If the function says it returns an object that shouldn't be modified I don't think you need to do this copy.

leighmcculloch, line 315: 💡 We might want to consider storing signatures directly on the envelope at the time the signatures are added. In multiple places we set them lazily on the envelope which I don't think is particularly intuitive. If we set them upfront functions like `ToXDR` would simply be:
```go
return t.envelope
```

leighmcculloch (approved):
I think reducing the number of functions that return the same thing, and therefore reducing the API, is good.

I think returning a transaction envelope that shouldn't be modified by the caller is also okay. As you say this API already does this in multiple places.

I think the errorless panic function is okay, because as far as I can tell you the SDK doesn't allow the panic case to occur unless there was a logical bug in the SDK.

I left comments inline containin a question (❓) and suggestion (💡). Regardless this looks 👍🏻. I defer to @stellar/horizon-committers.

txnbuild/transaction.go
```diff
@@ -260,18 +246,36 @@ func (t *Transaction) HashHex(network string) (string, error) {
 	return hashHex(t.envelope, network)
 }
 
+func (t *Transaction) clone(signatures []xdr.DecoratedSignature) *Transaction {
```
bartekn, line 249: This may be OK but this function does not clone objects behind pointers. The removed `cloneEnvelope` did so, so it's worth checking if any assumptions were not broken by this change.

bartekn (approved):

txnbuild/transaction.go
tamirms: `Transaction` never modifies the transaction envelope and when `ToXDR()` returns the transaction envelope we assume the caller will not modify the transaction envelope

PR #3385, opened 2021-02-03
Update golang to 1.14.14 and 1.15.7

bartekn:
Update golang to 1.14.14 and 1.15.7 because of CVE-2021-3115.

More information:
* https://groups.google.com/g/golang-announce/c/mperVMGa98w
* https://blog.golang.org/path-security

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -244,7 +244,7 @@ jobs:
   check_code_1_15:
     working_directory: /go/src/github.com/stellar/go
     docker:
-      - image: golang:1.15
+      - image: golang:1.15.7
     steps:
       - install_go_deps
       - check_go_deps
@@ -258,7 +258,7 @@ jobs:
   test_code_1_14:
     working_directory: /go/src/github.com/stellar/go
     docker:
-      - image: golang:1.14-stretch
+      - image: golang:1.14.14-stretch
         environment:
           GO111MODULE: "on"
           PGHOST: localhost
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -244,7 +244,7 @@ jobs:
   check_code_1_15:
     working_directory: /go/src/github.com/stellar/go
     docker:
-      - image: golang:1.15
+      - image: golang:1.15.7
```
leighmcculloch, line 247: Specifying the minor version tag is unnecessary. The `golang:1.15` tag is updated to point to the latest minor version, which has the advantage that whenever a new version is released we should be using that newer version.

Example:
Screen Shot 2021-02-03 at 7 19 02 AM
Ref: https://hub.docker.com/_/golang

It might be possible that CircleCI is caching the 1.15 build from a prior version though, which would make it worthwhile to specify the minor version specifically to make sure we're using a specific minor version. Is that your thought?

leighmcculloch (approved):
One question above but regardless this looks 👍.

.circleci/config.yml
bartekn: > It might be possible that CircleCI is caching the 1.15 build from a prior version though, which would make it worthwhile to specify the minor version specifically to make sure we're using a specific minor version. Is that your thought?

This was exactly my concern. If CircleCI is not caching we can switch to 1.14 and 1.15. Will check Circle logs to confirm.

leighmcculloch: I think they do cache, and that this change is good. We need to be more diligent to update our builds straight away when new releases come out, but I think that's fine, we can do that.

PR #3387, opened 2021-02-03
Add --pull flag to docker builds

jacekn:
### What

Add --pull flag to docker builds

### Why

By default if images exist locally they are used for builds.
This change will ensure we always pull latest versions which ensures we get all security patches.


brahman81 (approved):
LGTM, thanks Jacek

leighmcculloch (approved):

Issue #3340, opened 2021-01-13
Feature Request: Add functionality to history archive client for fetching transactions


tamirms:
Previously, we had a `HistoryArchiveBackend` implementation which used the history archives to fetch transactions belonging to a specific ledger. We decided to remove the implementation in https://github.com/stellar/go/issues/3321 because the transactions did not contain tx meta and that resulted in some inconsistencies with other `ingest.LedgerBackend` implementations.

However, the code to extract transactions from the history archives could be useful so we should add that functionality to the history archive client.

Issue #3321, opened 2021-01-08
Remove HistoryArchiveBackend


tamirms:
The `LedgerBackend` interface has several implementations:

- `CaptiveStellarCore`
- `DatabaseBackend`
- `RemoteCaptiveStellarCore`
- `HistoryArchiveBackend`

`HistoryArchiveBackend` is not consistent with the other `LedgerBackend` implementations because it produces `xdr.LedgerCloseMeta` instances which **omit** tx meta. This is a strong indication that `HistoryArchiveBackend` is not suited for the `LedgerBackend` interface. It's also worth noting that `HistoryArchiveBackend` is not used anywhere in the go monorepo.

I think it's best if we remove `HistoryArchiveBackend` since it is not used by Horizon and it will reduce confusion for users of the ingestion library.

PR #3389, opened 2021-02-03
Add function to fetch ledgers from history archive

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Close https://github.com/stellar/go/issues/3340

Previously, we had a HistoryArchiveBackend implementation which used the history archives to fetch transactions belonging to a specific ledger. We decided to remove the implementation in #3321 because the transactions did not contain tx meta and that resulted in some inconsistencies with other ingest.LedgerBackend implementations.

However, the code to extract transactions from the history archives could be useful so we should add that functionality to the history archive client.

### Known limitations

[N/A]

### DIFF
--- a/historyarchive/archive.go
+++ b/historyarchive/archive.go
@@ -71,7 +71,7 @@ type ArchiveInterface interface {
 	CategoryCheckpointExists(cat string, chk uint32) (bool, error)
 	GetLedgerHeader(chk uint32) (xdr.LedgerHeaderHistoryEntry, error)
 	GetRootHAS() (HistoryArchiveState, error)
-	GetLedgers(start, end uint32) (map[uint32]Ledger, error)
+	GetLedgers(start, end uint32) (map[uint32]*Ledger, error)
 	GetCheckpointHAS(chk uint32) (HistoryArchiveState, error)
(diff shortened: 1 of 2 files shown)


bartekn (commented):
LGTM! I wonder if we should create a new struct that contains only the data we can find in history archives. `xdr.LedgerCloseMeta` can still be confusing because it contains meta related fields that are not present in history archives. What about:
```go
type HistoryLedger struct {
	LedgerHeader      LedgerHeaderHistoryEntry
	Transaction       TransactionHistoryEntry
	TransactionResult TransactionHistoryResultEntry
}
```

tamirms:

historyarchive/archive.go
```diff
@@ -195,6 +202,74 @@ func (a *Archive) GetRootHAS() (HistoryArchiveState, error) {
 	return a.GetPathHAS(rootHASPath)
 }
 
+func (a *Archive) GetLedgers(start, end uint32) (map[uint32]Ledger, error) {
```
bartekn, line 205: Nit: Why not a pointer to `Ledger`? Tx results may potentially be huge so I wonder about implications on memory alocations.

bartekn (approved):

PR #3390, opened 2021-02-03
all: enable codeql code analysis

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Enable GitHub's CodeQL static analysis to run periodically and on PRs to `master`. It will open issues in the Security tab when it finds problems.

### Why

To get feedback about issues we might miss.

I wanted to enable this a while ago but there was no Go support, but I just stumbled on an advertisement in the GitHub UI suggesting it for this repo. I'm not sure how long ago they added support, but here we are. The yml file was mostly created by the GitHub UI suggestion box, but I edited it to include `release-*` branches.

### Known limitations

N/A

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -10,7 +10,7 @@ on:
     - 'master'
     - 'release-*'
   schedule:
-    - cron: '35 22 * * 6'
+    - cron: '30 17 * * 1,2,3,4,5'
 
 jobs:
   analyze:
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,38 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches:
+    - 'master'
+    - 'release-*'
+  pull_request:
+    branches:
+    - 'master'
+    - 'release-*'
+  schedule:
+    - cron: '35 22 * * 6'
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,38 @@
...
+  push:
+    branches:
+    - 'master'
+    - 'release-*'
+  pull_request:
+    branches:
+    - 'master'
+    - 'release-*'
+  schedule:
+    - cron: '35 22 * * 6'
```
ire-and-curses, line 13: 22.35 on a Saturday seems oddly specific?

If there's no downside, maybe this should be run more frequently than once a week. Once a day?

ire-and-curses (approved):
Thanks for this. Seems like a good thing to try out to me!

.github/workflows/codeql-analysis.yml
leighmcculloch: GitHub suggested running it once a week, and defaulted to that value. I agree, we should chnage it. It takes about 7 mins to run. Let's try it every weekday and see how it goes. I changed it in https://github.com/stellar/go/pull/3390/commits/2a8ec1aef28dde744c9c2e6e070f634c145c2383.

bartekn (approved):

PR #3391, opened 2021-02-04
services/horizon: Add docker image for checking reproducible builds

bartekn:
### What

Adds a docker image to allow comparing hashes of GitHub builds and local builds. Remove `buildTime` flag from build tags in `build_release_artifacts` to make the builds reproducible.

### Why

Creating reproducible builds and checking them will make our release process more secure.

### Known limitations

It doesn't check Debian packages yet. Will add this in a separate PR.

### DIFF
--- a/services/horizon/internal/scripts/check_release_hash/Dockerfile
+++ b/services/horizon/internal/scripts/check_release_hash/Dockerfile
@@ -1,0 +1,19 @@
+# Change to Go version used in CI or rebuild with --build-arg.
+ARG GO_IMAGE=golang:1.14.14-stretch
+FROM $GO_IMAGE
+
+WORKDIR /go/src/github.com/stellar/go
+
+ENV DEBIAN_FRONTEND=noninteractive
+# ca-certificates are required to make tls connections
+RUN apt-get update && apt-get install -y --no-install-recommends ca-certificates curl wget gnupg apt-utils git zip unzip
+RUN wget -qO - https://apt.stellar.org/SDF.asc | APT_KEY_DONT_WARN_ON_DANGEROUS_USAGE=true apt-key add -
+RUN echo "deb https://apt.stellar.org xenial stable" >/etc/apt/sources.list.d/SDF.list
+
+RUN git clone https://github.com/stellar/go.git /go/src/github.com/stellar/go
+# Fetch dependencies and prebuild binaries. Not necessary but will make check faster.
+RUN go run -v ./support/scripts/build_release_artifacts
+
+COPY check.sh .
+RUN chmod +x check.sh
+ENTRYPOINT ["./check.sh"]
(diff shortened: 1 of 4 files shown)


services/horizon/internal/scripts/check_release_hash/check.sh
```diff
@@ -0,0 +1,31 @@
...
+
+for S in "${suffixes[@]}"
+do
+	echo $TAG-$S
+    shasum -a 256 ./released/$TAG-$S
+    shasum -a 256 ./dist/$TAG-$S
+done
+
+apt-get update && apt-get install -y stellar-horizon
+shasum -a 256 $(which stellar-core)
```
tamirms, line 31: should this be stellar-horizon ?

services/horizon/internal/scripts/check_release_hash/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+
+ENV DEBIAN_FRONTEND=noninteractive
+# ca-certificates are required to make tls connections
+RUN apt-get update && apt-get install -y --no-install-recommends ca-certificates curl wget gnupg apt-utils git zip unzip
+RUN wget -qO - https://apt.stellar.org/SDF.asc | APT_KEY_DONT_WARN_ON_DANGEROUS_USAGE=true apt-key add -
+RUN echo "deb https://apt.stellar.org xenial stable" >/etc/apt/sources.list.d/SDF.list
+
+RUN git clone https://github.com/stellar/go.git /go/src/github.com/stellar/go
+# Fetch dependencies and prebuild binaries. Not necessary but will make check faster.
+RUN go run -v ./support/scripts/build_release_artifacts
```
tamirms, line 15: it seems like this is run again in the check script:

```
CIRCLE_TAG=$TAG go run -v ./support/scripts/build_release_artifacts
```

services/horizon/internal/scripts/check_release_hash/check.sh
bartekn: Good catch, pushing a fix.

services/horizon/internal/scripts/check_release_hash/Dockerfile
bartekn: Check the comment above. It's to precompile existing files during image build. This should make next build (when container is started) much faster.

tamirms: doesn't that assume that most files will stay the same from the moment the image is built until the moment the image is run? how will this image be used? are we going to build it once and then publish it to the public docker repo?

bartekn: Many packages (especially Go standard packages) stay the same between releases. This image is meant to be built once and then used with `docker run`. Image need only be rebuilt when we bump Golang version.

tamirms (approved):

Issue #2813, opened 2020-07-14
txnbuild: Use strings to represent accounts in txnbuild operations


tamirms:
We have an `Account` interface to represent Stellar accounts. The interface includes functions to get the account address and the account sequence number. This interface is useful for describing the source account of a Stellar transaction since we need to obtain the sequence number in order to define the transaction.

Many transaction operations also include `Account` fields. However, we never need the sequence number of an account to build an operation XDR. Instead of using the `Account` interface in operation structs, it would be more ergonomic to use strings since we only care about he account address when constructing operations.

Replacing the type of `Account` fields in all operation structs would require a major version bump. We should plan to include this change in the 4.0.0 release of `txnbuild`.

ire-and-curses:
This has been done, released in txnbuild v6.

PR #3393, opened 2021-02-05
txnbuild: Use strings to represent accounts in txnbuild operations

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Close https://github.com/stellar/go/issues/2813

Use strings to represent accounts in txnbuild operations.

This PR also fixes a bug `ClaimableBalanceID()`. The function builds an `xdr.OperationId` using the **operation** source account and the sequence number of the **operation** source account. It seems like we should be using the **transaction** source account and the sequence number found in the **transaction**. That is my understanding based on https://github.com/stellar/stellar-core/blob/b0cc4a9b8c3423358d1881b991834cf658bc38f5/src/transactions/CreateClaimableBalanceOpFrame.cpp#L278[…]L288

### Why

We have an `Account` interface to represent Stellar accounts. The interface includes functions to get the account address and the account sequence number. This interface is useful for describing the source account of a Stellar transaction since we need to obtain the sequence number in order to define the transaction.

Many transaction operations also include `Account` fields. However, we never need the sequence number of an account to build an operation XDR. Instead of using the `Account` interface in operation structs, it would be more ergonomic to use strings since we only care about he account address when constructing operations.

### Known limitations

[N/A]

### DIFF
--- a/txnbuild/transaction.go
+++ b/txnbuild/transaction.go
@@ -335,6 +335,10 @@ func (t *Transaction) ClaimableBalanceID(operationIndex int) (string, error) {
 		return "", errors.New("invalid operation index")
 	}
 
+	if _, ok := t.operations[operationIndex].(*CreateClaimableBalance); !ok {
+		return "", errors.New("operation is not CreateClaimableBalance")
+	}
+
 	// We mimic the relevant code from Stellar Core
 	// https://github.com/stellar/stellar-core/blob/9f3cc04e6ec02c38974c42545a86cdc79809252b/src/test/TestAccount.cpp#L285
 	operationId := xdr.OperationId{
--- a/txnbuild/transaction.go
+++ b/txnbuild/transaction.go
@@ -335,29 +335,13 @@ func (t *Transaction) ClaimableBalanceID(operationIndex int) (string, error) {
 		return "", errors.New("invalid operation index")
 	}
 
-	operation, ok := t.operations[operationIndex].(*CreateClaimableBalance)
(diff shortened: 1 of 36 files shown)


txnbuild/transaction.go
```diff
@@ -335,29 +335,13 @@ func (t *Transaction) ClaimableBalanceID(operationIndex int) (string, error) {
 		return "", errors.New("invalid operation index")
 	}
 
-	operation, ok := t.operations[operationIndex].(*CreateClaimableBalance)
-	if !ok {
-		return "", errors.New("operation is not CreateClaimableBalance")
-	}
```
Shaptic, line 341: I think this error message still has value :thinking: even if we don't need the returned `operation` anymore.

Shaptic (approved):
You should probably wait on another :heavy_check_mark: but this LGTM. Thanks again for catching the balance ID mistake.

txnbuild/transaction.go
tamirms: good point!

bartekn (approved):
Thanks! 🚀  It was always a pain to use...

leighmcculloch:
>This PR also fixes a bug ClaimableBalanceID()

The bug being fixed could be a minor release for the current major release of txnbuild. What's the motivation for bundling this into the next major release?

marcelosalloum:
I agree, that fix could be shipped in a minor release IMO.

PR #3402, opened 2021-02-11
historyarchive: Introduce a History Archive pool that's selected from for all calls

Shaptic:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
This is an improvement over #3375 which just chooses a random URL from the list. Instead, we should actually create a pool of all history archive connections and cycle through them randomly for each HA-related operation we need.

### Why
With this change, we actually use the entire URL list in the configuration that the user provides. This adds resilience and redundancy. See stellar/horizon-internal#3 for more context.

### Known limitations
n/a

### DIFF
--- a/historyarchive/pool.go
+++ b/historyarchive/pool.go
@@ -11,9 +11,9 @@ import (
 	"github.com/stellar/go/xdr"
 )
 
-// A PooledArchive is just a collection of `ArchiveInterface`s so that we can
+// A ArchivePool is just a collection of `ArchiveInterface`s so that we can
 // distribute requests fairly throughout the pool.
-type PooledArchive []ArchiveInterface
+type ArchivePool []ArchiveInterface
 
 // CreatePool tries connecting to each of the provided history archive URLs,
 // returning a pool of valid archives.
@@ -24,7 +24,7 @@ type PooledArchive []ArchiveInterface
 //
 // Possible FIXME for the above limitation: return []error instead? but then
 // users need to check `len(pool) > 0` instead of `err == nil`.
-func CreatePool(archiveURLs []string, config ConnectOptions) (PooledArchive, error) {
+func CreatePool(archiveURLs []string, config ConnectOptions) (ArchivePool, error) {
 	if len(archiveURLs) <= 0 {
 		return nil, errors.New("No history archives provided")
 	}
@@ -32,7 +32,7 @@ func CreatePool(archiveURLs []string, config ConnectOptions) (PooledArchive, err
 	var lastErr error = nil
 
 	// Try connecting to all of the listed archives, but only store valid ones.
-	var validArchives PooledArchive
+	var validArchives ArchivePool
 	for _, url := range archiveURLs {
 		archive, err := Connect(
 			url,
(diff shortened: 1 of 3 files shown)


ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -146,21 +146,35 @@ func NewCaptive(config CaptiveCoreConfig) (*CaptiveStellarCore, error) {
...
+		}
+
+		validArchives = append(validArchives, archive)
+	}
+
+	pool, err := historyarchive.CreatePool(validArchives...)
 	if err != nil {
 		cancel()
-		return nil, errors.Wrap(err, "error connecting to history archive")
+		config.Log.Error("Error connecting to ALL history archives.")
```
tamirms, line 172: I think you can return a wrapped error `errors.Wrap(err, "Error connecting to ALL history archives.")` instead of logging the error

historyarchive/pool.go
```diff
@@ -0,0 +1,97 @@
...
+import (
+	"math/rand"
+
+	"github.com/stellar/go/support/errors"
+	"github.com/stellar/go/xdr"
+)
+
+// Type PooledArchive forwards all API calls to a random ArchiveInterface within
+// its internal pool.
+type PooledArchive struct {
```
tamirms, line 16: I think it would be better if you defined the type as:

`type PooledArchive []ArchiveInterface`

Then you can have the constructor which takes a list of urls and a config options struct and move the code from captive core to here:

```golang
	for _, url := range config.HistoryArchiveURLs {
		archive, err := historyarchive.Connect(
			url,
			historyarchive.ConnectOptions{
				NetworkPassphrase:   config.NetworkPassphrase,
				CheckpointFrequency: config.CheckpointFrequency,
				Context:             config.Context,
			},
		)

		if err != nil {
			config.Log.Warnf("Error connecting to history archive (%s): %s", url, err)
			continue
		}

		validArchives = append(validArchives, archive)
	}
```

historyarchive/pool.go
```diff
@@ -0,0 +1,97 @@
+// Copyright 2016 Stellar Development Foundation and contributors. Licensed
+// under the Apache License, Version 2.0. See the COPYING file at the root
+// of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
+
```
bartekn, line 4: I don't think we need copyright comment on top of this file. At least we don't have it in most of other files. @ire-and-curses?
```suggestion
```

bartekn (commented):
LGTM! I agree with @tamirms's `type PooledArchive []ArchiveInterface` idea, it should be easier to use from outside.

historyarchive/pool.go
ire-and-curses: Re: copyright - let's just follow whatever we do everywhere else in the codebase.

Shaptic: Yeah so I was confused by this, too. In the `historyarchive` package, **all** of the files have copyright notices. The years are mixed, but all of them have one. So I added it here.

historyarchive/pool.go
Shaptic: There's a limitation with this approach (cc @bartekn): 

We either change `type CaptiveStellarCore` to take the new `PooledArchive`, or we just pass `archive:          historyarchive.GetRandomArchive(archivePool)` (I did this in 3da69c13) but this is basically equivalent to #3375 . Essentially, with this approach we don't actually distribute *actions* to the pool.

tamirms: @Shaptic you can define methods on `PooledArchive` so you can keep the functions like:

```
func (pa PooledArchive) GetAnyArchive() ArchiveInterface {
	return pa[rand.Intn(len(pa.pool))]
}

func (pa PooledArchive) GetPathHAS(path string) (HistoryArchiveState, error) {
	return pa.GetAnyArchive().GetPathHAS(path)
}

func (pa PooledArchive) PutPathHAS(path string, has HistoryArchiveState, opts *CommandOptions) error {
	return pa.GetAnyArchive().PutPathHAS(path, has, opts)
}```

File: historyarchive/pool.go [hunk 4]
```diff
@@ -0,0 +1,62 @@
...
+// CreatePool tries connecting to each of the provided history archive URLs,
+// returning a pool of valid archives.
+//
+// If none of the archives work, this returns the error message of the last
+// failed archive. Note that the errors for each individual archive are hard to
+// track if there's success overall.
+//
+// Possible FIXME for the above limitation: return []error instead? but then
+// users need to check `len(pool) > 0` instead of `err == nil`.
+func CreatePool(archiveURLs []string, config ConnectOptions) (*PooledArchive, error) {
```
tamirms, line 26: I would return `PooledArchive`. the pointer is not necessary

historyarchive/pool.go
Shaptic: :facepalm: of course. I got an error when I tried it initially and just assumed it was because arrays couldn't have methods on them... Done in c3429511.

historyarchive/pool.go
Shaptic: Haven't really figured this pattern out yet, so thank you for the tip!

historyarchive/pool.go
```diff
@@ -0,0 +1,136 @@
...
+// CreatePool tries connecting to each of the provided history archive URLs,
+// returning a pool of valid archives.
+//
+// If none of the archives work, this returns the error message of the last
+// failed archive. Note that the errors for each individual archive are hard to
+// track if there's success overall.
+//
+// Possible FIXME for the above limitation: return []error instead? but then
+// users need to check `len(pool) > 0` instead of `err == nil`.
+func CreatePool(archiveURLs []string, config ConnectOptions) (ArchivePool, error) {
```
tamirms, line 27: I prefer the name `NewPool()` so that it's consistent with other constructors which have the New prefix

tamirms (approved):
looks great! I left just one minor comment about the constructor name

historyarchive/pool.go
Shaptic: Good call! Renamed to `NewArchivePool` (and its filename) to match the type name.

tamirms (approved):

PR #3403, opened 2021-02-11
all: Update app.version string build flag

bartekn:
### What

Update app version string in `build_release_artifacts` script to match build other pipelines.

### Why

We want final binaries to be the same to allow reproducible builds.

### Known limitations

This will change the version string for other apps built by `build_release_artifacts`.


tamirms (approved):

leighmcculloch (commented):
Nice!

PR #3406, opened 2021-02-12
txnbuild: Use correct account in ClaimableBalanceID()

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

ClaimableBalanceID() builds an xdr.OperationId using the operation source account and the sequence number of the operation source account. It seems like we should be using the transaction source account and the sequence number found in the transaction. That is my understanding based on https://github.com/stellar/stellar-core/blob/b0cc4a9b8c3423358d1881b991834cf658bc38f5/src/transactions/CreateClaimableBalanceOpFrame.cpp#L278[…]L288

This fix was actually included in the v6 release https://github.com/stellar/go/pull/3393 . However, @leighmcculloch and @marcelosalloum would like to create a patch release for just this fix instead of including it in a major version release.

### Known limitations

[N/A]

### DIFF
--- a/txnbuild/CHANGELOG.md
+++ b/txnbuild/CHANGELOG.md
@@ -3,6 +3,10 @@
 All notable changes to this project will be documented in this
 file.  This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [v5.0.1](https://github.com/stellar/go/releases/tag/horizonclient-v5.0.1) - 2021-02-12
+
+* Fix a bug in `ClaimableBalanceID()` where the wrong account was used to derive the claimable balance id ([#3404](https://github.com/stellar/go/pull/3404))
+
 ## [v5.0.0](https://github.com/stellar/go/releases/tag/horizonclient-v5.0.0) - 2020-11-12
 
 ### Breaking changes
--- a/txnbuild/create_claimable_balance_test.go
+++ b/txnbuild/create_claimable_balance_test.go
@@ -2,6 +2,11 @@ package txnbuild
 
 import (
 	"testing"
+	"time"
+
+	"github.com/stellar/go/keypair"
+	"github.com/stellar/go/network"
+	"github.com/stretchr/testify/assert"
 )
 
 func TestCreateClaimableBalanceRoundTrip(t *testing.T) {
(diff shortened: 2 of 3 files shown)


txnbuild/create_claimable_balance_test.go
```diff
@@ -31,3 +36,45 @@ func TestCreateClaimableBalanceRoundTrip(t *testing.T) {
...
+			Operations:           []Operation{&claimableBalanceEntry},
+		},
+	)
+	assert.NoError(t, err)
+	tx, err = tx.Sign(network.TestNetworkPassphrase, aKeys)
+	assert.NoError(t, err)
+
+	balanceId, err := tx.ClaimableBalanceID(0)
+	assert.NoError(t, err)
+	assert.Equal(t, "0000000095001252ab3b4d16adbfa5364ce526dfcda03cb2258b827edbb2e0450087be51", balanceId)
```
Shaptic, line 79: How did you determine this ID to compare against? 

This depends on your answer to the above, but do you think it'd be possible to make/modify one of the integration tests such that it actually fails _without_ your changes? The existing `TestProtocol15Basics/BalanceIDs` test passes without this change, which obviously led to a false sense of security in the feature's correctness, so it'd be good to have deeper validation for this fix in future test runs.

txnbuild/transaction.go
```diff
@@ -353,18 +353,12 @@ func (t *Transaction) ClaimableBalanceID(operationIndex int) (string, error) {
...
 	}
 
-	// Use the operation's source account or the transaction's source if not.
-	var account Account = &t.sourceAccount
-	if operation.SourceAccount != nil {
-		account = operation.GetSourceAccount()
-	}
-
-	seq, err := account.GetSequenceNumber()
+	seq, err := t.sourceAccount.GetSequenceNumber()
```
Shaptic, line 361: As long as we're sure `t.sourceAccount` is never `nil` this is safe, but I can't make that guarantee off the top of my head.

txnbuild/CHANGELOG.md
```diff
@@ -3,6 +3,10 @@
 All notable changes to this project will be documented in this
 file.  This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [v5.0.1](https://github.com/stellar/go/releases/tag/horizonclient-v5.0.1) - 2021-02-12
+
+* Fix a bug in `ClaimableBalanceID()` where the wrong account was used to derive the claimable balance id ([#3404](https://github.com/stellar/go/pull/3404))
```
Shaptic, line 8: hey it was only the wrong account _sometimes_ :sweat_smile:

Shaptic (approved):
LGTM! Minor notes added.

txnbuild/create_claimable_balance_test.go
tamirms: this test will fail without my changes because the claimable balance operation source account is different from the transaction source account

txnbuild/transaction.go
tamirms: yeah, a transaction always requires a source account. if you take a look at the constructor for a transaction you'll see which check that the source account is not nil:

```golang

// NewTransaction returns a new Transaction instance
func NewTransaction(params TransactionParams) (*Transaction, error) {
	var sequence int64
	var err error

	if params.SourceAccount == nil {
		return nil, errors.New("transaction has no source account")
	}
```

Shaptic: Of course, good point! I must have mixed it up with operations since those take the tx account when not present.

Issue #3404, opened 2021-02-12
Feature Request: add `X-Forwarded-For` header value to `logStartOfRequest` 


Zagan202:
### What problem does your feature solve?
Add a `X-Forwarded-For` field to `logStartOfRequest`

### What would you like to see?

```
l := log.Ctx(r.Context()).WithFields(log.F{
		"subsys":    "http",
		"path":      r.URL.String(),
		"method":    r.Method,
		"ip":        r.RemoteAddr,
		"host":      r.Host,
		"useragent": r.Header.Get("User-Agent"),
              "X-Forwarded-For":r.Header.Get("X-Forwarded-For"),
	})
```

### What alternatives are there?

Not logging that value and trying to find the value in kubernetes

PR #3407, opened 2021-02-13
go/support/http 3404 add x forwarded for header to log start of request

Zagan202:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add xff field to the http middleware logger.
Closes: #3404 

### Why

Logging the The X-Forwarded-For (XFF) header will help identify the originating IP address of a client . 

### Known limitations

N/A

### DIFF
--- a/support/http/logging_middleware.go
+++ b/support/http/logging_middleware.go
@@ -50,12 +50,13 @@ func logStartOfRequest(
 	r *stdhttp.Request,
 ) {
 	l := log.Ctx(r.Context()).WithFields(log.F{
-		"subsys":    "http",
-		"path":      r.URL.String(),
-		"method":    r.Method,
-		"ip":        r.RemoteAddr,
-		"host":      r.Host,
-		"useragent": r.Header.Get("User-Agent"),
+		"subsys":          "http",
+		"path":            r.URL.String(),
+		"method":          r.Method,
+		"ip":              r.RemoteAddr,
+		"host":            r.Host,
+		"useragent":       r.Header.Get("User-Agent"),
+		"X-Forwarded-For": r.Header.Get("X-Forwarded-For"),
 	})
 	l.Info("starting request")
 }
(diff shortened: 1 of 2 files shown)


tamirms:
I would prefer using the [RealIP middleware](https://github.com/go-chi/chi/blob/master/middleware/realip.go) instead of including `X-Forwarded-For` in the logging middleware

support/http/logging_middleware.go
```diff
@@ -50,12 +50,13 @@ func logStartOfRequest(
...
-		"ip":        r.RemoteAddr,
-		"host":      r.Host,
-		"useragent": r.Header.Get("User-Agent"),
+		"subsys":          "http",
+		"path":            r.URL.String(),
+		"method":          r.Method,
+		"ip":              r.RemoteAddr,
+		"host":            r.Host,
+		"useragent":       r.Header.Get("User-Agent"),
+		"X-Forwarded-For": r.Header.Get("X-Forwarded-For"),
```
leighmcculloch, line 59: 💡 This log field should probably be named using a similar pattern to others, e.g.:
```suggestion
		"xforwardedfor": r.Header.Get("X-Forwarded-For"),
```
Without special characters will probably make it easier to use in queries in Kibana. @stellar/ops-team may be able to shed more light if this will be an issue or not as is.

leighmcculloch (approved):
One suggestion (💡), otherwise :shipit:.

support/http/logging_middleware_test.go
```diff
@@ -11,10 +12,17 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+func setXFFMiddleWare(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		r.Header.Set("X-Forwarded-For", "4.1.1")
```
marcelosalloum, line 17: nitpick: using a real IP would make the tests easier to read.

Examples [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For#examples).

marcelosalloum (approved):
LGTM! Just added one suggestion ⬇️

support/http/logging_middleware_test.go
marcelosalloum: `4.1.1` looks more like a software version rather than an IP.

PR #3411, opened 2021-02-16
all: update go1.16rc1 ci to use go1.16

leighmcculloch:
### What

Update Go 1.16rc1 CI builds to use the final release Go 1.16.

### Why

It was just released, and we should continue to test our code on the latest versions of Go as they become available.

### Known limitations

This change makes no changes to the supported Go versions that are already listed or tested, or the versions of Go that artifacts are built with. If we are to move to officially supporting only 1.15 and 1.16 there are changes still required to remove builds using 1.14 and update artifact building.

This PR CI runs won't pass until the golang:1.16 images are finalized and pushed.


leighmcculloch:
@bartekn Ready for review.

bartekn (approved):

PR #3412, opened 2021-02-17
historyarchive: Fix ArchivePool functions

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

`NewArchivePool() ` returns a `ArchivePool` instance which I expected to conform to `ArchiveInterface`. 

However, it turns out that `ArchivePool` does not conform to the interface because all of its functions are defined on pointer receivers. Since `ArchivePool` is defined to be a slice we should rarely (if ever) need to create a pointer to an `ArchivePool` instance. Therefore it would be better if all of the functions acted on `ArchivePool` instances instead of `ArchivePool` pointers.

### DIFF
--- a/historyarchive/archive_pool.go
+++ b/historyarchive/archive_pool.go
@@ -56,7 +56,7 @@ func NewArchivePool(archiveURLs []string, config ConnectOptions) (ArchivePool, e
 }
 
 // Ensure the pool conforms to the ArchiveInterface
-var _ ArchiveInterface = &ArchivePool{}
+var _ ArchiveInterface = ArchivePool{}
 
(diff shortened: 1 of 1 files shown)


Shaptic (approved):
A misunderstanding of Go on my part! See [the commit](https://github.com/stellar/go/pull/3402/commits/8a80c1a66e69118e4e7c983d80f92bd050ae0894) that actually touched on my uncertainty about this:

> I think that methods on a `*Type` are way less restrictive than methods on a `Type`, right?

Thanks! :heavy_check_mark:

PR #3413, opened 2021-02-18
services/horizon: Remove "Starting request" log entries

bartekn:
### What

This commit removes `Starting request` log entries from Horizon. It also changes `Finished request` to `HTTP request`

### Why

`Finished request` contains exactly the same subset of fields as `Starting request`. This change will decrease number of log entries from HTTP request by around a half.

### Known limitations

N/A

### DIFF
--- a/services/horizon/internal/httpx/middleware.go
+++ b/services/horizon/internal/httpx/middleware.go
@@ -79,11 +79,8 @@ func loggerMiddleware(serverMetrics *ServerMetrics) func(next http.Handler) http
 			acceptHeader := r.Header.Get("Accept")
 			streaming := strings.Contains(acceptHeader, render.MimeEventStream)
 
-			logStartOfRequest(ctx, r, streaming)
 			then := time.Now()
-
 			next.ServeHTTP(mw, r.WithContext(ctx))
-
 			duration := time.Since(then)
 			logEndOfRequest(ctx, r, serverMetrics.RequestDurationSummary, duration, mw, streaming)
 		})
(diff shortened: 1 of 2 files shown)


tamirms (approved):

services/horizon/internal/httpx/middleware.go
```diff
@@ -184,7 +159,7 @@ func logEndOfRequest(ctx context.Context, r *http.Request, requestDurationSummar
 		"status":         mw.Status(),
 		"streaming":      streaming,
 		"referer":        referer,
-	}).Info("Finished request")
+	}).Info("HTTP request")
```
leighmcculloch, line 162: ❔ Is changing this message valuable? You'll have a period of time where you'll need to filter by `Finished requests` or `HTTP request` to find all requests logs unless there is some other field you can filter on. If changing this is superficial it might be simpler to keep the existing log message.

leighmcculloch (approved):
🎉 One question (❔) and some suggestions (💡), but LGTM.

💡 The title of this should probably be prefixed with `services/horizon: `, as this is not changing this behavior for all services in the repo.

💡 Could you make the same change to the support packages that use the same pattern so that all our services follow the same pattern?

services/horizon/internal/httpx/middleware.go
bartekn: That's a good question and I don't have a good answer. Both has pros a cons but I think keeping the message the same is better. Going to changes this in a following commit.

bartekn:
> Could you make the same change to the support packages that use the same pattern so that all our services follow the same pattern?

We're not using this in Horizon so I'd rather not change it. Also, it looks like in `support/http` starting and finished requests logs have different set of fields.

PR #3414, opened 2021-02-18
services/horizon: Add metric with local latest ledger

bartekn:
### What

Adds a new metric: `horizon_ingest_local_latest_ledger` that exposes the local latest ledger sequence which is the last ledger ingested by this ingesting instance.

### Why

While checking metrics after Horizon 2.0.0 RC deployment I noticed that existing metric: `horizon_history_latest_ledger` can be confusing because it loads the ledger sequence from DB. However, it's possible that one of ingesting instances is out of sync and is waiting for Stellar-Core sync. It would be great to be able to measure the delay in ledgers and a new metric allows this.

### DIFF
--- a/services/horizon/internal/ingest/fsm.go
+++ b/services/horizon/internal/ingest/fsm.go
@@ -362,6 +362,8 @@ func (r resumeState) run(s *system) (transition, error) {
 		return start(), errors.New("unexpected latestSuccessfullyProcessedLedger value")
 	}
 
+	s.metrics.LocalLatestLedger.Set(float64(r.latestSuccessfullyProcessedLedger))
+
 	if err := s.historyQ.Begin(); err != nil {
 		return retryResume(r),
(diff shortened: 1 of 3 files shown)


tamirms:
> However, it's possible that one of ingesting instances is out of sync and is waiting for Stellar-Core sync. It would be great to be able to measure the delay in ledgers and a new metric allows this.

Could we observe this from the `horizon_stellar_core_latest_ledger` metric?

bartekn:
> Could we observe this from the `horizon_stellar_core_latest_ledger` metric?

This won't always be accurate:
* This metric value is fetched from the core defined by `stellar-core-url` which in some deployments is not captive stellar-core.
* Ledgers are streamed via `bufferedLedgerMetaReader` so it's possible that the actual processed ledger is behind `horizon_stellar_core_latest_ledger` (up to 20 ledgers).

tamirms (approved):

PR #3415, opened 2021-02-18
Merge cap35 into master

2opremio:


.circleci/config.yml
```diff
@@ -150,7 +150,8 @@ commands:
           command: |
             sudo wget -qO - https://apt.stellar.org/SDF.asc | APT_KEY_DONT_WARN_ON_DANGEROUS_USAGE=true sudo apt-key add -
             sudo bash -c 'echo "deb https://apt.stellar.org xenial unstable" > /etc/apt/sources.list.d/SDF-unstable.list'
-            sudo apt-get update && sudo apt-get install -y stellar-core
+            # TODO: replace by the stable version when ready
```
bartekn, line 153: I wonder if we should actually create a new job that tests CAP-35 using a custom core.

bartekn (approved):

.circleci/config.yml
2opremio: Good catch, I don't think we should merge until that is resolved

PR #3416, opened 2021-02-18
historyarchive: Fix bugs in GetLedgers()

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This commit fixes a few bugs found in the `GetLedgers()` function of the history archive client. It also includes a test case which should prevent future regressions.

### Known limitations

[N/A]

### DIFF
--- a/historyarchive/archive.go
+++ b/historyarchive/archive.go
@@ -206,10 +206,11 @@ func (a *Archive) GetLedgers(start, end uint32) (map[uint32]*Ledger, error) {
 	if start > end {
 		return nil, errors.Errorf("range is invalid, start: %d end: %d", start, end)
 	}
-	checkpointRange := a.GetCheckpointManager().MakeRange(start, end)
+	startCheckpoint := a.GetCheckpointManager().GetCheckpoint(start)
+	endCheckpoint := a.GetCheckpointManager().GetCheckpoint(end)
 	cache := map[uint32]*Ledger{}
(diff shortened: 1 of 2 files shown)


historyarchive/archive.go
```diff
@@ -206,10 +206,11 @@ func (a *Archive) GetLedgers(start, end uint32) (map[uint32]*Ledger, error) {
 	if start > end {
 		return nil, errors.Errorf("range is invalid, start: %d end: %d", start, end)
 	}
-	checkpointRange := a.GetCheckpointManager().MakeRange(start, end)
+	startCheckpoint := a.GetCheckpointManager().GetCheckpoint(start)
+	endCheckpoint := a.GetCheckpointManager().GetCheckpoint(end)
 	cache := map[uint32]*Ledger{}
-	for cur := checkpointRange.Low; cur <= checkpointRange.High; cur = a.GetCheckpointManager().NextCheckpoint(cur) {
```
bartekn, line 211: Why didn't previous method work? Looking at `MakeRange` it should work.

bartekn (approved):

historyarchive/archive.go
tamirms: The low part of the range is actually too low. it starts the range at the checkpoint before `start` instead of the checkpoint which includes `start`. For example if `start` was 1000 then `MakeRange()` would begin from checkpoint 959 instead of checkpoint 1023

PR #3418, opened 2021-02-19
services/horizon: Add Stellar-Core sync metrics

bartekn:
### What

This commit adds two new metrics connected to Stellar-Core:
* `horizon_ ingest_captive_stellar_core_synced` determines if Captive Stellar-Core instance started by ingesting server is synced.
* `horizon_stellar_core_synced` determines if Stellar-Core defined by `--stellar-core-url` is synced.

### Why

We want to detect when connected Stellar-Core is out of sync without having to check Stellar-Core metrics.

### Known limitations

The reason for adding two separate metrics is that in some deployments Stellar-Core connected via `--stellar-core-url` and Captive Stellar-Core are separate instances.

### DIFF
--- a/services/horizon/internal/app.go
+++ b/services/horizon/internal/app.go
@@ -38,6 +38,7 @@ type coreSettingsStore struct {
 func (c *coreSettingsStore) set(resp *proto.InfoResponse) {
 	c.Lock()
 	defer c.Unlock()
+	c.Synced = resp.IsSynced()
 	c.CoreVersion = resp.Info.Build
 	c.CurrentProtocolVersion = int32(resp.Info.Ledger.Version)
 	c.CoreSupportedProtocolVersion = int32(resp.Info.ProtocolVersion)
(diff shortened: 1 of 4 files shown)


services/horizon/internal/app.go
```diff
@@ -80,6 +81,7 @@ type App struct {
 	dbWaitCountCounter                prometheus.CounterFunc
 	dbWaitDurationCounter             prometheus.CounterFunc
 	coreLatestLedgerCounter           prometheus.CounterFunc
+	coreSynced                        prometheus.GaugeFunc
```
Shaptic, line 84: Even though there are two deployment methods, aren't they mutually exclusive? You can't have both Captive Core and pass `--stellar-core-url`, as far as I know. Can we have a single field that uses the appropriate metric internally?

tamirms: when horizon runs with captive core enabled it sets the stellar core url to the captive core endpoint on localhost so I don't think there needs to be two different metrics

bartekn: This is true when `--stellar-core-url` is not set, however if set it can be a different URL than captive on localhost. This is the case in which ingesting instance is also serving requests but it using external core too (to submit transactions). This is how Horizon is deployed to testnet at SDF.

tamirms (approved):

PR #3423, opened 2021-02-23
[CAP 35] Add AccountCredited effect to ClawbackCB operation

2opremio:
Part of #3348 

Leftover from https://github.com/stellar/go/pull/3358#discussion_r561784015 (CC @tamirms )

### DIFF
--- a/services/horizon/internal/ingest/processors/effects_processor.go
+++ b/services/horizon/internal/ingest/processors/effects_processor.go
@@ -209,7 +209,7 @@ func (operation *transactionOperationWrapper) effects() ([]effect, error) {
 	case xdr.OperationTypeClawback:
 		err = wrapper.addClawbackEffects()
 	case xdr.OperationTypeClawbackClaimableBalance:
-		err = wrapper.addClawbackClaimableBalanceEffects()
+		err = wrapper.addClawbackClaimableBalanceEffects(changes)
 	case xdr.OperationTypeSetTrustLineFlags:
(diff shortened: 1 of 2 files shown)


services/horizon/internal/ingest/processors/effects_processor.go
```diff
@@ -986,6 +988,22 @@ func (e *effectsWrapper) addClawbackClaimableBalanceEffects() error {
 		history.EffectClaimableBalanceClawedBack,
 		details,
 	)
+
+	// Generate the account credited effect (although the funds will be burned) for the asset issuer
+	for _, c := range changes {
+		if c.Type == xdr.LedgerEntryTypeClaimableBalance && c.Post == nil && c.Pre != nil {
+			cb := c.Pre.Data.ClaimableBalance
+			addAssetDetails(details, cb.Asset, "")
+			details := map[string]interface{}{"amount": amount.String(cb.Amount)}
```
bartekn, line 997: Shouldn't these two lines be reversed? Can you add the check to the test?
```suggestion
			details := map[string]interface{}{"amount": amount.String(cb.Amount)}
			addAssetDetails(details, cb.Asset, "")
```

2opremio: Really good catch. I would had hoped this to get caught by the static checker

2opremio:
@bartekn PTAL

bartekn (approved):

PR #3424, opened 2021-02-24
[CAP 35] Add happy-path test for SetTrustLineFlags operation

2opremio:
Part of #3348

### DIFF
--- a/services/horizon/internal/codes/main.go
+++ b/services/horizon/internal/codes/main.go
@@ -442,7 +442,7 @@ func String(code interface{}) (string, error) {
 	return "", errors.New(ErrUnknownCode)
 }
 
-// ForOperationResult returns the strong represtation used by horizon for the
+// ForOperationResult returns the strong representation used by horizon for the
 // error code `opr`
 func ForOperationResult(opr xdr.OperationResult) (string, error) {
 	if opr.Code != xdr.OperationResultCodeOpInner {
@@ -491,6 +491,12 @@ func ForOperationResult(opr xdr.OperationResult) (string, error) {
 		ic = ir.MustEndSponsoringFutureReservesResult().Code
 	case xdr.OperationTypeRevokeSponsorship:
 		ic = ir.MustRevokeSponsorshipResult().Code
+	case xdr.OperationTypeClawback:
+		ic = ir.MustClawbackResult().Code
+	case xdr.OperationTypeClawbackClaimableBalance:
+		ic = ir.MustClawbackClaimableBalanceResult().Code
+	case xdr.OperationTypeSetTrustLineFlags:
+		ic = ir.MustSetTrustLineFlagsResult().Code
 	}
 
 	return String(ic)
--- a/services/horizon/internal/integration/protocol16_test.go
+++ b/services/horizon/internal/integration/protocol16_test.go
@@ -63,10 +63,7 @@ func TestHappyClawback(t *testing.T) {
 
 	// Make sure the clawback flag was set
 
-	accountDetails, err := itest.Client().AccountDetail(horizonclient.AccountRequest{
-		AccountID: master.Address(),
-	})
-	tt.NoError(err)
+	accountDetails := itest.MustGetAccount(master)
 	tt.True(accountDetails.Flags.AuthClawbackEnabled)
 
 	// Create another account from which to claw an asset back
@@ -87,10 +84,7 @@ func TestHappyClawback(t *testing.T) {
 	}
 	itest.MustSubmitOperations(itest.MasterAccount(), master, &pesetasPayment)
 
-	accountDetails, err = itest.Client().AccountDetail(horizonclient.AccountRequest{
-		AccountID: accountKeyPair.Address(),
-	})
-	tt.NoError(err)
+	accountDetails = itest.MustGetAccount(accountKeyPair)
 	if tt.Len(accountDetails.Balances, 2) {
 		pts := accountDetails.Balances[0]
 		tt.Equal("PTS", pts.Code)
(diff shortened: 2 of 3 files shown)


services/horizon/internal/integration/protocol16_test.go
```diff
@@ -256,3 +258,94 @@ func TestHappyClawbackClaimableBalance(t *testing.T) {
...
+		SetFlags: []txnbuild.TrustLineFlag{
+			txnbuild.TrustLineClawbackEnabled,
+		},
+	}
+	_, err = itest.SubmitOperations(itest.MasterAccount(), master, &setTrustlineFlags)
+	if tt.Error(err) {
+		clientErr, ok := err.(*horizonclient.Error)
+		if tt.True(ok) {
+			tt.Equal(400, clientErr.Problem.Status)
+			tt.Equal(codes.OpMalformed, clientErr.Problem.Extras["result_codes"].(map[string]interface{})["operations"].([]interface{})[0].(string))
```
2opremio, line 346: I wonder if there's a better way to do this (without all the castings)

services/horizon/internal/codes/main.go
```diff
@@ -491,6 +491,12 @@ func ForOperationResult(opr xdr.OperationResult) (string, error) {
 		ic = ir.MustEndSponsoringFutureReservesResult().Code
 	case xdr.OperationTypeRevokeSponsorship:
 		ic = ir.MustRevokeSponsorshipResult().Code
+	case xdr.OperationTypeClawback:
+		ic = ir.MustClawbackResult().Code
+	case xdr.OperationTypeClawbackClaimableBalance:
+		ic = ir.MustClawbackClaimableBalanceResult().Code
+	case xdr.OperationTypeSetTrustLineFlags:
+		ic = ir.MustSetTrustLineFlagsResult().Code
```
2opremio, line 499: This could had gone easily unnoticed (I just happened to test against the opcode).

We have a few functions in which we return a default value if the type value isn't found. Instead I think we should have something like `MaxKnownOperationType`, which we change when new operations are added (`MaxKnownOperationType` would had been updated to  `xdr.OperationTypeSetTrustLineFlags` when changing the for CAP 35). In that way we would had been able to do:

```suggestion
	  case ir.Type <= MaxKnownOperationType:
	    return "", errors.New("internal error: not implemented!")
```

Then again, we could also forget to bump `MaxKnownOperationType` when adding new operations 🤷 

Thoughts?

services/horizon/internal/integration/protocol16_test.go
bartekn, line 346: I think it should work:
```suggestion
			tt.Equal(codes.OpMalformed, clientErr.Problem.ResultCodes().OperationCodes[0])
```

bartekn (approved):

services/horizon/internal/codes/main.go
bartekn: Why not `default:` case? We use it in other places, we should here too.

services/horizon/internal/integration/protocol16_test.go
```diff
@@ -256,3 +258,94 @@ func TestHappyClawbackClaimableBalance(t *testing.T) {
...
+			txnbuild.TrustLineClawbackEnabled,
+		},
+	}
+	itest.MustSubmitOperations(itest.MasterAccount(), master, &setTrustlineFlags)
+
+	// make sure it was cleared
+	accountDetails, err = itest.Client().AccountDetail(horizonclient.AccountRequest{
+		AccountID: accountKeyPair.Address(),
+	})
+	tt.NoError(err)
```
Shaptic, line 326:
```suggestion
	accountDetails = itest.MustGetAccount(accountKeyPair)
```

services/horizon/internal/integration/protocol16_test.go
```diff
@@ -256,3 +258,94 @@ func TestHappyClawbackClaimableBalance(t *testing.T) {
...
+	account := accounts[0]
+
+	// Time machine to Spain before Euros were a thing
+	pesetasAsset := txnbuild.CreditAsset{Code: "PTS", Issuer: master.Address()}
+	itest.MustEstablishTrustline(accountKeyPair, account, pesetasAsset)
+	// Confirm that the Trustline has the clawback flag
+	accountDetails, err = itest.Client().AccountDetail(horizonclient.AccountRequest{
+		AccountID: accountKeyPair.Address(),
+	})
+	tt.NoError(err)
```
Shaptic, line 303:
```suggestion
	accountDetails = itest.MustGetAccount(accountKeyPair)
```

services/horizon/internal/integration/protocol16_test.go
```diff
@@ -256,3 +258,94 @@ func TestHappyClawbackClaimableBalance(t *testing.T) {
...
+			txnbuild.AuthClawbackEnabled,
+		},
+	}
+	itest.MustSubmitOperations(itest.MasterAccount(), master, &setClawBackFlag)
+
+	// Make sure the clawback flag was set
+	accountDetails, err := itest.Client().AccountDetail(horizonclient.AccountRequest{
+		AccountID: master.Address(),
+	})
+	tt.NoError(err)
```
Shaptic, line 288:
```suggestion
	accountDetails = itest.MustGetAccount(accountKeyPair)
```

Shaptic (approved):
Just suggesting some minor readability improvements!

services/horizon/internal/codes/main.go
2opremio: I thought there was no default error in order to be forward compatible.

2opremio: But, to be fair, we already error in other places, which will stop ingestion (notably, the stats update processing).

services/horizon/internal/integration/protocol16_test.go
2opremio: Nice! It did require an extra step for error-checking `ResultCodes()` but it's much cleaner.

PR #3428, opened 2021-02-26
services/horizon: Support integration-testing a remote Horizon

2opremio:
Introduce `integration.NewTestForRemoteHorizon()`, which allows integration-testing
a remote Horizon instance. This is useful for testing a Horizon deployment
(e.g. when deploying a new cluster or a new Horizon release).

TODO: add a simple way to run all the integration tests without changing
          the constructor invocation (i.e. s/NewTest()/NewTestForRemoteHorizon()/) of all the tests.

### DIFF
--- a/services/horizon/internal/test/integration/integration.go
+++ b/services/horizon/internal/test/integration/integration.go
@@ -29,7 +29,7 @@ import (
 )
 
 const (
-	NetworkPassphrase           = "Standalone Network ; February 2017"
+	StandaloneNetworkPassphrase = "Standalone Network ; February 2017"
 	stellarCorePostgresPassword = "mysecretpassword"
 	adminPort                   = 6060
 	stellarCorePort             = 11626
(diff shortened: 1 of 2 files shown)


tamirms (approved):

services/horizon/internal/test/integration/integration.go
```diff
@@ -377,7 +391,7 @@ func (i *Test) CreateAccounts(count int, initialBalance string) ([]*keypair.Full
 	request := sdk.AccountRequest{AccountID: master.Address()}
 	account, err := client.AccountDetail(request)
 	if err == nil {
-		seq, err = strconv.ParseInt(account.Sequence, 10, 8) // why is this a string?
+		seq, err = strconv.ParseInt(account.Sequence, 10, 64) // why is this a string?
```
bartekn, line 394: Good catch!

bartekn (approved):

Shaptic:
For the TODO in the description (allow it to be part of `NewTest()`), we could create a `type RemoteConfig` and add it to the `Config` struct, e.g. 

```go
type RemoteConfig struct {
	URL        string
	MasterKey  string
	Passphrase string
}

type Config struct {
	// ...
	Remote *RemoteConfig
}
```

where `nil` means local?

paulbellamy:
Afaict this won't work without rewriting the integration tests, as they make all sorts of assumptions about being able to start/stop core, change the protocol verstion, etc.

Issue #3396, opened 2021-02-09
Add horizon health check endpoint


tamirms:
It would be useful to have a healthcheck endpoint that guarantees horizon is fully operational. Fully operation could be defined as being able to submit transactions to stellar core. Assuming that definition, the health check could query its stellar core (or captive core) instance to see that its synced to the network.

jacekn:
Most load balancers I know are happy with healthcheck endpoints returning http 200 so that will be enough.
But if it's easy it might be useful to also have health info returned in text format so that operators can query it to check status easily. Maybe something like:
```
{
    "ready": false,
    "database_connected": true,
    "core_up": true,
    "core_synced": false
}
```

bartekn:
For the future assignee: adding the code for healtcheck somewhere around https://github.com/stellar/go/pull/3414 and https://github.com/stellar/go/pull/3418 can help implementing this faster.

bartekn:
Also, my 2 cents: while I think this could be really useful I think we should keep in mind how health check are used in many platforms. In ELB for example I think the default is 2 failures will mark an instance as unavailable but then it will require 10 successes to bring it back. This means that a short blip can actually make the instance unavailable for a longer time.

Other than that, it's ok for me.

Issue #3397, opened 2021-02-09
Feature Request: more verbose logging during horizon startup


jacekn:
### What problem does your feature solve?

It makes troubleshooting easier

### What would you like to see?

Currently when horizon starts with database URL that's misconfigured or blocked by a firewall there are no log or console entries at all. It would be useful to add some messages to help with troubleshooting.

For example even before the DB connection is attempted we could log `Starting horizon` to indicate the process is up. If horizon can't write to the logfile it would be good to print error on the console so that it's visible in `systemctl status stellar-horizon` output.
After that `Connecting to database` would be useful. If horizon retries on failures it might be good to print error after each failed attempt.
Finally once it's up and running it would be good to see `Ready to serve requests` message

PR #3435, opened 2021-03-02
services/horizon: Add horizon health check endpoint

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Close https://github.com/stellar/go/issues/3396

This commit adds a health check endpoint which can be used to check if horizon is operational. Fully operation is defined as being able to submit transactions to stellar core and being able to access the Horizon DB. On success the health check responds with a 200 http status code. On failure the health check responds with a 503.

### Why

This endpoint can be queried by load balancers to determine whether traffic should be routed away from a horizon instance if it is unhealthy.

### Known limitations

I think it's worth adding a rule on the ops side so that the endpoint is only accessible internally (e.g. from the loadbalancer or monitoring agents).


jacekn:
> I think it's worth adding a rule on the ops side so that the endpoint is only accessible internally (e.g. from the loadbalancer or monitoring agents).

@tamirms above is a really good point, probably worth further discussion. Maybe `/health` should be exposed on the admin port instead, next to `/metrics`? But at the same time I think there is some value in anyone being able to hit `/health` to confirm the service should be responding. Do you think there is a performance impact if the endpoint is used heavily?

tamirms:
@jacekn 
> Maybe /health should be exposed on the admin port instead, next to /metrics? But at the same time I think there is some value in anyone being able to hit /health to confirm the service should be responding.

yes, I had the same thought and came to the conclusion that it would be better to serve the endpoint out of the same port used by the rest of the horizon API

 > Do you think there is a performance impact if the endpoint is used heavily?

I think it could be possible that stellar core is overwhelmed with info requests if the /health is hit excessively. But there should be rate limiting in place to mitigate that. Still I think it would be make sense to restrict access to /health just to be on the safe side

brahman81:
@tamirms am not sure I understand the motivation of having the `/health` endpoint exposed on the main socket if we are  suggesting the operator restricts and/or rate limit access to it.

Would it be best to expose this internally only on the same socket as `/metrics` ?

tamirms:
@brahman81 the benefit of having `/health` on the main socket is that it would verify that the http server on the main socket is operational and can respond to http requests. imagine if we exposed `/health` on the internal socket and for whatever reason the internal socket was working fine but the main socket was unable to receive requests. In that scenario the health check requests would succeed even though requests to the horizon API would fail.

jacekn:
@tamirms I can see your point about validating the main http server, it's definitely be a good idea to verify it's running.
But at the same time if we say that access to the `/health` endpoint should be restricted by ops then I think we should not expose it by default. It's very easy to miss the fact that restrictions are needed and I'm sure many operators would end up with the endpoint exposed unintentionally.

Maybe we should decide if we consider `/health` safe to be exposed globally without restrictions and work from there?
If we decide it's safe then we expose it by default on the main port and don't require ops to add custom rules.
If we decide it's unsafe I think the safest default behaviour is for `/health` to not be exposed. In this case we can either make it optional via config flag (so that ops must explicitly enable it after adding restrictions) or decide we run on the admin port and accept the risk that the main http server could be dead while the admin port is up.

ire-and-curses:
Can we just rate limit this endpoint in the implementation? e.g. just cache the stellar core response in Horizon with a lifetime of 1s or something?

paulbellamy:
The cache implemented here still has a thundering-herd problem when it expires. It's probably overkill, but the cache lock should be around the whole health-check, so the backend doesn't get a traffic spike every 500ms.

tamirms:
@paulbellamy good catch! would https://github.com/stellar/go/pull/3435/commits/5196749b14e68bd23d56ba0d3e49bd483d3a7dac fix the problem?

paulbellamy:
Yeah that looks ideal, I'd say. You could keep a RWMutex for performance, but probably not worth it. We'll still have the herd problem across multiple horizon instances, but also not worth fixing that now.

services/horizon/internal/health.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+func (h healthCheck) runCheck() healthResponse {
+	response := healthResponse{
+		DatabaseConnected: true,
+		CoreUp:            true,
+		CoreSynced:        true,
+	}
+	if err := h.session.Ping(); err != nil {
+		log.Warnf("could ping db: %s", err)
```
bartekn, line 64:
```suggestion
		log.Warnf("could not ping db: %s", err)
```

services/horizon/internal/health.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+func (h healthCheck) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	response := h.cache.get(h.runCheck)
+
+	if !response.DatabaseConnected || !response.CoreSynced || !response.CoreUp {
+		w.WriteHeader(http.StatusServiceUnavailable)
+	}
+
+	json.NewEncoder(w).Encode(response)
```
bartekn, line 85: We should handle the returned error here just in case.
```suggestion
	err := json.NewEncoder(w).Encode(response)
	if err != nil {
		// ...
	}
```

support/db/session.go
```diff
@@ -294,6 +294,12 @@ func (s *Session) Rollback() error {
 	return err
 }
 
+// Ping verifies a connection to the database is still alive,
+// establishing a connection if necessary.
+func (s *Session) Ping() error {
+	return s.DB.PingContext(s.Ctx)
```
bartekn, line 300: I can see one problem here similar to https://github.com/stellar/go/issues/3397. It's possible that the PING query will hang indefinitely because `s.Ctx` is a Horizon app context and, as a result, can hang health updates. I think we should either create a context with reasonable timeout inside `Ping()` or allow passing `ctx` with timeout via method argument.

services/horizon/internal/health.go
```diff
@@ -0,0 +1,86 @@
...
+		DatabaseConnected: true,
+		CoreUp:            true,
+		CoreSynced:        true,
+	}
+	if err := h.session.Ping(); err != nil {
+		log.Warnf("could ping db: %s", err)
+		response.DatabaseConnected = false
+	}
+	if resp, err := h.core.Info(h.ctx); err != nil {
+		log.Warnf("request to stellar core failed: %s", err)
```
bartekn, line 68: Worth calling `WithField` with a descriptive tag to understand this is happening in health checks. Same with `Warnf` above.

services/horizon/internal/health.go
tamirms: how should we handle it?

bartekn: Just log it and return it `500` HTTP error.

services/horizon/internal/health.go
```diff
@@ -60,12 +66,12 @@ func (h healthCheck) runCheck() healthResponse {
 		CoreUp:            true,
 		CoreSynced:        true,
 	}
-	if err := h.session.Ping(); err != nil {
-		log.Warnf("could ping db: %s", err)
+	if err := h.session.Ping(dbPingTimeout); err != nil {
+		log.WithField("component", "healthCheck").Warnf("could not ping db: %s", err)
```
bartekn, line 70: Nit: Maybe `service` will be better than `component`. We use service in other places like `ingest` or `txsub`.

services/horizon/internal/health.go
tamirms: in some cases we won't be able to respond with a 500. For example, if we already invoked `w.WriteHeader(http.StatusServiceUnavailable)` we cannot call `w.WriteHeader()` again to change the response code. 

`json.NewEncoder(w).Encode(response)` could also return an error for potentially 2 reasons: 

1) the json serialization failed 
2) writing to the `http.ResponseWriter` failed at some point

If case 2 occurs then it is possible we won't be able to respond with a 500 because the response writer will set the status to 200 before writing anything unless `w.WriteHeader()` has already been called

should we try to account for all these cases to try and respond with a 500 while avoiding multiple response.WriteHeader calls? or is just logging the error sufficient?

bartekn: Good points, I think it's OK if we just log the error. My main concern was ignoring the return error.

services/horizon/internal/health.go
```diff
@@ -67,11 +67,11 @@ func (h healthCheck) runCheck() healthResponse {
 		CoreSynced:        true,
 	}
 	if err := h.session.Ping(dbPingTimeout); err != nil {
-		log.WithField("component", "healthCheck").Warnf("could not ping db: %s", err)
+		log.WithField("service", "healthCheck").Warnf("could not ping db: %s", err)
```
bartekn, line 70: Last NIT: you can create a local log with field and reuse it later to avoid duplication:
```go
localLog := log.WithField("service", "healthCheck")
// ...
localLog.Warnf("could not ping db: %s", err)
```

bartekn (approved):

Issue #3447, opened 2021-03-04
[CAP 35] Improve integration test coverage


2opremio:
Right now we only have happy-path coverage (and for `SetTrustlineFlags` we still need to test the details and effects).

We should add more tests, for instance, the tests suggested by @leighmcculloch at https://github.com/stellar/go/pull/3368#discussion_r562827974

Part of #3348

paulbellamy:
From reading the comment on #3368, and [cap35](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0035.md), the list of extra potential tests I have is below. Not sure we need to hit *all* of these, as a lot should be handled by core.

## Happy path
  - [ ] Claiming back a partial amount.

## Sad Path
  - [ ] enabling and disabling it (tested implicitly by below?)
    - [ ] trying to set AUTH_CLAWBACK_ENABLED_FLAG after trustline created
  - [ ] set options should fail to set clawback enabled when not revocable
  - [ ] low threshold signature
  - [ ] negative amount
  - [ ] from == issuer
    - [ ] Shouldn't be able to clawback yourself/
  - [ ] invalid asset
  - [ ] claimable balance
    - [ ] already claimed
    - [ ] already cancelled?
    - [ ] source account not clawbackable
    - [ ] source account not allowed to clawback
    - [ ] source account is not issuer
    - [ ] created when clawback not enabled on issuer account, then clawback
        enabled on issuer account after. should still fail.
    - [ ] can't clawback it twice.
  - [ ] accounts
    - [ ] clawback not enabled on from account
    - [ ] clawback not enabled on issuer account
    - [ ] clawback enabled, but revocable not enalbed
    - [ ] trustline source != issuer (Is this a scenario?)
    - [ ] clawback should still work when holder is unauthorized
    - [ ] clawback should still work when holder is authorized_to_maintain_liabilities
    - [ ] from account has no trustline for asset
    - [ ] from account does not hold asset
    - [ ] from account insufficient funds
    - [ ] from account sufficient funds including selling liabilities
      - [ ] should fail unless issuer deauthorizes/reauthorizes the from account
      - [ ] deauthorize/reauthorize should work

paulbellamy:
Per discussion, for stuff where it is behaviour around core, it makes more sense to maybe sanity-check the error response gets generated correctly, and the horizon state is in the expected state. Possibly a table-driven test makes sense here.

PR #3451, opened 2021-03-05
[CAP 35] Improve SetTrustLineFlags integration testing

2opremio:
Part of https://github.com/stellar/go/issues/3447

### DIFF
--- a/services/horizon/internal/integration/protocol16_test.go
+++ b/services/horizon/internal/integration/protocol16_test.go
@@ -11,6 +11,7 @@ import (
 	"github.com/stellar/go/services/horizon/internal/codes"
 	"github.com/stellar/go/services/horizon/internal/test/integration"
 	"github.com/stellar/go/txnbuild"
+	"github.com/stellar/go/xdr"
 )
 
 var protocol16Config = integration.Config{ProtocolVersion: 16}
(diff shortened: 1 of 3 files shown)


services/horizon/internal/integration/protocol16_test.go
```diff
@@ -309,6 +310,44 @@ func TestHappySetTrustLineFlags(t *testing.T) {
...
+	if tt.Len(opDetailsResponse.Embedded.Records, 1) {
+		setTrustlineFlagsDetails := opDetailsResponse.Embedded.Records[0].(operations.SetTrustLineFlags)
+		tt.Equal("PTS", setTrustlineFlagsDetails.Code)
+		tt.Equal(master.Address(), setTrustlineFlagsDetails.Issuer)
+		tt.Equal(accountKeyPair.Address(), setTrustlineFlagsDetails.Trustor)
+		if tt.Len(setTrustlineFlagsDetails.ClearFlags, 1) {
+			tt.True(xdr.TrustLineFlags(setTrustlineFlagsDetails.ClearFlags[0]).IsClawbackEnabledFlag())
+		}
+		if tt.Len(setTrustlineFlagsDetails.ClearFlagsS, 1) {
+			tt.Equal(setTrustlineFlagsDetails.ClearFlagsS[0], "clawback_enabled")
```
2opremio, line 327: I wonder if the Go SDK should have a symbol for this (I don't think we do for the `SetOptions` operation, which has a very similar structure though)

bartekn (approved):

services/horizon/internal/integration/protocol16_test.go
bartekn: Sounds good. Maybe we should track it in a separate issue?

2opremio: Yep, good point, I will do that

Issue #3241, opened 2020-11-20
Feature Request: improve the assets endpoint


howardtw:
### What problem does your feature solve?

For regulated assets, the issuers often want to track the assets the've issued/distributed. However, regulated asset holders will have their account set to `UNAUTHORZIED` or `AUTHORIZED_TO_MAINTAIN_LIABILITIES` when they are not transacting the asset. And it seems like the assets endpoint doesn't return the assets in accounts with the flag to those two values.

### What would you like to see?

The response from the assets endpoint includes all assets regardless the authorization state in asset holders' accounts.

While there an [issue](https://github.com/stellar/go/issues/3240) targeted to address it, there are more things we can consider when we implement a solution.

Currently the response includes the flags in the issuer account, and those flags can be set with `set_options` ops in the issuer's account. There are a few things to consider regarding it:

1. The flags in the response only reflect the current state of the issuing accounts. However,  the `authorization_required ` flag in the issuer's account will only determine the default state of the trustline at the time when a holder creates one, and it will not affect existing trustlines. In other words, if an issuer issues some of the asset with `authorization_required` to set `false` and issues some of the assset with `authorization_required` to set `true`, I'm not sure whether it makes sense to add all accounts with different default states of the trustline together and display them under the current state of the issuer's account.
2. Even if we decide to pursue point above by categorizing the response by the default state of the trustline, we are only looking at the response from the issuer's standpoint. It will be useful if we can categorize the response by the asset holder's account state as well. The state of asset holders' accounts can be set by `allow_trust` ops. Currently, there are three authorization states an asset holder can have, `AUTHORIZED`, `AUTHORIZED_TO_MAINTAIN_LIABILITIES`, and `UNAUTHORIZED`. And the assets endpoint only counts the accounts in the `AUTHORIZED` state.

I understand there could be several combinations (24 at maximum, with 8 from the first point and 3 from the second point) if we do both of the above. If we can only pick one, it would think the second is more important than the first.

### What alternatives are there?

N/A

bartekn:
I think it is fixed in https://github.com/stellar/go/pull/3454. Will be part of Horizon 2.1.0 release.

PR #3454, opened 2021-03-08
services/horizon: Issue #3241, Include unauthorized assets on /assets

paulbellamy:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Include unauthorized assets on the /assets endpoint, and expose the trustline flags.

### Why

As per issue #3241.

### Known limitations

[TODO or N/A]

### DIFF
--- a/protocols/horizon/main.go
+++ b/protocols/horizon/main.go
@@ -193,6 +193,18 @@ type AssetStatAccounts struct {
 	Unauthorized                    int32 `json:"unauthorized"`
 }
 
+func (a AssetStatAccounts) Add(b AssetStatAccounts) AssetStatAccounts {
+	return AssetStatAccounts{
+		Authorized:                      a.Authorized + b.Authorized,
+		AuthorizedToMaintainLiabilities: a.AuthorizedToMaintainLiabilities + b.AuthorizedToMaintainLiabilities,
+		Unauthorized:                    a.Unauthorized + b.Unauthorized,
+	}
+}
+
+func (a AssetStatAccounts) Sum() int32 {
+	return a.Authorized + a.AuthorizedToMaintainLiabilities + a.Unauthorized
+}
+
 // Balance represents an account's holdings for a single currency type
 type Balance struct {
 	Balance                           string `json:"balance"`
--- a/protocols/horizon/main.go
+++ b/protocols/horizon/main.go
@@ -168,9 +168,9 @@ type AssetStat struct {
 	base.Asset
 	PT          string            `json:"paging_token"`
 	Accounts    AssetStatAccounts `json:"accounts"`
-	Amount      string            `json:"amount"`
+	Amount      string            `json:"amount"` // Deprecated
 	Balances    AssetStatBalances `json:"balances"`
-	NumAccounts int32             `json:"num_accounts"`
+	NumAccounts int32             `json:"num_accounts"` // Deprecated
 	Flags       AccountFlags      `json:"flags"`
 }
 
--- a/protocols/horizon/main.go
+++ b/protocols/horizon/main.go
@@ -166,12 +166,14 @@ type AssetStat struct {
 	} `json:"_links"`
 
 	base.Asset
-	PT          string            `json:"paging_token"`
-	Accounts    AssetStatAccounts `json:"accounts"`
-	Amount      string            `json:"amount"` // Deprecated
-	Balances    AssetStatBalances `json:"balances"`
-	NumAccounts int32             `json:"num_accounts"` // Deprecated
-	Flags       AccountFlags      `json:"flags"`
+	PT       string            `json:"paging_token"`
+	Accounts AssetStatAccounts `json:"accounts"`
+	// Action needed in release: horizon-v3.0.0: deprecated field
+	Amount   string            `json:"amount"`
+	Balances AssetStatBalances `json:"balances"`
+	// Action needed in release: horizon-v3.0.0: deprecated field
+	NumAccounts int32        `json:"num_accounts"`
+	Flags       AccountFlags `json:"flags"`
 }
 
 // PagingToken implementation for hal.Pageable
--- a/protocols/horizon/main.go
+++ b/protocols/horizon/main.go
@@ -193,18 +193,6 @@ type AssetStatAccounts struct {
 	Unauthorized                    int32 `json:"unauthorized"`
 }
 
-func (a AssetStatAccounts) Add(b AssetStatAccounts) AssetStatAccounts {
-	return AssetStatAccounts{
-		Authorized:                      a.Authorized + b.Authorized,
-		AuthorizedToMaintainLiabilities: a.AuthorizedToMaintainLiabilities + b.AuthorizedToMaintainLiabilities,
-		Unauthorized:                    a.Unauthorized + b.Unauthorized,
-	}
-}
-
-func (a AssetStatAccounts) Sum() int32 {
-	return a.Authorized + a.AuthorizedToMaintainLiabilities + a.Unauthorized
-}
-
 // Balance represents an account's holdings for a single currency type
 type Balance struct {
 	Balance                           string `json:"balance"`
--- a/protocols/horizon/main.go
+++ b/protocols/horizon/main.go
@@ -166,15 +166,31 @@ type AssetStat struct {
 	} `json:"_links"`
 
 	base.Asset
-	PT          string       `json:"paging_token"`
-	Amount      string       `json:"amount"`
-	NumAccounts int32        `json:"num_accounts"`
-	Flags       AccountFlags `json:"flags"`
+	PT          string               `json:"paging_token"`
+	Accounts    AssetStatNumAccounts `json:"accounts"`
+	Amount      string               `json:"amount"`
+	Balances    AssetStatBalances    `json:"balances"`
+	NumAccounts int32                `json:"num_accounts"`
+	Flags       AccountFlags         `json:"flags"`
 }
 
 // PagingToken implementation for hal.Pageable
 func (res AssetStat) PagingToken() string {
 	return res.PT
+}
+
+// AssetStatBalances represents the summarized balances for a single Asset
+type AssetStatBalances struct {
+	Authorized                      string `json:"authorized"`
+	AuthorizedToMaintainLiabilities string `json:"authorized_to_maintain_liabilities"`
+	Unauthorized                    string `json:"unauthorized"`
+}
+
+// AssetStatNumAccounts represents the summarized acount numbers for a single Asset
+type AssetStatNumAccounts struct {
+	Authorized                      int32 `json:"authorized"`
+	AuthorizedToMaintainLiabilities int32 `json:"authorized_to_maintain_liabilities"`
+	Unauthorized                    int32 `json:"unauthorized"`
 }
 
 // Balance represents an account's holdings for a single currency type
--- a/protocols/horizon/main.go
+++ b/protocols/horizon/main.go
@@ -166,12 +166,12 @@ type AssetStat struct {
 	} `json:"_links"`
 
 	base.Asset
-	PT          string               `json:"paging_token"`
-	Accounts    AssetStatNumAccounts `json:"accounts"`
-	Amount      string               `json:"amount"`
-	Balances    AssetStatBalances    `json:"balances"`
-	NumAccounts int32                `json:"num_accounts"`
-	Flags       AccountFlags         `json:"flags"`
+	PT          string            `json:"paging_token"`
+	Accounts    AssetStatAccounts `json:"accounts"`
+	Amount      string            `json:"amount"`
+	Balances    AssetStatBalances `json:"balances"`
+	NumAccounts int32             `json:"num_accounts"`
+	Flags       AccountFlags      `json:"flags"`
 }
 
 // PagingToken implementation for hal.Pageable
@@ -186,8 +186,8 @@ type AssetStatBalances struct {
 	Unauthorized                    string `json:"unauthorized"`
 }
 
-// AssetStatNumAccounts represents the summarized acount numbers for a single Asset
-type AssetStatNumAccounts struct {
+// AssetStatAccounts represents the summarized acount numbers for a single Asset
+type AssetStatAccounts struct {
 	Authorized                      int32 `json:"authorized"`
 	AuthorizedToMaintainLiabilities int32 `json:"authorized_to_maintain_liabilities"`
 	Unauthorized                    int32 `json:"unauthorized"`
--- a/services/horizon/internal/db2/history/main.go
+++ b/services/horizon/internal/db2/history/main.go
@@ -4,6 +4,8 @@ package history
 
 import (
 	"database/sql"
+	"database/sql/driver"
+	"encoding/json"
 	"fmt"
 	"sync"
 	"time"
@@ -12,7 +14,6 @@ import (
 	"github.com/guregu/null"
 	"github.com/jmoiron/sqlx"
 
-	"github.com/stellar/go/protocols/horizon"
 	"github.com/stellar/go/services/horizon/internal/db2"
 	"github.com/stellar/go/support/db"
 	"github.com/stellar/go/support/errors"
@@ -322,13 +323,13 @@ type Asset struct {
 
 // ExpAssetStat is a row in the exp_asset_stats table representing the stats per Asset
 type ExpAssetStat struct {
-	AssetType   xdr.AssetType             `db:"asset_type"`
-	AssetCode   string                    `db:"asset_code"`
-	AssetIssuer string                    `db:"asset_issuer"`
-	Accounts    horizon.AssetStatAccounts `db:"accounts"`
-	Balances    horizon.AssetStatBalances `db:"balances"`
-	Amount      string                    `db:"amount"`
-	NumAccounts int32                     `db:"num_accounts"`
+	AssetType   xdr.AssetType        `db:"asset_type"`
+	AssetCode   string               `db:"asset_code"`
+	AssetIssuer string               `db:"asset_issuer"`
+	Accounts    ExpAssetStatAccounts `db:"accounts"`
+	Balances    ExpAssetStatBalances `db:"balances"`
+	Amount      string               `db:"amount"`
+	NumAccounts int32                `db:"num_accounts"`
 }
 
 // PagingToken returns a cursor for this asset stat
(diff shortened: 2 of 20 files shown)


protocols/horizon/main.go
```diff
@@ -142,9 +142,11 @@ func (a *Account) SignerSummary() map[string]int32 {
...
 // AccountFlags represents the state of an account's flags
 type AccountFlags struct {
-	AuthRequired  bool `json:"auth_required"`
-	AuthRevocable bool `json:"auth_revocable"`
-	AuthImmutable bool `json:"auth_immutable"`
+	AuthRequired                      bool `json:"auth_required"`
+	AuthRevocable                     bool `json:"auth_revocable"`
+	AuthImmutable                     bool `json:"auth_immutable"`
+	IsAuthorized                      bool `json:"is_authorized,omitempty"`
+	IsAuthorizedToMaintainLiabilities bool `json:"is_authorized_to_maintain_liabilities,omitempty"`
```
bartekn, line 149: I think it's not necessary because `Balance` struct already contains new flags.

services/horizon/internal/ingest/processors/asset_stats_processor.go
```diff
@@ -206,46 +207,27 @@ func (p *AssetStatsProcessor) adjustAssetStat(
...
+		// removing a trustline
+		trustline = *preTrustline
+		deltaAccounts = -1
+		deltaBalance = -preTrustline.Balance
+	default:
 		return ingest.NewStateError(errors.New("both pre and post trustlines cannot be nil"))
 	}
 
-	err := p.assetStatSet.AddDelta(trustline.Asset, int64(deltaBalance), deltaAccounts)
+	err := p.assetStatSet.AddDelta(trustline.Asset, int64(deltaBalance), deltaAccounts, xdr.TrustLineFlags(trustline.Flags))
```
bartekn, line 230: I think you need to add two deltas in case flags are changed. say, somebody edited flag from authorized to unauthorized. You need to decrease authorized stats and then increase unauthorized stats.

services/horizon/internal/ingest/main.go
```diff
@@ -49,7 +49,8 @@ const (
 	// - 11: Protocol 14: CAP-23 and CAP-33.
 	// - 12: Trigger state rebuild due to `absTime` -> `abs_time` rename
 	//       in ClaimableBalances predicates.
-	CurrentVersion = 12
+	// - 13: Trigger state rebuild to include more than just authorized assets.
+	CurrentVersion = 13
```
bartekn, line 53: We also need to update state verifier.

services/horizon/internal/ingest/processors/asset_stats_processor.go
paulbellamy: Wait, do we want to have separate stats for each variant of the trustlineflags?

bartekn: Copying a few points from our discussion on Slack:
* Right now all trust lines by a given issuer inherit the authorization flag from account flags. However it will soon change when [CAP-35](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0035.md) is live.
* Even now, it's possible that if trust lines are unauthorized by default they can authorized with `allow_trust` operation (and reverse authorization). Then we need to update both rows given the current schema. (@paulbellamy brought a very good point about API design and is going to post it here so we can discuss if we want to stay with the current schema change in this PR)

paulbellamy: Basically it's how we want to incorporate the trustlines into the `/assets` endpoint. Options that jump to mind immediately are:

1.  Summarized stats at the asset level:
      ```js
      {
        // ... existing asset fields
        "num_accounts": {
          "authorized": 9390,
          "authorized_to_maintain_liabilities": 2408,
        },
        "balances": {
          "authorized": "1347404.4083346",
          "authorized_to_maintain_liabilities": "0.4083346",
        }
      }
      ```
      - Pros:
        - Concise and extensible.
        - Easy to query and render
      - Cons:
        - Assumes the different asset balances map cleanly to the trustline flags.
        - Will probably want a different api endpoint to break down balances per-trustline.

1. Breakdown by accounts:
      ```js
      {
        // ... existing asset fields
        "accounts": {
          "id": "GAYOLLLUIZE4DZMBB2ZBKGBUBZLIOYU6XFLW37GBP2VZD3ABNXCW4BVA",
          "flags": {
            "is_authorized": true,
            "is_authorized_to_maintain_liabilities": false,
          }
        }
      }
      ```
      - Pros:
        - More explicit and granular
        - Includes space for balances and other per-trustline metadata
      - Cons:
        - I'm nervous that it would be inefficient to query
        - Will return way too much data.

bartekn: @howardtw could you check the proposed API design options and let us know what would work better for you?

bartekn: @paulbellamy point 2. is no option. We need aggregate data and as you noted this will be inefficient.

Another option we discussed was to have a separate response object for each flag combination, ex.
```json
{
    "asset_type": "credit_alphanum4",
    "asset_code": "USD",
    "asset_issuer": "GDSNPSOVLFI24Y3IAMHTFO6QYZBSLPBNE7Z5U53WZQ6O5ATHR7BOL2WD",
    "paging_token": "USD_GDSNPSOVLFI24Y3IAMHTFO6QYZBSLPBNE7Z5U53WZQ6O5ATHR7BOL2WD_credit_alphanum4",
    "amount": "9980000.0000000",
    "num_accounts": 1,
    "trustline_flags": {
        "authorized": true
    }
},
{
    "asset_type": "credit_alphanum4",
    "asset_code": "USD",
    "asset_issuer": "GDSNPSOVLFI24Y3IAMHTFO6QYZBSLPBNE7Z5U53WZQ6O5ATHR7BOL2WD",
    "paging_token": "USD_GDSNPSOVLFI24Y3IAMHTFO6QYZBSLPBNE7Z5U53WZQ6O5ATHR7BOL2WD_credit_alphanum4",
    "amount": "1000.0000000",
    "num_accounts": 1,
    "trustline_flags": {
        "authorized": false
    }
},
```

howardtw: @paulbellamy the breakdown of the pros and cons is awesome 🔥 

> could you check the proposed API design options and let us know what would work better for you?

It's really up to our partner how they want to track the regulated asset. I think it would be valuable to see how they want to use the endpoint before we make a decision. Since option 2 is no go, I'll just comment on options 1 and [3](https://github.com/stellar/go/pull/3454#discussion_r590229697) as a user. In my view, I prefer option 1 over option 3 given that it has less repetition for the same combination of asset code and asset issuer.

I assume option 1 would not aggregate all assets with the same asset code when querying with just the asset code. In other words, I don't think it would be too helpful to provide the total number of accounts in the `authorized` state holding an asset with a code `USD` over the network. Same for the asset issuer.

In option 1:
```
{
  // ... existing asset fields
  "num_accounts": {
    "authorized": 9390,
    "authorized_to_maintain_liabilities": 2408,
  },
  "balances": {
    "authorized": "1347404.4083346",
    "authorized_to_maintain_liabilities": "0.4083346",
  }
}
```
We are omitting `unauthorized` in the above example, is that correct? 

Do you think it would be valuable to add a query param to include the authorization state of the asset holder's account so that people can do `/assets?asset_code=USD&asset_issuer=GDSNPSOVLFI24Y3IAMHTFO6QYZBSLPBNE7Z5U53WZQ6O5ATHR7BOL2WD&authorization_state=authorized_to_maintain_liabilities"`?

howardtw: @paulbellamy @bartekn our partner got back to us saying
```
the proposed solution looks perfect for me, total balances for an asset with breakdown per flags is very informative.
```

paulbellamy: > I assume option 1 would not aggregate all assets with the same asset code when querying with just the asset code.

Agreed, yeah. That makes sense.

> We are omitting unauthorized in the above example, is that correct?

Apologies, that was a mistake when I was typing out the example. It should be:
```js
{
  // ... existing asset fields
  "num_accounts": {
    "authorized": 9390,
    "authorized_to_maintain_liabilities": 2408,
    "unauthorized": 12
  },
  "balances": {
    "authorized": "1347404.4083346",
    "authorized_to_maintain_liabilities": "0.4083346",
    "unauthorized": "0.123456"
  }
}
```

> Do you think it would be valuable to add a query param to include the authorization state of the asset holder's account so that people can do `/assets?asset_code=USD&asset_issuer=GDSNPSOVLFI24Y3IAMHTFO6QYZBSLPBNE7Z5U53WZQ6O5ATHR7BOL2WD&authorization_state=authorized_to_maintain_liabilities"`?

Will need to have a think about that. If we were returning more verbose information, then yes. But since we are only returning summary stats at `/assets`, then maybe it's not worth it. Wdyt?

paulbellamy: Also, @howardtw, I'm happy to park this for ~a week or so if you want to check with the partner.

howardtw: They got back to us today. See https://github.com/stellar/go/pull/3454#discussion_r591761402 😄

paulbellamy:
Perfect then!

services/horizon/internal/db2/schema/migrations/43_asset_stat_trust_line_flags.sql
```diff
@@ -0,0 +1,16 @@
+-- +migrate Up
+
+ALTER TABLE asset_stats
+ADD COLUMN accounts JSONB;
+ADD COLUMN balances JSONB;
+-- Previously, all the asset_stats we stored were authorized.
+UPDATE asset_stats
+  SET
+    accounts = jsonb_build_object('authorized', num_accounts),
+    balances = jsonb_build_object('authorized', amount);
```
marcelosalloum, line 10: Shouldn't this be `json_build_object` instead of `jsonb_build_object`?

```suggestion
    accounts = json_build_object('authorized', num_accounts),
    balances = json_build_object('authorized', amount);
```

services/horizon/internal/ingest/main.go
marcelosalloum: ❓ When we upgrade this version will horizon re-ingest the balances so the new columns `balances` and `accounts` are correctly populated? Or will the new columns only be populated with data happening after this change?

services/horizon/internal/db2/schema/migrations/43_asset_stat_trust_line_flags.sql
paulbellamy: No, json and jsonb are different (but similar) things in postgres.

services/horizon/internal/ingest/main.go
bartekn: Horizon will reingest entire state automatically.

services/horizon/internal/db2/schema/migrations/43_asset_stat_trust_line_flags.sql
marcelosalloum: Got it, thanks for explaining.

services/horizon/internal/ingest/verify.go
```diff
@@ -27,7 +27,7 @@ const assetStatsBatchSize = 500
 // check them.
 // There is a test that checks it, to fix it: update the actual `verifyState`
 // method instead of just updating this value!
-const stateVerifierExpectedIngestionVersion = 12
```
bartekn, line 30: You also need to update actual state verification code to check if stats for new flags match, this function:
https://github.com/stellar/go/blob/1f586e6d19fe5572825c09f2fa1244cfb438911b/services/horizon/internal/ingest/verify.go#L257-L305

paulbellamy:
Per discussion, for now, I think let's leave out the query param for filtering. The output is not super verbose so it's probably not that useful.

services/horizon/internal/ingest/verify.go
paulbellamy: It uses struct equality to check for differences. So because the new fields are not pointers, it actually JustWorks(tm). But, it's a bit subtle so I've added a test to make sure we don't accidentally break it in future.

bartekn: I see what you did here! OK, will check everything again when PR is ready for review.

services/horizon/internal/ingest/processors/asset_stats_set.go
```diff
@@ -13,9 +13,78 @@ type assetStatKey struct {
...
+
+func (a assetStatBalances) Finish() history.ExpAssetStatBalances {
+	return history.ExpAssetStatBalances{
+		Authorized:                      a.Authorized.String(),
+		AuthorizedToMaintainLiabilities: a.AuthorizedToMaintainLiabilities.String(),
+		Unauthorized:                    a.Unauthorized.String(),
+	}
+}
+
+type assetStatAccounts struct {
```
tamirms, line 71: why not use `history.ExpAssetStatAccounts` instead of this struct?

services/horizon/internal/db2/history/main.go
```diff
@@ -338,6 +342,60 @@ func (e ExpAssetStat) PagingToken() string {
...
+
+// ExpAssetStatBalances represents the summarized balances for a single Asset
+type ExpAssetStatBalances struct {
+	Authorized                      string `json:"authorized"`
+	AuthorizedToMaintainLiabilities string `json:"authorized_to_maintain_liabilities"`
+	Unauthorized                    string `json:"unauthorized"`
+}
+
+func (e ExpAssetStatBalances) Value() (driver.Value, error) {
+	j, err := json.Marshal(e)
```
tamirms, line 386: it seems like you can just return `json.Marshal(e)`

services/horizon/internal/db2/history/main.go
```diff
@@ -338,6 +342,60 @@ func (e ExpAssetStat) PagingToken() string {
...
 
+// ExpAssetStatAccounts represents the summarized acount numbers for a single Asset
+type ExpAssetStatAccounts struct {
+	Authorized                      int32 `json:"authorized"`
+	AuthorizedToMaintainLiabilities int32 `json:"authorized_to_maintain_liabilities"`
+	Unauthorized                    int32 `json:"unauthorized"`
+}
+
+func (e ExpAssetStatAccounts) Value() (driver.Value, error) {
+	j, err := json.Marshal(e)
```
tamirms, line 353: it seems like you can just return `json.Marshal(e)`

services/horizon/internal/ingest/processors/asset_stats_processor.go
```diff
@@ -202,50 +219,37 @@ func (p *AssetStatsProcessor) adjustAssetStat(
...
 		}
-		// else, trustline was unauthorized and remains unauthorized
-		// so there is no change to accounts or balances
-	} else {
+	case preTrustline != nil && postTrustline == nil:
+		// removing a trustline
+		trustline = *preTrustline
+		deltaAccounts[trustline.Flags] = -1
+		deltaBalances[trustline.Flags] = int64(-preTrustline.Balance)
+	default:
```
bartekn, line 248: Nit: I think you can replace entire `switch` with a simple (please confirm):
```go
if preTrustline != nil {
	deltaAccounts[preTrustline.Flags] = -1
	deltaBalances[preTrustline.Flags] = int64(-preTrustline.Balance)
}
if postTrustline != nil {
	deltaAccounts[postTrustline.Flags] = deltaAccounts[preTrustline.Flags] + 1
	deltaBalances[postTrustline.Flags] = deltaBalances[preTrustline.Flags] + int64(postTrustline.Balance)
}
```
(but let's leave impossible case check: both nils)

services/horizon/internal/ingest/processors/asset_stats_set.go
```diff
@@ -24,16 +93,31 @@ type AssetStatSet map[assetStatKey]*assetStatValue
 // Add updates the set with a trustline entry from a history archive snapshot
 // if the trustline is authorized.
```
bartekn, line 94:
```suggestion
// Add updates the set with a trustline entry from a history archive snapshot.
```

services/horizon/internal/ingest/processors/asset_stats_set.go
```diff
@@ -13,9 +13,78 @@ type assetStatKey struct {
...
+	}
+}
+
+type assetStatAccounts struct {
+	Authorized                      int32
+	AuthorizedToMaintainLiabilities int32
+	Unauthorized                    int32
+}
+
+func (value assetStatValue) Finish() history.ExpAssetStat {
```
bartekn, line 77: Nit, name change?
```suggestion
func (value assetStatValue) ConvertToHistoryObject() history.ExpAssetStat {
```

services/horizon/internal/db2/schema/migrations/44_asset_stat_accounts_and_balances.sql
```diff
@@ -0,0 +1,19 @@
+-- +migrate Up
+
+ALTER TABLE exp_asset_stats
+ADD COLUMN accounts JSONB,
+ADD COLUMN balances JSONB;
+UPDATE exp_asset_stats
+  SET
+    accounts = jsonb_build_object('authorized', num_accounts),
+    balances = jsonb_build_object('authorized', amount);
```
bartekn, line 9: Maybe we don't need to keep old fields (`num_accounts`, `amount`) in a DB? We can easily fetch it from JSON and it's easy to forget to update it.

bartekn (approved):
Looks great! 🚀 Added some small improvement ideas, mostly 'nit's!

services/horizon/internal/ingest/processors/asset_stats_set.go
paulbellamy: Was just for symmetry with the `type assetStatBalances`, but sure, they're the same anyway.

services/horizon/internal/db2/schema/migrations/44_asset_stat_accounts_and_balances.sql
paulbellamy: Yeah was debating if we should minimize the db change, or clean it up proactively. I figured for now, remove it when the field is deprecated in the API. But happy to go either way.

services/horizon/internal/ingest/processors/asset_stats_processor.go
paulbellamy: This version misses the case the flags have changed, but found an (even simpler) version which works. :+1:

services/horizon/internal/ingest/processors/asset_stats_processor.go
```diff
@@ -202,50 +219,37 @@ func (p *AssetStatsProcessor) adjustAssetStat(
...
-			!xdr.TrustLineFlags(postTrustline.Flags).IsAuthorized() {
-			// trustline was authorized and became unauthorized
-			deltaAccounts = -1
-			deltaBalance = -preTrustline.Balance
-		} else if !xdr.TrustLineFlags(preTrustline.Flags).IsAuthorized() &&
-			xdr.TrustLineFlags(postTrustline.Flags).IsAuthorized() {
-			// trustline was unauthorized and became authorized
-			deltaAccounts = 1
-			deltaBalance = postTrustline.Balance
+		trustline = *postTrustline
```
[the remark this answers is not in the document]
bartekn: OK, I can see it now.

Issue #3405, opened 2021-02-12
Feature Request:  Make `LoggingMiddleware` extendable


Zagan202:
### What problem does your feature solve?
Update LoggingMiddleware to accept a list of header names. 
If a system using this feature wants to log values that are not in the default logger we should accept additional headers, such as `CF-Connecting-IP`

### What would you like to see?
Update the signature of the current constructor to accept a list of headers to log

### What alternatives are there?
Building another custom logger to accept custom headers.

PR #3456, opened 2021-03-08
go/support/http 3405 extend logging middleware

Zagan202:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Extends logStartOfRequest and adds a new function `LoggingMiddlewareWithOptions` to pass a slice of header names for the logger to capture.
 Closes:#3405
### Why

Logs values that are not in the default logger we should accept additional headers, such as `CF-Connecting-IP`

### Known limitations

Any extra headers given will need to be set in the `Request` ahead of time before calling.
Ex) `r.Header.Set("X-Forwarded-For", "203.0.113.195")`


support/http/logging_middleware.go
```diff
@@ -44,20 +44,54 @@ func LoggingMiddleware(next stdhttp.Handler) stdhttp.Handler {
 	})
 }
 
+// LoggingMiddlewareWithOptions is a middleware that logs requests to the logger.
+// With extra headers
+func LoggingMiddlewareWithOptions(extraHeaders ...string) func(stdhttp.Handler) stdhttp.Handler {
```
leighmcculloch, line 49: ❗ `WithOptions` makes sense if the additional parameter was an `Options`, but the additional params is extra headers, so I think you should either wrap the extra headers into an Options struct, which will make this extendable, or you should rename this function to `WithExtraHeaders`.

support/http/logging_middleware.go
```diff
@@ -44,20 +44,54 @@ func LoggingMiddleware(next stdhttp.Handler) stdhttp.Handler {
...
+			r = r.WithContext(ctx)
+
+			logStartOfRequest(r, extraHeaders...)
+
+			then := time.Now()
+			next.ServeHTTP(mw, r)
+			duration := time.Since(then)
+
+			logEndOfRequest(r, duration, mw)
+		})
```
leighmcculloch, line 68: 💡 This functionality, and the responsibility it captures, is identical to the functionality in LoggingMiddleware. I think you should have `LoggingMiddleware` call this function so that the responsibility lives in one place.

support/http/logging_middleware.go
```diff
@@ -44,20 +44,54 @@ func LoggingMiddleware(next stdhttp.Handler) stdhttp.Handler {
...
+	fields["method"] = r.Method
+	fields["ip"] = r.RemoteAddr
+	fields["host"] = r.Host
+	fields["useragent"] = r.Header.Get("User-Agent")
+	l := log.Ctx(r.Context()).WithFields(fields)
+
+	message := "starting request"
+	if len(fields) > 6 {
+		message = message + " w/Options"
+	}
```
leighmcculloch, line 93: ❗ This few lines is a bit brittle. If someone removes or adds a field they may not see that they need to update this magic number 6. I don't think the message needs a `w/Options` parameter since an application using this middleware will use one or the other, and won't need to distinguish between logs with options and logs without.

support/http/logging_middleware.go
```diff
@@ -44,20 +44,54 @@ func LoggingMiddleware(next stdhttp.Handler) stdhttp.Handler {
...
-		"path":      r.URL.String(),
-		"method":    r.Method,
-		"ip":        r.RemoteAddr,
-		"host":      r.Host,
-		"useragent": r.Header.Get("User-Agent"),
-	})
-	l.Info("starting request")
+	fields := log.F{}
+	for _, header := range extraHeaders {
+		fields[header] = r.Header.Get(header)
```
leighmcculloch, line 80: 💡 I think it might be easier to query these fields if they are named without their hashes and are lower case for consistency with the other fields we already name. I suspect it would mean when writing queries we don't need to escape the variables, but that might be unnecessary.

leighmcculloch (commented):
Nice! A left some comments below that I think need addressing (❗) and some suggestions (💡). Otherwise this looks good! Thanks!

support/http/logging_middleware.go
marcelosalloum: > I think you should either wrap the extra headers into an Options struct, which will make this extendable

I also prefer using an Options struct.

@Zagan202: using `Options{}` will make it easier to add further configuration options to this function in the future if we need to.

support/http/logging_middleware.go
```diff
@@ -44,20 +44,54 @@ func LoggingMiddleware(next stdhttp.Handler) stdhttp.Handler {
 	})
 }
 
+// LoggingMiddlewareWithOptions is a middleware that logs requests to the logger.
+// With extra headers
```
marcelosalloum, line 48: :nit:

```suggestion
// LoggingMiddlewareWithOptions is a middleware that logs requests to the logger
// with extra headers
```

support/http/logging_middleware_test.go
```diff
@@ -11,6 +11,21 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+// set "X-Forwarded-For" header to test LoggingMiddlewareWithOptions
+func setXFFMiddleWare(next stdhttp.Handler) stdhttp.Handler {
```
marcelosalloum, line 15: nit: method documentation usually starts with the method name and ends with a period `.`

You can see some examples at https://blog.golang.org/godoc.

```suggestion
// setXFFMiddleware sets "X-Forwarded-For" header to test LoggingMiddlewareWithOptions
func setXFFMiddleware(next stdhttp.Handler) stdhttp.Handler {
```

Also, Middleware is a single word so you don't need the capital `W` as in MiddleWare.

support/http/logging_middleware_test.go
```diff
@@ -11,6 +11,21 @@ import (
...
+	return stdhttp.HandlerFunc(func(w stdhttp.ResponseWriter, r *stdhttp.Request) {
+		r.Header.Set("X-Forwarded-For", "203.0.113.195")
+		next.ServeHTTP(w, r)
+	})
+}
+
+// set "Content-MD5" header to test LoggingMiddlewareWithOptions
+func setCMDFiveMiddleWare(next stdhttp.Handler) stdhttp.Handler {
```
marcelosalloum, line 23: Better to write MD5 rather than MDFive as it connects better with its meaning and how people are used to read it.

Also, abbreviating `Content` to `C` is a big leap. Better not to spare words and make it clearer.

```suggestion
// setContentMD5Middleware sets "Content-MD5" header to test LoggingMiddlewareWithOptions
func setContentMD5Middleware(next stdhttp.Handler) stdhttp.Handler {
```

support/http/logging_middleware_test.go
```diff
@@ -81,6 +96,85 @@ func TestHTTPMiddleware(t *testing.T) {
...
+	if assert.Len(t, logged, 7, "unexpected log line count") {
+		assert.Equal(t, "starting request w/Options", logged[0].Message)
+		assert.Equal(t, "http", logged[0].Data["subsys"])
+		assert.Equal(t, "GET", logged[0].Data["method"])
+		assert.NotEmpty(t, logged[0].Data["req"])
+		assert.Equal(t, "/path/1234", logged[0].Data["path"])
+		assert.Equal(t, "Go-http-client/1.1", logged[0].Data["useragent"])
+		assert.Equal(t, "203.0.113.195", logged[0].Data["X-Forwarded-For"])
+		assert.Equal(t, "Q2hlY2sgSW50ZWdyaXR5IQ==", logged[0].Data["Content-MD5"])
+		req1 := logged[0].Data["req"]
```
marcelosalloum, line 131: nit: you could additionally test the length of `logged[0].Data`.

marcelosalloum (commented):
Nice job! 

I'm adding a few suggestions below ⬇️  but I think Leigh already spotted the most important improvements.

support/http/logging_middleware.go
Zagan202: Sounds good I'll remove it then 👍🏾

support/http/logging_middleware.go
Zagan202: > 💡 I think it might be easier to query these fields if they are named without their hashes and are lower case for consistency with the other fields we already name. I suspect it would mean when writing queries we don't need to escape the variables, but that might be unnecessary.

Just to test my understanding...
Are you suggesting I lower the keys of `extraHeaders` given in order to make it easier for users to query those keys?
In the case of `"User-Agent"` for example we lower it we also remove the hyphen making it `useragent`.
I could do that but I think it would be a bit strange from the user's POV where they input an extra header a specific style (non caps) and query it another style for example ⬇️ 
```
extraHeaders := []string{"X-Forwarded-For", "Content-MD5"}
options := &Options{extraHeaders: extraHeaders}
mux.Use(LoggingMiddlewareWithOptions(options))
...
assert.Equal(t, "203.0.113.195", logged[0].Data["xforwardedfor"])
assert.Equal(t, "U3RlbGxhciBpcyBBd2Vzb21lIQ==", logged[0].Data["contentmd5"])
```

leighmcculloch: I agree, it is a bit odd that the query value is different compared to the value specified here, but our other fields already follow this format and it's unclear to me if that is accidental or intentional. I'm also not sure how specifying hyphenated fields will show up in Kibana. I know in some other log parsing platforms it is problematic and makes writing queries difficult because the hyphens have to be escaped in the queries.

Zagan202: > I'm also not sure how specifying hyphenated fields will show up in Kibana. I know in some other log parsing platforms it is problematic and makes writing queries difficult because the hyphens have to be escaped in the queries.

Thats good point, I didn't consider. 
I'll sanitize the input to make it consistent to the style we have currently and write a short comment explaining why.

Zagan202: Change made here
765c68e

support/http/logging_middleware_test.go
Zagan202: Change made here
eaf51c2

support/http/logging_middleware_test.go
Zagan202: Addressed here a18f383

support/http/logging_middleware_test.go
Zagan202: Addressed here a18f383

support/http/logging_middleware.go
Zagan202: Addressed here a18f383

support/http/logging_middleware.go
Zagan202: Addressed here 8d34fdc

support/http/logging_middleware.go
Zagan202: I believe this (52a721b) should address the suggestion.

support/http/logging_middleware.go
marcelosalloum: Both should work but in this case there is no need for creating a third function. Since the logic in `LoggingMiddleware()` should be the same as calling `LoggingMiddlewareWithOptions(nil)`, you can just have:

```go
func LoggingMiddleware() func(stdhttp.Handler) stdhttp.Handler {
    return LoggingMiddlewareWithOptions(nil)
}
```

This approach ☝️  is much easier to maintain than having two functions with almost the same logic inside.

marcelosalloum (commented):
Nice. I just left one suggestion 💡 regarding the regex, and a couple of nitpicks:

support/http/logging_middleware.go
Zagan202: Hmm I'm having trouble getting that to work properly as you suggested.

```
func LoggingMiddleware(next stdhttp.Handler) stdhttp.Handler {
	return LoggingMiddlewareWithOptions(nil)
}
```
```
cannot use LoggingMiddlewareWithOptions(nil) (type func(http.Handler) http.Handler) as type http.Handler in return argument:
	func(http.Handler) http.Handler does not implement http.Handler (missing ServeHTTP method)
```

Zagan202: > ```go
> func LoggingMiddleware() func(stdhttp.Handler) stdhttp.Handler {
>     return LoggingMiddlewareWithOptions(nil)
> }
> ```

```
 cannot use LoggingMiddleware (type func() func(http.Handler) http.Handler) as type func(http.Handler) http.Handler in argument to mux.Use
 ```
 Is there something I'm doing wrong?

Comment (marcelosalloum) [reply support/http/logging_middleware.go]: It looks like the method signatures are different, `LoggingMiddlewareWithOptions` returns a variable while `LoggingMiddleware` returns a function.

Comment (Zagan202) [reply support/http/logging_middleware.go]: I'm still a bit stumped as to why why this doesnt work 
```go
func LoggingMiddleware(next stdhttp.Handler) stdhttp.Handler {
	return stdhttp.HandlerFunc(func(w stdhttp.ResponseWriter, r *stdhttp.Request) {
		LoggingMiddlewareWithOptions(nil)
	})
}
```
It seems like it doesnt actually call the w/ options func
```

--- FAIL: TestHTTPMiddleware (0.00s)
    /Users/lawrence/go/src/go/support/http/printer.go:25: GET http://127.0.0.1:57924/path/1234
    /Users/lawrence/go/src/go/support/http/printer.go:25: GET http://127.0.0.1:57924/not_found
    /Users/lawrence/go/src/go/support/http/reporter.go:23: 
        	Error Trace:	reporter.go:23
        	            				chain.go:21
        	            				response.go:508
        	            				response.go:126
        	            				logging_middleware_test.go:43
        	Error:      	
        	            	expected status equal to:
        	            	 "404 Not Found"
        	            	
        	            	but got:
        	            	 "200 OK"
        	Test:       	TestHTTPMiddleware
    /Users/lawrence/go/src/go/support/http/printer.go:25: GET http://127.0.0.1:57924/really_not_found
    /Users/lawrence/go/src/go/support/http/reporter.go:23: 
        	Error Trace:	reporter.go:23
        	            				chain.go:21
        	            				response.go:508
        	            				response.go:126
        	            				logging_middleware_test.go:44
        	Error:      	
        	            	expected status equal to:
        	            	 "404 Not Found"
        	            	
        	            	but got:
        	            	 "200 OK"
        	Test:       	TestHTTPMiddleware
    /Users/lawrence/go/src/go/support/http/logging_middleware_test.go:49: 
        	Error Trace:	logging_middleware_test.go:49
        	Error:      	"[]" should have 7 item(s), but has 0
        	Test:       	TestHTTPMiddleware
        	Messages:   	unexpected log line count
FAIL
FAIL	github.com/stellar/go/support/http	0.245s
FAIL
```

support/http/logging_middleware.go
```diff
@@ -44,19 +51,55 @@ func LoggingMiddleware(next stdhttp.Handler) stdhttp.Handler {
...
+	fields := log.F{}
+	reg, err := regexp.Compile("[^a-zA-Z0-9]+")
+	if err != nil {
+		panic(err)
+	}
+
+	for _, header := range extraHeaders {
+		var headerkey = reg.ReplaceAllString(header, "")
+		fields[strings.ToLower(headerkey)] = r.Header.Get(header)
+	}
```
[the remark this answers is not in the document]
Zagan202: This regex removes any special characters from the header given such that `X-Forwarded-For` -> `XForwardedFor`
Then we lowercase the string making it`xforwardedfor`.
This will make the field keys uniform with the other keys in the logger.
@leighmcculloch brought up a good point to that kibana might not play friendly with the hyphens and such
https://github.com/stellar/go/pull/3456#discussion_r589728317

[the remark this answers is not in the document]
leighmcculloch: I'd focus on removing the character that matters to us right now, or that we know is commonly used in headers, and that is the hyphen. Regex's are expensive, and even though Go's regex implementation is linear, it's not ideal to use it in a middleware of every endpoint. I'd recommend using `strings.Replace` instead.

support/http/logging_middleware.go
Zagan202: Worked with @marcelosalloum offline and found that I wasnt properly calling `LoggingMiddlewareWithOptions` within `LoggingMiddleware`.
Because the `LoggingMiddlewareWithOptions` returns a `func(stdhttp.Handler)` that returns `stdhttp.Handler`
We need to give `LoggingMiddlewareWithOptions` nil for `Options` pointer and `next` for its function that requires a `stdhttp.Handler`.
Learned something new 🧠 💪🏾  
Addressed here 3a443f7

support/http/logging_middleware.go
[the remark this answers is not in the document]
Zagan202: Awesome, this looks a lot cleaner too.
Addressed here `0899ae0`

support/http/logging_middleware.go
```diff
@@ -10,6 +11,11 @@ import (
 	"github.com/stellar/go/support/log"
 )
 
+// Options allow the middleware logger to accept additional information.
+type Options struct {
+	extraHeaders []string
```
leighmcculloch, line 16: ❗ The extra headers needs to be exported for users of this package to be able to set it. Capitalize identifiers to make them accessible outside the current package.
```suggestion
	ExtraHeaders []string
```

support/http/logging_middleware.go
```diff
@@ -24,39 +30,58 @@ func SetLoggerMiddleware(l *log.Entry) func(stdhttp.Handler) stdhttp.Handler {
...
-				"req": middleware.GetReqID(r.Context()),
-			})
-		})
+	return LoggingMiddlewareWithOptions(nil)(next)
+}
 
-		r = r.WithContext(ctx)
+// LoggingMiddlewareWithOptions is a middleware that logs requests to the logger.
+// Requires an Options struct to accept additional information.
+func LoggingMiddlewareWithOptions(options *Options) func(stdhttp.Handler) stdhttp.Handler {
```
leighmcculloch, line 38: 💡 I would personally prefer to see `options Options` here, and then at the call sites above `Options{}`, because there is otherwise no need for this to be a pointer, and it makes the call site easier to understand for a reader. But this isn't a requirement.

support/http/logging_middleware.go
```diff
@@ -24,39 +30,58 @@ func SetLoggerMiddleware(l *log.Entry) func(stdhttp.Handler) stdhttp.Handler {
...
+				})
+			})
+			r = r.WithContext(ctx)
 
-		logStartOfRequest(r)
+			extraHeaders := []string{}
+			if options != nil {
+				extraHeaders = options.extraHeaders
+			}
+			logStartOfRequest(r, extraHeaders...)
```
leighmcculloch, line 53: 💡 If the `...` syntax is removed here and in the definition of `logStartOfRequest`, so prototype of the function is `logStartOfRequest(r *stdhttp.Request, extraHeaders []string)` then these lines of code can be reduced to:
```suggestion
			extraHeaders := []string(nil)
			if options != nil {
				extraHeaders = options.extraHeaders
			}
			logStartOfRequest(r, extraHeaders)
```
And if the pointer for the `Options` parameter is removed, then it gets even simpler:
```suggestion
			logStartOfRequest(r, options.extraHeaders)
```

leighmcculloch (approved):
One change requested (❗), otherwise LGTM. I also left couple suggestions that aren't required (💡).

support/http/logging_middleware.go
Zagan202: Good suggestion.

support/http/logging_middleware.go
Zagan202: That would make it a bit different to make the original `LoggingMiddleware` as flexible as it is currently though right?
```go
// LoggingMiddleware is a middleware that logs requests to the logger.
func LoggingMiddleware(next stdhttp.Handler) stdhttp.Handler {
	return LoggingMiddlewareWithOptions(nil)(next)
}
```
Which allows us to call it like so `mux.Use(LoggingMiddleware)`

Zagan202: Oh I think I can do this instead ⬇️ 
```go
// LoggingMiddleware is a middleware that logs requests to the logger.
func LoggingMiddleware(next stdhttp.Handler) stdhttp.Handler {
	return LoggingMiddlewareWithOptions(Options{})(next)
}
```

leighmcculloch: I'm not sure I'm following 100%, but it should make the call:
```go
// LoggingMiddleware is a middleware that logs requests to the logger.
func LoggingMiddleware(next stdhttp.Handler) stdhttp.Handler {
	return LoggingMiddlewareWithOptions(Options{})(next)
}
```

support/http/logging_middleware.go
Zagan202: These are really great refactor tips thanks @leighmcculloch  🥳

support/http/logging_middleware.go
Zagan202: Yup yup, addressed here 8074feb

support/http/logging_middleware.go
Zagan202: Done 8074feb

Zagan202:
@leighmcculloch When you have a chance give me a final thumbs up or down, just want to see if I addressed all your suggestions properly.

support/http/logging_middleware_test.go
```diff
@@ -11,6 +11,21 @@ import (
...
+}
+
+// setContentMD5MiddleWare sets header to test LoggingMiddlewareWithOptions.
+func setContentMD5Middleware(next stdhttp.Handler) stdhttp.Handler {
+	return stdhttp.HandlerFunc(func(w stdhttp.ResponseWriter, r *stdhttp.Request) {
+		r.Header.Set("Content-MD5", "U3RlbGxhciBpcyBBd2Vzb21lIQ==")
+		next.ServeHTTP(w, r)
+	})
+}
 func TestHTTPMiddleware(t *testing.T) {
```
leighmcculloch, line 29: I'm surprised gofmt didn't error on this, but there should be a new line between these functions.

leighmcculloch (approved):

leighmcculloch:
Nice work!

support/http/logging_middleware_test.go
Zagan202: 4243b1e

PR #3463, opened 2021-03-10
services/horizon: Stop generating separate Captive Core logs.

### LINKED ISSUE #3438: Allow setting LOG_FILE_PATH when starting Stellar-Core in captive mode

Some users would like to specify where Stellar-Core stores it's log file files because a default `/tmp` location is not always right.

Ideally, Stellar-Core team implements https://github.com/stellar/stellar-core/issues/2943 so there is no need to store log file anywhere (we embed it in Horizon log anyway).

Shaptic:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Passes along `LOG_FILE_PATH=""` to Stellar Core so that there is no separate log file.

### Why
Closes #3438: the log file is unnecessary since Horizon already (re)logs all of Core's output as the `stellar-core` subservice. This also saves space in the system's `/tmp` directory.

### Known limitations
This is not customizable. 

The original issue #3438 asks for adding a **parameter** to customize the output log (e.g. something like `--captive-core-log-path`), but after the [clarifications](https://github.com/stellar/stellar-core/issues/2943), it's unclear whether or not there's an active need in storing a separate log at all. If this is still an ask, it will be trivial to expand this PR.


ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -116,6 +116,7 @@ func (r *stellarCoreRunner) generateConfig() (string, error) {
 		fmt.Sprintf(`NETWORK_PASSPHRASE="%s"`, r.networkPassphrase),
 		fmt.Sprintf(`BUCKET_DIR_PATH="%s"`, filepath.Join(r.tempDir, "buckets")),
 		fmt.Sprintf(`HTTP_PORT=%d`, r.httpPort),
+		"LOG_FILE_PATH=\"\"",
```
bartekn, line 119: Nit
```suggestion
		`LOG_FILE_PATH=""`,
```

bartekn (approved):

ingest/ledgerbackend/stellar_core_runner.go
Shaptic: Ah, good call! Looks cleaner.

PR #3471, opened 2021-03-17
all: update all docker images to use Go 1.16.2

leighmcculloch:
### What

Update all docker images to use Go 1.16.2 and Ubuntu 18.04.

### Why

Go 1.14 is no longer supported or receiving security updates. Ubuntu needs upgrading to align the versions of Ubuntu in the runtime images with the build images.

### Known limitations

N/A

### Reviewing

If I have requested you review this PR it is because I believe you own or are the most knowledgeable person for a service impacted by this change. Please review the PR, and handle, or delegate someone else to handle, any deployment after merge for your respective service.

@accordeiro Ticker
@bartekn @ire-and-curses Horizon
@marcelosalloum Friendbot, Regulated Assets
@leighmcculloch Webauth, Recoverysigner, Keystore

### DIFF
--- a/services/horizon/docker/Dockerfile.dev
+++ b/services/horizon/docker/Dockerfile.dev
@@ -1,4 +1,4 @@
-FROM golang:1.14-stretch AS builder
+FROM golang:1.16.2 AS builder
 
 WORKDIR /go/src/github.com/stellar/go
 COPY go.mod go.sum ./
--- a/services/horizon/internal/scripts/check_release_hash/Dockerfile
+++ b/services/horizon/internal/scripts/check_release_hash/Dockerfile
@@ -1,5 +1,5 @@
 # Change to Go version used in CI or rebuild with --build-arg.
-ARG GO_IMAGE=golang:1.16.2-stretch
+ARG GO_IMAGE=golang:1.16.2
 FROM $GO_IMAGE
 
 WORKDIR /go/src/github.com/stellar/go
--- a/services/horizon/internal/scripts/check_release_hash/Dockerfile
+++ b/services/horizon/internal/scripts/check_release_hash/Dockerfile
@@ -1,5 +1,5 @@
 # Change to Go version used in CI or rebuild with --build-arg.
-ARG GO_IMAGE=golang:1.14.15-stretch
+ARG GO_IMAGE=golang:1.16.2-stretch
 FROM $GO_IMAGE
 
 WORKDIR /go/src/github.com/stellar/go
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -244,52 +244,11 @@ jobs:
       - staticcheck
       - build_packages
 
-  # test_code_1_14 performs all package tests using Go 1.14.
-  test_code_1_14:
-    working_directory: /go/src/github.com/stellar/go
-    docker:
-      - image: golang:1.14.15-stretch
-        environment:
-          GO111MODULE: "on"
-          PGHOST: localhost
-          PGPORT: 5432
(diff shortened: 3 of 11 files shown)


accordeiro (approved):
The Ticker part looks good to me, thanks @leighmcculloch!

services/horizon/internal/scripts/check_release_hash/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Change to Go version used in CI or rebuild with --build-arg.
-ARG GO_IMAGE=golang:1.14.15-stretch
+ARG GO_IMAGE=golang:1.16.2-stretch
```
marcelosalloum, line 2: ❓  Does it need to be a `-stretch` version?

marcelosalloum (approved):
The regulated assets reference server LGTM! I just added a question ❓  below.

Also, any reason to go with Ubuntu 18.04 rather than 20.04? Are you just playing safe there?

services/horizon/internal/scripts/check_release_hash/Dockerfile
bartekn, line 2:
```suggestion
ARG GO_IMAGE=golang:1.16.1-stretch
```
SDF's build server currently uses 1.16.1. To allow reproducible builds let's stick to 1.16.1.

services/horizon/docker/Dockerfile.dev
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.14-stretch AS builder
+FROM golang:1.16.2 AS builder
```
bartekn, line 1:
```suggestion
FROM golang:1.16.1 AS builder
```
SDF's build server currently uses 1.16.1. To allow reproducible builds let's stick to 1.16.1. And thanks for fixing this, we forgot to update this file.

bartekn (approved):

services/horizon/internal/scripts/check_release_hash/Dockerfile
leighmcculloch: Can we bump that to `.2`?

jacekn: Good point. Looking dockerhub tags it should be fine to switch to `golang:1.16.2` which defaults to buster

jacekn: @bartekn SDF Horizon builds are now Dockerized so they don't use go from the buildserver. We just need to bump image in the build job to match this

bartekn: We can bump it to `.2`. I'm just thinking that the next time we bump Golang version in our org we should create a meta issue with a list of places we need to update (right now I think it's Circle-CI, docker images, build server).

howardtw (approved):
keystore change lgtm. Thanks!

services/horizon/internal/scripts/check_release_hash/Dockerfile
leighmcculloch: Thanks, fixed in ae7f8ab384ef5d824449a4d14b91e176362e703f.

leighmcculloch: Resolving since we've updated to 1.16.1 now.

leighmcculloch:
I realize the CircleCI yml versions are also behind and need updating too, and we still had the old 1.14 builds listed, which aren't needed, so I removed them too.

howardtw (approved):
CircleCI config update also looks good to me.

PR #3472, opened 2021-03-18
services/horizon: Allow customization of the Captive Core log file.

### LINKED ISSUE #3438: Allow setting LOG_FILE_PATH when starting Stellar-Core in captive mode

Shaptic:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
This introduces the `--captive-core-log-path` parameter, which sets Stellar Core's `LOG_FILE_PATH` in the auto-generated configuration accordingly. If it's not set, Horizon will continue to log Captive Core's output to its log path with `subservice=stellar-core`.

### Why
Closes #3438

### Known limitations
n/a

### DIFF
--- a/ingest/ledgerbackend/captive_core_backend.go
+++ b/ingest/ledgerbackend/captive_core_backend.go
@@ -3,6 +3,7 @@ package ledgerbackend
 import (
 	"context"
 	"encoding/hex"
+	"io/ioutil"
 	"os"
 	"sync"
 
@@ -136,11 +137,21 @@ type CaptiveCoreConfig struct {
 func NewCaptive(config CaptiveCoreConfig) (*CaptiveStellarCore, error) {
 	// Here we set defaults in the config. Because config is not a pointer this code should
 	// not mutate the original CaptiveCoreConfig instance which was passed into NewCaptive()
+
+	// Create a default logging instance for Captive Core's output if no other
+	// logging configuration was specified.
 	if config.Log == nil {
 		config.Log = log.New()
-		config.Log.Logger.SetOutput(os.Stdout)
-		config.Log.SetLevel(logrus.InfoLevel)
+		if config.LogPath != "" {
+			// We still create an empty logger instance so there's no nil
+			// pointer floating around beyond this call.
+			config.Log.Logger.SetOutput(ioutil.Discard)
+		} else {
+			config.Log.Logger.SetOutput(os.Stdout)
+			config.Log.SetLevel(logrus.InfoLevel)
+		}
 	}
+
 	parentCtx := config.Context
 	if parentCtx == nil {
 		parentCtx = context.Background()
--- a/ingest/ledgerbackend/captive_core_backend.go
+++ b/ingest/ledgerbackend/captive_core_backend.go
@@ -111,6 +111,8 @@ type CaptiveCoreConfig struct {
 	NetworkPassphrase string
 	// HistoryArchiveURLs are a list of history archive urls
 	HistoryArchiveURLs []string
+	// LogPath is the path in which to store Core logs
+	LogPath string
 
 	// Optional fields
 
--- a/ingest/ledgerbackend/captive_core_backend.go
+++ b/ingest/ledgerbackend/captive_core_backend.go
@@ -3,7 +3,6 @@ package ledgerbackend
 import (
 	"context"
 	"encoding/hex"
-	"io/ioutil"
 	"os"
 	"sync"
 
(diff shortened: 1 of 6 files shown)


ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -116,7 +118,7 @@ func (r *stellarCoreRunner) generateConfig() (string, error) {
 		fmt.Sprintf(`NETWORK_PASSPHRASE="%s"`, r.networkPassphrase),
 		fmt.Sprintf(`BUCKET_DIR_PATH="%s"`, filepath.Join(r.tempDir, "buckets")),
 		fmt.Sprintf(`HTTP_PORT=%d`, r.httpPort),
-		`LOG_FILE_PATH=""`,
+		fmt.Sprintf(`LOG_FILE_PATH="%s"`, r.logPath),
```
tamirms, line 121: when `LOG_FILE_PATH` does stellar core log the same contents to both stdout and the log file path?

ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -111,6 +112,8 @@ type CaptiveCoreConfig struct {
 	NetworkPassphrase string
 	// HistoryArchiveURLs are a list of history archive urls
 	HistoryArchiveURLs []string
+	// LogPath is the path in which to store Core logs
+	LogPath string
```
bartekn, line 116: We should move it under `// Optional fields` line below. Also, let's update `Log` field comment (below).

services/horizon/internal/flags.go
```diff
@@ -255,6 +255,12 @@ func Flags() (*Config, support.ConfigOptions) {
 			OptType:   types.String,
 			Usage:     "name of the file where logs will be saved (leave empty to send logs to stdout)",
 		},
+		&support.ConfigOption{
+			Name:      "captive-core-log-path",
+			ConfigKey: &config.CaptiveCoreLogPath,
+			OptType:   types.String,
+			Usage:     "name of the path for Core logs (leave empty to log w/ Horizon)",
```
bartekn, line 262: To confirm with ops team: in SDF Horizon it would be great to still be able to see Stellar-Core logs embedded in Horizon log with external Stellar-Core log. I think it's not possible after this PR.

bartekn: OK, I have a confirmation from #ops that we can have logs stored in both. Could you add an option to support this?

ingest/ledgerbackend/stellar_core_runner.go
Shaptic: Empirically-speaking, I think so. We grok the output slightly differently, but it doesn't look like anything is excluded from either.

services/horizon/internal/init.go
```diff
@@ -72,6 +72,7 @@ func initIngester(app *App) {
 		CaptiveCoreBinaryPath:       app.config.CaptiveCoreBinaryPath,
 		CaptiveCoreConfigAppendPath: app.config.CaptiveCoreConfigAppendPath,
 		CaptiveCoreHTTPPort:         app.config.CaptiveCoreHTTPPort,
+		CaptiveCoreLogPath:          app.config.CaptiveCoreLogPath,
 		RemoteCaptiveCoreURL:        app.config.RemoteCaptiveCoreURL,
 		EnableCaptiveCore:           app.config.EnableCaptiveCoreIngestion,
```
Shaptic, line 77: As an informational question, why do we do this, rather than passing along app.config directly?

ingest/ledgerbackend/captive_core_backend.go
Shaptic: Good call, done in 6ce55527424a27725d3d9ea6dfbb70de0b18b53a.

services/horizon/internal/flags.go
Shaptic: Okay, done in 8d4aa56b6085142d03090b58c7f00762cecf750a.

bartekn, line 262:
```suggestion
			Usage:     "name of the path for Core logs (leave empty to log w/ Horizon only)",
```

services/horizon/internal/init.go
bartekn: It's clear which global config variables are used by the component. I know that we sometimes pass entire `config` in other places in Horizon but I believe we should only pass specific values.

bartekn (approved):

PR #3479, opened 2021-03-22
services/horizon: Add flag to customize storage directory for Captive Core data

### LINKED ISSUE #3437: Allow setting BUCKET_DIR_PATH when starting Stellar-Core in captive mode

Some users would like to specify where Stellar-Core stores bucket files because a default `/tmp` location is not always right (small partition, etc.)

Ideally, Stellar-Core team implements https://github.com/stellar/stellar-core/issues/2942 so there is no need to store bucket files on disk.

Shaptic:
### What
This introduces the `--captive-core-storage-path=./` parameter, which allows users to control where Horizon stores Captive Core's data. This includes the auto-generated configuration, the log, and most-importantly Core's bucket data. 

It defaults to the current directory rather than `/tmp` in order to raise visibility of the large amount of data Core stores in that directory. Internally, `/captive-core` is appended to it since we wipe the directory on shutdown.

### Why
Closes #3437.

### Known limitations
n/a


services/horizon/internal/flags.go
```diff
@@ -133,6 +133,14 @@ func Flags() (*Config, support.ConfigOptions) {
 			Usage:       "HTTP port for Captive Core to listen on (0 disables the HTTP server)",
 			ConfigKey:   &config.CaptiveCoreHTTPPort,
 		},
+		&support.ConfigOption{
+			Name:        "captive-core-storage-path",
+			OptType:     types.String,
+			FlagDefault: "./",
```
tamirms, line 139: I'm not sure if this is will work on windows. perhaps it would be more platform independent to have horizon call `os.Getwd()` if no value is provided

ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -157,7 +176,8 @@ func (r *stellarCoreRunner) generateConfig() (string, error) {
 }
 
 func (r *stellarCoreRunner) getConfFileName() string {
-	return filepath.Join(r.tempDir, "stellar-core.conf")
+	path, _ := filepath.Abs(filepath.Join(r.storagePath, "stellar-core.conf"))
```
bartekn, line 179: We should handle the error here.

services/horizon/internal/flags.go
bartekn: You can use `CustomSetValue` if we need to do some manual defaults processing.

Shaptic: `.` has resolved to the current directory in Windows for as long as I can remember, but I've updated this in the latest push.

ingest/ledgerbackend/stellar_core_runner.go
Shaptic: Shouldn't be fatal, so I did my best in 9d2d48502af27a4456a60b756b47d3eeaf963d06.

Shaptic: Shouldn't be fatal, so I did my best in 9d2d485.

services/horizon/internal/flags.go
```diff
@@ -133,6 +133,24 @@ func Flags() (*Config, support.ConfigOptions) {
 			Usage:       "HTTP port for Captive Core to listen on (0 disables the HTTP server)",
 			ConfigKey:   &config.CaptiveCoreHTTPPort,
 		},
+		&support.ConfigOption{
+			Name:        "captive-core-storage-path",
+			OptType:     types.String,
+			FlagDefault: "",
+			CustomSetValue: func(opt *support.ConfigOption) {
+				existingValue := *opt.ConfigKey.(*string)
```
bartekn, line 141: I can be wrong but I think the `ConfigKey` isn't set at this point (can you check). In the other places I can see we are using `viper.Get*` method so in this case it should be:
```suggestion
				existingValue := viper.GetString(co.Name)
```

bartekn (approved):

services/horizon/internal/flags.go
Shaptic: Oh interesting; I could've sworn I modeled this directly after the other uses of `CustomSetValue` but I guess not exactly.

Shaptic: Fixed in 32e72c6de9b1ea5b26a4b60792328a46d82b0938!

Shaptic:
I swear if I need to merge master and pass tests before merging **one** more friggin' time...

PR #3480, opened 2021-03-22
services/regulated-assets-approval-server: SEP-8 approval server should identify "Rejected" transactions

Zagan202:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

* [X] I've updated any docs ([developer docs](https://www.stellar.org/developers/reference/), `.md`

### Release planning

</details>

### What

SEP-8 approval server should reject transactions that don't meet the following criteria.

- If no transaction is submitted, respond with `{"status":"rejected", "error":"Missing parameter \"tx\""}`
- If can't parse XDR respond with `{"status":"rejected", "error":"Invalid parameter \"tx\""}`
- If the transaction sourceAccount the same as the server issuer account? If so, reject with `{"status":"rejected", "error":"The source account is invalid."}`
- If any operations where the sourceAccount is the server issuer account? If that operation is not of the type `AllowTrust` then respond with `{"status":"rejected", "error":"There is one or more unauthorized operations in the provided transaction"}`
- If all of that pass we can respond with `{"status":"rejected", "error":"Not implemented."}` for now.

Closes https://github.com/stellar/stellar-demo-wallet/issues/119
Update toml file to include [SEP-8 fields](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0008.md#sep-1-stellartoml)
### Why

We currently want to create the skeleton for the `tx_approve/` endpoint.
As a starting point we want to reject any and all transactions given to the server.
Next steps are to
- Extend rejection criteria
- Revise transaction payments for edge cases
- Revise transaction payments for the "happy path"
-  Evaluate a transaction to "Pending"
- Approve transactions

### Known limitations

We're assuming that the issuer is also the distributor of the regulated asset. 
Such that, any regulated asset that has a distribution account will not be able reject transactions with the correct response that have the following criteria.
- If the transaction sourceAccount the same as the regulated asset's distributor account
- If any operations where the sourceAccount is the regulated asset's distributor account
The aforementioned cases will be rejected via the default response `{"status":"rejected", "error":"Not implemented."}` for now.


services/regulated-assets-approval-server/serve/tx_approve.go
```diff
@@ -0,0 +1,134 @@
...
+	tx, ok := parsed.Transaction()
+	if !ok {
+		log.Ctx(ctx).Error(errors.Wrapf(err, "Transaction %s is not a simple transaction.", in.Transaction))
+		return &txApproveResponse{
+			Status: rejectedStatus,
+			Error:  invalidParamErr,
+		}, nil
+	}
+
+	issuerKP, err := keypair.Parse(h.issuerAccountSecret)
```
leighmcculloch, line 90: 💡 I would recommend parsing the keypair in the application boot and storing the parsed keypair in the handler. By doing so you can remove the need to handle this error condition here, and an operator of the application gets an immediate error at boot if they provide an invalid secret rather than an error at the first HTTP request.

You can use `*keypair.Full` for a key that is from an S secret key, or `*keypair.FromAddress` for a G address.

Example:
https://github.com/stellar/go/blob/ec5600bd6b2b6900d26988ff670b9ca7992313b8/exp/services/webauth/internal/serve/challenge.go#L17-L20

services/regulated-assets-approval-server/serve/tx_approve.go
```diff
@@ -0,0 +1,134 @@
...
+	if err != nil {
+		log.Ctx(ctx).Error(errors.Wrap(err, "Parsing issuer secret failed."))
+		return &txApproveResponse{
+			Status: rejectedStatus,
+			Error:  internalErrErr,
+		}, NewHTTPError(http.StatusBadRequest, `Parsing issuer secret failed.`)
```
leighmcculloch, line 96: ❓ These few lines are confusing, because the response is filled out but won't be used if an error is returned. It's unclear if this code is doing what it intends to do. Or is the goal to return two JSON objects?

services/regulated-assets-approval-server/serve/tx_approve.go
```diff
@@ -0,0 +1,134 @@
...
+	if err != nil {
+		httpErr, ok := err.(*httpError)
+		if !ok {
+			httpErr = serverError
+		}
+		httpErr.Render(w)
+	}
+	if rejectedResponse != nil {
+		httpjson.RenderStatus(w, http.StatusBadRequest, rejectedResponse, httpjson.JSON)
+	}
```
leighmcculloch, line 60: ❓ Will this return two JSON objects if a response and error are returned?

leighmcculloch (commented):
Couple questions that will probably result in changes or follow up (❓), otherwise looks great!

I also left one suggestion (💡).

services/regulated-assets-approval-server/serve/tx_approve.go
Zagan202: Thats a great idea!
Let me make those changes.

Zagan202: Thats a great idea!
Let me make those changes.

services/regulated-assets-approval-server/serve/tx_approve.go
Zagan202: After I change the keypair parsing logic this will be removed ideally.

services/regulated-assets-approval-server/serve/tx_approve.go
Zagan202: After some thought I think what should happen is if `err != nil` then render error and return.
That way it shouldn't return two JSOn responses causing confusion.

services/regulated-assets-approval-server/serve/tx_approve.go
Zagan202: Also, if in the future there are other internal errors that we don't want to send a SEP-8 like response for we should return `return nil, NewHTTPError(http.StatusBadRequest, `Something bad happened.`)` for those cases and have `ServeHTTP` handle and Render the error.

services/regulated-assets-approval-server/serve/tx_approve.go
Zagan202: addressed here d5e90c1

services/regulated-assets-approval-server/serve/tx_approve.go
Zagan202: addressed here b2d0328

services/regulated-assets-approval-server/serve/tx_approve.go
Zagan202: addressed here b2d0328

services/regulated-assets-approval-server/serve/handlers.go
```diff
@@ -3,10 +3,24 @@ package serve
...
 )
 
 func stellarTOMLHandler(cfg Options) http.HandlerFunc {
-	return http.HandlerFunc(func(rw http.ResponseWriter, _ *http.Request) {
+	return http.HandlerFunc(func(rw http.ResponseWriter, r *http.Request) {
+		issuerKP, err := keypair.Parse(cfg.IssuerAccountSecret)
+		if err != nil {
+			return
+		}
+		approvalCriteria := fmt.Sprintf("Currently %q is not approving any %q transactions at this time", r.Host, cfg.AssetCode)
```
marcelosalloum, line 16: nit: 

```suggestion
		approvalCriteria := fmt.Sprintf("Currently %q is not approving any %q transactions", r.Host, cfg.AssetCode)
```

services/regulated-assets-approval-server/serve/handlers.go
```diff
@@ -3,10 +3,24 @@ package serve
...
+		approvalServer := fmt.Sprintf("%s/tx_approve", r.Host)
 		fmt.Fprintf(rw, "NETWORK_PASSPHRASE=%q\n", cfg.NetworkPassphrase)
+		fmt.Fprintf(rw, "[[CURRENCIES]]\n")
+		fmt.Fprintf(rw, "code=%q\n", cfg.AssetCode)
```
marcelosalloum, line 20: 💡 could you also update the variable name `cfg` to `opts`? I used an improper name when I first implemented this method 😅 .

services/regulated-assets-approval-server/serve/handlers.go
```diff
@@ -3,10 +3,24 @@ package serve
 import (
 	"fmt"
 	"net/http"
+
+	"github.com/stellar/go/keypair"
 )
 
 func stellarTOMLHandler(cfg Options) http.HandlerFunc {
```
marcelosalloum, line 10: At some point (not required for this PR) we should update the stellarTOML handler to use the same pattern of the other handlers, i.e. use dependency injection + the `ServeHTTP` method.

services/regulated-assets-approval-server/serve/serve.go
```diff
@@ -23,10 +25,16 @@ type Options struct {
 }
 
 func Serve(opts Options) {
+	handler, err := handleHTTP(opts)
+	if err != nil {
+		err = errors.Wrap(err, "parsing secret")
+		log.Fatal(err)
+		return
```
marcelosalloum, line 32: 💡  There's no need in updating the value of err if you'll just use it once. Also, you don't need the return statement since log.Fatal will terminate immediately.

```suggestion
		log.Fatal(errors.Wrap(err, "parsing secret"))
```

services/regulated-assets-approval-server/serve/serve.go
```diff
@@ -43,7 +51,13 @@ func Serve(opts Options) {
...
 }
 
-func handleHTTP(opts Options) *chi.Mux {
+func handleHTTP(opts Options) (http.Handler, error) {
+	issuerKP, err := keypair.ParseFull(opts.IssuerAccountSecret)
+	if err != nil {
+		err = errors.Wrap(err, "parsing secret")
+		log.Error(err)
+		return nil, err
+	}
```
marcelosalloum, line 60: 💡  I don't think you need to return an error here to terminate the code in the parent/caller. in this case you can call `log.Fatal` here since handleHTTP is called only once, at boot time, and we usually want to fail fast at boot time.

services/regulated-assets-approval-server/serve/tx_approve.go
```diff
@@ -0,0 +1,126 @@
...
+
+const (
+	rejectedStatus    = "rejected"
+	missingParamErr   = "Missing parameter \"tx\"."
+	invalidParamErr   = "Invalid parameter \"tx\"."
+	internalErrErr    = "Internal Error."
+	invalidSrcAccErr  = "The source account is invalid."
+	unauthorizedOpErr = "There is one or more unauthorized operations in the provided transaction."
+	notImplementedErr = "Not implemented."
+)
```
marcelosalloum, line 23: 🤔  What exactly is the motivation for using these constants? Personally I think the code is more readable without them, but that's just my personal choice.

services/regulated-assets-approval-server/serve/tx_approve.go
```diff
@@ -0,0 +1,126 @@
...
+	issuerKP  *keypair.Full
+	assetCode string
+}
+
+type txApproveRequest struct {
+	Transaction string `json:"tx" form:"tx"`
+}
+
+type txApproveResponse struct {
+	Status string `json:"status"`
```
marcelosalloum, line 35: 💡  In this case I think we should use constants, even better if they were custom string-based type. Something like:

```go
type txApprovalStatus string

const txApprovalStatusRejected = "rejected"
```

We don't want to be returning any unexpected status that's not covered by [SEP-8](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0008.md).

services/regulated-assets-approval-server/serve/tx_approve.go
```diff
@@ -0,0 +1,126 @@
...
+		httpjson.RenderStatus(w, http.StatusBadRequest, rejectedResponse, httpjson.JSON)
+	}
+}
+
+func (h txApproveHandler) isRejected(ctx context.Context, in txApproveRequest) (*txApproveResponse, error) {
+	if in.Transaction == "" {
+		return &txApproveResponse{
+			Status: rejectedStatus,
+			Error:  missingParamErr,
+		}, nil
```
marcelosalloum, line 69: 💡  I'd make this into a method similar to `NewHTTPError` so we can make sure we're setting the `rejectedStatus` correctly in this one place. It'd make this code safer to maintain.

services/regulated-assets-approval-server/serve/tx_approve.go
```diff
@@ -0,0 +1,126 @@
...
+	tx, ok := parsed.Transaction()
+	if !ok {
+		log.Ctx(ctx).Error(errors.Wrapf(err, "Transaction %s is not a simple transaction.", in.Transaction))
+		return &txApproveResponse{
+			Status: rejectedStatus,
+			Error:  invalidParamErr,
+		}, nil
+	}
+
+	// Check if transaction's sourceaccount is the same as the server issuer account.
```
marcelosalloum, line 91: nit:
```suggestion
	// Check if transaction's source account is the same as the server issuer account.
```

services/regulated-assets-approval-server/serve/tx_approve.go
```diff
@@ -0,0 +1,126 @@
...
+	if tx.SourceAccount().AccountID == h.issuerKP.Address() {
+		log.Ctx(ctx).Error(errors.Wrapf(err,
+			"Transaction %s sourceAccount %s the same as the server issuer account %s",
```
marcelosalloum, line 94: 💡  I think there's no need to repeat the source account twice:

```suggestion
			"Transaction %s sourceAccount is the same as the server issuer account %s",
```

services/regulated-assets-approval-server/serve/tx_approve.go
```diff
@@ -0,0 +1,126 @@
...
+		}, nil
+
+	}
+
+	// Check if transaction's operation(s)' sourceaccount is the same as the server issuer account.
+	for _, op := range tx.Operations() {
+		opSourceAccount := op.GetSourceAccount()
+		if opSourceAccount == "" {
+			continue
+		}
```
marcelosalloum, line 110: ❕   You don't need this chunk of code.

services/regulated-assets-approval-server/serve/tx_approve.go
```diff
@@ -0,0 +1,126 @@
...
+		if op.GetSourceAccount() == h.issuerKP.Address() {
+			log.Ctx(ctx).Error(errors.Wrapf(err,
+				"Unauthorized operation from %s in the provided transaction",
+				op.GetSourceAccount()))
```
marcelosalloum, line 114: ❕  There's no error to be logged here:

```suggestion
```

services/regulated-assets-approval-server/serve/tx_approve_test.go
```diff
@@ -0,0 +1,652 @@
...
+		Issuer: issuerAccKeyPair.Address(),
+	}
+	// Test if no transaction is submitted.
+	req := txApproveRequest{
+		Transaction: "",
+	}
+	rejectedResponse, err := txApproveHandler{
+		issuerKP:  issuerAccKeyPair,
+		assetCode: assetGOAT.GetCode(),
+	}.isRejected(ctx, req)
```
marcelosalloum, line 33: ❕  Instead of re-declaring this exact same `txApproveHandler` throughout this function you should define it as a new variable at the top of the function and reuse that variable.

services/regulated-assets-approval-server/serve/tx_approve_test.go
```diff
@@ -0,0 +1,652 @@
...
+	require.NoError(t, err)
+	wantRejectedResponse = txApproveResponse{
+		Status: rejectedStatus,
+		Error:  invalidSrcAccErr,
+	}
+	assert.Equal(t, &wantRejectedResponse, rejectedResponse)
+
+	// Test if the transaction's operation sourceaccount the same as the server issuer account
```
marcelosalloum, line 127: nit:
```suggestion
	// Test if the transaction's operation sourceAccount the same as the server issuer account
```

services/regulated-assets-approval-server/serve/tx_approve_test.go
```diff
@@ -0,0 +1,652 @@
...
+	assert.Equal(t, "application/json; charset=utf-8", resp.Header.Get("Content-Type"))
+
+	body, err = ioutil.ReadAll(resp.Body)
+	require.NoError(t, err)
+	wantBody = `{
+		"status":"rejected", "error":"The source account is invalid."
+	}`
+	require.JSONEq(t, wantBody, string(body))
+
+	// Test if the transaction's operation sourceaccount the same as the server issuer account
```
marcelosalloum, line 343: nit:
```suggestion
	// Test if the transaction's operation sourceAccount the same as the server issuer account
```

services/regulated-assets-approval-server/serve/tx_approve_test.go
```diff
@@ -0,0 +1,652 @@
...
+	resp = w.Result()
+
+	assert.Equal(t, http.StatusBadRequest, resp.StatusCode)
+	assert.Equal(t, "application/json; charset=utf-8", resp.Header.Get("Content-Type"))
+
+	body, err = ioutil.ReadAll(resp.Body)
+	require.NoError(t, err)
+	wantBody = `{
+		"status":"rejected", "error":"Invalid parameter \"tx\"."
+		}`
```
marcelosalloum, line 481: nit: 
```suggestion
	}`
```

services/regulated-assets-approval-server/serve/tx_approve.go
```diff
@@ -0,0 +1,126 @@
...
+		httpErr, ok := err.(*httpError)
+		if !ok {
+			httpErr = serverError
+		}
+		httpErr.Render(w)
+		return
+	}
+	if rejectedResponse != nil {
+		httpjson.RenderStatus(w, http.StatusBadRequest, rejectedResponse, httpjson.JSON)
+	}
```
marcelosalloum, line 61: You don't need the `if` statement here. If you arrived at this point the `rejectedResponse` shouldn't be nil.

As a go pattern, whenever you have something like:
```go
result, err := fooBar()
```

You should have either `err` **or** `result` be not-empty while the other will be empty/null.

marcelosalloum (commented):
Great code coverage! That's a pretty solid implementation! 🎉 

About the tests: when you test all use cases for the `ServeHTTP` for both `json` and `form` and also all edge cases for `h.isRejected` you're kind of testing the same parts of the code three times. Having a full coverage of all edge cases for the `json` and covering just the happy path for the other two options should be enough. There's no need to remove any tests from this PR, I'm just letting you know it can be leaner for future PRs.

About the TOML file, would you mind adding a unit test for it? Could be in a separate PR though.

services/regulated-assets-approval-server/serve/tx_approve.go
marcelosalloum: I know this is not being used in this PR anymore, but I'd like to touch base on this to share my opinion on how we should handle these errors.

We'll want to return a **rejected** response only if there's something wrong with the data sent from the user. The [SEP-8 rejection section] says:

> **Rejected**
>
> This response means that the transaction is not compliant and could not be revised to be made compliant.

In case we get an error parsing a secret or signing a transaction for instance, it's likely it's the server's fault by mis-constructing the transaction or using an invalid account secret. In cases like these we should return a `500 - internal server error` so the client knows it's not the client fault but rather the server's.

[SEP-8 rejection section]: https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0008.md#rejected

Zagan202:
> Great code coverage! That's a pretty solid implementation! 🎉
> 
> About the tests: when you test all use cases for the `ServeHTTP` for both `json` and `form` and also all edge cases for `h.isRejected` you're kind of testing the same parts of the code three times. Having a full coverage of all edge cases for the `json` and covering just the happy path for the other two options should be enough. There's no need to remove any tests from this PR, I'm just letting you know it can be leaner for future PRs.
> 
> About the TOML file, would you mind adding a unit test for it? Could be in a separate PR though.

I made an issue for improving the tests So its addressed in another PR
https://github.com/stellar/go/issues/3495

services/regulated-assets-approval-server/serve/handlers.go
Zagan202: Addressed here 29da569

services/regulated-assets-approval-server/serve/handlers.go
Zagan202: Good idea I'll make an issue for that

Zagan202: Issue here https://github.com/stellar/go/issues/3496

services/regulated-assets-approval-server/serve/serve.go
Zagan202: Great catch, I'll keep that in mind when or if we include more processing at boot time.
Addressed here 
aa97336

services/regulated-assets-approval-server/serve/tx_approve.go
Zagan202: I prefer these constants because some of the messages are reused in isRejected and it makes it easier to edit text in one spot.

And in the future it may make it easier to define these in and .env file or what not.
Its mostly inspired from how HTTP response codes are implemented in some packages.

services/regulated-assets-approval-server/serve/tx_approve_test.go
Zagan202: Addressed here b730dd8

services/regulated-assets-approval-server/serve/tx_approve.go
Zagan202: I can change this now but I'll end up reverting it when the revise/approve states are added to the code.
I've been trying to tackle how I would do this elegantly there but I'm hitting a wall.
As it stands now it looks something like this. ⬇️ 
```go
        resp, err := h.isRejected(ctx, in)
	if err != nil {
		...
	}
	if resp != nil {
		httpjson.RenderStatus(w, http.StatusBadRequest, resp, httpjson.JSON)
		return
	}
	resp, err = h.Approve(ctx, in)
	if err != nil {
		...
	}
	if resp != nil {
		httpjson.RenderStatus(w, http.StatusOK, resp, httpjson.JSON)
		return
	}
```
Its functional but ultimately kinda redundant. 
I'm very open to refactor suggestions if you have any. :)

marcelosalloum: Can we focus on this one PR for now and then discuss the snippet you just shared in your next PR?

I have some ideas but it's hard to be opinionated without seeing the full code... if I give you a suggestion based on how I suppose your future implementation will be chances are I'll miss the point.

services/regulated-assets-approval-server/serve/handlers.go
```diff
@@ -3,10 +3,24 @@ package serve
...
 
-func stellarTOMLHandler(cfg Options) http.HandlerFunc {
-	return http.HandlerFunc(func(rw http.ResponseWriter, _ *http.Request) {
-		fmt.Fprintf(rw, "NETWORK_PASSPHRASE=%q\n", cfg.NetworkPassphrase)
+func stellarTOMLHandler(opts Options) http.HandlerFunc {
+	return http.HandlerFunc(func(rw http.ResponseWriter, r *http.Request) {
+		issuerKP, err := keypair.Parse(opts.IssuerAccountSecret)
+		if err != nil {
+			return
+		}
```
marcelosalloum, line 15: 💡  You're already instantiating the keypair at boot time, right? You should be passing it as a parameter to `stellarTOMLHandler` instead of parsing it at request time, as Leigh suggested in that other thread.

services/regulated-assets-approval-server/serve/handlers.go
```diff
@@ -3,10 +3,24 @@ package serve
...
+		approvalCriteria := fmt.Sprintf("Currently %q is not approving any %q transactions", r.Host, opts.AssetCode)
+		approvalServer := fmt.Sprintf("%s/tx_approve", r.Host)
```
marcelosalloum, line 17: Nice! I didn't know you could grab the host so easily like that 🙌 .

services/regulated-assets-approval-server/serve/handlers.go
```diff
@@ -3,10 +3,24 @@ package serve
...
+		fmt.Fprintf(rw, "NETWORK_PASSPHRASE=%q\n", opts.NetworkPassphrase)
+		fmt.Fprintf(rw, "[[CURRENCIES]]\n")
+		fmt.Fprintf(rw, "code=%q\n", opts.AssetCode)
+		fmt.Fprintf(rw, "issuer=%q\n", issuerKP.Address())
+		fmt.Fprintf(rw, "regulated=true\n")
+		fmt.Fprintf(rw, "approval_server=%q\n", approvalServer)
+		fmt.Fprintf(rw, "approval_criteria=%q\n", approvalCriteria)
```
marcelosalloum, line 24: 💡  I don't see a reason for defining `approvalServer` and `approvalCriteria` in advance since you're using it only once. You could be building the result right in this line the same way you did for  `code`, `issuer` and `regulated`.

services/regulated-assets-approval-server/serve/serve.go
```diff
@@ -43,7 +49,12 @@ func Serve(opts Options) {
 	supporthttp.Run(serverConfig)
 }
 
-func handleHTTP(opts Options) *chi.Mux {
+func handleHTTP(opts Options) (http.Handler, error) {
```
marcelosalloum, line 52: ❕  I think you forgot to revert the method signature to its original state. You never return an error inside this method, so you can simplify its signature.

services/regulated-assets-approval-server/serve/serve.go
```diff
@@ -43,7 +49,12 @@ func Serve(opts Options) {
...
+	issuerKP, err := keypair.ParseFull(opts.IssuerAccountSecret)
+	if err != nil {
+		err = errors.Wrap(err, "parsing secret")
+		log.Fatal(err)
```
marcelosalloum, line 56: 💡  Since you're only using the `err` once there's no need to define it in advance:

```suggestion
		log.Fatal(errors.Wrap(err, "parsing secret"))
```

services/regulated-assets-approval-server/serve/tx_approve.go
```diff
@@ -0,0 +1,122 @@
...
+		log.Ctx(ctx).Error(errors.Wrapf(err, "Parsing transaction %s failed", in.Transaction))
+		return &txApproveResponse{
+			Status: rejectedStatus,
+			Error:  invalidParamErr,
+		}, nil
+	}
+
+	tx, ok := parsed.Transaction()
+	if !ok {
+		log.Ctx(ctx).Error(errors.Wrapf(err, "Transaction %s is not a simple transaction.", in.Transaction))
```
marcelosalloum, line 84: ❓ That's ok to log an error message but why are you wrapping the `err` here?

services/regulated-assets-approval-server/serve/tx_approve.go
```diff
@@ -0,0 +1,122 @@
...
+			Error:  invalidParamErr,
+		}, nil
+	}
+
+	// Check if transaction's source account is the same as the server issuer account.
+	if tx.SourceAccount().AccountID == h.issuerKP.Address() {
+		log.Ctx(ctx).Error(errors.Wrapf(err,
+			"Transaction %s sourceAccount is the same as the server issuer account %s",
+			in.Transaction,
+			h.issuerKP.Address()))
```
marcelosalloum, line 96: ❓ That's ok to log an error message but why are you wrapping the `err` here?

marcelosalloum (commented):
Good job, I've just added a few extra details to be improved.

Also, I think there are a couple comments from the last review that were not addressed, like https://github.com/stellar/go/pull/3480#discussion_r600883368 and https://github.com/stellar/go/pull/3480#discussion_r600880765.

services/regulated-assets-approval-server/serve/tx_approve.go
Zagan202: I dont think I follow, I'm doing this above no?
```
const (
	rejectedStatus    = "rejected"
	missingParamErr   = "Missing parameter \"tx\"."
	invalidParamErr   = "Invalid parameter \"tx\"."
	internalErrErr    = "Internal Error."
	invalidSrcAccErr  = "The source account is invalid."
	unauthorizedOpErr = "There is one or more unauthorized operations in the provided transaction."
	notImplementedErr = "Not implemented."
)
```

services/regulated-assets-approval-server/serve/handlers.go
Zagan202: Yeah its pretty sweet

services/regulated-assets-approval-server/serve/handlers.go
Zagan202: I was planning on porting those constants elsewhere like an .env file in a future PR.
I can see the `approvalCriteria` value changing semi frequently.

services/regulated-assets-approval-server/serve/serve.go
Zagan202: Whoops addressed here 2e5d15d

services/regulated-assets-approval-server/serve/tx_approve.go
Zagan202: Ah, poor copy paste job 😅

Zagan202: Addressed here 585f4a9

services/regulated-assets-approval-server/serve/tx_approve.go
Zagan202: Good catch

Zagan202: 0001d78

services/regulated-assets-approval-server/serve/tx_approve.go
Zagan202: Addressed here 57e36ce

services/regulated-assets-approval-server/serve/tx_approve.go
Zagan202: Great idea!
7764975

services/regulated-assets-approval-server/serve/tx_approve.go
Zagan202: Missed this one.
Addressed here bda8e13

services/regulated-assets-approval-server/serve/handlers.go
Zagan202: I did a major refactor on this one here d68ad2e
This is to make it more like the other handlers in order to.
A) Make it easier to maintain
B) Pull in only the data we need including the issuer's public address. Which we can get from the kp we instantiated at boot time.

I'll make a PR to rename this function from handlers.go -> stellarTOMLHandler.go (or something).

services/regulated-assets-approval-server/serve/tx_approve.go
marcelosalloum: My point is: `txApproveResponse.Status` currently allows any kind of string to be set, but [SEP-8] only defines 5 possible states:

```go
type txApproveResponse struct {
	Status string `json:"status"`
	Error  string `json:"error"`
}
```

This is error prone and chances are a developer will eventually add a typo or a wrong state there. 

This is a great use case for enums. There's a good text about it [here](https://www.ribice.ba/golang-enums/) and sounds like the following snippet could work:

```go
type sep8Status string

const(
    Sep8StatusRejected sep8Status = "rejected"
)

type txApproveResponse struct {
	Status sep8Status `json:"status"`
	Error  string     `json:"error"`
}
```

This would make it safer to assume this property won't be misconfigured that easily.

[SEP-8]: https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0008.md#regulated-assets-approval-flow

services/regulated-assets-approval-server/serve/tx_approve.go
```diff
@@ -0,0 +1,103 @@
...
+	Status string `json:"status"`
+	Error  string `json:"error"`
+}
+
+func (h txApproveHandler) craftTXApproveResponse(status string, errorMessage string) *txApproveResponse {
+	return &txApproveResponse{
+		Status: status,
+		Error:  errorMessage,
+	}
+}
```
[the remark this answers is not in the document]
Zagan202: Got it.
Both this and https://github.com/stellar/go/pull/3480#discussion_r601802803 make a lot of sense.
Thanks for your input!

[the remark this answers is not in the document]
Zagan202: Addressed here 332c027

Zagan202:
@marcelosalloum addressed [this request](https://github.com/stellar/go/pull/3480#discussion_r601802803) here 332c027

services/regulated-assets-approval-server/serve/handlers.go
marcelosalloum: Nice!

services/regulated-assets-approval-server/serve/handlers.go
marcelosalloum: SGTM!

services/regulated-assets-approval-server/serve/tx_approve.go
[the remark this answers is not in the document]
marcelosalloum: Awesome!

marcelosalloum (approved):
Looks great! 🚀

Issue #3305, opened 2020-12-22
Feature Request: Expose the transaction and operation(s) that created the claimable balance


leighmcculloch:
### What problem does your feature solve?

There is no effecient method for a receiver (claimant) to identify the sender of a [two-part payment](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0023.md), i.e. the creator of a claimable balance, or the memo of the transaction that created it.

It's possible to discover the sender by getting the ledger ID that the claimable balance was created in, and iterating over all operations in that ledger, looking for the operation that created it. This is very inefficient and cumbersome for the developer.

Claimable balances have a sponsor field and that field is set to the account that created the claimable balance if the ledger entry isn't sponsored. However, claimable balances that are sponsored won't have the creator set there and no application should rely on the sponsor being the sender.

Claimable balances were created with the intention of facilitating two-part payments. While they provide generic types for separating an amount into a balance disconnected from any account they still need to facilitate payments like how direct payments do. The sender shouldn't be obfuscated from the receiver, and it should be equally easy to find the memo of the transaction that created it so that if the memo is used as an order ID or a virtual account identifier, it can be looked up.

Example:
https://horizon.stellar.org/claimable_balances/00000000178826fbfe339e1f5c53417c6fedfe2c05e8bec14303143ec46b38981b09c3f9
```json
{
  "_links": {
    "self": {
      "href": "https://horizon.stellar.org/claimable_balances/00000000178826fbfe339e1f5c53417c6fedfe2c05e8bec14303143ec46b38981b09c3f9"
    }
  },
  "id": "00000000178826fbfe339e1f5c53417c6fedfe2c05e8bec14303143ec46b38981b09c3f9",
  "asset": "BODHI:GDCJIHD3623OCYNH65UUQC3NLG2D6YCNCDPZULRLCLOA76TBQRL6A3TF",
  "amount": "0.1000000",
  "sponsor": "GDCJIHD3623OCYNH65UUQC3NLG2D6YCNCDPZULRLCLOA76TBQRL6A3TF",
  "last_modified_ledger": 32747318,
  "last_modified_time": "2020-11-23T16:02:38Z",
  "claimants": [
    {
      "destination": "GBEUDKANIFPTFHPWJ5T3R6RIO36RQBFGHYPAQ6STH7KMNDHAT36LHOLD",
      "predicate": {
        "unconditional": true
      }
    }
  ],
[...]

PR #3483, opened 2021-03-22
services/horizon: Issue 3305, Claimable Balance API Extension

paulbellamy:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Track transactions and operations for claimable balances, even after claimable balances are claimed.

### Why

https://github.com/stellar/go/issues/3305

### Known limitations

[TODO or N/A]


2opremio:
We still need to look into (and extend) the integration tests.

bartekn (commented):
Looks good! Before 👍 I think we need to change two things:
1. Store IDs as hex instead of base64 in a DB. It will allow removing most of the marshaling/unmarshaling code and will make it much easier to read.
2. Check the comments connected to validation. I think we can simplify the code in `actions`.

services/horizon/internal/ingest/processors/operations_processor.go
```diff
@@ -689,3 +689,55 @@ func operationsParticipants(transaction ingest.LedgerTransaction, sequence uint3
...
+		found := false
+		for _, change := range changes {
+			if change.Pre != nil ||
+				change.Post == nil ||
+				change.Type != xdr.LedgerEntryTypeClaimableBalance {
+				continue
+			}
+			found = true
+			cb := change.Post.Data.MustClaimableBalance()
+			cbs = append(cbs, cb.BalanceId)
```
Shaptic, line 713: Can there be more than one ID found in this loop? If not, `break`.

You also might be able to use `txnbuild.Transaction.ClaimableBalanceID()` here, instead? Though it might be just as verbose.

services/horizon/internal/actions/operation.go
```diff
@@ -93,6 +108,12 @@ func (handler GetOperationsHandler) GetResourcePage(w HeaderWriter, r *http.Requ
 	switch {
 	case qp.AccountID != "":
 		query.ForAccount(qp.AccountID)
+	case qp.ClaimableBalanceID != "":
+		cbID, parseErr := qp.BalanceID()
+		if parseErr != nil {
+			return nil, parseErr
+		}
```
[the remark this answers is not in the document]
2opremio: I don't think we should be ignoring the error (regardless of the validator)

services/horizon/internal/db2/history/history_claimable_balances.go
```diff
@@ -0,0 +1,159 @@
...
+func (q *Q) ClaimableBalancesByIDs(dest interface{}, ids []xdr.ClaimableBalanceId) error {
+	sql := selectHistoryClaimableBalance.Where(map[string]interface{}{
+		"hcb.claimable_balance_id": ids, // hcb.claimable_balance_id IN (...)
+	})
+	return q.Select(dest, sql)
+}
+
+// ClaimableBalancesByID loads a row from `history_claimable_balances`, by claimable_balance_id
+func (q *Q) ClaimableBalancesByID(dest interface{}, cbIDy xdr.ClaimableBalanceId) error {
+	sql := selectHistoryClaimableBalance.Limit(1).Where("hcb.claimable_balance_id = ?", cbIDy)
```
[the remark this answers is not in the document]
2opremio: We initially used an hex-string representation in the db, but then we saw that we already use a base64 representation in other places and changed it for consistency.

I think we should keep it consistent with what we already have.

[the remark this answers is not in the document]
bartekn: I think we use base64 for low level xdr structures that are hard to represent in a different way. The similar example is account IDs, we don't store them as base64 and decoded strkey to find an account in a DB. We just use strkey directly. Because we use hex representation of claimable balance ID in Horizon and SDKs I think we should use it in a DB too. Plus, the code will be much simpler.

services/horizon/internal/actions/operation.go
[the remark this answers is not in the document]
bartekn: I'm not sure there will be any error here that's worth handling. The only case I can think of is the user passing invalid claimable balance ID and in such case it just won't be found.

services/horizon/internal/db2/history/history_claimable_balances.go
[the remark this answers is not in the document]
2opremio: @bartekn While I agree, we are already using a base64 representation for `ClaimableBalanceID` in the `claimable_balances` table. Thus, I think we should respect consistency and not use a different representation for `ClaimableBalanceID` in two different places.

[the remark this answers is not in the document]
2opremio: (For context, we initially used an hex representation, using the same reasoning as you, but then we learned about the representation in the `claimable_balances` table).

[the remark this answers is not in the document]
bartekn: For the record, we can easily change the representation in `claimable_balances` table because it just requires a state rebuild. But we won't be able to change the code here so easily because it requires history reingestion.

It's an extra work so if not now we can do it next sprint. I think it's worth it because this PR alone will be literally hundreds lines of code smaller. No strong opinion, decision is yours.

services/horizon/internal/actions/operation.go
```diff
@@ -57,6 +59,19 @@ func (qp OperationsQuery) Validate() error {
...
 }
 
+// BalanceID returns the xdr.ClaimableBalanceId from the request query
+func (q OperationsQuery) BalanceID() (xdr.ClaimableBalanceId, error) {
+	var balanceID xdr.ClaimableBalanceId
+	err := xdr.SafeUnmarshalHex(q.ClaimableBalanceID, &balanceID)
+	if err != nil {
+		return balanceID, supportProblem.MakeInvalidFieldProblem(
+			"claimable_balance_id",
+			fmt.Errorf("Invalid claimable balance ID"),
```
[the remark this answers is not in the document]
paulbellamy: It's an error, not a string. Used `errors.New` for consistency with other similar funcs in this file though.

services/horizon/internal/httpx/router.go
```diff
@@ -222,6 +222,15 @@ func (r *Router) addRoutes(config *RouterConfig, rateLimiter *throttled.HTTPRate
 			})
 		})
 	})
+	// claimable balance actions
+	r.Group(func(r chi.Router) {
+		r.Use(historyMiddleware)
+		r.Method(http.MethodGet, "/claimable_balances/{claimable_balance_id:[[:xdigit:]]+}/operations", streamableHistoryPageHandler(ledgerState, actions.GetOperationsHandler{
+			LedgerState:  ledgerState,
+			OnlyPayments: false,
```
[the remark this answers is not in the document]
paulbellamy: We've included it explicitly as `false` every other time we use it.

services/horizon/internal/actions/operation.go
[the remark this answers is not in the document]
paulbellamy: It would be a nice usability thing to differentiate between totally invalid and not found though. But seems like a nice-to-have.

services/horizon/internal/ingest/processors/claimable_balances_transaction_processor.go
```diff
@@ -0,0 +1,297 @@
...
+		transaction,
+	)
+	if err != nil {
+		return errors.Wrap(err, "Could not determine claimable balances for transaction")
+	}
+
+	for _, cb := range transactionClaimableBalances {
+		entry := cbSet[cb]
+		entry.addTransactionID(transactionID)
+		cbSet[cb] = entry
```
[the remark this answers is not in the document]
paulbellamy: No. `cannot call pointer method addTransactionID on claimableBalance`. The final `cbSet[cb] = entry` is unnecessary though.

services/horizon/internal/ingest/processors/operations_processor.go
```diff
@@ -689,3 +689,41 @@ func operationsParticipants(transaction ingest.LedgerTransaction, sequence uint3
...
 }
+
+// ClaimableBalances returns the accounts taking part in the operation.
+func (operation *transactionOperationWrapper) ClaimableBalances() ([]xdr.ClaimableBalanceId, error) {
+	cbs := []xdr.ClaimableBalanceId{}
+
+	switch operation.OperationType() {
+	case xdr.OperationTypeCreateClaimableBalance:
+		op := operation.OperationResult().MustCreateClaimableBalanceResult()
+		if op.BalanceId != nil {
```
2opremio, line 700: If it's nil, I think the caller can choke (breaking ingestion). I think we should explicitly consider that possibility.

2opremio: Scratch that, the callers won't choke. It was the prior change lookup that choked if there was no change was found.

services/horizon/internal/ingest/processors/operations_processor.go
paulbellamy: Ended up using the Results object from @bartekn suggested above.

services/horizon/internal/actions/operation.go
```diff
@@ -30,6 +31,7 @@ func (qp Joinable) IncludeTransactions() bool {
 type OperationsQuery struct {
 	Joinable                  `valid:"optional"`
 	AccountID                 string `schema:"account_id" valid:"accountID,optional"`
+	ClaimableBalanceID        string `schema:"claimable_balance_id" valid:"claimableBalanceID,optional"`
```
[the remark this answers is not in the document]
2opremio: I think we do exactly the same thing in `TransactionsQuery`

services/horizon/internal/actions/transaction.go
```diff
@@ -57,15 +58,30 @@ func (handler GetTransactionByHashHandler) GetResource(w HeaderWriter, r *http.R
...
+	var balanceID xdr.ClaimableBalanceId
+	err := xdr.SafeUnmarshalHex(q.ClaimableBalanceID, &balanceID)
+	if err != nil {
+		return balanceID, supportProblem.MakeInvalidFieldProblem(
+			"claimable_balance_id",
+			fmt.Errorf("Invalid claimable balance ID"),
+		)
+	}
+	return balanceID, nil
+}
```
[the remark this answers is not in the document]
2opremio: We still need it because we need to fetch it (the validation just makes sure it's correct)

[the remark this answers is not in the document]
2opremio: I will look into sharing the code, although it may not be worth it.

services/horizon/internal/actions/operation.go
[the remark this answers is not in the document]
bartekn: I'm specifically thinking about `Validate` method: https://github.com/stellar/go/pull/3483/files#diff-e72d127b557b53946e05bb68a3b9db8060418c54219f7782236be9708b7f2740R80.

services/horizon/internal/db2/history/history_claimable_balances.go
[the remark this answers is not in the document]
2opremio: > But we won't be able to change the code here so easily because it requires history reingestion.

I think that's out of the scope of this PR. That said, I don't think it would required a full reingestion. We could just write a migration (@paul and I checked and it seems you can convered base64 to hex).

> I think it's worth it because this PR alone will be literally hundreds lines of code smaller.

I am not sure I agree, we used hex-encoding before and the change didn't remove hundreds of lines 571d47e6f35eeb27bc2d32dd399f2045c4f7d940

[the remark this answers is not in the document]
bartekn: > We could just write a migration (@paul and I checked and it seems you can convered base64 to hex).

That would be cool.

> I am not sure I agree, we used hex-encoding before and the change didn't remove hundreds of lines

I was estimating the code responsible for decoding hex / encoding base64 for DB query purposes and XDR decoding used in validation. But yeah, maybe it will sum to hundred not hundreds.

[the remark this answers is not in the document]
2opremio: OK, we have decided to use `hex` for the new table (`history_claimable_balances` introduced in this PR) and later use a migration to change the representation in `claimable_balances`

For the record, @paulbellamy thinks this should be possible through `select encode(decode('abcd-base64-here', 'base64'), 'hex');`

services/horizon/internal/db2/schema/migrations/46_convert_claimable_balance_ids_to_hex.sql
```diff
@@ -0,0 +1,10 @@
+-- +migrate Up
+
+UPDATE claimable_balances SET id = encode(decode(id, 'base64'), 'hex');
+UPDATE history_claimable_balances SET claimable_balance_id = encode(decode(claimable_balance_id, 'base64'), 'hex');
```
2opremio, line 4: Why is this necessary? (we are creating this table from scratch in `45_`) which hasn't yet been deployed anywhere.

Also we could (optionally) merge this file with the `45_` one.

And, finally, I think we need an explicit test for this (i.e. making sure a `claimable_balances` table with the old format is successfully migrated and used in the new format).

services/horizon/internal/db2/schema/migrations/45_add_claimable_balances_history.sql
```diff
@@ -0,0 +1,56 @@
+-- +migrate Up
+
+UPDATE claimable_balances SET id = encode(decode(id, 'base64'), 'hex');
```
bartekn, line 3: I think we should remove this:
1. The state will be rebuilt anyway because of `CurrentVersion` bump.
2. In case of a rolling deployment, if another ingesting instance starts verifying state after this migration it will be marked invalid (unnecessary alerts etc.).

2opremio: I am confused (probably due to may shallow knowledge of the ingestion system), do you mean that an explicit data migration isn't needed and that the migration will happen implicitly because we rebuild all the _state_ after an upgrade?

Could you shed some light (for my own learning) on what _state_ means in this case? (And in particular, what tables are considered part of the state?)

tamirms: When the ingestion system starts up one of the checks it performs is to see if the recorded ingestion version in the horizon db matches `CurrentVersion`:

https://github.com/stellar/go/blob/master/services/horizon/internal/ingest/fsm.go#L161-L200

If there's a mismatch, ingestion will transition to the build state. In the build state we truncate all state tables:

https://github.com/stellar/go/blob/master/services/horizon/internal/ingest/fsm.go#L303

https://github.com/stellar/go/blob/ec5600bd6b2b6900d26988ff670b9ca7992313b8/services/horizon/internal/db2/history/ingestion.go#L8

Ingested data is divided into two categories: state and historical.

State data is basically a cumulative aggregation. An example of state data would be the current balance of an account. State data is derived by applying ChangeProcessors for every `ingest.Change` generated by a ledger:

https://github.com/stellar/go/blob/master/services/horizon/internal/ingest/processors/change_processors.go

In order to keep state data up to date we always need to feed in the latest changes from each ledger into the ChangeProcessors, and the change processors mutate entries in the state tables so that the data is up to date.

However, when we truncate all the state tables we do not need to ingest all the changes for every single ledger starting from the genesis ledger. Instead, we can get all the ledger entries in stellar  from a recent history archive snapshot checkpoint. That is what the the "build state" does in the ingestion system.

The other type of data that horizon ingests is historical data. Historical data is not cumulative. Historical data is an  immutable time series. Examples of historical data are transactions, ledgers, operations, and effects. 

Once a particular effect is ingested into the horizon db we never update or remove that entry. The only case where we need to update / remove historical data is if we have a bug in horizon ingestion which introduce faulty data. In that case, we need to reingest the affected ledgers.

Historical ingestion data is derived from TransactionProcessors:

https://github.com/stellar/go/blob/master/services/horizon/internal/ingest/processors/transactions_processor.go

services/horizon/internal/db2/schema/migrations/45_add_claimable_balances_history.sql
```diff
@@ -0,0 +1,56 @@
...
+
+-- TODO: Do we need to start from existing count(claimable_balances) and backfill?
```
tamirms, line 5: no, I don't think we need to backfill. when we start ingesting ledgers with the ClaimableBalancesTransactionProcessor it will start adding the missing entries in history_claimable_balances for every claimable balance it sees

services/horizon/internal/db2/history/history_claimable_balances.go
```diff
@@ -0,0 +1,158 @@
...
+// HistoryClaimableBalance is a row of data from the `history_claimable_balances` table
+type HistoryClaimableBalance struct {
+	BalanceID  xdr.ClaimableBalanceId `db:"claimable_balance_id"`
+	InternalID int64                  `db:"id"`
+}
+
+var selectHistoryClaimableBalance = sq.Select("hcb.*").From("history_claimable_balances hcb")
+
+// ClaimableBalancesByIDs loads rows from `history_claimable_balances`, by claimable_balance_id
+func (q *Q) ClaimableBalancesByIDs(dest interface{}, ids []xdr.ClaimableBalanceId) error {
```
tamirms, line 85: I think it would be better if this function and ClaimableBalanceByID returned the `HistoryClaimableBalance` values instead of passing pointers as `dest interface{}` parameters

services/horizon/internal/db2/schema/migrations/45_add_claimable_balances_history.sql
2opremio: Fantastic explanation. Thanks Tamir.

However, as I understand it, the state reset won't happen when ingestion isn't enabled.

Doesn't this mean that querying instances (at SDF, temporarily, or elsewhere where, for whatever reason ingestion isn't enabked) would break without the explicit database migration?

bartekn: > Doesn't this mean that querying instances (at SDF, temporarily, or elsewhere where, for whatever reason ingestion isn't enabled) would break without the explicit database migration?

In case of a rolling deployment it depends on the order. If any of ingesting instances is upgraded first, then there's no problem because web instances will detect that state is rebuild and will return `Still Ingesting` error. If web instance is deployed first then you are right that users will get `500 Internal Server Error` due to an error in decoding IDs.

However, even the later is better that updating IDs via migration because migration can affect state verifier which runs on ingesting instances every 5 mins. The following sequence is possible:
1. State verifier starts on instance A.
2. Instance B migrates a DB.
3. Instance A starts `REPEATABLE READ` tx and sees updated (hex) IDs.
4. Decoding would error and (need to check but quite possibly) state will be marked as invalid.

It's not end of the world but requires some manual steps to recover.

2opremio: Alright, let's remove the migration. I think we should probably add some release notes on the possibility of getting 500s before the state is truncated.

2opremio:
OK, this should be good to go.

I also added some notes in the CHANGELOG, about the new endpoints and the potential 500 errors in `GET /claimable_balances` until any ingestion node is upgraded. 

I am still slightly worried about a potential usecase in which ingestion is not enabled at all (leading to perpetual 500 errors) though but @bartekn 's analysis doesn't consider it a big problem.

services/horizon/CHANGELOG.md
```diff
@@ -9,7 +9,10 @@ file. This project adheres to [Semantic Versioning](http://semver.org/).
...
-* Add a flag `--captive-core-peer-port`/`CAPTIVE_CORE_PEER_PORT` that allows users to control which port the Captive Core subprocess will bind to for connecting to the Stellar swarm. ([3483](https://github.com/stellar/go/pull/3484)). 
+* Add a flag `--captive-core-peer-port`/`CAPTIVE_CORE_PEER_PORT` that allows users to control which port the Captive Core subprocess will bind to for connecting to the Stellar swarm. ([3483](https://github.com/stellar/go/pull/3484)).
+* Add 2 new HTTP endpoints `GET claimable_balances/{id}/transactions` and `GET claimable_balances/{id}/operations`, which respectively return the transactions and operations related to a provided Claimable Balance Identifier `{id}`.
+* Internal DB represenatation change: the `claimable_balances` table now represents the claimable balance identifiers as an hexadecimal string (as opposed to base64). The migration will be performed by the ingestion system, thus, if some of your Horizon nodes are not ingestors (i.e. no `--ingestion` flag enabled) you may experience 500s in the `GET /claimable_balances/` requests until an ingestion node is upgraded.
```
tamirms, line 14: it's worth putting this in bold. Also, it's worth noting that the rebuild process will take several minutes and no new ledgers will be ingested until the rebuild is finished.

clients/horizonclient/operation_request.go
```diff
@@ -27,6 +27,10 @@ func (op OperationRequest) BuildURL() (endpoint string, err error) {
 	if op.ForAccount != "" {
 		endpoint = fmt.Sprintf("accounts/%s/%s", op.ForAccount, op.endpoint)
 	}
+	if op.ForClaimableBalance != "" {
+		// The only endpoint which makes sense for claimable balances is /operations (so, we hardcode it)
+		endpoint = fmt.Sprintf("claimable_balances/%s/operations", op.ForClaimableBalance)
```
tamirms, line 32: nit: `op.endpoint` should be passed in instead of hardcoding "operations" to be consistent with the other code in this function

services/horizon/internal/actions/operation.go
```diff
@@ -93,6 +95,12 @@ func (handler GetOperationsHandler) GetResourcePage(w HeaderWriter, r *http.Requ
 	switch {
 	case qp.AccountID != "":
 		query.ForAccount(qp.AccountID)
+	case qp.ClaimableBalanceID != "":
+		cbID, parseErr := balanceIDHex2XDR(qp.ClaimableBalanceID, "claimable_balance_id")
+		if parseErr != nil {
+			return nil, parseErr
+		}
+		query.ForClaimableBalance(cbID)
```
bartekn, line 103: Now when we have everything in hex maybe we should just use `string` in DB queries? Then we will be able to remove remove `balanceIDHex2XDR` entirely. The only difference is that for invalid IDs users will see 404 error instead of 500 error but that's fine.

xdr/db.go
```diff
@@ -45,12 +45,12 @@ func (t Asset) Value() (driver.Value, error) {
 
 // Scan reads from src into a ClaimableBalanceId
```
bartekn, line 46:
```suggestion
// Scan reads from src into a ClaimableBalanceId
// This method decodes values from hex.
```

xdr/db.go
```diff
@@ -45,12 +45,12 @@ func (t Asset) Value() (driver.Value, error) {
...
 func (c *ClaimableBalanceId) Scan(src interface{}) error {
-	return safeBase64Scan(src, c)
+	return safeHexScan(src, c)
 }
 
 // Value implements the database/sql/driver Valuer interface.
```
bartekn, line 51:
```suggestion
// Value implements the database/sql/driver Valuer interface.
// This method encodes values to hex.
```

bartekn (approved):
LGTM! I think it's worth exploiting the fact the we use hex for both HTTP API and DB so we can `string` from request query directly (without decoding into XDR value).

services/horizon/internal/ingest/processors/claimable_balances_transaction_processor.go
```diff
@@ -0,0 +1,289 @@
...
+
+	return nil
+}
+
+func claimableBalancesForTransaction(
+	sequence uint32,
+	transaction ingest.LedgerTransaction,
+) ([]xdr.ClaimableBalanceId, error) {
+	cbs := []xdr.ClaimableBalanceId{}
+	c, err := claimableBalancesForMeta(transaction.Meta)
```
tamirms, line 83: I think it would be better to get the changes using `transaction.GetChanges()` instead of inspecting `transaction.Meta` but @bartekn would know best

bartekn: @tamirms great catch! Yes, `transaction.Meta.Operations` will likely be empty because we're at V2 right now so the proper way would be to get `transaction.Meta.V2.Operations`. So yes, we should use `transaction.GetChanges()` and maybe we should make `LedgerTransaction.Meta` private (in a separate PR).

bartekn: https://github.com/stellar/go/issues/3491

paulbellamy: This whole bit was a bit belt-and-braces. Is there a way for a claimable balance to be related to a transaction that wouldn't be part of the operations?

bartekn: No but the problem is that `xdr.TransactionMeta` can be one of 3 versions. Using `xdr.TransactionMeta.Operations` is v0 and is empty for some time (left there for backwards compatibility reasons). The version we are using now is V2 and operations' meta can be found in `xdr.TransactionMeta.V2.Operations`.

paulbellamy: Ok, so we don't need any of this bit of code then..

services/horizon/internal/actions/operation.go
paulbellamy: I agree, but let's do this later?

bartekn: Sounds good.

2opremio:
Good to go?

services/horizon/internal/ingest/processors/operations_processor.go
```diff
@@ -689,3 +689,41 @@ func operationsParticipants(transaction ingest.LedgerTransaction, sequence uint3
 
 	return participants, nil
 }
+
+// ClaimableBalances returns the accounts taking part in the operation.
```
tamirms, line 693: I think this is now dead code

tamirms (approved):
🎉

PR #3484, opened 2021-03-23
services/horizon: Add flag to choose a custom peering port for Captive Core

### LINKED ISSUE #3441: Captive core peer port can clash with other stellar core instances running on the machine
On stellar core the peer port is configured to 11625 by default. When captive core is launched it will fail to start if the peer port is occupied by another process (e.g. a stellar core instance running on the same host). To mitigate this we could either disable the peer port if that's possible or we could select a random port which is available on the host.

Shaptic:
### What
This introduces the `--captive-core-peer-port=11625` parameter, which allows users to control which port Captive Core binds to when joining the Stellar network swarm.

### Why
Closes #3441.

### Known limitations
No validation is done on the actual value passed in (aside from type == uint), so non-port values work. However, then Core will fail to start (with an appropriate error message, of course). We could try to "fail fast" instead, but we don't do this for the HTTP port either.

### DIFF
--- a/ingest/ledgerbackend/stellar_core_runner.go
+++ b/ingest/ledgerbackend/stellar_core_runner.go
@@ -54,6 +54,7 @@ type stellarCoreRunner struct {
 	networkPassphrase string
 	historyURLs       []string
 	httpPort          uint
+	peerPort          uint
 	mode              stellarCoreRunnerMode
 
 	started      bool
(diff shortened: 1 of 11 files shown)


ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -153,6 +159,8 @@ func (r *stellarCoreRunner) generateConfig() (string, error) {
 	}
 	result += strings.Join(lines, "\n")
 
+	fmt.Println(result)
```
tamirms, line 162: do we need to print this?

tamirms (approved):

ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -153,6 +159,8 @@ func (r *stellarCoreRunner) generateConfig() (string, error) {
...
+
```
bartekn, line 163: What @tamirms said:
```suggestion
```

bartekn (approved):
👍 Please update CHANGELOG.

ingest/ledgerbackend/stellar_core_runner.go
Shaptic: oops, I knew I left it somewhere

PR #3493, opened 2021-03-24
Point STELLAR_CORE_VERSION at private test build

rokopt:
This draft PR is not for merging, but for running integration tests on a private test build.


2opremio:
I think we can close this, it already served its purpose.

Issue #3495, opened 2021-03-24
Feature Request: Improve services/regulated-assets-approval-server tests


Zagan202:
### What would you like to see?

A few outstanding items 
- Use cases for the `ServeHTTP` for both `json` and `form` and also all edge cases for `h.isRejected` tests the same parts of the code three times. 
    - Have full coverage of all edge cases for the `json`. 
    - Cover just the happy path for the other two options. 

- Add a unit test for the TOML file.

Zagan202:
Tests were improved in prior PRs
https://github.com/stellar/go/pull/3595
https://github.com/stellar/go/pull/3592
https://github.com/stellar/go/pull/3590

PR #3499, opened 2021-03-25
services/regulated-assets-approval-server: SEP-8 approval server should revise transaction payments for happy path

Zagan202:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Revise the transaction, going from:
```
Operation 1: Payment from A to B
```
into:
```
Operation 1: AllowTrust op where issuer fully authorizes account A, asset X
Operation 2: AllowTrust op where issuer fully authorizes account B, asset X
Operation 3: Payment from A to B
Operation 4: AllowTrust op where issuer fully deauthorizes account B, asset X
Operation 5: AllowTrust op where issuer fully deauthorizes account A, asset X
```

- Adds response [revise response](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0008.md#revised)
- Updates TOML with relevant criteria changes

Closes: https://github.com/stellar/stellar-demo-wallet/issues/120
### Why

Focus on landing a code that addresses the happy path scenario where a transaction that meets the criteria is sandwiched by the server.

### Known limitations

Not a limitation but there is some [discussion](https://stellarfoundation.slack.com/archives/C01D4MW5DPT/p1616605600010400) on whether or not the server should be creating the sandwich transaction at all.

For a SEP-8 reference server we should build and demo an implementation that the SDF and the stellar community agrees is the best(or good enough) representation of how a regulated asset server/anchor should behave.


services/regulated-assets-approval-server/serve/tx_approve.go
```diff
@@ -97,12 +129,98 @@ func (h txApproveHandler) isRejected(ctx context.Context, in txApproveRequest) (
...
+		Code:   h.assetCode,
+		Issuer: h.issuerKP.Address(),
+	}
+
+	tx, err = txnbuild.NewTransaction(txnbuild.TransactionParams{
+		SourceAccount:        &txnbuild.SimpleAccount{AccountID: tx.SourceAccount().AccountID},
+		IncrementSequenceNum: true,
+		Operations: []txnbuild.Operation{
+			&txnbuild.AllowTrust{
+				Trustor:   tx.SourceAccount().AccountID,
```
marcelosalloum, line 184: ❕ The transaction source account and the payment source account could be different, which means the sender address is the payment source account (if it's not empty). We only default to the transaction source account if payment.SourceAccount is empty.

services/regulated-assets-approval-server/serve/tx_approve.go
```diff
@@ -97,12 +129,98 @@ func (h txApproveHandler) isRejected(ctx context.Context, in txApproveRequest) (
...
+				Type:      asset,
+				Authorize: false,
+			},
+			&txnbuild.AllowTrust{
+				Trustor:   op.Destination,
+				Type:      asset,
+				Authorize: false,
+			},
```
marcelosalloum, line 203: ❓  Will this succeed if submitted to the ledger? I had the impression we should disallow addresses in the reverse order we allowed them, as in https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0008.md#transaction-composition. 

I never tried executing the operations in this order though.

services/regulated-assets-approval-server/serve/tx_approve.go
```diff
@@ -97,12 +129,98 @@ func (h txApproveHandler) isRejected(ctx context.Context, in txApproveRequest) (
...
+	tx, err = tx.Sign(h.networkPassphrase, h.issuerKP)
+	if err != nil {
+		log.Ctx(ctx).Error(errors.Wrap(err, "signing transaction"))
+		return nil, NewHTTPError(http.StatusBadRequest, `Failed to sign transaction.`)
+	}
+
+	txEnc, err := tx.Base64()
+	if err != nil {
+		log.Ctx(ctx).Error(errors.Wrap(err, "unable to serialize tx"))
+		return nil, NewHTTPError(http.StatusBadRequest, `unable to serialize tx.`)
```
marcelosalloum, line 222: 💡  In these cases the problem seems more like a 500 internal error than a bad request. It's nice to log things here but returning a 400 would imply the client did something wrong, which probably isn't the case here.

services/regulated-assets-approval-server/serve/tx_approve.go
```diff
@@ -16,36 +17,51 @@ type sep8Status string
...
 	}
+
+}
+
+func NewRevisedTXApproveResponse(message string, tx string) *txApproveResponse {
+	return &txApproveResponse{
+		Status:      Sep8StatusRevised,
+		Message:     message,
+		Transaction: tx,
+	}
```
marcelosalloum, line 64: 💡  If the happy path message is always the same you could hardcode it and simplify this method signature to receive just the txe.

Also, a more intuitive name IMO could be tx -> txe | txXDR | xdr.

services/regulated-assets-approval-server/serve/tx_approve_test.go
```diff
@@ -170,11 +242,90 @@ func TestTxApproveHandler_isRejected(t *testing.T) {
...
+	kp01 := keypair.MustRandom()
+	kp02 := keypair.MustRandom()
+	tx, err := txnbuild.NewTransaction(
+		txnbuild.TransactionParams{
+			SourceAccount:        &txnbuild.SimpleAccount{AccountID: kp01.Address()},
+			IncrementSequenceNum: true,
+			Operations: []txnbuild.Operation{
+				&txnbuild.Payment{
+					SourceAccount: kp01.Address(),
+					Destination:   kp02.Address(),
```
marcelosalloum, line 264: Taking this as an example, you could have different accounts as the tx.SourceAccount and the payment.SourceAccount.

The tx Source account is the one paying the transaction fees and it becomes the operation source account automatically for any operation that doesn't have their source account set.

services/regulated-assets-approval-server/serve/tx_approve_test.go
```diff
@@ -170,11 +242,90 @@ func TestTxApproveHandler_isRejected(t *testing.T) {
...
+	assert.Equal(t, op4.Trustor, kp01.Address())
+	assert.Equal(t, op4.Type.GetCode(), assetGOAT.GetCode())
+	require.False(t, op4.Authorize)
+
+	// Check Operation 5: AllowTrust op where issuer fully deauthorizes account A, asset X.
+	op5, ok := tx.Operations()[4].(*txnbuild.AllowTrust)
+	require.True(t, ok)
+	assert.Equal(t, op5.Trustor, kp02.Address())
+	assert.Equal(t, op5.Type.GetCode(), assetGOAT.GetCode())
+	require.False(t, op5.Authorize)
```
marcelosalloum, line 328: Awesome test coverage 🙌  🎉  🚀

services/regulated-assets-approval-server/serve/tx_approve_test.go
```diff
@@ -398,15 +550,23 @@ func TestTxApproveHandler_serveHTTPJson(t *testing.T) {
...
+
+	_, exists := response["status"]
+	require.NoError(t, err)
+	assert.True(t, exists)
+	assert.Equal(t, response["status"], "revised")
+	_, exists = response["status"]
+	require.NoError(t, err)
+	assert.True(t, exists)
+	assert.Equal(t, response["message"], "Authorization and deauthorization operations were added.")
+	_, exists = response["tx"]
```
marcelosalloum, line 567: ❓  Can you add a test to validate the tx returned here? Or is the value of tx going to change in every test execution?

marcelosalloum (commented):
Great job! I've added a few suggestions below, but nothing serious ⬇️

services/regulated-assets-approval-server/serve/tx_approve.go
```diff
@@ -97,12 +129,98 @@ func (h txApproveHandler) isRejected(ctx context.Context, in txApproveRequest) (
...
+	// Check if transaction has only one operation.
+	if len(tx.Operations()) > 1 {
+		log.Ctx(ctx).Errorf("Transaction has %d operations.", len(tx.Operations()))
+		return NewRejectedTXApproveResponse(exceededOpsLenErr), nil
+	}
+
+	// Check if operation is a payment.
+	op, ok := tx.Operations()[0].(*txnbuild.Payment)
+	if !ok {
+		log.Ctx(ctx).Errorf("Transaction contains a %q operation.", reflect.TypeOf(op))
```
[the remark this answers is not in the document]
Zagan202: I really struggled trying to figure out how to do the go equivalent of python's built-in [type()](https://docs.python.org/3/library/functions.html#type) function.
Thanks for the tip!

services/regulated-assets-approval-server/serve/tx_approve.go
```diff
@@ -97,12 +129,98 @@ func (h txApproveHandler) isRejected(ctx context.Context, in txApproveRequest) (
...
+	}
+
+	return nil, nil
+}
+
+func (h txApproveHandler) Approve(ctx context.Context, in txApproveRequest) (*txApproveResponse, error) {
+	// Decode the request transaction.
+	parsed, err := txnbuild.TransactionFromXDR(in.Transaction)
+	if err != nil {
+		log.Ctx(ctx).Error(errors.Wrapf(err, "Parsing transaction %s failed", in.Transaction))
```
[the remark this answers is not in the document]
Zagan202: Thanks for the reminder 😅

services/regulated-assets-approval-server/serve/tx_approve.go
```diff
@@ -97,12 +129,98 @@ func (h txApproveHandler) isRejected(ctx context.Context, in txApproveRequest) (
...
+	if !ok {
+		log.Ctx(ctx).Errorf("Transaction %s is not a simple transaction.", in.Transaction)
+		return nil, NewHTTPError(http.StatusBadRequest, `Transaction submitted is not a simple transaction.`)
+	}
+
+	op, ok := tx.Operations()[0].(*txnbuild.Payment)
+	if !ok {
+		log.Ctx(ctx).Errorf("Transaction contains a %q operation.", reflect.TypeOf(op))
+		return nil, NewHTTPError(http.StatusBadRequest, `Transaction contains is not a Payment operation.`)
+	}
```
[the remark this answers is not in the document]
Zagan202: I think I like the idea of calling `Approve` inside of `isRejected`.
It would require a bit of refactoring but its totally doable. 

❓ Would you be open to having that be a separate PR?

services/regulated-assets-approval-server/serve/tx_approve.go
Zagan202: Oh this is a good catch I didn't consider. 
Let me adjust this.

services/regulated-assets-approval-server/serve/tx_approve.go
Zagan202: I'll be honest I've never tried submitting these transactions on the network.

🧠 However I think this sparks a good feature request idea for this [issue](https://github.com/stellar/go/issues/3495).

I can write a test that would mock a client(acting like a wallet) where it makes calls to the SEP8 ref server and attempts to submit them to the network.

Zagan202: 👍🏾 Also I think its a good idea to reverse the order of the disallow ops to be as consistent with the SEP as you pointed out.

Zagan202: In theory it shouldn't make a difference (as far as I understand it) because the transactions are ATOMIC.
So I'm unsure of any documented barriers that would prevent the current way its crafted to be submitted.

services/regulated-assets-approval-server/serve/tx_approve.go
Zagan202: Great point I'll change that :)

services/regulated-assets-approval-server/serve/tx_approve.go
Zagan202: 👍🏾 I agree that `tx` should be  `xdr` to make it much more descriptive of what format it should be.

Zagan202: > If the happy path message is always the same you could hardcode it and simplify this method signature to receive just the txe.

I think I would like to keep it modular for now in case new requirements make it so we need to pass other messages and such.

services/regulated-assets-approval-server/serve/tx_approve_test.go
Zagan202: I just tried it without the random kp I generate (giving the transaction static account ids and a static `txnbuild.SimpleAccount`) and the marshaled Base64 rep of the XDR changes each time.

I think just checking if it exists is good enough for the time being.

services/regulated-assets-approval-server/serve/tx_approve.go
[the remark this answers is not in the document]
Zagan202: Addressed here 2de2647

services/regulated-assets-approval-server/serve/tx_approve.go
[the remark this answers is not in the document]
Zagan202: Addressed here 90c4e12

services/regulated-assets-approval-server/serve/tx_approve.go
Zagan202: Addressed here 
c41562c

services/regulated-assets-approval-server/serve/tx_approve.go
Zagan202: Addressed here 
73d3674

services/regulated-assets-approval-server/serve/tx_approve.go
Zagan202: Addressed here 
1b925e6

services/regulated-assets-approval-server/serve/tx_approve.go
Zagan202: Rename addressed here 0d1b665

services/regulated-assets-approval-server/serve/tx_approve_test.go
Zagan202: c41562c

services/regulated-assets-approval-server/serve/tx_approve_test.go
Zagan202: rename addressed here 0d1b665

services/regulated-assets-approval-server/serve/tx_approve.go
```diff
@@ -79,30 +107,124 @@ func (h txApproveHandler) isRejected(ctx context.Context, in txApproveRequest) (
...
+
+	// Check if transaction's operation sourceaccount is the same as the server issuer account.
+	if op.GetSourceAccount() == h.issuerKP.Address() {
+		log.Ctx(ctx).Errorf("transaction's operation sourceaccount %q is the same as issuer account", op.GetSourceAccount())
+		return NewRejectedTXApproveResponse(unauthorizedOpErr), nil
+	}
+
+	// Check if transaction's sequence number is the same as the user account sequence number + 1
+	srcAccount := &txnbuild.SimpleAccount{AccountID: op.GetSourceAccount()}
+	srcAccountSeq, _ := srcAccount.GetSequenceNumber()
```
marcelosalloum, line 149: ⁉️ wait, this sequence number will always be zero! You're not gathering any information from the ledger there, you're just instantiating a `txnbuild.SimpleAccount` object with Sequence equals zero:

https://github.com/stellar/go/blob/6145dda1700f23542aeecd4202d518092b2bff29/txnbuild/simple_account.go#L4-L7

marcelosalloum: Also, plz make sure to write tests that fail when the sequence number is inconsistent with this rule.

services/regulated-assets-approval-server/serve/tx_approve.go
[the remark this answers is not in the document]
marcelosalloum: Hum, I don't think that's such a big of a change (please correct me if I'm wrong).

Also, having a repeated logic violates the [DRY principle](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself), making the code redundant, harder to maintain and less predictable.

services/regulated-assets-approval-server/serve/tx_approve.go
marcelosalloum: ⁉️  I don't think your implementation addresses what I meant. You probably forgot the part where I say the payment source account can be empty and we need to default to the transaction source account in that case:
* The payment source can be achieved from `paymentOp.SourceAccount`, **but that could be empty**.
* If the payment source account is empty, the ledger will default to the transaction source account.

So what you actually need to do is something like:
```go
sourceAccount := paymentOp.GetSourceAccount()
if sourceAccount == "" {
  sourceAccount = tx.GetSourceAccount()
}
```

services/regulated-assets-approval-server/serve/tx_approve.go
marcelosalloum: I'd get as close to the SEP spec as possible. Also, you can try using stellar laboratory to check if your order would work.

Zagan202:
Will address these changes and suggestion during https://github.com/stellar/nebula/issues/182

PR #3504, opened 2021-03-26
all: remove stellar_core_commander and update bundler version

### LINKED ISSUE #3503: Update ruby and Rakefile scripts to be ruby 3.2 compatible
The ruby/Rakefile scripts and their bundler requirements require bundler 1.17.2. When running bundler 1.17.2 on Ruby 3 I see this warning message:

```
/home/vscode/.rvm/gems/ruby-3.0.0/gems/bundler-1.17.2/lib/bundler/shared_helpers.rb:118: warning: Pathname#untaint is deprecated and will be removed in Ruby 3.2.
```

We should probably upgrade to bundler 2 at some point so that bundling doesn't break in the future.

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Remove stellar_core_commander and update bundler version and xdrgen.

### Why

Mostly to fix xdrgen in this repo for new Ruby versions.

Stellar_core_commander has really old dependencies which are no longer supported on the latest version of Ruby. Also, according to [internal discussions](https://stellarfoundation.slack.com/archives/C02B04RMK/p1616797881149900?thread_ts=1616797044.148000&cid=C02B04RMK) we don't actually use stellar_core_commander anymore.

Bundler is being updated because v1.17.2 of bundler calls functions that are deprecated and soon to be removed from Ruby. Using bundler in this repo outputs warnings saying this.

Close #3503 

### Known limitations

N/A


bartekn:
Do we actually need the ruby stuff here? It looks like after this change it will only call `xdrgen` we can just run from CLI.

leighmcculloch:
I think we do unless we make more changes. The `Rakefile` file uses Octokit to download the files and hooks everything together. I think we could simplify this even more by moving that logic into the `Makefile` and replacing Octokit with `curl`. I'm happy to do that, but will do that in a separate PR.

leighmcculloch:
@bartekn To clarify this PR doesn't change how xdrgen gets used, so the same ruby scripts exist and it still gets generated the same way. This PR removes stellar_commander from the bundle because it is old and pulled in old dependencies that were causing problems. I think what you're proposing is good, but is a different and larger change than this.

bartekn (approved):

PR #3506, opened 2021-03-27
verify-range: Bump base branch to Horizon 2.0

2opremio:


2opremio:
@bartekn how was the image supposed to pushed? is it still a manual process?

bartekn (approved):

bartekn:
Yes, the process is manual but if it's a problem you can push the image to any other account and use the new image in tests.

PR #3513, opened 2021-03-29
Only consistency-check the Captive Core flags when captive-core ingestion is enabled

### LINKED ISSUE #3507: horizon: captive-core flags shouldn't be checked when captive core ingestion is disabled
The following invocation of horizon shouldn't error:

```
export NETWORK_PASSPHRASE="Public Global Stellar Network ; September 2015"
export HISTORY_ARCHIVE_URLS="https://s3-eu-west-1.amazonaws.com/history.stellar.org/prd/core-live/core_live_001"
export DATABASE_URL="postgres://postgres:postgres@localhost:5432/horizon?sslmode=disable"
export ENABLE_CAPTIVE_CORE_INGESTION="false"
export INGEST_FAILED_TRANSACTIONS=true
export STELLAR_CORE_URL="http://localhost:11626"
export STELLAR_CORE_DATABASE_URL="postgres://postgres:postgres@localhost:5432/core?sslmode=disable"
export INGEST="true"
horizon ingest verify-range --from 1 --to 64 --verify-state

2021/03/27 19:50:11 Invalid config: stellar-core-binary-path = /Users/fons/stellar-core/src/stellar-core, but corresponding option captive-core-config-append-path is not configured
```

It only works after setting `CAPTIVE_CORE_CONFIG_APPEND_PATH` to a phony value (e.g. `CAPTIVE_CORE_CONFIG_APPEND_PATH=/dev/null`).

This is an ugly workaround the user shouldn't go through.

2opremio:
Fixes #3507

### DIFF
--- a/services/horizon/internal/flags.go
+++ b/services/horizon/internal/flags.go
@@ -449,20 +449,6 @@ func ApplyFlags(config *Config, flags support.ConfigOptions) {
 	validateBothOrNeither("tls-cert", "tls-key")
 
 	if config.Ingest {
-		binaryPath := viper.GetString(StellarCoreBinaryPathName)
-
-		// If the user didn't specify a Stellar Core binary, we can check the
-		// $PATH and possibly fill it in for them.
-		if binaryPath == "" {
-			if result, err := exec.LookPath("stellar-core"); err == nil {
-				binaryPath = result
-				viper.Set(StellarCoreBinaryPathName, binaryPath)
-				config.CaptiveCoreBinaryPath = binaryPath
-			}
-		}
-
-		// When running live ingestion a config file is required too
-		validateBothOrNeither(StellarCoreBinaryPathName, CaptiveCoreConfigAppendPathName)
 
 		// config.HistoryArchiveURLs contains a single empty value when empty so using
 		// viper.GetString is easier.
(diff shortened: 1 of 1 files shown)


services/horizon/internal/flags.go
```diff
@@ -471,6 +457,21 @@ func ApplyFlags(config *Config, flags support.ConfigOptions) {
...
 		if config.EnableCaptiveCoreIngestion {
+			binaryPath := viper.GetString(StellarCoreBinaryPathName)
+
+			// If the user didn't specify a Stellar Core binary, we can check the
+			// $PATH and possibly fill it in for them.
+			if binaryPath == "" {
+				if result, err := exec.LookPath("stellar-core"); err == nil {
+					binaryPath = result
+					viper.Set(StellarCoreBinaryPathName, binaryPath)
+					config.CaptiveCoreBinaryPath = binaryPath
```
2opremio, line 468: I think it would be better to only check the path and use the binary in the path when using the configuration value (i.e. when using captive core). Modifying the configuration taints what we know about what the user provided.

2opremio: (that's for later though)

paulbellamy (approved):

PR #3518, opened 2021-03-30
services/horizon: Removed LedgerBackend calls inside ingest critical section

### LINKED ISSUE #3517: Move LedgerBackend operations outside ingest locked section
In `resumeState` of ingestion state machine we get ledger data from `LedgerBackend` **inside** critical section when lock is acquired (line below between tx start and `LedgerBackend` usage):
https://github.com/stellar/go/blob/48e6dfc9e9b18e3738917cd514a86ad7a0adbb6b/services/horizon/internal/ingest/fsm.go#L367-L426

It should be possible to move the code responsible for getting ledger data outside of DB transaction. In case of `DatabaseBackend`/`CaptiveStellarCore` connection issues, the ingestion lock will be free for other ingesting instances to acquire. This should improve reliability of ingesion.

Other ideas for improving reliability can be found in #3474.

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This commit moves calls to `LedgerBackend` in `buildState` and `resumeState` outside critical section which is a DB transaction acquiring ingestion lock. To clean the code `LedgerBackend.GetLedgerBlocking` method was introduced that always blocks until the ledger is available in the backend. Additionally, `GetLatestLedger` method name was changed to `GetLatestHistoryLedger` for clarity and `ProcessorRunner` methods have changed and do not depend on `LedgerBackend` anymore.

Close #3517.

### Why

In case of access to `LedgerBackend` hangs (ex. broken connection to Core DB or remote captive core) inside a critical section ingestion will halt in entire cluster. With this change only a single ingesting node will halt.

### Known limitations

All of the other states that are reading from `LedgerBackend` were not changed (still access it in critical section). This is due to the fact that multiple ledgers are read (ex. `historyRangeState`) and they may not fit into memory.


ingest/ledger_transaction_reader.go
```diff
@@ -13,13 +13,13 @@ import (
 // LedgerTransactionReader reads transactions for a given ledger sequence from a backend.
 // Use NewTransactionReader to create a new instance.
 type LedgerTransactionReader struct {
-	ledgerCloseMeta xdr.LedgerCloseMeta
```
tamirms, line 16: the definition of `xdr.LedgerCloseMeta` is:

```
type LedgerCloseMeta struct {
	V  int32
	V0 *LedgerCloseMetaV0
}
```

Since the struct isn't so large we will not be saving much in terms of performance. Is there any other reason besides improving performance for changing it to a pointer type?

services/horizon/internal/ingest/fsm.go
```diff
@@ -991,3 +1101,33 @@ func (s *system) maybePrepareRange(from uint32) (bool, error) {
 
 	return false, nil
 }
+
+// getLedgerFromBackend gets ledger from the backend. If range is not prepared
+// it will prepare it.
+func (s *system) getLedgerFromBackend(sequence uint32) (bool, *xdr.LedgerCloseMeta, error) {
```
tamirms, line 1107: When I read `getLedgerFromBackend()` I thought it was strange that we had a wrapper around `ledgerBackend.GetLedger()`. I think if we changed the function signature to the one below it would reduce confusion.

```golang
func prepareAndGet(ledgerBackend  ledgerbackend.LedgerBackend, sequence uint32) (bool, *xdr.LedgerCloseMeta, error) {
}
```

services/horizon/internal/ingest/fsm.go
```diff
@@ -991,3 +1101,33 @@ func (s *system) maybePrepareRange(from uint32) (bool, error) {
...
+
+		log.WithFields(logpkg.F{
+			"ledger":   sequence,
+			"duration": time.Since(startTime).Seconds(),
+		}).Info("Range prepared")
+
+		return false, &xdr.LedgerCloseMeta{}, nil
+	}
+
+	return s.ledgerBackend.GetLedger(sequence)
```
tamirms, line 1132: it might be too much to include in this PR but I think we should consider changing the LedgerBackend API so that we can add an argument to `GetLedger()` which configures whether the call should be blocking or non blocking. Currently, the blocking behavior of `GetLedger()` is configured implicitly depending on whether you call PrepareRange with a bounded or unbounded range. Which makes you have to internalize implementation details of `CaptiveStellarCore` to understand how `GetLedger()` will behave

tamirms: having the blocking configuration passed into `GetLedger()` would also allow us to remove the fast-forward logic from buildState and resume. because we could change `return s.ledgerBackend.GetLedger(sequence)` so that it blocks until it finds ledger `sequence`

services/horizon/internal/ingest/fsm.go
```diff
@@ -376,8 +472,6 @@ func (r resumeState) run(s *system) (transition, error) {
 		return retryResume(r), errors.Wrap(err, getLastIngestedErrMsg)
 	}
 
-	ingestLedger := r.latestSuccessfullyProcessedLedger + 1
-
 	if ingestLedger > lastIngestedLedger+1 {
 		return start(), errors.New("expected ingest ledger to be at most one greater " +
 			"than last ingested ledger in db")
```
tamirms, line 477: I think we need to retryResume if `ingestLedger < lastIngestedLedger`. otherwise we end up ingesting the same ledger more than once when we run `s.runner.RunAllProcessorsOnLedger(ledgerCloseMeta)` below

ingest/ledger_transaction_reader.go
bartekn: Good point, forgot about versioning. Reverted in 172843572d.

services/horizon/internal/ingest/fsm.go
```diff
@@ -570,7 +630,16 @@ func (h historyRangeState) run(s *system) (transition, error) {
 	}
```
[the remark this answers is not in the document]
bartekn: Yes, unfortunately we have to keep access to `LedgerBackend` inside db transaction in this state. Alternatively, we could load all ledgers in a range into memory but for larger ranges it can cause OOM.

services/horizon/internal/ingest/fsm.go
bartekn: This is a valid point. I will try to do it in this PR.

services/horizon/internal/ingest/fsm.go
bartekn: I think we already have this check, one line below your comment, [here](https://github.com/stellar/go/pull/3518/files#diff-33cc075976ed019ba7c91300d22ddb3939d928f34e95951ceef77d39a7779496L384).

tamirms: the code below my comment sets `ingestLedger` to `lastIngestedLedger + 1` but afterwards it proceeds with ingestion; it does not exit with retryResume. before your changes that control flow was ok because afterwards we checked to see if stellar core had `ingestLedger`. But with your code changes, you move that check to before the lock is acquired

bartekn: I can see it now. Great catch!

bartekn: Fixed in 7d7e3b154c.

services/horizon/internal/ingest/fsm.go
```diff
@@ -564,13 +572,25 @@ func (h historyRangeState) run(s *system) (transition, error) {
 		return start(), nil
 	}
 
-	lockReleased, err := s.maybePrepareRange(h.fromLedger)
-	if lockReleased || err != nil {
+	didPrepare, err := s.maybePrepareRange(h.fromLedger)
+	if didPrepare || err != nil {
+		// Release distributed ingestion lock and prepare the range
+		s.historyQ.Rollback()
```
tamirms, line 578: there is a `defer s.historyQ.Rollback()` at the top of the function so we should rollback automatically after the next line when we return out of the function

ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -452,6 +454,18 @@ func (c *CaptiveStellarCore) isPrepared(ledgerRange Range) bool {
 	return false
 }
 
+// GetLedgerBlocking works as GetLedger but will block until the ledger is
+// available in the backend (even for UnboundedRange).
+// Please note that requesting a ledger sequence far after current ledger will
+// block the execution for a long time.
+func (c *CaptiveStellarCore) GetLedgerBlocking(sequence uint32) (xdr.LedgerCloseMeta, error) {
```
tamirms, line 461: given that we have this function, should we make `GetLedger()` always be non blocking?

paulbellamy: This also looks like it's not really threadsafe with `GetLedger()`, given it modifies the underlying thing. Wouldn't "the go way" be to have the default way be blocking, and use a goroutine/channel when you want it async? (When *do* we want it async, anyway?)

services/horizon/internal/ingest/fsm.go
```diff
@@ -644,7 +664,16 @@ func (h reingestHistoryRangeState) ingestRange(s *system, fromLedger, toLedger u
 	}
 
 	for cur := fromLedger; cur <= toLedger; cur++ {
-		if err = runTransactionProcessorsOnLedger(s, cur); err != nil {
+		exists, ledgerCloseMeta, err := s.ledgerBackend.GetLedger(cur)
```
paulbellamy, line 667: Seems odd that this func gets passed `ledgerCloseMeta`, then re-fetches it?

paulbellamy:
To some extent, it seems like the code around `fsm.go`, and the processor mix up concerns around "How do I get ledgers", and "What do I do with the data". Both are combined in the `history.IngestionQ`, which leads the code to mix together their usage. Splitting the ledger fetching methods out of `history.IngestionQ` might help.

I guess the structure I expected to see was something more like: `LedgerReader -> IngestionProcessor -> Database`, then with a top-level loop something like:
```go
db := NewDB()
processor := NewIngestionProcessor(db)
for ledgers.Next() {
  ledger, err := ledgers.Get()
  err = processor.Ingest(ledger) // or Verify(ledger)
}
```

services/horizon/internal/ingest/fsm.go
bartekn: Good point but I realized that the lock should be released sooner (when preparing starts). To solve this I moved `maybePrepareRange` call before starting a DB transaction: ace6341aff.

ingest/ledgerbackend/captive_core_backend.go
bartekn: I agree that we probably should make `GetLedger` blocking by default but before doing this I think we should 1) ensure it doesn't cause unexpected issues (we should be able to tell after testing Horizon release with this PR), 2) we are modifying stable `ingest` package so we should do it via major release. Going to create an issue about it soon.

bartekn:
@paulbellamy I'm not sure I understand.
> seems like the code around fsm.go, and the processor mix up concerns around "How do I get ledgers", and "What do I do with the data". 

I think "How do I get ledgers" concern was really mixed before that PR because `ProcessorRunner` called `GetLedger` of the backend to get ledger data. This PR actually solves this issue, due to different reasons but it seems to be solved. Right now we get a single ledger data in `fsm.go` and pass it to `ProcessorRunner` to process.

> I guess the structure I expected to see was something more like: LedgerReader -> IngestionProcessor -> Database, then with a top-level loop something like:

The code you shared was actually how the previous ingesting system was running (pre 1.0.0) and there was nothing wrong with the main loop in particular. However, because we added state ingestion (from history buckets) and distributed ingestion it complicated this a lot: right now it's possible that when Horizon is ingesting ledgers one-by-one it may escape to another FSM state, ex. to rebuild ledger state because of the upgrade etc. OR it may need to skip one or more ledger ingested by another machine.

Also, the FSM was designed when non-blocking `GetLedger` was used. I _think_ that if we are happy with the blocking version we may want to rewrite `resumeState` to be a loop similar to your snippet. But the rest of the states will need to stay like this.

Please elaborate, maybe I'm missing something.

paulbellamy:
No thats’s pretty close to my understanding now as well. Wasn’t a comment
on this pr in particular, just generally.

bartekn:
Awesome!

services/horizon/internal/ingest/fsm.go
```diff
@@ -958,36 +1001,31 @@ func (s *system) completeIngestion(ledger uint32) error {
...
 
 	if !prepared {
-		// Release distributed ingestion lock and prepare the range
-		s.historyQ.Rollback()
-		log.Info("Released ingestion lock to prepare range")
-		log.WithFields(logpkg.F{"ledger": from}).Info("Preparing range")
+		log.WithFields(logpkg.F{"from": from}).Info("Preparing range")
 		startTime := time.Now()
 
 		err = s.ledgerBackend.PrepareRange(ledgerRange)
```
tamirms, line 1017: `PrepareRange()` already checks if the range is prepared. I think we can get rid of the `maybePrepareRange(from uint32) error` function entirely and just replace all call sites with `s.ledgerBackend.PrepareRange(ledgerRange)`

bartekn:
@tamirms FYI I changed `GetLedger` to `GetLedgerBlocking` in `historyRangeState`: https://github.com/stellar/go/pull/3518/commits/1a06520f3f42518a30e81811c7da9b157aef423a.

services/horizon/internal/ingest/fsm.go
```diff
@@ -559,16 +559,20 @@ func (h historyRangeState) run(s *system) (transition, error) {
...
 	for cur := h.fromLedger; cur <= h.toLedger; cur++ {
-		var exists bool
 		var ledgerCloseMeta xdr.LedgerCloseMeta
-		exists, ledgerCloseMeta, err = s.ledgerBackend.GetLedger(cur)
+
+		log.WithField("ledger", cur).Info("Waiting for ledger to be available in the backend...")
+		startTime := time.Now()
+
+		ledgerCloseMeta, err = s.ledgerBackend.GetLedgerBlocking(cur)
 		if err != nil {
```
tamirms, line 568: I think if we encounter an error when fetching the ledger we can safely commit the work we have already processed.

I would be in favor of calling `s.historyQ.Commit()` before returning out of the function in this case.

tamirms: We could also keep it as non blocking and commit when `!exists || err != nil`. But I don't see a reason why using the non blocking calls are better than using `GetLedgerBlocking()`

services/horizon/internal/ingest/fsm.go
bartekn: Good idea, I implemented it after your suggestion but then realized that we log duration of prepare which can be useful information. If you don't feel strongly about it (I don't) I can leave it as it is now.

services/horizon/internal/ingest/fsm.go
bartekn: Great idea, added in https://github.com/stellar/go/pull/3518/commits/364f7e9dbd3d62395d923d51c6705d4a2173a079.

services/horizon/internal/ingest/fsm.go
```diff
@@ -644,7 +655,16 @@ func (h reingestHistoryRangeState) ingestRange(s *system, fromLedger, toLedger u
 	}
 
 	for cur := fromLedger; cur <= toLedger; cur++ {
-		if err = runTransactionProcessorsOnLedger(s, cur); err != nil {
+		exists, ledgerCloseMeta, err := s.ledgerBackend.GetLedger(cur)
```
tamirms, line 658: should this also be blocking? I think it should be blocking for captive core. But, when using a db ledger backend I guess we expect that the ledger should already be in the db. If that reasoning is correct we could just add a comment to make this blocking once we remove the db ledger backend

tamirms (approved):
I think in a future PR we should make captive core always block on `GetLedger()` because I can't think of any cases in fsm.go where want it to be non-blocking

services/horizon/internal/ingest/fsm.go
bartekn: It doesn't need to use `GetLedgerBlocking` because we prepare a bounded range in `reingestHistoryRangeState` which will make CC blocking. But I agree, we we can make it blocking explicitly after updating ledger backend `GetLedger` to be always blocking (and when we remove DB backend). Not adding a comment because when above is done, it will not be possible to get ledgers in a non-blocking way.

Issue #3443, opened 2021-03-04
Add tool to generate random XDR values


tamirms:
Given an XDR data type (e.g. `LedgerEntryChange`) we would like to generate a set of XDR values which cover all possible structures of the data type (e.g. given an xdr union we want to make sure all branches of a union are represented in the set). The fields within the XDR values should be set to random values.

The primary motivation behind building such a tool is that we could use it to generate a corpus for fuzz testing. We could also potentially integrate this tool into a property testing framework. This tool would be particularly helpful when working on protocol upgrades which require us to update the Stellar XDR definitions (e.g. https://github.com/stellar/go/issues/3430)

Issue #3430, opened 2021-03-01
Automatically test XDR type-switches


2opremio:
In Horizon there are several places where we exhaustively go over union type tags.

e.g. 1 :

```
	switch operation.OperationType() {
	case xdr.OperationTypeCreateAccount:
		foo()
	case xdr.OperationTypeCreateAccount:
		bar()
	...
	}
```

e.g. 2 :

```
	switch effectType {
	case EffectTypeNames[EffectAccountCreated]:
		foo()
	case EffectTypeNames[EffectAccountCredited]:
		bar()
	...
	}
```

When adding new types (e.g. a new operation or effect) it's easy to miss updating some of those switches.

We should strive to create automatic unit tests which notify a `case` entry was missed whenever we add a new type.

2opremio:
For operation types we can use reflection, since we have all the symbol names are automatically generated in `xdr.operationTypeMap `.

For effects we can go over the keys of `effects.EffectTypeNames`.

We should do something similar for all the `history.Effect*` symbols, but we will probably need to add some extra machinery for it (or fix the replication of `effects.Effect*` and `history.Effect*` once and for all).

PR #3523, opened 2021-04-01
xdr: Add random XDR generator

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Close https://github.com/stellar/go/issues/3443

This commit introduces the `randxdr` package to the monorepo. `randxdr` is used for generating XDR values with random data. 

`randxdr` was built using https://github.com/xdrpp/goxdr . Check out this presentation https://drive.google.com/file/d/1NG9aTGlf3yKMfIHdxXeGMpMNMDkOfGre/view to see how goxdr is better than other xdr compilers.

### Why

> The primary motivation behind building such a tool is that we could use it to generate a corpus for fuzz testing. We could also potentially integrate this tool into a property testing framework. This tool would be particularly helpful when working on protocol upgrades which require us to update the Stellar XDR definitions (e.g. #3430)

### Known limitations

1. introduces another dependency to the monorepo ( https://github.com/xdrpp/goxdr )
2. introduces another autogenerated xdr go file ( gxdr/xdr_generated.go ) when we aleady xdr/xdr_generated.go

Counterpoint to (1): I think goxdr has no external dependencies itself.
Counterpoint to (2): I added a script to ensure that  gxdr/xdr_generated.go is always in sync with the xdr/*.x files


randxdr/marshaller.go
```diff
@@ -0,0 +1,121 @@
...
+// Marshal populates a given goxdr.XdrType with random values.
+//
+// Every complex goxdr.XdrType has functions like XdrRecurse() which
+// allow you to visit subfields of the complex type. That is how
+// randMarshaller is able to populate subfields of a complex type with
+// random values.
+//
+// Note that randMarshaller is stateful because of how union types are handled.
+// Therefore Marshal() should not be used concurrently.
+func (rm *randMarshaller) Marshal(field string, i goxdr.XdrType) {
```
paulbellamy, line 53: Nit: I feel like [`filepath.Walk`](https://golang.org/pkg/path/filepath/#Walk) would be a better name and approach, than the `Marshal`/`XdrRecurse` combo, but that's more of a `goxdr` pull request, so, c'est la vie.

randxdr/marshaller.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	switch t := goxdr.XdrBaseType(i).(type) {
+	case goxdr.XdrVarBytes:
+		bound := t.XdrBound()
+		if bound > rm.maxBytesSize {
+			bound = rm.maxBytesSize
+		}
+		bound++
+		bs := make([]byte, rm.rand.Uint32()%bound)
```
paulbellamy, line 65: Could this OOM-kill the tests if it tries to claim 4GB of ram? Good to test though... 🤔

paulbellamy: Ah, no the maxBytesSize is 256. But, should we actually test larger values? Are they possible?

randxdr/marshaller.go
```diff
@@ -0,0 +1,121 @@
...
+			bound = rm.maxVecLen
+		}
+		bound++
+		vecLen := rm.rand.Uint32() % bound
+		t.SetVecLen(vecLen)
+		t.XdrMarshalN(rm, field, vecLen)
+	case goxdr.XdrPtr:
+		present := rm.rand.Uint32()&1 == 1
+		t.SetPresent(present)
+		t.XdrMarshalValue(rm, field)
```
paulbellamy, line 84: Shouldn't we only marshal this if it is present?

randxdr/marshaller.go
```diff
@@ -0,0 +1,121 @@
...
+		// We must rely on these two facts:
+		// * when XdrRecurse() is called on a union, the first field which will be marshalled is the tag field
+		// * the tag field can be one of two types: uint32 or enum
+		if m := t.XdrValidTags(); m != nil {
+			rm.tag = uint32(rm.randomKey(m))
+			rm.useTag = true
+			// The next field the marshaller will visit is the tag field.
+			// Once the tag is set, we need to toggle rm.useTag to false.
+		}
+		t.XdrRecurse(rm, field)
```
paulbellamy, line 115: Nit: You could potentially avoid the statefulness and make this thread-safe by passing a new `randomMarshaller` to this function. Something like:
```suggestion
		t.XdrRecurse(&randomMarshaller{
			tag: uint32(rm.randomKey(m)),
			rand: rm.rand,
			maxBytesSize: rm.maxBytesSize,
			maxVecLen: rm.maxVecLen,
			presets: rm.presets,
		}, field)
```

Then you don't need the `useTag` thing at all.

randxdr/presets.go
```diff
@@ -0,0 +1,99 @@
...
+
+// FieldMatches returns a Selector which matches on a field name by regexp
+func FieldMatches(r *regexp.Regexp) Selector {
+	return func(name string, xdrType goxdr.XdrType) bool {
+		return r.MatchString(name)
+	}
+}
+
+// IsPtr is a Selector which matches on all XDR pointer fields
+var IsPtr Selector = func(name string, xdrType goxdr.XdrType) bool {
```
paulbellamy, line 38: Why vars for these? do they change somewhere?

paulbellamy (approved):

randxdr/generator.go
```diff
@@ -0,0 +1,52 @@
...
+)
+
+// NewGenerator returns a new Generator instance configured with default settings.
+// The returned Generator is deterministic but it is not thread-safe.
+func NewGenerator() Generator {
+	return Generator{
+		MaxBytesSize: DefaultMaxBytesSize,
+		MaxVecLen:    DefaultMaxVecLen,
+		// rand.NewSource returns a source which is *not* safe for concurrent use
+		Source: rand.NewSource(DefaultSeed),
```
bartekn, line 39: Wouldn't this produce deterministic sequence on every run? So in tests in the loop it will test the same cases over and over again? This script prints same values for me on every run:
```go
package main

import (
	"fmt"
	"math/rand"
)

func main() {
	source := rand.NewSource(99)
	r := rand.New(source)
	for i := 0; i < 100; i++ {
		fmt.Println(r.Int())
	}
}
```

.circleci/config.yml
```diff
@@ -78,6 +78,14 @@ commands:
           name: Check generated code
           command: ./gogenerate.sh
 
+  # checkgoxdr validates that any goxdr generated code has been updated if needed.
+  checkgoxdr:
```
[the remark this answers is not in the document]
tamirms: I had that thought as well. The reason I created a separate script is that `gogenerate.sh` was for all workflows that use https://golang.org/pkg/cmd/go/internal/generate/ whereas `checkgoxdr` uses a make file to generate the code. so I wasn't sure if it was appropriate to merge the 2 scripts

randxdr/marshaller.go
tamirms: This is configured by the `MaxBytesSize` field in `Generator`. 256 is an arbitrary default so I'm not sure if it's the best one

bartekn: I agree we should test larger values - up to what's possible (I guess it's MAX_INT64 for xdr). Should be probably set to smaller values in envs with smaller memory (like Circle).

randxdr/marshaller.go
tamirms: the go doc string for `XdrMarshalValue()` is:

	// If the present/not-present bool is false, this function does
	// nothing.  Otherwise, it marshals just the value, not the bit.
	XdrMarshalValue(XDR, string)

services/horizon/internal/ingest/processors/offers_processor_test.go
```diff
@@ -3,16 +3,56 @@
...
+	gen := randxdr.NewGenerator()
+
+	var changes []xdr.LedgerEntryChange
+	for i := 0; i < 1000; i++ {
+		change := xdr.LedgerEntryChange{}
+		shape := &gxdr.LedgerEntryChange{}
+		gen.Next(
+			shape,
+			[]randxdr.Preset{
+				{randxdr.FieldEquals("type"), randxdr.SetU32(gxdr.LEDGER_ENTRY_CREATED.GetU32())},
```
bartekn, line 36: We may want to generate updated and remove changes in the future. Do you have a plan how to do that?

services/horizon/internal/ingest/processors/offers_processor_test.go
```diff
@@ -3,16 +3,56 @@
...
+				// the offers postgres table is configured with some database constraints which validate the following
+				// fields:
+				{randxdr.FieldEquals("created.lastModifiedLedgerSeq"), randxdr.SetPositiveNum32()},
+				{randxdr.FieldEquals("created.data.offer.amount"), randxdr.SetPositiveNum64()},
+				{randxdr.FieldEquals("created.data.offer.price.n"), randxdr.SetPositiveNum32()},
+				{randxdr.FieldEquals("created.data.offer.price.d"), randxdr.SetPositiveNum32()},
```
bartekn, line 42: This is cool but we need to remove cases that would not pass Stellar-Core validation (like flags - basically reimplement `doCheckValid` methods from Core here). We should probably also generate only offer changes (or at least set probability of offer change to something like 90%).

randxdr/presets.go
tamirms: they do not change. the only reason I use vars is because I wanted to communicate that the values are of the `Selector` type

xdr/json_test.go
```diff
@@ -59,6 +63,34 @@ func TestClaimPredicateJSON(t *testing.T) {
...
+func TestRandClaimPredicateJSON(t *testing.T) {
+	gen := randxdr.NewGenerator()
+	for i := 0; i < 1000; i++ {
+		cp := &ClaimPredicate{}
+		shape := &gxdr.ClaimPredicate{}
+		gen.Next(
+			shape,
+			[]randxdr.Preset{
+				{randxdr.IsPtr, randxdr.SetPtrToPresent},
+			},
```
bartekn, line 75: As in offers test: we should make sure that it does not generate objects that wouldn't be accepted by Core. Ex. `AbsBefore` < 0.

randxdr/generator.go
tamirms: I get a different values when I run that code https://play.golang.org/p/XOBZMkfZuuN

services/horizon/internal/ingest/processors/offers_processor_test.go
tamirms: I think you can do it in two steps. First generate a state entry containing an offer. For the update case generate an updated entry and preset the offer so it has the same id, allow the other offer fields to be random. In the remove case it would be similar except the type of the 2nd entry would be removed instead of updated

services/horizon/internal/ingest/processors/offers_processor_test.go
tamirms: I'm not sure we should do more than the bare minimum of validation because it will make our ingestion implementation more robust. If we do have any assumptions implicit in the ingestion code doing this type of testing will help make those assumptions more explicit

xdr/json_test.go
tamirms: Since we're testing if the code crashes / produces an error I don't think it's necessary to restrict the inputs more than the bare minimum

randxdr/marshaller.go
tamirms: I don't think that would that would work. Here is what  `XdrRecurse()` looks like for a union:

```golang
func (u *LedgerCloseMeta) XdrRecurse(x XDR, name string) {
	if name != "" {
		name = x.Sprintf("%s.", name)
	}
	XDR_int32(&u.V).XdrMarshal(x, x.Sprintf("%sv", name))
	switch u.V {
	case 0:
		x.Marshal(x.Sprintf("%sv0", name), XDR_LedgerCloseMetaV0(u.V0()))
		return
	}
	XdrPanic("invalid V (%v) in LedgerCloseMeta", u.V)
}
```

The marshaller is going to get applied on the union tag field and it will set `u.V` to `randomMarshaller.tag` which is what we want, so that's good.

However, the same marshaller will then be applied on the union arm. So, any uint32 or enum fields contained within the arm will be fixed to `randomMarshaller.tag` which we don't want to happen

randxdr/generator.go
bartekn: I can see same values on my laptop **and** in play.golang.

tamirms: Screen Shot 2021-04-08 at 8 04 36 AM

tamirms: another indication that it produces different values is `TestFuzzOffers`. If the same offer was generated on every run then the ingestion processor would return an error from postgres for attempting to create multiple rows with the same primary key

randxdr/marshaller.go
paulbellamy: Ah yeah, I was thinking in the non-union ones you would clear the tag. But I think you're right. 🤔 Ok. gefn.

randxdr/generator.go
paulbellamy: Yeah, I was under the impression you wanted it to return the same values every run. If you *don't* want that, the seed should be set to something like `time.UnixNano()`

services/horizon/internal/ingest/processors/offers_processor_test.go
paulbellamy: It could be interesting to model a state-machine around an account (e.g. create account, send a txn, updateData, etc) then generate sequences by a random walk through the state machine transitions.

xdr/json_test.go
paulbellamy: IMO, it should generate anything which could make it through the parser.

.circleci/config.yml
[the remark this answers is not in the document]
paulbellamy: well, two approaches, I guess. Chuck a comment in `./goxdr/main.go`(or wherever) to call the makefile: `// go:generate make whatever`, or it could just do it directly. I'm not super fussed either way, but think that have a single entrypoint to "regenerate all the codegen" would be a good thing.

randxdr/generator.go
bartekn: What got me thinking is that in tests we are testing the same cases over and over again but I understand now. Thanks!

bartekn: > The determinism is actually intentional because if there's a test failure I would like to be able to reproduce it

You can actually also just print the XDR that caused a failure.

services/horizon/internal/ingest/processors/offers_processor_test.go
bartekn: I understand what you mean. I think this is basically _more validation_ vs _more false positives_ dillema.

bartekn (approved):
LGTM. I think it's worth doing more in terms of validation to decrease number of false positives but maybe let's check if we get too many of them.

randxdr/marshaller.go
tamirms: cc @stanford-scs

PR #3524, opened 2021-04-02
services/horizon Add command for migration file hashes

### LINKED ISSUE #1832: Add Horizon command to return hash of each schema migration file
Horizon should have a command that returns hash of each schema migration file. It can prevent situations where release candidate is deployed for tests but some schema changes are required (ex. due to performance reasons). That way, code responsible for deploying Horizon can check if shared migration files are different and, if so, stop deployment process.

Adding to Horizon 0.23.0 milestone because looks simple but feel free to move to next release milestone.

JeroenVanSteijn:
closes #1832 

This PR adds a new service/horizon db command that prints out a hash for each .sql file in the migrations folder.

An open question is the output format. This was not specified in the request, and thus I opted to add the filenames as well for easier comparison between two outputs.

I don't think this change needs a version bump, since it does not impact any functionality and is not used yet. However, I am open to suggestions on this.


services/horizon/cmd/db.go
```diff
@@ -107,6 +112,41 @@ var dbMigrateCmd = &cobra.Command{
 	},
 }
 
+var dbMigrateHashCmd = &cobra.Command{
+	Use:   "migrate-hash",
+	Short: "outputs hash for each migration file",
+	Long:  "migrate-hash hashes each migration file of the database and outputs it",
+	Run: func(cmd *cobra.Command, args []string) {
+		migrationFolder := "./services/horizon/internal/db2/schema/migrations/"
```
tamirms, line 120: the migrations are actually embedded into the horizon binary using https://github.com/kevinburke/go-bindata . The code you have here assumes that the migrations are located on disk at the relative path of `./services/horizon/internal/db2/schema/migrations/`. If you were to build the horizon binary and then run the binary from outside the git repo, this command would fail.

Instead of assuming the files exist on disk, you need to use the interface provided by go-bindata to access the embedded files. Check out services/horizon/internal/db2/schema/main_test.go for some example code which inspects the embedded files.

JeroenVanSteijn: Very clear, thank you for the pointers. I thought the use-case was primarly for developers (inside the git repository), but making it work in just the binary seems like a better solution.

PR #3527, opened 2021-04-06
txnbuild: Reenable Multiplexed strkeys (SEP23 M-addresses) behind flags

### LINKED ISSUE #3490: Feature Request (txnbuild): (Re-)implement multiplexed accounts in the SDK
[Multiplexed account](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0027.md) support was implemented in the Stellar protocol in Protocol 13. Full ecosystem support was not implemented because the lift was large: see https://github.com/stellar/stellar-protocol/issues/617.

Adding SDK support (only) for multiplexed addresses now will allow developers to experiment and identify valuable concrete use cases to drive full adoption ([rationale](https://docs.google.com/document/d/1o0FUU3BLYtxVTPwT0RGlKjzAwfb8pPOr3hgcEilTNaY)). This issue does not propose adding any support to Horizon for M addresses.

M address support should **default to off**, so most users will see no functionality change.

- expand G address validation to also cover M (opt-in)
- allow construction of M addresses (opt-in)
- allow M addresses in transactions in place of G, where allowed (opt in)
- decode an M address from a provided hash (opt in)
- API helpers for comparing M addresses and extracting underlying G addresses

One complication: M addresses are not fully interchangeable with G. If you `grep` for AccountID in [Stellar-transaction.x](https://github.com/stellar/stellar-core/blob/master/src/xdr/Stellar-transaction.x), you'll see the places where AccountID was preserved. These are: CreateAccount, SetOptions, AllowTrust, and SponsoringFutureReserves, as well as various operation results. The parsing implementation should disallow such invalid address uses.

Note also: This [change to SEP 23](https://github.com/stellar/stellar-protocol/pull/895) will change the `strkey` implementation.

Prior work was done and then removed here: https://github.com/stellar/go/pull/2612. Based on the requirements above it is probably not sufficient.

Once the implementation is clear, we can raise equivalent issues on the other SDKs.

2opremio:
Fixes #3490 

This is ~an initial RFC draft~ a PR to re-introduce Multiplexed addresses in txnbuild, considering all the requirements from #3490 .

My main criteria for this proposal has been that no user should be enabling M-addresss unknowingly, either for new or existing code. For that reason:
  * New public functions/methods enabling SEP23 state so explicitly in their name
  * Existing public functions/methods keep their name and enable SEP23 through explicit options
  
Key changes:

1. the `txnbuild.Operation`  interface now adds a `withMuxedAccount` parameter to most of its methods:
```go
    type Operation interface {
	BuildXDR(withMuxedAccounts bool) (xdr.Operation, error)
	FromXDR(xdrOp xdr.Operation, withMuxedAccounts bool) error
	Validate(withMuxedAccounts bool) error
	GetSourceAccount() string
}
```
   ~For now, only `txnbuild.AccountMerge` supports the extended interface.~

2. `NewFeeBumpTransaction()`, `TransactionFromXDR()` and `NewFeeBumpTransaction()` now accept an option to enable SEP23, while keeping backwards compatibility

 TODO:
 - [x] Implement support for the new `SEP23`-methods in all operations
 - [x] Implement support for the updated M-strkey structure from https://github.com/stellar/stellar-protocol/pull/895
 - [x] Add tests
 - ~Add support for `SEP23` in the the `SEP10` functions (`BuildChallengeTx()`, `ReadChallengeTx()` ...). I am not even sure we should, TBH (I would appreciate feedback on this).~ We agreed offline not to do it.


txnbuild/transaction.go
```diff
@@ -604,6 +634,7 @@ type TransactionParams struct {
 	BaseFee              int64
 	Memo                 Memo
 	Timebounds           Timebounds
+	Options              []TransactionOption
```
tamirms, line 637: `TransactionParams` already represents options for constructing the Transaction. I would prefer to have a `SEP23` bool field

txnbuild/transaction.go
```diff
@@ -514,18 +514,33 @@ func (t GenericTransaction) FeeBump() (*FeeBumpTransaction, bool) {
 	return t.feeBump, t.feeBump != nil
 }
 
+type TransactionOption int
```
tamirms, line 517: I guess the reason this is necessary is because we don't want to break the `TransactionFromXDR ` API.

If we add a field to `TransactionParams` to toggle the SEP23 feature, we can change the name of this type to `ParseXDROption` to convey that these options should only be used with `TransactionFromXDR()`

txnbuild/operation.go
```diff
@@ -12,6 +12,9 @@ type Operation interface {
 	FromXDR(xdrOp xdr.Operation) error
 	Validate() error
 	GetSourceAccount() string
+	BuildXDRWithSEP23() (xdr.Operation, error)
```
tamirms, line 15: alternatively, we could keep the interface same and instead have a `SEP23` boolean flag in each of the operation structs which influences the behavior of `BuildXDR()` and `FromXDR(xdrOp xdr.Operation)`.

I think it would make sense to take this approach if we anticipate that we will eventually be supporting muxed accounts without opt in. If we transition to supporting muxed accounts without opt in it would be nice to minimize any breaking changes on the txnbuild interfaces

txnbuild/transaction.go
2opremio: > I guess the reason this is necessary is because we don't want to break the `TransactionFromXDR ` API.

Exactly

> If we add a field to `TransactionParams` to toggle the SEP23 feature, we can change the name of this type to `ParseXDROption` to convey that these options should only be used with `TransactionFromXDR()`

Sure thing

txnbuild/operation.go
2opremio: Sure, I can do that. The reason for adding new methods is that I didn't want to break backwards compatibility.

txnbuild/account_merge.go
```diff
@@ -27,7 +36,11 @@ func (am *AccountMerge) BuildXDR() (xdr.Operation, error) {
 		return xdr.Operation{}, errors.Wrap(err, "failed to build XDR OperationBody")
 	}
 	op := xdr.Operation{Body: body}
-	SetOpSourceAccount(&op, am.SourceAccount)
+	if withSEP23 {
+		SetOpSourceAccountWithSEP23(&op, am.SourceAccount)
```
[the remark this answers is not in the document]
2opremio: I am fine with the name change in the functions from the `txnbuild` package.

However, switching `SEP23` to `MuxedAccount` in the functions from the `xdr` package doesn't make sense, because they already operate on a `MuxedAccount`. That is:

```go
func (m *MuxedAccount) SetAddressWithSEP23(address string) error
func SEP23AddressToMuxedAccount(address string) (MuxedAccount, error)
func (m *MuxedAccount) SEP23Address() string
```

txnbuild/operation.go
```diff
@@ -24,8 +27,18 @@ func SetOpSourceAccount(op *xdr.Operation, sourceAccount string) {
 	op.SourceAccount = &opSourceAccountID
 }
 
+// SetOpSourceAccount sets the source account ID on an Operation, allowing M-strkeys (as defined in SEP23).
+func SetOpSourceAccountWithSEP23(op *xdr.Operation, sourceAccount string) {
```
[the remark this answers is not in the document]
2opremio: I have been extra careful about this after prior conversations with @tomerweller which led to reverting the initial implementation.

AFAIU this PR's purpose is enabling an initial evaluation of Muxed accounts and should be opt-in. By keeping the old symbol names and changing behavior, existing code will start accepting M-strkeys just by updating the txnbuild dependency. As I see, what you are suggesting is not opt-in (and in-fact there wouldn't event be an opt-out).

strkey/main.go
```diff
@@ -163,7 +166,7 @@ func checkValidVersionByte(version VersionByte) error {
 	switch version {
 	// intentionally disallow M-strkeys  (versionByteMuxedAccount)
 	// until SEP23 leaves the Draft status.
```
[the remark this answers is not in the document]
2opremio: Good catch

txnbuild/account_merge.go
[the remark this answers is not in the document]
2opremio: To be more specific, @bartekn are you OK with renaming the symbols in `txnbuild` and leaving the ones in `xdr` as is? Or do you have an alternative naming suggestion?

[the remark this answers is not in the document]
bartekn: We should definitely do that in `txnbuild` especially given that @lydiat and the team are pushing for human-friendly names. For `xdr` (all just suggestions):
```go
func (m *MuxedAccount) SetAddressWithSEP23(address string) error
```
Can we update existing `SetAddress` to handle muxed addresses?
```go
func SEP23AddressToMuxedAccount(address string) (MuxedAccount, error)
```
`AddressToMuxedAccount` and created based on the prefix (`G`/`M`).
```go
func (m *MuxedAccount) SEP23Address() string
````
Change `SEP23Address` to `Address` and return `G` for `KEY_TYPE_ED25519` and `M` for `KEY_TYPE_MUXED_ED25519`.

File: txnbuild/account_merge.go [hunk 7]
```diff
@@ -37,10 +50,22 @@ func (am *AccountMerge) FromXDR(xdrOp xdr.Operation) error {
...
-	am.SourceAccount = accountFromXDR(xdrOp.SourceAccount)
+	return am.fromXDR(xdrOp, false)
+}
+
+func (am *AccountMerge) fromXDR(xdrOp xdr.Operation, withSEP23 bool) error {
+	if xdrOp.Body.Type != xdr.OperationTypeAccountMerge {
+		return errors.New("error parsing account_merge operation from xdr")
+	}
+
+	am.SourceAccount = accountFromXDR(xdrOp.SourceAccount, withSEP23)
```
Shaptic, line 61: +1 towards @bartekn's comment about human-friendly names: if it wasn't mentioned in the PR title that SEP23 <==> M-addresses and I just saw this in the codebase, I'd be lost. However, I do think mentioning/linking SEP-23 in a docstring would be invaluable since that's where you'd go for more detailed information.

Anyway, for this specific case, maybe something like `withMuxedAccount bool` instead?

txnbuild/account_merge.go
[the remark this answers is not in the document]
2opremio: > Can we update existing `SetAddress` to handle muxed addresses?

I separated it for the same `opt-in` reason I mentioned in a separate comment thread (the function and package are public and it's fair to assume that there's code out there already using it)

> `AddressToMuxedAccount` and created based on the prefix `(G/M)`.

That would be fine since I just introduced it, but I would like to keep it consistent with `SetAddressWithSEP23()`

(BTW, it already does the G/M dance)

> Change SEP23Address to Address and return G for KEY_TYPE_ED25519 and M for KEY_TYPE_MUXED_ED25519.

It already does the G/M dance. But again, I would like to keep the naming consistent with`SetAddressWithSEP23()`

txnbuild/allow_trust.go
```diff
@@ -21,7 +21,7 @@ type AllowTrust struct {
 }
```
tamirms, line 21: Sorry my earlier comment was probably not clear enough. I thought you could include a `AllowMuxed bool` field to the struct definition and then  `BuildXDR()` and `FromXDR()` would condition on `at.AllowMuxed` in the function body. That would allow us to not break the existing interface

2opremio: You mean, in the type definition of every operation? I think that's an awkward contract, when using the `Operation()` methods in isolation, particularly for `FromXDR()` in which the user expects the struct fields to be filled in.

Your suggestion would require the user to initialize the `AllowMuxed` before calling `FromXDR()`.

txnbuild/account_merge.go
2opremio: Yeah, I have already renamed that. Note, however, that `accountFromXDR` is private.

txnbuild/allow_trust.go
2opremio: Also, adding a field to each operation struct is a bit strange. Mixing configuration options with data.

2opremio: we could make it a private field ... but then there would be no way to externally call the `Operation` while enabling MuxedAccounts.

txnbuild/account_merge.go
2opremio: Also note that the `APPEND_PATH` parameter was a collective effort 🤣

txnbuild/allow_trust.go
tamirms: we discussed this offline:

it doesn't make sense to have public `AllowMuxed` fields on operations because an operation can have an `AllowMuxed` field which contradicts the `EnableMuxedAccounts ` field in `TransactionParams`.

We could make `allowMuxed` a private field which is only set by the transaction constructor or in `TransactionFromXDR()`. But we agreed that the current solution would be cleaner even though it is a breaking change.

In practice, most users of txnbuild are not calling `Build()` or `Validate()` on individual operations so most people upgrading their txnbuild dependency will not need to make any code changes to support the new version.

2opremio: Thanks @tamirms

2opremio:
@tamirms Tests (Which are cooking) and https://github.com/stellar/go/pull/3527#discussion_r608892387 aside, this is ready for review. PTAL while I am on it so that we can reach the release quicker.

Same for @bartekn (but I know he is off today :) ).

strkey/main.go
```diff
@@ -207,6 +208,10 @@ func decodeString(src string) ([]byte, error) {
 	//    which we have for checked above). If there are any leftover bits, they should be set to 0
 	if leftoverBits > 0 {
 		lastChar := srcBytes[len(srcBytes)-1]
+		if int(lastChar) >= len(decodingTable) {
```
tamirms, line 211: I left a comment in https://github.com/stellar/go/pull/3529 about this:

decodingTable has a size of 256. assuming lastChar is a byte shouldn't it always be in the range of 0-255?

txnbuild/clawback.go
```diff
@@ -59,15 +63,19 @@ func (cb *Clawback) BuildXDR() (xdr.Operation, error) {
...
 	result, ok := xdrOp.Body.GetClawbackOp()
 	if !ok {
 		return errors.New("error parsing clawback operation from xdr")
 	}
 
-	cb.SourceAccount = accountFromXDR(xdrOp.SourceAccount)
-	fromAID := result.From.ToAccountId()
-	cb.From = fromAID.Address()
+	cb.SourceAccount = accountFromXDR(xdrOp.SourceAccount, withMuxedAccounts)
+	if withMuxedAccounts {
```
tamirms, line 73: can we use accountFromXDR() to assign cb.From?

txnbuild/path_payment.go
```diff
@@ -88,18 +92,28 @@ func (pp *PathPaymentStrictReceive) BuildXDR() (xdr.Operation, error) {
...
-func (pp *PathPaymentStrictReceive) FromXDR(xdrOp xdr.Operation) error {
+func (pp *PathPaymentStrictReceive) FromXDR(xdrOp xdr.Operation, withMuxedAccounts bool) error {
 	result, ok := xdrOp.Body.GetPathPaymentStrictReceiveOp()
 	if !ok {
 		return errors.New("error parsing path_payment operation from xdr")
 	}
 
-	pp.SourceAccount = accountFromXDR(xdrOp.SourceAccount)
+	pp.SourceAccount = accountFromXDR(xdrOp.SourceAccount, withMuxedAccounts)
+	if withMuxedAccounts {
```
tamirms, line 111: I think accountFromXDR can be used to assign to pp.Destination

txnbuild/transaction.go
```diff
@@ -691,11 +729,10 @@ func NewTransaction(params TransactionParams) (*Transaction, error) {
 	}
 
 	for _, op := range tx.operations {
-		if verr := op.Validate(); verr != nil {
+		if verr := op.Validate(false); verr != nil {
```
tamirms, line 732: should we use params.EnableMuxedAccounts instead of false?

txnbuild/account_merge.go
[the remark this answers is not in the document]
bartekn: I think that `xdr` package should be independent of `txnbuild` (people can use it without `txnbuild`). What about this:
* `MuxedAccount.SetAddressWithSEP23`:
  * Existing `SetAddress` handles both ed25519 and muxed accounts.
  * We add a new `SetEd25519Address` that works only for `G` addresses.
  * We add a new `SetMuxedEd25519Address` that works only for `M` addresses.
 * `SEP23AddressToMuxedAccount` (similar logic to above):
   * `AddressToMuxedAccount` handles both types,
   * `Ed25519AddressToMuxedAccount` that works only for `G` addresses.
   * `MuxedEd25519AddressToMuxedAccount` that works only for `M` addresses.
 * `MuxedAccount.SEP23Address`:
   *  `Address` - returns `G`/`M` addresses.
   *  `Ed25519Address` - returns `G` addresses, errors on muxed.
   *  `MuxedEd25519Address` - returns `M` addresses, errors on _normal_.

txnbuild/transaction.go
2opremio: Absolutely. Good catch!

That was just a leftover default when using GoLand's interface refactoring.

strkey/main.go
2opremio: I responded there, 🤦

txnbuild/account_merge.go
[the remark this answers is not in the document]
bartekn: @tomerweller we had a discussion about it with @2opremio on slack.

I feel quite strongly about mixing concerns of `txnbuild` with `xdr`. Specifically, it looks like some muxed account features were built as an opt-in feature in `xdr`. Instead, I think that `xdr` should support all features of the latest protocol (ex. methods like `MuxedAccount.SetAddress` should support both `G` and `M` addresses). If we need opt-in we can do that at `txnbuild` level (ex. if the user wants to use both: we use general method that supports both addresses type, if not: we use `G` supported methods only). I think that this way we will unblock everyone else that builds apps using low-level `xdr` package.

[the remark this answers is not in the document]
2opremio: Thanks @bartekn I will do that

txnbuild/path_payment.go
2opremio: Yep, we can, since `From` cannot be nil.

txnbuild/clawback.go
2opremio: Yes, we can, since `From` cannot be `nil`

txnbuild/path_payment.go
2opremio: Not really, because `accountFromXDR()` accepts `nil` values.

txnbuild/account_merge.go
```diff
@@ -13,10 +13,14 @@ type AccountMerge struct {
 }
 
 // BuildXDR for AccountMerge returns a fully configured XDR Operation.
-func (am *AccountMerge) BuildXDR() (xdr.Operation, error) {
+func (am *AccountMerge) BuildXDR(withSEP23 bool) (xdr.Operation, error) {
```
leighmcculloch, line 16: ❗ I don't think `SEP23` should show up in the code. The feature is muxed accounts str keys, and the code should refer to that feature I think. I see us using `withMuxedAccounts` elsewhere, could we use that here too?

txnbuild/account_merge.go
```diff
@@ -13,10 +13,14 @@ type AccountMerge struct {
...
 	var xdrOp xdr.MuxedAccount
-
-	err := xdrOp.SetAddress(am.Destination)
+	var err error
+	if withSEP23 {
+		err = xdrOp.SetAddress(am.Destination)
+	} else {
+		err = xdrOp.SetEd25519Address(am.Destination)
```
leighmcculloch, line 22: ❔ I find it interesting that the old behavior here was `SetAddress` and now the new default behavior will be `SetEd25519Address`, and this differs from 10 lines down where the old behavior is `SetOpSourceAccount` and the new default behavior is `SetOpSourceAccount`.

It appears to me like the default behavior is changing in some places, but not in others.

Why is that?

txnbuild/account_merge.go
leighmcculloch, line 16: 💡 Instead of introducing a boolean I think we should introduce a struct type that contains the boolean and can be used in all the places. It will keep the API thinner if we add more options. If we need to make a similar change again we won't need to break the APIs of these functions.

I think it will also make the call sites and tests clearer. For example I see this in the tests:
```go
p.FromXDR(xdrEnv.Operations()[1], false)
```

This would be clearer:
```go
p.FromXDR(xdrEnv.Operations()[1], XDROptions{WithMuxedAccounts: false})
```

I see you have a type defined for the TransactionFromXDR function that is sort of this. Any reason we aren't using the same type everywhere and just passing it down?

Thoughts?

txnbuild/claim_claimable_balance.go
```diff
@@ -54,7 +58,7 @@ func (cb *ClaimClaimableBalance) FromXDR(xdrOp xdr.Operation) error {
 
 // Validate for ClaimClaimableBalance validates the required struct fields. It returns an error if any of the fields are
 // invalid. Otherwise, it returns nil.
-func (cb *ClaimClaimableBalance) Validate() error {
+func (cb *ClaimClaimableBalance) Validate(bool) error {
```
leighmcculloch, line 61: ❗ Can you name the bool variable here? It's unclear what it is without giving it a name. I realize in the context of this PR it is obvious, but after the PR is merged looking at the code in isolation won't provide that context.

tamirms (approved):

txnbuild/transaction.go
```diff
@@ -514,18 +514,33 @@ func (t GenericTransaction) FeeBump() (*FeeBumpTransaction, bool) {
 	return t.feeBump, t.feeBump != nil
 }
 
+type ParseXDROption int
```
leighmcculloch, line 517: This type is named ParseXDROption, but it is for the TransactionFromXDR function, not the ParseXDR function. I think it would make it easier to understand the relationship between this type and the function it is intended for if it was named `TransactionFromXDROption`.

txnbuild/transaction.go
```diff
@@ -514,18 +514,33 @@ func (t GenericTransaction) FeeBump() (*FeeBumpTransaction, bool) {
...
+			return true
+		}
+	}
+	return false
+}
+
 // TransactionFromXDR parses the supplied transaction envelope in base64 XDR
 // and returns a GenericTransaction instance.
-func TransactionFromXDR(txeB64 string) (*GenericTransaction, error) {
+func TransactionFromXDR(txeB64 string, options ...ParseXDROption) (*GenericTransaction, error) {
```
leighmcculloch, line 534: 💡 I have a couple thoughts about how we're approaching options here:

---

1 – Boolean only options types

The definition of this options type as an `int` makes it only possible to use these options for booleans in the future. If we're going to use optional option types, we may as well go all the way to using optional function types which will look the same for the user but will allow us to add non-boolean options in the future. It'd look something like this:

```go
type transactionFromXDROptions struct {
    EnableMuxedAccounts bool
}

type TransactionFromXDROption interface {
    Apply(transactionFromXDROptions)
}

type transactionFromXDROptionFunc func(transactionFromXDROptions)

func (f transactionFromXDROptionFunc) Apply(o transactionFromXDROptions) {
    f(o)
}

func TransactionFromXDROptionEnableMuxedAccounts() TransactionFromXDROption {
    return transactionFromXDROptionFunc(func (o transactionFromXDROptions) {
        o.EnableMuxedAccounts = true
    }
}

func TransactionFromXDR(txeB64 string, options ...TransactionFromXDROption) (*GenericTransaction, error) {
    optionsApplied := transactionFromXDROptions{}
    for _, o := options {
        o.Apply(optionsApplied)
    }
    ...
```

---

2 – Consistency with transaction building

I also think it would be better to have something like `TransactionParams` in this function. It won't be backwards compatible, but other changes in this PR aren't either and then the interface to parsing will look much more similar to the interface to building new transactions.

---

If I was choosing, I'd be choosing something like in 2, but if we really want variable list options then I think we should do 1 otherwise we're not really getting all the benefits of using them.

Thoughts?

xdr/muxed_account.go
```diff
@@ -35,14 +35,103 @@ func (m *MuxedAccount) SetAddress(address string) error {
...
+// of the provided strkey G-address or M-address, as described in SEP23.
+func (m *MuxedAccount) SetAddress(address string) error {
+	if m == nil {
+		return nil
+	}
+
+	switch len(address) {
+	case 56:
+		return m.SetEd25519Address(address)
+	case 69:
```
leighmcculloch, line 60: ❗  What's the motivation for forking on the length of the address rather than the first character, which is used to uniquely identify each of the types of addresses?

The length of the strkeys is a concern that should be owned by the strkey package.

We could add a function to strkey to get the version byte, then fork on that. This keeps the logic about identifying the addresses inside the strkey package, which should be the package to own that concern.

leighmcculloch (commented):
Really exciting to see this rolling.

I left comments about a couple things I think we should change (❗).

If you're open, I also left some suggestions (💡) that are to do with how the API is changing for the user, and how I think we could make the API is easier to understand.

txnbuild/claim_claimable_balance.go
2opremio: Sure

txnbuild/account_merge.go
2opremio: See prior resolved discussion with @bartek

txnbuild/transaction.go
2opremio: Sure

txnbuild/transaction.go
2opremio: I decided to keep it backwards compatible

txnbuild/account_merge.go
2opremio: Yes, that was the idea, this is a leftover I forgot to change. Thanks.

xdr/muxed_account.go
2opremio: The length approach was already present before this PR. Please create a ticket if you think it's important to change it (I don't want to delay this PR further)

txnbuild/transaction.go
2opremio: PS :see the other conversations resolved conversations.

txnbuild/account_merge.go
2opremio: This isn't much of a difference and I think it's overkill if we don't envision other upcoming options. The idea is, in fact, to remove the option in the future, once (and if) the experimentation with muxed accounts is successful.

2opremio:
It took considerable effort (due to the lack of prior coverage) but all operations should now be tested.

This should be good to go. @tamirms @bartekn please take a final look.

tamirms (approved):

txnbuild/CHANGELOG.md
```diff
@@ -8,11 +8,16 @@ file.  This project adheres to [Semantic Versioning](http://semver.org/).
...
 
 * `AllowTrustOpAsset` was renamed to `AssetCode`, `{Must}NewAllowTrustAsset` was renamed to `{Must}NewAssetCodeFromString`.
+* Some methods from the `Operation` interface (`BuildXDR()`,`FromXDR()` and `Validate()`) now take an additional `bool` parameter (`withMuxedAccounts`)
+  to indicate whether [SEP23](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0023.md) M-strkeys should be enabled.
 
 ### New features
 
-Add support for Stellar Protocol 16 (CAP35): `Clawback` operations.
-
+* Add support for Stellar Protocol 16 (CAP35): `Clawback`, `ClawbackClaimableBalance` and `SetTrustlineFlags` operations.
```
bartekn, line 16:
```suggestion
* Add support for Stellar Protocol 17 (CAP35): `Clawback`, `ClawbackClaimableBalance` and `SetTrustlineFlags` operations.
```

txnbuild/account_merge.go
```diff
@@ -13,10 +13,14 @@ type AccountMerge struct {
 }
 
 // BuildXDR for AccountMerge returns a fully configured XDR Operation.
-func (am *AccountMerge) BuildXDR() (xdr.Operation, error) {
+func (am *AccountMerge) BuildXDR(withMuxedAccounts bool) (xdr.Operation, error) {
```
bartekn, line 16: Nit: I know this PR has been reviewed multiple times but I have one (simple) idea how to improve the API. Passing bool value to methods can be misleading (user need to check the description of the method to understand what is the param). We can create additional extra methods for each method accepting `withMuxedAccounts`. `BuildXDR` as an example:
* Change `BuildXDR(withMuxedAccounts bool) (xdr.Operation, error)` to `buildXDR(withMuxedAccounts bool) (xdr.Operation, error)`
* `BuildXDR()` stays in a form before this PR. Calls `buildXDR(false)` internally.
* A new `BuildXDRWithMuxedAccounts()` is added and calls `buildXDR(true)` internally.

bartekn (approved):

xdr/muxed_account.go
```diff
@@ -1,15 +1,15 @@
...
 	"fmt"
 
+	"github.com/pkg/errors"
 	"github.com/stellar/go/strkey"
 )
 
 func MustMuxedAddress(address string) MuxedAccount {
 	muxed := MuxedAccount{}
-	err := muxed.SetAddress(address)
+	err := muxed.SetEd25519Address(address)
```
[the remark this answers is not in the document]
2opremio: Uhm, good point, although it breaks the opt-in approach.

I will change it, since otherwise it can be confusing.

txnbuild/account_merge.go
2opremio: @bartekn that's **exactly** what I implemented originally and I discarded after the design discussion. See https://github.com/stellar/go/pull/3527/commits/27a79c6bded74710d05f8870e19e44653179c5bc#diff-0ef2e7d165d2f196939c937afadef91206823b0671f32120b265843fd291332aR15 and the issue body edit history (which originally presented the same design you suggested).

We should be more cohesive when discussing designs (but that's for a retro discussion). The purpose of creating an early draft was exactly avoiding this situation.

I personally think this needs to go out (it's already late) but if @ire-and-curses thinks it's worth another rewrite, I will cave in and restore my original design.

2opremio: @ire-and-curses  In order to get more context, on what happened.

1. I originally proposed what @bartekn is proposing now
2. @tamirms suggested using an internal boolean (appended to each operation struct) at https://github.com/stellar/go/pull/3527#discussion_r608575810
3. We discarded it at https://github.com/stellar/go/pull/3527#discussion_r609032676 and went for the design as it is now.
4. @leighmcculloch proposed yet another design modification at https://github.com/stellar/go/pull/3527#discussion_r611925837 which I discarded.
5. @bartekn suggests my original design.

ire-and-curses: 6. Profit??!!
Let's discuss in team meeting today. I really want to get this out.

2opremio: We decided offline to keep it as it is. Thanks for the suggestion though @bartekn

PR #3528, opened 2021-04-08
historyarchive: Archivist slow scan missing

graydon:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

* [X] (NA) I've updated any docs ([developer docs](https://www.stellar.org/developers/reference/), `.md`

### Release planning

* [X] (NA) I've updated the relevant CHANGELOG ([here](services/horizon/CHANGELOG.md) for Horizon) if
* [X] (No) I've decided if this PR requires a new major/minor version according to
</details>

### What

This change fixes a bug in archivist that causes it to mis-report archives as "not missing files" when they are, in fact, missing files. Fairly serious condition. It's caused on archives accessed through http -- which resorts to a slow-scan path that probes each file one by one using HEAD -- rather than archives like s3 that can be fast-scanned with a list command.

### Why

Existing behaviour is wrong. It treats "we got information about a file at all" as meaning "the file is there", even when the information we got is "the file is not there". Classic sort of 3-value-logic mistake on my part.

### Known limitations

None I'm aware of.

### DIFF
--- a/historyarchive/scan.go
+++ b/historyarchive/scan.go
@@ -363,8 +363,8 @@ func (arch *Archive) CheckCheckpointFilesMissing(opts *CommandOptions) map[strin
 	for _, cat := range Categories() {
 		missing[cat] = make([]uint32, 0)
 		for ix := range opts.Range.GenerateCheckpoints(arch.checkpointManager) {
-			fileExists, mapEntryExists := arch.checkpointFiles[cat][ix]
-			if !fileExists || !mapEntryExists {
+			fileExists := arch.checkpointFiles[cat][ix]
+			if !fileExists {
 				missing[cat] = append(missing[cat], ix)
 			}
 		}
--- a/historyarchive/scan.go
+++ b/historyarchive/scan.go
@@ -348,8 +348,8 @@ func (arch *Archive) NoteExistingBucket(bucket Hash) {
 func (arch *Archive) NoteReferencedBucket(bucket Hash) bool {
 	arch.mutex.Lock()
 	defer arch.mutex.Unlock()
-	_, exists := arch.referencedBuckets[bucket]
-	if exists {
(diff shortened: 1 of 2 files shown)


historyarchive/scan.go
```diff
@@ -377,8 +377,8 @@ func (arch *Archive) CheckBucketsMissing() map[Hash]bool {
 	defer arch.mutex.Unlock()
 	missing := make(map[Hash]bool)
 	for k := range arch.referencedBuckets {
-		_, ok := arch.allBuckets[k]
-		if !ok {
+		bucketExists, mapEntryExists := arch.allBuckets[k]
+		if !bucketExists || !mapEntryExists {
```
bartekn, line 381: nit: `allBuckets` is `map[Hash]bool` so we can exploit a default value:
```suggestion
		bucketExists := arch.allBuckets[k]
		if !bucketExists {
```

historyarchive/scan.go
```diff
@@ -363,8 +363,8 @@ func (arch *Archive) CheckCheckpointFilesMissing(opts *CommandOptions) map[strin
 	for _, cat := range Categories() {
 		missing[cat] = make([]uint32, 0)
 		for ix := range opts.Range.GenerateCheckpoints(arch.checkpointManager) {
-			_, ok := arch.checkpointFiles[cat][ix]
-			if !ok {
+			fileExists, mapEntryExists := arch.checkpointFiles[cat][ix]
+			if !fileExists || !mapEntryExists {
```
bartekn, line 367: nit: I think we should be able to do the same thing here:
```suggestion
			fileExists := arch.checkpointFiles[cat][ix]
			if !fileExists {
```

bartekn (approved):
LGTM, 2 nit comments.

PR #3529, opened 2021-04-08
strkey: Check for leftover character to be in the expected alphabet

2opremio:
Found this while working on #3527

### DIFF
--- a/strkey/main.go
+++ b/strkey/main.go
@@ -207,11 +207,12 @@ func decodeString(src string) ([]byte, error) {
 	//    which we have for checked above). If there are any leftover bits, they should be set to 0
 	if leftoverBits > 0 {
 		lastChar := srcBytes[len(srcBytes)-1]
-		if int(lastChar) >= len(decodingTable) {
-			// let's output an error matching the errors from the base32 decoder invocation below
+		decodedLastChar := decodingTable[lastChar]
+		if decodedLastChar == 0xff {
+			// The last character from the input wasn't in the expected input alphabet.
+			// Let's output an error matching the errors from the base32 decoder invocation below
 			return nil, errors.Wrap(base32.CorruptInputError(len(srcBytes)), "base32 decode failed")
 		}
-		decodedLastChar := decodingTable[lastChar]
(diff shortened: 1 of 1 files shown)


strkey/main.go
```diff
@@ -207,6 +207,10 @@ func decodeString(src string) ([]byte, error) {
 	//    which we have for checked above). If there are any leftover bits, they should be set to 0
 	if leftoverBits > 0 {
 		lastChar := srcBytes[len(srcBytes)-1]
+		if int(lastChar) >= len(decodingTable) {
```
tamirms, line 210: `decodingTable` has a size of 256. assuming `lastChar` is a byte shouldn't it always be in the range of 0-255?

2opremio: That is totally true (🤦 ). Instead we should be checking for a decoded value of `0xff` for the error (otherwise we can get an unintuitive `non-canonical strkey` error when the problem is the input alphabet).

2opremio:
@tamirms PTAL

tamirms (approved):

strkey/main.go
```diff
@@ -208,6 +208,11 @@ func decodeString(src string) ([]byte, error) {
 	if leftoverBits > 0 {
 		lastChar := srcBytes[len(srcBytes)-1]
 		decodedLastChar := decodingTable[lastChar]
+		if decodedLastChar == 0xff {
+			// The last character from the input wasn't in the expected input alphabet.
+			// Let's output an error matching the errors from the base32 decoder invocation below
+			return nil, errors.Wrap(base32.CorruptInputError(len(srcBytes)), "base32 decode failed")
+		}
```
leighmcculloch, line 215: Could you add a test for this bug being fixed? This seems pretty critical to be tested, but I don't see tests being updated or added.

leighmcculloch (commented):
Nice find! One request below.

strkey/main.go
2opremio: Unlike my initial analysis, it's not critical at all (check the comment from @tamirms ). This PR ended up just being a small error refinement.

Issue #3344, opened 2021-01-14
Add timeouts to all functions called from Tick()


bartekn:
### What version are you using?

1.13.1

### What did you do?

We don't set timeouts in functions in `horizon.App.Tick()` so if the connection gets stuck for any reason it will block entire `Tick()` method forever. It will affect metrics, fee stats and all the components of the system that rely on ingestion/ledger related stats.

Thanks @jacekn for reporting this!

bartekn:
Additionally check other places where this can cause issues.

2opremio:
Related https://github.com/stellar/go/issues/613

PR #3545, opened 2021-04-15
all: Refactor db.Session to take ctx into each call, instead of storing it on the session level

paulbellamy:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

The root change is:
```diff
 package db

 type Session struct {
        // DB is the database connection that queries should be executed against.
        DB *sqlx.DB

-       // Ctx is the context in which the repo is operating under.
-       Ctx context.Context
-
        tx        *sqlx.Tx
        txOptions *sql.TxOptions
 }
```
And changing all the `db.Session` methods (where needed) to take a `ctx context.Context` params. Then passing that through all the places to get it there.

### Why

For https://github.com/stellar/go/issues/3344, we want to pass different timeouts in different contexts. This change makes that easier, and generally follows the go norms of passing the context late into the method.

### Known limitations

It's big. Will be difficult to review and merge. Most of the param passing is pretty mechanical. The interesting parts are around where we create the contexts initially. But, most of that is in tests.

### Questions

- [ ] RemoteCaptiveStellarCore.createContext and cancellation semantics. Should we do it everywhere? Only for PrepareRange?
  - Ignored this, keeping the existing semantics, for now. But seems like we should cancel all outstanding requests, not just PrepareRange.
- [ ] OrderBookStream.Run, should we make a new timeout ctx each tick?

### Bonus

~~I also fixed `TestOpen_openAndPingFails`, because it was intermittently failing and being annoying.~~

### DIFF
--- a/exp/services/captivecore/internal/api.go
+++ b/exp/services/captivecore/internal/api.go
@@ -63,7 +63,7 @@ func (c *CaptiveCoreAPI) Shutdown() {
 	c.core.Close()
 }
 
-func (c *CaptiveCoreAPI) startPrepareRange(ledgerRange ledgerbackend.Range) {
+func (c *CaptiveCoreAPI) startPrepareRange(ctx context.Context, ledgerRange ledgerbackend.Range) {
 	defer c.wg.Done()
 
 	err := c.core.PrepareRange(ledgerRange)
@@ -100,7 +100,7 @@ func (c *CaptiveCoreAPI) startPrepareRange(ledgerRange ledgerbackend.Range) {
 }
 
 // PrepareRange executes the PrepareRange operation on the captive core instance.
-func (c *CaptiveCoreAPI) PrepareRange(ledgerRange ledgerbackend.Range) (ledgerbackend.PrepareRangeResponse, error) {
+func (c *CaptiveCoreAPI) PrepareRange(ctx context.Context, ledgerRange ledgerbackend.Range) (ledgerbackend.PrepareRangeResponse, error) {
 	c.activeRequest.Lock()
 	defer c.activeRequest.Unlock()
 	if c.ctx.Err() != nil {
@@ -121,7 +121,7 @@ func (c *CaptiveCoreAPI) PrepareRange(ledgerRange ledgerbackend.Range) (ledgerba
 		c.activeRequest.valid = true
 
 		c.wg.Add(1)
-		go c.startPrepareRange(ledgerRange)
+		go c.startPrepareRange(ctx, ledgerRange)
 
 		return ledgerbackend.PrepareRangeResponse{
 			LedgerRange:   ledgerRange,
@@ -140,7 +140,7 @@ func (c *CaptiveCoreAPI) PrepareRange(ledgerRange ledgerbackend.Range) (ledgerba
 }
 
 // GetLatestLedgerSequence determines the latest ledger sequence available on the captive core instance.
-func (c *CaptiveCoreAPI) GetLatestLedgerSequence() (ledgerbackend.LatestLedgerSequenceResponse, error) {
+func (c *CaptiveCoreAPI) GetLatestLedgerSequence(ctx context.Context) (ledgerbackend.LatestLedgerSequenceResponse, error) {
 	c.activeRequest.Lock()
 	defer c.activeRequest.Unlock()
 
@@ -159,7 +159,7 @@ func (c *CaptiveCoreAPI) GetLatestLedgerSequence() (ledgerbackend.LatestLedgerSe
 }
 
 // GetLedger fetches the ledger with the given sequence number from the captive core instance.
-func (c *CaptiveCoreAPI) GetLedger(sequence uint32) (ledgerbackend.LedgerResponse, error) {
+func (c *CaptiveCoreAPI) GetLedger(ctx context.Context, sequence uint32) (ledgerbackend.LedgerResponse, error) {
 	c.activeRequest.Lock()
 	defer c.activeRequest.Unlock()
 
--- a/exp/services/captivecore/internal/api.go
+++ b/exp/services/captivecore/internal/api.go
@@ -63,7 +63,7 @@ func (c *CaptiveCoreAPI) Shutdown() {
 	c.core.Close()
 }
 
-func (c *CaptiveCoreAPI) startPrepareRange(ctx context.Context, ledgerRange ledgerbackend.Range) {
+func (c *CaptiveCoreAPI) startPrepareRange(ledgerRange ledgerbackend.Range) {
 	defer c.wg.Done()
 
 	err := c.core.PrepareRange(ledgerRange)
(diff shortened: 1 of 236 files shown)


ingest/ledgerbackend/captive-core-026b1473/stellar-core.conf
```diff
@@ -0,0 +1,14 @@
+# Generated file -- do not edit
```
tamirms, line 1: was this included by mistake?

paulbellamy: d'oh, yes

ingest/ledgerbackend/ledger_backend.go
```diff
@@ -1,32 +1,34 @@
...
 import (
+	"context"
+
 	"github.com/stellar/go/xdr"
 )
 
 // LedgerBackend represents the interface to a ledger data store.
 type LedgerBackend interface {
 	// GetLatestLedgerSequence returns the sequence of the latest ledger available
 	// in the backend.
```
tamirms, line 12: This is a breaking change for the ingestion library since we're changing the function signatures. Can we pull these changes out from this PR? We do plan on changing the interface so `GetLedger()` is always blocking, see https://github.com/stellar/go/pull/3518#pullrequestreview-632145899

So, I think it would be good to combine both those API changes in a single release

bartekn: That's right. We can make the change to `LedgerBackend` when working on https://github.com/stellar/go/issues/3549.

paulbellamy: The point of this was to set things up so we can add ctx into the methods called from `horizon.App.Tick()`, which are mostly calls to the LedgerBackend... But yeah, we could definitely separate it out into a couple smaller PRs, and do the other API change as well.

tamirms: @paulbellamy I don't think any of the methods invoked in `App.Tick()` involve `LedgerBackend`
```golang
func (a *App) Tick() {
	var wg sync.WaitGroup
	log.Debug("ticking app")
	// update ledger state, operation fee state, and stellar-core info in parallel
	wg.Add(3)
	go func() { a.UpdateLedgerState(); wg.Done() }()
	go func() { a.UpdateFeeStatsState(); wg.Done() }()
	go func() { a.UpdateStellarCoreInfo(); wg.Done() }()
	wg.Wait()

	wg.Add(2)
	go func() { a.reaper.Tick(); wg.Done() }()
	go func() { a.submitter.Tick(a.ctx); wg.Done() }()
	wg.Wait()

	log.Debug("finished ticking app")
}
```
`UpdateLedgerState()` uses the stellar core client and queries the horizon db
`UpdateFeeStatsState()` queries the horizon db
`UpdateStellarCoreInfo()` uses the stellar core client

`LedgerBackend` is only used for ingestion purposes

paulbellamy: Yeah, just realized that while I was pulling out the changes.

paulbellamy: The changes I just pulled out are: https://github.com/paulbellamy/stellar-go/compare/3344/tick-timeouts...paulbellamy:3344/tick-timeouts-ledgerbackend?expand=1

.gitignore
```diff
@@ -11,3 +11,4 @@ debug
 *.crt
 *.csr
 *.key
+/ingest/ledgerbackend/captive-core-*/
```
tamirms, line 14: this probably shouldn't be included

exp/services/captivecore/internal/api.go
```diff
@@ -63,7 +63,7 @@ func (c *CaptiveCoreAPI) Shutdown() {
 	c.core.Close()
 }
```
tamirms, line 65: note to self: changes to this file requires a careful review

ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -241,7 +241,7 @@ func (c *CaptiveStellarCore) openOfflineReplaySubprocess(from, to uint32) error
 	return nil
 }
```
tamirms, line 242: can you move these changes to the ledgerbackend PR ? it would help to reduce the cognitive load

ingest/ledgerbackend/database_backend.go
```diff
@@ -74,8 +75,9 @@ func (*DatabaseBackend) IsPrepared(ledgerRange Range) (bool, error) {
 
 // GetLatestLedgerSequence returns the most recent ledger sequence number present in the database.
 func (dbb *DatabaseBackend) GetLatestLedgerSequence() (uint32, error) {
+	ctx := context.TODO()
```
tamirms, line 78: this will change the behavior because now we cannot cancel a query via context termination. maybe it's best to move changes to this file to the ledgerbackend PR as well

.gitignore
paulbellamy: in gitignore? or in this PR specifically?

tamirms: in gitignore

paulbellamy: It gets created by tests. So we should either have it in gitignore, or (better) put the files somewhere already ignored. Either way, not really related to this PR, so...

paulbellamy:
Stuff here is difficult without including the ledgerbackend changes. Going to try doing those first, and see if that makes this change simpler.

ingest/ledgerbackend/database_backend.go
paulbellamy: I think this is fixed, with the new changes, but double-check me on that, please.

paulbellamy:
One catch with this, is that we now need to call an explicit `.Close()` method on places where we relied on the context cancellation to implicitly shutdown everything.

Places where I've found that so far:
- [x] LedgerBackend shutdown (we were already doing that)
- [x] LedgerHashStore shotdown (also already doing that)

ingest/ledgerbackend/database_backend.go
```diff
@@ -26,21 +27,32 @@ var _ LedgerBackend = (*DatabaseBackend)(nil)
 
 // DatabaseBackend implements a database data store.
 type DatabaseBackend struct {
+	cancel            context.CancelFunc
+	ctx               context.Context
```
2opremio, line 31: Why is it stored instead of using it in method calls? (it would only be OK if a database backend is ephemeral)

ingest/ledgerbackend/remote_captive_core.go
```diff
@@ -67,6 +67,7 @@ type RemoteCaptiveStellarCore struct {
 	client                   *http.Client
 	lock                     *sync.Mutex
 	cancel                   context.CancelFunc
+	parentCtx                context.Context
```
2opremio, line 70: Same here

2opremio (approved):

PR #3551, opened 2021-04-19
exp/services/recoverysigner: Fix TestOpen_openAndPingFails

paulbellamy:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Extracted from https://github.com/stellar/go/pull/3545. I fixed `TestOpen_openAndPingFails`. The two issues were, Open on port 0, and sometimes the error would be different due to IPv6. Now it handles both ipv4 and ipv6, as well as finding a free port for the test to run on.

### Why

Because it was intermittently failing and being annoying.

### Known limitations

[TODO or N/A]

### DIFF
--- a/exp/services/recoverysigner/internal/db/db_test.go
+++ b/exp/services/recoverysigner/internal/db/db_test.go
@@ -1,6 +1,9 @@
 package db
 
 import (
+	"fmt"
+	"net"
+	"regexp"
 	"testing"
 
 	"github.com/stellar/go/support/db/dbtest"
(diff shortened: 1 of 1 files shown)


exp/services/recoverysigner/internal/db/db_test.go
```diff
@@ -20,10 +23,25 @@ func TestOpen_openAndPingSucceeds(t *testing.T) {
 }
 
 func TestOpen_openAndPingFails(t *testing.T) {
-	sqlxDB, err := Open("postgres://localhost:0")
+	// Find an empty port
+	listener, err := net.Listen("tcp", ":0")
+	require.NoError(t, err)
+	port := listener.Addr().(*net.TCPAddr).Port
+	require.NoError(t, listener.Close())
+	// Slight race here with other stuff on the system, which could claim this port.
```
leighmcculloch, line 31: ❓ What problem is this change fixing? Is the reason for this change so that this code knows what the port is for the error check below? Instead of doing this how about changing the regex to just check for any integer. It isn't particularly important to verify the port in the error message.

leighmcculloch (approved):
Thanks for fixing this! One question below, 👍🏻 regardless.

exp/services/recoverysigner/internal/db/db_test.go
paulbellamy: Trying to dial to port `0` (as the test used to) on OSX says: `dial tcp [::1]:0: connect: can't assign requested address`. So, to get the test running correctly we need a real (unused) port number.

We could widen the regex to just check for a digit, but we use the port to connect to anyway, so may as well use it in the regex, I figured...

Issue #2097, opened 2020-01-09
exp/ingest/io: Remove support for v0 TransactionMeta


bartekn:
Ingestion of `TransactionMeta` with version 0 using `LedgerTransaction.GetChanges` is currently broken according to https://github.com/stellar/stellar-core/pull/944 (`LEDGER_ENTRY_STATE` is emitted when a ledger entry is first modified in a ledger). Because version 0 is not longer emitted in the latest version of stellar-core remove support for it in `expingest`. We can add it back in the future if there's a need for this (ex. ledgers ingested by legacy stellar-core versions).

`LedgerTransaction.GetChanges` should return error for version 0 meta.

bartekn:
Fixed in #2101.

PR #3555, opened 2021-04-20
services/horizon:  Make LedgerTransaction's Meta a private field

### LINKED ISSUE #3491: Make ingest.LedgerTransaction.Meta private
`LedgerTransaction.Meta` is low-level object and can be misused. We should make it private (`GetChanges()` should be used instead).

> @tamirms great catch! Yes, `transaction.Meta.Operations` will likely be empty because we're at V2 right now so the proper way would be to get `transaction.Meta.V2.Operations`. So yes, we should use `transaction.GetChanges()` and maybe we should make `LedgerTransaction.Meta` private (in a separate PR).

_Originally posted by @bartekn in https://github.com/stellar/go/pull/3483#r600465388_

2opremio:
This change transforms 

```go
type LedgerTransaction struct {
	...
	Meta       xdr.TransactionMeta
}
```

into 

```go
type LedgerTransaction struct {
	...
	meta       xdr.TransactionMeta
}
```

Unfortunately I've had to add a getter and a setter:

1. `func (t *LedgerTransaction) UnsafeSetMeta(meta xdr.TransactionMeta)`, needed for test fixtures
2. `(t *LedgerTransaction) UnsafeGetMeta() xdr.TransactionMeta)`, needed for db ingestion (see `transactionToRow(transaction ingest.LedgerTransaction, sequence uint32)`).

As a result the participants processor (`participantsForTransaction()`) now uses `func (t *LedgerTransaction) GetChanges()` which constitutes a change of behavior (for the better, I believe, although new participants may be found).

To make sure `participantsForTransaction()`works with old transactions, `GetChanges()` now also returns the changes of Meta version `0`.

In addition, and for consistency, `participantsForTransaction()` now uses `GetFeeChanges()` instead of `LedgerTransaction`s `FeeChanges` field.

I think `FeeChanges` should probably be made private in the future.

Fixes #3491

### DIFF
--- a/ingest/ledger_transaction.go
+++ b/ingest/ledger_transaction.go
@@ -47,7 +47,14 @@ func (t *LedgerTransaction) GetChanges() ([]Change, error) {
 	// Transaction meta
 	switch t.meta.V {
 	case 0:
-		return changes, errors.New("TransactionMeta.V=0 not supported")
+		if t.meta.Operations != nil {
+			for _, operationMeta := range *t.meta.Operations {
+				opChanges := GetChangesFromLedgerEntryChanges(
+					operationMeta.Changes,
+				)
+				changes = append(changes, opChanges...)
+			}
+		}
 	case 1:
 		v1Meta := t.meta.MustV1()
 		txChanges := GetChangesFromLedgerEntryChanges(v1Meta.TxChanges)
--- a/ingest/ledger_transaction.go
+++ b/ingest/ledger_transaction.go
@@ -14,7 +14,17 @@ type LedgerTransaction struct {
 	// Use LedgerTransaction.GetChanges() for higher level access to ledger
 	// entry changes.
 	FeeChanges xdr.LedgerEntryChanges
-	Meta       xdr.TransactionMeta
+	meta       xdr.TransactionMeta
+}
+
+// UnsafeSetMeta should only be used when you know what you are doing.
+func (t *LedgerTransaction) UnsafeSetMeta(meta xdr.TransactionMeta) {
+	t.meta = meta
+}
+
+// UnsafeGetMeta should only be used when you know what you are doing. Normally you should use GetChanges()
+func (t *LedgerTransaction) UnsafeGetMeta() xdr.TransactionMeta {
+	return t.meta
 }
 
 func (t *LedgerTransaction) txInternalError() bool {
@@ -35,11 +45,11 @@ func (t *LedgerTransaction) GetChanges() ([]Change, error) {
 	var changes []Change
 
 	// Transaction meta
-	switch t.Meta.V {
+	switch t.meta.V {
 	case 0:
 		return changes, errors.New("TransactionMeta.V=0 not supported")
 	case 1:
-		v1Meta := t.Meta.MustV1()
+		v1Meta := t.meta.MustV1()
 		txChanges := GetChangesFromLedgerEntryChanges(v1Meta.TxChanges)
 		changes = append(changes, txChanges...)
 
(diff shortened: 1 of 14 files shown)


ingest/ledger_transaction.go
```diff
@@ -35,11 +45,18 @@ func (t *LedgerTransaction) GetChanges() ([]Change, error) {
 	var changes []Change
 
 	// Transaction meta
-	switch t.Meta.V {
+	switch t.meta.V {
 	case 0:
-		return changes, errors.New("TransactionMeta.V=0 not supported")
+		if t.meta.Operations != nil {
```
bartekn, line 50: We shouldn't support V0 because it's broken in Stellar-Core, see: https://github.com/stellar/go/issues/2097.
> To make sure participantsForTransaction()works with old transactions, GetChanges() now also returns the changes of Meta version 0.

We don't need this because meta version is not related to the ledger but Stellar-Core version. I.e. Latest Stellar-Core will produce V2 for all the ledgers, also very old ledgers.

ingest/ledger_transaction_test.go
```diff
@@ -331,17 +331,6 @@ func TestMetaV2Order(t *testing.T) {
 
 }
 
-func TestMetaV0(t *testing.T) {
```
bartekn, line 334: Since we won't support V0, we can revert this test.

services/horizon/internal/ingest/processors/participants_processor.go
```diff
@@ -147,14 +115,17 @@ func participantsForTransaction(
 	if transaction.Envelope.IsFeeBump() {
 		participants = append(participants, transaction.Envelope.FeeBumpAccount().ToAccountId())
 	}
-
-	p, err := participantsForMeta(transaction.Meta)
+	changes, err := transaction.GetChanges()
+	if err != nil {
```
bartekn, line 119: > As a result the participants processor (participantsForTransaction()) now uses func (t *LedgerTransaction) GetChanges() which constitutes a change of behavior (for the better, I believe, although new participants may be found).

I believe this is the change described in the PR message. I think we shouldn't do that because it will affect all txs in which tx source is different than op source (because `GetChanges` also include fee changes). This is a breaking change we shouldn't introduce.

bartekn (commented):
LGTM, but we need to fix/check the issues from my comments.

services/horizon/internal/ingest/processors/participants_processor.go
```diff
@@ -108,35 +105,6 @@ func participantsForLedgerEntry(le xdr.LedgerEntry) *xdr.AccountId {
...
-
-func participantsForMeta(
-	meta xdr.TransactionMeta,
-) ([]xdr.AccountId, error) {
-	var participants []xdr.AccountId
-	if meta.Operations == nil {
-		return participants, nil
-	}
-
-	for _, op := range *meta.Operations {
```
2opremio, line 127: @bartekn given what you have mentioned about `V0`, doesn't this mean that this function doesn't really do anything?

ingest/ledger_transaction.go
2opremio: I was just trying to keep compatibility with what `participantsForMeta()` does (it only looks at the `V0`operations).

services/horizon/internal/ingest/processors/participants_processor.go
2opremio: OK, then I see very little point in this PR (since access to the meta field will remain the same, only through Getters/Setters).

I think we should just rename the field. Let me know if you agree.

2opremio: I've created  #3561

ingest/ledger_transaction.go
```diff
@@ -35,11 +45,18 @@ func (t *LedgerTransaction) GetChanges() ([]Change, error) {
 	var changes []Change
 
 	// Transaction meta
-	switch t.Meta.V {
+	switch t.meta.V {
```
Shaptic, line 48: I'm not sure what proper convention is for Golang, but should we also be using `.UnsafeGetMeta()` here (and elsewhere) instead of accessing `meta` directly within private methods, as well, or no?

2opremio:
Close in favor of #3561

Issue #3486, opened 2021-03-23
Duplicate UNSAFE_QUORUM entry in stellar-core temp config when reingesting range in Horizon + Captive Core Docker Setup


lorenzobersano:
### What version are you using?

Cloned this repo as in https://github.com/stellar/go/commit/17275e80a6821fd057ad3313b651a1ba07f9bf80

### What did you do?

I started a Horizon + Captive Core setup using the start script in pubnet mode as found here https://github.com/stellar/go/tree/master/services/horizon/docker.
The base setup works fine, but once I get into the Horizon container the `horizon db reingest range <from> <to>` doesn't work .

### What did you expect to see?

I expected to see the reingestion start correctly.

### What did you see instead?

I'm getting this error:
```
INFO[2021-03-23T14:47:27.292Z] Ingestion system initial state                current_state="reingestHistoryRange(fromLedger=18587775, toLedger=34573119, force=false)" pid=307 service=ingest
INFO[2021-03-23T14:47:27.292Z] Preparing ledger backend to retrieve range    from=18587775 pid=307 service=ingest to=34573119
INFO[2021-03-23T14:47:28.063Z] default: Config from /tmp/captive-stellar-core996790614/stellar-core.conf  pid=307 service=ingest subservice=stellar-core
ERRO[2021-03-23T14:47:28.065Z] default: Got an exception: Failed to parse '/tmp/captive-stellar-core996790614/stellar-core.conf' :Key UNSAFE_QUORUM already present at line 13  pid=307 service=ingest subservice=stellar-core
ERRO[2021-03-23T14:47:28.065Z] default: Please report this bug along with this log file if this was not expected  pid=307 service=ingest subservice=stellar-core
ERRO[2021-03-23T14:47:28.075Z] Error in ingestion state machine              current_state="reingestHistoryRange(fromLedger=18587775, toLedger=34573119, force=false)" error="error preparing range: error starting prepare range: opening subprocess: error running stellar-core: error waiting for `stellar-core new-db` subprocess: error waiting for `stellar-core [new-db]` subprocess: exit status 1" next_state=stop pid=307 service=ingest
INFO[2021-03-23T14:47:28.076Z] Shut down                                     pid=307 service=ingest
2021/03/23 14:47:28 error preparing range: error starting prepare range: opening subprocess: error running stellar-core: error waiting for `stellar-core new-db` subprocess: error waiting for `stellar-core [new-db]` subprocess: exit status 1
```

tamirms:
[...]

Issue #3442, opened 2021-03-04
Feature Request: Add toml file specifically for captive core


tamirms:
### What problem does your feature solve?

The configuration space for the captive core append path file is not fully documented. Operators don't know what is possible to configure in captive core so they try to copy parts of the stellar-core.cfg file from their stand-alone stellar core instances into the captive core configuration. However, it's not clear which fields  can and can't be set, which forces operators to go through a trial and error process of trying different combinations of fields until they get their setup to work.

### What would you like to see?

We should create a restricted toml file just for captive core which is a subset of [stellar-core.cfg](https://github.com/stellar/stellar-core/blob/master/docs/stellar-core_example.cfg).

The restricted toml file should be documented with an example cfg file just like the one in the stellar core repo: https://github.com/stellar/stellar-core/blob/master/docs/stellar-core_example.cfg

We should strive to make sure that configuration is not duplicated between horizon command line flags and the captive core toml file.

ire-and-curses:
Let's discuss this. What are the pros and cons?

Shaptic:
One of the cons we should definitely keep in mind is that by pulling in a TOML parser, we're now exposing a potential attack vector to malicious user input. I'm not sure we should worry about that much, given that if you can control the config file, you can control everything else about Horizon anyway, but it's still worth keeping in mind, especially given our recent discussions about security.

tamirms:
If we have a separate toml file for captive core we should consider allowing `LOG_FILE_PATH` to be configured. If `LOG_FILE_PATH` is set we will store captive core logs in that location and remove captive core output from the horizon log stream. If `LOG_FILE_PATH` is not set then we can default to the current behavior of including captive core logs in the horizon log stream.

One point of duplication we need to figure out is how to deal with the configuration of history archive urls. History archive urls are part of horizon configuration and stellar core configuration. Ideally we should not have to configure both. But the difficulty is that stellar core's configuration is more expressive than horizon. It allows for specifying how core should obtain the history archives (e.g. `curl -sf %s/{0} -o {1}` or `aws s3 cp s3://history.stellar.org/{0} {1}`)

Shaptic:
After the discussion with ops, I think this has progressed into a far bigger, arguably meta-issue. 

[...]

PR #3558, opened 2021-04-21
ingest/ledgerbackend: Parse captive core toml to implement better merging

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This PR fixes https://github.com/stellar/go/issues/3486 and lays the groundwork for implementing https://github.com/stellar/go/issues/3442 .

This PR changes how we interpret captive core append file (configured via `--captive-core-config-append-path`). Previously, we automatically generated the stellar core toml file for the captive core subprocess in ingest/ledgerbackend/stellar_core_runner.go by presetting some defaults and then concatenating the generated configuration with the user supplied append file.

This strategy did not work out so well because sometimes horizon operators included configuration in the append file which was already configured in the presets generated by ingest/ledgerbackend/stellar_core_runner.go. An example of this type of error can be found in https://github.com/stellar/go/issues/3486 .

Now, in this PR, instead of generating the stellar core toml file via simple string concatenation we actually parse the user supplied append file and try to do more robust merging between the user supplied configuration and the captive core preset defaults. The merging code is implemented in ingest/ledgerbackend/toml.go .

The merging strategy is unfortunately complicated by the fact that captive core configuration can come from 3 places:

1) User supplied append file
2) horizon command line flag (e.g. `--captive-core-peer-port`, `--captive-core-http-port`, etc)
3) default values which we preset (e.g. `DISABLE_XDR_FSYNC`)

In theory the user supplied append file should only contain the quorum set but in practice we have never validated the append file to ensure that it configures a quorum set and nothing else. We could simplify the merging logic in this PR by choosing to validate the append file in this way. I decided not to take that approach because I think that it would be a breaking change for many horizon operators (including SDF) who have configuration in the append file which is not related to quorum sets. If you disagree with this design decision please let me know.

The merging algorithm generally works as follows:

If there is a configuration field which is set both in source (1) and source (2) we check that both sources provide equivalent values, otherwise we return an error.

The exception to this rule is the history archive configuration. We take the history archive configuration from source (1) and completely disregard what is provided in source (2) unless there is no history archive configuration in source (1). The reason for this is that it's difficult to compare the history archive values from source (1) and source (2).

If a configuration field is defined in source (1) but not source (2), we will use the field which is explicitly defined. The same logic applies if the field is defined in source (2) but not source (1).

Finally, if there is no configuration defined in either source (1) or source (2), we use the value from source (3).

In the next PR, my plan is to fully implement https://github.com/stellar/go/issues/3442 by introducing a new horizon command line flag `--captive-core-config-file` which deprecates `--captive-core-config-append-path`. The difference between `--captive-core-config-file` and `--captive-core-config-append-path` is that the validation is stricter.

`--captive-core-config-file` will not allow any unknown fields to be present in the toml file.

Furthermore, if `--captive-core-config-file` is provided then you will not be allowed to use any horizon command line flags to configure captive core. In other words, `--captive-core-config-file` deprecates source (2).

Along with defining  `--captive-core-config-file`, the next PR will also add documentation for the captive core configuration file in an captive-core_example.cfg file (similar to how stellar core does it https://github.com/stellar/stellar-core/blob/master/docs/stellar-core_example.cfg ). The example file will document the full range of configuration options available to captive core instances.

Unfortunately there will be some slight differences between the stellar core example cfg and the captive core example cfg because captive core has different defaults compared to stellar core. For example, captive core sets `DISABLE_XDR_FSYNC` to true by default whereas stellar core sets `DISABLE_XDR_FSYNC` to false by default. We should think carefully about each of the cases where captive core diverges from stellar core.
 
### Known limitations

The golang toml library doesn't handle nested tables in a way which is consistent with stellar core. As a workaround I implemented some hacky code to flatten nested tables using string manipulation.


ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"context"
 	"fmt"
+	"github.com/BurntSushi/toml"
```
Shaptic, line 7: The very first line of this repo is 

> THIS PROJECT IS UNMAINTAINED

Should we use something more recent like [this](https://github.com/pelletier/go-toml), instead?

ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -112,84 +106,154 @@ func newStellarCoreRunner(config CaptiveCoreConfig, mode stellarCoreRunnerMode)
...
+		runner.log.Debugf("captive core config file contents:\n%s", conf)
 	}
 
 	return runner, nil
 }
 
-func (r *stellarCoreRunner) generateConfig() (string, error) {
-	if r.mode == stellarCoreRunnerModeOnline && r.configAppendPath == "" {
-		return "", errors.New("stellar-core append config file path cannot be empty in online mode")
+// writes the stellar core config file to the given location; called by the
```
Shaptic, line 129: by the??? I'm dying in suspense here :laughing:

ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -112,84 +106,154 @@ func newStellarCoreRunner(config CaptiveCoreConfig, mode stellarCoreRunnerMode)
...
+			return "", fmt.Errorf(
+				"HTTP_PORT in captive core config file: %d does not match Horizon captive-core-http-port value: %d",
+				generatedConfig.HTTPPort,
+				*config.HTTPPort,
+			)
+		} else if !def && config.HTTPPort != nil {
+			generatedConfig.HTTPPort = *config.HTTPPort
+		} else if !def && config.HTTPPort == nil {
+			generatedConfig.HTTPPort = defaultHTTPPort
+		}
```
Shaptic, line 171: If only Go had generics... this set of similar blocks is begging for a helper function!

ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -112,84 +106,154 @@ func newStellarCoreRunner(config CaptiveCoreConfig, mode stellarCoreRunnerMode)
...
-	if r.configAppendPath != "" {
-		appendConfigContents, err := ioutil.ReadFile(r.configAppendPath)
-		if err != nil {
-			return "", errors.Wrap(err, "reading quorum config file")
+	if !generatedConfig.historyIsConfigured() {
+		generatedConfig.HistoryEntries = map[string]history{}
+		for i, val := range config.HistoryArchiveURLs {
+			generatedConfig.HistoryEntries[fmt.Sprintf("h%d", i)] = history{
+				Get: fmt.Sprintf("curl -sf %s/{0} -o {1}", val),
+			}
```
Shaptic, line 244: Do we want to do validation on the URLs here?

ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -112,84 +106,154 @@ func newStellarCoreRunner(config CaptiveCoreConfig, mode stellarCoreRunnerMode)
...
 	}
 
-	lines = []string{}
-	for i, val := range r.historyURLs {
-		lines = append(lines, fmt.Sprintf("[HISTORY.h%d]", i))
-		lines = append(lines, fmt.Sprintf(`get="curl -sf %s/{0} -o {1}"`, val))
+	var sb strings.Builder
+	if _, err = sb.WriteString("# Generated file -- do not edit\n"); err != nil {
+		return "", errors.Wrap(err, "could not write toml header")
+	}
```
Shaptic, line 251: According to the [documentation](https://golang.org/pkg/strings/#Builder.WriteString), this never fails!

>  WriteString appends the contents of s to b's buffer. It returns the length of s and a nil error.

ingest/ledgerbackend/stellar_core_runner_test.go
```diff
@@ -39,27 +39,24 @@ func TestGenerateConfig(t *testing.T) {
...
 				ConfigAppendPath:   testCase.appendPath,
 				StoragePath:        "./test-temp-dir",
-				PeerPort:           12345,
 				Context:            context.Background(),
 				NetworkPassphrase:  "Public Global Stellar Network ; September 2015",
-			}, testCase.mode)
-			assert.NoError(t, err)
+			}
+			config.SetHTTPPort(6789)
+			config.SetPeerPort(12345)
```
Shaptic, line 51: `const`s at the top? not sure

Shaptic (commented):
I know this is in-progress, but just a few thoughts

ingest/doc_test.go
```diff
@@ -71,7 +71,6 @@ func Example_changes() {
 	backend, err := ledgerbackend.NewCaptive(
 		ledgerbackend.CaptiveCoreConfig{
 			BinaryPath:         "/bin/stellar-core",
-			ConfigAppendPath:   "/opt/stellar-core.cfg",
```
tamirms, line 74: need to update this example

ingest/ledgerbackend/stellar_core_runner.go
tamirms: I don't think it's the right place to validate the urls. those urls come from the horizon flag so if we want to validate them we should do it closer to the flag parsing code

ingest/ledgerbackend/stellar_core_runner_test.go
tamirms: not sure if it's worth it for these test cases

ingest/ledgerbackend/stellar_core_runner_test.go
```diff
@@ -12,6 +12,12 @@ import (
 	"github.com/stellar/go/support/log"
 )
 
+func newUint(v uint) *uint {
+	p := new(uint)
+	*p = v
+	return p
+}
```
bartekn, line 19: Nit, `v` is a copy so I think below will work too:
```suggestion
func newUint(v uint) *uint {
	return &v
}
```

ingest/ledgerbackend/toml.go
```diff
@@ -0,0 +1,345 @@
...
+		if err != nil {
+			return nil, errors.Wrap(err, "could not load toml path")
+		}
+		if err = captiveCoreToml.Unmarshal(text); err != nil {
+			return nil, errors.Wrap(err, "could not unmarshall captive core toml")
+		}
+		if err = captiveCoreToml.validate(params); err != nil {
+			return nil, errors.Wrap(err, "invalid captive core toml")
+		}
+	}
```
bartekn, line 155: Maybe we should have another constructor like `NewCaptiveCoreTomlFromFile(path string)`? It can be a little bit more clear and helpful in places like `flags.go` in Horizon.

bartekn (commented):
It's looking good! I don't have any major comments except `NewCaptiveCoreTomlFromFile`.

ingest/ledgerbackend/stellar_core_runner.go
ire-and-curses: Since we already use this dependency and we know it works for us, I'm in favour of using it here. Let's also open an issue to review it and move to something more maintainable if that makes sense.

tamirms:
@stellar/horizon-committers this ready for a final review now, PTAL!

exp/services/captivecore/main.go
```diff
@@ -14,16 +14,17 @@ import (
 	"github.com/stellar/go/ingest/ledgerbackend"
 	"github.com/stellar/go/network"
 	"github.com/stellar/go/support/config"
+	support "github.com/stellar/go/support/config"
```
bartekn, line 17: Isn't it imported line above?
```suggestion
```

exp/services/captivecore/main.go
```diff
@@ -98,12 +99,13 @@ func main() {
...
-			OptType:     types.Uint,
-			FlagDefault: uint(11626),
-			Required:    false,
-			Usage:       "HTTP port for captive core to listen on (0 disables the HTTP server)",
+			Name:           "stellar-captive-core-http-port",
+			ConfigKey:      &captiveCoreTomlParams.HTTPPort,
+			OptType:        types.Uint,
+			CustomSetValue: support.SetOptionalUint,
+			Required:       false,
+			FlagDefault:    uint(0),
```
paulbellamy, line 107: Should this be 11626?

ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -112,84 +105,53 @@ func newStellarCoreRunner(config CaptiveCoreConfig, mode stellarCoreRunnerMode)
...
-	for i, val := range r.historyURLs {
-		lines = append(lines, fmt.Sprintf("[HISTORY.h%d]", i))
-		lines = append(lines, fmt.Sprintf(`get="curl -sf %s/{0} -o {1}"`, val))
+	if !captiveCoreToml.QuorumSetIsConfigured() {
+		return nil, errors.New("stellar-core append config file does not define any quorum set")
 	}
-	result += strings.Join(lines, "\n")
 
-	return result, nil
+	text, err := captiveCoreToml.Marshall()
```
paulbellamy, line 150: Typo?
```suggestion
	text, err := captiveCoreToml.Marshal()
```

ingest/ledgerbackend/toml.go
```diff
@@ -0,0 +1,509 @@
...
+
+// unflattenTables is the inverse of flattenTables, it restores the
+// text back to its original form by replacing all placeholders with their
+// original values.
+func unflattenTables(text string, tablePlaceHolders *placeholders) string {
+	re := regexp.MustCompile(`\[.*\]`)
+
+	return re.ReplaceAllStringFunc(text, func(match string) string {
+		insideBrackets := match[1 : len(match)-1]
+		original := tablePlaceHolders.get(insideBrackets)
```
paulbellamy, line 174: Super nit:
```suggestion
		original, ok := tablePlaceHolders.get(insideBrackets)
```

ingest/ledgerbackend/toml.go
```diff
@@ -0,0 +1,509 @@
...
+		if len(original) == 0 {
+			return match
+		}
+		return "[" + original + "]"
+	})
+}
+
+// Marshall serializes the CaptiveCoreToml into a toml document.
+func (c *CaptiveCoreToml) Marshall() ([]byte, error) {
```
paulbellamy, line 183: Typo?
```suggestion
func (c *CaptiveCoreToml) Marshal() ([]byte, error) {
```

ingest/ledgerbackend/toml.go
```diff
@@ -0,0 +1,509 @@
...
+	return tree.Unmarshal(dest)
+}
+
+func (c *CaptiveCoreToml) unmarshal(data []byte) error {
+	var body captiveCoreTomlValues
+	quorumSetEntries := map[string]QuorumSet{}
+	historyEntries := map[string]History{}
+	// The toml library has trouble with nested tables so we need to flatten all nested
+	// QUORUM_SET and HISTORY tables as a workaround.
+	// In Marshall() we apply the inverse process to unflatten the nested tables.
```
paulbellamy, line 220:
```suggestion
	// In Marshal() we apply the inverse process to unflatten the nested tables.
```

ingest/ledgerbackend/toml.go
```diff
@@ -0,0 +1,509 @@
...
+	captiveCoreToml.setDefaults(params)
+	return &captiveCoreToml, nil
+}
+
+// CatchupToml returns a new CaptiveCoreToml instance based off the existing
+// instance with some modifications which are suitable for running
+// the catchup command on captive core.
+func (c *CaptiveCoreToml) CatchupToml() (*CaptiveCoreToml, error) {
+	// clone the existing toml file
+	data, err := c.Marshall()
```
paulbellamy, line 331:
```suggestion
	data, err := c.Marshal()
```

ingest/ledgerbackend/toml.go
```diff
@@ -0,0 +1,509 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "could not clone toml")
+	}
+	var offline CaptiveCoreToml
+	if err = offline.unmarshal(data); err != nil {
+		return nil, errors.Wrap(err, "could not unmarshall captive core toml")
+	}
```
paulbellamy, line 338: Should we have one of these, just to make this more clear what we're doing here?
```go
func (c *CaptiveCoreToml) clone() (*CaptiveCoreToml, error) {
```

bartekn (approved):

paulbellamy (approved):

tamirms:
This PR fixes #3486 and lays the groundwork for implementing #3442 .

This PR changes how we interpret captive core append file (configured via --captive-core-config-append-path). Previously, we automatically generated the stellar core toml file for the captive core subprocess in ingest/ledgerbackend/stellar_core_runner.go by presetting some defaults and then concatenating the generated configuration with the user supplied append file.

This strategy did not work out so well because sometimes horizon operators included configuration in the append file which was already configured in the presets generated by ingest/ledgerbackend/stellar_core_runner.go. An example of this type of error can be found in #3486 .

Now, in this PR, instead of generating the stellar core toml file via simple string concatenation we actually parse the user supplied append file and try to do more robust merging between the user supplied configuration and the captive core preset defaults. The merging code is implemented in ingest/ledgerbackend/toml.go .

The merging strategy is unfortunately complicated by the fact that captive core configuration can come from 3 places:

User supplied append file
horizon command line flag (e.g. --captive-core-peer-port, --captive-core-http-port, etc)
default values which we preset (e.g. DISABLE_XDR_FSYNC)
In theory the user supplied append file should only contain the quorum set but in practice we have never validated the append file to ensure that it configures a quorum set and nothing else. We could simplify the merging logic in this PR by choosing to validate the append file in this way. I decided not to take that approach because I think that it would be a breaking change for many horizon operators (including SDF) who have configuration in the append file which is not related to quorum sets. If you disagree with this design decision please let me know.

The merging algorithm generally works as follows:

If there is a configuration field which is set both in source (1) and source (2) we check that both sources provide equivalent values, otherwise we return an error.

The exception to this rule is the history archive configuration. We take the history archive configuration from source (1) and completely disregard what is provided in source (2) unless there is no history archive configuration in source (1). The reason for this is that it's difficult to compare the history archive values from source (1) and source (2).

If a configuration field is defined in source (1) but not source (2), we will use the field which is explicitly defined. The same logic applies if the field is defined in source (2) but not source (1).

Finally, if there is no configuration defined in either source (1) or source (2), we use the value from source (3).

In the next PR, my plan is to fully implement #3442 by introducing a new horizon command line flag --captive-core-config-file which deprecates --captive-core-config-append-path. The difference between --captive-core-config-file and --captive-core-config-append-path is that the validation is stricter.

--captive-core-config-file will not allow any unknown fields to be present in the toml file.

Furthermore, if --captive-core-config-file is provided then you will not be allowed to use any horizon command line flags to configure captive core. In other words, --captive-core-config-file deprecates source (2).

Along with defining --captive-core-config-file, the next PR will also add documentation for the captive core configuration file in an captive-core_example.cfg file (similar to how stellar core does it https://github.com/stellar/stellar-core/blob/master/docs/stellar-core_example.cfg ). The example file will document the full range of configuration options available to captive core instances.

Unfortunately there will be some slight differences between the stellar core example cfg and the captive core example cfg because captive core has different defaults compared to stellar core. For example, captive core sets DISABLE_XDR_FSYNC to true by default whereas stellar core sets DISABLE_XDR_FSYNC to false by default. We should think carefully about each of the cases where captive core diverges from stellar core.

PR #3560, opened 2021-04-21
services/horizon: Test full coverage of operation type switches

2opremio:
Part of #3430

I incorporated new tests, which use `xdr`'s  `operationTypeMap` (now  exported `OperationTypeToStringMap`) to go over all possible operations, even when we incorporate new ones.

The idea is that, from now on, every time we add a new XDR operation (i.e. through a CAP), the developer gets guided on what needs to be modified based on the failing tests.

The tests I added can be considered finicky, but it's hard to do it generically in a robust way.

### DIFF
--- a/protocols/horizon/operations/main_test.go
+++ b/protocols/horizon/operations/main_test.go
@@ -1,0 +1,32 @@
+package operations
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/stellar/go/xdr"
+)
+
+func TestTypeNamesAllCovered(t *testing.T) {
+	for typ, s := range xdr.OperationTypeToStringMap {
+		_, ok := TypeNames[xdr.OperationType(typ)]
+		assert.True(t, ok, s)
+	}
+}
+
+func TestUnmarshalOperationAllCovered(t *testing.T) {
+	mistmatchErr := errors.New("Invalid operation format, unable to unmarshal json response")
+	for typ, s := range xdr.OperationTypeToStringMap {
+		_, err := UnmarshalOperation(typ, []byte{})
+		assert.Error(t, err, s)
+		// it shoudl be a parsing error, not the default branch
+		assert.NotEqual(t, mistmatchErr, err, s)
+	}
+
+	// make sure the check works for an unreasonable operation type
+	_, err := UnmarshalOperation(200000, []byte{})
+	assert.Error(t, err)
+	assert.Equal(t, mistmatchErr, err)
+}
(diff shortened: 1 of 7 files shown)


tamirms (approved):

protocols/horizon/operations/main_test.go
```diff
@@ -0,0 +1,32 @@
...
+		_, err := UnmarshalOperation(typ, []byte{})
+		assert.Error(t, err, s)
+		// it shoudl be a parsing error, not the default branch
+		assert.NotEqual(t, mistmatchErr, err, s)
+	}
+
+	// make sure the check works for an unreasonable operation type
+	_, err := UnmarshalOperation(200000, []byte{})
+	assert.Error(t, err)
+	assert.Equal(t, mistmatchErr, err)
```
bartekn, line 31: Nit, there's [`EqualError`](https://pkg.go.dev/github.com/stretchr/testify/assert#EqualError) that can be useful here.

protocols/horizon/operations/main_test.go
```diff
@@ -0,0 +1,32 @@
...
+		assert.True(t, ok, s)
+	}
+}
+
+func TestUnmarshalOperationAllCovered(t *testing.T) {
+	mistmatchErr := errors.New("Invalid operation format, unable to unmarshal json response")
+	for typ, s := range xdr.OperationTypeToStringMap {
+		_, err := UnmarshalOperation(typ, []byte{})
+		assert.Error(t, err, s)
+		// it shoudl be a parsing error, not the default branch
```
bartekn, line 24:
```suggestion
		// it should be a parsing error, not the default branch
```

services/horizon/internal/codes/main_test.go
```diff
@@ -7,6 +7,34 @@ import (
...
+		result := xdr.OperationResult{
+			Code: xdr.OperationResultCodeOpInner,
+			Tr: &xdr.OperationResultTr{
+				Type: xdr.OperationType(typ),
+			},
+		}
+		f := func() {
+			ForOperationResult(result)
+		}
+		// it must panic because the operation restult is not set
```
bartekn, line 21:
```suggestion
		// it must panic because the operation result is not set
```

services/horizon/internal/codes/main_test.go
```diff
@@ -7,6 +7,34 @@ import (
...
+	result := xdr.OperationResult{
+		Code: xdr.OperationResultCodeOpInner,
+		Tr: &xdr.OperationResultTr{
+			Type: xdr.OperationType(200000),
+		},
+	}
+	f := func() {
+		ForOperationResult(result)
+	}
+	// it doesn'' panic because it doesn't branch out into the operation type
```
bartekn, line 34:
```suggestion
	// it doesn't panic because it doesn't branch out into the operation type
```

bartekn (approved):
Awesome!

protocols/horizon/operations/main_test.go
2opremio: I didn't know, thanks! However, the error is currently identical, so I think it's fine.

bartekn: Yeah, I mean, you can merge `Error` and `Equal` intro `EqualError`.

PR #3562, opened 2021-04-22
horizon: Test consistency and coverage of effect typenames (and cleanup unused effects)

2opremio:
Part of #3430

### DIFF
--- a/protocols/horizon/effects/main.go
+++ b/protocols/horizon/effects/main.go
@@ -46,7 +46,7 @@ const (
 	// unused
 	// EffectAccountInflationDestinationUpdated effects occur when an account changes its
 	// inflation destination.
-	// EffectAccountInflationDestinationUpdated EffectType = 7 // from set_options
+	EffectAccountInflationDestinationUpdated EffectType = 7 // from set_options
 
 	// signer effects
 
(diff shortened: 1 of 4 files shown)


tamirms (approved):

protocols/horizon/effects/main.go
```diff
@@ -43,9 +43,10 @@ const (
 	// account flags, either clearing or setting.
 	EffectAccountFlagsUpdated EffectType = 6 // from set_options
 
+	// unused
 	// EffectAccountInflationDestinationUpdated effects occur when an account changes its
 	// inflation destination.
-	EffectAccountInflationDestinationUpdated EffectType = 7 // from set_options
+	// EffectAccountInflationDestinationUpdated EffectType = 7 // from set_options
```
bartekn, line 49: It looks like Horizon is producing it here:
https://github.com/stellar/go/blob/f362e0c5d7f8a013d92ff191d37c1daac0555a2f/services/horizon/internal/ingest/processors/effects_processor.go#L567-L573
I think it's worth keeping the value here.

2opremio: Really good catch, thanks @bartekn

PR #3563, opened 2021-04-22
services/horizon/txsub: Batch status check queries for pending transactions in txsub

### LINKED ISSUE #3552: Batch status check queries for pending transactions in txsub

2.1.0

### What did you do?

@tomerweller was running into performance issues when submitting thousands of transactions to a standalone network. It looks like the problem is related to this code:
https://github.com/stellar/go/blob/56d94a2f370fde4e8a7500658cc7fa57ddfc08e4/services/horizon/internal/txsub/system.go#L307-L325

With many pending transactions getting its' results, even when separate queries are fast, can cumulate to a very long updates time (few seconds time). The chart below shows the time distribution of tx status updates after an example ledger ingestion:
Screenshot 2021-04-20 at 00 21 07

When checking the status of pending transaction we should send a single batch query instead of thousands queries asking for a single tx result. This should make this magnitudes faster.

When working on this please remember about two things:
* Postgres param limit of around 65k. If a batch query needs to send more params it should be called multiple times.
* The query used right now is searching for both normal txs and fee bump txs. The batch query should do the same thing.

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This commit modifies `txsub.Tick` to check the status of transactions in the queue by sending a batch query instead of sending separate queries to get status of each transaction in the queue.

Close #3552.

### Why

This was done due performance reasons. If the number of transactions in the queue is large this slows down entire `Tick` function. Data in https://github.com/stellar/go/issues/3552 suggest that previous method can take even 4s per `Tick` execution for around 1000 transactions in the queue.

### Known limitations

N/A


services/horizon/internal/txsub/system_test.go
```diff
@@ -355,6 +355,8 @@ func (suite *SystemTestSuite) TestTick_FinishesTransactions() {
 }
 
 func (suite *SystemTestSuite) TestTickFinishFeeBumpTransaction() {
+	// Temporarily skip this test
+	return
```
bartekn, line 359: @stellar/horizon-committers I'm temporary skipping this test because I'm thinking that a behaviour tested here is wrong.

What is happening here is we simulate submitting a transaction `T` and then updating it using results of fee bumped tx `F` (with the previously mentioned tx `T` as inner transaction of fee bump `F`). Should we do that? Fee bump tx has a different hash, has a different structure in `TransactionResult` etc. I think the correct behaviour would be to timeout `T` indefinitely (returning an error can be confusing).

Thoughts?

Shaptic: I may be missing a little bit of context covered in our meeting, but why wouldn't we just return `tx_insufficient_fee` for T? Are we trying to be "too helpful" to our users in that we are returning the fee bump result for their non-bumped tx?

bartekn: `tx_insufficient_fee` can introduce double spend because the submitter may think that transaction did not go through (but it did, embedded in a fee bump).

Shaptic: Just to make sure I understand the flow, it's like this, right?

-> Submit T
<- Get insufficient_fee
-> Submit F (inner T)
<- Success
-> Get T status
<- See success (currently), but I claim they should continue to see insufficient_fee

Isn't it the responsibility of the submitter to see that F succeeded and thus not resubmit T?

bartekn: > Just to make sure I understand the flow, it's like this, right?

That looks correct, except in point 2 it can be just timeout.

> Isn't it the responsibility of the submitter to see that F succeeded and thus not resubmit T?

I think that the problem is anyone can wrap any transaction inside fee bump and submit it.

Shaptic: Wait, what do you mean anyone? Anyone can resubmit a failed transaction w/in a fee-bump and they don't need it re-signed?

tamirms: > What is happening here is we simulate submitting a transaction T and then updating it using results of fee bumped tx F (with the previously mentioned tx T as inner transaction of fee bump F). Should we do that?

this behavior is intentional because anyone can fee bump your transaction. so if you submit your transaction and someone else fee bumps it and your ledger gets included in the ledger we want you to know that your ledger was included

bartekn: OK, reverted the behaviour in https://github.com/stellar/go/pull/3563/commits/ca5e147b9b4380479b2b1a97f9ac0a13e94c0622.

services/horizon/internal/db2/history/transaction.go
```diff
@@ -28,6 +28,25 @@ func (q *Q) TransactionByHash(ctx context.Context, dest interface{}, hash string
 	return q.Get(ctx, dest, union)
 }
 
+// TransactionsByHashesSinceLedger fetches transactions from the `history_transactions`
+// table which match the given hash since the given ledger sequence (for perf reasons).
+func (q *Q) TransactionsByHashesSinceLedger(ctx context.Context, dest interface{}, hashes []string, sinceLedgerSeq uint32) error {
```
tamirms, line 33: why is it necessary to add the `ht.ledger_sequence >= sinceLedgerSeq` condition? I thought the transaction hash is globally unique?

services/horizon/internal/txsub/system.go
```diff
@@ -304,23 +307,58 @@ func (sys *System) Tick(ctx context.Context) {
...
+			logger.WithError(err).Error("error getting latest history ledger")
+			return
 		}
 
-		if _, ok := err.(*FailedTransactionError); ok {
-			logger.WithField("hash", hash).Debug("finishing open submission")
-			sys.Pending.Finish(ctx, hash, Result{Transaction: tx, Err: err})
-			continue
+		var sinceLedgerSeq int32 = int32(latestLedger)
+		sinceLedgerSeq -= 100
```
2opremio, line 320: @bartekn Isn't `100` a bit too conservative? Wouldn't this lead to unnecessary extra iterations when the number of pending transactions is high? (In the body of the PR you mentioned 4s-ticks for 1000 transactions, but I don't know if 1000 transactions is realistic ).

services/horizon/internal/db2/history/transaction.go
2opremio, line 33: Do we really need the extra performance provided by `sinceLedgerSeq`?

* It complicates the logic
* It makes the code harder to read (i didn't initially understand the need of it, since hashes were also provided)
* It may actually lead to a longer total execution time due to multiple loop iterations (due to iterating over `sinceLedgerSeq`.

services/horizon/internal/txsub/system.go
```diff
@@ -304,23 +307,58 @@ func (sys *System) Tick(ctx context.Context) {
...
+			txMap[tx.TransactionHash] = tx
+			if tx.InnerTransactionHash.Valid {
+				txMap[tx.InnerTransactionHash.String] = tx
+			}
+		}
+
+		for _, hash := range pending {
+			tx, found := txMap[hash]
+			if !found {
+				continue
```
2opremio, line 343: How come we can happily continue here?

2opremio: Couldn't it be that the transaction is in the DB but isn't found by `TransactionsByHashesSinceLedger()` due to the `sinceLedgerSeq` passed?

services/horizon/internal/txsub/system.go
2opremio: I don't understand the logic around subtracting 100 to the latest history ledger.

services/horizon/internal/db2/history/transaction.go
bartekn: Unfortunately the query without the `ledger_sequence` limit is extremely slow. I realized it after to.mer tested it and added a fix in https://github.com/stellar/go/pull/3563/commits/e8701abc90e0ce8774f5efab0b4e7dce2573fb1e. Technically the 100 ledgers is too much because `TransactionsByHashesSinceLedger` is used only when checking the status of transactions in `Tick` so this could be 12 (one minute of ledgers).

services/horizon/internal/txsub/system.go
bartekn: Let's discuss here: https://github.com/stellar/go/pull/3563#discussion_r657978386.

services/horizon/internal/db2/history/transaction.go
bartekn: Let's discuss here: https://github.com/stellar/go/pull/3563#discussion_r657978386.

services/horizon/internal/txsub/system.go
bartekn: If `txMap` doesn't contain the `hash` it means that tx with `hash` has not been included in the ledger yet. So we continue to the next pending transaction.

bartekn:
@2opremio @tamirms thanks for the comments. I decided to put off this PR because of the issues found while testing 2.5.0 (Captive-Core persistent storage). I answered your questions and we can include this in the next release.

services/horizon/internal/txsub/system.go
```diff
@@ -304,23 +307,58 @@ func (sys *System) Tick(ctx context.Context) {
...
+		if err != nil {
+			logger.WithError(err).Error("error getting latest history ledger")
+			return
 		}
 
-		if _, ok := err.(*FailedTransactionError); ok {
-			logger.WithField("hash", hash).Debug("finishing open submission")
-			sys.Pending.Finish(ctx, hash, Result{Transaction: tx, Err: err})
-			continue
+		var sinceLedgerSeq int32 = int32(latestLedger)
```
tamirms, line 319: `sinceLedgerSeq := int32(latestLedger) - 100`

services/horizon/internal/db2/history/transaction.go
tamirms, line 33: instead of taking a `dest interface{}` this function should return `([]history.Transaction, error)`

tamirms (approved):

services/horizon/internal/db2/history/transaction.go
bartekn: Fixed in 6d0e082a3a90a3bca5ce2475222b5abdef92b9dc.

PR #3567, opened 2021-04-23
services/horizon: Add 1s timeout to horizon.App.Tick

### LINKED ISSUE #3344: Add timeouts to all functions called from Tick()

paulbellamy:
Fixes #3344

<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add 1s timeout to total runtime of `horizon.App.Tick`.

### Why

See https://github.com/stellar/go/issues/3344

### Known limitations

Just a single timeout, not per-item. But that seems to make sense given we want them all to complete within 1s total.

### DIFF
--- a/services/horizon/internal/app.go
+++ b/services/horizon/internal/app.go
@@ -88,13 +88,15 @@ func (a *App) GetCoreSettings() actions.CoreSettings {
 	return a.coreSettings.get()
 }
 
+const tickerDuration = 1 * time.Second
+
 // NewApp constructs an new App instance from the provided config.
 func NewApp(config Config) (*App, error) {
 	a := &App{
 		config:         config,
 		ledgerState:    &ledger.State{},
 		horizonVersion: app.Version(),
-		ticks:          time.NewTicker(1 * time.Second),
+		ticks:          time.NewTicker(tickerDuration),
 		done:           make(chan struct{}),
 	}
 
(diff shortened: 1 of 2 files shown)


services/horizon/internal/app.go
```diff
@@ -530,7 +534,12 @@ func (a *App) run() {
 	for {
 		select {
 		case <-a.ticks.C:
-			a.Tick()
+			ctx, cancel := context.WithTimeout(a.ctx, tickerDuration)
```
Shaptic, line 537: Will this change cause more timeouts, or is the behavior identical to before just with better context management?

tamirms (approved):

services/horizon/internal/app.go
paulbellamy: Well, before they would never timeout, it would just launch another new one every second (ish), and hope the previous had finished. So by definition, yeah this'll cause more timeouts.

services/horizon/internal/app.go
```diff
@@ -88,13 +88,15 @@ func (a *App) GetCoreSettings() actions.CoreSettings {
 	return a.coreSettings.get()
 }
 
+const tickerDuration = 1 * time.Second
```
bartekn, line 91: Just a suggestion regarding the name, but I think we should increase timeout to 10 seconds. The aim of this PR is to make sure ticker doesn't hang indefinitely but called `Tick` functions can take more than 10 seconds in certain circumstances. Ex. txsub (without #3563) can take a few sec. when many txs are submitted, possibly reap can be slow with many ledgers to delete.
```suggestion
const maxTickerDuration = 10 * time.Second
```

bartekn (approved):
LGTM. We should just increase the duration.

services/horizon/internal/app.go
paulbellamy: Are you also suggesting we change it to ensure there is only a single `.Tick()` running at a time?

bartekn: Can you elaborate? Before and after this change there can be only one `App.Tick()` running concurrently because of the way `App.run()` works.

paulbellamy: i.e. Are you suggesting we should attempt to launch a `.Tick()` every second, but abort if one is already running. Or, should we only attempt to launch a `.Tick()` every 10s.

bartekn: We should attempt to lunch it 1s after the previous was done (this is existing behaviour). We should make sure it's not running more than 10s (by sending a context with timeout). The code suggestion from the first comment implements this.

paulbellamy: > We should attempt to lunch it 1s after the previous was done (this is existing behaviour).

The existing behaviour is that it launches a new one every second (with no regard to the previous).


> We should make sure it's not running more than 10s (by sending a context with timeout). 

👍 


> The code suggestion from the first comment implements this.

Not quite, but sgtm. Will update to do this.

PR #3574, opened 2021-04-28
services/horizon: Add middleware for checking read-replica lag

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

* Adds a new `RO_DATABASE_URL` that allows setting a connection to a read-replica DB.
* When the flag is set:
  * A new middleware checking if the read-replica is behind a primary DB will be enabled. If such check fails three times, the `stale_history` error will be returned. The middleware will sleep 20ms and 40ms on the first and second failure.
  * All DB queries will be sent to read-replica (except the one that checks the primary sequence number in the middleware above).
* Adds a new `horizon_http_replica_lag_errors_count` metric to be able to track number of error responses due to replica lag.

### Why

Horizon should provide clients with strongly consistent responses when using read-replicas. Specifically, for consecutive HTTP requests: A and B, response for B must always return data at ledger sequence higher or equal to the response for A.

### Known limitations

In case of a higher replica lag, the middleware may increase HTTP error rate.

### DIFF
--- a/services/horizon/internal/httpx/middleware.go
+++ b/services/horizon/internal/httpx/middleware.go
@@ -354,3 +354,51 @@ func (m *StateMiddleware) WrapFunc(h http.HandlerFunc) http.HandlerFunc {
 func (m *StateMiddleware) Wrap(h http.Handler) http.Handler {
 	return m.WrapFunc(h.ServeHTTP)
 }
+
+type ReplicaSyncCheckMiddleware struct {
+	PrimaryHistoryQ *history.Q
+	ReplicaHistoryQ *history.Q
+	ServerMetrics   *ServerMetrics
+}
+
+// WrapFunc executes the middleware on a given HTTP handler function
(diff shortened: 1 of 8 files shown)


services/horizon/internal/httpx/middleware.go
```diff
@@ -354,3 +354,51 @@ func (m *StateMiddleware) WrapFunc(h http.HandlerFunc) http.HandlerFunc {
...
+			}
+
+			replicaIngestLedger, err := m.ReplicaHistoryQ.GetLastLedgerIngestNonBlocking()
+			if err != nil {
+				problem.Render(r.Context(), w, err)
+				return
+			}
+
+			if replicaIngestLedger >= primaryIngestLedger {
+				break
```
tamirms, line 381: would it make sense to have metrics on the distribution of requests which satisfy this condition on attempt 1, 2, 3, and 4?

tamirms (approved):

Issue #2761, opened 2020-06-29
Improve verify-range testing


2opremio:
While testing captive core with verify-range in AWS batch I encountered a few problems which made the experience unpleasant:

1. Manually editing the Job Definitions/Queues/Compute environments in the AWS console was hairy and error-prone. 
2. Spawning jobs suffers from (1) aggravated by the need to calculate the ledger ranges (we have a spreadsheet for that, but still)
3. When a job fails, there is no way to simply respawn jobs. You need to manually create a new job by copy-pasting the fields of the failed job (increasing the number of retries is often not enough).

To improve the situation, I propose the following:
1. Use [Terraform to keep track of the definiton AWS Batch resources](https://www.terraform.io/docs/providers/aws/r/batch_compute_environment.html) and apply them automatically on merge. This will probably require having a separate repo for verify-range.
2. Spawn test jobs semi-automatically through CI (specify the base branch, click a button and have CI spawn the jobs). For GitHub Actions we could use https://www.actionspanel.app/ (I don't know if there is an equivalent solution for CircleCI)
3. CI could monitor jobs for failures and respawn them when the job failure is transient. We could also have a manual CI action which clones a job given its identifier.

2opremio:
Some other things which we could do:
1. Automatically push the `verify-range` image when modifying it
2. Automatically test the `verify-range` image in CI with a small range

ire-and-curses:
More improvements: https://github.com/stellar/pipelines/pull/202

PR #3575, opened 2021-04-29
services/horizon: Build and test verify-range Docker image

2opremio:
Part of #2761

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -394,7 +394,7 @@ jobs:
           command: |
             docker build -f services/horizon/docker/verify-range/Dockerfile -t stellar/horizon-verify-range services/horizon/docker/verify-range/
             # Any range should do for basic testing, this range was chosen pretty early in history so that it only takes a few mins to run
-            docker run -ti -e BRANCH=05e09089828a0528e5545a04d7f713dee8f8e379 -e FROM=15000063 -e TO=15000127 stellar/horizon-verify-range
+            docker run -ti -e BRANCH=master -e FROM=15000063 -e TO=15000127 stellar/horizon-verify-range
 
   publish_latest_horizon_docker_image:
     docker:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -383,6 +383,18 @@ jobs:
...
+    docker:
+      - image: docker:18.04-git
+    steps:
+      - checkout
+      - setup_remote_docker
+      - run:
+          name: Build and test Verify Range Docker image
+          command: |
+            docker build -f services/horizon/docker/verify-range/Dockerfile -t stellar/horizon-verify-range services/horizon/docker/verify-range/
+            docker run -ti -e BRANCH=master -e FROM=15000063 -e TO=15000127 stellar/horizon-verify-range
```
2opremio, line 396: Maybe there is a better range to test, but I didn't want it to take more than a few minutes.

@bartekn let me know if you have other range suggestions

paulbellamy: If there's other ranges to test, we could do a few different ranges in parallel on circle?

paulbellamy: Also, it would be nice to comment in the circle file why we chose these particular ranges.

paulbellamy (approved):

.circleci/config.yml
bartekn: I think it's fine to have a single range here. Isn't it added here just to check if the image works? I have some ideas when it comes to ranges but the image is run using the code from `master` so it wouldn't really test the PR.

bartekn (approved):

2opremio:
CodeQL seems stuck, I will merge without it.

2opremio:
Uhm, I rebased and the verify-range test gets stuck, I suspect it could be due to #3576

I will remove the PR and see if it passes

2opremio:
I think it is due to #3576  (it works with `BRANCH=05e09089828a0528e5545a04d7f713dee8f8e379`). @paulbellamy could you take a look?

2opremio:
OK, it seems to _randomly_ get stuck (it just worked with master).

2opremio:
I am thinking it may be a memory problem, so I am going to move to an even earlier range.

PR #3576, opened 2021-04-29
services/horizon: Change ingest/ledgerbackend/LedgerBackend.GetLedger to be always blocking

paulbellamy:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

- Added `context.Context` as first param to most LedgerBackend methods, so that blocking ones can be aborted early due to shutdown or request timeout.
- Remove `ingest/ledgerbackend/LedgerBackend.GetLedgerBlocking`.
- Change `ingest/ledgerbackend/LedgerBackend.GetLedger` to always be blocking.
- Add 5s timeout (with `408 Request Timeout`) to `exp/services/captivecore/internal/server.go`'s `/ledger/{sequence}`, so that `ingest/ledgerbackend/RemoteCaptiveCore.GetLedger`, can retry and long-poll until the requested ledger is ready.

### Why

- We don't need both blocking and non-blocking, so it is simpler to not maintain it.
- Having it block, and passing a ctx to abort is (arguable) more idiomatic go.

### Known limitations

- I tried to keep the changes as non-invasive as possible, so there could probably be more cleanup done in the -CaptiveCoreBackend.
- It would be nice to push the single-threaded ness and some of the correctness guarantees down from the `exp/services/captivecore/internal/client.go` into the `ingest/ledgerbackend/CaptiveCoreBackend` itself, to ensure we always expose the same constraints.

### DIFF
--- a/exp/services/captivecore/internal/api.go
+++ b/exp/services/captivecore/internal/api.go
@@ -181,7 +181,6 @@ func (c *CaptiveCoreAPI) GetLedger(ctx context.Context, sequence uint32) (ledger
 	// TODO: We are always true here now, so this changes the semantics of this
 	// call a bit. We need to change the client to long-poll this endpoint.
 	return ledgerbackend.LedgerResponse{
-		Present: true,
-		Ledger:  ledgerbackend.Base64Ledger(ledger),
+		Ledger: ledgerbackend.Base64Ledger(ledger),
 	}, err
 }
--- a/exp/services/captivecore/internal/api.go
+++ b/exp/services/captivecore/internal/api.go
@@ -125,7 +125,7 @@ func (c *CaptiveCoreAPI) PrepareRange(ctx context.Context, ledgerRange ledgerbac
 		c.activeRequest.valid = true
 
 		c.wg.Add(1)
-		go c.startPrepareRange(ctx, ledgerRange)
+		go c.startPrepareRange(c.ctx, ledgerRange)
 
 		return ledgerbackend.PrepareRangeResponse{
 			LedgerRange:   ledgerRange,
--- a/exp/services/captivecore/internal/api.go
+++ b/exp/services/captivecore/internal/api.go
@@ -174,7 +174,7 @@ func (c *CaptiveCoreAPI) GetLedger(ctx context.Context, sequence uint32) (ledger
 		return ledgerbackend.LedgerResponse{}, ErrPrepareRangeNotReady
 	}
 
-	ledger, err := c.core.GetLedgerBlocking(ctx, sequence)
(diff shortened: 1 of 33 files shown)


ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -375,7 +363,7 @@ func (c *CaptiveStellarCore) startPreparingRange(ledgerRange Range) (bool, error
 	if ledgerRange.bounded {
 		err = c.openOfflineReplaySubprocess(ledgerRange.from, ledgerRange.to)
```
tamirms, line 364: do we need to pass in the context here?

exp/services/captivecore/internal/api.go
```diff
@@ -121,7 +125,7 @@ func (c *CaptiveCoreAPI) PrepareRange(ledgerRange ledgerbackend.Range) (ledgerba
 		c.activeRequest.valid = true
 
 		c.wg.Add(1)
-		go c.startPrepareRange(ledgerRange)
+		go c.startPrepareRange(ctx, ledgerRange)
```
tamirms, line 128: I'm not sure we want to use the request context here because this go routine should survive beyond the life time of this http request. perhaps we should be passing in c.ctx?

ingest/ledgerbackend/captive_core_backend.go
paulbellamy: no?

exp/services/captivecore/internal/api.go
```diff
@@ -170,12 +174,14 @@ func (c *CaptiveCoreAPI) GetLedger(sequence uint32) (ledgerbackend.LedgerRespons
...
-	present, ledger, err := c.core.GetLedger(sequence)
+	ledger, err := c.core.GetLedger(ctx, sequence)
 	if err != nil {
 		c.activeRequest.valid = false
 	}
+	// TODO: We are always true here now, so this changes the semantics of this
+	// call a bit. We need to change the client to long-poll this endpoint.
 	return ledgerbackend.LedgerResponse{
-		Present: present,
+		Present: true,
```
tamirms, line 184: should we remove the `Present` field entirely from the response?

tamirms (approved):

exp/services/captivecore/internal/api.go
paulbellamy: was debating. We could either:
- return Present: false in the body for the timeout
- remove it all together
- leave it as is, to change the api less (but given we're the only ones using it...)
wdyt?

tamirms: I'd be ok with leaving it as is or removing it all together, with a preference for removing it

PR #3583, opened 2021-05-04
all: remove codecov bash script

leighmcculloch:
### What

Remove the codecov.io bash script from CI runs.

### Why

As far as I am aware most folks who contribute to this repository are not using the codecov check regularly. Executing bash scripts that are curled from other environments without any checksum checks or versioning seems unwise since we have no knowledge of when it changes or how it changes. If we aren't using the check regularly, the value to risk doesn't make it worth having in my opinion.

### Known limitations

The why states that as far as I know folks who contribute to this repository are not using the codecov check regularly. I don't know for sure if this is the case. Is anyone using it and want it to stay? Please speak up here.


bartekn (approved):

jacekn (approved):
LGTM!

PR #3588, opened 2021-05-05
`services/regulated-assets-approval-server/{cmd, internal}`: Add `base-url` and `approval-server-payment-threshold` env/cli args to sep8 server

Zagan202:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

* [X] I've updated any docs ([developer docs](https://www.stellar.org/developers/reference/), `.md`

### Release planning

</details>

### What

Add env/CLI arguments to sep8-regulated-assets:
* base-url
* price-threshold

### Why

To prevent hardcoded values.

### Known limitations

N/A


services/regulated-assets-approval-server/README.md
```diff
@@ -46,6 +46,8 @@ Flags:
       --issuer-account-secret string   Secret key of the asset issuer's stellar account. (ISSUER_ACCOUNT_SECRET)
       --network-passphrase string      Network passphrase of the Stellar network transactions should be signed for (NETWORK_PASSPHRASE) (default "Test SDF Network ; September 2015")
       --port int                       Port to listen and serve on (PORT) (default 8000)
+      --base-url string                The base url address to this server
+      --approval-server-payment-threshold int The amount threshold to require KYC
```
leighmcculloch, line 50: 💡 All the configs are for the approval server, so we probably don't need `approval-server` in parameters.

Suggestion:
```suggestion
      --kyc-required-payment-amount-threshold int The amount threshold to require KYC
```

services/regulated-assets-approval-server/internal/serve/serve.go
```diff
@@ -62,6 +65,8 @@ func handleHTTP(opts Options) http.Handler {
 		assetCode:         opts.AssetCode,
 		issuerAddress:     issuerKP.Address(),
 		networkPassphrase: opts.NetworkPassphrase,
+		approvalServer:    buildURLString(opts.BaseURL, "/tx_approve"),
```
leighmcculloch, line 68: This is out-of-scope of this PR, so this doesn't need to block or impact this PR at all, but just noticing we use underscores on this endpoint, but hyphens in the well-known endpoint. Consistency would be beneficial I think and underscores are rarely seen in URLs. Could we use `tx-approve`? cc @marcelosalloum

services/regulated-assets-approval-server/internal/serve/serve.go
```diff
@@ -84,3 +89,13 @@ func (opts Options) horizonClient() horizonclient.ClientInterface {
...
 }
+
+func buildURLString(baseURL, endpoint string) string {
+	urlRegexp, err := regexp.Compile(`([^:])(//+)`)
+	if err != nil {
+		log.Fatal(errors.Wrap(err, "compiling regex"))
+	}
+
+	fullPath := fmt.Sprintf("%s/%s", baseURL, endpoint)
+	return urlRegexp.ReplaceAllString(fullPath, "$1/")
```
leighmcculloch, line 100: ❗ Could you use the `net/url` package to parse the incoming base URL, then append the appropriate path? You should be able to do that without using regex. The type you get back from parsing will let you append a path, then you can regenerate the final URL using the type. I think we should do this because it is safer, and means we aren't reinventing something that exists in stdlib, and one less thing for us to get right and test many variations.

services/regulated-assets-approval-server/README.md
leighmcculloch, line 50: 💡 I think we should also mention how is the amount denominated, so it is clear. The fact the type is `int` is helpful, but if I enter 100 I might be surprised to find out that is `0.0000100`. In fact maybe a string is better, and then you can use the `amount` package to unwrap it to an int.

leighmcculloch (commented):
One thing I think we should change (❗), and a couple suggestions (💡) that I defer to you folks on, otherwise looks great 👏🏻.

services/regulated-assets-approval-server/README.md
Zagan202: Good suggestion!

Zagan202: Thats an interesting point.
@marcelosalloum What do you think?

services/regulated-assets-approval-server/internal/serve/serve.go
Zagan202: @leighmcculloch How do you feel about this change? a6e437d
Also addresses this too https://github.com/stellar/go/pull/3588#discussion_r626955906

services/regulated-assets-approval-server/internal/serve/serve.go
```diff
@@ -84,3 +89,12 @@ func (opts Options) horizonClient() horizonclient.ClientInterface {
...
 }
+
+func buildURLString(baseURL, endpoint string) string {
+	fullPath := fmt.Sprintf("%s/%s", baseURL, endpoint)
+	URL, err := url.Parse(fullPath)
+	if err != nil {
+		log.Fatal(errors.Wrapf(err, "Unable to parse URL: %s", baseURL))
+	}
+	return URL.String()
+}
```
leighmcculloch, line 100: 💡 Nice. Couple small changes I'd suggest:

- Use the `path` package to build the path. It is intended for constructing URL paths. (Not to be confused with the `path/filepath` package which is not for URLs.) It'll make sure there's one and only-one slash if someone provides a baseURL that has or doesn't have a slash.

- Use the `Parse` function to parse the baseURL, so that if you error, you error based on the user input only. Then use the `url.URL` type, the type of your `URL` variable, to construct the final result.

Take a look at this example of doing these two things:

https://play.golang.org/p/iFMkJPChaM3

leighmcculloch (commented):
Small suggested change to the construction of the base URL 👇🏻 , otherwise looks great. 👏🏻

leighmcculloch (approved):
Small suggested change to the construction of the base URL ☝🏻 , otherwise looks great. 👏🏻

services/regulated-assets-approval-server/internal/serve/serve.go
marcelosalloum: I was in favor of `/tx_approve` because it's the same wording used in [SEP-8](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0008.md#example) but after taking another look at the SEP, I can see it only shows up in the example, so probably not something to be enforced... I agree `tx-approve` should be better.

services/regulated-assets-approval-server/README.md
marcelosalloum: >💡 I think we should also mention how is the amount denominated, so it is clear.

Yup, totally agree with that!

> In fact maybe a string is better (...)

The `amount` package seems like a good option, it's part of the monorepo and seems to be meant for this kind of usage. It's my first time coming across it as well but might be worth the try.

services/regulated-assets-approval-server/internal/serve/toml.go
```diff
@@ -7,19 +7,18 @@ import (
...
 	fmt.Fprintf(rw, "NETWORK_PASSPHRASE=%q\n", h.networkPassphrase)
 	fmt.Fprintf(rw, "[[CURRENCIES]]\n")
 	fmt.Fprintf(rw, "code=%q\n", h.assetCode)
 	fmt.Fprintf(rw, "issuer=%q\n", h.issuerAddress)
 	fmt.Fprintf(rw, "regulated=true\n")
-	fmt.Fprintf(rw, "approval_server=%q\n", approvalServer)
-	fmt.Fprintf(rw, "approval_criteria=%q\n", approvalCriteria)
-
+	fmt.Fprintf(rw, "approval_server=%q\n", h.approvalServer)
+	fmt.Fprintf(rw, `approval_criteria="The approval server currently only accepts payments and the transaction must have exactly one operation and it must be of type payment. If an account is registered in the internal list of unauthorized addresses, all payments to/from that account will be rejected. Also, if the payment amount exceeds %.2f %s it will need KYC approval."`, h.kycThreshold, h.assetCode)
```
marcelosalloum, line 23: 💡  I don't think this description is correct. The old description text seems more correct for the current state of the application.

marcelosalloum (approved):
LGTM! I just included two wording updates below ⬇️ .

services/regulated-assets-approval-server/internal/serve/serve.go
leighmcculloch: We could also change the SEP too, not for consistency with our code, but just because underscores are less common in APIs. It's not a big deal though.

services/regulated-assets-approval-server/internal/serve/serve.go
Zagan202: > * Use the `path` package to build the path. It is intended for constructing URL paths. (Not to be confused with the `path/filepath` package which is not for URLs.) It'll make sure there's one and only-one slash if someone provides a baseURL that has or doesn't have a slash.

oooo wow thats an awesome package thats for the tip!

> * Use the `Parse` function to parse the baseURL, so that if you error, you error based on the user input only. Then use the `url.URL` type, the type of your `URL` variable, to construct the final result.

Good point 😄

Zagan202: applied here 69c9373

leighmcculloch (approved):

services/regulated-assets-approval-server/cmd/serve.go
```diff
@@ -61,6 +61,21 @@ func (c *ServeCommand) Command() *cobra.Command {
...
+			Usage:     "The base url address to this server",
+			OptType:   types.String,
+			ConfigKey: &opts.BaseURL,
+			Required:  true,
+		},
+		{
+			Name:        "kyc-required-payment-amount-threshold",
+			Usage:       "The amount threshold to require KYC",
+			OptType:     types.Int,
+			ConfigKey:   &opts.KYCPaymentThreshold,
```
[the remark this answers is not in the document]
Zagan202: Would you consider it more idiomatic to rename it  `KYCRequiredPaymentAmountThreshold`?

services/regulated-assets-approval-server/internal/serve/toml.go
Zagan202: Agreed let me switch it back. 😅

Zagan202: resolved here 0da586e

services/regulated-assets-approval-server/cmd/serve.go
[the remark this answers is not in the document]
Zagan202: implemented here eb3d1fd

services/regulated-assets-approval-server/README.md
Zagan202: @marcelosalloum @leighmcculloch Thoughts on this before I squash and merge?

services/regulated-assets-approval-server/README.md
```diff
@@ -62,6 +62,8 @@ Flags:
       --issuer-account-secret string   Secret key of the asset issuer's stellar account. (ISSUER_ACCOUNT_SECRET)
       --network-passphrase string      Network passphrase of the Stellar network transactions should be signed for (NETWORK_PASSPHRASE) (default "Test SDF Network ; September 2015")
       --port int                       Port to listen and serve on (PORT) (default 8000)
+      --base-url string                The base url address to this server
+      --kyc-required-payment-amount-threshold string The amount threshold when KYC is required, may contain decimals (default 500 units)
```
marcelosalloum, line 66: 👀  FYI this text is copied from the output of the command `go run serve --help` as described in line 51 above.

You should be updating the env description in services/regulated-assets-approval-server/cmd/serve.go and then run `go run serve --help` to update this section of the readme. Updating the CLI tool is more important than updating the Readme because that's ultimately the source of truth while the Readme could be outdated.

services/regulated-assets-approval-server/cmd/serve.go
```diff
@@ -69,6 +69,21 @@ func (c *ServeCommand) Command() *cobra.Command {
...
+		{
+			Name:      "base-url",
+			Usage:     "The base url address to this server",
+			OptType:   types.String,
+			ConfigKey: &opts.BaseURL,
+			Required:  true,
+		},
+		{
+			Name:        "kyc-required-payment-amount-threshold",
+			Usage:       "The amount threshold to require KYC",
```
marcelosalloum, line 81: This is the place where you should be updating the description.

marcelosalloum (commented):
Found an inconsistency between the CLI tool and the Readme.md. Could you plz open a new PR fixing that (should be a quick fix)? Explanation follows below ⬇️

Issue #3593, opened 2021-05-10
services/horizon: Add metrics on DB query rate, errors, and duration


paulbellamy:
### What problem does your feature solve?

Helps give operators into how horizon's db connection health is.

### What would you like to see?

prometheus metrics for db query rate, errors, and duration.

PR #3597, opened 2021-05-12
services/horizon: Add prometheus metrics to database queries

paulbellamy:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add prometheus metrics for query count and duration to horizon database calls. Turns out the metrics we had before were actually missing the ingest and core database clients as well, so I've added the same metrics to those too.

The big code changes here are:
- Actually use `db.SessionInterface`, instead of just using `*db.Session` everywhere. Nice cleanup anyway.
- Have to pass the `StateMiddleware` and `HistoryMiddleware` into each route individually, as they have to happen after routing. Big OOF.

### Why

So we can verify and really understand which endpoints are causing db load and why.

Fixes https://github.com/stellar/go/issues/3593

### Known limitations

- [x] TODO: Needs a bit more testing, like firing it up and seeing it work.
- I'm not sure if the register/unregister calls are in the right place. wdyt?
- @jacekn Do these metrics and labels look right?

### DIFF
--- a/services/horizon/internal/db2/history/accounts.go
+++ b/services/horizon/internal/db2/history/accounts.go
@@ -8,6 +8,7 @@ import (
 	"github.com/lib/pq"
 
 	"github.com/stellar/go/services/horizon/internal/db2"
+	"github.com/stellar/go/support/db"
 	"github.com/stellar/go/support/errors"
 	"github.com/stellar/go/xdr"
 )
@@ -190,7 +191,7 @@ func (q *Q) UpsertAccounts(ctx context.Context, accounts []xdr.LedgerEntry) erro
 		num_sponsoring = excluded.num_sponsoring`
 
 	_, err := q.ExecRaw(
-		ctx,
+		context.WithValue(ctx, &db.QueryTypeContextKey, "upsert"),
 		sql,
 		pq.Array(accountID),
 		pq.Array(balance),
--- a/services/horizon/internal/db2/history/accounts.go
+++ b/services/horizon/internal/db2/history/accounts.go
@@ -191,7 +191,7 @@ func (q *Q) UpsertAccounts(ctx context.Context, accounts []xdr.LedgerEntry) erro
 		num_sponsoring = excluded.num_sponsoring`
 
 	_, err := q.ExecRaw(
-		context.WithValue(ctx, &db.QueryTypeContextKey, "upsert"),
+		context.WithValue(ctx, &db.QueryTypeContextKey, db.UpsertQueryType),
 		sql,
 		pq.Array(accountID),
 		pq.Array(balance),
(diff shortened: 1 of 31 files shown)


ire-and-curses:
Just want to confirm: are we getting raw numbers from these metrics? It would be great to be able to derive e.g. max and min as well as average for these calls.

jacekn:
> @jacekn Do these metrics and labels look right?

Yeah they look good to me.
We're splitting one of them by `route` that that one will generate 225 new metrics so in total it should be 270 metrics which if my math is right. That's not too bad, it's less than 25% increase and those should be really useful

jacekn:
> Just want to confirm: are we getting raw numbers from these metrics? It would be great to be able to derive e.g. max and min as well as average for these calls.

Averages we'll get out of the box - in prometheus we'll be able to divide total of all queries by number of queries to get average.
Min/max we won't get out of the box but it's doable too. We'll just need a small "recording rule" in prometheus (which is a common thing and very simple to set up). With the recording rule we'll be able to calculate min/max in any timerange we choose.

services/horizon/internal/db2/history/accounts.go
```diff
@@ -190,7 +191,7 @@ func (q *Q) UpsertAccounts(ctx context.Context, accounts []xdr.LedgerEntry) erro
 		num_sponsoring = excluded.num_sponsoring`
 
 	_, err := q.ExecRaw(
-		ctx,
+		context.WithValue(ctx, &db.QueryTypeContextKey, "upsert"),
```
2opremio, line 194: Can we get a symbol for the uses of `"upsert"`?

services/horizon/internal/httpx/middleware.go
```diff
@@ -141,16 +141,16 @@ func sanitizeMetricRoute(routePattern string) string {
 	route = strings.ReplaceAll(route, "\\", "\\\\")
 	route = strings.ReplaceAll(route, "\"", "\\\"")
 	route = strings.ReplaceAll(route, "\n", "\\n")
+	if route == "" {
+		// Can be empty when request did not reached the final route (ex. blocked by
```
2opremio, line 145:
```suggestion
		// Can be empty when request did not reach the final route (ex. blocked by
```

services/horizon/internal/init.go
```diff
@@ -44,22 +44,24 @@ func mustInitHorizonDB(app *App) {
 	}
 
 	app.historyQ = &history.Q{mustNewDBSession(
+		"history",
```
2opremio, line 47: can we get symbols for the subsystems?

2opremio (approved):

services/horizon/internal/init.go
```diff
@@ -18,15 +18,15 @@ import (
...
+func mustNewDBSession(subsystem db.Subsystem, databaseURL string, maxIdle, maxOpen int, registry *prometheus.Registry) db.SessionInterface {
 	session, err := db.Open("postgres", databaseURL)
 	if err != nil {
 		log.Fatalf("cannot open Horizon DB: %v", err)
 	}
 
 	session.DB.SetMaxIdleConns(maxIdle)
 	session.DB.SetMaxOpenConns(maxOpen)
-	return session
+	return db.RegisterMetrics(session, "horizon", subsystem, registry)
```
bartekn, line 29: I think it changes metrics' names. Can we revert it? It will break Grafana dashboard.

PR #3601, opened 2021-05-17
Horizon: re-enable core 17 for CI (reverts #3600)

2opremio:

### DIFF
--- a/services/horizon/docker/Dockerfile.dev
+++ b/services/horizon/docker/Dockerfile.dev
@@ -9,7 +9,7 @@ RUN go install github.com/stellar/go/exp/services/captivecore
 
 FROM ubuntu:20.04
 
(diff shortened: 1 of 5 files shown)


services/horizon/docker/Dockerfile.dev
```diff
@@ -9,7 +9,7 @@ RUN go install github.com/stellar/go/exp/services/captivecore
 
 FROM ubuntu:20.04
 
-ENV STELLAR_CORE_VERSION 15.0.0-40
+ENV STELLAR_CORE_VERSION 17.0.0-557.096f6a7.focal
```
paulbellamy, line 12: Should this be the same build as above? Might make it easier to grep for if we keep them in sync?

paulbellamy (approved):

PR #3607, opened 2021-05-19
`regulated-assets-approval-server/`: Copy migrations to image

Zagan202:
### What

Include `/app/internal/db/dbmigrate/migrations/` and their files to docker image

### Why

The init deployed container is failing because the migration .sql files are not included in the image.
We need to copy them from the codebase into the image.
[Slack Thread](https://stellarfoundation.slack.com/archives/C02U19A2A/p1621380918020400)

### DIFF
--- a/services/regulated-assets-approval-server/docker/Dockerfile
+++ b/services/regulated-assets-approval-server/docker/Dockerfile
@@ -9,6 +9,7 @@ FROM ubuntu:20.04
 
 RUN apt-get update && apt-get install -y --no-install-recommends ca-certificates
 COPY --from=build /bin/regulated-assets-approval-server /app/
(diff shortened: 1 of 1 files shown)


leighmcculloch (approved):
Approving to unblock you, but we should change to embedding the migrations into the resulting binary like how we store migrations in our other projects.

The approach of building the migrations into the docker images will work for docker deployments but not for other deployments. Our services are released as binaries usually, and the migrations won't be included in those releases when this service gets it's first release. Deployers will have to go to the repository, find the appropriate revision, and copy the migrations. The final directory the migrations need to be in is also not intuitive.

Note that once Go 1.17 releases we can shift to using the builtin go embed feature that'll make embedding files much easier. Until then unfortunately we need to use the gobin approach like we do in recoverysigner, horizon, etc.

Zagan202:
@leighmcculloch What you suggested sounds like a much better long term solution.
I made an issue for it [here](https://github.com/stellar/go/issues/3608).

PR #3603, opened 2021-05-17
services/horizon: Add Muxed Account details to operation responses

2opremio:
Part of #3591 

It follows the design discussed at #3591 

It abuses the operation details to encode `source_account_muxed` and `source_account_muxed_id`

TODO:
- [x] Add tests

Issue #3591, opened 2021-05-07
Horizon: support querying by Muxed Account and incorporate Muxed accounts in responses


2opremio:
Following the description of https://github.com/stellar/stellar-protocol/issues/617 we should add support for Muxed Accounts in:

1. All account-parameterized requests
2. Responses

With respect to requests ~we should support Muxed Accounts when:~ (edit: we decided to not do this for now)

- ~Querying transactions by account~
- ~Querying operations by account~
- ~Querying effects by account? (we may want to stick to G-addresses for effects)~

With respect to responses, we should support Muxed Accounts when:
- [x] Responding with Operation objects (source account and payment/merge/clawback destination)
- [ ] Responding with Transaction objects (source account)
- [ ] Responding with Effects?

Note: this description may not be complete yet, I will be incorporating new tasks while going through the API and DB.

2opremio:
The first thing we should clarify/design is how Horizon should respond based on whether it receives an M-acount or or a G-account as a query parameter. Depending on what we decide we could end up with a different DB schema.


There are a bunch of approaches which could be followed:

1. Be strict about G-addreses:
   * If the user provides a G-address (lets call it `GADDR` ), she strictly means that G-address explicitly. Responses should only contain data related to ` GADDR` . In particular, M-addresses with the same underlying `GADDR`  should be discarded.

2. Be strict about M-addresses: 
   * If the user provides an M-address (lets call it `MADDR` , with underlying G-address GADDR), she strictly means that M-address explicitly. Responses should only contain data related to that M-address. Other M-addresses with the same underlying ` GADDR`  as `MADDR` should be discarded. In addition, data related to `GADDR` (i.e. the `Ed25519` variant of `MADDR`, without an `Id`) should also be discarded.

3. Glob G-addresess:
   * If the user provides a G-address (lets call it ` GADDR` ), she means any `Ed2551` or `Med25519` account with the same `Ed25519` data as GADDR. In particular the response should include all the M-addresses with the same `Ed25519` value as `GADDR`, regardless of their ID. And, of course, `GADDR` should be included.

4. Glob M-addresses:
   * If the user provides an M-address (lets call it `MADDR` , with underlying G-address `GADDR`). she means any `Ed2551` or `Med25519` account with the same `Ed25519` data as `GADDR`. In particular the response should include all the M-addresses with the same `Ed25519` value as `GADDR`, regardless of their ID. And, of course, `GADDR` should be included.


[...]

PR #3610, opened 2021-05-19
services/horizon: Add Muxed Account details to effect responses 

2opremio:
Addresses part of #3591 

Depends on #3603 (whose commits are included here for now)

TODO:
- [x] Add more tests

### DIFF
--- a/protocols/horizon/effects/main.go
+++ b/protocols/horizon/effects/main.go
@@ -4,10 +4,9 @@ import (
 	"encoding/json"
 	"time"
 
-	"github.com/stellar/go/xdr"
-
 	"github.com/stellar/go/protocols/horizon/base"
 	"github.com/stellar/go/support/render/hal"
+	"github.com/stellar/go/xdr"
 )
 
 // Peter 30-04-2019: this is copied from the history package "github.com/stellar/go/services/horizon/internal/db2/history"
--- a/protocols/horizon/effects/main.go
+++ b/protocols/horizon/effects/main.go
@@ -2,8 +2,9 @@ package effects
 
 import (
 	"encoding/json"
-	"github.com/stellar/go/xdr"
 	"time"
+
+	"github.com/stellar/go/xdr"
 
 	"github.com/stellar/go/protocols/horizon/base"
 	"github.com/stellar/go/support/render/hal"
@@ -249,6 +250,8 @@ type Base struct {
 	ID              string    `json:"id"`
 	PT              string    `json:"paging_token"`
 	Account         string    `json:"account"`
+	AccountMuxed    string    `json:"account_muxed,omitempty"`
+	AccountMuxedID  uint64    `json:"account_muxed_id,omitempty"`
 	Type            string    `json:"type"`
 	TypeI           int32     `json:"type_i"`
 	LedgerCloseTime time.Time `json:"created_at"`
@@ -382,6 +385,8 @@ type TrustlineDeauthorized struct {
 type Trade struct {
 	Base
 	Seller            string `json:"seller"`
+	SellerMuxed       string `json:"seller_muxed,omitempty"`
+	SellerMuxedID     uint64 `json:"seller_muxed_id,omitempty"`
 	OfferID           int64  `json:"offer_id,string"`
 	SoldAmount        string `json:"sold_amount"`
 	SoldAssetType     string `json:"sold_asset_type"`
--- a/services/horizon/internal/ingest/processors/effects_processor.go
+++ b/services/horizon/internal/ingest/processors/effects_processor.go
@@ -234,14 +234,24 @@ type effectsWrapper struct {
 	operation *transactionOperationWrapper
 }
 
-func (e *effectsWrapper) add(address string, effectType history.EffectType, details map[string]interface{}) {
+func (e *effectsWrapper) add(address *xdr.AccountId, effectType history.EffectType, details map[string]interface{}) {
 	e.effects = append(e.effects, effect{
-		address:     address,
+		address:     address.Address(),
 		operationID: e.operation.ID(),
 		effectType:  effectType,
 		order:       uint32(len(e.effects) + 1),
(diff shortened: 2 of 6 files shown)


services/horizon/internal/ingest/processors/effects_processor.go
```diff
@@ -234,16 +234,26 @@ type effectsWrapper struct {
...
 
+func (e *effectsWrapper) addMuxed(address *xdr.MuxedAccount, effectType history.EffectType, details map[string]interface{}) {
+	if address.Type == xdr.CryptoKeyTypeKeyTypeMuxedEd25519 {
+		// We abuse the details to inject muxed-account information without changing the DB schema
+		details["account_muxed"] = address.Address()
+		details["account_muxed_id"] = uint64(address.Med25519.Id)
+	}
+	accID := address.ToAccountId()
+	e.add(&accID, effectType, details)
+}
```
Shaptic, line 255: Should we have an `func (m *MuxedAccount) GetId() uint64` method to simplify his code? It's a bit of a nuisance to have to think about how the muxed ID is accessed/encoded.

services/horizon/internal/ingest/processors/effects_processor.go
```diff
@@ -366,7 +376,11 @@ func (e *effectsWrapper) addLedgerEntrySponsorshipEffects(change ingest.Change)
 		return nil
 	}
 
-	var accountAddress string
+	var (
+		accountID    *xdr.AccountId
+		muxedAccount *xdr.MuxedAccount
+	)
+
```
Shaptic, line 383: Would it be better to use `xdr.MuxedAccount`'s `SetEd25519Address(accountId)` to unify the code paths to *all* use `MuxedAccount`, then switching on the `CryptoKeyType` at the lowest possible level if necessary? 

I was under the impression that this switch is the point of muxed accounts. Otherwise, I feel like we're going to do this "if this then muxed account otherwise account ID" dance in a *lot* of places with high-level code.

protocols/horizon/effects/main.go
```diff
@@ -2,9 +2,10 @@ package effects
 
 import (
 	"encoding/json"
-	"github.com/stellar/go/xdr"
 	"time"
 
+	"github.com/stellar/go/xdr"
+
```
Shaptic, line 8: weird nit but why is there a newline here haha

Shaptic (commented):
Added some thoughts about whether or not we can unify code paths to always use the `MuxedAccount` object.

services/horizon/internal/ingest/processors/effects_processor.go
2opremio: In this case I don't think it's so bad, since you need to check the Muxed Account type regardless

services/horizon/internal/ingest/processors/effects_processor.go
2opremio: Good point! I think I will do that

protocols/horizon/effects/main.go
2opremio: Uhm, I think you are right there shouldn't be a newline. My IDE (Goland) did this automatically.

2opremio: I will check why.

2opremio: Fixed

services/horizon/internal/ingest/processors/effects_processor.go
2opremio: I gave it a try and I think the resulting code is harder to understand

services/horizon/internal/ingest/processors/effects_processor.go
```diff
@@ -234,16 +234,26 @@ type effectsWrapper struct {
...
 		details:     details,
 	})
 }
 
+func (e *effectsWrapper) addMuxed(address *xdr.MuxedAccount, effectType history.EffectType, details map[string]interface{}) {
+	if address.Type == xdr.CryptoKeyTypeKeyTypeMuxedEd25519 {
+		// We abuse the details to inject muxed-account information without changing the DB schema
+		details["account_muxed"] = address.Address()
+		details["account_muxed_id"] = uint64(address.Med25519.Id)
+	}
```
bartekn, line 252: Sorry, this might have been discussed in previous PR but why not add a new column to the table? Adding a `DEFAULT NULL` column is fast as far as I can remember.

services/horizon/internal/ingest/processors/effects_processor.go
bartekn: I'm thinking that having separate `add` and `addMuxed` methods increase chances of a bug in the future. Say, we implement a new effect. With the current solution you need to remember if the field is normal or muxed account. Developer can also forget about muxed accounts and try to fetch it using `address.ToAccountId()`.

I think that alternatively we can introduce a new interface on shared methods of both `AccountId` and `MuxedAccount` (which are `MustEd25519` and `GetEd25519`). Then we only have `add` which checks if the passed object is normal or muxed account.

The problem with that solution is that 3 more structs that implement these methods but none of them is used in muxed account context.

bartekn (approved):
LGTM! Added two questions/ideas.

services/horizon/internal/ingest/processors/effects_processor.go
2opremio: I discarded that because I agreed with @ire-and-curses to minimize schema changes.

That said, if we agree to make a schema change, it would also solve the performance problem with #3619

ire-and-curses: ok, let's add the column. I don't like having to deserialise the whole envelope every time. Lesser of two evils.

2opremio: OK, I will merge this as is and then change the approach.

tamirms (approved):

PR #3614, opened 2021-05-20
keypair: rename Master to Root

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Add a `Root` function that returns a networks root account keypair derived from the network passphrase, and deprecate the `Master` function that does same.

### Why
The `Master` function is difficult to discover because the term 'master' is not normally what we call the keypair for the networks root account.

The term master is what we use to refer to the keypair of an account that the account ID is derived from, which is the default keypair that can sign for an account. Every account on a network has a master keypair.

In stellar-core and the quickstart image this keypair is referred to as the root account keypair. It would be easier to discover functionality in the Go SDK if it uses language that is consistent with stellar-core and other components.

The `Master` function returns a `KP` interface type. The intention for this function is to return a full keypair, a public and private key, so there is little benefit to the caller to obfuscate the `*Full` type from the caller. In fact I see in code that we routinely cast the `KP` to a `*Full` anyway, so we have a dependence on this always being a `*Full`. For this reason the `Root` function returns a `*Full`.

The `Master` function is exported, so rather than rename it, keeping it and marking it as deprecated preserves backwards compatibility.

### Known limitations

The keypair package doesn't have a CHANGELOG, so no CHANGELOG was updated.


marcelosalloum (approved):
LGTM!

bartekn (approved):

PR #3616, opened 2021-05-20
txnbuild: add SequenceNumber func to Transaction

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Add `SequenceNumber` function to `Transaction`.

### Why
Stellar transactions have a sequence number. If a transaction is valid and it is
accepted into a ledger the source account of the transaction is updated such
that its sequence number becomes the sequence number of the transaction.

In the Go SDK a parsed transaction has no sequence number field or function
directly on it. For a developer to get the sequence number they must look at the
source account object. This is not intuitive and not consistent conceptually
because a transaction has a sequence number and it should be a first class
attribute of the transaction.

In my own development using the Go SDK I spent time trying to figure out where
the sequence was stored and how to get at it. I work on Stellar everyday, so if
this is not intuitive to me, I suspect it isn't intuitive to others.

Other Stellar SDKs, such as the JS SDK, expose it as an attribute on a
Transaction.

This is likely to become even less intuitive as we plan to change the sequence
number rules where the sequence number on an account does not need to be
sequential. Regardless of this, it isn't intuitive today.

### Known limitations

N/A


acharb (approved):
lgtm!

txnbuild/transaction.go
```diff
@@ -212,6 +212,11 @@ func (t *Transaction) SourceAccount() SimpleAccount {
 	return t.sourceAccount
 }
 
+// SequenceNumber returns the sequence number of the transaction.
+func (t *Transaction) SequenceNumber() int64 {
+	return t.sourceAccount.Sequence
+}
+
```
marcelosalloum, line 219: ❓ Shouldn't we return `t.sourceAccount.Sequence + 1` when `t.IncrementSequenceNumber` is true?

marcelosalloum (commented):
Shouldn't the transaction sequence number depend on the `IncrementSequenceNumber` flag as well?

txnbuild/transaction.go
leighmcculloch: The `Transaction` type has no `IncrementSequenceNumber` field or function. I think you're thinking of `TransactionParams` which has the confusing and unexpected-side-effect-causing `IncrementSequenceNumber`.

marcelosalloum: Cool, thanks for explaining.

ire-and-curses (approved):
Great, this looks really helpful, especially for any non-trivial application (i.e. one that assumes validity, and tracks sequence numbers locally rather than looking up Horizon each time).

PR #3617, opened 2021-05-20
ingest: Prepares the package for a v2.0.0 release.

Shaptic:
This is the first new release of the `ingest` package, so this introduces a CHANGELOG that outlines everything since [v1.0.0](https://github.com/stellar/go/releases/tag/stellar-ingest-v1.0.0) (have a look at `git difftool stellar-ingest-v1.0.0 -- ingest/` for where this came from).


ingest/CHANGELOG.md
```diff
@@ -0,0 +1,30 @@
+# Changelog
+
+All notable changes to this project will be documented in this file. This project adheres to [Semantic Versioning](http://semver.org/).
+
+
+## Unreleased
+
+
+## v2.0.0
```
paulbellamy, line 9: If they're tightly coupled, should we just change the version to match horizon `v2.3.0`, then?

paulbellamy (approved):

ingest/CHANGELOG.md
Shaptic: Maybe "tightly coupled" isn't the right phrasing to use; I was more trying to say that "a lot of these changes came about with the Horizon v2.3 release." I'll amend the wording and merge this boi.

PR #3618, opened 2021-05-20
regulated-assets-approval-server/internal/serve/kyc-status/get_detail_handler.go: BUGFIX get request render json

Zagan202:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Properly render kyc-status GET response

### Why

kyc-status GET response was failing to render due to the use of `httpjson.RenderStatus` without populating the status code.
Thus resulting in a 502 error

### Known limitations
N/A


leighmcculloch (approved):

accordeiro:
@Zagan202 maybe I'm missing some context, but maybe we should add test coverage here to make sure there is no regression? Ideally we'd approach this fix by first writing a test that fails, then fixing the code.

Zagan202:
> @Zagan202 I might be missing some context, but maybe we should add test coverage here to make sure there is no regression? Ideally we'd approach this fix by first writing a test that fails, then fixing the code.

Apologies, the tests I wrote didn't detect this rendering error caused by `httpjson.RenderStatus`.
I was moving quickly because it was a simple fix.
I'll make a PR to test against this problem.

PR #3622, opened 2021-05-21
txnbuild: fix bug where base fee is not allowed to be zero

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Allow the base fee to be zero when building transactions.

### Why

The `txnbuild.NewTransaction` function errors when creating a transaction with a base fee of zero.

This is a bug.

Transactions with a zero base fee are accepted to the Stellar network when they are wrapped in a fee bump transaction envelope that pays the fee of the transaction. Applications use a zero base fee on the inner transaction to support fee sponsorship by an account that isn't the transaction source account.

This is required for payment channels, wallets like Vibrant, and other uses.

### Known limitations

N/A

### DIFF
--- a/txnbuild/CHANGELOG.md
+++ b/txnbuild/CHANGELOG.md
@@ -7,8 +7,7 @@ file.  This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Bug Fix
 
-* `BaseFee` in `TransactionParams` when calling `NewTransaction` is now allowed
-to be zero.
+* `BaseFee` in `TransactionParams` when calling `NewTransaction` is allowed to be zero because the fee can be paid by wrapping a `Transaction` in a `FeeBumpTransaction`. ([#3622](https://github.com/stellar/go/pull/3622))
 
 ## [v7.0.0](https://github.com/stellar/go/releases/tag/horizonclient-v7.0.0) - 2021-05-15
 
--- a/txnbuild/CHANGELOG.md
+++ b/txnbuild/CHANGELOG.md
@@ -5,6 +5,11 @@ file.  This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+### Bug Fix
+
+* `BaseFee` in `TransactionParams` when calling `NewTransaction` is now allowed
+to be zero.
+
 ## [v7.0.0](https://github.com/stellar/go/releases/tag/horizonclient-v7.0.0) - 2021-05-15
 
 ### Breaking changes
--- a/txnbuild/transaction_fee_test.go
+++ b/txnbuild/transaction_fee_test.go
@@ -4,56 +4,35 @@ import (
 	"math"
 	"testing"
 
+	"github.com/stellar/go/keypair"
 	"github.com/stretchr/testify/assert"
 )
 
-func TestBaseFeeCanBeZero(t *testing.T) {
-	kp0 := newKeypair0()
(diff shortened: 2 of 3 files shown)


Zagan202 (approved):
LGTM

howardtw (approved):
LGTM

txnbuild/transaction_fee_test.go
```diff
@@ -7,20 +7,54 @@ import (
...
+		TransactionParams{
+			SourceAccount: &sourceAccount,
+			Operations:    []Operation{&Inflation{}},
+			BaseFee:       MinBaseFee,
+			Timebounds:    NewInfiniteTimeout(),
+		},
+	)
+
+	assert.NoError(t, err)
+	assert.Equal(t, int64(MinBaseFee), tx.baseFee)
```
Shaptic, line 41: Any reason why these are separate tests rather than something like `func TestBaseFeeCanBeNonNegative()` with a loop testing a couple values?

txnbuild/CHANGELOG.md
```diff
@@ -5,6 +5,11 @@ file.  This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+### Bug Fix
+
+* `BaseFee` in `TransactionParams` when calling `NewTransaction` is now allowed
+to be zero.
```
Shaptic, line 11: I see we don't link to PRs in this changelog like we do in Horizon, but it'd be nice imo to have context for anyone wondering why there's no network minimum base fee.

```suggestion
* `BaseFee` in `TransactionParams` when calling `NewTransaction` is now allowed to be zero ([#3622](https://github.com/stellar/go/pull/3622)).
```

Shaptic (approved):
Nice :+1:

txnbuild/transaction_fee_test.go
leighmcculloch: Mostly because I wasn't optimizing for less lines of code, and these tests are small. I find tests are easier to maintain and evolve if we don't try and DRY them. If the tests had been simple input-output tests I probably would have written them as a table driven test with a for loop.

txnbuild/CHANGELOG.md
leighmcculloch: Good point. This changelog entry should explain why this change is being made too. Adding the link as you suggested along with a why:
```suggestion
* `BaseFee` in `TransactionParams` when calling `NewTransaction` is allowed to be zero because the fee can be paid by wrapping a `Transaction` in a `FeeBumpTransaction`. ([#3622](https://github.com/stellar/go/pull/3622))
```

txnbuild/transaction_fee_test.go
leighmcculloch: I rewrote the successful cases as you suggested, with the values looped over: 8e12127.

When I first read your comment I was thinking the alternative is to have all the successful and unsuccessful cases together, with ifs for the error cases, etc. It didn't occur to me until I re-read your message that you might be meaning to loop over the valid values, as I now realize you might have been. Thanks

Shaptic: Cool, agreed re: de-prioritizing LoC and DRY for tests and how it's a little weird to loop over just two happy-path values, but I appreciate you making the change! And walking me through your thought process, too.

Shaptic (approved):
:tada:

Issue #3580, opened 2021-05-04
Documentation for reingestion is incompatible with Horizon 2.2.0


dimaa6:
stellar-core 17.0.0
stellar-horizon version
2.2.0-7d7e58007af55e7697223009fd24daf76aa908e2
go1.16.2

Installed on Ubuntu 18.04 with apt.

I followed instructions on reingestion found here:

https://github.com/stellar/go/blob/master/services/horizon/internal/docs/captive_core.md#using-captive-core-to-reingest-the-full-public-network-history

My /etc/default/stellar-horizon is:

export DATABASE_URL=postgres://horizon:horizon@localhost:5433/horizon
export APPLY_MIGRATIONS=true
export HISTORY_ARCHIVE_URLS=https://s3-eu-west-1.amazonaws.com/history.stellar.org/prd/core-live/core_live_001
export NETWORK_PASSPHRASE="Public Global Stellar Network ; September 2015"
export STELLAR_CORE_BINARY_PATH=$(which stellar-core)
export ENABLE_CAPTIVE_CORE_INGESTION=true
export PARALLEL_JOB_SIZE=100
export RETRIES=10
export RETRY_BACKOFF_SECONDS=20

Database exists at provided URL.

Then I ran:

. /etc/default/stellar-horizon
stellar-horizon db reingest range --parallel-workers=24 17000000 17001000

I expected reingestion to start and to load 1001 ledgers as instructed.

Instead, Horizon produced the following output:

2021/05/03 12:49:43 successfully applied 47 horizon migrations
2021/05/03 12:49:43 Invalid config: one or more captive core params passed (--stellar-core-binary-path or --captive-core-config-append-path) but --ingest not set. If you are migrating from Horizon 1.x.y read the Migration Guide here: https://github.com/stellar/go/blob/master/services/horizon/internal/docs/captive_core.md

2opremio:
I think `--ingest` should be implicit when running the reingest command. It may actually have been the case and we could have messed it up.

PR #3625, opened 2021-05-24
services/horizon/cmd: Fix prepare range bug

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Fixes https://github.com/stellar/go/issues/3580

The `horizon db reingest range` command failed because it requires that the top level horizon ingest flag to be true.
In other words, `horizon --ingest db reingest range` works but `horizon db reingest range` responds with the following error:
```
2021/05/03 12:49:43 Invalid config: one or more captive core params passed (--stellar-core-binary-path or --captive-core-config-append-path) but --ingest not set. 
```

This PR fixes this bug by making all ingestion subcommands set ingest to true implicitly.

This PR also fixes the captive core config validation by only requiring the config file to be present when ingesting in online mode. It is not necessary to have a captive core config file when ingesting bounded ledger ranges.

### DIFF
--- a/services/horizon/CHANGELOG.md
+++ b/services/horizon/CHANGELOG.md
@@ -9,6 +9,7 @@ file. This project adheres to [Semantic Versioning](http://semver.org/).
 
 * Add more in-depth Prometheus metrics (count & duration) for db queries.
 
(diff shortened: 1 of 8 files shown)


paulbellamy (approved):
LGTM, but the code scanning warning does look like it might be a thing.

PR #3628, opened 2021-05-24
horizenclient: add IsTimeoutError helper

howardtw:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This PR adds a new helper function `IsTimeoutError` to determine if an error returned by the Horizon client is 504 or not.

### Why

Applications that use Horizon client will want to handle 504s. Some applications might want to retry the same transaction, some applications might want to return a special error so that its client can retry or handle it in another graceful way.

### Known limitations

N/A

### DIFF
--- a/clients/horizonclient/error_helpers.go
+++ b/clients/horizonclient/error_helpers.go
@@ -23,6 +23,26 @@ func IsNotFoundError(err error) bool {
 	return hErr.Problem.Type == "https://stellar.org/horizon-errors/not_found"
 }
 
+// IsTimeoutError returns true if the error is a horizonclient.Error with
+// a timeout problem indicating that Horizon timed out.
+func IsTimeoutError(err error) bool {
+	var hErr *Error
+
+	err = errors.Cause(err)
+	switch err := err.(type) {
+	case *Error:
+		hErr = err
+	case Error:
+		hErr = &err
+	}
+
+	if hErr == nil {
+		return false
+	}
+
+	return hErr.Problem.Type == "https://stellar.org/horizon-errors/timeout"
+}
+
 // GetError returns an error that can be interpreted as a horizon-specific
 // error. If err cannot be interpreted as a horizon-specific error, a nil error
 // is returned. The caller should still check whether err is nil.
--- a/clients/horizonclient/error_helpers_test.go
+++ b/clients/horizonclient/error_helpers_test.go
@@ -127,6 +127,125 @@ func TestIsNotFoundError(t *testing.T) {
 	}
 }
 
+func TestIsTimeoutError(t *testing.T) {
+	testCases := []struct {
+		desc string
+		err  error
+		is   bool
+	}{
+		{
+			desc: "nil error",
+			err:  nil,
+			is:   false,
+		},
+		{
+			desc: "another Go type of error",
+			err:  errors.New("error"),
+			is:   false,
+		},
+		{
+			desc: "timeout problem (pointer)",
+			err: &Error{
+				Problem: problem.P{
+					Type:   "https://stellar.org/horizon-errors/timeout",
+					Title:  "Timeout",
+					Status: 504,
+				},
+			},
+			is: true,
+		},
+		{
+			desc: "wrapped timeout problem (pointer)",
+			err: errors.Wrap(&Error{
+				Problem: problem.P{
+					Type:   "https://stellar.org/horizon-errors/timeout",
+					Title:  "Timeout",
+					Status: 504,
+				},
+			}, "wrap message"),
+			is: true,
+		},
+		{
+			desc: "timeout problem (not a pointer)",
+			err: Error{
+				Problem: problem.P{
+					Type:   "https://stellar.org/horizon-errors/timeout",
+					Title:  "Timeout",
+					Status: 504,
+				},
+			},
+			is: true,
+		},
+		{
+			desc: "wrapped timout problem (not a pointer)",
+			err: errors.Wrap(Error{
+				Problem: problem.P{
+					Type:   "https://stellar.org/horizon-errors/timeout",
+					Title:  "Timeout",
+					Status: 504,
+				},
+			}, "wrap message"),
+			is: true,
+		},
+		{
+			desc: "some other problem (pointer)",
+			err: &Error{
+				Problem: problem.P{
+					Type:   "https://stellar.org/horizon-errors/server_error",
+					Title:  "Server Error",
+					Status: 500,
+				},
+			},
+			is: false,
+		},
+		{
+			desc: "wrapped some other problem (pointer)",
+			err: errors.Wrap(&Error{
+				Problem: problem.P{
+					Type:   "https://stellar.org/horizon-errors/server_error",
+					Title:  "Server Error",
+					Status: 500,
+				},
+			}, "wrap message"),
+			is: false,
+		},
+		{
+			desc: "some other problem (not a pointer)",
+			err: Error{
+				Problem: problem.P{
+					Type:   "https://stellar.org/horizon-errors/server_error",
+					Title:  "Server Error",
+					Status: 500,
+				},
+			},
+			is: false,
+		},
+		{
+			desc: "wrapped some other problem (not a pointer)",
+			err: errors.Wrap(Error{
+				Problem: problem.P{
+					Type:   "https://stellar.org/horizon-errors/server_error",
+					Title:  "Server Error",
+					Status: 500,
+				},
+			}, "wrap message"),
+			is: false,
+		},
+		{
+			desc: "a nil *horizonclient.Error",
+			err:  (*Error)(nil),
+			is:   false,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.desc, func(t *testing.T) {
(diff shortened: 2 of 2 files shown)


morleyzhi (approved):

clients/horizonclient/CHANGELOG.md
```diff
@@ -3,6 +3,12 @@
 All notable changes to this project will be documented in this
 file.  This project adheres to [Semantic Versioning](http://semver.org/).
 
+## Unreleased
+
+### Add
+
+- Add `IsTimeoutError`
```
marcelosalloum, line 10: nit: 
```suggestion
- Add `IsTimeoutError`.
```

marcelosalloum (approved):
LGTM!

clients/horizonclient/CHANGELOG.md
howardtw: https://github.com/stellar/go/pull/3628/commits/783087e590b8523e5fe8af63868e74da78b00f73

clients/horizonclient/error_helpers.go
```diff
@@ -23,6 +23,26 @@ func IsNotFoundError(err error) bool {
...
+		hErr = err
+	case Error:
+		hErr = &err
+	}
+
+	if hErr == nil {
+		return false
+	}
+
+	return hErr.Problem.Type == "https://stellar.org/horizon-errors/timeout"
```
leighmcculloch, line 43: >Applications that use Horizon client will want to handle 504s.

I don't think this is sufficient for detecting all 504s. You could get other errors when a 504 occurs. You will also likely get other Go errors (non-horizonclient errors) when a timeout occurs.

The PR description says the goal is to handle 504 errors, but this code will only capture one specific case where 504 gets returned from Horizon in the problem format, and not from other components such as ELBs, etc, since this code doesn't check for the 504 status code, it relies on a fully formed Horizon response.

This might be fine for the scope of Vibrant, which I assume is what this is for, but maybe not all applications, so the scope might be too narrow for general use.

>Some applications might want to retry the same transaction

Applications probably need to consider a wider range of error conditions to know if they need to retry. Basically any error that isn't a transaction failed or bad request would be possibly retry-able.

leighmcculloch (commented):
I think error helper functions are helpful. We may also want to explore error.Is/As compatibility too, which would make it possible for people to write their own inspectors/matches without adding a function for each case to the lib.

The code doesn't entirely align with the description. I'm not sure if that is intentional or by mistake. See inline for more on that. I defer to @stellar/horizon-committers though.

clients/horizonclient/error_helpers.go
```diff
@@ -23,6 +23,26 @@ func IsNotFoundError(err error) bool {
...
+func IsTimeoutError(err error) bool {
+	var hErr *Error
+
+	err = errors.Cause(err)
+	switch err := err.(type) {
+	case *Error:
+		hErr = err
+	case Error:
+		hErr = &err
+	}
```
Shaptic, line 37: We should leverage `GetError()`, below, here:

```suggestion
	hErr := GetError(err)
```

I know the other helper does it this way, but I'd take issue with that, too.

clients/horizonclient/error_helpers.go
Shaptic, line 43: I agree with @leighmcculloch here that this is a _very_ specific error being detected. A more-accurate name is `MightBeTimeoutError()`, haha. The big difference between 404s and 504s is that the former is guaranteed (afaik) to be well-formed since it always comes from the API, whereas the latter can come from many places (and, in fact, so can a transient connection error or other outage). I would be okay with this change with a rename, like possibly `IsHorizonAPITimeoutError`, or some other language that lets the helper be, well, helpful, but also clarifies that this isn't a catch-all for timeout scenarios. It almost seems like it belongs in the client, honestly.

Aside: I'm *really* against hard-coding the error string here; why not check `hErr.Problem.Status == 504`? Again, I know the other helper does it this way, but I take issue with that, too.

clients/horizonclient/CHANGELOG.md
```diff
@@ -3,6 +3,12 @@
 All notable changes to this project will be documented in this
 file.  This project adheres to [Semantic Versioning](http://semver.org/).
 
+## Unreleased
+
+### Add
+
+- Add `IsTimeoutError`.
```
Shaptic, line 10: I'd love a little more description here, or at least a cross-reference to this PR for context. If you're an SDK consumer and you see this line, it's hard to learn more about the new helper quickly.

Shaptic (commented):
I like the idea of this change, but I think it needs more thought, especially w/ Leigh's comments in mind. We don't want to mislead SDK consumers into thinking this is a catch-all for timeouts. More thoughts inline!

clients/horizonclient/error_helpers.go
leighmcculloch: >It almost seems like it belongs in the client, honestly.

I think keeping this in the calling code might be the best path too. There is definitely room for improvement in the horizonclient for helping a client know if they should retry, or check the status of a tx, but that's a larger issue that needs a holistic plan.

Shaptic: Completely agreed: I would love to see a function that does this in the codebase (perhaps as a larger enhancement helping devs with tx unhappy-path flows); I'm just not sure we can throw this in without deeper design thoughts, and this is short enough that the caller can implement it for their specific needs.

howardtw: Good points all around.

I think renaming the function can solve the issue of it being confusing as there could be other reasons that lead to the timeout as you guys pointed out.

I don't feel strongly about merging it since Vibrant is already doing it. I opened this PR simply because I thought it would be a nice addition before we come up with a holistic plan.

clients/horizonclient/error_helpers.go
howardtw: https://github.com/stellar/go/pull/3628/commits/efd88f403ac05dd75f25adf41948d7532194218b

clients/horizonclient/error_helpers.go
howardtw: > I'm really against hard-coding the error string here; why not check hErr.Problem.Status == 504?

https://github.com/stellar/go/pull/3628/commits/0f547a79e1710d3972cc970df3aac4087dd3ac60

> I would be okay with this change with a rename, like possibly IsHorizonAPITimeoutError

https://github.com/stellar/go/pull/3628/commits/a2cdb3f037f1f78556e48ea4582c316164865ddd

clients/horizonclient/CHANGELOG.md
howardtw: https://github.com/stellar/go/pull/3628/commits/ee8fc706b8898558e6142ed68443d0099b6f04d7

howardtw:
@Shaptic @leighmcculloch PTAL again. If you still prefer a more holistic approach, I can close this PR.

clients/horizonclient/error_helpers.go
```diff
@@ -1,26 +1,32 @@
...
+	hErr := GetError(err)
+	if hErr == nil {
+		return false
 	}
 
+	return hErr.Problem.Type == "https://stellar.org/horizon-errors/not_found"
+}
+
+// IsHorizonAPITimeoutError returns true if the error is a horizonclient.Error with
+// a timeout problem indicating that Horizon timed out.
```
Shaptic, line 22: Can we add some more commentary here outlining the things we've discussed in this PR? I don't want anyone to rely on this in ways they shouldn't.

```suggestion
// IsHorizonAPITimeoutError returns true if the error is a horizonclient.Error 
// with a timeout Problem. This indicates that Horizon timed out in a 
// well-defined, intentional way:
//     https://stellar.org/horizon-errors/not_found
// but is not a catch-all method for detecting other timeout errors.
```

Shaptic (approved):
Cool, LGTM! Thanks for discussing this at length with us @howardtw.

clients/horizonclient/error_helpers.go
```diff
@@ -1,26 +1,32 @@
...
+func IsHorizonAPITimeoutError(err error) bool {
+	hErr := GetError(err)
 	if hErr == nil {
 		return false
 	}
 
-	return hErr.Problem.Type == "https://stellar.org/horizon-errors/not_found"
+	return hErr.Problem.Status == http.StatusGatewayTimeout
```
leighmcculloch, line 29: I think it is a little premature to add this to the horizonclient API given there's some consensus more thought needs to go into solving the problem holistically. It feels overly specific to one use case, and the function is very small, so not abstracting much for the user. The function also sets some precedence that isn't clear, should we add other error functions like this, should they be prefixed with `IsHorizonAPI`. If we follow this pattern forward, will the package API be cluttered and difficult to understand.

However, I defer to @Shaptic and @stellar/horizon-committers.

2opremio: @Shaptic thoughts?

clients/horizonclient/error_helpers.go
paulbellamy: Yeah, this would fail on testnet, or custom horizon sintallations.

PR #3636, opened 2021-05-26
txnbuild: Injects the source account as the issuer in AllowTrust.FromXDR

Shaptic:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
When decoding an `AllowTrustOp` from raw XDR, the resulting `Asset.Issuer` field will be blank, because its not included in the XDR directly. However, the issuer will [always](https://developers.stellar.org/docs/start/list-of-operations/#allow-trust) be the operation's source account, so we can backfill intelligently:

> **Allow Trust**
> This can only be called by the issuer of a trustline’s asset

### Why
We should decode XDR to be as close as possible to their original pre-encoded structures.

### Known limitations
If the operation is created without an issuer, decoding it will inject the issuer, meaning the higher-level operations (i.e. the `txnbuild.AllowTrust` structure) aren't identical. This is unavoidable: we either **never** have an issuer (there isn't one in the XDR) or **usually** have an issuer (pull it from the source account, if possible). Given that custom assets should always have issuers, the latter is preferred.

Also, since an operation's source account is optional, this will only _sometimes_ back-fill.


txnbuild/CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ file.  This project adheres to [Semantic Versioning](http://semver.org/).
 
 * `BaseFee` in `TransactionParams` when calling `NewTransaction` is allowed to be zero because the fee can be paid by wrapping a `Transaction` in a `FeeBumpTransaction`. ([#3622](https://github.com/stellar/go/pull/3622))
 
+* `AllowTrustOp` now fills in the asset issuer from the source account when parsing from the XDR ([#3635](https://github.com/stellar/go/pull/3635)).
```
Shaptic, line 16: TODO: Update me to be #3636.

txnbuild/allow_trust.go
```diff
@@ -94,6 +94,7 @@ func (at *AllowTrust) FromXDR(xdrOp xdr.Operation, withMuxedAccounts bool) error
 	if err != nil {
 		return errors.Wrap(err, "error parsing allow_trust operation from xdr")
 	}
+	t.Issuer = at.SourceAccount // asset issuer is *always* source account
```
leighmcculloch, line 97: It's surprising that no tests change for this. Can you add tests that fail without this change?

leighmcculloch, line 97: What happens if you create the operation without the issuer, then serialize, then deserialize, are the operations still identical?

leighmcculloch, line 97: Does the same change need making to `SetTrustLineFlagsOp`? Its asset field works the same way I believe, not storing the issuer because the issuer is the source account.

Shaptic: No, the asset field in `SetTrustLineFlags` is a proper `Asset`, rather than an `AssetCode`, meaning it [encodes the issuer information](https://github.com/stellar/go/blob/6a326806bb7142122489736643d0840298d7773d/txnbuild/asset.go#L103) within itself.

Shaptic: Good catch, no they aren't. However, I don't think this is problematic: assets **always** have issuers, so I would consider creating one _without_ an issuer to be UB. Up for discussion, of course.

Shaptic: Heh, the tests were using a `CreditAsset{"CODE", ""}`, aka no issuer, so they passed just fine. Fixed in latest push.

leighmcculloch: Maybe we should make the building the xdr fail if the issuer inside the asset != the source account. It might catch situations where someone is trying to allow trust from the wrong source account. If we added that validation, then this situation would never arise.

leighmcculloch: Such a change should be safe because it would only break applications that aren't sensible.

txnbuild/allow_trust.go
```diff
@@ -94,6 +94,11 @@ func (at *AllowTrust) FromXDR(xdrOp xdr.Operation, withMuxedAccounts bool) error
 	if err != nil {
 		return errors.Wrap(err, "error parsing allow_trust operation from xdr")
 	}
+	// The asset issuer is *always* the source account, but muxed accounts are a
+	// high-level representation, so we ensure here that we use the underlying
+	// address that would represent the "true" issuer on the Stellar network.
```
leighmcculloch, line 99: Good catch to handle the muxed source account!

txnbuild/allow_trust_test.go
```diff
@@ -71,8 +71,46 @@ func TestAllowTrustRoundtrip(t *testing.T) {
...
+		Type:          CreditAsset{"USD", "GA7QYNF7SOWQ3GLR2BGMZEHXAVIRZA4KVWLTJJFC7MGXUA74P7UJVSGZ"},
 		Authorize:     true,
 	}
+
 	testOperationsMarshallingRoundtrip(t, []Operation{&allowTrust}, true)
 }
+
+// Ensures that the issuer is back-filled from the source account correctly.
+// See https://github.com/stellar/go/pull/3636 for more context.
+func TestDecoderIncludesIssuer(t *testing.T) {
```
leighmcculloch, line 83: Would be helpful to include the scope of the test, the type it tests, in the test name.
```suggestion
func TestAllowTrustDecoderIncludesIssuer(t *testing.T) {
```

leighmcculloch (approved):
Looks great. Great fix.

Also, nice catching the muxed address case.

txnbuild/allow_trust_test.go
Shaptic: Isn't this implied by being in allow_trust_test? :smile:

txnbuild/allow_trust.go
Shaptic: Gotta give the existing AllowTrust tests credit for that one!

txnbuild/allow_trust.go
Shaptic: Great call, but this of course reveals a new can of worms: the existence of a `TestAllowTrustNoIssuer`... So enabling this was actually _intended_ behavior at some point. (`git blame` reveals [poliha](https://github.com/poliha), added June 2019).

I'm at a loss now about what to do now :joy: maybe enforcing them to match and/or forcing the issuer to exist is not the right call?

Shaptic:
New snafu found in 8582fe52: since an operation's source account is optional, this will only _sometimes_ back-fill.

txnbuild/allow_trust_test.go
leighmcculloch: Functions aren't scoped by their files, so when you run the tests for the package the function name is all you got. What if two operations in the package need a TestDecoderIncludesIssuer test?

leighmcculloch:
> an operation's source account is optional, this will only _sometimes_ back-fill.

Oh, good point. In those cases it comes from the transaction source account, but we don't have that in this scope.

What a conundrum. 

What if we remove the issuer from the AllowTrustOp?

I guess ignore the problem is also an option, if benefit to fixing it is less than the effort and consequences.

This has been a helpful learning experience for me regardless. We frequently make decisions at the protocol layer on the assumption a higher up layer can abstract the inconvenience, but abstractions are leaky.

txnbuild/allow_trust.go
ire-and-curses: I think we did it like this back in the day because we were trying to implement something that matches the underlying operation and we understood that this issuer could be empty because the XDR doesn't care.

This doesn't mean it's the best choice for users though. I like @leighmcculloch's idea of doing the sanity check during build (and removing that test if necessary).

leighmcculloch: I'm not sure that's the best path forward now because of https://github.com/stellar/go/pull/3636#issuecomment-849195313.

Shaptic:
Eric and I discussed this offline, and ultimately decided that this should be put on ice given the design & implementation challenges / time commitments appear to outweigh the benefits. This clearly needs more thought, more dev time (especially to work through test cases that appear to depend on this arguably-broken but inarguably-confusing behavior), and perhaps even breaking changes to the SDK.

-------

Just to summarize our collaborative investigations for posterity,

### The original problem
If you encode an `AllowTrust` operation with a valid `CreditAsset` (i.e. the `Issuer` field set) into XDR, then decode that XDR back into an `AllowTrust` object, the asset's issuer will be empty.

**Why?** Well, only the issuer of the asset can create a valid `AllowTrustOp` in the first place, so this redundant information isn't stored in the XDR.

However, this is problematic and unintuitive, because SDK consumers expect consistency between a pre-encoded structure (i.e. a structure they crafted and sent on the wire) and the post-decoded structure (i.e. one they receive from the wire). In this case, the SDK is over-promising and under-delivering in its higher-level abstraction: `xdr.AllowTrustOp.Asset` (of type `xdr.AssetCode`) can only partially fill `txnbuild.AllowTrust.Type` (of type `txnbuild.CreditAsset`).

### The original solution
The solution seemed straightforward: backfill `AllowTrust.Asset.Issuer` from the source account when parsing. 

### A can of bugs
Even with this best-effort attempt, there are some non-trivial considerations:
  
  - We have to be careful about muxed accounts.
  - The operation's source account is optional, so it only sometimes work. If the goal is to make the pre-encoding and post-decoding structures match, we still fall short.

  - Certain test cases (e.g. `TestAllowTrustNoIssuer`) seemed to have *expected* this behavior.
  - Other projects in the monorepo (e.g. `regulated-assets-approval-server`, with [this test](https://github.com/stellar/go/blob/dae1442926d3dfa3d0fcbf27e3893cd9c363168e/services/regulated-assets-approval-server/internal/serve/api_tx_approve_test.go#L262-L289)) also broke with this change.

### Possible resolutions

  1. We can continue with the "best-effort backfill," only working works when the operation's source account is set. This will require non-trivial refactoring to test cases across the monorepo.

```go
func (at *AllowTrust) FromXDR(xdrOp xdr.Operation, withMuxedAccounts bool) error {
	// [...]
	t.Issuer = accountFromXDR(xdrOp.SourceAccount, false)
	// [...]
}
```

  2. We can create a new type that conforms closer to the XDR (like `CreditAssetCode`) and doesn't include an issuer. This would literally only be used for the now-deprecated `AllowTrust` operation.

  3. We can add more sanity-checking when building the `AllowTrust` operation, for example ensuring that if the source account is set, it matches the asset issuer, and that the asset issuer must be set, since a `CreditAsset` object without an issuer is somewhat nonsensical. This doesn't have many tangible benefits, though.

  4. We can call this "working as intended," or say "prefer `SetTrustLineFlags` instead of this deprecated op," and just move on.

-------

Regardless, this is no longer trivial and should be put on hold until time, bandwidth, desire, or need permits.

Issue #632, opened 2018-09-05
Optimize trade aggregations query


bartekn:

bartekn:
Yes, we do log slow queries. Here's an example:
```
2019-01-14 13:01:32 UTC:172.17.20.113(55478):stellar@horizon:[79213]:LOG: duration: 1353.979 ms execute <unnamed>: SELECT timestamp, count(*) as count, sum(base_amount) as base_volume, sum(counter_amount) as counter_volume, sum(counter_amount)/sum(base_amount) as avg, max_price(price) as high, min_price(price) as low, first(price) as open, last(price) as close FROM (SELECT div((cast((extract(epoch from ledger_closed_at) * 1000 ) as bigint) - 0), 900000)*900000 + 0 as timestamp, history_operation_id, "order", counter_asset_id as base_asset_id, counter_amount as base_amount, base_asset_id as counter_asset_id, base_amount as counter_amount, ARRAY[price_d, price_n] as price FROM history_trades WHERE counter_asset_id = $1 AND base_asset_id = $2 AND ledger_closed_at >= $3 AND ledger_closed_at < $4 ORDER BY history_operation_id , "order") AS htrd GROUP BY timestamp ORDER BY timestamp desc LIMIT 200
2019-01-14 13:01:32 UTC:172.17.20.113(55478):stellar@horizon:[79213]:DETAIL: parameters: $1 = '7', $2 = '2', $3 = '1970-01-18 13:00:00', $4 = '2018-11-16 08:30:00'
```
Explain analyze:
```
[...]

PR #3641, opened 2021-05-28
services/horizon: Precompute 1m Trade Aggregation buckets

paulbellamy:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Incremental pre-computation for the 1-minute trade-aggregation buckets with a database-trigger. Compute higher-resolution queries on-the-fly.

### Why

Fixes https://github.com/stellar/go/issues/632

Per discussion pre-computing seems the best option for performance and scalability, as the amount of computation is known and bounded. The concern is disk usage. Given a raw `history_trades` table of 4gb, a naive 1m bucket pre-computation results in a new table approx 1.6GB. Not outrageous, but could grow quickly if more trading pairs become active. When generating that as a materialized view, I accidentally forgot to include the `base_asset_id`, and `counter_asset_id`, and noticed the table was much smaller. To compress the table this PR uses wide jsonb columns, so there is one row per year/month/asset-pair, with a `values` column containing the month's 1m buckets. This adds complexity and a small overhead on querying (~15%), but saves a further ~70% on disk size.

In order to handle the `offset` param the maximum resolution we could potentially pre-compute is 15 minutes. For this PR, I've only pre-computed the 1 minute buckets, and aggregated higher resolutions on the fly. As higher resolutions are quite quick to aggregate on-the-fly. In my past experience this has been enough, as the 1m aggregation gives higher resolutions a big boost. (TODO: benchmark this for concrete numbers)

Initially I hoped we could just use a materialized view to do the pre-computation (to avoid changes to horizon ingestion code). However, updating materialized views is all-or-nothing, and takes ~7 minutes on my laptop. Instead, this PR uses a trigger to incrementally update the relevant buckets on insert. This implementation is still quite fragile, in that it doesn't handle `UPDATE` or `DELETE` queries, and assumes all trades are inserted in-order. Non-incremental recompute of a 1m bucket takes ~300ms. I didn't want to add that overhead to each insert, but it might be an easy way to handle update/delete queries.

### Known limitations

This trigger is still quite fragile, in that it doesn't handle `UPDATE` or `DELETE` queries, and assumes all trades are inserted in-order.

The reduced query load on `history_trades` means we could probably get rid of some of the now-unused indexes on that table, which would save a few gigabytes of disk.

## TODO

- [x] fix high/low calculation
- [x] fix open/close calculation
- [x] benchmark on-the-fly higher-resolution aggregations
- [x] try benchmarks without jsonb as well to see if disk usage tradeoff is worth it or not.
- [x] handle offsets correctly in the new sql queries
- [x] handle reingestion (delete buckets, insert, rebuild start+end buckets)
- [x] try rebuilding buckets from go (instead of db triggers)
  - [x] test this actually works
  - [x] check this is actually fast enough to not slow down ingestion
- [x] testing the complicated sql aggregation queries (or pre-compute 5 and 15 minute intervals)
- [x] Check it works with postgres 9.6
- [x] Figure out if we need to create the index concurrently
- [x] benchmark with big data
  - [x] 1m queries
  - [ ] 1d queries
  - [ ] ingestion
  - [x] reingestion
  
 ## Later
 
- [ ] Figure out if we can drop any old indexes on the history_trades.


paulbellamy:
For reading this, I'd recommend starting with the migration, as it'll help make sense of the rest.

bartekn (approved):
It's looking good. Looking forward to some response durations stats. My only worry is reingestion in ranges. In such case ranges can be ingested in non-increasing order so the triggers here will calculate values wrong.

I think we can fix this but it requires adding an extra step once reingestion is done:
* We move the trigger to Go code and run it after a ledger is ingested using normal operations (not reingestion).
* In reingestion code we add an extra step that runs once all ranges are completed which is responsible for calculating 1m resolutions.

bartekn:
Idea to fix `open` and `close` when reingesting out of order. We can add two new columns `open_ledger_seq` and `close_ledger_seq` that will indicate the sequence number of the previously set open/close price. If a new ledger in a given bucket appears and it's before `open_ledger_seq` then we update the `open` price, otherwise `open` is not updated. We update `close` price in a similar way.

paulbellamy:
> In reingestion code we add an extra step that runs once all ranges are completed which is responsible for calculating 1m resolutions.

This seems like a good plan. We actually only need to completely-recalculate the first and last buckets, as those are the only ones which would be only partially-ingested.

paulbellamy:
Ok, so benchmarking on my local laptop. Let's find the biggest most-traded pairs, to set a worst-case scenario bound...
```
postgres=# select base_asset_id, counter_asset_id, count(*) from history_trades group by base_asset_id, counter_asset_id order by count desc limit 10;
 base_asset_id | counter_asset_id |  count
---------------+------------------+---------
             2 |              149 | 2819451
             2 |         17581803 | 1660065
             2 |         27869101 | 1424528
             2 |             2351 |  761635
             2 |             1519 |  616861
             2 |            40569 |  604979
             2 |            28007 |  590901
      20772946 |         20772947 |  562536
             2 |         46998801 |  555386
             2 |          1649457 |  449990
```

Pre-computed 1m buckets query:
```
postgres=# EXPLAIN ANALYZE SELECT
  (j.value->>'timestamp')::bigint as timestamp,
  (j.value->>'count')::integer as count,
  (j.value->>'base_volume')::bigint as base_volume,
  (j.value->>'counter_volume')::bigint as counter_volume,
  ARRAY[(j.value->>'high_n')::bigint, (j.value->>'high_d')::bigint] as high,
  ARRAY[(j.value->>'low_n')::bigint, (j.value->>'low_d')::bigint] as low,
  ARRAY[(j.value->>'open_n')::bigint, (j.value->>'open_d')::bigint] as open,
  ARRAY[(j.value->>'close_n')::bigint, (j.value->>'close_d')::bigint] as close
FROM
  history_trades_60000 h, jsonb_each(h.values) j
WHERE h.base_asset_id = '2'
  AND h.counter_asset_id = '149'
ORDER BY timestamp DESC LIMIT 200;


                                                                                                  QUERY PLAN
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=957.86..958.36 rows=200 width=156) (actual time=910.185..910.196 rows=200 loops=1)
   ->  Sort  (cost=957.86..958.61 rows=300 width=156) (actual time=910.183..910.188 rows=200 loops=1)
         Sort Key: (((j.value ->> 'timestamp'::text))::bigint) DESC
         Sort Method: top-N heapsort  Memory: 131kB
         ->  Nested Loop  (cost=0.29..945.51 rows=300 width=156) (actual time=0.356..848.402 rows=371658 loops=1)
               ->  Index Scan using history_trades_60000_year_month_base_asset_id_counter_asset_key on history_trades_60000 h  (cost=0.29..912.51 rows=3 width=618) (actual time=0.099..0.607 rows=39 loops=1)
                     Index Cond: ((base_asset_id = '2'::bigint) AND (counter_asset_id = '149'::bigint))
               ->  Function Scan on jsonb_each j  (cost=0.00..1.00 rows=100 width=32) (actual time=5.995..7.002 rows=9530 loops=39)
 Planning Time: 0.192 ms
 Execution Time: 910.512 ms
(10 rows)

Time: 911.161 ms
```

Daily buckets query:
```
postgres=# EXPLAIN ANALYZE SELECT
  bucket as timestamp,
  sum("count") as count,
  sum(base_volume) as base_volume,
  sum(counter_volume) as counter_volume,
  sum(counter_volume)/sum(base_volume) as avg,
  (max_price(high))[1] as high_n,
  (max_price(high))[2] as high_d,
  (min_price(low))[1] as low_n,
  (min_price(low))[2] as low_d,
  (first(open))[1] as open_n,
  (first(open))[2] as open_d,
  (last(close))[1] as close_n,
  (last(close))[2] as close_d
  FROM (
    SELECT
      div(((j.value->>'timestamp')::bigint - 0), 86400000)*86400000 + 0 as bucket,
      (j.value->>'timestamp')::bigint as timestamp,
      (j.value->>'count')::integer as count,
      (j.value->>'base_volume')::bigint as base_volume,
      (j.value->>'counter_volume')::bigint as counter_volume,
      ARRAY[(j.value->>'high_n')::bigint, (j.value->>'high_d')::bigint] as high,
      ARRAY[(j.value->>'low_n')::bigint, (j.value->>'low_d')::bigint] as low,
      ARRAY[(j.value->>'open_n')::bigint, (j.value->>'open_d')::bigint] as open,
      ARRAY[(j.value->>'close_n')::bigint, (j.value->>'close_d')::bigint] as close
    FROM
      history_trades_60000 h, jsonb_each(h.values) j
    WHERE h.base_asset_id = '2'
      AND h.counter_asset_id = '149'
    ORDER BY timestamp ASC
  ) as htrd
  GROUP BY bucket ORDER BY timestamp DESC LIMIT 200;

                                                                                                           QUERY PLAN
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=979.95..1622.70 rows=200 width=296) (actual time=1218.986..1223.731 rows=200 loops=1)
   ->  GroupAggregate  (cost=979.95..1622.70 rows=200 width=296) (actual time=1218.984..1223.723 rows=200 loops=1)
         Group Key: htrd.bucket
         ->  Sort  (cost=979.95..980.70 rows=300 width=180) (actual time=1218.870..1219.016 rows=1148 loops=1)
               Sort Key: htrd.bucket DESC
               Sort Method: external merge  Disk: 69184kB
               ->  Subquery Scan on htrd  (cost=963.86..967.61 rows=300 width=180) (actual time=1032.856..1084.236 rows=371658 loops=1)
                     ->  Sort  (cost=963.86..964.61 rows=300 width=188) (actual time=1032.854..1057.811 rows=371658 loops=1)
                           Sort Key: (((j.value ->> 'timestamp'::text))::bigint)
                           Sort Method: external merge  Disk: 74928kB
                           ->  Nested Loop  (cost=0.29..951.51 rows=300 width=188) (actual time=0.309..901.815 rows=371658 loops=1)
                                 ->  Index Scan using history_trades_60000_year_month_base_asset_id_counter_asset_key on history_trades_60000 h  (cost=0.29..912.51 rows=3 width=618) (actual time=0.060..0.521 rows=39 loops=1)
                                       Index Cond: ((base_asset_id = '2'::bigint) AND (counter_asset_id = '149'::bigint))
                                 ->  Function Scan on jsonb_each j  (cost=0.00..1.00 rows=100 width=32) (actual time=5.279..6.170 rows=9530 loops=39)
 Planning Time: 0.385 ms
 Execution Time: 1279.927 ms
(16 rows)
```

Keep in mind these are worst-cases (loading all of time for the most traded asset pair). Seems like most of the time is spent unwinding the jsonb. Intuitively I feel like that could be reduced, as it seems like the limit param isn't being propagated down as far as it could. Maybe I'm missing an index I had before? 🤔 
It also might be worth doing a benchmark without the jsonb storage. It ~doubles the data storage, but might make the query times much happier.

paulbellamy:
Thinking about this more, and benchmarking a bit more, I think we should *not* do the jsonb thing for now. The faster query times will take more load off the DB, and hard-drives are cheap(ish). Let's opt for the ~1600MB disk usage, and fastest query times.

Because all the data is available, if/when disk usage later becomes an issue we can reconsider a move to jsonb (via db migration, no reingest needed), with more knowledge about the usage patterns.

paulbellamy:
# reingestion benchmarking

<details>
<summary>baseline w 2.4.1: ~100ms/ledger</summary>
Screenshot 2021-06-23 at 10 27 39
</details>

services/horizon/internal/ingest/fsm.go
```diff
@@ -753,6 +760,11 @@ func (h reingestHistoryRangeState) run(s *system) (transition, error) {
 		}
 	}
 
+	err = s.historyQ.RebuildTradeAggregationBuckets(s.ctx, h.fromLedger, h.toLedger)
```
paulbellamy, line 763: Note, running this once at the end of the reingestion means that the aggregations will be out-of-date until the reingestion finishes.

paulbellamy: We could decide we're fine with that, or do some detection and rebuild it after each minute chunk. Running it once at the end is way more efficient though, as it can just load the whole range of trades at once (instead of loading each minute's).

paulbellamy: Per benchmarking with @jacekn it seems like the previous slowness was caused by missing an index to use for the DeleteRangeAll call. If we're concerned with consistency during reingestion we can try returning to rebuilding the buckets after each ledger, and time that.

services/horizon/internal/db2/history/trade_aggregation.go
```diff
@@ -212,12 +205,96 @@ func bucketTrades(resolution int64, offset int64) sq.SelectBuilder {
...
+		"counter_asset_id",
+		"count(*) as count",
+		"sum(base_amount) as base_volume",
+		"sum(counter_amount) as counter_volume",
+		"sum(counter_amount::numeric)/sum(base_amount::numeric) as avg",
+		"(max_price(price))[1] as high_n",
+		"(max_price(price))[2] as high_d",
+		"(min_price(price))[1] as low_n",
+		"(min_price(price))[2] as low_d",
+		"first(history_operation_id) as open_ledger",
```
bartekn, line 286:
```suggestion
		"first(history_operation_id) as open_ledger_toid",
```

services/horizon/internal/db2/history/trade_aggregation.go
```diff
@@ -212,12 +205,96 @@ func bucketTrades(resolution int64, offset int64) sq.SelectBuilder {
...
+		"(first(price))[1] as open_n",
+		"(first(price))[2] as open_d",
+		"last(history_operation_id) as close_ledger",
```
bartekn, line 289:
```suggestion
		"last(history_operation_id) as close_ledger_toid",
```

services/horizon/internal/ingest/fsm.go
```diff
@@ -456,6 +456,11 @@ func (r resumeState) run(s *system) (transition, error) {
 		return retryResume(r), errors.Wrap(err, "Error running processors on ledger")
 	}
 
+	err = s.historyQ.RebuildTradeAggregationBuckets(s.ctx, ingestLedger, ingestLedger)
```
bartekn, line 459: Can we add a metric here tracking how long rebuilding buckets took? Should be easy, check `s.Metrics().LedgerIngestionDuration.Observe(float64(duration))` below.

services/horizon/internal/db2/history/trade_aggregation.go
```diff
@@ -212,12 +205,96 @@ func bucketTrades(resolution int64, offset int64) sq.SelectBuilder {
...
+func (q Q) RebuildTradeAggregationBuckets(ctx context.Context, fromSeq, toSeq uint32) error {
+	fromLedger := toid.New(int32(fromSeq), 0, 0).ToInt64()
+	// toLedger should be inclusive here.
+	toLedger := toid.New(int32(toSeq+1), 0, 0).ToInt64()
+
+	// Clear out the old bucket values.
+	_, err := q.Exec(ctx, sq.Delete("history_trades_60000").Where(
+		sq.GtOrEq{"open_ledger": fromLedger},
+		sq.Lt{"open_ledger": toLedger},
+	))
```
bartekn, line 252: I think something is wrong here or I misunderstand what `history_trades_60000` holds. Given that `history_trades_60000` holds one minute trade aggregations buckets, check this chart:
```
---|-------------------|----------
   O      |            C      |
          |                   |
          X                   Y
```
The line in the top is one of the already aggregated buckets in a DB, it starts at ledger O (open) and ends at C (close). Then we run reingest of X-Y ledgers. What we want to delete buckets starting at O and C and I think that the highlighted code doesn't do that.

services/horizon/internal/db2/history/trade_aggregation.go
```diff
@@ -212,12 +205,96 @@ func bucketTrades(resolution int64, offset int64) sq.SelectBuilder {
...
-	)
+		"base_asset_id",
+		"base_amount",
+		"counter_asset_id",
+		"counter_amount",
+		"ARRAY[price_n, price_d] as price",
+	).From("history_trades").Where(
+		// TODO: Check if this is the right thing
+		sq.GtOrEq{"history_operation_id": fromLedger},
+		sq.Lt{"history_operation_id": toLedger},
```
bartekn, line 270: I think this looks good but please check the comment about the delete part above.

services/horizon/internal/db2/history/trade_aggregation.go
```diff
@@ -212,12 +205,96 @@ func bucketTrades(resolution int64, offset int64) sq.SelectBuilder {
...
+		"(first(ARRAY[open_n, open_d]))[2] as open_d",
+		"(last(ARRAY[close_n, close_d]))[1] as close_n",
+		"(last(ARRAY[close_n, close_d]))[2] as close_d",
+	).FromSelect(query, "htrd").GroupBy("timestamp")
+}
+
+// RebuildTradeAggregationBuckets rebuilds a specific set of trade aggregation buckets.
+// to ensure complete data in case of partial reingestion.
+func (q Q) RebuildTradeAggregationBuckets(ctx context.Context, fromSeq, toSeq uint32) error {
+	fromLedger := toid.New(int32(fromSeq), 0, 0).ToInt64()
```
bartekn, line 244:
```suggestion
	fromLedgerToid := toid.New(int32(fromSeq), 0, 0).ToInt64()
```

services/horizon/internal/db2/history/trade_aggregation.go
```diff
@@ -212,12 +205,96 @@ func bucketTrades(resolution int64, offset int64) sq.SelectBuilder {
...
+	// toLedger should be inclusive here.
+	toLedger := toid.New(int32(toSeq+1), 0, 0).ToInt64()
```
bartekn, line 246:
```suggestion
	toLedgerToid := toid.New(int32(toSeq+1), 0, 0).ToInt64()
```

bartekn (commented):
OK, it looks good to me! I added one major comment about delete part of `RebuildTradeAggregationBuckets` and a few minor comments. I know you checked the speed of the new thing (which looks really good!) but have you checked the correctness? I think what we could do is: deploy your branch to stg, rebuild buckets and then get as many `/trade_aggregation` requests from the access log and use `horizon-cmp` to check if the stable release returns the same data.

go.list
```diff
@@ -2,7 +2,7 @@ github.com/stellar/go
 cloud.google.com/go v0.34.0
 firebase.google.com/go v3.12.0+incompatible
 github.com/BurntSushi/toml v0.3.1
-github.com/Masterminds/squirrel v0.0.0-20161115235646-20f192218cf5
+github.com/Masterminds/squirrel v1.5.0
```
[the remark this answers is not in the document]
paulbellamy: The older version doesn't support inserting from a select, like in: `sq.Insert("history_trades_60000").Select(rebuilt)`.

services/horizon/internal/db2/schema/migrations/47_precompute_trade_aggregations.sql
```diff
@@ -0,0 +1,95 @@
...
+        first(history_operation_id) as open_ledger,
+        (first(price))[1] as open_n,
+        (first(price))[2] as open_d,
+        last(history_operation_id) as close_ledger,
+        (last(price))[1] as close_n,
+        (last(price))[2] as close_d
+      FROM htrd
+      GROUP by base_asset_id, counter_asset_id, timestamp
+      ORDER BY timestamp
+    );
```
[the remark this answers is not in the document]
paulbellamy: Would this actually need to cause downtime? I guess so, as the table locks would conflict with ingestion... 🤔

services/horizon/internal/db2/history/trade_aggregation.go
paulbellamy: You are right, yeah. We actually needed to find the first and last bucket timestamps from the given ledger range, not the actual ledger values. Otherwise we would only include partial data when rebuilding.

paulbellamy:
Note, when ingesting this failed with:
```
error rebuilding trade aggregations: could not rebuild trade aggregation bucket: exec failed: pq: duplicate key value violates unique constraint \"history_trades_60000_pkey\"
```
So, we are not clearing out the correct old buckets.

services/horizon/internal/db2/schema/migrations/47_precompute_trade_aggregations.sql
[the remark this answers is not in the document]
paulbellamy: After actually deploying this and using it, I think I agree. Having the migrations take a long time is a pain, as is maintaining the query in 2 places. Let's remove it from the migration.

PR #3648, opened 2021-06-01
services/horizon: Add script to run verify-range test AWS Batch job

2opremio:
~This PR creates a GitHub action which allows manually running a verify-range AWS Batch job~

~See https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/ for details on how this is supposed to work.~

This PR adds a script which allows running a verify-range AWS Batch job

Addresses part of #2761


2opremio:
It seems we need to merge this into master to test it :(

leighmcculloch:
What's the motivation for having this be a GitHub Action vs being a Jenkins pipeline?

2opremio:
> What's the motivation for having this be a GitHub Action vs being a Jenkins pipeline?

Unfamiliarity with Jenkins plus easier access.

Why would we use Jenkins instead?

.github/workflows/verify-range-test.yml
```diff
@@ -0,0 +1,58 @@
...
+      baseBranch:
+        description: 'Base Branch'
+        required: false
+        default: 'horizon-v2.0.0'
+      batchStartLedger:
+        description: 'Batch Start Ledger'
+        required: false
+        default: '1'
+      batchSize:
+        description: 'Batch Start Ledger'
```
Shaptic, line 18: Should be 'Batch Size', right?

.github/workflows/verify-range-test.yml
```diff
@@ -0,0 +1,58 @@
...
+        required: false
+        default: '2'
+      jobDefinition:
+        description: 'Job Definition'
+        required: false
+        default: 'verify-range-c5-large-job:4'
+      jobQueue:
+        description: 'Job Queue'
+        required: false
+        default: 'verfy-range-c5-large-queue'
```
Shaptic, line 32: I think we use 30016 for large and 10048 for xlarge, right? It seems like we only specify one of them here.

.github/workflows/verify-range-test.yml
```diff
@@ -0,0 +1,58 @@
...
+        required: false
+        default: '30016'
+      arraySize:
+        description: 'Array Size'
+        required: false
+        default: '2'
+      jobDefinition:
+        description: 'Job Definition'
+        required: false
+        default: 'verify-range-c5-large-job:4'
```
Shaptic, line 28: Can we make the revision programmable? Not sure the limitations here.

.github/workflows/verify-range-test.yml
```diff
@@ -0,0 +1,58 @@
...
+        required: false
+        default: '1'
+      batchSize:
+        description: 'Batch Start Ledger'
+        required: false
+        default: '30016'
+      arraySize:
+        description: 'Array Size'
+        required: false
+        default: '2'
```
Shaptic, line 24: This will only verify-range on the first 30016*2 ledgers by default, right? Is that what we want?

services/horizon/docker/verify-range/run_verify_range_test.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+
+set -e
+
+: ${JOB_NAME:=test_verify_range$RANDOM}
+: ${JOB_QUEUE:=verfy-range-c5-large-queue}
+: ${ARRAY_SIZE:=2}
+ARRAY_PROPERTIES="size=$ARRAY_SIZE"
+: ${JOB_DEFINITION:=verify-range-c5-large-job:4}
+: ${BRANCH:=release-horizon-v2.4.0}
```
Shaptic, line 10: Is `master` a better default?

services/horizon/docker/verify-range/run_verify_range_test.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+
+set -e
+
+: ${JOB_NAME:=test_verify_range$RANDOM}
+: ${JOB_QUEUE:=verfy-range-c5-large-queue}
```
Shaptic, line 6: Typo? verfy -> verify

Shaptic (commented):
I know there's some discussion about possibly abandoning this in lieu of Jenkins, but adding my thoughts here nonetheless.

services/horizon/docker/verify-range/run_verify_range_test.sh
2opremio: There was a typo when naming the queue, this is just consistent with it

2opremio:
@Shaptic It seems we are switching to Jenkins, I will remove the github job definition and just leave the script.

services/horizon/docker/verify-range/run_verify_range_test.sh
2opremio: Yeah, probably

PR #3652, opened 2021-06-02
services/regulated-assets-approval-server: implement SEP-8 "success" response

marcelosalloum:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Support SEP-8 [success](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0008.md#success) response.

### Why

According to SEP-8, if a client submits a transaction in a format already compliant with SEP-8 the server should be able to identify that case and sign the submitted transaction without revising it.

### Known limitations

N/A

### DIFF
--- a/services/regulated-assets-approval-server/internal/serve/api_tx_approve_test.go
+++ b/services/regulated-assets-approval-server/internal/serve/api_tx_approve_test.go
@@ -429,3 +429,152 @@ func TestAPI_txAprove_actionRequiredFlow(t *testing.T) {
 	}`
 	require.JSONEq(t, wantBody, string(body))
 }
+
+func TestAPI_txApprove_success(t *testing.T) {
+	ctx := context.Background()
+	db := dbtest.Open(t)
+	defer db.Close()
+	conn := db.Open()
+	defer conn.Close()
+
+	senderKP := keypair.MustRandom()
+	receiverKP := keypair.MustRandom()
+	issuerKP := keypair.MustRandom()
+	assetGOAT := txnbuild.CreditAsset{
+		Code:   "GOAT",
+		Issuer: issuerKP.Address(),
+	}
+	kycThresholdAmount, err := amount.ParseInt64("500")
+	require.NoError(t, err)
+
+	horizonMock := horizonclient.MockClient{}
+	horizonMock.
+		On("AccountDetail", horizonclient.AccountRequest{AccountID: senderKP.Address()}).
+		Return(horizon.Account{
+			AccountID: senderKP.Address(),
+			Sequence:  "5",
+		}, nil)
+
+	handler := txApproveHandler{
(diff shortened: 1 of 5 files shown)


services/regulated-assets-approval-server/internal/serve/api_tx_approve_test.go
```diff
@@ -429,3 +429,150 @@ func TestAPI_txAprove_actionRequiredFlow(t *testing.T) {
...
+			},
+			&txnbuild.Payment{
+				Destination: receiverKP.Address(),
+				Amount:      "1",
+				Asset:       assetGOAT,
+			},
+			&txnbuild.AllowTrust{
+				Trustor:       receiverKP.Address(),
+				Type:          assetGOAT,
+				Authorize:     false,
```
leighmcculloch, line 498: Do we place to support authorized to maintain liabilities in the future? CAP-18 introduced this and is what I expected we'd probably use.

leighmcculloch (approved):
Looks great. One comment, but looks good regardless.

services/regulated-assets-approval-server/internal/serve/api_tx_approve_test.go
marcelosalloum: AFAIK "authorized to maintain liabilities" is used when we're authorizing an Offer to remain open, right?

The idea for the reference server was to start supporting "payments" at first, I don't think we have plans to support "offers" in this service yet. I've just added a note explaining that in https://github.com/stellar/go/pull/3652/commits/b97ddb9d432230cadd77cfc10dfa77074052c3be.

leighmcculloch: If the account has an offer, then later does a payment, and you set authorized: false without maintaining liabilities, it will cancel any existing offers the account has. So if you want to allow offers to stay open you need to use maintain liabilities in all cases, not just when the sandwich contains the offer.

But if you don't plan to support offers at all, then yeah this makes sense.

marcelosalloum: > If the account has an offer, then later does a payment, and you set authorized: false without maintaining liabilities, it will cancel any existing offers the account has

That makes sense, thanks for highlighting that!

We don't have plans to support offers right now, so that's something we can handle when we add "offers" support.

leighmcculloch: One thing we should consider is that someone may use this as an example, and it would be no cost to us to use the maintain liabilities flag here, but it may improve the safety of the example if someone copies it.

leighmcculloch (approved):

PR #3653, opened 2021-06-02
services/horizon: Return 503 error in POST /transactions if Stellar-Core not synced

### LINKED ISSUE #3550: Horizon txsub endpoint should return 503 Service Unavailable when underlying Stellar-Core not synced

2.1.0

### What did you do?

During the SDF outage on April, 6th 2021 Horizon `POST /transactions` was accepting transactions and forwarding them to Stellar-Core. While Stellar-Core (maybe?) broadcasted these transactions to the network Horizon was not able to ingest ledgers (because Stellar-Core was not synced) thus it couldn't check the status of submitted transactions.

Horizon should detect if Stellar-Core it's using for submitting transactions is synced. If not, it should return `503 Service Unavailable` and do not attempt to submit transactions.

The status can be probably checked using [`stellarcore.Client.Info`](https://pkg.go.dev/github.com/stellar/go@v0.0.0-20210416120342-3b1409bd844f/clients/stellarcore#Client.Info).

### What did you see instead?

All valid transactions submitted during the outage were forwarded to Stellar-Core and were getting `504 Gateway Timeout`.

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This commit changes `POST /transactions` behaviour to return `503 Service Unavailable` (`stale_history`) instead of `504 Gateway Timeout` (`timeout`) response if connected Stellar-Core is not synced.

Close #3550.

### Why

`timeout` error is confusing when the connected Stellar-Core is not synced because it means that the transaction has been broadcasted but there is no result yet. What actually happens is that the transaction is not broadcasted (and Horizon cannot ingest it) when Stellar-Core is not in synced. The new response is returned as soon as Horizon determines that the Stellar-Core status has changed.

### Known limitations

[TODO or N/A]

### DIFF
--- a/services/horizon/internal/init.go
+++ b/services/horizon/internal/init.go
@@ -208,7 +208,7 @@ func initDbMetrics(app *App) {
 			Help: "determines if Stellar-Core defined by --stellar-core-url is synced with the network",
 		},
 		func() float64 {
-			if app.coreState.Synced {
+			if app.coreState.Get().Synced {
 				return 1
 			} else {
 				return 0
--- a/services/horizon/internal/init.go
+++ b/services/horizon/internal/init.go
@@ -208,7 +208,7 @@ func initDbMetrics(app *App) {
 			Help: "determines if Stellar-Core defined by --stellar-core-url is synced with the network",
 		},
 		func() float64 {
-			if app.coreSettings.Synced {
+			if app.coreState.Synced {
 				return 1
 			} else {
 				return 0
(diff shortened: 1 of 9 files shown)


services/horizon/internal/init.go
```diff
@@ -208,7 +208,7 @@ func initDbMetrics(app *App) {
 			Help: "determines if Stellar-Core defined by --stellar-core-url is synced with the network",
 		},
 		func() float64 {
-			if app.coreSettings.Synced {
```
tamirms, line 211: should you use the Get() function so that the read lock is acquired?

bartekn: Good catch! Updated `corestate.Store` to not expose `State` directly in 4dc19c65f4a47eec98d1f7a13942e19f991f66af.

tamirms (approved):

PR #3658, opened 2021-06-02
services/horizon: Fix route value for rate limited requests

### LINKED ISSUE #3635: high number of route=undefined in logs for rate-limited requests
From Tom LS:

 We are getting a high number (95%) of route=undefined for rate limited requests even though the path field is defined correctly in the same request.

To demonstrate this I am using requests to 
https://horizon.stellar.org/accounts/GCCUFCDJ5WL6JQEXGJONFZOT5RXDQLMMLCXP3JYKCRW4VFRJ2MQP7J37/transactions?cursor=now

This request for that route was rate limited and logged 

`route=/accounts/{account_id}/transactions`
and
`path=/accounts/GCCUFCDJ5WL6JQEXGJONFZOT5RXDQLMMLCXP3JYKCRW4VFRJ2MQP7J37/transactions?cursor=now`

correctly but this request was rate limited and logged  route=undefined with 

`path=/accounts/GCCUFCDJ5WL6JQEXGJONFZOT5RXDQLMMLCXP3JYKCRW4VFRJ2MQP7J37/transactions?cursor=now`

From what I can tell both of these requests appear to be identical, they hit the same instance horizon-blue-004c  and got the same response (429) yet they log differently.

Although we still get some route=undefined for non rate limited requests (6%) it is a much more significant for rate limited requests (95%).
Am I missing something obvious?

Links to logging data here ([slack](https://stellarfoundation.slack.com/archives/C02B04RMK/p1621987864135600?thread_ts=1621987830.135500&cid=C02B04RMK)).

bartekn:
### What

This commit fixes the value of `route` in the logs for rate-limited requests (previously it was set to `undefined`).

Close #3635.

### Known limitations

This solution was found in https://github.com/go-chi/chi/issues/270#issuecomment-479184559. Can we add this to our repo @ire-and-curses?


services/horizon/internal/httpx/middleware.go
```diff
@@ -149,8 +149,34 @@ func sanitizeMetricRoute(routePattern string) string {
...
+	}
+
+	// tctx has the updated pattern, since Match mutates it
+	return tctx.RoutePattern()
+}
+
 func logEndOfRequest(ctx context.Context, r *http.Request, requestDurationSummary *prometheus.SummaryVec, duration time.Duration, mw middleware.WrapResponseWriter, streaming bool) {
-	route := sanitizeMetricRoute(chi.RouteContext(r.Context()).RoutePattern())
+	route := sanitizeMetricRoute(getRoutePattern(r))
+	if route == "" {
```
tamirms, line 177: I don't think this check is necessary since it is included in `sanitizeMetricRoute()`

bartekn: Correct! Removed.

tamirms (approved):

PR #3661, opened 2021-06-04
services/horizon: Fix reingestion for the `history_{transaction,operation}_claimable_balances` tables

paulbellamy:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Also delete the ranges from `history_{transaction,operation}_claimable_balances` tables.

### Why

So that they don't get duplicate insertion conflicts when reingesting with existing claimable balance data.

### Known limitations

I didn't add any tests for this yet. How to test reingestion generally needs a bit more thought (IMO) than I wanted to have scope in this PR.


services/horizon/internal/db2/history/main.go
```diff
@@ -826,34 +826,22 @@ func (q *Q) CloneIngestionQ() IngestionQ {
 // DeleteRangeAll deletes a range of rows from all history tables between
 // `start` and `end` (exclusive).
 func (q *Q) DeleteRangeAll(ctx context.Context, start, end int64) error {
-	err := q.DeleteRange(ctx, start, end, "history_effects", "history_operation_id")
-	if err != nil {
-		return errors.Wrap(err, "Error clearing history_effects")
+	for _, pair := range [][2]string{
```
bartekn, line 829: Nit: Maybe anonymous struct would be more readable?

bartekn (approved):

services/horizon/internal/db2/history/main.go
paulbellamy: does the order actually matter? Cause a map a bit cleaner as well

bartekn: I think the order doesn't matter because we don't have any foreign keys in the schema. But I haven't checked if changing the order doesn't break it.

paulbellamy: Seems to work ok (on the broken range, anyway).

paulbellamy:
Ok, so had to update kahuna to get the tests passing again. Basically, just did a diff between the base-horizon.sql, and kahuna-horizon.sql. @bartekn does that seem like the right approach here? Not really sure on the normal maintenance for those files...

2opremio:
@paulbellamy I think that approach is fine. The scenarios are deprecated.

PR #3663, opened 2021-06-04
services/regulated-assets-approval-server: add missing postgres driver

marcelosalloum:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add missing Postgres driver that was being added automagically in the xdr package.

### Why

Builds started failing due to this change: https://github.com/stellar/go/commit/11f0c3f96df0ac113341d8a00fdc295fe30f246f.

### Known limitations

Ideally, we should add a test to prevent this from happening again but I didn't find out a way to do so since the database used in the tests and in the actual API are not the same.

### DIFF
--- a/services/regulated-assets-approval-server/internal/db/db.go
+++ b/services/regulated-assets-approval-server/internal/db/db.go
@@ -2,6 +2,7 @@ package db
 
 import (
 	"github.com/jmoiron/sqlx"
+	_ "github.com/lib/pq"
 )
 
 func Open(dataSourceName string) (*sqlx.DB, error) {


leighmcculloch (approved):
>Ideally, we should add a test to prevent this from happening again but I didn't find out a way to do so since the database used in the tests and in the actual API are not the same.

It is unfortunate that the method for importing and registering db drivers was used for the database/sql package in the Go ecosystem. There's actually another way, that if we used wouldn't suffer from this "remember to import it so it can register itself" pattern.

If we used `database/sql.OpenDB`, then we'd pass in the lib/pq connector, rather than rely on the magical registration.

Either way, thanks for catching. This may be broken in other applications now too 😕.

marcelosalloum:
> If we used `database/sql.OpenDB`, then we'd pass in the lib/pq connector, rather than rely on the magical registration.

Thanks for the suggestion! I can take a look at it next week to check how hard it is to make this change.

> Either way, thanks for catching. This may be broken in other applications now too 😕.

I'll ping some other Stellar folks so they can take a look at their end.

leighmcculloch:
> This may be broken in other applications now too 😕.

I just checked recoverysigner, and it is fine. I think the only other one would be Ticker, which I haven't checked.

PR #3670, opened 2021-06-07
ingest: Reuse Stellar-Core's cached buckets if possible.

### LINKED ISSUE #3631: Reuse Captive-Core storage path
Stellar-Core 17.1.0 allows reusing buckets for faster catchup (https://github.com/stellar/stellar-core/pull/2994). Right now we generate random subfolder at Captive-Core startup that prevents from using this feature. This should be done after protocol upgrade to ensure version 17.1.0 is installed.

Shaptic:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Stellar-Core can intelligently reuse existing bucket files (stellar/stellar-core#2994) as of v17.1.0 (see the [release notes](https://github.com/stellar/stellar-core/releases/tag/v17.1.0)). We store these under the `--captive-core-storage-path=./` directory. It used to be a randomly-generated subfolder (e.g. `captive-core-d34db33f/`) on startup; now it's kept consistent (just `captive-core/`) to take advantage of this feature.

### Why
Performance: faster startup. Closes #3631.

### Known limitations
This feature is only available with Stellar-Core >= v17.1.0. We can introduce version checks to keep having random buckets with < 17.1, but this will complicate startup significantly, since we'd now need to run & parse `stellar-core version`.

Needs #3683 to be merged first for proper 17.1 support in our builds.

### DIFF
--- a/ingest/CHANGELOG.md
+++ b/ingest/CHANGELOG.md
@@ -6,7 +6,8 @@ All notable changes to this project will be documented in this file. This projec
 ## Unreleased
 
 ### New Features
-* Performance improvement: Captive Core now reuses bucket files whenever it finds existing ones in the corresponding `--captive-core-storage-path` (introduced in [v2.0](#v2.0.0) rather than generating a one-time temporary sub-directory ([](https://github.com/stellar/go/pull/XXXX)). **This requires Stellar-Core version 17.1 or later.**
+* **Performance improvement**: the Captive Core backend now reuses bucket files whenever it finds existing ones in the corresponding `--captive-core-storage-path` (introduced in [v2.0](#v2.0.0)) rather than generating a one-time temporary sub-directory ([#3670](https://github.com/stellar/go/pull/3670)). Note that taking advantage of this feature requires [Stellar-Core v17.1.0](https://github.com/stellar/stellar-core/releases/tag/v17.1.0) or later.
+
 
 ## v2.0.0
 
--- a/ingest/CHANGELOG.md
+++ b/ingest/CHANGELOG.md
@@ -5,6 +5,8 @@ All notable changes to this project will be documented in this file. This projec
 
 ## Unreleased
 
+### New Features
+* Performance improvement: Captive Core now reuses bucket files whenever it finds existing ones in the corresponding `--captive-core-storage-path` (introduced in [v2.0](#v2.0.0) rather than generating a one-time temporary sub-directory ([](https://github.com/stellar/go/pull/XXXX)). **This requires Stellar-Core version 17.1 or later.**
 
 ## v2.0.0
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -147,12 +147,12 @@ commands:
     parameters:
       core-version:
         type: string
-        default: "17.1.0-600.a5e7028.xenial"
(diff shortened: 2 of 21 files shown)


ire-and-curses:
> The next release of Horizon will require Stellar-Core >= v17.1.0.

What's the reason Horizon is not backwards-compatible?

Shaptic:
@ire-and-curses I think I misspoke. More accurately: to take advantage of the performance improvement, we need >= 17.1. Otherwise, I think Core will just flush the buckets folder. I'm building an older binary as we speak to make sure it's still backwards compatible. The biggest change in both cases, though, is that the directory isn't cleaned up on shutdown anymore.

ire-and-curses:
> Otherwise, I think Core will just flush the buckets folder.

Ok, that's what I was expecting, thanks. Yeah as you say let's confirm for sure it will still work with an older version (and just won't be any faster).

bartekn:
> Otherwise, I think Core will just flush the buckets folder. I'm building an older binary as we speak to make sure it's still backwards compatible.

Have you checked this? I think Stellar-Core does not flush the folder between runs.

I checked the [failing test](https://app.circleci.com/pipelines/github/stellar/go/8432/workflows/a2b2d1d7-2968-4c3b-873d-8607d1b65eac/jobs/46170): `TestReingestDB`. I think the problem is that the test starts Horizon normally in the background somewhere (maybe by our integration test framework) because the log output doesn't make sense. In reingestion we don't ingest state nor verify state but the logs suggest something else. I think if we disable starting Horizon before we actually run reingestion command it should work. It's possible that it worked before because different (random) bucket folder was used.

Shaptic:
> I think Stellar-Core does not flush the folder between runs.

I tested with 17.0 and confirmed that it does delete and redownload the bucket files @bartekn.

> I think the problem is that the test starts Horizon normally in the background somewhere

Yep, I came to the same exact conclusion and introduced `IntegrationTest.StopHorizon()` to alleviate this in f9aaba985a033bdf5b657cabe89ef139ea0f3770. All tests pass now!

ingest/CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ All notable changes to this project will be documented in this file. This projec
 
 ## Unreleased
 
+### New Features
+* **Performance improvement**: the Captive Core backend now reuses bucket files whenever it finds existing ones in the corresponding `--captive-core-storage-path` (introduced in [v2.0](#v2.0.0)) rather than generating a one-time temporary sub-directory ([#3670](https://github.com/stellar/go/pull/3670)). Note that taking advantage of this feature requires [Stellar-Core v17.1.0](https://github.com/stellar/stellar-core/releases/tag/v17.1.0) or later.
```
bartekn, line 9: Shouldn't we use value from `Config` instead of `--captive-core-storage-path`?

bartekn (approved):
> I tested with 17.0 and confirmed that it does delete and redownload the bucket files @bartekn.

Sounds good! On the contrary 17.1 does not delete the buckets, correct?

LGTM!

Shaptic:
@bartekn correct! The catchup is much faster as expected.

PR #3677, opened 2021-06-08
xdr: Easily created MuxedAccount objects from a G-address and a memo ID.

Shaptic:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Introduces `MuxedAccountFromAccountId`, which makes it easy to turn an account ID (`G...`) into a full `MuxedAccount` with a memo ID (i.e. representing an `M...` address). Also introduces `MuxedAccount.GetId()`, making it easy to extract the underlying memo ID if it's present.

### Why
If you have a G-address, it should be easy to create M-addresses from it.

### Known limitations
n/a


xdr/muxed_account.go
```diff
@@ -21,6 +21,43 @@ func MustMuxedAddressPtr(address string) *MuxedAccount {
...
+	idBytes, err := muxed.Med25519.Id.MarshalBinary()
+	if err != nil {
+		return MuxedAccount{}, err
+	}
+
+	mAddress := strkey.MustEncode(
+		strkey.VersionByteMuxedAccount,
+		append(keyBytes, idBytes...),
+	)
+	return AddressToMuxedAccount(mAddress)
```
bartekn, line 58: I think I'm missing something here but why can't we return muxed instead of all these lines? `strkey` roundtrip seems unnecessary?
```suggestion
	return muxed, err
```

Shaptic: I must've been written this at 3am or something... that's the only explanation. That or working with muxed account objects is just *confusing*, hence the need for said helpers lol. Anyway, tysm for the catch, resolved in latest push.

xdr/muxed_account.go
```diff
@@ -21,6 +21,24 @@ func MustMuxedAddressPtr(address string) *MuxedAccount {
...
 }
 
+func MuxedAccountFromAccountId(gAddress string, id uint64) (MuxedAccount, error) {
+	bytes, err := strkey.Decode(strkey.VersionByteAccountID, gAddress)
+	if err != nil {
+		return MuxedAccount{}, err
+	}
+
+	var publicKey Uint256
+	copy(publicKey[:], bytes)
```
bartekn, line 31: I haven't noticed it before but I think it's worth replacing this with `xdr.AddressToAccountId` (then you can use `PublicKey.Ed25519` in `MuxedAccountMed25519`. The reason is that it's using `AccountId.SetAddress` which performs key length validation so it can prevent some subtle bugs.

bartekn (approved):

xdr/muxed_account.go
Shaptic: Nice, wasn't aware of this method.

bartekn:
It would be super helpful for me when writing tests for #3722. Please merge.

PR #3680, opened 2021-06-09
services/regulated-assets-approval-server: refactor tests targetting the "/kyc-status/*" endpoints

marcelosalloum:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Refactor tests targetting the `/kyc-status/*` endpoints. Now the unit tests are covering not only the happy path but also all edge cases from the transaction approval logic. On the other hand, the API tests were reduced to cover the happy path only, so they become easier to maintain and we don't repeat ourselves (DRY).

### Why

There was confusion between unit tests and API tests in this service so the package was not properly tested in terms of unit tests and a lot of the edge cases were being validated only in the API tests. That setup makes the code and tests harder to maintain as a developer would need to maintain package-related tests outside the package.

### Known limitations

N/A

### DIFF
--- a/services/regulated-assets-approval-server/internal/serve/api_kyc_status_test.go
+++ b/services/regulated-assets-approval-server/internal/serve/api_kyc_status_test.go
@@ -21,44 +21,36 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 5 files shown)


acharb (approved):
lgtm!

Issue #3602, opened 2021-05-17
Restart captive core when a new version of core is detected on disk


ire-and-curses:
### What problem does your feature solve?

Horizon running in captive-core mode spawns stellar-core as a subprocess. If the stellar-core package is upgraded after Horizon is started the core binary changes on disk but the captive-core child process is NOT restarted.

This means that we end up in a situation where the running core version can be different from the version installed. This behaviour is confusing, non-obvious and hard to spot and can have unexpected impact on services. An unscheduled reboot can cause surprising behaviour some time after the update.

In the case of a network-wide protocol upgrade although the operator may have updated the captive-core version, the running captive-core might be a version of the core binary that does not support new protocol, leading to surprise downtime.

Forgetting to stop/start the binary has caught us out in our own deployments more than once so is likely a problem for others.

The behaviour of restarting services is standard in the Linux package world (see e.g. Postgres). To control upgrade timing, the best practice approach is to [hold packages](https://www.cyberciti.biz/faq/apt-get-hold-back-packages-command/). We believe this should be familiar to operators but we can consider providing an informative message in the Horizon logs.

### What would you like to see?
Horizon monitors captive-core on disk, and restarts captive-core if it detects a change (e.g. a more recent file timestamp for the captive-core binary). This was discussed internally at SDF ([doc](https://docs.google.com/document/d/1ykVAWoJ1ZHlYChkw93AN5T1xYvxwsxrlV1iVWVuhAZE/edit#)).

This is a break with previous behaviour. We should document this change.

There is a risk that some users will experience unscheduled downtime as a result. However, latest core release candidate (17.1.0rc1) shrinks core sync time from over 5 minutes to around 30s (see https://github.com/stellar/stellar-core/issues/2960), so in this case downtime will be small. Also, auto-updating Horizon is a bad idea, and forgetting to hold the package is only a mistake you make once as an operator.

### What alternatives are there?

- Keep things the way they are
- Push responsibility for detecting changes on disk to operators (who would need additional tooling, e.g. Puppet)
- It might be possible to build a meta package that forces the dependency to update in a post-installation step

PR #3687, opened 2021-06-11
ingest/ledgerbackend: Restart captive core when a new version of core is detected on disk

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Close https://github.com/stellar/go/issues/3602

* Add a ticker to stellarCoreRunner which will execute every 10 seconds. The stellarCoreRunner context will interrupt the goroutine attached to the ticker so that if captive core shutsdown the ticker will also stop.

* Upon every tick, we will check to see if the stellar core binary has been modified. If it has been modified we log a warning message to announce we will be shutting down captive core and then we call stellarCoreRunner.close() to terminate the captive core instance. Once stellarCoreRunner.close() is called any pending operations (PrepareRange() or GetLedger()) should return immediately with an error and the ingestion state machine will handle the error by retrying.

* When the retry happens it will create a new instance of stellarCoreRunner which should use the newest stellar core binary.

### Why

See https://github.com/stellar/go/issues/3602 for motiviation.

### DIFF
--- a/ingest/ledgerbackend/stellar_core_runner.go
+++ b/ingest/ledgerbackend/stellar_core_runner.go
@@ -249,6 +249,11 @@ func (r *stellarCoreRunner) catchup(from, to uint32) error {
 		return errors.Wrap(err, "error waiting for `stellar-core new-db` subprocess")
 	}
 
+	binaryWatcher, err := newFileWatcher(r)
+	if err != nil {
+		return errors.Wrap(err, "could not create captive core binary watcher")
+	}
+
 	rangeArg := fmt.Sprintf("%d/%d", to, to-from+1)
 	cmd := r.createCmd(
 		"catchup", rangeArg,
@@ -256,7 +261,6 @@ func (r *stellarCoreRunner) catchup(from, to uint32) error {
 		"--in-memory",
 	)
 
-	var err error
 	r.pipe, err = r.start(cmd)
 	if err != nil {
 		r.closeLogLineWriters(cmd)
@@ -266,6 +270,7 @@ func (r *stellarCoreRunner) catchup(from, to uint32) error {
 	r.started = true
 	r.ledgerBuffer = newBufferedLedgerMetaReader(r.pipe.Reader)
 	go r.ledgerBuffer.start()
+	go binaryWatcher.loop()
 	r.wg.Add(1)
 	go r.handleExit(cmd)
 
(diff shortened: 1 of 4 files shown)


bartekn (approved):

services/horizon/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ file. This project adheres to [Semantic Versioning](http://semver.org/).
 
 * Performance improvement: Captive Core now reuses bucket files whenever it finds existing ones in the corresponding `--captive-core-storage-path` (introduced in [v2.1.0](#v2.1.0) rather than generating a one-time temporary sub-directory ([](https://github.com/stellar/go/pull/XXXX)). **This feature requires Stellar-Core version 17.1 or later.**
```
Shaptic, line 12: Unrelated but could you fix the PR link here to reference #3670? Forgot to update it after the fact.

services/horizon/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ file. This project adheres to [Semantic Versioning](http://semver.org/).
...
 
+* Horizon monitors captive-core on disk, and restarts captive-core if it detects a change (e.g. a more recent file timestamp for the captive-core binary). ([3687](https://github.com/stellar/go/pull/3687)).
```
Shaptic, line 14: It monitors the Stellar Core binary, right? Well, more specifically, the binary pointed to by `--stellar-core-binary-path`. Maybe something like:

```suggestion
* Horizon now monitors the Stellar Core binary on disk (pointed to by `--stellar-core-binary-path`/`STELLAR_CORE_BINARY_PATH`) and restarts its Captive Core subprocess if it detects changes (i.e a more recent file timestamp for the Stellar Core binary) ([3687](https://github.com/stellar/go/pull/3687)).
```

Shaptic (approved):
:+1:

ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -249,14 +249,18 @@ func (r *stellarCoreRunner) catchup(from, to uint32) error {
 		return errors.Wrap(err, "error waiting for `stellar-core new-db` subprocess")
 	}
 
+	binaryWatcher, err := newFileWatcher(r)
+	if err != nil {
+		return errors.Wrap(err, "could not create captive core binary watcher")
+	}
+
```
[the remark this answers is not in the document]
Shaptic: Though it seems like the only error case is where we can't `stat` the stellar-core binary, which means we probably failed way before this anyway...

bartekn:
@tamirms please merge when ready.

PR #3693, opened 2021-06-15
all: Bump Golang minor versions in accordance with latest release.

Shaptic:
### What
Bumps from 1.16.3 -> 1.16.5 and 1.15.11 -> 1.15.13.

### Why
This bump includes security issues. Refer to the [release announcement](https://groups.google.com/g/golang-announce/c/RgCMkAEQjSI/m/r_EP-NlKBgAJ) for more.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -227,7 +227,7 @@ jobs:
   check_code_1_16:
     working_directory: /go/src/github.com/stellar/go
     docker:
-      - image: golang:1.16.5
+      - image: golang:1.16
     steps:
       - install_go_deps
       - check_go_deps
@@ -282,7 +282,7 @@ jobs:
   test_code_1_16:
     working_directory: /go/src/github.com/stellar/go
     docker:
-      - image: golang:1.16.5
+      - image: golang:1.16
         environment:
           GO111MODULE: "on"
           PGHOST: localhost
@@ -302,7 +302,7 @@ jobs:
   test_code_1_16_postgres10:
     working_directory: /go/src/github.com/stellar/go
     docker:
-      - image: golang:1.16.5
+      - image: golang:1.16
         environment:
           GO111MODULE: "on"
           PGHOST: localhost
@@ -325,7 +325,7 @@ jobs:
   publish_artifacts:
     working_directory: /go/src/github.com/stellar/go
     docker:
-      - image: golang:1.16.5
+      - image: golang:1.16
     steps:
       - check_deprecations
       - install_go_deps
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -241,7 +241,7 @@ jobs:
   test_code_1_15:
     working_directory: /go/src/github.com/stellar/go
     docker:
-      - image: golang:1.15.13
+      - image: golang:1.15
         environment:
           GO111MODULE: "on"
           PGHOST: localhost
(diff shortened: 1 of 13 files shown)


tamirms (approved):

Shaptic:
I found these references based on Ctrl+F on things like `go:`, `1.16`, `golang:`, and `1.15`, therefore it may not be all of the possible references, but it damn-sure is a lot of them.

.circleci/config.yml
```diff
@@ -227,7 +227,7 @@ jobs:
   check_code_1_16:
     working_directory: /go/src/github.com/stellar/go
     docker:
-      - image: golang:1.16
+      - image: golang:1.16.5
```
bartekn, line 230: Maybe we should replace it with:
```suggestion
      - image: golang:1.16
```
so that it will always pull latest Golang 1.16.x. This was recommended by @jacekn and @satyamz in the issue about bumping Golang versions in SDF. Same thing for 1.15.x.

ire-and-curses: I like this idea. We basically always want to be on latest patch release. Let's do it.

For minor versions let's stick to manual update for now.

Shaptic: Cool, I didn't realize 1.16 would always be the latest

Shaptic: Done!

services/horizon/docker/Dockerfile.dev
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.16.3 AS builder
+FROM golang:1.16.5 AS builder
```
bartekn, line 1:
```suggestion
FROM golang:1.16 AS builder
```

.github/workflows/go.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        go: [1.16.4]
```
bartekn, line 14: Is it possible to use `1.16` and `1.14` in Go actions?

Shaptic: Hmm, reading [this article](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix) now to figure that out...

Shaptic: Looks like no, because the `go` variable here refers to a version (from [here](https://github.com/actions/go-versions/blob/main/versions-manifest.json)) rather than an image.

bartekn (approved):

services/friendbot/docker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.16.3 as build
```
leighmcculloch, line 1: Why did we remove the patch version from some locations like this one, but we kept the patch versions in other services?

ire-and-curses: are you talking about these? https://github.com/stellar/go/pull/3693#discussion_r652960689

leighmcculloch: That link doesn't seem to go anywhere for me. In the CI files patch versions are used, and in the Ticker, Keystore, and other docker files a patch version is used. It'd be helpful if we are consistent.

Shaptic: Using the patch version means updating the patch version. 
Using the minor version for Docker images (e.g. `golang:1.16`) means we always get the latest patch version.
We can't do this for things like download URLs, however, so we still need the patch version sometimes.

leighmcculloch: Got it. What's the motivation for only making that change for Friendbot and Horizon, and not the other services?

PR #3694, opened 2021-06-16
ingest/ledgerbackend: Captive-Core fixes to support Stellar-Core 17.1.0

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Multiple fixes to support Captive-Core persistent storage in 17.1.0:
* Switched from `exec.CommandContext` to `exec.Command` when starting Stellar-Core. This was done because `CommandContext` kills the process (`Process.Kill()`) without waiting for a graceful shutdown. This can corrupt minimal DB or buckets. We send `signal.Interrupt` right now so Stellar-Core can exit gracefully. There's one gotcha: `signal.Interrupt` is not supported in Windows so we kill the process and empty storage folder that can contain corrupted data.
* Removed `DISABLE_XDR_FSYNC` setting in `stellar-core.cfg`. This can corrupt buckets when Stellar-Core is killed.
* Changed the behaviour of `CaptiveStellarCore.nextLedger`. Previously it was set to a value calculated using `CaptiveStellarCore.runFromParams` however these calculations are no longer correct when Stellar-Core is restarted with persistent storage. To prevent some Stellar-Core version checks we now set `CaptiveStellarCore.nextLedger` after the first ledger is streamed.
* Reverted the code creating and removing storage directory on Windows.

### Why

Stellar-Core 17.1.0 now persists the minimal DB and buckets between executions. It allows faster catchup on restart.

### Known limitations

[TODO or N/A]


ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -274,7 +296,7 @@ func (r *stellarCoreRunner) catchup(from, to uint32) error {
 	}
 
 	r.wg.Add(1)
-	go r.handleExit(cmd)
```
tamirms, line 277: if you look a few lines above this code, I think `r.cmd` should be assigned to the result of `r.createCmd()`

ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -398,5 +471,11 @@ func (r *stellarCoreRunner) close() error {
 		r.pipe.Reader.Close()
 	}
 
-	return nil
+	if runtime.GOOS != "windows" {
```
tamirms, line 474: imo it would be clearer to have

```golang

if runtime.GOOS == "windows" {
   return os.RemoveAll(storagePath)
}
return nil
```

ingest/ledgerbackend/stellar_core_runner.go
bartekn: Thanks, fixed in 2c0e5e01c3.

ingest/ledgerbackend/stellar_core_runner.go
bartekn: Good point, fixed in 12be1180a8.

ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -340,7 +410,10 @@ func (r *stellarCoreRunner) handleExit(cmd *exec.Cmd) {
 	}
 
 	r.processExited = true
-	r.processExitError = exitErr
+	if interruptErr := <-errc; interruptErr != nil {
+		r.processExitError = interruptErr
+	}
+	r.processExitError = waitErr
```
tamirms, line 416: I think this needs to be inside an else

bartekn: Good catch, fixed in ed5e87807d.

ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -398,5 +472,11 @@ func (r *stellarCoreRunner) close() error {
 		r.pipe.Reader.Close()
 	}
 
+	if runtime.GOOS == "windows" {
+		// It's impossible to send SIGINT on Windows so buckets can become
+		// corrupted. If we can't reuse it, then remove it.
+		return os.RemoveAll(storagePath)
```
tamirms, line 478: if the buckets are corrupted what happens if we don't remove the directory? will captive core be unable to start at all? should we also remove the directory on linux in the scenario that captive core does not shutdown gracefully and we have to use sigkill?

bartekn: Good point. I believe the change in 3d28e9e466 should fix it (remove folder if there was an error terminating the process).

tamirms: I don't see how `r.processExitError != nil` implies that the process must have been terminated by sigkill. Isn't the scenario below possible?

1. context is cancelled
2. we send sigint to captive core
3. captive core terminates cleanly before the 10 second timeout (we don't need to send sigkill)
4. `r.processExitError` is assigned the context error which is non nil

bartekn: I forgot about `context.Cancelled` - does 3200baf6d7 look good now?

tamirms: @bartekn I ran it locally and it seems to work. it might be worth adding an assertion in the integration tests here:

https://github.com/stellar/go/blob/master/services/horizon/internal/test/integration/integration.go#L252

if we're running the integration tests on windows with captive core then we expect that the buckets directory to still exist after horizon has shut down

ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -495,7 +488,7 @@ func (c *CaptiveStellarCore) handleMetaPipeResult(sequence uint32, result metaRe
 	}
 
 	seq := result.LedgerCloseMeta.LedgerSequence()
-	if seq != c.nextLedger {
+	if c.nextLedger != 0 && seq != c.nextLedger {
```
tamirms, line 491: when `c.nextLedger ` is 0, do we need to have a check that `sequence` must be less than or equal to `c.nextExpectedSequence()` ?

tamirms: Sorry, I meant should we check that `seq` must be less than or equal to `c.nextExpectedSequence()`.

bartekn: I think we shouldn't. `seq` is the sequence number of the ledger from the stream and it can be smaller than `nextExpectedSequence()` when we start streaming (Stellar-Core sometimes streams ledger starting from the checkpoint ledger before the requested ledger range start).

tamirms: right, `seq` can be smaller than `nextExpectedSequence()` but it should never be greater than `nextExpectedSequence()`

bartekn: OK, I see what you mean now. Sounds good, will add an extra check for security.

paulbellamy:
> ... without waiting for a graceful shutdown. This can corrupt ...

Isn't that an issue if it gets OOM-killed, or powered-off too?

ingest/ledgerbackend/captive_core_backend.go
bartekn: Added in https://github.com/stellar/go/pull/3694/commits/36251ee7536bd423a66803c0ead9825ccefd084d.

bartekn:
> Isn't that an issue if it gets OOM-killed, or powered-off too?

Correct but I believe there's nothing we can other than, maybe, creating a lock file as Stellar-Core does. But it will still not help with data corruption. In such case storage dir need to be removed manually.

tamirms (approved):

Issue #3696, opened 2021-06-16
CAPTIVE_CORE_LOG_PATH config value is ignored


bartekn:
### What version are you using?

2.4.1

### What did you do?

Set `CAPTIVE_CORE_LOG_PATH` to a path.

### What did you expect to see?

Captive-Core logs sent to a specified file.

### What did you see instead?

Log file not created, generated `stellar-core.cfg` contains: `LOG_FILE_PATH = ""`.

PR #3702, opened 2021-06-17
Fix configuration via environment variables for optional parameters

tamirms:
Fixes https://github.com/stellar/go/issues/3696

This PR fixes a bug where the configuration variables listed below were ignored if they were configured via environment variables instead of command line arguments:

```
CAPTIVE_CORE_LOG_PATH
CAPTIVE_CORE_PEER_PORT
CAPTIVE_CORE_HTTP_PORT
```


bartekn (approved):
LGTM! I'm just wondering if `CaptiveCoreTomlParams.HTTPPort` can't be `uint` instead of `*uint`. `0` value disabled the HTTP server so should work with a non-pointer default value.

tamirms:
@bartekn if http port is omitted the default value assumed by stellar core is actually is 11626 not 0

PR #3712, opened 2021-06-22
horizon: add package-based Dockerfile and Makefile

jacekn:
### What

Add horizon package-based Dockerfile and Makefile

### Why

This change will allow anyone to rebuild the official docker image and makes the build process more transparent.

### Known limitations

N/A


satyamz (approved):
LGTM!

tamirms (approved):

services/horizon/docker/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM ubuntu:focal
+
+ARG VERSION
+ARG DEBIAN_FRONTEND=noninteractive
```
paulbellamy, line 4: Should we also lock the core versions here like we do in the `.dev` version?
```
ENV STELLAR_CORE_VERSION 17.1.0-584.fbc0325.focal
ENV STELLAR_CORE_BINARY_PATH /usr/bin/stellar-core
```

paulbellamy (approved):

services/horizon/docker/Dockerfile
jacekn: We're planning to use this Dockerfile as part existing build pipeline. This means that we can build and tag images automatically when horizon packages are released. For example we'd bump horizon version in the `stellar-horizon:stable` image when we promote deb pacakges to the stable pocket. We'll probably publish `stellar-horizon:testing` whenever we push packages to the `testing` apt repo pocket

PR #3721, opened 2021-06-24
exp/services/webauth: make JWT token deterministic according with the challenge transaction

### LINKED ISSUE #3717: exp/services/webauth: make jwt deterministic based on challenge transaction
### What
Make the JWT that the webauth service generates deterministic based off the challenge transaction so that a single challenge transaction has a single JWT object.

### Why
To make our reference implementation consistent with stellar/stellar-protocol#982.

marcelosalloum:
<details>
  
### PR Structure

* x ] This PR avoids mixing refactoring changes with feature changes (split into two PRs
  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Make the payload of the JWT token deterministic based on the challenge transaction. The changes include:
* The JWT `iat` (issued at) parameter is set to the challenge transaction minimum time bound
* The JWT `exp` (expiration) parameter is set to the a fixed distance from that `iat`
* The tests in `token_test.go` where updated to validate the exact value of `iat` and `exp` since they're now deterministic.
* Add a new test to make sure replaying the same challenge transaction in a different period of time will return a JWT with the same payload.

### Why

Closes #3717.

### Further discussion

Although the JWT header and payload are deterministic, the JWT signature can't currently be made deterministic because the signing method uses `rand.Reader` to generate a random secure number [here](https://github.com/square/go-jose/blob/v2/asymmetric.go#L525).

With this, two JWTs from the same challenge transaction will look like this:
* `headers.payload.signatureA`
* `headers.payload.signatureB`


exp/services/webauth/internal/serve/token_test.go
```diff
@@ -121,12 +121,141 @@ func TestToken_formInputSuccess(t *testing.T) {
...
+	token2, err := jwt.Parse(res2.Token, func(token *jwt.Token) (interface{}, error) {
+		if _, ok := token.Method.(*jwt.SigningMethodECDSA); !ok {
+			return nil, fmt.Errorf("unexpected signing method: %v", token.Header["alg"])
+		}
+		return &jwtPrivateKey.PublicKey, nil
+	})
+	require.NoError(t, err)
+
+	claims2 := token2.Claims.(jwt.MapClaims)
+	require.Equal(t, claims1, claims2)
```
leighmcculloch, line 256: I think we should take the raw JWT token string, and strip off the signature manually (you can do this by splitting on the second separator), and checking the rest of the token is identical. My suggestion for doing that is because there are more than claims in a JWT.

leighmcculloch (approved):
The change looks great, there's one suggestion regarding the test below. Otherwise looks great. Thanks!

exp/services/webauth/internal/serve/token_test.go
marcelosalloum: Cool, doing that in https://github.com/stellar/go/pull/3721/commits/d2255bbf449bd1db8c4d1df75e1a8cd1cc23167d.

leighmcculloch (approved):
Looks great!

PR #3727, opened 2021-06-25
all: cache go module downloads and other build and test artifacts

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Cache Go module downloads and other build and test results so that for a single branch subsequent commits after the first commit use cached data.

The caching applies only to branches that aren't master.

On a branches first run it may use a cache from a different branch.

### Why

To speed up GitHub Actions test runs. They appear to be somewhat slower than CircleCI test runs, and @bartekn pointed out this might be because in CircleCI test runs we cache the Go module downloads.

Caching module downloads is already what we do in our CircleCI builds, so adding that here is not controversial.

Caching test results is new though. The go toolchain has great support for test result caching. This is noticeable when running tests locally: Go will only run tests for packages that have changed, or some external input to the test has changed, such as an environment variable or a file the test reads.

### Known limitations

N/A


leighmcculloch:
The builds on the first commit in this PR passed in these time periods:

Build | Duration
---
Go / check (ubuntu-latest, 1.16.5) | 2m
Go / build (ubuntu-latest, 1.16.5) | 55s
Go / build (ubuntu-latest, 1.15.13) | 59s
Go / test (ubuntu-latest, 1.16.5, 9.6.5) | 4m
Go / test (ubuntu-latest, 1.16.5, 10) | 5m
Go / test (ubuntu-latest, 1.15.13, 9.6.5) | 5m
Go / test (ubuntu-latest, 1.15.13, 10) | 5m

After I pushed an empty commit, the builds – that should be using the cache – passed in these time periods:

Build | Duration
---
Go / check (ubuntu-latest, 1.16.5) | 2m
Go / build (ubuntu-latest, 1.16.5) | 27s
Go / build (ubuntu-latest, 1.15.13) | 58s
Go / test (ubuntu-latest, 1.16.5, 9.6.5) | 4m
Go / test (ubuntu-latest, 1.16.5, 10) | 5m
Go / test (ubuntu-latest, 1.15.13, 9.6.5) | 5m
Go / test (ubuntu-latest, 1.15.13, 10) | 4m

There doesn't appear to be much difference, probably because we use the `-race` option when running tests that ignores the test cache.

It might be still worth keeping these cache lines for caching the Go module dependencies, at least any outage in any server we download dependencies from won't impact our builds, but it appears to have no material difference on build time.

bartekn (approved):

2opremio:
@leighmcculloch are you still pursuing this or aren't the preformance improvements good enough?

leighmcculloch:
> @leighmcculloch are you still pursuing this or aren't the preformance improvements good enough?

Yup, I am still working on this. There are concerns that @Shaptic raised offline that I plan to look into.

leighmcculloch:
Reopening this PR because @2opremio DM'd me saying he thinks there would be value in adding this now. @Shaptic I don't think the concerns around integration tests will be impacted by this because integration tests do not run in this build.

leighmcculloch:
Commit 0c1a1d186ddca5371f06b44cbe41e679cdcbe7ab, which should have had no caches available, had the following build times:
Screen Shot 2022-02-23 at 10 27 29 AM

leighmcculloch:
Commit da834c946b978d93fc539a296432f278fdc6de39, which should have had all caches available, had the following build and test times:
Screen Shot 2022-02-23 at 10 31 58 AM

leighmcculloch:
Looking at the two screenshots above the tests are faster, but not world changing. Total build time is reduced from 6 minutes to 1 minute. Is 5 minutes worth adding the additional complexity to our builds and additional risk that comes with running less tests in our PRs? I'm leaning no its not worth it, but I've heard some interest so:

@stellar/horizon-committers @stellar/product-eng-team @stellar/platform-committers What do you think?

leighmcculloch:
@2opremio I added the caching to the Horizon integration test that you just added and merge into this branch. The integration tests do not appear to be cacheable though. If they're executing docker commands and interacting with the environment and shelling out, that will probably be the reason. Go tracks all those interactions to determine if the test is reasonably cacheable.

ire-and-curses:
> Is 7 minutes worth adding the additional complexity to our builds and additional risk that comes with running less tests in our PRs?

Not sure I see what the additional risk is? Go is not running tests where it determines nothing could have changed. Seems solid to me. Can we think of an example where this could allow a silent failure? Leaning towards yes on that basis.

leighmcculloch:
That's a good point. Even if we cache the wrong thing and there's bugs in that, we are ultimately relying on the Go build / test code choosing to use the cache or not. We can ignore the statement about risk I think. Also, this PR ignores the cache on the master branch, so we do run all tests there everytime if any risk ever develops in the future.

2opremio:
I think 7 mins is a great improvement. I've been looking at how to factor a `uses` step out, and it seems we would need to create our own action, which would probably be fine.

We could also fix https://github.com/actions/setup-go/issues/130 and upstream it

leighmcculloch:
We could break this out into a composite action. We've done that once before here: https://github.com/stellar/quickstart/blob/master/.github/actions/push/action.yml

2opremio:
Yes, that's exactly what I am trying out. And it seems to work just fine. I will factor out the rest of the code.

.github/actions/cache-go/action.yml
```diff
@@ -0,0 +1,17 @@
...
+    - uses: actions/cache@v2
+      with:
+        path: ~/go/pkg/mod
+        key: ${{ github.workflow }}-${{ github.job }}-${{ matrix.go }}-${{ runner.os }}-go-mod-${{ hashFiles('**/go.sum') }}
+        restore-keys: ${{ github.workflow }}-${{ github.job }}-${{ matrix.go }}-${{ runner.os }}-go-mod
+    - uses: actions/cache@v2
+      if: ${{ github.ref != 'master' }}
+      with:
+        path: ~/.cache/go-build
+        key: ${{ github.workflow }}-${{ github.job }}-${{ matrix.go }}-${{ runner.os }}-go-all-${{ github.ref }}-${{ hashFiles('**', '!.git') }}
```
leighmcculloch, line 14: @2opremio This key needs all matrix entries for each job, not just the `go` entry. We need caches to be separate for the version of Postgres in use, and any other parameters, otherwise the results can be shared even though the environment is different.

2opremio: We can pass those as parameters, in fact we can also factor out the checkout and the go installation to make it even more compact (renaming the compose action accordingly).

I am off for today, but feel free to give it a go.

leighmcculloch: FYI @ire-and-curses this is the risk that you asked about here https://github.com/stellar/go/pull/3727#issuecomment-1049148752: Go can't determine everything about our environment is correct. If we hope to test across two versions of Postgres, we need to make sure ourselves we aren't reusing a cache from the wrong Postgres.

2opremio:
Caching integration tests doesn't seem to work that well. They practically take the same time.

Before: https://github.com/stellar/go/runs/5308340934?check_suite_focus=true

After: https://github.com/stellar/go/runs/5315578032?check_suite_focus=true


This is probably due to the system calls in the integration tests: https://youtu.be/balVSY4Qrj0?t=1248

I am going to debug it with `GODEBUG=gocache{hash,test}=1` (see https://github.com/golang/go/issues/35301 ) and try to fix it.

2opremio:
After quite a bit of digging, it seems like the problem comes from the `stat` of directories from `/home/runner/work/go/go` to `/home/runner/work/go/go/services/horizon/internal/integration` changing from run to run.

(I wonder why the tests open those directories though)

2opremio:
> 1. The `stat(2)` changes from run to run.

This is simply how `git clone` works.

2opremio:
> > 1. The `stat(2)` changes from run to run.
> 
> This is simply how `git clone` works.

After using `git-restore-mtime` to set the modification times of the working directory after checkout, the unit tests are all cached between runs (like it happens locally), leading to`Go/test` job times of ~17s. The bottleneck of the `Go` workflow is now `Go/check`, which takes 2 minutes (@leighmcculloch do you think we could improve this? we should at least be able to run the check steps in parallel ).

Even with `git-restore-mtime` I am not able to cache integration tests (I am suspecting this is due to files being generated by Horizon). I will dig further.

2opremio:
> 2. Why the tests open those directories.

BTW, the fact that files like `/home/runner/work/go/go/services/horizon/internal/integration/captive-core-f3ca03d7/stellar-core.conf` explain this.

leighmcculloch:
>After using git-restore-mtime to set the modification times of the working directory after checkout, the unit tests are all cached between runs

Why is git-restore-mtime needed on unit tests? The unit tests were cached already when I tested them without changing the time.

2opremio:
> Why is git-restore-mtime needed on unit tests? The unit tests were cached already when I tested them without changing the time.

Some tests (like the ingest tests) didn't.

2opremio:
To elaborate further, those tests (which take up to 30s) read `services/horizon/internal/ingest/testdata/sample-changes.xdr`

There are other tests which weren't cached, but that was the most meaningful one.

2opremio:
This is finally ready! All tests, including integration tests, cache properly 🥳

2opremio:
Actually, I need to fix the mdate of files (fixing it will cause tests to be cached when fixtures change, which is not good!)

2opremio:
Ah, `git-restore-mtime` didn't work because GitHub actions clone the repo with `--depth=1` (thus it doesn't know the last modification time of files)

leighmcculloch:
>Actually, I need to fix the mdate of files (fixing it will cause tests to be cached when fixtures change, which is not good!)

Can you elaborate on this? Won't the contents of the file changing cause the cache key to be different anyway because we use GitHub's `hashFiles`?

Unfortunately using git-restore-mtimes binds the useful of this action to Ubuntu only, and creates a weird dependency on depth=0.

2opremio:
I am not sure I understand your comment. The cache key generated will be different, but we don't use the file hashes to restore the key:

```yaml
      restore-keys: |
        ${{ env.PREFIX }}-go-build-${{ github.ref }}
        ${{ env.PREFIX }}-go-build
```

Thus, you can restore the build cache of another run with different fixture contents. If both files have the same modification time, go will cache the test.

leighmcculloch:
Ah yes, good point.

leighmcculloch:
So what youre saying if the mtimes are the same, but file contents are different, Go will cache the test that reads the file? I thought it always used the file contents.

2opremio:
No, I don't think it does (otherwise I guess there wouldn't be a need to use modification times). See https://cs.opensource.google/go/go/+/refs/tags/go1.17.7:src/cmd/go/internal/test/test.go;l=1583-1586;drc=refs%2Ftags%2Fgo1.17.7;bpv=0;bpt=1

2opremio:
it turns out that `git-restore-mtime` doesn't change the mtime of some directories:

```
> "open /home/runner/work/go/go bb9598098875243c0dbd385a488fb5414ddfcd49f22602f81cf53ab3bfcd841c\n"
> "open /home/runner/work/go/go/services/horizon/internal 24e92446952b505a6dd3917d834b58f24e66d3f39e1bca77e5489ac11bbe3026\n"
> "stat /home/runner/work/go/go/.git 4a45c85e5426f757d802f36ed9b24761a2ee8e098f0bebbf5841636a22f1d660\n"
> "stat /home/runner/work/go/go/docs c21656fb0d652a1ed825aa3d8f6ff51fe7e4d96913887ba9124046fc5fc17743\n"
> "stat /home/runner/work/go/go/handlers 6cbcb56c1d83bb49c6127b72fc392957a73ce3e9287a87c25cece275c4740606\n"
> "stat /home/runner/work/go/go/services/horizon/internal/assets e8325ba2d8b8012ea7d9c76243cf502da96f1da1c152066aa2a91e9c39434de2\n"
> "stat /home/runner/work/go/go/services/horizon/internal/codes e8325ba2d8b8012ea7d9c76243cf502da96f1da1c152066aa2a91e9c39434de2\n"
> "stat /home/runner/work/go/go/services/horizon/internal/db2 50af35c929f6129e1d4633fc2ef9701e3c8f8885c855717cab98bcc9d3ea358f\n"
> "stat /home/runner/work/go/go/services/horizon/internal/errors 7a0c973de020bc0828080e6d689e800c6c1e27b12964d5b8d680d572b3babff9\n"
> "stat /home/runner/work/go/go/services/horizon/internal/reap eb634dba2c0cfc9db06be86a937d519eaf2194187a57fbaaba1709de76c1c0ab\n"
> "stat /home/runner/work/go/go/services/horizon/internal/render eb634dba2c0cfc9db06be86a937d519eaf2194187a57fbaaba1709de76c1c0ab\n"
> "stat /home/runner/work/go/go/services/horizon/internal/tls 038519f3ba2f0d1043a681d29b2a496f2ff508961dc77187e757b94150112b54\n"
> "stat /home/runner/work/go/go/services/horizon/internal/utf8 038519f3ba2f0d1043a681d29b2a496f2ff508961dc77187e757b94150112b54\n"
> "stat /home/runner/work/go/go/support 00b9d9e1980020205b34ef036b67bcdb5d92275aa6c6c8507add40535e7f2429\n"
> "stat /home/runner/work/go/go/toid 00b9d9e1980020205b34ef036b67bcdb5d92275aa6c6c8507add40535e7f2429\n"
> "stat /home/runner/work/go/go/tools 5c07b44606decb0197ce6806bb42bbafd7cf0847c494a9e6f8d5f0469825e116\n"
```

leighmcculloch:
Ah directories are not stored in git, so that makes sense that they wouldn't have an mtime stored in git to restore.

2opremio:
> Ah directories are not stored in git, so that makes sense that they wouldn't have an mtime stored in git to restore.

But `git-restore-mtime` seems to be able to restore the mtime of directories (inferred from the mtime of files), so that seems to be a bug.

.github/actions/setup-go/action.yml
```diff
@@ -18,9 +18,12 @@ runs:
   - name: Restore modification time of checkout files
     shell: bash
     run: |
-      # git-restore-mtime doesn't correct restore the mtime of some directories.
-      # to make up for it we set a base, fixed modification time of all files
-      touch -m -t '202201010101' $(find . -type d -not -path '.git/*')
+      # Set a base, fixed modification time of all directories.
+      # git-restore-mtime doesn't set the mtime of directories because git
+      # doesn't store directories, but changing mtimes of directories can
```
2opremio, line 23: That's not the reason, git-restore-mtime does update the mtime of directories (infered from the mtime of files).

leighmcculloch: Fixed comment. Thanks.

.github/actions/setup-go/action.yml
```diff
@@ -19,7 +19,8 @@ runs:
     shell: bash
     run: |
       # Set a base, fixed modification time of all directories.
-      # git-restore-mtime doesn't set the mtime of directories because git
+      # git-restore-mtime doesn't set the mtime of directories that contain only
+      # subdirectories because there's no files to infer an mtime and git
```
2opremio, line 23: That isn't the reason either 😆 , `services/horizon/internal/assets/` contains files and isn't properly restored

leighmcculloch: Interesting, the go restore-mtime changeling says its won't change directories with subdirectories, so it is at least some of the reason we won't see changes.

2opremio:
Alright! This should be ready to go!

leighmcculloch:
Awesome. Thanks for making this better and adding caching for integration tests.

I'm going to run some tests, then I'll merge this after I'm finished.

leighmcculloch:
@bartekn You already approved this, but I accidentally merged this before you reviewed the changes. Please feel free to leave any feedback.

PR #3737, opened 2021-07-01
services/horizon: require COUNT param for horizon db migrate commands

### LINKED ISSUE #3720: services/horizon: Make `migrate up/down [COUNT]` param mandatory to prevent wiping your db

Currently `horizon db migrate down` command accepts optional `[COUNT]` parameter. When not provided the command will default to all migrations and will start reverting them resulting in empty database.
This means that a user who, for example, forgets to add `--help` flat can accidentally wipe horizon data. Some commands, for example `horizon db`, will print help instead of performing actions so it's easy to incorrectly assume `horizon db migrate down` behaves in the same way.

Make the `[COUNT]` param mandatory so people know more about what they're doing.

It's not clear how to establish what the `[COUNT]` should be when downgrading horizon. An alternative approach could be version number that we want to go back to. For example when downgrading to `2.3.1` users could run:
```
horizon db migrate to 2.3.1
```
With above they don't have to know how many (if any at all) schema downgrades are required.

paulbellamy:
Fixes #3720

<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Require the `COUNT` param for `horizon db migrate down COUNT` (and `redo`, to prevent accidentally running all downward migrations. `up` is unchanged, so you can run all upwards migrations.

Add a `horizon db migrate status` command, printing migrations and the time they were applied. A short-term fix to make figuring out the count possible. Output like:
```sh
$ horizon db migrate status
Migration                                                       Applied
1_initial_schema.sql                                            2021-06-29 16:29:19.728113 +0000 UTC
2_index_participants_by_toid.sql                                2021-06-29 16:29:19.729313 +0000 UTC
3_use_sequence_in_history_accounts.sql                          2021-06-29 16:29:19.730471 +0000 UTC
4_add_protocol_version.sql                                      2021-06-29 16:29:19.731004 +0000 UTC
5_create_trades_table.sql                                       2021-06-29 16:29:19.732911 +0000 UTC
6_create_assets_table.sql                                       2021-06-29 16:29:19.734534 +0000 UTC
7_modify_trades_table.sql                                       2021-06-29 16:29:19.737889 +0000 UTC
8_add_aggregators.sql                                           2021-06-29 16:29:19.739113 +0000 UTC
...
```

### Why

It's a bit of a footgun at the moment. The default behaviour of `horizon db migrate down` is maximum destruction.

### Known limitations

- [ ] Figure out where to document this. (especially using `status` to figure out how many steps to migrate down). Suggestions welcome...


2opremio:
I think that for the `migrate status` command it would be really useful for each column to also print what `migrate down COUNT` command is needed to get there. I can create a separate ticket for that.

2opremio (approved):
Optionally, it would be useful to have a integration test for the migration commands.

paulbellamy:
Opened https://github.com/stellar/go/issues/3744 to track any improvements to `horizon db migrate status`

Issue #3193, opened 2020-11-04
Create tests for setting configuration values via CLI params and env variables.


bartekn:
Horizon 1.11.0 contained a bug preventing setting `--db-url` config via CLI params. We should create tests (integration tests) checking this. See: https://github.com/stellar/go/issues/3191

PR #3741, opened 2021-07-02
services/horizon: Refactors integration test framework to allow custom parameter testing.

Shaptic:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
This introduces the `IntegrationTest.Config.HorizonParameters` field, which allows you to customize the parameters passed to Horizon, and refactors the tests to use these accordingly. It also introduces `FatalTestSuite`, which is a way to run integration tests and see if they *should* fail.

**Note**: This is mostly intended a design proposal and an opportunity to get feedback before I dive into actual testing of the combinatorial explosion of parameters we have. There are a few tests and validators included to demonstrate how said tests would look.

### Why
In order to facilitate configuration testing (see #3193), we need to be able to run Horizon with arbitrary combinations of parameters.


ingest/ledgerbackend/toml.go
```diff
@@ -21,6 +21,10 @@ const (
 	// if LOG_FILE_PATH is omitted stellar core actually defaults to "stellar-core.log"
 	// however, we are overriding this default for captive core
 	defaultLogFilePath = "" // by default we disable logging to a file
+
+	// if DISABLE_XDR_FSYNC is omitted stellar core actually defaults to false
+	// however, we are overriding this default for captive core
+	defaultDisableXDRFsync = true
```
Shaptic, line 27: Ignore changes to this file: this is an out-dated merge from master, I think.

services/horizon/internal/test/integration/integration.go
```diff
@@ -189,76 +243,93 @@ func (i *Test) startHorizon(
...
+	var env strings.Builder
+	for key, value := range i.config.HorizonEnvironment {
+		env.WriteString(fmt.Sprintf("%s=%s", key, value))
+	}
+	i.t.Logf("Horizon environmental variables: %s\n", env.String())
+
+	// prepare env
 	cmd.SetArgs(args)
+	for key, value := range i.config.HorizonEnvironment {
+		os.Setenv(key, value)
```
tamirms, line 310: I think we need to restore the environment variables to what they were after stopping horizon

services/horizon/internal/test/integration/integration.go
```diff
@@ -70,64 +99,92 @@ func NewTestForRemoteHorizon(t *testing.T, horizonURL string, passPhrase string,
...
-	// Only run Stellar Core container and its dependencies
-	runComposeCommand("up", "--detach", "--quiet-pull", "--no-color", "core")
+	if value, ok := config.HorizonParameters[horizon.StellarCoreBinaryPathName]; ok {
+		i.coreConfig.binaryPath = value
+	}
+	if value, ok := config.HorizonParameters[horizon.CaptiveCoreConfigPathName]; ok {
+		i.coreConfig.configPath = value
+	}
 
+	if !bothOrNeitherString(i.coreConfig.binaryPath, i.coreConfig.configPath) {
```
tamirms, line 132: doesn't horizon enforce this when it parses the command line flags? If so maybe we can remove this code because there's no need to do the check twice

services/horizon/internal/test/integration/integration.go
```diff
@@ -70,64 +99,92 @@ func NewTestForRemoteHorizon(t *testing.T, horizonURL string, passPhrase string,
...
+		t:           t,
+		config:      config,
+		composePath: composePath,
+		passPhrase:  StandaloneNetworkPassphrase,
 	}
 
-	var captiveCoreBinaryPath, captiveCoreConfigPath string
+	// We either test Captive Core through environment variables, or through
+	// custom Horizon parameters.
 	if os.Getenv("HORIZON_INTEGRATION_ENABLE_CAPTIVE_CORE") != "" {
```
tamirms, line 120: now that we're using this check in multiple places, I think it would make sense to create a global variable:

```golang
var RunWithCaptiveCore = os.Getenv("HORIZON_INTEGRATION_ENABLE_CAPTIVE_CORE") != ""
```

services/horizon/internal/integration/parameters_test.go
```diff
@@ -0,0 +1,205 @@
...
+		horizon.StellarCoreDBURLFlagName:  "",
+	})
+	defer os.RemoveAll(STORAGE_PATH)
+
+	err := test.StartHorizon()
+	assert.NoError(t, err)
+	// FIXME: IntegrationTest needs a big refactor so we can properly wait for
+	//        Captive Core. This needs a lot more thought so this workaround is
+	//        an unfortunate temporary necessity...
+	time.Sleep(10 * time.Second)
```
tamirms, line 115: why is this necessary, doesn't `waitForHorizon()` which is called by `StartHorizon()` ensure that horizon is ingesting before it returns?

services/horizon/internal/integration/parameters_test.go
```diff
@@ -0,0 +1,205 @@
...
+	// This is a bit of a hacky workaround.
+	//
+	// In CI, we run our integration tests twice: once with Captive Core
+	// enabled, and once without. *These* tests only run with Captive Core
+	// configured properly (specifically, w/ the CAPTIVE_CORE_BIN envvar set).
+	if _, ok := os.LookupEnv("HORIZON_INTEGRATION_ENABLE_CAPTIVE_CORE"); !ok {
+		suite.T().Skip()
+	}
+
+	defer createCaptiveCoreConfig(
```
tamirms, line 76: this is a bit confusing because when I read `defer createCaptiveCoreConfig....` I assume that `createCaptiveCoreConfig()` is the operation that is getting deferred. I think it would be cleaner to make `createCaptiveCoreConfig()` write the file and not return anything. Then in the next line you could call `defer os.Remove("./captive-core.toml")`

services/horizon/internal/integration/parameters_test.go
```diff
@@ -0,0 +1,205 @@
...
+		"./captive-core.toml", BUCKET_DIR_DISALLOWED_TOML)()
+
+	const STORAGE_PATH string = "./test_no-bucket-dir"
+	test := NewParameterTest(suite.T(), map[string]string{
+		"captive-core-storage-path":       STORAGE_PATH,
+		horizon.CaptiveCoreConfigPathName: "./captive-core.toml",
+		horizon.StellarCoreBinaryPathName: os.Getenv("CAPTIVE_CORE_BIN"),
+	})
+	defer os.RemoveAll(STORAGE_PATH)
```
tamirms, line 85: does "./test_no-bucket-dir" exist? if not then calling `os.RemoveAll(STORAGE_PATH)` is not necessary

services/horizon/internal/integration/parameters_test.go
```diff
@@ -0,0 +1,205 @@
...
+
+	suite.Exits(func() { test.StartHorizon() })
```
tamirms, line 87: I don't understand what's happening here. Could you explain what suite.Exits is doing and why it's necessary to implement the control flow in this way?

services/horizon/internal/integration/parameters_test.go
paulbellamy: Or at least:
```go
cleanup := createCaptiveCoreConfig("./captive-core.toml", BUCKET_DIR_DISALLOWED_TOML)
defer cleanup()
```

services/horizon/internal/integration/parameters_test.go
paulbellamy: As I understand it, (correct me if I'm wrong @Shaptic) we need this because our config is injected via global variables, and ad-hoc places in the code fetching environment variables. So to run the app with different configs we have to launch a whole new subprocess...

If we had a more controlled way of passing config into horizon we could do something like:
```go
import "github.com/services/horizon/cmd"

c := cmd.RootCmd
c.SetEnv(map[string]string{"ENV_VARS_HERE": "1"}) // Not sure how to do this with cobra
c.SetArgs([]string{"db", "migrate", "up"})
err := c.Execute()
assert.NoError(t, c.Execute()) // or assert.Error(t, c.Execute(), "whatever we expect")
```

But yeah the control flow around the subprocess thing in `Exits` is quite weird either way.

services/horizon/internal/integration/parameters_test.go
```diff
@@ -0,0 +1,205 @@
...
+	assert.NoError(t, err)
+	// FIXME: IntegrationTest needs a big refactor so we can properly wait for
+	//        Captive Core. This needs a lot more thought so this workaround is
+	//        an unfortunate temporary necessity...
+	time.Sleep(10 * time.Second)
+
+	runParameterMatrix(test, []ValidatorFunc{
+		func() { validateCaptiveCoreDiskState(test, STORAGE_PATH) },
+		func() { validateNoBucketDirPath(test, STORAGE_PATH) },
+	})
```
paulbellamy, line 120: I'd do these as two separate tests, and extract the setup (which is the common part) instead of extracting the test (which is the unique part).

```go

// Should this be a tempdir?
const storagePath string = "./test_captive-core-works"

var defaultCaptiveCoreConfig = map[string]string{
  "captive-core-storage-path":       storagePath,
  "captive-core-reuse-storage-path": "true",
  horizon.StellarCoreBinaryPathName: os.Getenv("CAPTIVE_CORE_BIN"),
  horizon.CaptiveCoreConfigPathName: "./captive-core.toml",
  horizon.StellarCoreURLFlagName:    "",
  horizon.StellarCoreDBURLFlagName:  "",
}

func TestCaptiveCoreCreatesExpectedDirsOnLaunch(t *testing.T) {
  // ensure that running Captive Core creates a
  // sensible directory structure.
  //
  // Pass "rootDirectory" set to whatever it is you pass to
  // "--captive-core-storage-path".
    
  if _, ok := os.LookupEnv("HORIZON_INTEGRATION_ENABLE_CAPTIVE_CORE"); !ok {
    t.Skip() // explained above
  }
  
  cleanup := launchHorizon(t, defaultConfig)
  defer cleanup()

  storageDir := path.Join(rootDir, "captive-core")
  coreConf := path.Join(storageDir, "stellar-core.conf")
  
  assert.DirExists(t, rootDir)
  assert.DirExists(t, storageDir)
  assert.FileExists(t, coreConf)
}

func TestCaptiveCoreConfigExcludesBucketDirPathEntry(t *testing.T) {
  // ensure the Captive Core auto-generated configuration
  // file doesn't contain the BUCKET_DIR_PATH entry, which is forbidden when using
  // Captive Core.
  //
  // Pass "rootDirectory" set to whatever it is you pass to
  // "--captive-core-storage-path".
  
  if _, ok := os.LookupEnv("HORIZON_INTEGRATION_ENABLE_CAPTIVE_CORE"); !ok {
    t.Skip() // explained above
  }
  
  cleanup := launchHorizon(t, defaultConfig)
  defer cleanup()
  
  coreConf := path.Join(rootDir, "captive-core", "stellar-core.conf")
  assert.FileExists(t, coreConf)
  
  result, err := ioutil.ReadFile(coreConf)
  assert.NoError(t, err)
  
  isBucketPathSet := strings.Contains(string(result), "BUCKET_DIR_PATH")
  assert.False(t, isBucketPathSet)
}

// Paul: Or whatever you need to do to launch horizon here.
func launchHorizon(t *testing.T, config map[string]string) (func()) {
  // Paul: Should this be a temp file? Actually, why does this need to be a file?
  cleanupConfig := createCaptiveCoreConfig("./captive-core.toml", CAPTIVE_CORE_CONFIG_STATE_TOML)
  
  test := NewParameterTest(t, config)
  defer os.RemoveAll(storagePath)
  
  err := test.StartHorizon()
  assert.NoError(t, err)
  // FIXME: IntegrationTest needs a big refactor so we can properly wait for
  //        Captive Core. This needs a lot more thought so this workaround is
  //        an unfortunate temporary necessity...
  time.Sleep(10 * time.Second)
  
  // Paul: Do we need to kill horizon when the test finishes?
  return cleanupConfig
}
```

services/horizon/internal/test/integration/integration.go
```diff
@@ -42,14 +44,41 @@ type Config struct {
 	ProtocolVersion       int32
 	SkipContainerCreation bool
 	CoreDockerImage       string
+
+	// Weird naming here because bools default to false, but we want to start
+	// Horizon by default.
+	DontStartHorizon bool
```
paulbellamy, line 50: Nit/Suggestion:
```suggestion
	SkipHorizon bool
```
Above we called it `SkipContainerCreation`. Maybe there's a similar phrasing here? `SkipHorizonStart`?

services/horizon/internal/integration/parameters_test.go
```diff
@@ -0,0 +1,205 @@
...
+	storageDir := path.Join(rootDir, "captive-core")
+	coreConf := path.Join(storageDir, "stellar-core.conf")
+
+	tt.DirExists(rootDir)
+	tt.DirExists(storageDir)
+	tt.FileExists(coreConf)
+}
+
+func createCaptiveCoreConfig(filename, contents string) func() {
+	tomlFile, err := os.Create(filename)
```
paulbellamy, line 193: again could this be a tempdir or tempfile to avoid hardcoding paths into tests?

services/horizon/internal/test/integration/integration.go
```diff
@@ -42,14 +44,41 @@ type Config struct {
...
+
+	composePath string
+
 	config        Config
+	coreConfig    CaptiveConfig
 	horizonConfig horizon.Config
-	hclient       *sdk.Client
-	cclient       *stellarcore.Client
+
+	hclient *sdk.Client
```
paulbellamy, line 79: To match the above `horizonConfig`:
```suggestion
	horizonClient *sdk.Client
```

services/horizon/internal/test/integration/integration.go
```diff
@@ -42,14 +44,41 @@ type Config struct {
...
+	cclient *stellarcore.Client
```
paulbellamy, line 80:
```suggestion
	coreClient *stellarcore.Client
```

services/horizon/internal/test/integration/integration.go
```diff
@@ -70,64 +99,92 @@ func NewTestForRemoteHorizon(t *testing.T, horizonURL string, passPhrase string,
...
+		}
+
+		i.waitForHorizon()
+	}
+
+	return i
+}
+
+// Runs a docker-compose command applied to the above configs
+func (i *Test) runComposeCommand(args ...string) {
```
paulbellamy, line 157: 💯

services/horizon/internal/test/integration/integration.go
```diff
@@ -189,76 +243,93 @@ func (i *Test) startHorizon(
...
-		fmt.Sprintf("http://%s:%d", hostname, historyArchivePort),
-		"--ingest",
-		"--db-url",
-		horizonPostgresURL,
-		"--stellar-core-db-url",
-		fmt.Sprintf(
+	coreBinaryPath := i.coreConfig.binaryPath
+	captiveCoreConfigPath := i.coreConfig.configPath
+
+	var DEFAULT_ARGS = map[string]string{
```
paulbellamy, line 262:
```suggestion
	defaultArgs := map[string]string{
```

services/horizon/internal/test/integration/integration.go
```diff
@@ -189,76 +243,93 @@ func (i *Test) startHorizon(
...
-		"21626",
-
-		"--enable-captive-core-ingestion=" + strconv.FormatBool(len(captiveCoreBinaryPath) > 0),
-
-		"--network-passphrase",
-		i.passPhrase,
-		"--apply-migrations",
-		"--admin-port",
-		strconv.Itoa(i.AdminPort()),
+	merged := mergeMaps(DEFAULT_ARGS, i.config.HorizonParameters)
```
paulbellamy, line 286:
```suggestion
	merged := mergeMaps(defaultArgs, i.config.HorizonParameters)
```

services/horizon/internal/test/integration/integration.go
paulbellamy: Eugh yeah. modifying the global os env here sucks. Would be good to stop doing that, but definitely out of scope for this PR.

services/horizon/internal/test/integration/integration.go
```diff
@@ -302,6 +375,7 @@ func (i *Test) waitForCore() {
 			time.Sleep(time.Second)
 			continue
 		}
+		i.t.Log("Core is up.")
```
paulbellamy, line 378: 🆙 nice.

services/horizon/internal/test/integration/integration.go
```diff
@@ -302,6 +375,7 @@ func (i *Test) waitForCore() {
...
 		return
 	}
 	i.t.Fatal("Core could not sync after several attempts")
```
paulbellamy, line 381:
```suggestion
	i.t.Fatal("Core could not sync after 30s")
```

services/horizon/internal/test/integration/integration.go
```diff
@@ -660,3 +730,24 @@ func findDockerComposePath() string {
 	// Directly jump down to the folder that should contain the configs
 	return filepath.Join(current, "services", "horizon", "docker")
 }
+
+// mergeMaps returns a new map which contains the keys and values of both input
+// maps, preferring the latter newValues map if there are duplicate entries.
+func mergeMaps(defaults map[string]string, newValues map[string]string) map[string]string {
```
paulbellamy, line 736: Ironically it's even simpler in the more generic case... 😂 

```go
func mergeMaps(ms ...map[string]string) map[string]string {
  merged := map[string]string{}
  for _, m := range ms {
    for k, v := range m {
      merged[k] = v
    }
  }
  return merged
}
```

services/horizon/internal/integration/parameters_test.go
Shaptic: You're sorta there but not exactly, @paulbellamy.

The main problem is that Horizon dies via `log.Fatal` if the configuration options are invalid. Since the purpose of this framework is exactly to test whether or not Horizon treats invalid combinations as invalid (and that it treats valid combinations like it should), we basically need a way to say "did Horizon die as expected?" Since `log.Fatal` kills the process, the test also immediately dies. Thus, we need this (hacky, but functional) `suite.Exits()` approach to detect Horizon exiting _expectedly_.

services/horizon/internal/integration/parameters_test.go
Shaptic: Oh good tip: `StartHorizon` and `waitForHorizon` are called separately, so I didn't see the latter.

services/horizon/internal/integration/parameters_test.go
Shaptic: Yep, good points!

services/horizon/internal/integration/parameters_test.go
Shaptic: I think it looks this way because there aren't enough example tests yet, but the intent of the framework here is essentially to associate certain "validation functions" with each parameter. You grouped the parameters together into `defaultCaptiveCoreConfig`, but in fact this is supposed to be the unique part. Otherwise, we aren't actually testing the parameter combinations!

However, I do think there's merit to having this default, and then instead modifying only what we need within the test. I'll tinker with it and push my ideas.

services/horizon/internal/test/integration/integration.go
Shaptic: yep, hated my naming; great suggestion

services/horizon/internal/integration/parameters_test.go
Shaptic: great idea :+1:

services/horizon/internal/test/integration/integration.go
Shaptic: Clever! Love it

services/horizon/internal/integration/parameters_test.go
```diff
@@ -0,0 +1,205 @@
...
+	// enabled, and once without. *These* tests only run with Captive Core
+	// configured properly (specifically, w/ the CAPTIVE_CORE_BIN envvar set).
+	if _, ok := os.LookupEnv("HORIZON_INTEGRATION_ENABLE_CAPTIVE_CORE"); !ok {
+		suite.T().Skip()
+	}
+
+	defer createCaptiveCoreConfig(
+		"./captive-core.toml", BUCKET_DIR_DISALLOWED_TOML)()
+
+	const STORAGE_PATH string = "./test_no-bucket-dir"
```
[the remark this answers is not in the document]
Shaptic: :joy:

services/horizon/internal/test/integration/integration.go
Shaptic: wasn't a big change, so I introduced `EnvironmentManager` to help with this!

services/horizon/internal/test/integration/integration.go
Shaptic: you're right!

services/horizon/internal/integration/parameters_test.go
paulbellamy: Yeah, I'm saying that, in an ideal world, only place in the codebase that should do `log.Fatal`, or `os.Exit` is `func main()`

services/horizon/internal/integration/parameters_test.go
paulbellamy: Got it, ok. In that case, maybe we should use `t.Run("some name", ...)` to label them instead of the matrix function?

services/horizon/internal/integration/parameters_test.go
Shaptic: Sure, agreed! But that's a bigger refactor than is in scope here, unfortunately. 

The alternative is to run the entirety of Horizon as a subprocess rather than "natively" in a `go func() {}`; that refactor may actually be "better" in a way because it enables us to test subcommands, too (e.g. `migrate`).

services/horizon/internal/integration/parameters_test.go
Shaptic: Hm, seems reasonable. Done in 2d98e427.

paulbellamy: Cool. Looks better at least. As a nit, I'd probably inline `validateNoBucketDirPath`, and the other one (since they're only used in one place). But it's good enough for now.

services/horizon/internal/integration/parameters_test.go
paulbellamy: Nit, should this bit about log.Fatal and it dying be a comment on the `Exits` method?

paulbellamy (approved):
Definitely more refactoring we could do to make testing this simpler, but much better than we had, and more than good enough for now. 💯

services/horizon/internal/integration/parameters_test.go
Shaptic: Fair, but again the idea is that these would be reused in new tests that also wanted to confirm the state of e.g. `--captive-core-storage-path`.

PR #3743, opened 2021-07-05
services/horizon: Check for overlap error before preparing range in reingestion

### LINKED ISSUE #3638: services/horizon: Earlier warning about reingest overlapping with current ledger

When running `horizon db reingest START STOP`, it goes for a while, then I get:
```
ERRO[2021-05-27T15:13:13.310Z] Error in ingestion state machine              current_state="reingestHistoryRange(fromLedger=1171022, toLedger=1171121, force=false)" error="reingest range overlaps with horizon ingestion" next_state=stop pid=94970 service=ingest
INFO[2021-05-27T15:13:13.310Z] Shut down                                     pid=94970 service=ingest
2021/05/27 15:13:13
                        The range you have provided overlaps with Horizon's most recently ingested ledger.
                        It is not possible to run the reingest command on this range in parallel with
                        Horizon's ingestion system.
                        Either reduce the range so that it doesn't overlap with Horizon's ingestion system,
                        or, use the force flag to ensure that Horizon's ingestion system is blocked until
                        the reingest command completes.
 ```             

It would be nice if it warned me earlier, that this command wouldn't work. For bonus points, it would be great if it told me how to fix the command.

### PR DESCRIPTION
**Pull request by 2opremio**
Fixes #3638

### DIFF
--- a/services/horizon/internal/ingest/fsm.go
+++ b/services/horizon/internal/ingest/fsm.go
@@ -19,10 +19,17 @@ import (
 
 var (
 	defaultSleep = time.Second
-	// ErrReingestRangeConflict indicates that the reingest range overlaps with
-	// horizon's most recently ingested ledger
-	ErrReingestRangeConflict = errors.New("reingest range overlaps with horizon ingestion")
 )
+
+// ErrReingestRangeConflict indicates that the reingest range overlaps with
+// horizon's most recently ingested ledger
+type ErrReingestRangeConflict struct {
+	maximumLedgerSequence uint32
+}
+
+func (e ErrReingestRangeConflict) Error() string {
+	return fmt.Sprintf("reingest range overlaps with horizon ingestion, supplied range shouldn't contain ledger %d", e.maximumLedgerSequence)
+}
 
 type stateMachineNode interface {
 	run(*system) (transition, error)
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: services/horizon/internal/ingest/fsm.go [hunk 1]
```diff
@@ -725,7 +737,12 @@ func (h reingestHistoryRangeState) run(s *system) (transition, error) {
 		}
 
 		if lastIngestedLedger > 0 && h.toLedger >= lastIngestedLedger {
-			return stop(), ErrReingestRangeConflict
+			return stop(), ErrReingestRangeConflict{lastIngestedLedger}
+		}
+
+		// Only prepare the range after checking the bounds to enable an early error return
+		if t, err := prepareRange(); err != nil {
```
tamirms, line 744: don't we need to call `prepareRange()` in the `h.force` case as well?

services/horizon/internal/ingest/fsm.go
```diff
@@ -681,24 +688,29 @@ func (h reingestHistoryRangeState) run(s *system) (transition, error) {
...
+			"from":     h.fromLedger,
+			"to":       h.toLedger,
+			"duration": time.Since(startTime).Seconds(),
+		}).Info("Range ready")
 
-	startTime = time.Now()
+		startTime = time.Now()
+
+		return transition{}, nil
+	}
```
bartekn, line 713: I feel it would be simpler if we just move the condition before `PrepareRange` call, like:
```go
		if !h.force && lastIngestedLedger > 0 && h.toLedger >= lastIngestedLedger {
			return stop(), ErrReingestRangeConflict{lastIngestedLedger}
		}
```

services/horizon/internal/ingest/fsm.go
2opremio: Really good catch

services/horizon/internal/ingest/fsm.go
bartekn: > then it would apply to both the force and non-force

There's `!h.force` in the condition so it won't be executed in forced call.

2opremio: Right, I think it's less clear that way because it's disconnected from the rest of the `!h.force` branch.

services/horizon/internal/ingest/fsm.go
```diff
@@ -690,52 +697,66 @@ func (h reingestHistoryRangeState) run(s *system) (transition, error) {
 		h.fromLedger = 2
 	}
 
-	log.WithFields(logpkg.F{
-		"from": h.fromLedger,
-		"to":   h.toLedger,
-	}).Info("Preparing ledger backend to retrieve range")
-	startTime := time.Now()
+	var startTime time.Time
+	prepareRange := func() (transition, error) {
```
tamirms, line 701: I think it would be cleaner to move this into a `prepareRange()` function on the `reingestHistoryRangeState` struct instead of making it a closure

2opremio: Really good point, it's much cleaner now. Thanks.

tamirms (approved):

PR #3760, opened 2021-07-09
services/horizon: [v2.6.0] Add migration to rebuild trade aggregations when upgrading

paulbellamy:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add back in the migration to rebuild the trade aggregations when running normal db migrations.

### Why

This seems like the easiest and simplest way to make sure node operators don't need to reingest all of history.

### Known limitations

[TODO or N/A]


2opremio (approved):
Did you test this will the full History?

bartekn (approved):
LGTM! I have an idea how we can do it at migration stage but prevent code duplication. We can add the method call that rebuilds the buckets to `services/horizon/internal/db2/schema.Migrate` (`MigrateUp` case).

paulbellamy:
re: testing, yeah. When initially writing this, I ran it with full history, then ran the rebuild, and verified it generated the same data.

paulbellamy:
re code duplication, i was thinking that db migrations shouldn't really change anyway. So we shouldn't need to maintain this, but potentially add another migration later if we need to change the format or w/e.

2opremio (approved):

PR #3766, opened 2021-07-13
services/horizon: Purging log.Fatal, to make horizon CLI easier to test.

paulbellamy:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Remove `log.Fatal` and `os.Exit` calls from Horizon.
- `log.Fatal` and `os.Exit` should only be called in `func main()`, and `func init()`.
- Everything else which could fail returns an `error`, which propagates up to `func main()`

### Why

As discussed in https://github.com/stellar/go/pull/3741#discussion_r668142648, using `log.Fatal` and `os.Exit` make integration testing Horizon harder. They force you to launch it as a subprocess and faff around with managing that. Removing these *should* make it simpler.

Reducing friction for CLI-type tests would have helped catch https://github.com/stellar/go/pull/3762 type bugs sooner/easier.

### Known limitations

`[WIP]` at this point, as I want to get people's opinions, and if there are any gotchas to watch out for here before spending more time on this. If we're happy with the approach we should merge https://github.com/stellar/go/pull/3741 as-is, then use this to simplify it, and add a few more cli tests.

The one big outstanding thing here re: testing are the places we use `os.Getenv`. But.. One thing at a time.

### DIFF
--- a/services/horizon/cmd/db.go
+++ b/services/horizon/cmd/db.go
@@ -32,32 +32,34 @@ var dbMigrateCmd = &cobra.Command{
 	Short: "commands to run schema migrations on horizon's postgres db",
 }
 
-func requireAndSetFlag(name string) {
+func requireAndSetFlag(name string) error {
 	for _, flag := range flags {
 		if flag.Name == name {
 			flag.Require()
 			flag.SetValue()
-			return
+			return nil
 		}
 	}
-	log.Fatalf("could not find %s flag", name)
+	return fmt.Errorf("could not find %s flag", name)
 }
 
 var dbInitCmd = &cobra.Command{
 	Use:   "init",
 	Short: "install schema",
 	Long:  "init initializes the postgres database used by horizon.",
-	Run: func(cmd *cobra.Command, args []string) {
-		requireAndSetFlag(horizon.DatabaseURLFlagName)
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if err := requireAndSetFlag(horizon.DatabaseURLFlagName); err != nil {
+			return err
+		}
 
 		db, err := sql.Open("postgres", config.DatabaseURL)
 		if err != nil {
-			log.Fatal(err)
+			return err
 		}
 
 		numMigrationsRun, err := schema.Migrate(db, schema.MigrateUp, 0)
 		if err != nil {
-			log.Fatal(err)
+			return err
 		}
 
 		if numMigrationsRun == 0 {
@@ -65,18 +67,19 @@ var dbInitCmd = &cobra.Command{
 		} else {
 			log.Printf("Successfully applied %d migrations.\n", numMigrationsRun)
 		}
+		return nil
 	},
 }
 
-func migrate(dir schema.MigrateDir, count int) {
+func migrate(dir schema.MigrateDir, count int) error {
 	dbConn, err := db.Open("postgres", config.DatabaseURL)
 	if err != nil {
-		log.Fatal(err)
+		return err
 	}
 
 	numMigrationsRun, err := schema.Migrate(dbConn.DB.DB, dir, count)
 	if err != nil {
-		log.Fatal(err)
+		return err
 	}
 
 	if numMigrationsRun == 0 {
@@ -84,29 +87,32 @@ func migrate(dir schema.MigrateDir, count int) {
 	} else {
 		log.Printf("Successfully applied %d migrations.\n", numMigrationsRun)
 	}
+	return nil
 }
 
 var dbMigrateDownCmd = &cobra.Command{
 	Use:   "down COUNT",
 	Short: "run upwards db schema migrations",
 	Long:  "performs a downards schema migration command",
-	Run: func(cmd *cobra.Command, args []string) {
-		requireAndSetFlag(horizon.DatabaseURLFlagName)
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if err := requireAndSetFlag(horizon.DatabaseURLFlagName); err != nil {
+			return err
+		}
 
 		// Only allow invokations with 1 args.
 		if len(args) != 1 {
 			cmd.Usage()
-			os.Exit(1)
+			return ErrUsage
 		}
 
 		count, err := strconv.Atoi(args[0])
 		if err != nil {
 			log.Println(err)
 			cmd.Usage()
-			os.Exit(1)
+			return ErrUsage
 		}
 
-		migrate(schema.MigrateDown, count)
+		return migrate(schema.MigrateDown, count)
 	},
 }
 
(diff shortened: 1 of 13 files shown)


services/horizon/internal/flags.go
```diff
@@ -39,48 +39,60 @@ const (
...
 	migrationsToApplyUp := schema.GetMigrationsUp(config.DatabaseURL)
 	if len(migrationsToApplyUp) > 0 {
-		stdLog.Printf(`There are %v migrations to apply in the "up" direction.`, len(migrationsToApplyUp))
-		stdLog.Printf("The necessary migrations are: %v", migrationsToApplyUp)
-		stdLog.Printf("A database migration is required to run this version (%v) of Horizon. Run \"horizon db migrate up\" to update your DB. Consult the Changelog (https://github.com/stellar/go/blob/master/services/horizon/CHANGELOG.md) for more information.", apkg.Version())
-		os.Exit(1)
+		return fmt.Errorf(`
+		There are %v migrations to apply in the "up" direction.
+		The necessary migrations are: %v
+		A database migration is required to run this version (%v) of Horizon. Run "horizon db migrate up" to update your DB. Consult the Changelog (https://github.com/stellar/go/blob/master/services/horizon/CHANGELOG.md) for more information.
```
Shaptic, line 77: pedantic but there's preceding whitespace here now

support/config/config_option.go
```diff
@@ -148,32 +172,35 @@ func (co *ConfigOption) setFlag(cmd *cobra.Command) error {
 }
 
 // SetDuration converts a command line int to a duration, and stores it in the final config.
-func SetDuration(co *ConfigOption) {
+func SetDuration(co *ConfigOption) error {
 	*(co.ConfigKey.(*time.Duration)) = time.Duration(viper.GetInt(co.Name)) * time.Second
+	return nil
```
Shaptic, line 177: Prob missing something, but why change func sig if always `return nil`?

Shaptic: Oh it's to conform to the new `CustomSetValue`, huh

Shaptic (commented):
Bit of a Catch-22 here: can't write CLI tests without this refactor, can't test this refactor without CLI tests! Everything seems pretty straightforward aside from the `WaitGroup` changes, though. My vote is to get this in really early into a sprint so that we can all interact with it a bunch throughout.

support/config/config_option.go
paulbellamy: Yeah. Some of them return an error, unfortunately.

services/horizon/internal/app.go
```diff
@@ -382,19 +385,22 @@ func (a *App) DeleteUnretainedHistory(ctx context.Context) error {
...
 	log.Debug("ticking app")
 
 	// update ledger state, operation fee state, and stellar-core info in parallel
-	wg.Add(3)
-	go func() { a.UpdateLedgerState(ctx); wg.Done() }()
-	go func() { a.UpdateFeeStatsState(ctx); wg.Done() }()
-	go func() { a.UpdateStellarCoreInfo(ctx); wg.Done() }()
+	wg.Go(func() error { a.UpdateLedgerState(ctx); return nil })
+	wg.Go(func() error { a.UpdateFeeStatsState(ctx); return nil })
+	wg.Go(func() error { return a.UpdateStellarCoreInfo(ctx) })
```
2opremio, line 394: Nice!

2opremio:
I am in favor of this.

My only concern: could this have any negative operational implications? (e.g. the monitoring system not catching bad restarts anymore because the failure wasn't logged as expected)

paulbellamy:
Yeah, we'll need to be super careful not to change error output messages too much with this. Good call.

bartekn (approved):
LGTM

services/horizon/internal/app.go
```diff
@@ -415,10 +417,14 @@ func (a *App) Tick(ctx context.Context) error {
...
 	// update ledger state, operation fee state, and stellar-core info in parallel
 	wg.Add(3)
+	var err error
 	go func() { a.UpdateLedgerState(ctx); wg.Done() }()
 	go func() { a.UpdateFeeStatsState(ctx); wg.Done() }()
-	go func() { a.UpdateStellarCoreInfo(ctx); wg.Done() }()
+	go func() { err = a.UpdateStellarCoreInfo(ctx); wg.Done() }()
 	wg.Wait()
+	if err != nil {
+		return err
```
tamirms, line 426: what's the reasoning for returning early here? do we not want to execute the submitter go routine?

services/horizon/cmd/db.go
```diff
@@ -182,37 +195,37 @@ var dbMigrateUpCmd = &cobra.Command{
...
 var dbReingestCmd = &cobra.Command{
 	Use:   "reingest",
 	Short: "reingest commands",
 	Long:  "reingest ingests historical data for every ledger or ledgers specified by subcommand",
-	Run: func(cmd *cobra.Command, args []string) {
+	RunE: func(cmd *cobra.Command, args []string) error {
 		fmt.Println("Use one of the subcomands...")
 		cmd.Usage()
-		os.Exit(1)
+		return ErrUsage
```
Shaptic, line 228: Alternatively, we print usage when we get `ErrUsage` as a returned `error` value? Just a thought.

Shaptic (approved):
This will be great for param testing :heart_eyes:. Get someone else's :heavy_check_mark: though heh.

services/horizon/internal/app.go
paulbellamy: From: `UpdateStellarCoreInfo`
```go
	// Check if NetworkPassphrase is different, if so exit Horizon as it can break the
	// state of the application.
	if resp.Info.Network != a.config.NetworkPassphrase {
		return fmt.Errorf(
			"Network passphrase of stellar-core (%s) does not match Horizon configuration (%s). Exiting...",
			resp.Info.Network,
			a.config.NetworkPassphrase,
		)
	}
```
Previously it did a log.Fatal.

services/horizon/internal/app.go
```diff
@@ -372,9 +374,9 @@ func (a *App) UpdateFeeStatsState(ctx context.Context) {
 // UpdateStellarCoreInfo updates the value of CoreVersion,
 // CurrentProtocolVersion, and CoreSupportedProtocolVersion from the Stellar
 // core API.
-func (a *App) UpdateStellarCoreInfo(ctx context.Context) {
+func (a *App) UpdateStellarCoreInfo(ctx context.Context) error {
```
tamirms, line 377: it seems like UpdateStellarCoreInfo should only return an error if it's fatal. I think someone who edits this code later on could later modify it so that it returns non fatal errors (e.g. request to stellar core times out). Maybe add a comment here to make the expected behavior of UpdateStellarCoreInfo more explicit?

tamirms (approved):

PR #3768, opened 2021-07-13
all: reduce go.list to deps used during build

leighmcculloch:
### What
Shorten the list of dependencies in `go.list` to be dependencies that are used during the build of code and tests.

### Why
Our `go.list` file contains all dependencies in the graph, even dependencies that don't make it to code. It makes assessing the actual impact of upgrades like #3767 difficult.

### Known Limitations
Moving away from using the `-m` flag means tags apply. Any code behind a build tag won't be included. It's not trivial to use `go list` with multiple tags. There's an issue open at golang/go#42504 that will hopefully make it possible to run the go list command with all tags included.

Thanks to @bcmills for help on the Gophers Slack in figuring out what command to use to get the modules that are built.


.github/workflows/go.yml
```diff
@@ -10,7 +10,7 @@ jobs:
   check:
     strategy:
       matrix:
-        os: [ubuntu-latest]
+        os: [ubuntu-latest, windows-latest, macos-latest]
```
2opremio, line 13: was committing this intentional?

leighmcculloch: Thanks for catching that, I need to revert that. Otherwise the PR is ready for review.

leighmcculloch: Updated.

2opremio (approved):

Issue #3764, opened 2021-07-13
Running Horizon via Docker Compose no longer works on Linux.


Shaptic:
### What version are you using?
master

### What did you do?
Tried running a standalone network via docker compose:

```bash
cd $GOPATH/src/github.com/stellar/go/services/horizon/docker
./start.sh standalone
```

### What did you expect to see?
The 4 running containers.

### What did you see instead?
A docker error:

```
[long stack trace omitted]
 File "/usr/lib/python3.9/site-packages/docker/api/container.py", line 598, in create_host_config
    return HostConfig(*args, **kwargs)
  File "/usr/lib/python3.9/site-packages/docker/types/containers.py", line 338, in __init__
    raise host_config_incompatible_error(
docker.errors.InvalidArgument: "host" network_mode is incompatible with port_bindings
```

### More information
This new validation check appears [as early as January](https://forums.docker.com/t/docker-errors-invalidargument-host-network-mode-is-incompatible-with-port-bindings/103492). As the link implies, it doesn't make sense to have port bindings when using the host's networking, but docker never made this check previously.

I work around this by dropping all of the port mappings in the YAMLs I need, but it'd be nice to get a more... realistic fix and enable the Docker Compose workflow for all platforms again.

**Workarounds include:** separate YAMLs for Linux, manually dropping all port mappings, downgrading your docker version

PR #3769, opened 2021-07-15
services/horizon: Update Docker Compose flow to work with latest stable Stellar-Core

Shaptic:
### What
Updates the Docker-Compose YAML files to be compatible with the latest version of Stellar-Core.

### Why
Running Core+Horizon via the compose workflow will cause issues:
 - the core container will fail because `force scp` is deprecated
 - the upgrade container will use P15 which is outdated
 - [stellar/stellar-core:latest](https://hub.docker.com/layers/stellar/stellar-core/latest/images/sha256-d6b97102fe154afca3fb8e8e5d2f5976bc8c8a878f2d83905cbb577c3a75a96f?context=explore) defines an ENTRYPOINT which is incompatible with the compose `command`

### Known limitations
#3764 still applies.

### DIFF
--- a/services/horizon/docker/core-start.sh
+++ b/services/horizon/docker/core-start.sh
@@ -12,6 +12,9 @@ cat stellar-core.cfg
 stellar-core new-db
 
 if [ "$1" = "standalone" ]; then
+  # start a network from scratch
+  FLAGS="--wait-for-consensus"
+
   # initialize history archive for standalone network
   stellar-core new-hist vs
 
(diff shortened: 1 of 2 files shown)


Shaptic:
Upon further investigation, this is just one of a handful of issues facing the current Docker Compose situation. I think we need a deeper look & refactor into this (especially considering that we don't recommend running standalone Watcher Core yet... do exactly that here) than this mild adjustment.

Shaptic:
Reopened @ #3774 now that #3773 is in.

PR #3771, opened 2021-07-19
services/horizon/internal: Add metric to observe the age of response items in history endpoints

tamirms:


services/horizon/internal/actions/ledger.go
```diff
@@ -56,7 +56,13 @@ type LedgerByIDQuery struct {
...
 
 type GetLedgerByIDHandler struct {
-	LedgerState *ledger.State
+	LedgerState       *ledger.State
+	ResponseAgeMetric ResponseAgeMetric
+}
+
+// ResponseAgeMetric is a metric which tracks how old a response is compared to the most recently ingested ledger
+type ResponseAgeMetric interface {
+	ObserveLedgerAge(r *http.Request, ledger int)
```
bartekn, line 65:
```suggestion
	ObserveLedgerAge(r *http.Request, ledgerSequence uint32)
```

services/horizon/internal/actions/ledger.go
```diff
@@ -56,7 +56,13 @@ type LedgerByIDQuery struct {
 }
 
 type GetLedgerByIDHandler struct {
-	LedgerState *ledger.State
+	LedgerState       *ledger.State
+	ResponseAgeMetric ResponseAgeMetric
+}
+
+// ResponseAgeMetric is a metric which tracks how old a response is compared to the most recently ingested ledger
+type ResponseAgeMetric interface {
```
bartekn, line 64: Maybe we should move it to `main.go`?

services/horizon/internal/httpx/handler.go
```diff
@@ -216,6 +227,58 @@ func (handler pageActionHandler) renderPage(w http.ResponseWriter, r *http.Reque
...
+
+func (m responseAgeMetric) ObserveLedgerAgePage(r *http.Request, records []hal.Pageable) {
+	latestLedger := int(m.ledgerState.CurrentStatus().ExpHistoryLatest)
+	route := db.Route(r.Context())
+	for _, record := range records {
+		var ledgersAgo float64
+		ledger, err := ledgerFromPagingToken(record.PagingToken())
+		if err != nil {
+			log.Warnf("invalid paging token %v", record.PagingToken())
+		}
```
bartekn, line 257: I think we should just ignore invalid cursors. I'm afraid it will pollute the logs.

bartekn (approved):
It looks good! I wonder if we want to keep these metrics permanently? If not, we should probably merge into a new feature branch.

ire-and-curses:
We discussed this today in Horizon team meeting. We aren't sure whether we want these permanently in production, and if the number of metrics is a concern. We plan to deploy this PR to staging and run for a few days and take a look at the information before we make a call on permanent inclusion.

bartekn:
I cherry-picked this on top of `resource-age-analysis` branch and modified (https://github.com/stellar/go/commit/cb63378638337a9e00713257bc60e07c4d5bf465) it use logs instead. This way we can have a better understanding of the data and decide which statistic we want to use if/when merged into master.

services/horizon/internal/httpx/handler.go
tamirms: we shouldn't be generating any invalid cursors in the response. I would expect there to be no logs for this case. If we were using the paging token from the requests then I would agree with you

2opremio:
@bartekn should we keep this PR or is it OK to close it?

bartekn:
Let's close it. We obtained the information we needed in stg and I don't think we need this metric in prd.

PR #3773, opened 2021-07-19
services/horizon/docker: Remove NETWORK_MODE=host configuration for linux users

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Fixes https://github.com/stellar/go/issues/3764

Docker for linux recently implemented a fix which allows a container to access the host machine by setting the `extra_hosts` property https://stackoverflow.com/questions/31324981/how-to-access-host-port-from-docker-container/61424570#61424570

With that fix we can get rid of the `network_mode=host` configuration

### DIFF
--- a/services/horizon/docker/README.md
+++ b/services/horizon/docker/README.md
@@ -20,21 +20,6 @@ The script takes one optional parameter which configures the Stellar network use
 
 `./start.sh standalone` will run the containers on a private standalone Stellar network.
 
-## Set up a .env file
-
-Mac OS X and Windows users should create an [`.env`](https://docs.docker.com/compose/environment-variables/#the-env_file-configuration-option) file which consists of:
-
-`NETWORK_MODE=bridge`
-
-Linux users should also create an `.env` file. However, the contents of the file should look like:
-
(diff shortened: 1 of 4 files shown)


Shaptic:
**TL;DR:** This works, but is insufficient; we also need #3769. Thank you @tamirms!

With this fix, I can confirm that the error from #3764 goes away on Arch Linux. I can also confirm the following:

 - By itself, it isn't sufficient in getting all 4 containers _running_ in `standalone` mode.
 - However, when used alongside #3769, then all 4 _do_ run.
 - In `testnet` mode, it _is_ sufficient for getting the 2 containers going.

Under `standalone` mode, we have a new issue to overcome: the Horizon ingestion state machine fails to start.

```
Error in ingestion state machine

current_state="buildFromCheckpoint(checkpointLedger=0)" 
error="unexpected checkpointLedger value" 
next_state=start pid=1 service=ingest
```

However, this warrants a separate issue which I'll open.

Shaptic (approved):

PR #3783, opened 2021-07-26
services/horizon: Allow starting Horizon with minimal configuration

### LINKED ISSUE #3719: Remove unnecessary config params required to start fully functional Horizon instance.
To improve easy of use of Horizon @Shaptic suggested to decrease number of mandatory config params required to start fully functional Horizon. Below is a list of config params required to start ingesting instance with some ideas how we can change make them optional:
* `DATABASE_URL` - There's probably nothing we can do because Horizon requires a Postgres DB. We could potentially use a default `postgres://localhost:5432/horizon?sslmode=disable` but errors about invalid credentials can be more confusing than asking for setting this param.
* `INGEST` - This is currently `false` by default but we could make it `true` so starting Horizon without the param will always start ingestion.
* `STELLAR_CORE_URL` - The default ingestion is using Captive-Core so we could default to Captive-Core HTTP port.
* `STELLAR_CORE_BINARY_PATH` - We can require this param only if `stellar-core` binary is not available in $PATH.
* `CAPTIVE_CORE_CONFIG_PATH` - We discussed it internally and it's tempting to provide a default config file with tier 1 validators. There's a risk that people will never provide a production ready config file.
* `HISTORY_ARCHIVE_URLS` - The values of history archives could be extracted from `CAPTIVE_CORE_CONFIG_PATH`.
* `NETWORK_PASSPHRASE` - The value could be extracted from `CAPTIVE_CORE_CONFIG_PATH`.

To sum it up, it should be possible to start fully functional Horizon by only setting a single config param: `DATABASE_URL`. While it will definitely improve easy of use there are some risks connected to using default values in production. We should make sure we understand the risks before implementing this.

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This commit adds multiple changes to config parameters:
* `--ingest` is set to `true` by default,
* `--captive-core-config-path` is generated with default values based on the network passphrase.

If the pubnet Horizon is started without custom config, it will be stopped with an error message after one hour.

Close #3719.

### Why

We want to make the time from Horizon installation to execution to be minimal. With less config values and reasonable defaults developers can start Horizon instance faster.

### Known limitations

Quorum set default values for public network should be changed to to organisations developer really trust so Horizon will be stopped after one hour if an admin does not provide custom config. However, service managers like `systemd` may restart the service making it worth permanently.


ingest/ledgerbackend/toml.go
```diff
@@ -348,6 +349,264 @@ func NewCaptiveCoreTomlFromFile(configPath string, params CaptiveCoreTomlParams)
 	return &captiveCoreToml, nil
 }
 
+// NewDefaultTestnetCaptiveCoreToml constructs a new CaptiveCoreToml instance
+// based off the default testnet configuration.
+func NewDefaultTestnetCaptiveCoreToml() *CaptiveCoreToml {
+	var captiveCoreToml CaptiveCoreToml
```
Shaptic, line 355: Would it be easier to maintain these defaults if we used raw strings? 

For example, we could use [this config](https://github.com/stellar/stellar-core/blob/master/docs/stellar-core_testnet.cfg) for the testnet and just write it to / read it from an [`ioutil.TempFile`](https://pkg.go.dev/io/ioutil#TempFile). I think it'd be easier to maintain if the configs changed over time, since you could just re-copy-paste the config instead of trying to identify the delta and adapting it to our internal data structure.

services/horizon/internal/app.go
```diff
@@ -119,6 +119,21 @@ func (a *App) Serve() {
 	// configure shutdown signal handler
 	signalChan := make(chan os.Signal, 1)
 	signal.Notify(signalChan, os.Interrupt, syscall.SIGINT, syscall.SIGTERM)
+
+	if a.config.UsingDefaultPubnetConfig {
```
Shaptic, line 123: Sorry, I'm not seeing where this flag is being set?

Shaptic: Nvm it's line 549 in the `switch` below.

services/horizon/internal/flags.go
```diff
@@ -538,8 +539,18 @@ func ApplyFlags(config *Config, flags support.ConfigOptions, options ApplyOption
 
 			if config.RemoteCaptiveCoreURL == "" && (binaryPath == "" || config.CaptiveCoreConfigPath == "") {
 				if options.RequireCaptiveCoreConfig {
-					stdLog.Fatalf("Invalid config: captive core requires that both --%s and --%s are set. %s",
-						StellarCoreBinaryPathName, CaptiveCoreConfigPathName, captiveCoreMigrationHint)
+					switch config.NetworkPassphrase {
+					case network.TestNetworkPassphrase:
```
Shaptic, line 543: Should we add some `stdLog.Warn`s here?

Shaptic (commented):
This test seems like a great opportunity to add a new parameter test that confirms, for example, that the TOML file being used by Horizon matches the one we expect, or that it fails (i.e. `FatalTestCase.Exits()`) with a standalone network passphrase.

Whether or not this needs to be Horizon 3.0 (because of `--ingest=true` instead of `--ingest=false` by default), I'm still unsure about. You could argue that a breaking change is one that requires *more* parameters to get started rather than *less*...

Also we need to update the [deployment guide](https://developers.stellar.org/docs/run-api-server/configuring/#parameters) alongside this.

ingest/ledgerbackend/toml.go
tamirms: I think we should have the default pubnet and testnet validators represented in strong form. Then we can implement `NewDefaultTestnetCaptiveCoreToml()` and `NewDefaultPubnetCaptiveCoreToml()` by calling `newCaptiveCoreTomlFromFile(tomlFileContents string, params CaptiveCoreTomlParams) (*CaptiveCoreToml, error)`

tamirms:
@jacekn what is your opinion on this:

>Quorum set default values for public network should be changed to to organisations developer really trust so Horizon will be stopped after one hour if an admin does not provide custom config. However, service managers like systemd may restart the service making it worth permanently.

jacekn:
> @jacekn what is your opinion on this?
> 
> > Quorum set default values for public network should be changed to to organisations developer really trust so Horizon will be stopped after one hour if an admin does not provide custom config. However, service managers like systemd may restart the service making it worth permanently.

I think having hidden 1h delay after which horizon will stop will be confusing and, like you say, systemd might even hide it from operators making it worse. Personally I have never seen similar behaviour in linux software so I think it would be best to avoid it.

Horizon binary is likely wrong place to store default quorumset too. ubuntu packages we build ship with default configs already so I think it would be better to leverage that mechanism instead. We can do the same with docker images and we'll achieve the same goal. The benefit of doing that using package manager is that it's consistent with other linux software and operators should be familiar with the process. It also means that if quorumset changes are needed there no need to compile and rollout new horizon binary.

services/horizon/internal/flags.go
bartekn: I'm not sure, using default config in testnet is not risky because we're not operating on real assets.

bartekn:
I think there's one misconception regarding this property. This is really for development only to allow devs to start Horizon for the first time quickly. Creating a custom captive core config is actually a huge barrier of entry especially for people new to Stellar and we want to mimic the behaviour of `stellar/quickstart` in which you only define which network you want to start. In production devs should switch to custom config and 1h restarts are there to make sure they notice that something is not working properly.

> I think having hidden 1h delay after which horizon will stop will be confusing and, like you say, systemd might even hide it from operators making it worse. Personally I have never seen similar behaviour in linux software so I think it would be best to avoid it.

I agree that maybe it's not visible enough but I didn't want it to be hidden - we print `level=error` message on startup and after 1h right before terminating.

> Horizon binary is likely wrong place to store default quorumset too. ubuntu packages we build ship with default configs already so I think it would be better to leverage that mechanism instead. We can do the same with docker images and we'll achieve the same goal. The benefit of doing that using package manager is that it's consistent with other linux software and operators should be familiar with the process. It also means that if quorumset changes are needed there no need to compile and rollout new horizon binary.

As explained in the first paragraph, feature is mostly for devs (often not using Linux) who download the binary from GH and want to start development quickly. Both ubuntu packages and docker images already have the default config (and the one here is actually copied from `stellar/quickstart`).

@stellar/horizon-committers let me know what you think about terminating Horizon with default config after 1h.

ingest/ledgerbackend/toml.go
bartekn: After @jacekn comment and given that this feature will be used probably only by users who use GH release maybe instead of keeping the file in the binary we could add testnet and pubnet config files in the packed `.tar.gz` release file. Then we can change the code to try to load the default file from disk if the value is not explicitly set.

jacekn:
> I think there's one misconception regarding this property. This is really for development only to allow devs to start Horizon for the first time quickly. Creating a custom captive core config is actually a huge barrier of entry especially for people new to Stellar and we want to mimic the behaviour of `stellar/quickstart` in which you only define which network you want to start. In production devs should switch to custom config and 1h restarts are there to make sure they notice that something is not working properly.

I see. Maybe there is a middle ground that could work for both use cases? Having hardcoded testnet config feels like less or a problem, especially if we don't stop after 1h.
With regards to pubnet maybe there is a good way to make it easy to generate the config for developers? Maybe an error message saying something similar to "Pubnet quorumset missing. You can download latest recommended config curl https://github.com/stellar/go/xxx" ? If we make the process simple and clear enough I think maybe it will be acceptable?

bartekn:
Regarding pubnet I also proposed this: https://github.com/stellar/go/pull/3783#discussion_r679053721.

ingest/ledgerbackend/toml.go
ire-and-curses: I like that idea

services/horizon/internal/flags.go
ire-and-curses: I don't think it matters for testnet (also there isn't much choice of quorum)

Shaptic:
A comfortable middle ground (gentler nudging of the user) could be to do something like `Warn()` about default config use every hour.

ingest/ledgerbackend/toml.go
ire-and-curses: I prefer the file idea to the curl idea because curl'd endpoints can change or go away (or be compromised), so less reproducible.

jacekn: > I prefer the file idea to the curl idea because curl'd endpoints can change or go away (or be compromised), so less reproducible.

If you use curl it would be best to put the files in a location controlled by SDF, possibly in the go repo next to horizon source code. That way security profile won't change, we'd essentially get the same level of security as for the horizon source code itself.

bartekn: I just pushed a change adding two default config files to release archive here: https://github.com/stellar/go/pull/3783/commits/8b85c497aa5f837b46d4b060f53463ba91a8663d.

services/horizon/internal/flags.go
```diff
@@ -538,8 +540,37 @@ func ApplyFlags(config *Config, flags support.ConfigOptions, options ApplyOption
...
+					default:
+						stdLog.Fatal(errorMessage)
+					}
+
+					executablePath, err := os.Executable()
+					if err != nil {
+						stdLog.Fatal(errorMessage)
+					}
+
+					config.CaptiveCoreConfigPath = filepath.Join(filepath.Dir(executablePath), configFileName)
```
tamirms, line 568: should we check if the file exists first?

bartekn: It is checked in `NewCaptiveCoreTomlFromFile` but agree that checking it first will make the error message clear. Added in e1c4cc65b5.

tamirms (approved):

bartekn:
@tamirms I can't see the release branch yet so please merge if you feel you'll be able to add it to the next release. If not, I'll merge it next week.

PR #3786, opened 2021-07-27
all: update jwt-go

crossle:
Use `github.com/golang-jwt/jwt` replace `github.com/dgrijalva/jwt-go`

because https://github.com/advisories/GHSA-w73w-5m7g-f7qc


leighmcculloch:
Thanks @crossle for opening this PR.

`github.com/dgrijalva/jwt-go` is only used in tests in this repository so there is no immediate consequence to not upgrading this dependency, however we should update it to keep a vuln out of the dependency graph.

According to https://github.com/dgrijalva/jwt-go/issues/462 the module has moved to `golang-jwt/jwt`, and that matches the fix details in https://github.com/advisories/GHSA-w73w-5m7g-f7qc as well that @crossle linked to above.

I'll do a pass over the changes being imported, then approve, test, and merge.

leighmcculloch (approved):

PR #3787, opened 2021-07-28
all: add ci job to close stale issues and prs

leighmcculloch:
### What

Add a CI job to comment on stale issues and pull requests after 30 days of inactivity and close them 30 days after that, so 60 days total.

### Why

There are a lot of stale issues. We've had success with using this same pattern to clean up the stellar-protocol repo.

I've noticed that @2opremio routinely checks in on long lived pull requests to see if they should be closed. We don't have a huge amount of open pull requests so there isn't as big of an issue with them as we have with issues, but checking in on long lived pull requests is something a bot can do rather than it being on a human on the team.

### Known limitations

In the stellar-protocol repo we had some labels that clearly signaled an issue shouldn't be closed, and so we configured the stale checker to ignore those issues. In this repo there is no clear label that signals that issue should stay around, so there are no issues/prs exempt from the check.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,21 @@
+name: 'Stale Issues'
+
+on:
+  schedule:
+    - cron: '0 18 * * *' # approx 9:30am daily
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v3
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          debug-only: false
+          days-before-stale: 30
+          days-before-close: 30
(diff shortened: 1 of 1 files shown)


marcelosalloum (approved):
LGTM!

ire-and-curses:
I'd rather not do this. We're aware that we need to do a better job of triaging our issues on this repo. It's something we're ramping up on as we hire more people. Auto-closing issues is not helpful when there is backlog work here that we would actually like to do.

ire-and-curses (changes requested):
I'd rather not do this. We're aware that we need to do a better job of triaging our issues on this repo. It's something we're ramping up on as we hire more people. Auto-closing issues is not helpful when there is backlog work here that we would actually like to do.

PR #3795, opened 2021-08-02
services/horizon/internal: Warn about change for ingest flag

tamirms:
@jacekn has advised that we warn about the new default for `--ingest` and give some time for horizon maintainers to update their configurations.

Depends on https://github.com/stellar/go/pull/3794

### DIFF
--- a/services/horizon/internal/flags.go
+++ b/services/horizon/internal/flags.go
@@ -388,11 +388,19 @@ func Flags() (*Config, support.ConfigOptions) {
 			Usage:     "TLS private key file to use for securing connections to horizon",
 		},
 		&support.ConfigOption{
-			Name:        "ingest",
-			ConfigKey:   &config.Ingest,
-			OptType:     types.Bool,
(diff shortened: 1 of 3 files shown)


jacekn (commented):
I love this! Very good and clear warning message

services/horizon/internal/flags.go
```diff
@@ -391,8 +391,15 @@ func Flags() (*Config, support.ConfigOptions) {
 			Name:        "ingest",
 			ConfigKey:   &config.Ingest,
 			OptType:     types.Bool,
-			FlagDefault: true,
-			Usage:       "causes this horizon process to ingest data from stellar-core into horizon's db",
+			FlagDefault: false,
```
bartekn, line 394: Maybe we should add Action Needed tag somewhere here:
```suggestion
			// Action needed in release: horizon-v2.9.0: make --ingest default true
			FlagDefault: false,
```

bartekn (approved):

PR #3797, opened 2021-08-02
horizon: Publish verify-range image in master branch

### LINKED ISSUE #2761: Improve verify-range testing

### LINKED ISSUE #3796: Automatically publish verify-range image to docker hub
`stellar/verify-range` image can now be published to docker hub similarly to `stellar/horizon`.

2opremio:
Fixes #2761 
Fixes #3796

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -374,32 +374,26 @@ jobs:
             docker build -f services/horizon/docker/Dockerfile.dev -t stellar/horizon:$VERSION .
             docker push stellar/horizon:$VERSION
 
-  publish_verify_range_docker_image:
-    docker:
-      - image: docker:18.04-git
-    steps:
-      - checkout
-      - setup_remote_docker
-      - run:
-          name: Build, test and Push Verify Range Docker image
-          command: |
-            docker build -f services/horizon/docker/verify-range/Dockerfile -t stellar/horizon-verify-range services/horizon/docker/verify-range/
-            # Any range should do for basic testing, this range was chosen pretty early in history so that it only takes a few mins to run
-            docker run -ti -e BRANCH=${CIRCLE_SHA1} -e FROM=10000063 -e TO=10000127 stellar/horizon-verify-range
-            docker push stellar/horizon-verify-range
-
   test_verify_range_docker_image:
+    parameters:
+      publish:
+        type: boolean
+        default: false
     docker:
       - image: docker:18.04-git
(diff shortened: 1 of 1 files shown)


bartekn (commented):
There's a actually a lot of code duplication. Maybe we could use params (see `enable-captive-core` param)? Then if, say, `publish-image` is `true` we just run `docker publish` command, everything else stays the same.

.circleci/config.yml
```diff
@@ -471,12 +478,17 @@ workflows:
       - test_horizon_integration:
           name: test_horizon_integration_with_captive_core
           enable-captive-core: true
-      - test_verify_range_docker_image
+      - test_verify_range_docker_image:
+          # test pusblishing
+          publish: true
```
bartekn, line 483: We don't want to publish on non-master, correct?
```suggestion
          publish: false
```

bartekn (approved):

.circleci/config.yml
2opremio: I'm testing that it works before merging

2opremio: But ... it seems like the secrets are not there (probably due to this being a PR from my fork).

I will revert it and test from `master`

2opremio:
@bartekn PTAL

bartekn (approved):

Issue #3755, opened 2021-07-09
Horizon: DB Runway: design DB benchmark


2opremio:
- [x] Figure out if postgres has io metrics and table metrics. Are those useful for our benchmarking? pgbench?
- [x] Decide if we should include reingestion or not?
- [x] Find/Make a representative workload
- [x] Make a way to get a big representative DB
   - [ ] Copy from the last year - COPY INTO query?
   - [x] Reingestion might also be viable
   - [x] If we wanted random new data, could use Fuzzer

paulbellamy:
# Should we include reingestion or not?

I think not, because:
- It is a constant-ish amount of data coming in (based on ledger size and rate).
- It is only processed on the write-only db node, so shouldn't affect query workloads.
- Since the incoming data will be quite small compared to the entire database size, replication load should be comparatively small.

paulbellamy:
# Representative Workload & Benchmarking

First: Let's sample some logs from prod at different times of day, and feed them into [siege](https://github.com/JoeDog/siege).
  - simple, and quick to set up. Might be good enough.
  - siege lets us use a list of urls to hit, and mixes through them, so is a good fit for mixed workloads
  - To Measure: throughput/latency of requests. Distribution. DB load.
 

Next: If horizon turns out to be the bottleneck, we can log the db queries horizon is making, and feed them into pgbench.
 - Cuts out horizon, so it is more isolated to just the db load
 - Less realistic


Last: If we the response-time/load/etc distribution is not even, or we want to do more digging, add more in-depth postgres metrics to find bottlenecks.

paulbellamy:
# Ingesting a big DB

Conveniently, the current db is 4TB (a power of 2), so we could double it and run tests each time to see how it scales.

To account for changing network usage, we *could* try to filter out more recent data, and clone that. But that seems like a pain, and wouldn't account for the network usage changing again in the future. The more recent data is bigger anyway, so if we clone all existing data, it should be more varied, while still being biased towards recent network usage.

## Options for Loading & Cloning Data

### Restore from backup?

- From talking with ops, the restore takes a few hours, but the db can take several days afterwards to become performant. Seems like that would make it difficult to trust benchmarks.
- Wouldn't help us get a db bigger than we have now.

### Reingest

- Reingest takes ~3 days, but performance is good right after.
- 3 days / 4TB = 18hr/TB, 64TB*18hr = 48 days, assuming linear. 🐌 

### Fuzzer

[...]

PR #3799, opened 2021-08-03
Add alb-replay tool

2opremio:
`alb-replay` replays the successful GET requests found in an AWS ALB log file

It will be used as a way to benchmark the Horizon DB, as part of #3755


tools/alb-replay/main.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+}
+
+func queryURLs(timeout time.Duration, urlChan chan NumberedURL, stop chan struct{}) {
+	client := http.Client{
+		Timeout: timeout,
+	}
+	for numURL := range urlChan {
+		select {
+		case <-stop:
```
paulbellamy, line 128: Would a `case <-ctx.Done():` be more idiomatic here?

paulbellamy (approved):
one minor suggestion, but lgtm.

tools/alb-replay/main.go
```diff
@@ -0,0 +1,203 @@
...
+	parsed, err := url.Parse(urlStr)
+	if err != nil {
+		return "", fmt.Errorf("error parsing url %q: %v", urlStr, err)
+	}
+
+	return parsed.RequestURI(), nil
+}
+
+func parseURLs(startFrom int, baseURL string, logReader *ALBLogEntryReader, urlChan chan NumberedURL, stop chan struct{}) {
+	counter := 1
```
tamirms, line 91: should `counter` start at 0 given that you increment counter before the `if counter < startFrom` check?

tools/alb-replay/main.go
2opremio: Much cleaner now, thanks.

tools/alb-replay/main.go
```diff
@@ -0,0 +1,203 @@
...
+		wg.Done()
+	}()
+
+	// setup interrupt cleanup code
+	c := make(chan os.Signal, 2)
+	signal.Notify(c, os.Interrupt, syscall.SIGTERM)
+	go func() {
+		<-c
+		close(stop)
+		wg.Wait()
```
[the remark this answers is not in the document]
2opremio: true, you are right.

tools/alb-replay/main.go
2opremio: Yeah, good point.

PR #3801, opened 2021-08-03
horizon: Respect the --connection-timeout flag in path-finding endpoints

2opremio:
~Apart from adding the flag,~ this change also propagates the request context down to the DFS code, evaluating the deadline in every recursive call.

### DIFF
--- a/exp/orderbook/dfs.go
+++ b/exp/orderbook/dfs.go
@@ -69,11 +69,9 @@ func dfs(
 	currentAsset xdr.Asset,
 	currentAssetAmount xdr.Int64,
 ) error {
-	select {
-	case <-ctx.Done():
-		return context.DeadlineExceeded
-	default:
-		// do not block
+	// exit early if the context was cancelled
+	if err := ctx.Err(); err != nil {
+		return err
 	}
 	if currentAssetAmount <= 0 {
 		return nil
--- a/exp/orderbook/dfs.go
+++ b/exp/orderbook/dfs.go
@@ -1,6 +1,8 @@
 package orderbook
 
 import (
+	"context"
+
 	"github.com/stellar/go/price"
 	"github.com/stellar/go/xdr"
 )
@@ -58,6 +60,7 @@ type searchState interface {
 }
 
 func dfs(
+	ctx context.Context,
 	state searchState,
 	maxPathLength int,
 	visited map[string]bool,
@@ -66,6 +69,12 @@ func dfs(
 	currentAsset xdr.Asset,
 	currentAssetAmount xdr.Int64,
 ) error {
+	select {
+	case <-ctx.Done():
+		return context.DeadlineExceeded
+	default:
+		// do not block
+	}
 	if currentAssetAmount <= 0 {
 		return nil
 	}
@@ -103,6 +112,7 @@ func dfs(
 		}
 
 		err = dfs(
+			ctx,
 			state,
 			maxPathLength,
 			visited,
--- a/services/horizon/internal/flags.go
+++ b/services/horizon/internal/flags.go
@@ -290,6 +290,14 @@ func Flags() (*Config, support.ConfigOptions) {
 			Usage:          "defines the timeout of connection after which 504 response will be sent or stream will be closed, if Horizon is behind a load balancer with idle connection timeout, this should be set to a few seconds less that idle timeout, does not apply to POST /transactions",
 		},
 		&support.ConfigOption{
+			Name:           "path-find-timeout",
+			ConfigKey:      &config.PathFindTimeout,
+			OptType:        types.Int,
+			FlagDefault:    20,
+			CustomSetValue: support.SetDuration,
+			Usage:          "defines the timeout for path-finding requests (/paths/*) after which 504 response will be sent",
+		},
+		&support.ConfigOption{
 			Name:        "per-hour-rate-limit",
 			ConfigKey:   &config.RateQuota,
 			OptType:     types.Int,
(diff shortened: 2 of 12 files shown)


services/horizon/internal/flags.go
```diff
@@ -289,6 +289,14 @@ func Flags() (*Config, support.ConfigOptions) {
 			CustomSetValue: support.SetDuration,
 			Usage:          "defines the timeout of connection after which 504 response will be sent or stream will be closed, if Horizon is behind a load balancer with idle connection timeout, this should be set to a few seconds less that idle timeout, does not apply to POST /transactions",
 		},
+		&support.ConfigOption{
+			Name:           "path-find-timeout",
+			ConfigKey:      &config.PathFindTimeout,
+			OptType:        types.Int,
+			FlagDefault:    20,
```
Shaptic, line 296: Do we care about bubbling the option all the way up to the user?

Shaptic (commented):
This is similar to how I foresaw this change after realizing we could check in `dfs`, but much more elegant :+1:. I won't :heavy_check_mark: until we think of a way to test this but the code makes sense to me.

exp/orderbook/dfs.go
```diff
@@ -66,6 +69,12 @@ func dfs(
 	currentAsset xdr.Asset,
 	currentAssetAmount xdr.Int64,
 ) error {
+	select {
+	case <-ctx.Done():
+		return context.DeadlineExceeded
```
bartekn, line 74:
```suggestion
		return ctx.Err()
```

bartekn: ...or we can rewrite select to:
```go
if err := ctx.Err(); err != nil {
	return err
}
```

services/horizon/internal/httpx/router.go
```diff
@@ -184,11 +185,12 @@ func (r *Router) addRoutes(config *RouterConfig, rateLimiter *throttled.HTTPRate
 			MaxAssetsParamLength: maxAssetsForPathFinding,
 			PathFinder:           config.PathFinder,
 		}}
-
-		r.With(stateMiddleware.Wrap).Method(http.MethodGet, "/paths", findPaths)
-		r.With(stateMiddleware.Wrap).Method(http.MethodGet, "/paths/strict-receive", findPaths)
-		r.With(stateMiddleware.Wrap).Method(http.MethodGet, "/paths/strict-send", findFixedPaths)
-
+		r.Route("/paths", func(r chi.Router) {
+			r.Use(timeoutMiddleware(config.PathFindTimeout))
```
bartekn, line 189: I think that the _global_ timeout middleware will cancel the request first. We probably need to add some logic there to ignore `/paths`. I think it may also be good to have a single global timeout (so no `PathFindTimeout` param at all).

bartekn (approved):

exp/orderbook/dfs.go
2opremio: Much nicer, thanks

services/horizon/internal/httpx/router.go
2opremio: I am not sure I follow. What do you mean by the request being cancelled first?

bartekn: We have a [middleware](https://github.com/stellar/go/blob/90666abbac1d2c1a571eed390377ec9bb1fc42e7/services/horizon/internal/httpx/middleware.go#L92) that adds a context timeout on all requests' context. So I think this middleware will trigger the cancel first.

2opremio: Sure, but the shortest timeout will dominate the context deadline

2opremio: (i e. I don't see the problem)

services/horizon/internal/flags.go
2opremio: I personally think it should be exposed to give control to the user

services/horizon/internal/httpx/router.go
bartekn: Not if the global timeout is smaller than paths timeout. Ex. horizon.stellar.org timeout is set to 10s, default paths is set to 20s so paths will be canceled after 10s not 20s.

2opremio: But I think that's totally fine. If the global timeout is shorter, the request should be cancelled earlier and the `context.WithTimeout()` call wouldn't have any effect

2opremio: (i.e. that's how contexts are supposed to work)

2opremio: I had a discussion with @bartekn offline and the conclusions were

> 1. timeout doesn’t work for paths and I want it to work - it’s a bug
> 2. i’m fine if /paths just respect global timeout (ex. of 10s in horizon.stellar.org)
> 3. @tamirms  wanted 20s timeout for paths

As it stands right now (with the current code in this PR) you cannot have a longer timeout for paths (since they are also affected by the global connection timeout). We need to decide whether we want the path endpoints to be excluded from the global connection timeout so that they can have a longer timeout.

services/horizon/internal/flags.go
2opremio: We decided against it in the weekly meeting

services/horizon/internal/httpx/router.go
2opremio: During the Horizon meeting we decided against having a dedicated flag for the path-finding endpoints timeout. I have updated the PR to simply respect the `--connetion-timeout` flag in the path-finding endpoints.

tamirms (approved):

PR #3823, opened 2021-08-12
services/horizon: Reap in batches of 100k ledgers per second, to play nicely with others

### LINKED ISSUE #3819: services/horizon: Throttle the reaper

When people first turn on the reaper, it monopolizes their CPU.

Add some sort of limit to how fast the reaper runs so that it runs slower, but leaves cpu/ram available for other processes.

e.g. Throttle the reap at 2x what’s needed to keep up with new stuff. Might be tricky depending on how it batches.

It does a delete all. Maybe we could add a limit.

- Leave it and warn people

paulbellamy:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Limit reaper to 100k ledgers/second..

### Why

This should keep it running well under an hour (370seconds currently), but stop it from using all available CPU, starving other processes.

Fixes #3819 

### Known limitations

- [x] TODO: Actually launch it, and see it work now.

### DIFF
--- a/services/horizon/internal/reap/system.go
+++ b/services/horizon/internal/reap/system.go
@@ -70,28 +70,49 @@ func (r *System) runOnce(ctx context.Context) {
 	}
 }
 
+// Work backwards in 100k ledger blocks to prevent using all the CPU.
+//
+// This runs every hour, so we need to make sure it doesn't
+// run for longer than an hour.
+//
+// Current ledger at 2021-08-12 is 36,827,497, so 100k means 368 batches. At 1
+// batch/second, that seems like a reasonable balance between running well
+// under an hour, and slowing it down enough to leave some CPU for other
+// processes.
+var batchSize = int32(100_000)
+var sleep = 1 * time.Second
+
 func (r *System) clearBefore(ctx context.Context, seq int32) error {
 	log.WithField("new_elder", seq).Info("reaper: clearing")
 
-	start, end, err := toid.LedgerRangeInclusive(1, seq-1)
-	if err != nil {
-		return err
-	}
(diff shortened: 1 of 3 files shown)


services/horizon/internal/reap/system.go
```diff
@@ -70,28 +70,49 @@ func (r *System) runOnce(ctx context.Context) {
...
+var sleep = 1 * time.Second
+
 func (r *System) clearBefore(ctx context.Context, seq int32) error {
 	log.WithField("new_elder", seq).Info("reaper: clearing")
 
-	start, end, err := toid.LedgerRangeInclusive(1, seq-1)
-	if err != nil {
-		return err
-	}
+	for endSeq := seq - 1; endSeq >= 1; endSeq -= batchSize {
```
tamirms, line 88: we should have a better lower bound than 1, I think we can use `latest.HistoryElder` from `DeleteUnretainedHistory ()` as the lower bound and pass that into `clearBefore()`

services/horizon/internal/reap/system.go
```diff
@@ -70,28 +70,49 @@ func (r *System) runOnce(ctx context.Context) {
...
+		if err != nil {
+			return errors.Wrap(err, "Error in DeleteRangeAll")
+		}
+
+		err = r.HistoryQ.Commit()
+		if err != nil {
+			return errors.Wrap(err, "Error in commit")
+		}
+
+		time.Sleep(sleep)
```
tamirms, line 115: should we log whenever we complete a batch?

tamirms (approved):

PR #3827, opened 2021-08-17
all: update supported go versions to 1.17 and 1.16

leighmcculloch:
### What

Update supported Go versions to 1.17 and 1.16.

### Why

Go 1.15 is no longer supported. Go 1.17 has been released.

### Known limitations

N/A


go.mod
```diff
@@ -30,7 +30,6 @@ require (
 	github.com/inconshreveable/mousetrap v1.0.0 // indirect
 	github.com/jarcoal/httpmock v0.0.0-20161210151336-4442edb3db31
 	github.com/jmoiron/sqlx v1.2.0
-	github.com/k0kubun/colorstring v0.0.0-20150214042306-9440f1994b88 // indirect
```
leighmcculloch, line 33: Go 1.16 and 1.17 narrow the definition of what needs to be listed in the go.mod file. The Go 1.16 rules apply because we specify `go 1.16` at the top of the file.

golist.sh
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
 
-go list -deps -test -f '{{with .Module}}{{.Path}} {{.Version}}{{end}}' ./... | LC_ALL=C sort -u
+go list -f '{{with .Module}}{{.Path}} {{.Version}}{{end}}' all | LC_ALL=C sort -u
```
leighmcculloch, line 3: Go 1.17 narrows the definition of `all` so that it is equivalent and using `all` here is the recommended approach for getting this info now.

ire-and-curses (approved):

services/horizon/internal/test/integration/integration.go
```diff
@@ -213,7 +213,7 @@ func (i *Test) prepareShutdownHandlers() {
 	// stopped even if ingestion or testing fails.
 	i.t.Cleanup(i.Shutdown)
 
-	c := make(chan os.Signal)
+	c := make(chan os.Signal, 1)
```
leighmcculloch, line 216: This change is to fix a go vet error. It's unsafe to use an unbuffered channel with signal.Notify because the runtime does an unbuffered send to the channel, so if the interrupt happens to arrive between the setup and the consumption the signal can be lost. I don't think this has any critical impact because this appears to be in integration tests and we call `<-c` so soon after setting up the notify there's only a tiny window. Fixing none-the-less to make the go vet error go away.

More details are available at https://github.com/golang/go/issues/45043.

leighmcculloch: @ire-and-curses You already approved, but FYI ☝🏻 additional change required.

support/render/problem/default_test.go
```diff
@@ -29,19 +29,19 @@ func TestRender(t *testing.T) {
 			500,
 		}, {
 			"renders the type correctly",
-			P{Type: "foo"},
+			P{Type: "foo", Status: 500},
 			[]string{"foo"},
-			0,
+			500,
```
leighmcculloch, line 34: The actual response writer used by the stdlib http.Server has never supported zero status codes, however in Go 1.16 and earlier the httptest.ResponseRecorder did. In Go 1.17 the recorder was updated to also error, so that it has the same behavior as the real response writer. We use the recorder in some tests and some tests avoided setting the status code to a non-zero value. For Go 1.17 they need updating. More details at https://github.com/golang/go/issues/45353.

2opremio:
@leighmcculloch I think we need to change the repo configuration so that it doesn't block on `check_code_1_16`  and `test_code_1_15` ?

leighmcculloch:
@2opremio Yup I'm waiting on @ire-and-curses to merge that change now.

Issue #3810, opened 2021-08-09
Ingest liquidity pool trades


tamirms:
Horizon already has a database table which stores all trades occurring on the Stellar DEX. We will need to also include trades occurring on liquidity pools.

AMM trades differ from orderbook trades in the following ways:

- The price fraction will represent its numerator and denominator as 64 bit integers.
- AMM trades do not refer to offer ids (e.g. base_offer_id or counter_offer_id)
- Only one side of the trade will be executed by an account (either base_account or counter_account). The other side of the trade will be executed against a liquidity pool (either base_pool_id or counter_pool_id).

Here is an example response of an AMM trade:
```
{
  "id": "157235845014249474-0",
  "paging_token": "157235845014249474-0",
  "ledger_close_time": "2021-07-29T21:10:53Z",
  "trade_type": "liquidity_pool",
  "base_liquidity_pool_id": "abcdef",
  "liquidity_pool_fee_bp": 30,
  "base_amount": "0.0002007",
  "base_asset_type": "native",
  "counter_account": "GDW634JZX3VMEF2RZTCJTT34RITIMNX46QOGTYHCJEJL3MM7BLOQ6HOW",
  "counter_amount": "0.0022300",
  "counter_asset_type": "credit_alphanum4",
  "counter_asset_code": "VZT",
  "counter_asset_issuer": "GBENYXZDFFR2J4F4DB3YPBBAM244TXYOTIOOUQI5DBT3OKUU4ZJ2M7NO",
  "base_is_seller": false,
  "price": {
    "n": “10000000”,
    "d": “899997”
  }
}
```

Tasks:

- [x]  Update trades table to support storing AMM trades
- [ ]  Allow queries to find all trades for a given pool id
- [ ] Allow queries to find all trades by trade_type
- [x] Update trades processor to ingest AMM trades
- [ ] Add a [fuzzing](https://github.com/stellar/go/blob/master/services/horizon/internal/ingest/processors/offers_processor_test.go#L22-L54) test if possible
- [x] Make sure that any changes to the trades schema will not break trade aggregations

bartekn:
There are also some skipped tests. Probably part of #3812.

tamirms:
all those tasks will be implemented in https://github.com/stellar/go/issues/3835

Issue #3835, opened 2021-08-23
Include liquidity pool trades in trades endpoint


tamirms:
https://github.com/stellar/go/issues/3810 will implement ingestion of liquidity pool trades. After that is complete we want to include liquidity pool trades in the trades response.

We will add an extra query parameter to the trades endpoint, `trade_type`, which defaults to `all`.

`GET /trades?trade_type={orderbook,liquidity_pools,all}`

The `trade_type` filter will work in addition to the base and counter asset pair filters. However, Horizon will reject requests which attempt to set `trade_type=liquidity_pools` when using the offer id filter.

The JSON definition of a trade will contain an extra field trade_type which indicates whether the trade occurred against the DEX or on a liquidity pool.

**Orderbook Trade JSON**

The price fraction will represent its numerator and denominator values as strings to be consistent with the liquidity pool trade JSON object.

{
  "id": "157235845014249474-0",
  "paging_token": "157235845014249474-0",
  "ledger_close_time": "2021-07-29T21:10:53Z",
  "offer_id": "738916928",
  "trade_type": "orderbook",
  "base_offer_id": "4768921863441637378",
  "base_account": "GAPP4CS5KSZSHF6A24JWZ2W6MZKPCQKCAW44Y7FFX46WS76WMBDYWKWH",
  "base_amount": "0.0002007",
  "base_asset_type": "native",
  "counter_offer_id": "738916928",
  "counter_account": "GDW634JZX3VMEF2RZTCJTT34RITIMNX46QOGTYHCJEJL3MM7BLOQ6HOW",
  "counter_amount": "0.0022300",
  "counter_asset_type": "credit_alphanum4",
  "counter_asset_code": "VZT",
  "counter_asset_issuer": "GBENYXZDFFR2J4F4DB3YPBBAM244TXYOTIOOUQI5DBT3OKUU4ZJ2M7NO",
  "base_is_seller": false,
  "price": {
    "n": “10000000”,
    "d": “899997”
  }
}

**Liquidity Pool Trade JSON**

The price fraction will represent its numerator and denominator values as strings because they will be 64 bit integers.
Fields that refer to offer ids (e.g. base_offer_id or counter_offer_id) are removed from the response object.
Only one side of the trade will be executed by an account (either base_account or counter_account). The other side of the trade will be executed against a liquidity pool (either base_liquidity_pool_id or counter_liquidity_pool_id).

```
{
  "id": "157235845014249474-0",
  "paging_token": "157235845014249474-0",
[...]

PR #3857, opened 2021-08-26
services/horizon: Ingest Liquidity Pool Trades

tamirms:
Fixes https://github.com/stellar/go/issues/3810

This PR implements ingestion the db functionality required for ingestion. I still need to implement functions and tests to query liquidity pool trades. The remaining functionality will be implemented in https://github.com/stellar/go/issues/3835


services/horizon/internal/ingest/processors/trades_processor.go
```diff
@@ -35,101 +33,166 @@ func (p *TradeProcessor) ProcessTransaction(ctx context.Context, transaction ing
...
+	}
+
+	var poolMap map[string]int64
+	poolMap, err = p.tradesQ.CreateHistoryLiquidityPools(ctx, poolIDs, maxBatchSize)
+	if err != nil {
+		return errors.Wrap(err, "Error creating pool ids")
+	}
+
+	for _, trade := range p.trades {
+		row := trade.row
```
bartekn, line 85: Isn't `row` a copy? So all updates below will update a copy?

services/horizon/internal/ingest/processors/trades_processor.go
```diff
@@ -35,101 +33,166 @@ func (p *TradeProcessor) ProcessTransaction(ctx context.Context, transaction ing
...
+	if err := key.SetLiquidityPool(poolID); err != nil {
+		return 0, errors.Wrap(err, "Could not create liquidity pool ledger key")
+
+	}
 
-	if !found {
-		return price, errors.Wrap(err, "could not find change for trade offer")
+	change, err := p.findOperationChange(transaction, opidx, key)
+	if err != nil {
+		return 0, errors.Wrap(err, "could not find change for trade offer")
```
bartekn, line 170:
```suggestion
		return 0, errors.Wrap(err, "could not find change for liquidity pool")
```

services/horizon/internal/ingest/processors/trades_processor.go
```diff
@@ -190,20 +253,39 @@ func (p *TradeProcessor) extractTrades(
...
+
+			var sellerAccount, liquidityPoolID string
+			if trade.Type == xdr.ClaimAtomTypeClaimAtomTypeLiquidityPool {
+				id := trade.MustLiquidityPool().LiquidityPoolId
+				liquidityPoolID = PoolIDToString(id)
+				var fee uint32
+				if fee, err = p.findPoolFee(transaction, opidx, id); err != nil {
+					return nil, err
+				}
+				row.LiquidityPoolFee = null.IntFrom(int64(fee))
```
bartekn, line 280: I can't find setting `row.*LiquidityPoolID` fields. Should we do it here?

bartekn (approved):
LGTM! When reading the code I started wondering if it's possible that the trade against LP is successful but the LP doesn't change (ex. <1 stroop trade) so the meta changes are not generated for it. If it's true then there are several places where we rely on this. It's not clear from CAP so we should check with Stellar-Core team.

services/horizon/internal/ingest/processors/trades_processor.go
tamirms: @bartekn that is done in `Commit()`. In this function we gather all the liqudity pool id strings. Then in commit we map the liquidity pool id strings to their int ids from the `history_liquidity_pools` table. Once we have the mapping we set the `row.*LiquidityPoolID` fields with the int id values

services/horizon/internal/ingest/processors/trades_processor.go
tamirms: that's ok because in the end we are calling on `batch.Add()` on `row`. `trade.row` is never used anywhere

PR #3865, opened 2021-08-27
xdr: Enforce asset ordering in xdr pool

paulbellamy:
<details>
  
### PR Structure

* [ ]
  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Enforce asset ordering in `xdr.NewPoolId`.

### Why

To try to catch potential bugs early.

### Known limitations

[TODO or N/A]


xdr/pool_id.go
```diff
@@ -8,6 +8,10 @@ import (
 )
 
 func NewPoolId(a, b Asset, fee Int32) (PoolId, error) {
+	if b.LessThan(a) {
+		return PoolId{}, errors.New("asset A must be < asset B")
```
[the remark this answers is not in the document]
paulbellamy: I figured, that case wouldn't lead to a mistake or an error, would it? I guess the pool wouldn't exist though.. 🤔

xdr/pool_id.go
```diff
@@ -8,6 +8,10 @@ import (
...
+	}
+
```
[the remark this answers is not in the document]
paulbellamy: Yeah, I had a brief play with this earlier, but didn't find a satisfactory API in go. Will give it another "go". 😂

txnbuild/liquidity_pool_deposit.go
```diff
@@ -18,6 +20,51 @@ type LiquidityPoolDeposit struct {
...
+// implicitly ordering assets to generate the correct pool id.
+func NewLiquidityPoolDeposit(
+	sourceAccount string,
+	assetA Asset,
+	maxAmountA string,
+	assetB Asset,
+	maxAmountB string,
+	minPrice,
+	maxPrice string,
+) (LiquidityPoolDeposit, error) {
```
paulbellamy, line 34: Not sure this is the cleanest go signature for this. The other option would be having something like:
```go
d, err := NewLiquidityPoolDeposit(
  sourceAccount,
  // Not sure if there's a better name here. Could use `DepositAsset`, but
  // then there'd have to be an identical `WithdrawalAsset`... In past apps I've called
  // this asset+amount combination a `Money`, or just an `Amount`, but neither feels
  // quite appropriate here..
  AssetAmount{assetA, maxAmountA},
  AssetAmount{assetB, maxAmountB},
  // ... something with minPrice/maxPrice here...
  // @jonjove suggested structs with numerator/denominator, but I expect that'll be painful to use.
  // Could also keep minPrice/maxPrice as strings, as below.
  minPrice,
  maxPrice,
)
```

xdr/pool_id.go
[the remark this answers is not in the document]
paulbellamy: Ok, there's a couple new API proposals here https://github.com/stellar/go/pull/3865#discussion_r699496188

Thoughts there?

txnbuild/liquidity_pool_deposit.go
tamirms: @paulbellamy I have a slight preference for the `AssetAmount` option. We use the same pattern in the protocols  package to define the horizon response structs https://github.com/stellar/go/blob/amm/protocols/horizon/base/main.go#L14

tamirms: I agree though that `AssetAmount` is a little bit weird as a name. but I can't really think of anything better. Maybe `AssetAndAmount` ?

leighmcculloch: >AssetAmount

I think this name is accurate, simple, and easy to understand. 👍🏻 

>Could also keep minPrice/maxPrice as strings

I think this looks better. I would stick with strings since that's what the txnbuild API uses for amounts. I realize looking at this it's not really clear to me how max and min prices and max amounts work together to form a deposit, but that's {

txnbuild/asset.go
```diff
@@ -127,6 +131,33 @@ func (ca CreditAsset) GetCode() string { return ca.Code }
...
+	}
+
+	if ca.GetCode() < other.GetCode() {
+		return true
+	} else if other.GetCode() < ca.GetCode() {
+		return false
+	}
+
+	return ca.GetIssuer() < other.GetIssuer()
+}
```
bartekn, line 159: Sorry to be late to review party. I can't really find the exact place where this is implemented in Core but I think it compares binary representations of both assets. If that's true I think there's a bug here. Credit assets have two types: `alphanum4` and `alphanum12` so we should also check the type first.

I think we should just mimic what Stellar-Core does to avoid issues like this. We can just marshal both assets to binary form and compare them.

bartekn: OK, I wasn't aware https://github.com/stellar/stellar-protocol/pull/1037 existed. Checking it first...

txnbuild/asset.go
```diff
@@ -127,6 +131,33 @@ func (ca CreditAsset) GetCode() string { return ca.Code }
...
+		return false
+	}
+
+	if ca.GetCode() < other.GetCode() {
+		return true
+	} else if other.GetCode() < ca.GetCode() {
+		return false
+	}
+
+	return ca.GetIssuer() < other.GetIssuer()
```
Shaptic, line 158: I'm a little confused as to why this code is near-duplicated above (in `xdr.Asset.LessThan`)? Is it not possible to do something akin to `return ca.xdrAsset.LessThan(other.xdrAsset)`?

xdr/pool_id_test.go
```diff
@@ -8,6 +8,18 @@ import (
...
+	b, err := hex.DecodeString(hexKey)
+	require.NoError(t, err)
+	var key Uint256
+	copy(key[:], b)
+	a, err := NewAccountId(PublicKeyTypePublicKeyTypeEd25519, key)
+	require.NoError(t, err)
+	addr, err := a.GetAddress()
+	require.NoError(t, err)
+	return addr
+}
```
Shaptic, line 21: I think this could be simplified via `AddressToAccountId` from xdr/account_id.go.

xdr/asset.go
```diff
@@ -431,3 +431,47 @@ func (a Asset) ToTrustLineAsset() TrustLineAsset {
...
+	}
+
+	if a.GetCode() < b.GetCode() {
+		return true
+	} else if b.GetCode() < a.GetCode() {
+		return false
+	}
+
+	return a.GetIssuer() < b.GetIssuer()
+}
```
Shaptic, line 477: I've seen this logic a handful of times now, and each time I've had to triple-check it for correctness. Obviously, that's also what tests are for, but maybe there's a better way to code the logic. I think the following is equivalent:

```suggestion
func (a *Asset) LessThan(b Asset) bool {
	if int32(a.Type) == int32(b.Type) {
		if a.GetCode() == b.GetCode() {
			return a.GetIssuer() < b.GetIssuer()
		}
		return a.GetCode() < b.GetCode()
	}
	return int32(a.Type) < int32(b.Type)
}
```
Harder to understand, or easier?

xdr/pool_id_test.go
paulbellamy: Probably, but these tests are copied from core, to make sure we get the exact same serialization.

txnbuild/asset.go
bartekn: OK, it looks like we should check the credit asset type as I said before. From https://github.com/stellar/stellar-protocol/pull/1037:
```
if isUnion(T)
        if u.discriminant == v.discriminant
            // body means the contents corresponding to this value of the
            // discriminant
            return lessThan(u.body, v.body)
        else
            return u.discriminant < v.discriminant // <--
```

txnbuild/liquidity_pool_deposit.go
```diff
@@ -18,6 +20,49 @@ type LiquidityPoolDeposit struct {
...
+			return LiquidityPoolDeposit{}, errors.Errorf("cannot parse minPrice: %v", minPrice)
+		}
+		minPrice = min.Inv(min).FloatString(7)
+
+		// Invert the maxPrice. amountA/amountB -> amountB/amountA
+		max := &big.Rat{}
+		if _, ok := max.SetString(maxPrice); !ok {
+			return LiquidityPoolDeposit{}, errors.Errorf("cannot parse maxPrice: %v", minPrice)
+		}
+		maxPrice = max.Inv(max).FloatString(7)
```
bartekn, line 48: Given there can be rounding error after the price inversion I wonder if we should simply return an error here.

paulbellamy: yeah, was debating that as well...

txnbuild/asset.go
bartekn: <facepalm> I was reviewing the old version. New one looks good!

txnbuild/liquidity_pool_deposit.go
paulbellamy: Per discussion, this opens a can of worms. Let's just go back to throwing an error if the assets are out-of-order. That is the safest option, matches the JS SDK, and shouldn't happen often IRL, as the assets returned from horizon will be in-order.

xdr/pool_id_test.go
Shaptic: Good point, plus it's a hex address (as opposed to the base64 `G...` addresses) which I didn't realize.

PR #3872, opened 2021-08-30
Fix operation ordering when querying by claimable balance and liq pools

2opremio:
It seems like the ordering was broken for claimable balances as well


bartekn (approved):
LGTM! Is it possible to add a quick assert in one of existing tests?

PR #3887, opened 2021-09-02
services/horizon: Add Liquidity Pools to asset stats

### LINKED ISSUE #3838: Add liquidity pool shares and total amount to asset stats
We should update `/assets` with liquidity pools shares and total amounts so users have quick access to total supply of any issued asset.

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This commit adds Liquidity Pools to asset stats.

Close #3838.

### Why

Assets stats should include all data from all ledger entries so the sum of balances equals the sum of circulating supply of an issued asset.

### Known limitations

[TODO or N/A]


services/horizon/internal/ingest/processors/asset_stats_set.go
```diff
@@ -203,6 +217,76 @@ func (s AssetStatSet) AddTrustline(change ingest.Change) error {
...
+		return ingest.NewStateError(errors.New("both pre and post liquidity pools cannot be nil"))
+	}
+
+	var lpType xdr.LiquidityPoolType
+	if pre != nil {
+		lpType = pre.Body.Type
+	}
+	if post != nil {
+		lpType = post.Body.Type
+	}
```
Shaptic, line 246: Sanity check: can both pre and post be set? If so, this block will overwrite `lpType`.

tamirms (approved):

services/horizon/internal/ingest/processors/asset_stats_set.go
bartekn: Both can be set but it doesn't matter because the type cannot be changed. We basically want to extract ledger entry type from the change. I guess this should be a method on `ingest.Change`.

Issue #3836, opened 2021-08-23
Integrate liquidity pools in pathfinding endpoint


tamirms:
The interface for trading against a liquidity pool is the path payment operation.  

Therefore Horizon will need to update its path finding endpoint so that, for each step in the path, the exchange will be routed to the order book or liquidity pool that yields the best price.

To implement this we will need to:

- [x] (https://github.com/stellar/go/pull/3898) Add a deleted column to the `liquidity_pools` table 
- [x] (https://github.com/stellar/go/pull/3898) Periodically prune deleted rows similar to what we do for offers
- [x] (https://github.com/stellar/go/pull/3909) Implement something similar to OrderBookStream which polls AMM pool state rows found in the DB and updates in the memory AMM pool state
- [x] (@Shaptic) Update order book graph algorithm to examine in memory AMM pool state  (more details about the algorithm can be found in [CAP-38](https://stellar.org/protocol/cap-38#pathpaymentstrictsendop-and-pathpaymentstrictreceiveop))

PR #3888, opened 2021-09-03
exp/orderbook: Add helper to simulate exchanges against liquidity pools.

Shaptic:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Introduces a helper function

```go
func makeTrade(asset xdr.Asset, deposit int64, pool xdr.LiquidityPoolEntry) (uint64, error)
```

to calculate the payout from an LP if you make an exchange against it.

**Benchmark result** (on my machine):

```
$ go test ./orderbook/... -bench=.
goos: linux
goarch: amd64
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
BenchmarkSingleLiquidityPoolExchanges-12         	1000000000	          0.0000060 ns/op
BenchmarkLiquidityPoolExchanges-12               	  2120850	        558.8 ns/op
BenchmarkAlternativeLiquidityPoolExchanges-12    	  1000000	       1201 ns/op
PASS
ok  	github.com/stellar/go/exp/orderbook	2.979s
```

(where `BenchmarkAlternativeLiquidityPoolExchanges` introduces a custom `Uint128` type, which leverages both `math/bits` and `math/big` in a [failed] effort to optimize certain operations.)

### Why
We need this in order to support liquidity pools in the path payment algorithm, see also #3836.

### Known limitations
`staticcheck` will fail until `makeTrade` is actually used somewhere.


exp/orderbook/graph.go
```diff
@@ -478,3 +480,131 @@ func sortAndFilterPaths(
...
+	quotient := tempA.Quo(tempA, tempB)
+	payout, accuracy := quotient.Int64() // floors
+	isOutOfRange := ((payout == math.MinInt64 && accuracy == big.Above) ||
+		(payout == math.MaxInt64 && accuracy == big.Below))
+
+	return xdr.Int64(payout), !isOutOfRange && payout >= 0
+}
+
+// copyPoolState returns a duplicate of the given pool entry.
+func copyPoolState(pool xdr.LiquidityPoolEntry) (xdr.LiquidityPoolEntry, error) {
```
tamirms, line 590: since `makeTrade()` will be used by the path finding code we want to make it efficient as possible. instead of marshalling and unmarshalling you can manually construct a copy. Luckily there aren't so many fields to copy over:
```
xdr.LiquidityPoolEntry{
		LiquidityPoolId: xdr.PoolId{0xca, 0xfe, 0xba, 0xbe, 0xde, 0xad, 0xbe, 0xef},
		Body: xdr.LiquidityPoolEntryBody{
			Type: xdr.LiquidityPoolTypeLiquidityPoolConstantProduct,
			ConstantProduct: &xdr.LiquidityPoolEntryConstantProduct{
				Params: xdr.LiquidityPoolConstantProductParameters{
					AssetA: xdr.MustNewCreditAsset("USD", "GC3C4AKRBQLHOJ45U4XG35ESVWRDECWO5XLDGYADO6DPR3L7KIDVUMML"),
					AssetB: xdr.MustNewNativeAsset(),
					Fee:    34,
				},
				ReserveA:                 450,
				ReserveB:                 500,
				TotalPoolShares:          412241,
				PoolSharesTrustLineCount: 52115,
			},
		},
	}
``` 


you can also add a unit test specifically for this function using the `randxdr` package to generate a random `LiquidityPoolEntry` then call `copyPoolState()` and check that the return value is equal to the source by marshalling both and comparing the binary forms

exp/orderbook/graph.go
```diff
@@ -478,3 +480,131 @@ func sortAndFilterPaths(
...
+	return
+}
+
+// calculateAmountDisbursed calculates the pool payout. From CAP-38:
+//
+//      y = floor[(1 - F) Yx / (X + x - Fx)]
+//
+// It returns false if the calculation overflows.
+func calculatePoolPayout(reserveA, reserveB, received xdr.Int64, fee xdr.Int32) (xdr.Int64, bool) {
+	X, Y := big.NewFloat(float64(reserveA)), big.NewFloat(float64(reserveB))
```
tamirms, line 561: try to avoid using the `math/big` package because it's quite slow. I think you should be able to use https://pkg.go.dev/math/bits to implement the calculations which will be much faster

tamirms: we use `math/bits` in `price/main.go`

exp/orderbook/graph.go
```diff
@@ -478,3 +480,131 @@ func sortAndFilterPaths(
...
+		return
+	}
+
+	payout = uint32(payoutXdr)
+	newDetails := newPool.Body.ConstantProduct
+
+	// Adjust reserves based on this exchange with the LP. Note that pool shares
+	// don't change because the exchange doesn't make you an actual participant
+	// in the pool.
+	if isAssetA {
```
tamirms, line 544: do we need to check for overflows on ReserveA orReserveB?

exp/orderbook/graph.go
```diff
@@ -478,3 +480,131 @@ func sortAndFilterPaths(
...
+// makeTrade simulates execution of an exchange with a liquidity pool.
+//
+// It returns the amount that would be paid out by the pool for depositing
+// `amount` of `asset` alongside the new state of the liquidity pool after this
+// exchange, or an error. Errors can occur because of negative amounts, invalid
+// assets, invalid pools, XDR issues, etc.
+//
+// Refer to https://github.com/stellar/stellar-protocol/blob/master/core/cap-0038.md#pathpaymentstrictsendop-and-pathpaymentstrictreceiveop
+// for details on the exchange algorithm.
+func makeTrade(
```
tamirms, line 493: can you add a benchmark for this function https://tutorialedge.net/golang/benchmarking-your-go-programs/ ?

exp/orderbook/graph.go
```diff
@@ -478,3 +480,131 @@ func sortAndFilterPaths(
...
+	asset xdr.Asset,
+	deposit uint32,
+	pool xdr.LiquidityPoolEntry,
+) (payout uint32, newPool xdr.LiquidityPoolEntry, err error) {
```
tamirms, line 497: I just realized that we don't need to calculate `newPool`. I thought it was necessary but I forgot that we don't allow cycles in the path finding algorithm.

Once we trade against a pool for a given path we will never revisit it. Because we will never revisit the pool we don't need to keep track of the state of the pool after the trade.

exp/orderbook/graph.go
tamirms: you can ignore this comment. we can get rid of `copyPoolState()` entirely, see https://github.com/stellar/go/pull/3888/files#r701706417

exp/orderbook/graph.go
Shaptic: Oof yeah, good point. We needed to do this for the orderbook because there could be more than one offer for a particular pair, right? Thankfully, the `calculatePoolPayout` function is still useful.

Shaptic: Ah, and we don't need to unwind the state either since we're operating on a copy.

tamirms: > We needed to do this for the orderbook because there could be more than one offer for a particular pair, right?

we don't need to modify the offers either in the path finding algorithm for the same reason. Once we visit an asset during a given path we will never go back so we don't need to look back on those offers

exp/orderbook/graph.go
```diff
@@ -478,3 +480,131 @@ func sortAndFilterPaths(
...
+		return
+	}
+
+	newPool, err = copyPoolState(pool)
+	if err != nil {
+		err = errors.Wrap(err, "Failed to duplicate LP state")
+		return
+	}
+
+	payout = uint32(payoutXdr)
```
bartekn, line 538: `calculatePoolPayout` returns `xdr.Int64` so likely this should be `uint64` or `int64` but not `uint32`.
```suggestion
	payout = uint64(payoutXdr)
```

exp/orderbook/graph.go
Shaptic: Done in ea172b68.

exp/orderbook/graph.go
Shaptic: Good call! Part of ea172b68.

exp/orderbook/graph.go
Shaptic: Okay, I'm looking into this now. The biggest challenge, I think, is the fact that the fee `F` is expressed in bips in the liquidity pool parameters, while being a raw value in the formula (30 vs. 0.0003). This means we need to do `F /= 1e4`, but that'll usually be a very small value, and I'm concerned about floating point precision limitations with `math.bits`.

exp/orderbook/graph.go
Shaptic: Definitely, I think I fixed the types now.

exp/orderbook/graph.go
```diff
@@ -478,3 +481,92 @@ func sortAndFilterPaths(
...
+
+	payoutXdr, ok := calculatePoolPayout(X, Y, depositXdr, details.Params.Fee)
+	if !ok {
+		return 0, errors.New("Liquidity pool overflows from this exchange")
+	}
+
+	// should never happen, right?? by definition of an AMM
+	if payoutXdr > Y {
+		return 0, errors.New("Not enough reserve to pay out this exchange")
+	}
```
Shaptic, line 528: @jonjove, will the invariant formula (from [here](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0038.md#pathpaymentstrictsendop-and-pathpaymentstrictreceiveop)):

```matlab
y = floor[(1 - F) Yx / (X + x - Fx)]
```

ever return a payout bigger than the current reserve? That is, will `y <= Y` always hold in this calculation? My gut says yes but I want to confirm.

jonjove: `y <= Y` always holds. For proof:

```
y  = floor[(1 - F) Yx / (X + x - Fx)]
  <= (1 - F) Yx / (X + x - Fx)
  <= (1 - F) Yx / (x - Fx)                   (because F <= 1, x >= 0, and X >= 0)
   = (1 - F) Yx / [(1 - F)x]
   = Y
```

Shaptic: Perfect, thank you!

Shaptic:
@tamirms I don't think the math/bits approach will work out to be faster. 

After introducing 77b2cd896f09055ded6f197125a4d92e6bf43c57, the benchmark showed that the `big.Float` version was faster. I think it's due to 
  - dealing with conversions between `uint64`s and a custom `Uint128`, as well as 
  - the fact that you still need to drop into `math/big` to do `int128` division (unsupported by Golang, nor by math/bits, but necessary for the formula as seen in the Stellar-Core code [here](https://github.com/jonjove/stellar-core/blob/522f52ebf7e88a798916f3314bbdfebab52cfcc0/src/util/numeric.cpp#L265) where B and C can be 128-bit because of prior 64-bit multiplies, e.g. with `reserveB * deposit`).

Obviously, my implementation could be (and likely is) very lackluster in places, but fundamentally I don't think we'll see the same speed-up that we may have in other places because 128-bit multiplies/divisions are non-trivial.

On the bright side, the Core reference code showed me how to do the calculation in bips, leading to a 2x speedup in b1ddf4c04266ba60866a5495a5f20f136cc0dcad since `big.Float` is no longer necessary.

tamirms (approved):
@Shaptic ok let's use the `big.Int` version since it's faster. thanks for doing the analysis to confirm which version is faster!

Shaptic:
As noted in the _Known Limitations_,  this can't be merged until `makeTrade` is actually used, so this PR can stay open yet "Done," and a new PR will build on top of it.

2opremio:
> As noted in the _Known Limitations_, this can't be merged until `makeTrade` is actually used, so this PR can stay open yet "Done," and a new PR will build on top of it.

you could simply add a annotation:

```go
//lint:ignore U1000 Ignore unused function until we make use of it later on
func makeTrade()
```

See https://staticcheck.io/docs/configuration#line-based-linter-directives

PR #3891, opened 2021-09-05
clients/horizonclient: allow Fund on other networks and rely on server to indicate support

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Allow the `Fund` function to submit a fund request to any horizon instance, not just the instance defined by `horizonclient.DefaultTestNetClient`, and rely on the horizon instance responding with a 404 as an indicator that funding is not supported.

### Why

The only way to construct a `horizonclient.Client` that supports the `Fund` operation is to use and mutate the `horizonclient.DefaultTestNetClient`. This is unnecessarily restrictive because a developer may wish to use the `Fund` operation with a non-default client that they have constructed themselves, or they may wish to use the `Fund` operation with their own private or standalone test network.

### Known limitations

N/A

### DIFF
--- a/clients/horizonclient/client.go
+++ b/clients/horizonclient/client.go
@@ -575,11 +575,11 @@ func (c *Client) Trades(request TradeRequest) (tds hProtocol.TradesPage, err err
 // Fund creates a new account funded from friendbot. It only works on test networks. See
 // https://www.stellar.org/developers/guides/get-started/create-account.html for more information.
 func (c *Client) Fund(addr string) (tx hProtocol.Transaction, err error) {
-	if !c.isTestNet {
-		return tx, errors.New("can't fund account from friendbot on production network")
-	}
 	friendbotURL := fmt.Sprintf("%sfriendbot?addr=%s", c.fixHorizonURL(), addr)
 	err = c.sendGetRequest(friendbotURL, &tx)
+	if IsNotFoundError(err) {
+		return tx, errors.Wrap(err, "funding is only available on test networks and may not supported by "+c.fixHorizonURL())
+	}
 	return
 }
 
--- a/clients/horizonclient/client.go
+++ b/clients/horizonclient/client.go
@@ -578,7 +578,7 @@ func (c *Client) Fund(addr string) (tx hProtocol.Transaction, err error) {
 	friendbotURL := fmt.Sprintf("%sfriendbot?addr=%s", c.fixHorizonURL(), addr)
 	err = c.sendGetRequest(friendbotURL, &tx)
 	if IsNotFoundError(err) {
-		return tx, errors.Wrap(err, "funding is only available on test networks and may not supported by "+c.fixHorizonURL())
+		return tx, errors.Wrap(err, "funding is only available on test networks and may not be supported by "+c.fixHorizonURL())
 	}
 	return
 }
--- a/clients/horizonclient/client_fund_test.go
+++ b/clients/horizonclient/client_fund_test.go
@@ -25,7 +25,6 @@ func TestFund(t *testing.T) {
 	client := &Client{
 		HorizonURL: "https://localhost/",
 		HTTP:       hmock,
-		isTestNet:  true,
 	}
 
 	hmock.On(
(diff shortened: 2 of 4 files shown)


clients/horizonclient/client.go
```diff
@@ -575,11 +575,11 @@ func (c *Client) Trades(request TradeRequest) (tds hProtocol.TradesPage, err err
 // Fund creates a new account funded from friendbot. It only works on test networks. See
 // https://www.stellar.org/developers/guides/get-started/create-account.html for more information.
 func (c *Client) Fund(addr string) (tx hProtocol.Transaction, err error) {
-	if !c.isTestNet {
-		return tx, errors.New("can't fund account from friendbot on production network")
-	}
 	friendbotURL := fmt.Sprintf("%sfriendbot?addr=%s", c.fixHorizonURL(), addr)
 	err = c.sendGetRequest(friendbotURL, &tx)
+	if IsNotFoundError(err) {
+		return tx, errors.Wrap(err, "funding is only available on test networks and may not supported by "+c.fixHorizonURL())
```
Shaptic, line 581: typo: may not **be**

clients/horizonclient/client_fund_test.go
```diff
@@ -37,3 +36,26 @@ func TestFund(t *testing.T) {
...
+		HTTP:       hmock,
+	}
+
+	hmock.On(
+		"GET",
+		"https://localhost/friendbot?addr=GBLPP2W3X3PJQXYMC7EFWM5G2QCZL7HTCTFNMONS4ITGAYJ3GNNZIQ4V",
+	).ReturnString(404, friendbotFundResponse)
+
+	_, err := client.Fund("GBLPP2W3X3PJQXYMC7EFWM5G2QCZL7HTCTFNMONS4ITGAYJ3GNNZIQ4V")
+	assert.EqualError(t, err, "funding is only available on test networks and may not supported by https://localhost/: horizon error: \"Resource Missing\" - check horizon.Error.Problem for more information")
```
Shaptic, line 60: This is more of a philosophical discussion, but I've been meaning to ask this for a while: why do we test for specific error messages like this rather than just `assert.Error`, or maybe even a regex match for a substring on the error? I've read that the latter is considered an anti-pattern, but don't fully understand why. 

My rationale against it:

 - awkward hard-coding of errors: we can't use constants, and so we have long strings here
 - increased test verbosity without (in my opinion) added value. For example, the message has a typo (see above), but this test wouldn't have caught that.
 - discourages changing error messages that may increase clarity or information density, since there are many places to edit it

Again, I get that this is more of a philosophical best-practices discussion, but this seemed like an isolated-enough PR to get myself some clarity on that.

clients/horizonclient/client_fund_test.go
```diff
@@ -37,3 +36,26 @@ func TestFund(t *testing.T) {
...
 	assert.Equal(t, int32(8269), tx.Ledger)
 }
+
+func TestFund_notSupported(t *testing.T) {
+	friendbotFundResponse := `{
+  "type": "https://stellar.org/horizon-errors/not_found",
+  "title": "Resource Missing",
+  "status": 404,
+  "detail": "The resource at the url requested was not found.  This usually occurs for one of two reasons:  The url requested is not valid, or no data in our database could be found with the parameters provided."
+}`
```
Shaptic, line 46: Maybe we actually want to intentionally make this response as minimal as possible, to ensure that the method only relies on a particular subset of fields? Specifically, I would expect this to behave identically if everything but the `type` and `status` were omitted.

Shaptic (approved):
I think this is a good change, but have some musings below (and a typo fix).

clients/horizonclient/client.go
leighmcculloch: Thanks for catching this!
```suggestion
		return tx, errors.Wrap(err, "funding is only available on test networks and may not be supported by "+c.fixHorizonURL())
```

clients/horizonclient/client_fund_test.go
leighmcculloch: I prefer to test with as real responses as possible, in this case it's just a few more lines so I don't think the tradeoffs of doing so are that great. I can reduce it if you prefer though.

Shaptic: I think for me this is somewhat related to my other comment given the verbosity of `detail`, but your rationale makes sense; no strong opinion on my part.

clients/horizonclient/client_fund_test.go
leighmcculloch: I think the distinction here is that this error string I'm wrapping the not found error in is just an annotation or a hint, like a comment about the error. The error is still the Horizon not found error in the propoer horizon client error object. Because of this I don't it is worth adding a new error type for the annotation.

In regards to more generally, the reason that regexing isn't preferred is it is brittle. The preferred approach for error matching is to use `errors.Is` or `errors.As`, which works great with custom error types that wrap an error, or error constants that don't wrap another error.

In this specific case we can't use a constant for the error, otherwise we'd lose the underlying horizon not found error, which I think is actually the most important error in this scenario. In this specific case I don't think it is worth defining an error type to wrap the not found error.

leighmcculloch: I misunderstood your question.

> why do we test for specific error messages like this rather than just `assert.Error`

`assert.Error` would tell us that any error occurred. Any error could occur for a variety of other reasons and we wouldn't necessarily be testing the exact scenario we intend.

`assert.EqualError` is maybe a little stricter than we need, but it works?

PR #3897, opened 2021-09-07
txnbuild: implement encoding.Text{M,Unm}arshaler in Transaction, FeeBumpTransaction, GenericTransaction

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Implement `encoding.TextMarshaler` and `encoding.TextUnmarshaler` in `Transaction`, `FeeBumpTransaction`, `GenericTransaction`.

### Why

Transaction, FeeBumpTransaction, and GenericTransaction all have well defined formats for serialization as strings. It would be convenient if these types were automatically serializable via encoders such as the `encoding/json` package.

Constructors for GenericTransaction are added so that code building Transaction and FeeBumpTransaction can get a GenericTransaction for those built transactions, for setting in a field that can hold both, in a type supporting serialization.

### Known limitations

N/A

### DIFF
--- a/txnbuild/transaction.go
+++ b/txnbuild/transaction.go
@@ -568,8 +568,8 @@ func (t *FeeBumpTransaction) Base64() (string, error) {
 	return marshallBase64(t.envelope, t.Signatures())
 }
 
-// NewGenericTransactionWithFeeBumpTransaction creates a GenericTransaction
-// containing the FeeBumpTransaction.
+// ToGenericTransaction creates a GenericTransaction containing the
+// FeeBumpTransaction.
 func (t *FeeBumpTransaction) ToGenericTransaction() *GenericTransaction {
 	return &GenericTransaction{feeBump: t}
 }
--- a/txnbuild/transaction.go
+++ b/txnbuild/transaction.go
@@ -382,6 +382,11 @@ func (t *Transaction) Base64() (string, error) {
 	return marshallBase64(t.envelope, t.Signatures())
 }
 
+// ToGenericTransaction creates a GenericTransaction containing the Transaction.
+func (t *Transaction) ToGenericTransaction() *GenericTransaction {
+	return &GenericTransaction{simple: t}
+}
+
 // ClaimableBalanceID returns the claimable balance ID for the operation at the given index within the transaction.
 // given index (which should be a `CreateClaimableBalance` operation).
 func (t *Transaction) ClaimableBalanceID(operationIndex int) (string, error) {
@@ -563,6 +568,12 @@ func (t *FeeBumpTransaction) Base64() (string, error) {
 	return marshallBase64(t.envelope, t.Signatures())
 }
 
+// NewGenericTransactionWithFeeBumpTransaction creates a GenericTransaction
+// containing the FeeBumpTransaction.
+func (t *FeeBumpTransaction) ToGenericTransaction() *GenericTransaction {
+	return &GenericTransaction{feeBump: t}
+}
+
 // InnerTransaction returns the Transaction which is wrapped by
 // this FeeBumpTransaction instance.
 func (t *FeeBumpTransaction) InnerTransaction() *Transaction {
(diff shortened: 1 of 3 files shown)


txnbuild/transaction.go
```diff
@@ -526,6 +578,18 @@ type GenericTransaction struct {
 	feeBump *FeeBumpTransaction
 }
 
+// NewGenericTransactionWithTransaction creates a GenericTransaction containing
+// the given Transaction.
+func NewGenericTransactionWithTransaction(tx *Transaction) *GenericTransaction {
```
tamirms, line 583: what's the motivation for adding these GenericTransaction constructors? A `GenericTransaction` should only be constructed via `TransactionFromXDR()`

leighmcculloch: Serialization and deserialization on `Transaction` and `FeeBumpTransaction` are useful, but only if you know whether it is one or the other. In cases where we serialize and deserialize a transaction we don't know the type of, we should be deserializing into the `GenericTransaction`. When serializing this might mean creating a Transaction in code, and then needing to put it inside a GenericTransaction so that it can be set on a field in a struct that supports serializing/deserializing both types of transactions.

leighmcculloch: > A `GenericTransaction` should only be constructed via `TransactionFromXDR()`

I don't think we should have to serialize a transaction to XDR base64 to put it in a GenericTransaction.

tamirms: Why do you have to wrap `tx` in your example into a `GenericTransaction` instance? You can call `json.Encode()` directly on `tx`. When you want to decode the json you can either decode the payload into a `Transaction` instance if you know it's not a fee bump transaction or you can decode it into a `GenericTransaction` instance.

leighmcculloch: If there's no constructor, then the above example becomes more complicated because I have to serialize different outer objects depending on whether it is a transaction or a fee bump:

```go
type MessageWithTransaction struct {
    Tx *txnbuild.Transaction
}
type MessageWithFeeBump struct {
    Tx *txnbuild.FeeBumpTransaction
}
type Message struct {
    Tx *txnbuild.GenericTransaction
}

// Sending Tx
tx := txnbuild.NewTransaction(...)
m := MessageWithTransaction{Tx: tx}
_, _ := json.Marshal(m)

// Receiving Tx
tx := txnbuild.NewTransaction(...)
m := Message{}
_ := json.Unmarshal(..., &m)
```

Or, I have to do a serialization round trip to put my `txnbuild.Transaction` into the `Tx *GenericTransaction` field.


```go
type Message struct {
    Tx *txnbuild.GenericTransaction
}

tx := txnbuild.NewTransaction(...)
b, _ := tx.Base64()
m := Message{Tx: txnbuild.TransactionFromXDR(b)}
_, _ := json.Encode(m)
```

tamirms: ok I understand why the constructors are necessary for your example now. What about adding `ToGenericTransaction()` functions to `txnbuild.Transaction` and `txnbuild.FeeBumpTransaction`? I think `Message{Tx: tx.ToGenericTransaction()}` is preferable to `Message{Tx: txnbuild.NewGenericTransactionWithTransaction(tx)}`

leighmcculloch: I considered that option, but if we add a `ToGenericTransaction()` we're coupling types since `GenericTransaction` is dependent on `Transaction` and `Transaction` is dependent on `GenericTransaction. By adding a constructor we continue to only embed because `GenericTransaction` contains `Transaction` and `Transaction` doesn't know about `GenericTransaction`.

I think a constructor is also more discoverable. For example, when I want to create a type, I look for constructors.

> I think `Message{Tx: tx.ToGenericTransaction()}` is preferable to

What are the benefits of `ToGenericTransaction`?

tamirms: there are two benefits:

1) `tx.ToGenericTransaction()` is easier to read than `txnbuild.NewGenericTransactionWithTransaction(tx)`
2) with your proposed change there are essentially 3 constructors for `GenericTransaction`: `NewGenericTransactionWithTransaction()`, `NewGenericTransactionWithFeeBumpTransaction()`, and `TransactionFromXDR()`. These 3 constructors are named inconsistently. To fix this we could rename `TransactionFromXDR()` to `NewGenericTransactionFromXDR()` or something similar. But that would be a breaking change. That is why my preference is to let `TransactionFromXDR()` be the only constructor.

Regarding the coupling, I don't see how it's a big deal in practice. Could you provide an example of how the coupling could lead to a bad outcome in the future?

leighmcculloch: There are still three constructors, and they still have different names: `TransactionFromXDR`, `ToGenericTransaction` and `ToGenericTransaction`, and they'll be spread out in the docs.

These are all small details, if the pattern you folks prefer is `To*` functions for creating superset types that works.

3d90e0e5

txnbuild/transaction.go
```diff
@@ -506,11 +542,38 @@ func (t *FeeBumpTransaction) MarshalBinary() ([]byte, error) {
...
+	*t = *fbtx
+	return nil
+}
+
 // Base64 returns the base 64 XDR representation of the transaction envelope.
 func (t *FeeBumpTransaction) Base64() (string, error) {
 	return marshallBase64(t.envelope, t.Signatures())
 }
 
+// NewGenericTransactionWithFeeBumpTransaction creates a GenericTransaction
```
tamirms, line 571: replace `NewGenericTransactionWithFeeBumpTransaction` with `ToGenericTransaction` in the comment

leighmcculloch: Thanks! af4891d5

tamirms (approved):

PR #3898, opened 2021-09-08
services/horizon: Soft delete liquidity pools

tamirms:
This PR implements the following two changes which are required for https://github.com/stellar/go/issues/3836 :

* Add a deleted column to the liquidity_pools table which we'll use to soft delete liquidity pool rows
* Periodically prune deleted rows in `CompactLiquidityPools()`

In a subsequent PR, the in memory orderbook will use `GetUpdatedLiquidityPools()` to poll for changes to any liquidity pools and apply those changes to its in-memory data structures.


bartekn (approved):
LGTM!

PR #3904, opened 2021-09-10
Update docker image build timestamp format

jacekn:
Golang time parser is not compatible with format rendered by the GNU date,
even though both claim to be RFC3339. This change replaces space separator
between date and time with "T" which will allow golang to parse it.
It also switches from `--utc` to `-u` that's supported by both GNU and BSD
date commands.

### DIFF
--- a/services/keystore/Makefile
+++ b/services/keystore/Makefile
@@ -4,7 +4,7 @@ SUDO := $(shell docker version >/dev/null 2>&1 || echo "sudo")
 # If TAG is not provided set default value
 TAG ?= stellar/keystore:$(shell git rev-parse --short HEAD)$(and $(shell git status -s),-dirty-$(shell id -u -n))
 # https://github.com/opencontainers/image-spec/blob/master/annotations.md
-BUILD_DATE := $(shell date --utc --rfc-3339=seconds)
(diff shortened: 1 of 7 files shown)


bartekn (approved):

services/keystore/Makefile
```diff
@@ -4,7 +4,7 @@ SUDO := $(shell docker version >/dev/null 2>&1 || echo "sudo")
 # If TAG is not provided set default value
 TAG ?= stellar/keystore:$(shell git rev-parse --short HEAD)$(and $(shell git status -s),-dirty-$(shell id -u -n))
 # https://github.com/opencontainers/image-spec/blob/master/annotations.md
-BUILD_DATE := $(shell date --utc --rfc-3339=seconds)
+BUILD_DATE := $(shell date -utc +%FT%TZ)
```
marcelosalloum, line 7: Shouldn't this be `-u` instead of `-utc`, like the others?

marcelosalloum (commented):
I think there's a place where you forgot to update from `-utc` to `-u`.

services/keystore/Makefile
jacekn: Absolutely! Well spotted, thank you

PR #3908, opened 2021-09-10
.vscode: fix tasks file

leighmcculloch:
### What

Format the VSCode tasks file.

### Why

When I open the stellar/go repo locally in the latest version of VSCode it displays a warning saying the file is "dirty". It appears the file is not formatted to how VSCode prefers it to be formatted. I don't really know what this file does or is for, but supposedly it was added because somebody uses it. This change resolves the warning.

### Known limitations

N/A

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -9,9 +9,15 @@
     "options": {
         "cwd": "${workspaceRoot}"
     },
-    "tasks": [{
-        "taskName": "test all",
-        "isWatching": false,
-        "args": ["test", "./..."]
-    }]
+    "tasks": [
+        {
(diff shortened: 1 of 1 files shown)


bartekn (approved):
LGTM but it reminded me about a discussion we had quite a long time ago about keeping the IDE specific files in this repo. Not everyone is using VS Code. Let's merge it but we can discuss it somewhere.

leighmcculloch:
@bartekn I agree, we should keep ide specific files out of the repo. I think VSCode is probably the exception since GitHub has VSCode built into it's UI and it's the IDE for GitHub Codespaces.

PR #3873, opened 2021-08-30
services/horizon: Allow filtering accounts by liquidity pool participation.

Shaptic:
<details>
  <summary>PR Checklist</summary>
  
### PR Structure

* [x] This PR has reasonably narrow scope (if not, break it down into smaller PRs).
* [x] This PR avoids mixing refactoring changes with feature changes (split into two PRs
  otherwise).
* [x] This PR's title starts with name of package that is most changed in the PR, ex.
  `services/friendbot`, or `all` or `doc` if the changes are broad or impact many
  packages.

### Thoroughness

* [ ] This PR adds tests for the most critical parts of the new functionality or fixes.
* [ ] I've updated any docs ([developer docs](https://www.stellar.org/developers/reference/), `.md`
  files, etc... affected by this change). Take a look in the `docs` folder for a given service,
  like [this one](https://github.com/stellar/go/tree/master/services/horizon/internal/docs).

### Release planning

* [x] I've updated the relevant CHANGELOG ([here](services/horizon/CHANGELOG.md) for Horizon) if
  needed with deprecations, added features, breaking changes, and DB schema changes.
* [x] I've decided if this PR requires a new major/minor version according to
  [semver](https://semver.org/), or if it's mainly a patch change. The PR is targeted at the next
  release branch if it's not a patch change.
</details>

### What
- Amends the API to include liquidity pool details in account balances.
- Allows only querying accounts participating in a particular liquidity pool (`GET /accounts?liquidity_pool=[...some id...]`).

### Why
Resolves #3832.

### Known limitations
Still needs to be tested, hence why it's in draft.

PR #3910, opened 2021-09-13
services/horizon/internal/actions: Fix liquidity_pool filter for accounts endpoint

tamirms:


bartekn (approved):

Shaptic:
Definitely partially my fault (#3873); I handed off the work before I could write tests. Thanks for the quick fix!

PR #3914, opened 2021-09-14
keypair: improve performance of FromAddress/Full Hint, Sign, Verify

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Store the generated ed25519 public and private keys within the `FromAddress` and `Full` types.

Also, simplified our implementation of crc16.

### Why

Benchmarking with stellar/experimental-payment-channels#318 identified that the app was spending as much time constructing keys as it was verifying and signing. It's expected that verifying and signing will be somewhat expensive, but it was a surprise that repeated use of an in memory keypair would result in repeated regeneration of the keys.

Regenerating keys doesn't only consume CPU time, it allocates memory.

The `FromAddress` and `Full` types store the public and private key in strkey [SEP-23] form in memory. Every `Hint`, `Sign`, and `Verify` call decodes the strkey then generates the ed25519 keys used by the Go stdlib. That key setup is pretty expensive, especially for verifying with a secret key, and for the hint comparisons.

There are some competing attributes of keypair that make introducing this change non-trivial.

1. The keypair types are safe for concurrent use if you treat them as immutable after construction. Introducing a cache and caching on first hint/sign/verify, would break that attribute because the types would be mutable and calling those functions concurrently would no longer be safe without a synchronization primitive such as `sync.Once` or `sync.Mutex`.
2. The keypair types are copyable. Introducing a simple `sync.Once` on the keypairs to address the first point would break the copyable attribute because synchronizing types like `sync.Once` and `sync.Mutex` are not safe to copy in Go.

These two points leave us with one reasonable solution: generating the ed25519 keys when the keypair FromAddress and Full values are created. This makes some assumptions about how the keypair types are used. For example, if you use the `Full` type to purely hold an `S...` strkey but use none of the signing, verifying, or hint functions, the `Full`s construction will now be slower. Also, if you call `keypair.Random` frequently that call will also slower. However, these slower operations are shifting existing time spent in the other functions, so this is not new time applications will spend, unless those applications are generating random keys without actually using the keys. Generally applications use types for functionality and these types are primarily present for signing and verifying.

The crc16 implementation is also simplified, removing use of the `binary.Write` and `bytes.Buffer` which are both unnecessary. Removing the use of those reduce the allocations from 24 to 18 in calls to `FromRawSeed` (used by `Random`) and helps to offset a little the impact of adding new weight to that function.

### Benchmarks

The following benchmarks were run with the `benchmarks_test.go` against the `master` branch and this branch, then the results were pushed through [benchstat](https://pkg.go.dev/golang.org/x/perf/cmd/benchstat) to get comparisons.

```
goos: darwin
goarch: amd64
pkg: github.com/stellar/go/keypair
cpu: Intel(R) Core(TM) i7-8569U CPU @ 2.80GHz
```

```
name                        old time/op    new time/op    delta
FromAddress_ParseAddress-8     464ns ± 0%     375ns ± 0%    -19.26%
FromAddress_Hint-8             468ns ± 0%       3ns ± 0%    -99.43%
FromAddress_Verify-8          54.0µs ± 0%    53.1µs ± 0%     -1.53%
Full_ParseFull-8               239ns ± 0%     250ns ± 0%     +4.56%
Full_FromRawSeed-8             439ns ± 0%   19445ns ± 0%  +4330.39%
Full_Hint-8                   19.4µs ± 0%     0.0µs ± 0%    -99.99%
Full_Verify-8                 73.0µs ± 0%    57.9µs ± 0%    -20.80%
Full_Sign-8                   42.0µs ± 0%    22.2µs ± 0%    -47.03%

name                        old alloc/op   new alloc/op   delta
FromAddress_ParseAddress-8      244B ± 0%      178B ± 0%    -27.05%
FromAddress_Hint-8              244B ± 0%        0B        -100.00%
FromAddress_Verify-8            244B ± 0%        1B ± 0%    -99.59%
Full_ParseFull-8                128B ± 0%      128B ± 0%      0.00%
Full_FromRawSeed-8              374B ± 0%      776B ± 0%   +107.49%
Full_Hint-8                     420B ± 0%        0B        -100.00%
Full_Verify-8                   420B ± 0%        0B        -100.00%
Full_Sign-8                     484B ± 0%        0B        -100.00%

name                        old allocs/op  new allocs/op  delta
FromAddress_ParseAddress-8      6.00 ± 0%      4.00 ± 0%    -33.33%
FromAddress_Hint-8              6.00 ± 0%      0.00        -100.00%
FromAddress_Verify-8            6.00 ± 0%      0.00        -100.00%
Full_ParseFull-8                2.00 ± 0%      2.00 ± 0%      0.00%
Full_FromRawSeed-8              10.0 ± 0%      18.0 ± 0%    +80.00%
Full_Hint-8                     10.0 ± 0%       0.0        -100.00%
Full_Verify-8                   10.0 ± 0%       0.0        -100.00%
Full_Sign-8                     11.0 ± 0%       0.0        -100.00%
```

I used a Makefile to automate running tests on master and the branch and comparing them. The Makefile is visible in this commit 0f172807cbcac80a5f4abac91889da2172d2d342 if anyone is of interest, but it has been removed from this PR now.

### Testing

Existing unit tests have been updated and cover the api surface area pretty well. There was a small area of the package without test coverage and I have added tests for that area in #3922 which will be merged ahead of this PR.

In addition to unit tests I wrote a regression test that is in this PR but won't be merged with the final merge. It uses random input and executes all of the functions of the FromAddress and Full types on both the master branch and on this PR and compares the outputs of all of them. The tests have shown no observable difference.

The regression test also runs all its assertions in parallel multiple times for the same input, and when the regression test was run with the race detector enabled there was no races detected. This provides some confidence over what we know thereotically, that this PR does not change the keypair package's safety for us in concurrent applications when the keypair values are used in an immutable fashion.

The regression tests have been removed from this PR now, but they are visible on this branch before commit 7372f9d94d98874d03db8ed0904e7f907afa1f80.

### Known limitations

N/A

[SEP-23]: https://stellar.org/protocol/sep-23


keypair/main_test.go
```diff
@@ -140,9 +140,14 @@ var _ = DescribeTable("keypair.ParseFull()",
...
 	Entry("a valid seed", ParseFullCase{
-		Input:    "SDHOAMBNLGCE2MV5ZKIVZAQD3VCLGP53P3OBSBI6UN5L5XZI5TKHFQL4",
-		FullCase: Equal(&Full{seed: "SDHOAMBNLGCE2MV5ZKIVZAQD3VCLGP53P3OBSBI6UN5L5XZI5TKHFQL4"}),
-		ErrCase:  BeNil(),
+		Input: "SDHOAMBNLGCE2MV5ZKIVZAQD3VCLGP53P3OBSBI6UN5L5XZI5TKHFQL4",
+		FullCase: Equal(&Full{
+			address:    "GBRPYHIL2CI3FNQ4BXLFMNDLFJUNPU2HY3ZMFSHONUCEOASW7QC7OX2H",
+			seed:       "SDHOAMBNLGCE2MV5ZKIVZAQD3VCLGP53P3OBSBI6UN5L5XZI5TKHFQL4",
+			publicKey:  ed25519.PublicKey{98, 252, 29, 11, 208, 145, 178, 182, 28, 13, 214, 86, 52, 107, 42, 104, 215, 211, 71, 198, 242, 194, 200, 238, 109, 4, 71, 2, 86, 252, 5, 247},
+			privateKey: ed25519.PrivateKey{206, 224, 48, 45, 89, 132, 77, 50, 189, 202, 145, 92, 130, 3, 221, 68, 179, 63, 187, 126, 220, 25, 5, 30, 163, 122, 190, 223, 40, 236, 212, 114, 98, 252, 29, 11, 208, 145, 178, 182, 28, 13, 214, 86, 52, 107, 42, 104, 215, 211, 71, 198, 242, 194, 200, 238, 109, 4, 71, 2, 86, 252, 5, 247},
```
leighmcculloch, line 148: I generated these values using the `master` version of the `keypair` package using this code, to ensure that the keys stored during construction in this PR match the keys that were getting lazily generated with the current version.

```go
func TestExtractKnownKeys(t *testing.T) {
	pk := "GBRPYHIL2CI3FNQ4BXLFMNDLFJUNPU2HY3ZMFSHONUCEOASW7QC7OX2H"
	t.Logf(pk)
	pkfa := MustParseAddress(pk)
	t.Logf("publicKey: %v", pkfa.publicKey())

	sk := "SDHOAMBNLGCE2MV5ZKIVZAQD3VCLGP53P3OBSBI6UN5L5XZI5TKHFQL4"
	t.Logf(sk)
	skfa := MustParseFull(sk)
	skpk, sksk := skfa.keys()
	t.Logf("publicKey: %v", skpk)
	t.Logf("secretKey: %v", sksk)
}
```

keypair/benchmarks_test.go
```diff
@@ -0,0 +1,136 @@
...
+)
+
+func BenchmarkFromAddress_ParseAddress(b *testing.B) {
+	// Secret key for setting up the components.
+	sk := keypair.MustRandom()
+	address := sk.Address()
+
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		_, _ = keypair.ParseAddress(address)
```
Shaptic, line 18: :question: why assign to anything?

keypair/full.go
```diff
@@ -36,15 +83,14 @@ func (kp *Full) Verify(input []byte, sig []byte) error {
 	if len(sig) != 64 {
 		return ErrInvalidSignature
 	}
-	if !ed25519.Verify(kp.publicKey(), input, sig) {
+	if !ed25519.Verify(kp.publicKey, input, sig) {
```
Shaptic, line 86: This is a really, _really_ big breaking change, right? This kind of code is *every*where, afaik. Maybe we do a "property-esque" approach and keep `publicKey()` which just `return _publicKey`s?

keypair/benchmarks_test.go
leighmcculloch: The default linters for VS Code's warns if you don't check errors. The function returns an error. So I'm assigning to `_` to make it clear to the reader I'm intentionally not wanting to check the error.

keypair/full.go
leighmcculloch: `publicKey` is not exported, so we are guaranteed that the only places calling `publicKey` are inside this package. The package is small, and the compiler will error if there are any call sides that attempt to treat the field as a function. I don't think there's much value in keeping the function.

In regards to if this PR as a whole is a breaking change, there's plenty of risk for sure, which is why I'm testing beyond the typical measure of testing. Also, I'm still working on the change and evaluating whether it is worth it or not. I still think it is, but that might change.

Shaptic: Oh, good point about being private :thinking: I've clearly spent too much time looking at the JS `Keypair`.

keypair/from_address.go
```diff
@@ -15,7 +15,27 @@ import (
...
+	}
+	pub := ed25519.PublicKey(payload)
+	return &FromAddress{
+		address:   address,
+		publicKey: pub,
+	}, nil
+}
+
+func newFromAddressWithPublicKey(address string, publicKey ed25519.PublicKey) *FromAddress {
+	return &FromAddress{
```
bartekn, line 35: As a security precaution maybe we should check if address match the publicKey here?

bartekn (approved):
LGTM! Good catch and great PR description!

keypair/from_address.go
leighmcculloch: If we did that, this function would `sdtrkey.Decode` the `address` (because that's faster than `strkey.Encode`ing the `publicKey`). If we're going to calculate the other component we may as well just call `newFromAddress` which does the same thing.

The main reason this function exists is an optimization for `Full` converting itself to a `FromAddress`.

leighmcculloch:
I had one thought of how this change might not be safe: if someone was constructing a key in the form `keypair.Full{"S..."}`. However, a quick check and it turns out it is not possible to do this in Go because the compiler disallows the implicit assignment of an unexported field. For example:
```
./main.go:10:21: implicit assignment of unexported field 'address' in keypair.Full literal
```

PR #3915, opened 2021-09-14
protocols/horizon: Represent trade prices as int64 fractions

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Represent trade prices as int64 fractions.

### Why

Prior to protocol 18 we represented trade prices as a fraction where the numerator and denominator took on 32 bit integer values:

```golang
type Price struct {
	N int32 `json:"n"`
	D int32 `json:"d"`
}
```
In protocol 18 we introduce liquidity pools and trades against liquidity pools involve exchanged amount values which can take on 64  bit integer values. We derive the price for a liquidity pool trade by dividing the amounts exchanged. Therefore we can no longer represent trade prices as fractions with 32 bit numerators and denominators.

### Known limitations

This is a breaking change for **both** the trades and the trade aggregations API. We were already expecting this to be a breaking change for the trades API. However, at that time, I did not realize that we would also need to do a breaking change for the trade aggregations API as well. So we will need to communicate this breaking API change to horizon stakeholders.

Luckily we are representing the aggregation values in the `history_trades_60000` table as `numeric` postgres types, so no db migrations are required on the history aggregations table in order to support 64 bit trade prices.

### DIFF
--- a/protocols/horizon/main.go
+++ b/protocols/horizon/main.go
@@ -5,6 +5,7 @@ package horizon
 import (
 	"encoding/base64"
 	"encoding/json"
+	"math/big"
 	"strconv"
 	"time"
 
@@ -297,7 +298,7 @@ func (p Path) PagingToken() string {
 	return ""
 }
 
-// Price represents a price
+// Price represents a price for an offer
 type Price base.Price
 
 // PriceLevel represents an aggregation of offers that share a given price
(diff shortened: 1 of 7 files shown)


2opremio (approved):
It would be good to have an integration test for this though

PR #3920, opened 2021-09-15
services/horizon: Horizon 2.8.2 CHANGELOG

bartekn:


services/horizon/CHANGELOG.md
```diff
@@ -3,6 +3,13 @@
 All notable changes to this project will be documented in this
 file. This project adheres to [Semantic Versioning](http://semver.org/).
 
+## v2.8.2
+**Upgrading to this version from <= v2.8.0 will trigger a state rebuild. During this process (which will take at least 10 minutes), Horizon will not ingest new ledgers.**
+
+**In the 2.9.0 Horizon release, the `--ingest` flag will be set to `true` by default.**
+
+* Improve performance of `OffersProcessor`. This should speed up ingestion of latest Stellar Public Network activity by up to 30%. ([3917](https://github.com/stellar/go/pull/3917))
```
tamirms, line 11: would it be worth adding a note that it won't improve reingestion speeds? If someone is reingesting history they might think that this patch will improve the performance of that operation

bartekn: Very good point, changed in: https://github.com/stellar/go/pull/3920/commits/c60dd4737b5041bdf4375a0c60abab931e8c3e12.

tamirms (approved):

PR #3921, opened 2021-09-15
exp/orderbook: Expand path payment algorithm to search liquidity pools.

Shaptic:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Allows bidirectional path-finding through both liquidity pools and offers.

Major changes:

 - Support for simulating exchanges w/ a liquidity pool: see `makeTrade` and its corresponding calculation functions
 - Modify `edgeSet` to represent all trade opportunities: both offers (as before) and now liquidity pools.
 - Modify `searchState` interface to process offers and LPs ("venues") properly:
   * `venues(asset)` retrieves all pools and LPs for a given asset
   * `chooseVenue(a, b)` determines whether the path should choose offers (`a`) or pools (`b`), since the choice depends on if we're going forwards (`FindFixedPath`) or backwards (`FindPath`) through the graph's edges.
   * `consumePool(pool, asset, amount)` evaluates an exchange with a pool, which is also direction-dependent
 - the new `processVenues()` is a catch-all helper to evaluate pools and offers (in that order) fairly
- Modify the `OrderBookGraph` to handle batch add-update-removal of liquidity pools

### Why
We need to enable users to find paths through both the existing DEX and through the newly-introduced liquidity pools. 
See #3836 for more.

### Known limitations
I'd like to incorporate Jon's tests to evaluate the actual liquidity pool exchange math, and I'd also like to ensure that the test cases sufficiently cover the new functionality I've introduced here.


Shaptic:
@tamirms Before I dive into enabling LP hops for `FindFixedPaths` (and its corresponding `buyingGraphSearchState`), I'd like your review on my current approach. There are still optimization improvements I'd like to make to this one(especially in `venues()` to retrieve pools and offers for a given asset), but those will come after a confirmation that I'm on the right track!

exp/orderbook/batch.go
```diff
@@ -106,13 +106,21 @@ func (tx *orderBookBatchedUpdates) apply(ledger uint32) error {
...
+				tx.orderbook.liquidityPoolsForAsset[a], *entry)
+			tx.orderbook.liquidityPoolsForAsset[b] = append(
+				tx.orderbook.liquidityPoolsForAsset[b], *entry)
+
 		case removeLiquidityPoolOperationType:
-			if _, ok := tx.orderbook.liquidityPools[operation.liquidityPoolAssets]; !ok {
-				panic(errors.New("liquidity pool not present in orderbook graph"))
-			}
-			delete(tx.orderbook.liquidityPools, operation.liquidityPoolAssets)
+			delete(tx.orderbook.liquidityPoolsForAsset, operation.liquidityPoolAssets.buyingAsset)
```
tamirms, line 121: I don't think this is right. you cannot delete the entire map entry. I think you need to remove the element of the slice which matches `operation.liquidityPoolAssets`

exp/orderbook/dfs.go
```diff
@@ -64,6 +73,7 @@ func dfs(
 	state searchState,
 	maxPathLength int,
 	visited []xdr.Asset,
+	visitedPools *IdSet,
```
tamirms, line 76: I don't think we need to track the pools we have visited. we should be able to remove this parameter entirely

exp/orderbook/dfs.go
```diff
@@ -91,49 +101,110 @@ func dfs(
...
+				offers,
+			)
+
+			// We should only error out if the LP trade didn't happen.
+			if err != nil {
+				if bestExchangeRate == 0 {
+					return err
+				}
+
+				continue
```
tamirms, line 163: we don't want to continue here because the continue statement is associated with the for loop

exp/orderbook/dfs.go
```diff
@@ -91,49 +101,110 @@ func dfs(
...
+			}
+
+			// TODO: Move this check into consumeOffers to optimize it.
+			//
+			// TODO: Should we prefer offers or LPs if the exchange is
```
tamirms, line 168: it actually doesn't matter which venue we pick because path finding is only concerned with constructing the path (starting asset and amount, intermediate assets, ending asset and amount). whether we choose a lp or orderbook in each individual step of the path is irrelevant

exp/orderbook/dfs.go
```diff
@@ -91,49 +101,110 @@ func dfs(
...
 		return nil
 	}
 
-	for nextAssetString, offers := range state.edges(currentAssetString) {
-		if len(offers) == 0 {
-			continue
+	updatedVisitedPools := visitedPools.Clone()
+	for nextAssetString, venues := range state.venues(currentAssetString) {
+		bestExchangeRate := xdr.Int64(0)
+		var bestAsset xdr.Asset
```
tamirms, line 114: I would rename this to `nextAsset`. `bestAsset` implies you're looking at multiple different trading pairs but in the body of the for loop we're only examining trades for the currentAsset -> nextAsset trade

exp/orderbook/dfs.go
```diff
@@ -91,49 +101,110 @@ func dfs(
...
+
+				continue
+			}
+
+			// TODO: Move this check into consumeOffers to optimize it.
+			//
+			// TODO: Should we prefer offers or LPs if the exchange is
+			//       equivalent? My gut says offers, because (a) there's no fee
+			//       and (b) we reduce the orderbook size.
+			if nextAssetAmount <= bestExchangeRate || bestExchangeRate == 0 {
```
tamirms, line 171: you actually want to maximize `nextAssetAmount`. it seems like you are minimizing here. I suggest renaming `nextAssetAmount` to `amount` and `bestExchangeRate` to `nextAssetAmount` to make more apparent we are trying to maximize the outcome of the trade

exp/orderbook/dfs.go
```diff
@@ -180,8 +251,38 @@ func (state *sellingGraphSearchState) appendToPaths(
...
+		// we're overshooting by up to 2x, but it's a reasonable size hint
+		len(state.graph.edgesForSellingAsset)+
+			len(state.graph.liquidityPoolsForAsset))
+
+	// FIXME: I really don't like this whole "check or set" approach; lookups
+	//        are suboptimal and the code itself isn't clean.
+
+	for nextAsset, offers := range state.graph.edgesForSellingAsset[currentAsset] {
+		if opp, ok := result[nextAsset]; ok {
+			opp.offers = offers
```
tamirms, line 265: we will never execute this line of code because `result` starts out empty

exp/orderbook/graph.go
```diff
@@ -514,3 +467,156 @@ func sortAndFilterPaths(
 
 	return filtered, nil
 }
+
+const (
+	tradeTypeDeposit     = iota // deposit into pool, what's the payout?
```
tamirms, line 472: I don't see where we are making use of the trade types in `difs.go`

exp/orderbook/graph.go
```diff
@@ -514,3 +467,156 @@ func sortAndFilterPaths(
 
 	return filtered, nil
 }
+
```
tamirms, line 470: this code should either be moved to dfs.go or it should be moved into its own file

exp/orderbook/dfs.go
```diff
@@ -180,8 +251,38 @@ func (state *sellingGraphSearchState) appendToPaths(
 	})
 }
 
-func (state *sellingGraphSearchState) edges(currentAssetString string) edgeSet {
-	return state.graph.edgesForSellingAsset[currentAssetString]
+func (state *sellingGraphSearchState) venues(currentAsset string) map[string]Venues {
```
tamirms, line 254: since `venues()` is getting called many times in the dfs I think it would be better to avoid the unioning operation here. I think it would be better to  maintain the `map[string]Venues` data structure in the graph struct so that it can be obtained quickly in the dfs

exp/orderbook/dfs.go
tamirms: once https://github.com/stellar/go/pull/3933 is merged, you will need to pull in those changes on to your branch

exp/orderbook/dfs.go
Shaptic: good catch ty

exp/orderbook/dfs.go
Shaptic: There are two ways to avoid cycles, or at the very least extra hops: we either check to see if either reserve in the current pool is part of the `visited []Asset` list, or we check to see if it's a pool we've seen before.

Doing this avoids scenarios where we may have checked the rate of A -> B (for some A/B liquidity pool), but have now moved our way through the graph and are now evaluating B -> A.

exp/orderbook/dfs.go
Shaptic: great point!

exp/orderbook/batch.go
Shaptic: I thought I understood this but it's less clear now; if `liquidityPoolsForAsset` is defined as a `map` of `string -> LiquidityPoolEntry`, what's wrong with removing both entries entirely?

tamirms: `liquidityPoolsForAsset` is not a `map[string]xdr. LiquidityPoolEntry` it's a `map[string][]xdr.LiquidityPoolEntry`. Although there is only one pool for a given trading pair there could be multiple pools which share a common asset (e.g. `USD` <-> `EUR` and `USD` <-> `CHF` will both be present under the `USD` key)

Shaptic: Wow, thanks for double-checking that. I'm not sure how I misread my own code :facepalm: Now your comment makes sense :+1:

exp/orderbook/dfs.go
Shaptic: Yep! This was part of the plan -- done in latest push :+1:

exp/orderbook/batch.go
```diff
@@ -106,13 +105,59 @@ func (tx *orderBookBatchedUpdates) apply(ledger uint32) error {
 			if err := tx.orderbook.remove(operation.offerID); err != nil {
 				panic(errors.Wrap(err, "could not apply update in batch"))
 			}
+
 		case addLiquidityPoolOperationType:
-			tx.orderbook.liquidityPools[operation.liquidityPoolAssets] = *operation.liquidityPool
+			ob := tx.orderbook
```
tamirms, line 110: to be consistent with the other operations we should either encapsulate the liquidity pool operations into functions on the orderbook graph, or, we should inline `orderbook.add()` and `orderbook.remove()` into `orderBookBatchedUpdates .apply()`. I think it would be better to encapsulate the liquidity pool operations into separate functions

exp/orderbook/dfs.go
```diff
@@ -51,21 +56,27 @@ type searchState interface {
...
+	// venues returns all possible trading opportunities for a particular asset.
+	//
+	// The result is grouped by the next asset hop, mapping to a sorted list of
+	// offers (by price) and a liquidity pool (if one exists for that trading
+	// pair).
+	venues(currentAsset string) edgeSet
+
+	// chooseVenue determines whether the offer or pool amount should be chosen,
+	// returning the amount and whether or not the offer was chosen.
+	chooseVenue(offerAmount, poolAmount xdr.Int64) (xdr.Int64, bool)
```
tamirms, line 68: I think we can do a slight refactoring to make the interface easier to understand. Instead of having `chooseVenue()` , `consumeOffers()`, and `consumePool()` we could replace it with a single function:

`trade(currentAsset xdr.Asset, currentAssetAmount xdr.Int64, venues Venues) (xdr.Asset, xdr.Int64, error)`

The `trade()` function will essentially do the logic to select the best venue and then execute the trade

paulbellamy:
Sorry for coming in swinging here, but maybe fresh eyes are useful. Maybe a more object-oriented approach here would make the code easier to maintain and understand?

If we think of each venue as a graph edge, the new situation AMMs introduce is that we have multiple edges/venues between each asset pair. But you could see that being extended in future with "Concentrated Liquidity Pools" or somesuch else, like:
graph dot

So, we could model the different venue types as an interface, which would let us do something like:

```go
  // All venues implement the Venue interface
  type Venue interface {
    // Calculate how much you would receive if you sent the inputs.
    Send(
      asset xdr.Asset,
      amount xdr.Int64,
    ) (xdr.Asset, xdr.Int64, error)

    // Calculate how much you would have to send in order to receive the expected.
    Receive(
      expectedAsset xdr.Asset,
      expectedAmount xdr.Int64,
    ) (xdr.Asset, xdr.Int64, error)

    // Utility, maybe if we need
    OtherAsset(asset xdr.Asset) xdr.Asset
  }

  // Implement Venue for liquidity pools
  type LiquidityPoolVenue xdr.LiquidityPoolEntry
  func (v *LiquidityPoolVenue) OtherAsset(asset xdr.Asset) xdr.Asset
  func (v *LiquidityPoolVenue) Send(asset xdr.Asset, amount xdr.Int64) (xdr.Asset, xdr.Int64, error)
  func (v *LiquidityPoolVenue) Receive(expectedAsset xdr.Asset, expectedAmount xdr.Int64) (xdr.Asset, xdr.Int64, error)

  // Implement Venue for orderbooks
  type OrderbookVenue []xdr.OfferEntry
  func (v *OrderbookVenue) OtherAsset(asset xdr.Asset) xdr.Asset
  func (v *OrderbookVenue) Send(asset xdr.Asset, amount xdr.Int64) (xdr.Asset, xdr.Int64, error)
  func (v *OrderbookVenue) Receive(expectedAsset xdr.Asset, expectedAmount xdr.Int64) (xdr.Asset, xdr.Int64, error)

  // Implement Venue for arrays, choosing the best path. To simplify graph traversal.
  type Venues []Venue
  func (v *Venues) OtherAsset(asset xdr.Asset) xdr.Asset
  func (v *Venues) Send(asset xdr.Asset, amount xdr.Int64) (xdr.Asset, xdr.Int64, error)
  func (v *Venues) Receive(expectedAsset xdr.Asset, expectedAmount xdr.Int64) (xdr.Asset, xdr.Int64, error)

  // then we can model each edge as an array of venues
  type edgeSet map[string]Venues
```

Just an idea, but maybe it's a big change, or has too much performance overhead?

exp/orderbook/dfs.go
Shaptic: Nice idea, I'll give it a whirl.

exp/orderbook/batch.go
Shaptic: Yep, agreed. Now that these are stable (:sweat_smile:) it makes sense to abstract them like for offers.

exp/orderbook/dfs.go
Shaptic: So after trying this out, I realized why I broke it up into parts like this. If you look at the `processVenues` helper (which is what DFS uses):

https://github.com/stellar/go/blob/1c7e42b7e389c8b1b1b3b1a997d4f0e08deb5f44/exp/orderbook/dfs.go#L432-L479

The only points of divergence are at `state.consumePool`, `state.consumeOffers`, and `state.chooseVenue`, all necessary because buys and sells interact with the pool/offers differently and the results need to be evaluated differently (`<` vs. `>`).

If we combine these into a single `state.trade(...)`, then we have ~45 lines of duplicated code across the `sellingGraphSearchState` and `buyingGraphSearchState`, with divergences exactly as they are above, except I guess they wouldn't need to be attached to the `searchState`.

I agree that this can be improved, but I'm not sure this suggestion would lead to cleaner code overall. A cleaner interface, maybe, but uglier interface implementations. I did do some interface cleanup in the latest push and dropped the awkward `chooseVenue()` method. Now it's just `consumePool` and `consumeOffers`, which I think is pretty straightforward to understand.

Shaptic: (see cd60906edd07f237e1e596a060b53724293fe538 for the delta)

exp/orderbook/dfs.go
```diff
@@ -378,3 +422,45 @@ func consumeOffersForBuyingAsset(
...
+		amount, err := state.consumePool(pool, currentAsset, currentAssetAmount)
+		if err == nil {
+			nextAsset = getOtherAsset(currentAsset, pool)
+			poolAmount = amount
+		}
+		// It's only a true error if the offers fail later, too
+	}
+
+	if poolAmount == 0 && len(venues.offers) == 0 {
+		return nextAsset, 0, errNoVenues
```
tamirms, line 451: it seems like we don't check for `errNoVenues` in a special way. If would rather return `return nextAsset, 0, nil` because we handle `errNoVenues` in the same way we handle a `nextAmount` value of 0

exp/orderbook/dfs.go
```diff
@@ -378,3 +422,45 @@ func consumeOffersForBuyingAsset(
...
+) (xdr.Asset, xdr.Int64, error) {
+	var nextAsset xdr.Asset
+
+	if currentAssetAmount == 0 {
+		return nextAsset, 0, errAssetAmountIsZero
+	}
+
+	// We evaluate the pool venue (if any) before offers, because pool exchange
+	// rates can only be evaluated with an amount.
+	poolAmount := xdr.Int64(0)
```
tamirms, line 440: nit: `var poolAmount xdr.Int64`

exp/orderbook/utils.go
```diff
@@ -0,0 +1,45 @@
...
+	return false
+}
+
+// getPoolAssets retrieves string representations of a pool's reserves
+func getPoolAssets(pool xdr.LiquidityPoolEntry) (string, string) {
+	params := pool.Body.MustConstantProduct().Params
+	return params.AssetA.String(), params.AssetB.String()
+}
+
+func max(vals ...xdr.Int64) xdr.Int64 {
```
tamirms, line 23: Although this implementation of `max()` and `positiveMin()` are more generic from my testing it seems much slower than these definitions:

```
func max(a ,b xdr.Int64) xdr.Int64 {
	if a > b {
		return a
	}
	return b
}

func min(a ,b xdr.Int64) xdr.Int64 {
	if a > b {
		return b
	}
	return a
}
```

I suggest replacing the implementation with the code listed above

tamirms:
@Shaptic the code looks good to me. I just left a couple of suggestions about some minor issues. if you can fix the tests and resolve any outsanding todos / fixmes then I think we can merge this

exp/orderbook/dfs.go
Shaptic: Done in 49c81a6e!

exp/orderbook/utils.go
Shaptic: Good point. I do need `positiveMin` because otherwise `-1` gets preferred, but I did stop generalizing.

tamirms (approved):

PR #3923, opened 2021-09-15
Merge release-horizon-v2.8.2 into master

bartekn:
I noticed that a few latest release branches were not merged. Any reason why? Was it because of XDR update PR?


bartekn:
@tamirms @paulbellamy you were working on latest releases. Can you check the question above?

tamirms:
> Was it because of XDR update PR?

I think the history is a bit messed up because for the 2.8.0 release I tried to exclude the XDR update.

bartekn:
@tamirms @paulbellamy I resolved conflicts. I think it's fine to merge because XDR changes commit is now both in `master` and `amm` branch and the next release will contain `amm`. If we need to create a release without `amm` we can always create a new branch off `horizon-v2.8.2`.

bartekn:
Ping

tamirms (approved):

PR #3925, opened 2021-09-16
`check_release_hash` improvements

### LINKED ISSUE #3651: stellar/horizon: check_release_binaries could be improved somewhat

- Right now you can't use the check_release_binaries against a deb package unless it is in the stable track.
- The output is very unintuitive as to whether it passed or failed.

- The check_release_binaries dockerfile to include unstable, so we can check that before promoting to master.
- Make the output more obvious whether it failed or not

bartekn:
Several `check_release_hash` script improvements:
* Run `apt-get` commands separately. Commands were previously joined by `&&` operator which returned 0 code in case any of the commands failure (🤔).
* Hide hashes if they match.
* Add unstable repo.

Close #3651.

### DIFF
--- a/services/horizon/internal/scripts/check_release_hash/check.sh
+++ b/services/horizon/internal/scripts/check_release_hash/check.sh
@@ -1,7 +1,9 @@
 #!/bin/bash
 set -e
 
-apt-get clean && apt-get update && apt-get install -y stellar-horizon=$PACKAGE_VERSION
+apt-get clean
+apt-get update
+apt-get install -y stellar-horizon=$PACKAGE_VERSION
 
 mkdir released
 cd released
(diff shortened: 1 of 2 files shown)


services/horizon/internal/scripts/check_release_hash/check.sh
```diff
@@ -23,30 +25,57 @@ git checkout $TAG
...
+            msg="$TAG-$S NO MATCH! github=$released compile=$dist"
+        fi
     fi
-done
 
-echo "debian package"
-shasum -a 256 $(which stellar-horizon)
\ No newline at end of file
+    echo $msg
+done
```
tamirms, line 81: will it exit with a non zero status if there is a mismatch?

bartekn: No, but it didn't before this update too. I'm not sure it's required, the image probably won't be used in any automated process.

tamirms (approved):

PR #3933, opened 2021-09-17
exp/orderbook: Speed up path finding dfs by avoiding calls to consumeOffers for visited nodes

tamirms:
While I was reviewing https://github.com/stellar/go/pull/3921 , I realized we could improve the performance of the path finding DFS by moving the visited check from the beginning of the function to the moment we consider each outgoing edge. 

Previously, we were doing **both** the call to `consumeOffers()` and the visited check for each outgoing edge. Now, we only call `consumeOffers()` if we pass the visited check. We improve the performance of the dfs by removing the expensive calls to `consumeOffers()` on edges which lead to nodes we have already visited.

Benchmark before changes:

```
goos: darwin
goarch: amd64
BenchmarkMultipleDestinationAssets-12    	       2	 501200867 ns/op
BenchmarkTestData
BenchmarkTestData-12                     	       1	19009783799 ns/op
PASS
```

Benchmark after changes:

```
goos: darwin
goarch: amd64
BenchmarkMultipleDestinationAssets-12    	       3	 402958734 ns/op
BenchmarkTestData
BenchmarkTestData-12                     	       1	15575703035 ns/op
PASS
```

There is about a 25% improvement in the benchmarks.


exp/orderbook/dfs.go
```diff
@@ -73,31 +83,25 @@ func dfs(
 	if err := ctx.Err(); err != nil {
 		return err
 	}
-	if currentAssetAmount <= 0 {
```
tamirms, line 76: we don't need to check if `currentAssetAmount <= 0` because do the same check (`nextAssetAmount <= 0`) when processing outgoing edges.

exp/orderbook/dfs.go
```diff
@@ -59,11 +59,21 @@ type searchState interface {
 	) (xdr.Asset, xdr.Int64, error)
 }
 
+func contains(list []string, want string) bool {
+	for i := 0; i < len(list); i++ {
+		if list[i] == want {
```
bartekn, line 64: Any reason why not use `xdr.Asset.Equal` as before? Is string comparison much faster?

bartekn (approved):
Good find!

exp/orderbook/dfs.go
tamirms: The string comparison is faster than `xdr.Asset.Equal` but both of those operations are very fast.

The reason why I can't use `xdr.Asset.Equal` is because I only have `nextAssetString` I don't have the `xdr.Asset` form. Calling `String()` on each `xdr.Asset` in the visited list so that I can compare against `nextAssetString` would be significantly slower because of the `String()` operation

bartekn: I see, maybe we should remove `visitedAssets` then? If I'm reading the code right we can replace it with `visitedAssetStrings` everywhere.

tamirms: `appendToPaths()` requires `visitedAssets`. I think we previously discussed having an asset type which caches the `String()` result. I can try that out in a separate PR

PR #3944, opened 2021-09-20
services/horizon: Batch liquidity pool updates/removals

### LINKED ISSUE #3931: `LiquidityPoolsChangeProcessor`: batch upsert and remove objects
https://github.com/stellar/go/issues/3926

2opremio:
Fixes #3931


services/horizon/internal/ingest/processors/liquidity_pools_change_processor.go
```diff
@@ -50,64 +50,30 @@ func (p *LiquidityPoolsChangeProcessor) ProcessChange(ctx context.Context, chang
...
-			if err != nil {
-				return errors.Wrap(err, "Error creating ledger key")
-			}
-			rowsAffected, err = p.qLiquidityPools.RemoveLiquidityPool(
-				ctx, PoolIDToString(lPool.LiquidityPoolId), p.sequence,
-			)
+			lp := p.ledgerEntryToRow(change.Pre)
+			lp.Deleted = true
+			lp.LastModifiedLedger = p.sequence
+			lps = append(lps, lp)
```
bartekn, line 66: I think we should create `RemoveLiquidityPools` method and hide implementation of deleted objects (`Deleted` flag) from the caller.

2opremio: Can you be more specific? By implementation of deleted objects you mean factoring out the creation of a deleted liquidity pool value?

The deleted field is public anyhow

bartekn: > By implementation of deleted objects you mean factoring out the creation of a deleted liquidity pool value?

Exactly. If we ever change the implementation of how objects are removed (idk: we also add a deleted time or we will just remove the row) we'll have to change every instance of deletion upstream and fix a bunch of tests. I agree that `Deleted` field is public. We should probably also change this (not in this PR).

2opremio: I have given it a try, but I don't think it makes sense to attempt that sort of encapsulation without removing the Deleted field (which I think is out of the context of this PR).

Can we handle this separately?

bartekn: OK.

bartekn (approved):

Issue #3939, opened 2021-09-17
Debug ingestion speed post RebuildTradeAggregationBuckets


bartekn:
### What version are you using?

2.8.2

### What did you do?

After https://github.com/stellar/go/pull/3641 was deployed we observed ingestion duration almost doubled. 

<img width="858" alt="Screenshot 2021-07-30 at 19 58 40" src="https://user-images.githubusercontent.com/464938/133853420-d04b2606-de32-484e-ac90-cb1710302d77.png">

It's very likely that it's connected to `RebuildTradeAggregationBuckets` method because "Ingestion Overhead" metric's values also increased. Let's see if we can improve this.

PR #3947, opened 2021-09-21
services/horizon: Add a brin index on history_trades_60000(timestamp)

paulbellamy:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add a BRIN index onto `history_trades_60000`, on the `timestamp` column.

### Why

Fixes https://github.com/stellar/go/issues/3939

I ran the queries that a RebuildTradeAggregationBuckets does on staging with timings, and found:
```
finding ledger start timestamp: 0.5ms
finding ledger end timestamp: 0.5ms
deleting 40 old records: 360ms
inserting 40 new records: 3ms
```

After some `explain analyze` on the delete query, I found that the time was being taken in finding which aggregates to delete. The primary key btree index it was using to filter timestamps is not very effective. BRINs are small, and great for time-series data, so I added one. Now delete takes ~1ms.

### Known limitations

What do we want to do with the migration here? AMM also has a migration number `49`. Should we rename this one? or the other?


tamirms:
> What do we want to do with the migration here? AMM also has a migration number 49. Should we rename this one? or the other?


I think we should rename the AMM migration because this branch is closer to getting merged into master than the amm branch

2opremio (approved):

PR #3948, opened 2021-09-22
services/horizon/internal/db2/history: Fix fee source muxed account ingestion bug

tamirms:


bartekn (approved):
Thanks for fixing this!

Issue #3936, opened 2021-09-17
`ClaimableBalancesProcessor`: move XDR ingestion code to processor


bartekn:
https://github.com/stellar/go/issues/3196

bartekn:
Can be fixed with: https://github.com/stellar/go/issues/3930

PR #3953, opened 2021-09-22
services/horizon: Batch claimable balance creation, updates and deletions

### LINKED ISSUE #3930: `ClaimableBalancesChangeProcessor`: batch upsert and remove objects
https://github.com/stellar/go/issues/3926

2opremio:
Also, avoid passing XDR ledger entries and XDR Claimable Balance IDs to the claimable balance DB interface

Closes #3930 
Closes https://github.com/stellar/go/issues/3936


services/horizon/internal/db2/history/claimable_balances.go
```diff
@@ -123,23 +122,12 @@ type ClaimableBalancesBatchInsertBuilder interface {
 
 // QClaimableBalances defines account related queries.
 type QClaimableBalances interface {
-	NewClaimableBalancesBatchInsertBuilder(maxBatchSize int) ClaimableBalancesBatchInsertBuilder
-	UpdateClaimableBalance(ctx context.Context, entry xdr.LedgerEntry) (int64, error)
-	RemoveClaimableBalance(ctx context.Context, cBalance xdr.ClaimableBalanceEntry) (int64, error)
+	UpsertClaimableBalances(ctx context.Context, cb []ClaimableBalance) error
+	RemoveClaimableBalances(ctx context.Context, cBalanceIds []xdr.ClaimableBalanceId) (int64, error)
```
bartekn, line 121: We're still using `xdr.ClaimableBalanceId` for ID field. Do you think we can remove it and use strings? This will also allow us to remove `xdr.ClaimableBalanceId.Scan` and `xdr.ClaimableBalanceId.Value`. We can do something similar to `PoolID` in `amm` branch.

bartekn (approved):
LGTM! Added one suggestion about `BalanceID`.

services/horizon/internal/db2/history/claimable_balances.go
2opremio: I tried but it felt really unnatural to mix the base64 representation  from the DB and the hex representation used in the frontend.

bartekn: The DB actually is hex representation not base64.

2opremio: Ah, that's right. OK, I will give it a try.

Should I do the same for the Asset field? that one really is base64 and I think it would be odd to move an opaque base64 value around (for the BalanceID that's fine since it's a hash)

bartekn: For asset to be consistent we should probably have `db2.Asset` type with `Scan` and `Value` that encodes/decodes base64. We could also split it into `asset_type`, `asset_issuer`, `asset_code` fields but it requires schema change. Maybe let's create an issue for this because we should probably change all occurrences of assets in one go.

2opremio: OK, I removed all instances of `xdr.ClaimableBalanceId` from the history package. PTAL

bartekn (approved):

PR #3954, opened 2021-09-22
protocols/horizon: Represent liquidity pool total_shares field as amount string

tamirms:


services/horizon/internal/resourceadapter/liquidity_pool.go
```diff
@@ -28,7 +28,7 @@ func PopulateLiquidityPool(
 	}
 	dest.Type = typ
 	dest.TotalTrustlines = liquidityPool.TrustlineCount
-	dest.TotalShares = liquidityPool.ShareCount
+	dest.TotalShares = amount.String(xdr.Int64(liquidityPool.ShareCount))
```
bartekn, line 31: Nit:
```suggestion
	dest.TotalShares = amount.StringFromInt64(liquidityPool.ShareCount)
```

bartekn (approved):

PR #3965, opened 2021-09-28
exp/orderbook: Speed up path finding by changing edgeSet data structure

tamirms:
I did some profiling on path finding and found that iterating through entries in the edgeSet `map` was particularly expensive. We never use `edgeSet` to do map lookups in the dfs. We only use the data structure to iterate through the key value pairs. So I changed the representation to a pair of lists and that resulted in a significant improvement in the path finding performance.

Before applying the changes in this commit:

```
goos: darwin
goarch: amd64
BenchmarkMultipleDestinationAssets-12    	       3	 417991332 ns/op
PASS
```

After applying the changes in this commit:

```
goos: darwin
goarch: amd64
BenchmarkMultipleDestinationAssets-12    	       4	 300770334 ns/op
PASS
```


exp/orderbook/edges.go
```diff
@@ -9,21 +9,34 @@ import (
...
+}
+
+func findKey(keys []string, key string) int {
+	for i := 0; i < len(keys); i++ {
+		if keys[i] == key {
+			return i
+		}
+	}
+	return -1
+}
```
Shaptic, line 24: Bit of a nitpick, but maybe we can merge this method with the `contains()` helper we use on the visited lists?

exp/orderbook/graph.go
```diff
@@ -287,12 +277,9 @@ func (graph *OrderBookGraph) removePool(pool xdr.LiquidityPoolEntry) {
...
 		} {
-			if venues, ok := table[asset]; ok {
-				venues.removePool(otherAsset)
-				if venues.isEmpty(otherAsset) {
-					delete(venues, otherAsset)
-				}
-			} // should we panic on !ok?
+			edges := table[asset]
+			edges.removePool(otherAsset)
+			table[asset] = edges
```
Shaptic, line 286: It bothers me so much that Go doesn't let us just `table[asset].removePool(otherAsset)` :unamused:

exp/orderbook/edges.go
```diff
@@ -9,21 +9,34 @@ import (
 // edgeSet maintains a mapping of strings (asset keys) to a set of venues, which
 // is composed of a sorted lists of offers and, optionally, a liquidity pool.
 // The offers are sorted by ascending price (in terms of the buying asset).
-type edgeSet map[string]Venues
+type edgeSet struct {
+	keys   []string
+	values []Venues
+}
```
Shaptic, line 15: This is out of the scope of this change, but as an idea, we might be able to squeeze out _even more_ performance by catering towards cache hits. We always access `keys[i]` and `values[i]` together, so by keeping them close in memory we might get faster lookups, e.g.

```go
type edge struct {
	key   string
	value Venues
}
type edgeSet []edge
```

would give a key and its value cache locality.

exp/orderbook/edges.go
```diff
@@ -35,22 +48,32 @@ func (e edgeSet) addOffer(key string, offer xdr.OfferEntry) {
...
-	venues := e[key]
-	offers := venues.offers
+func (e *edgeSet) removeOffer(key string, offerID xdr.Int64) bool {
+	i := findKey(e.keys, key)
+	if i < 0 {
+		return false
+	}
 
+	offers := e.values[i].offers
+	updatedOffers := offers
```
Shaptic, line 76: Can we modify `offers` directly?

exp/orderbook/graph.go
```diff
@@ -258,16 +248,20 @@ func (graph *OrderBookGraph) removeOffer(offerID xdr.Int64) error {
...
 
 	if set, ok := graph.venuesForBuyingAsset[pair.buyingAsset]; !ok {
 		return errOfferNotPresent
 	} else if !set.removeOffer(pair.sellingAsset, offerID) {
 		return errOfferNotPresent
-	} else if len(set) == 0 {
+	} else if len(set.values) == 0 {
 		delete(graph.venuesForBuyingAsset, pair.buyingAsset)
+	} else {
+		graph.venuesForBuyingAsset[pair.buyingAsset] = set
```
Shaptic, line 264: Wait, isn't this already true? That is, isn't `graph.venuesForBuyingAsset[pair.buyingAsset] == set` at the start of this `else` block?

exp/orderbook/graph_test.go
```diff
@@ -320,6 +321,15 @@ func assertPathEquals(t *testing.T, a, b []Path) {
...
 }
 
+func findByAsset(edges edgeSet, asset string) Venues {
+	for i, key := range edges.keys {
+		if key == asset {
+			return edges.values[i]
+		}
+	}
+	return Venues{}
+}
```
Shaptic, line 331: This can use the other helper `findKey`, right?

```suggestion
func findByAsset(edges edgeSet, asset string) Venues {
	i := findKey(edges.keys, asset)
	if i >= 0 {
		return edges.values[i]
	}
	return Venues{}
}
```

Shaptic (commented):
> We never use edgeSet to do map lookups in the dfs.

I think I passively observed this during the latest changeset, but was too deep in a refactor to give it proper consideration. Super exciting to see a follow-up on it! :tada: 

A few minor DRY-related comments and other questions below :arrow_down:

exp/orderbook/graph.go
tamirms: no, because removeOffer modifies the set

exp/orderbook/edges.go
tamirms: I didn't want to modify `offers` while iterating through it in the `for i, offer := range offers` loop

Shaptic (approved):
LGTM! I'm curious if you saw a boost after going for the `[]edge` approach?

tamirms:
> I'm curious if you saw a boost after going for the []edge approach?

There wasn't a noticeable difference but the code was cleaner using an `edge` struct so I kept it.

PR #3967, opened 2021-09-29
.vscode: fix tasks files again

leighmcculloch:
### What
Fix VSCode task file again.

### Why
VSCode keeps saying the file is out of date and needs updating on every open which creates permanent diffs in local checkouts. In #3908 I committed a fixed VSCode file, but it must have only been partial.


leighmcculloch:
I think @bartekn's previous comments about should we even have this file still is valid. I'm not sure. I'm also not sure the consequence of us deleting it. But at the very least this PR will stop the diffs.

bartekn (approved):

PR #3974, opened 2021-09-30
services/horizon: Update `asset_type` for trust line sponsorship effects on fly

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

29c16d3 adds a new `asset_type` field to trust line sponsorship effects in order to distinguish asset trust lines from liquidity pool trust lines. There is an issue with this approach: effects ingested before 2.9.0 release will not have these fields set so an effect response will contain empty `asset_type` JSON field. To solve it, we use `asset` value to derive `asset_type` in `Effect.UnmarshalDetails` method.

### DIFF
--- a/services/horizon/internal/db2/history/effect.go
+++ b/services/horizon/internal/db2/history/effect.go
@@ -8,6 +8,7 @@ import (
 
 	sq "github.com/Masterminds/squirrel"
 
+	"github.com/stellar/go/protocols/horizon/effects"
 	"github.com/stellar/go/services/horizon/internal/db2"
 	"github.com/stellar/go/services/horizon/internal/toid"
 	"github.com/stellar/go/support/errors"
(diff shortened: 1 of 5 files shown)


tamirms (approved):

services/horizon/internal/db2/history/effect.go
```diff
@@ -19,7 +20,37 @@ func (r *Effect) UnmarshalDetails(dest interface{}) error {
 		return nil
 	}
 
-	return errors.Wrap(json.Unmarshal([]byte(r.DetailsString.String), &dest), "unmarshal effect details failed")
+	err := errors.Wrap(json.Unmarshal([]byte(r.DetailsString.String), &dest), "unmarshal effect details failed")
+	if err == nil {
```
2opremio, line 24: Nit: it's much more readable to return if there is an error early, saving us from having indentation later on

```go
  if err := json.Unmarshal([]byte(r.DetailsString.String), &dest); err != nil {
    return  errors.Wrap(err, "unmarshal effect details failed")
  }
  // In 2.9.0 a new `asset_type` was introduced to include liquidity
  // pools. Instead of reingesting entire history, let's fill the
  // `asset_type` here if it's empty.
  // (I hate to convert to `protocol` types here but there's no other way
  // without larger refactor.)
  switch dest := dest.(type) {
   ...
  }
  return nil
}
```

PR #3975, opened 2021-09-30
exp/orderbook: Add parameter to exclude pools from path finding search

tamirms:


bartekn (approved):
LGTM but please wait for @Shaptic review. While I understand what's happening here I'm not up to date with previous changes to LP path finding. I'll try to catchup tomorrow.

exp/orderbook/graph_test.go
```diff
@@ -2059,7 +2104,7 @@ func TestInterleavedPaths(t *testing.T) {
 		InteriorNodes:     []xdr.Asset{usdAsset, eurAsset},
 	}, {
 		SourceAsset:       chfAsset,
-		SourceAmount:      58,
+		SourceAmount:      53,
```
Shaptic, line 2107: This happened to me once when adding LP support, but I thought I was just tripping out. Any idea why it happened here?

exp/orderbook/pools_test.go
```diff
@@ -37,7 +37,7 @@ func TestLiquidityPoolExchanges(t *testing.T) {
 		}{
 			{yenAsset, eurYenLiquidityPool, 100, 112},
 			{eurAsset, eurUsdLiquidityPool, 112, 127},
-			{nativeAsset, nativeUsdPool, 5, 27},
+			{nativeAsset, nativeUsdPool, 5, 25},
```
Shaptic, line 40: Hmm... path-finding aside, unless the pool spec changed, this shouldn't change. I'm wary of this delta.. are we accidentally modifying the pool by reference somewhere?

services/horizon/internal/flags.go
```diff
@@ -364,6 +364,14 @@ func Flags() (*Config, support.ConfigOptions) {
 			FlagDefault: uint(3),
 			Usage:       "the maximum number of assets on the path in `/paths` endpoint, warning: increasing this value will increase /paths response time",
 		},
+		&support.ConfigOption{
+			Name:        "enable-pool-path-finding",
+			ConfigKey:   &config.EnablePoolPathFinding,
+			OptType:     types.Bool,
+			FlagDefault: true,
```
Shaptic, line 371: Kind of a nitpick, but I feel that seeing `--enable=true`-type flags by default can sometimes be confusing.

As the CLI user, I'd prefer `--disable=false` to be the default. The rationale is that if I'm reading the `--help`, I think "ooh, lemme enable the thing," but when I do `./horizon --enable-pools=true` I see no behavioral change from just running `./horizon`. However, `./horizon --disable-pools` will actually change things, and indicates you're turning a default off.

Shaptic (commented):
Structure, LGTM, but I'm concerned about the LP math suddenly changing.

exp/orderbook/graph_test.go
tamirms: See the first commit https://github.com/stellar/go/pull/3975/commits/f750a0f225b97628a485e2a317cc44ec851ad067

`TestUpdateLiquidityPools` and `TestRemoveLiquidityPools` modified the state of some of the pools which are declared as global variables and used in a bunch of test cases. So if you ran all the tests in the file you would get one result but if you ran `TestInterleavedPaths` and `TestLiquidityPoolExchanges` individually you would get another result.

I fixed this by making sure that `TestUpdateLiquidityPools` and `TestRemoveLiquidityPools` do not modify the globally declared test pools. After this change there are no side effects from `TestUpdateLiquidityPools` and `TestRemoveLiquidityPools` and the test results are consistent whether you run all the tests together or individually.

exp/orderbook/pools_test.go
tamirms: https://github.com/stellar/go/pull/3975#discussion_r719977968

PR #3977, opened 2021-10-01
services/horizon: Use operation body to generate "claimant created" effect

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Switched to using operation body conditionally to fetch claimable balance predicated.

### Why

In https://github.com/stellar/go/commit/29c16d31daa7debe6a5bb8b1e3354a8c8dc06c22 `EffectClaimableBalanceClaimantCreated` started to be generated solely from tx meta. While this approach is fine it changes the way this effect details are ingested compared to 2.8.3. This is happening because all `rel_before` predicates present in operation body are changed to `abs_before` by Stellar-Core when the tx is included in the ledger.

### Known limitations

N/A


paulbellamy (approved):

paulbellamy:
Test would be nice...

services/horizon/internal/ingest/processors/effects_processor.go
```diff
@@ -959,7 +959,20 @@ func (e *effectsWrapper) addClaimableBalanceEntryCreatedEffects(source *xdr.Muxe
...
 		details,
 	)
-	for _, c := range cb.Claimants {
+	// EffectClaimableBalanceClaimantCreated can be generated by
+	// `create_claimable_balance` but also by `liquidity_pool_withdraw` which
+	// also removes LP.
+	// In case of `create_claimable_balance` we use `op.Claimants` to make
+	// effects backward compatible. The reason for this is that Stellar-Core
+	// changes all `rel_before` predicated to `abs_before` so effects generated
+	// by Horizon 2.8.3 and 2.9.0 are different.
```
2opremio, line 968: Could you rephrase this so that future readers can understand it better? In particular what is `abs_before` and `rel_before` and why are they different depending on where we obtain the claimants from?

I have also proposed some other changes in the comment. Among other things I don't think that the versions are important, since we are releasing with backwards compatibility (but they do add to the confusion).

```suggestion
	// EffectClaimableBalanceClaimantCreated can be generated by
	// `create_claimable_balance` operation but also by `liquidity_pool_withdraw` operation
	// causing a revocation.
	// In case of `create_claimable_balance` we use `op.Claimants` to make
	// effects backward compatible. The reason for this is that Stellar-Core
	// changes all `rel_before` predicated to `abs_before`.
```

2opremio: Actually, including the explanation from the PR description would be enough:

> This is happening because all rel_before predicates present in operation body are changed to abs_before by Stellar-Core when the tx is included in the ledger.

bartekn:
OK, added a test in https://github.com/stellar/go/pull/3977/commits/f3e8e8b635ba632b92c8d9af25523fe423a9964b. I thought that behaviour is already tested but it really wasn't.

Also, rephrased the comment.

bartekn:
Run `verify-range` on a small affected range and confirmed the fix is working.

Issue #3982, opened 2021-10-05
Handle negative sequences gracefully


2opremio:
1. Txnbuild should error when incrementing a transaction sequence (`Int64` ) would result in an wraparound (overflow). When a transaction sequence gets to MaxInt64 it simply keeps increasing the sequence, resulting in a wrap around (i.e. `MaxInt64` ++ results in `MinInt64`)

2. Horizon should handle submitted transactions with negative sequences properly
    * It currently times out (it gets stuck in `sys.waitUntilAccountSequence` until reaching the request context deadline)
    * The submission system assumes that sequences are always positive (it casts incoming sequences to `uint64s` which, for negative sequences results in unplanned bahavior, e.g. -9223372036854775808 gets transform 9223372036854775808)

PR #3983, opened 2021-10-05
Handle negative transaction sequences gracefully

2opremio:
Also: Create integration test for negative tx sequence submission

Closes https://github.com/stellar/go/issues/3982

### DIFF
--- a/protocols/horizon/main.go
+++ b/protocols/horizon/main.go
@@ -5,6 +5,8 @@ package horizon
 import (
 	"encoding/base64"
 	"encoding/json"
+	"fmt"
+	"math"
 	"math/big"
 	"strconv"
 	"time"
@@ -110,8 +112,11 @@ func (a *Account) IncrementSequenceNumber() (int64, error) {
 	if err != nil {
 		return 0, err
 	}
+	if seqNum == int64(math.MaxInt64) {
+		return 0, fmt.Errorf("sequence cannot be increased, it already reached MaxInt64 (%d)", int64(math.MaxInt64))
+	}
 	seqNum++
-	a.Sequence = strconv.FormatInt(int64(seqNum), 10)
+	a.Sequence = strconv.FormatInt(seqNum, 10)
 	return seqNum, nil
 }
 
--- a/services/horizon/internal/txsub/system.go
+++ b/services/horizon/internal/txsub/system.go
@@ -101,6 +101,11 @@ func (sys *System) Submit(
 		"tx":      rawTx,
 	}).Info("Processing transaction")
 
+	if envelope.SeqNum() < 0 {
+		sys.finish(ctx, hash, response, Result{Err: ErrBadSequence})
+		return
+	}
+
 	tx, sequenceNumber, err := checkTxAlreadyExists(ctx, db, hash, sourceAddress)
 	if err == nil {
 		sys.Log.Ctx(ctx).WithField("hash", hash).Info("Found submission result in a DB")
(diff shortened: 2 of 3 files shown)


services/horizon/internal/integration/protocol18_test.go
```diff
@@ -759,3 +760,61 @@ func TestLiquidityPoolFailedDepositAndWithdraw(t *testing.T) {
...
+		BaseFee:              txnbuild.MinBaseFee,
+		Timebounds:           txnbuild.NewInfiniteTimeout(),
+		IncrementSequenceNum: false,
+		EnableMuxedAccounts:  true,
+	}
+	tx, err := txnbuild.NewTransaction(txParams)
+	tt.NoError(err)
+	tx, err = tx.Sign(integration.StandaloneNetworkPassphrase, master)
+	tt.NoError(err)
+	txResp, err = itest.Client().SubmitTransaction(tx)
```
2opremio, line 817: This shouldn't fail, but it does

services/horizon/internal/txsub/system.go
```diff
@@ -113,6 +113,16 @@ func (sys *System) Submit(
...
 
+	// TODO: the submission queue system doesn't play well with negative sequence numbers
+	//       which are possible since the XDR uses Int64 (see the uint64(envelope.SeqNum()) casts below
+	//       which will translate sequence -9223372036854775808 into 9223372036854775808).
+	// We could do a bunch of things
+	// 1. Reject transactions with negative values. But this seems excesive since we should just be an intermediary with
+	//    core
+	// 2. Adapt the queuing system to use in64 instead of uint64 ... but that comes with its own problems
+	//    (the nextSequence logic should be handled carefully to avoid overflows)
+	// 3. Avoid the queuing system altogether for negative sequences (which we know are going to fail anyhow).
```
2opremio, line 124: @bartekn @tamirms I would appreciate your input here

2opremio: I am thinking that it Core rejects transactions with negative sequence numbers (i.e. they don't even get to be ingested as failed transactions). We could just reject them right away. Thoughts?

2opremio: Actually, the submission system already sends a Badsequence error without contacting core at all, so I think I will just do that

```
	// queue the submission and get the channel that will emit when
	// submission is valid
	seq := sys.SubmissionQueue.Push(sourceAddress, uint64(envelope.SeqNum()))

	// update the submission queue with the source accounts current sequence value
	// which will cause the channel returned by Push() to emit if possible.
	sys.SubmissionQueue.Update(map[string]uint64{
		sourceAddress: sequenceNumber,
	})

	select {
	case err := <-seq:
		if err == sequence.ErrBadSequence {
			// convert the internal only ErrBadSequence into the FailedTransactionError
			err = ErrBadSequence
		}

		if err != nil {
			sys.finish(ctx, hash, response, Result{Err: err})
			return
		}
```

bartekn (approved):

2opremio:
I agree about the change being small enough to be included in 2.9.0 , let's merge it!

protocols/horizon/main.go
```diff
@@ -110,8 +112,11 @@ func (a *Account) IncrementSequenceNumber() (int64, error) {
 	if err != nil {
 		return 0, err
 	}
+	if seqNum == int64(math.MaxInt64) {
```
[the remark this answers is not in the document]
2opremio: (i.e. untyped consts don't seem to exit)

PR #4035, opened 2021-10-28
Add `is_source` query parameter to transaction endpoints

### LINKED ISSUE #3688: Feature Request: Get Transactions for Account where Account is the Source
### Feature Request

Ability to query for Transactions where a given account is the _Source Account_ of the transaction using Horizon.

Currently, there is no Horizon way to query for Transactions by the _source account_. The `accounts/:account_id/transactions` endpoint functions by loading all transactions where the given account has participated in the transaction.

This means that to determine Transactions where an account is a source account, you have to wade through all transactions an account has participated in. This can be especially problematic with the current reality of "ad accounts" spamming 1 Stroop payments to an account. If this account is somewhat centrally related to an application or protocol, it also becomes a target of a 1 Stroop payment spam attack that could cause performance/stability issues for the app.

A new parameter on the `accounts/:account_id/transactions` to allow for only fetching transactions where the account is the source.

```
accounts/:account_id/transactions?[current query params]&is_source{true,false}

is_source would be optional and default to false
```

I don't really know Go - but a code change suggestion could be for `/services/horizon/internal/db2/history/transaction.go#L67`:
```
// ForAccount filters the transactions collection to a specific account
func (q *TransactionsQ) ForAccount(ctx context.Context, aid string, bool isSource) *TransactionsQ  {	
    var account Account
    q.Err = q.parent.AccountByAddress(ctx, &amp;account, aid)
    if q.Err != nil { 
        return q
    }
    
    if isSource {
        q.sql = selectTransaction.Where("ht.account = ?", account.ID)
    }
    else {
        q.sql = q.sql.
            Join("history_transaction_participants htp ON htp.history_transaction_id = ht.id").
            Where("htp.history_account_id = ?", account.ID)
   }
        
    return q
}
```
I want to note that I am not sure how performant this query might be. It might require some indexing of the `account` field of the `history_transactions` table to truly be efficient.

If the above query is not performant, an option to add an `is_source` column to the `history_transaction_participants` table could be explored to allow the query to still function in mostly the same manner as it does today. The disadvantage here would be to add more data to an already large table.

2opremio:
~I'm happy to use a different name for the endpoint.~

~I used `transactions_by_source` to distinguish it from the existing endpoint `GET /accounts/{account_id}/transactions`~

Closes #3688


services/horizon/internal/db2/schema/migrations/51_transactions_by_source_account.sql
```diff
@@ -0,0 +1,17 @@
+-- +migrate Up
+
+-- unused indices
+DROP INDEX by_account;
+DROP INDEX by_fee_account;
```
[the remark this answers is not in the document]
Shaptic: > by_account

I thought this would be used by `GET /account/:account_id/transactions`, but it looks like it joins/filters on the `history_transaction_participants` table, instead. Seems ok to remove, but is there a safer way to check that?

>  by_fee_account

I could've sworn I recalled either a PR or an issue that lets people filter fee-bump transactions by their fee source... can't find it now, maybe someone else remembers.

services/horizon/internal/actions/transaction.go
```diff
@@ -58,6 +58,7 @@ func (handler GetTransactionByHashHandler) GetResource(w HeaderWriter, r *http.R
 // TransactionsQuery query struct for transactions end-points
 type TransactionsQuery struct {
 	AccountID                 string `schema:"account_id" valid:"accountID,optional"`
+	SourceAccountID           string `schema:"source_account_id" valid:"accountID,optional"`
```
Shaptic, line 61: Wait, I think that's exactly what you've done here. Disregard if that's the case, but I'm leaving the comment here just in case.

Shaptic (approved):
Seems good to me, minor notes below :arrow_down:

services/horizon/internal/db2/schema/migrations/51_transactions_by_source_account.sql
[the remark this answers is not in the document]
tamirms: I haven't find any cases where the indexes are used. We can monitor the latencies in staging with mirroring on to verify there is no performance degradation

tamirms (approved):

services/horizon/internal/httpx/router.go
```diff
@@ -226,6 +226,7 @@ func (r *Router) addRoutes(config *RouterConfig, rateLimiter *throttled.HTTPRate
 		}, streamHandler))
 		r.With(historyMiddleware).Method(http.MethodGet, "/accounts/{account_id:\\w+}/trades", streamableHistoryPageHandler(ledgerState, actions.GetTradesHandler{LedgerState: ledgerState, CoreStateGetter: config.CoreGetter}, streamHandler))
 		r.With(historyMiddleware).Method(http.MethodGet, "/accounts/{account_id:\\w+}/transactions", streamableHistoryPageHandler(ledgerState, actions.GetTransactionsHandler{LedgerState: ledgerState}, streamHandler))
+		r.With(historyMiddleware).Method(http.MethodGet, "/accounts/{source_account_id:\\w+}/transactions_by_source", streamableHistoryPageHandler(ledgerState, actions.GetTransactionsHandler{LedgerState: ledgerState}, streamHandler))
```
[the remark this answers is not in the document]
bartekn: I agree with this. This is filtering and in most cases we add filters via GET params.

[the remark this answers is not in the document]
2opremio: Done

2opremio:
@bartekn @Shaptic  PTAL

services/horizon/CHANGELOG.md
```diff
@@ -3,6 +3,17 @@
...
+# unreleased
+
+### DB Schema Migration
+
+* This release comes with a DB migration removing unused indexes (`by_account` and `by_fee_account`) and adding a new index (`history_transactions_by_source_account`). It should not take more than 15 minutes to complete the migration.
+
+### Changes
+
+* Add `is_source=true` parameter to transaction endpoints (e.g. `GET /accounts/{account_id}/transactions?is_source=true` ) to
+  allow filtering transactions based on their source account.
```
bartekn, line 15: New line here a mistake?

services/horizon/internal/db2/schema/migrations/51_transactions_by_source_account.sql
```diff
@@ -0,0 +1,17 @@
...
+
+-- recreate by_account index without the sequence, which will be used to query transactions by_source_account
+CREATE INDEX history_transactions_by_source_account ON history_transactions USING btree (account);
```
bartekn, line 8: We already have multicolumn index which contains `account`:
```
"by_account" btree (account, account_sequence)
```
So I wonder if we need a new one. Postgres [docs](https://www.postgresql.org/docs/13/indexes-multicolumn.html) are not super clear on this but it seems queries should be fast if there are equality constraints on leading columns (so `account` in our case).

bartekn (approved):
LGTM!

services/horizon/internal/db2/schema/migrations/51_transactions_by_source_account.sql
2opremio: Why have an index on multiple columns if you only need it for one? (it would take less space)

services/horizon/CHANGELOG.md
2opremio: The line was too long and it shouldn't make a difference in rendering.

services/horizon/internal/db2/schema/migrations/51_transactions_by_source_account.sql
bartekn: Oh, we're removing it! I get it, sorry.

services/horizon/CHANGELOG.md
leighmcculloch, line 15: Does this filter by transaction source account only, or operation source account too, and what about fee account? Is the goal to find transactions that only consume sequence numbers of the account (transaction source account), or transactions where the account chose to do something (transaction source account, operation source account, fee account). cc @tyvdh

leighmcculloch: Moved this discussion to https://github.com/stellar/go/issues/3688#issuecomment-961930998

2opremio: It filters by transaction source account.

services/horizon/internal/db2/schema/migrations/51_transactions_by_source_account.sql
2opremio: I used 15 minutes as an estimation, I didn't get to meassure it (we will find out when deploying to staging as part of testing)

PR #4040, opened 2021-11-01
keypair: add BinaryMarsheler support to FromAddress

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Make `FromAddress` compatible with the `BinaryMarshaler` interface.

### Why

It makes the type compatible with encoders/decoders that support the `BinaryMarshaler` interface such as the `encoding/gob` package.

### Known limitations

N/A

### DIFF
--- a/keypair/from_address.go
+++ b/keypair/from_address.go
@@ -102,3 +102,31 @@ func (kp *FromAddress) UnmarshalText(text []byte) error {
 func (kp *FromAddress) MarshalText() ([]byte, error) {
 	return []byte(kp.address), nil
 }
+
+var (
+	_ = encoding.BinaryMarshaler(&FromAddress{})
+	_ = encoding.BinaryUnmarshaler(&FromAddress{})
+)
+
+func (kp *FromAddress) UnmarshalBinary(b []byte) error {
+	accountID := xdr.AccountId{}
+	err := xdr.SafeUnmarshal(b, &accountID)
+	if err != nil {
+		return err
+	}
+	address := accountID.Address()
+	binKP, err := ParseAddress(address)
+	if err != nil {
+		return err
+	}
+	*kp = *binKP
+	return nil
+}
+
+func (kp *FromAddress) MarshalBinary() ([]byte, error) {
+	accountID, err := xdr.AddressToAccountId(kp.address)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


acharb (commented):
this lgtm 👍 . I think to be safe someone from Horizon should also take a look though since `FromAddress` is widely used. Can you request someone from their team to also take a look?

keypair/from_address.go
```diff
@@ -102,3 +102,31 @@ func (kp *FromAddress) UnmarshalText(text []byte) error {
...
+	*kp = *binKP
+	return nil
+}
+
+func (kp *FromAddress) MarshalBinary() ([]byte, error) {
+	accountID, err := xdr.AddressToAccountId(kp.address)
+	if err != nil {
+		return nil, err
+	}
+	return accountID.MarshalBinary()
```
leighmcculloch, line 131: @stellar/horizon-committers @bartekn How would you feel if I didn't use XDR for this, and instead used the compressed XDR that you folks use internally in Horizon? i.e. instead of 4 byte discriminant, it'd be a 1 byte discriminant.

bartekn: I think it's OK. It's not `xdr` type so it doesn't need to follow strict XDR encoding. Just make sure it panics on types taking more than 1 byte.

leighmcculloch: Should we anticipate types of greater than one byte? If so, then I should just leave it as it is I think.

bartekn: You never know. XDR supports this so it is possible.

leighmcculloch: Although, according to SEP-23, it is always a byte. https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0023.md

bartekn: Yeah, it is now but SEP can be changed any time because of protocol upgrade etc. We may miss updating the code here because it's not in XDR package. A safe check can prevent potential issues in the future.

leighmcculloch: Once we commit this we shouldn't change it, so lets stick with XDR.

bartekn (approved):

PR #4043, opened 2021-11-04
services/horizon: Add an endpoint that allows querying for which liquidity pools an account is participating in

erika-sdf:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

`GET /liquidity_pools?account=<id>`returns a list of pools that have <id> as a participant

### Why

Getting this information in the current api requires querying an account for its balances to obtain pool ids, then querying each liquidity pool by id. This simplifies that look up into a single call.

### Known limitations

This change does not allow requests that specify both account if and existing reserves parameter in the same request.


services/horizon/internal/db2/history/liquidity_pools.go
```diff
@@ -150,8 +155,30 @@ func (q *Q) FindLiquidityPoolByID(ctx context.Context, liquidityPoolID string) (
...
+
+// GetLiquidityPools finds all liquidity pools where accountID owns assets
 func (q *Q) GetLiquidityPools(ctx context.Context, query LiquidityPoolsQuery) ([]LiquidityPool, error) {
+	log.Infof("%+v", query)
+	if len(query.AccountID) > 0 && len(query.Assets) > 0 {
+		return nil, fmt.Errorf("only one of `account id` or `assets` can be specified in a liquidity pool request")
+	}
+
+	if len(query.AccountID) > 0 {
+		log.Infof("find by account id")
```
Shaptic, line 178: Looks like some of our debug code snuck in

services/horizon/internal/db2/history/liquidity_pools.go
```diff
@@ -219,3 +246,71 @@ var liquidityPoolsSelectStatement = "lp.id, " +
 	"lp.last_modified_ledger"
 
 var selectLiquidityPools = sq.Select(liquidityPoolsSelectStatement).From("liquidity_pools lp")
+var selectLiquidityPoolsJoinedTrustlines = selectLiquidityPools.LeftJoin("trust_lines ON id = liquidity_pool_id")
+
+// MakeTestPool is a helper to make liquidity pools for testing purposes. It's
+// public because it's used in other test suites.
+func MakeTestPool(A xdr.Asset, a uint64, B xdr.Asset, b uint64) LiquidityPool {
```
Shaptic, line 253: note for reviewers: these are intentionally public as they're used both in these tests and in the actions' tests.

services/horizon/internal/db2/history/liquidity_pools.go
```diff
@@ -150,8 +155,30 @@ func (q *Q) FindLiquidityPoolByID(ctx context.Context, liquidityPoolID string) (
...
+	}
+
+	return results, nil
+}
+
+// GetLiquidityPools finds all liquidity pools where accountID owns assets
 func (q *Q) GetLiquidityPools(ctx context.Context, query LiquidityPoolsQuery) ([]LiquidityPool, error) {
+	log.Infof("%+v", query)
+	if len(query.AccountID) > 0 && len(query.Assets) > 0 {
+		return nil, fmt.Errorf("only one of `account id` or `assets` can be specified in a liquidity pool request")
```
Shaptic, line 174: nit: wrong param names here. We can also stay consistent with other error message (see e.g. `TradesQuery.Validate()`):
> `errors.New("this endpoint does not support filtering by both asset pair and offer id ")`

```suggestion
		return nil, fmt.Errorf("this endpoint does not support filtering by both account ID and reserve assets.")
```

Shaptic (commented):
obviously can't :heavy_check_mark: this but some minor notes

services/horizon/internal/db2/history/liquidity_pools.go
```diff
@@ -150,8 +155,30 @@ func (q *Q) FindLiquidityPoolByID(ctx context.Context, liquidityPoolID string) (
...
+	if err := q.Select(ctx, &results, sql); err != nil {
+		return nil, errors.Wrap(err, "could not run select join query")
+	}
+
+	return results, nil
+}
+
+// GetLiquidityPools finds all liquidity pools where accountID owns assets
 func (q *Q) GetLiquidityPools(ctx context.Context, query LiquidityPoolsQuery) ([]LiquidityPool, error) {
+	log.Infof("%+v", query)
```
tamirms, line 172: probably want to remove this debug log line as well

services/horizon/internal/db2/history/liquidity_pools.go
```diff
@@ -219,3 +246,71 @@ var liquidityPoolsSelectStatement = "lp.id, " +
...
+	if poolId == "" {
+		trustline.AssetType = asset.Type
+		switch asset.Type {
+		case xdr.AssetTypeAssetTypeNative:
+			trustline.AssetCode = "native"
+
+		case xdr.AssetTypeAssetTypeCreditAlphanum4:
+			fallthrough
+		case xdr.AssetTypeAssetTypeCreditAlphanum12:
+			fmt.Println("Code:", asset.GetCode())
```
tamirms, line 297: probably want to remove the print statements here and below

services/horizon/internal/db2/history/liquidity_pools.go
erika-sdf: Removed these debug statements.

services/horizon/internal/db2/history/liquidity_pools.go
erika-sdf: Updated the error message here.

services/horizon/internal/db2/history/liquidity_pools.go
erika-sdf: Removed

services/horizon/internal/db2/history/liquidity_pools.go
erika-sdf: Also removed

services/horizon/internal/actions/liquidity_pool_test.go
```diff
@@ -142,48 +92,79 @@ func TestGetLiquidityPools(t *testing.T) {
...
+			map[string]string{"account": accountId},
+			map[string]string{},
+			q,
+		)
+		assert.Contains(t, request.URL.String(), fmt.Sprintf("account=%s", accountId))
+
+		handler := GetLiquidityPoolsHandler{}
+		response, err := handler.GetResourcePage(httptest.NewRecorder(), request)
+		assert.NoError(t, err)
+		assert.Len(t, response, 1)
```
bartekn, line 168: Worth checking the actual response if the LP is the one we expect.

services/horizon/internal/actions/liquidity_pool.go
```diff
@@ -61,13 +61,14 @@ func (handler GetLiquidityPoolByIDHandler) GetResource(w HeaderWriter, r *http.R
...
 	Reserves string `schema:"reserves" valid:"optional"`
+	Account  string `schema:"account" valid:"optional"`
 
 	reserves []xdr.Asset
 }
 
 // URITemplate returns a rfc6570 URI template the query struct
 func (q LiquidityPoolsQuery) URITemplate() string {
-	return "/liquidity_pools?{?reserves}"
+	return "/liquidity_pools?{?reserves,?account}"
```
bartekn, line 71: I think we don't prepend with `?` in any other template.
```suggestion
	return "/liquidity_pools?{?reserves,account}"
```

services/horizon/internal/db2/history/liquidity_pools_test.go
```diff
@@ -8,30 +8,17 @@ import (
...
-				Reserve: 450,
-			},
-			{
-				Asset:   xdr.MustNewNativeAsset(),
-				Reserve: 450,
-			},
-		},
-		LastModifiedLedger: 123,
-	}
+	lp := MakeTestPool(usdAsset, 450, xlmAsset, 450)
```
bartekn, line 21: I can't see a test checking `Account` filter. Maybe worth adding one?

services/horizon/internal/db2/history/liquidity_pools.go
```diff
@@ -150,8 +154,28 @@ func (q *Q) FindLiquidityPoolByID(ctx context.Context, liquidityPoolID string) (
...
+}
+
+// GetLiquidityPools finds all liquidity pools where accountID owns assets
 func (q *Q) GetLiquidityPools(ctx context.Context, query LiquidityPoolsQuery) ([]LiquidityPool, error) {
+	if len(query.Account) > 0 && len(query.Assets) > 0 {
+		return nil, fmt.Errorf("this endpoint does not support filtering by both accountID and reserve assets.")
+	}
+
+	if len(query.Account) > 0 {
+		return q.findLiquidityPoolsByAccountId(ctx, query.Account)
```
bartekn, line 176: I don't think there's a need to extract this to a separate method. First, it's easy to miss stuff, for example currently the code does not apply cursor so pagination probably doesn't work. Second, it actually duplicates code: `deleted` where clause is added in two methods.

I think we can rewrite `selectLiquidityPools` to have named table `lp`. Then in a single `if` statement we can set the query we want to work on and add filters.

services/horizon/internal/db2/history/liquidity_pools_test.go
Shaptic: I was curious about this (did most of the test work): the db code path is already checked by using the actions code path. It seems like the tests are very similar to the action ones. Is there a benefit to having them in both places?

bartekn: The main idea we had when doing a refactoring was that `actions` package will not use DB directly but mocks. And because we have integration tests framework, the entire stack should be tested there. Unfortunately previous action tests did not have this rule so quite a lot new tests access DB too.

I think it's fine if we test it in one place. What we could also do instead here is add a very simple (a few lines of code) test sending a new query and ensuring the result is correct in `TestLiquidityPoolHappyPath` test.

services/horizon/internal/db2/history/liquidity_pools.go
erika-sdf: A test has been added here to test querying by account id. Thanks for @Shaptic for helping debug.

services/horizon/internal/db2/history/liquidity_pools_test.go
erika-sdf: A test has been added here to test querying by account id. Thanks for @Shaptic for helping debug.

services/horizon/internal/actions/liquidity_pool_test.go
erika-sdf: This check has been added.

bartekn (approved):
Great!

PR #4046, opened 2021-11-04
services/horizon: Add --max-assets-per-path-request flag

erika-sdf:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Introduce a flag that sets the max number of assets to consider during strict-send or strict-receive.

### Why

The maximum (15) is hardcoded, and making it configurable grants users some flexibility.

### Known limitations

N/A


services/horizon/CHANGELOG.md
```diff
@@ -7,7 +7,8 @@ file. This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changes
 
-* Add an endpoint that allows querying for which liquidity pools an account is participating in 
+* Add a new feature flag `--max-assets-per-path_request` (`15` by default) that sets the number of assets to consider for strict-send and strict-recieve ([4046](https://github.com/stellar/go/pull/4046))
```
tamirms, line 10: nit: I would replace "feature flag" with "horizon flag"

tamirms (approved):

services/horizon/CHANGELOG.md
bartekn:
```suggestion
* Add a new feature flag `--max-assets-per-path-request` (`15` by default) that sets the number of assets to consider for strict-send and strict-recieve ([4046](https://github.com/stellar/go/pull/4046))
```

bartekn (approved):

marcinx:
Is it possible to bump up the default though? 15 seems arbitrary and looking at response times of requests with 15 assets, the pathfinder should be able to handle a multiple of that easily.

Trying to avoid situations where we use, e.g. 100, on our deployment and are not able to fall back to SDF Horizon because of a much lower constraint.

services/horizon/CHANGELOG.md
erika-sdf: Done

ire-and-curses:
@marcinx We're not going to change this number for SDF's Horizon in the near future. It's unclear what the impact of raising the number is for our load. We have a large number of requests, and even small performance changes have large consequences when multiplied across all our traffic. We would need to simulate that before making any tweaks here.

As far as fall back goes, if your use case requires this very large path, it would be a great idea to consider a backup or secondary cluster for the scenarios you mentioned. It is dangerous in general, and not recommended, to rely on the SDF Horizon instance for production use.

marcinx:
@ire-and-curses Okay, thanks for the reply. That's a good idea and I had a setup with a separate Horizon instance just for path payments in mind already. Or maybe we'll come up with a way to iterate over path requests in our code eventually.

PR #4047, opened 2021-11-04
ingest: Remove base64 encoding when storing LedgerKey in memory

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This commit changes all key representations of `LedgerKey` from base64 string to raw bytes string. This should lower memory requirements of `ChangeCompactor`, `CheckpointChangeReader` and `StateVerifier` by around 25% (given that base64 encoding is roughly 4/3 of original length).

### Why

According to [Strings, bytes, runes and characters in Go](https://go.dev/blog/strings):
> Some people think Go strings are always UTF-8, but they are not: only string literals are UTF-8. As we showed in the previous section, string values can contain arbitrary bytes; as we showed in this one, string literals always contain UTF-8 text as long as they have no byte-level escapes.
>
> To summarize, strings can contain arbitrary bytes, but when constructed from string literals, those bytes are (almost always) UTF-8.

Because of that we don't need to base64-encode marshalled `LedgerKey` when using it as map key.

### Known limitations

[TODO or N/A]

### DIFF
--- a/ingest/change_compactor.go
+++ b/ingest/change_compactor.go
@@ -1,6 +1,7 @@
 package ingest
 
 import (
+	"encoding/base64"
 	"sync"
 
 	"github.com/stellar/go/support/errors"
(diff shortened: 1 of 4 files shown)


tamirms (approved):

bartekn:
@leighmcculloch could you confirm I'm not misunderstanding this? I run the new code to ingest state and normal ledger ingestion and it works.

leighmcculloch (approved):
Yup, you're spot on. Strings and []byte are interchangeable in the sense that you can convert from one to the other and back again and nothing changes about the underlying data.

If this is a performance sensitive area of code it may be beneficial to profile the memory allocations to check if the building to a []byte and then conversion to string is causing allocations. It's not clear to be that it is, but if it is, there might be value in us defining MarshalBinary functions that take in a io.Writer do that we can pass in a strings.Builder.

Regardless this looks good to me.

PR #4049, opened 2021-11-05
Revert "Add `is_source` query parameter to transaction endpoints"

2opremio:
Reverts stellar/go#4035


Shaptic (approved):
Looks like a clean revert, but what happened?

PR #4050, opened 2021-11-05
ingest/ledgerbackend: Let filewatcher check for binary hash to detect captive core updates.

erika-sdf:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Let `file_watcher` check for binary hash to detect captive core updates.

### Why

Horizon watches it's captive stellar-core binary timestamp to detect if a new version has been deployed. However, the binaries are timestamped with the package changelog, which is not dynamically updated during build.

### Known limitations

The hash function currently takes < ~30ms to finish. The 10 second duration for `file_watcher`'s loop should be sufficient.


ingest/ledgerbackend/file_watcher.go
```diff
@@ -1,37 +1,56 @@
...
 	"sync"
 	"time"
 
-	"github.com/pkg/errors"
-
+	"github.com/stellar/go/support/errors"
 	"github.com/stellar/go/support/log"
 )
 
+type hash []byte
```
tamirms, line 15: nit: I don't think this type alias provides much benefit since we end up using bytes.Equal to compare hashes anyways

tamirms (approved):

ingest/ledgerbackend/file_watcher.go
```diff
@@ -1,37 +1,56 @@
...
+}
+
+func hashFile(filename string) (hash, error) {
+	f, err := os.Open(filename)
+	if err != nil {
+		return hash{}, errors.Wrapf(err, "unable to open %v", f)
+	}
+	defer f.Close()
+
+	h := sha256.New()
```
Shaptic, line 34: Maybe a nitpick, but do we need sha256 here? There's no need for it to be a cryptographically-secure hash. I think something like crc (or anything else from [pkg/hash](https://pkg.go.dev/hash)) would be sufficient and faster.

Shaptic (approved):
We could've kept the old watcher, but I doubt anyone would miss it. LGTM :+1: Just one comment below :arrow_down:

ingest/ledgerbackend/file_watcher.go
bartekn: We paired with @erika-sdf on this PR and we actually discussed it. We picked SHA-256 because we really wanted to prevent collisions from happening. In case of collision Stellar-Core would not restart and it can be critical in some circumstances.

ingest/ledgerbackend/file_watcher.go
erika-sdf: Fair point. I slightly prefer to keep the type definition though, so I defined a `Equals` method for type `hash` that just calls `byte.Equals`.

ingest/ledgerbackend/file_watcher.go
```diff
@@ -1,37 +1,60 @@
...
 
-	"github.com/pkg/errors"
-
+	"github.com/stellar/go/support/errors"
 	"github.com/stellar/go/support/log"
 )
 
+type hash []byte
+
+func (h *hash) Equals(other hash) bool {
```
tamirms, line 17: what's the rationale for making the function have a pointer receiver?

erika-sdf: I wasn't thinking too much about it. I think defining it on the object itself makes a copy of it, which doesn't seem like a big deal either way. Do you have a reason to prefer one or the other?

tamirms: In https://github.com/golang/go/wiki/CodeReviewComments#receiver-type it states:

If the receiver is a slice and the method doesn't reslice or reallocate the slice, don't use a pointer to it.

erika-sdf: Sounds good to me. Updated.

ingest/ledgerbackend/file_watcher.go
Shaptic: Fair enough, @bartekn! However, in that case, [`crypto/sha1`](https://pkg.go.dev/crypto/sha1) should be sufficient. It's ~23% faster ([source](https://automationrhapsody.com/md5-sha-1-sha-256-sha-512-speed-performance/)):

> SHA-256 is 15.5% slower than SHA-1 for short strings and 23.4% for longer strings.

And _basically_ just as collision resistant. I'm aware of [SHAtter](https://shattered.io/), but that is well beyond the probability of any "accidental" hash collisions from the core binary:

> This attack required over 9,223,372,036,854,775,808 SHA1 computations.

erika-sdf: Swapped out `sha256` with `sha1`

PR #4053, opened 2021-11-08
strkey: add muxed accounts support

### LINKED ISSUE #4029: Feature Request: create strkey helper to validate if a muxed account address is valid
To make it easy to check if a muxed account address such as `MAQAA5L65LSYH7CQ3VTJ7F3HHLGCL3DSLAR2Y47263D56MNNGHSQSAAAAAAAAAAE2LP26` is valid.

Nowadays the only way to validate a muxed address is to check if the method `MuxedAccount{}.SetAddress(M...)` returns an error:
https://github.com/stellar/go/blob/5742327250c1468a68700a41c48ea16cb74dd8e4/xdr/muxed_account.go#L67

The `strkey` package could have a helper method similar to `strkey.IsValidEd25519PublicKey` that validates if the address is a valid muxed account, maybe something called `strkey.IsValidMuxedAccountMed25519PublicKey`

There could also be an additional method that checks if either one or the other is true. It would be useful when validating a destination address.

We could use another package to test for that, instead of `strkey` or even keep relying on `MuxedAccount{}.SetAddress(M...)`

marcelosalloum:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add mixed account helpers and packing/unpacking utilities to the `strkey` package.

### Why

In order to use muxed accounts, developers had to resort to the `xdr` package, while regular accounts could be handled through the `strkey` package. This change makes it so the strkey can be the only package developers will need, also adding a few utilities to make life easier 🎉 .

Closes #4029

### Known limitations

N/A


strkey/muxed_account.go
```diff
@@ -0,0 +1,99 @@
...
+	return Encode(VersionByteMuxedAccount, raw)
+}
+
+// ParseMuxedAccount receives a muxed account M-address and parses it into a
+// MuxedAccount object containing an ed25519 address and an id.
+func ParseMuxedAccount(mAddress string) (*MuxedAccount, error) {
+	if !IsValidMuxedAccountMed25519PublicKey(mAddress) {
+		return nil, errors.New("invalid muxed account")
+	}
+
```
leighmcculloch, line 82: ❗ Similar, you don't need to call this if you're calling Decode immediately following.

strkey/muxed_account.go
```diff
@@ -0,0 +1,99 @@
...
+	raw := make([]byte, 0, 40)
+	raw = append(raw, m.ed25519[:]...)
+	raw = append(raw, b.Bytes()...)
+
+	return Encode(VersionByteMuxedAccount, raw)
+}
+
+// ParseMuxedAccount receives a muxed account M-address and parses it into a
+// MuxedAccount object containing an ed25519 address and an id.
+func ParseMuxedAccount(mAddress string) (*MuxedAccount, error) {
```
leighmcculloch, line 78: ❗ Looking at the existing strkey API, I think we should give this another name.

Screen Shot 2021-11-08 at 11 13 15 AM

It looks like this function is the same as `Decode`, except the versionbyte is known, and therefore we want to return the muxed account type. So, let's use the same name.

```suggestion
func DecodeMuxedAccount(address string) (*MuxedAccount, error) {
```

strkey/muxed_account_test.go
```diff
@@ -0,0 +1,88 @@
+package strkey
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestMuxedAccount_id(t *testing.T) {
```
leighmcculloch, line 9: 💡 Looks like this primarily tests the `ID` function. I'd use the convention of naming the function:
```suggestion
func TestMuxedAccount_ID(t *testing.T) {
```

strkey/muxed_account_test.go
```diff
@@ -0,0 +1,88 @@
...
+	muxed := MuxedAccount{}
+	assert.Equal(t, uint64(0), muxed.ID())
+
+	muxed = MuxedAccount{id: uint64(9223372036854775808)}
+	assert.Equal(t, uint64(9223372036854775808), muxed.ID())
+}
+
+func TestMuxedAccount_setID(t *testing.T) {
```
leighmcculloch, line 17: 💡 
```suggestion
func TestMuxedAccount_SetID(t *testing.T) {
```

strkey/muxed_account_test.go
```diff
@@ -0,0 +1,88 @@
...
+	muxed := MuxedAccount{}
+	muxed.SetID(123)
+	assert.Equal(t, uint64(123), muxed.ID())
+
+	muxed.SetID(456)
+	assert.Equal(t, uint64(456), muxed.ID())
+}
+
+func TestMuxedAccount_address(t *testing.T) {
```
leighmcculloch, line 26: 💡 
```suggestion
func TestMuxedAccount_Address(t *testing.T) {
```

strkey/muxed_account_test.go
leighmcculloch: There is no official format for naming tests, but there is an official format for example tests, and the same pattern applies well to regular tests too:

```go
func Example_suffix() { ... }
func ExampleF_suffix() { ... }
func ExampleT_suffix() { ... }
func ExampleT_M_suffix() { ... }
```
Ref: https://pkg.go.dev/testing#hdr-Examples

leighmcculloch (commented):
Couple asks (❗) and suggestions (💡). Thanks for taking on this work. Thanks for adding thorough tests!

strkey/main.go
```diff
@@ -240,6 +240,19 @@ func IsValidEd25519PublicKey(i interface{}) bool {
 	return err == nil
 }
 
+// IsValidMuxedAccountMed25519PublicKey validates a Stellar SEP-23 muxed address.
+func IsValidMuxedAccountMed25519PublicKey(i interface{}) bool {
+	enc, ok := i.(string)
```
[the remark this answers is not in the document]
marcelosalloum: I made it an interface to be consistent with IsValidEd25519PublicKey and IsValidEd25519SecretSeed. I'm happy to change it to string if there's no reason to use interface.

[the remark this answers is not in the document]
leighmcculloch: Make it a string for now. It is a relatively safe change to make it an interface{} later if it turns out we were wrong. It's a more controversial change to make it a string later.

strkey/main.go
```diff
@@ -240,6 +240,19 @@ func IsValidEd25519PublicKey(i interface{}) bool {
 	return err == nil
 }
 
+// IsValidMuxedAccountMed25519PublicKey validates a Stellar SEP-23 muxed address.
+func IsValidMuxedAccountMed25519PublicKey(i interface{}) bool {
```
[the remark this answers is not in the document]
Shaptic: Nope, it's called that in the XDR:

```xdr
// Source or destination of a payment operation
union MuxedAccount switch (CryptoKeyType type)
{
case KEY_TYPE_ED25519:
    uint256 ed25519;
case KEY_TYPE_MUXED_ED25519:
    struct
    {
        uint64 id;
        uint256 ed25519;
    } med25519;
};
```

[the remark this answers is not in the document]
leighmcculloch: Ah, the address is a muxed ed25519, the `med` is just the abbreviation of muxed ed25519 for the variable name. In our function names I don't think we need to repeat the abbreviation for "muxed" immediately after saying "muxed account".

strkey/main_test.go
```diff
@@ -63,6 +63,32 @@ func TestIsValidEd25519PublicKey(t *testing.T) {
...
+
+	invalidKey = "MAQAA5L65LSYH7CQ3VTJ7F3HHLGCL3DSLAR2Y47263D56MNNGHSQSAAAAAAAAAAE2LP26"
+	isValid = IsValidEd25519PublicKey(invalidKey)
+	assert.Equal(t, false, isValid)
+}
+
+func TestIsValidMuxedAccountMed25519PublicKey(t *testing.T) {
+	validKey := "MAQAA5L65LSYH7CQ3VTJ7F3HHLGCL3DSLAR2Y47263D56MNNGHSQSAAAAAAAAAAE2LP26"
+	isValid := IsValidMuxedAccountMed25519PublicKey(validKey)
+	assert.Equal(t, true, isValid)
```
Shaptic, line 75: nit: it'd be _very slightly_ cleaner to use `assert.IsTrue/IsFalse` here

strkey/main.go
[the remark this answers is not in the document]
marcelosalloum: This naming was based on the xdr package type `MuxedAccountMed25519`:

https://github.com/stellar/go/blob/7a3ca3a531a71f26dc1ac593ce4eb23ece81db3a/xdr/xdr_generated.go#L13224-L13228

But I agree MuxedAccountEd25519 would be clear enough.

strkey/muxed_account.go
```diff
@@ -0,0 +1,99 @@
...
+	"fmt"
+
+	xdr "github.com/stellar/go-xdr/xdr3"
+	"github.com/stellar/go/support/errors"
+)
+
+type MuxedAccount struct {
+	id      uint64
+	ed25519 [32]byte
+}
```
Shaptic, line 14: Should we use the `xdr.MuxedAccount` type here, instead? It'll make it clear that this is just a wrapper around it.

strkey/muxed_account.go
```diff
@@ -0,0 +1,99 @@
...
+	if m.ed25519 == [32]byte{} {
+		return "", errors.New("muxed account has no ed25519 key")
+	}
+
+	return Encode(VersionByteAccountID, m.ed25519[:])
+}
+
+// MuxedAddress returns the muxed account M-address according with the SEP-23
+// definition for multiplexed accounts.
+func (m *MuxedAccount) MuxedAddress() (string, error) {
```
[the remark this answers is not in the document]
Shaptic: Agreed here: Address should return the true address, while something like AccountIDAddress, UnderlyingAddress, GAddress, BaseAddress, or AccountID (my preference), would return the G... version.

strkey/muxed_account.go
marcelosalloum: https://github.com/stellar/go/pull/4053/commits/b566ffa248b29fabe754b499dd4b73e7752012fc

strkey/muxed_account.go
```diff
@@ -0,0 +1,99 @@
...
+func (m *MuxedAccount) SetID(id uint64) {
+	m.id = id
+}
+
+// SetAddress populates the muxed account ed25519 address.
+func (m *MuxedAccount) SetAddress(address string) error {
+	if !IsValidEd25519PublicKey(address) {
+		return errors.New("invalid ed25519 public key")
+	}
+
```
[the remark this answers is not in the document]
marcelosalloum: https://github.com/stellar/go/pull/4053/commits/b566ffa248b29fabe754b499dd4b73e7752012fc

strkey/main.go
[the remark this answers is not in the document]
marcelosalloum: https://github.com/stellar/go/pull/4053/commits/5ce8134bae881d31c95f94772a0261436e20ed88

strkey/muxed_account_test.go
marcelosalloum: https://github.com/stellar/go/pull/4053/commits/8819f8b24892683470c90c74355c5422aafb5b23

strkey/muxed_account_test.go
marcelosalloum: https://github.com/stellar/go/pull/4053/commits/8819f8b24892683470c90c74355c5422aafb5b23

strkey/muxed_account_test.go
marcelosalloum: https://github.com/stellar/go/pull/4053/commits/8819f8b24892683470c90c74355c5422aafb5b23

strkey/muxed_account.go
marcelosalloum: https://github.com/stellar/go/pull/4053/commits/7638e65839fd33cc99f1e9b861c5b5ee2901e38f

strkey/main_test.go
```diff
@@ -63,6 +63,32 @@ func TestIsValidEd25519PublicKey(t *testing.T) {
...
+	isValid = IsValidMuxedAccountMed25519PublicKey(invalidKey)
+	assert.Equal(t, false, isValid)
+
+	invalidKey = ""
+	isValid = IsValidMuxedAccountMed25519PublicKey(invalidKey)
+	assert.Equal(t, false, isValid)
+
+	invalidKey = "SBCVMMCBEDB64TVJZFYJOJAERZC4YVVUOE6SYR2Y76CBTENGUSGWRRVO"
+	isValid = IsValidMuxedAccountMed25519PublicKey(invalidKey)
+	assert.Equal(t, false, isValid)
```
Shaptic, line 91: More invalid test cases in [stellar/js-stellar-base here](https://github.com/stellar/js-stellar-base/blob/master/test/unit/strkey_test.js#L358-L389) taken directly from the SEP, might be worth a for loop like [this](https://github.com/stellar/js-stellar-base/blob/master/test/unit/strkey_test.js#L391).

Shaptic (commented):
Echoing some of @leighmcculloch's comments, but also some minor suggestions below :arrow_down:!

strkey/muxed_account.go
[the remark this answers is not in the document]
leighmcculloch: AccountID works too.

strkey/muxed_account.go
marcelosalloum: We can't because it would cause a circular dependency since `xdr` already imports `strkey`.

strkey/muxed_account.go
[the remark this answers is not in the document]
marcelosalloum: Updated in https://github.com/stellar/go/pull/4053/commits/d39ab52a3b04aa0a44134509437b0f712fd30187.

strkey/main.go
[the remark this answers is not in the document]
marcelosalloum: Renamed to `IsValidMuxedAccountEd25519PublicKey` in https://github.com/stellar/go/pull/4053/commits/e6430a2cc77596a0916764b94b3ce14d703fedb5.

strkey/main_test.go
marcelosalloum: Thanks for the suggestion. Done in https://github.com/stellar/go/pull/4053/commits/7184f623df20af7071eb8d6ef02a3ffebe4d4d6f.

strkey/main_test.go
marcelosalloum: Done as part of https://github.com/stellar/go/pull/4053/commits/7184f623df20af7071eb8d6ef02a3ffebe4d4d6f.

strkey/muxed_account.go
marcelosalloum: `strkey` is already being imported by `xdr` in these files:
* https://github.com/stellar/go/blob/7a3ca3a531a71f26dc1ac593ce4eb23ece81db3a/xdr/account_id.go#L7
* https://github.com/stellar/go/blob/7a3ca3a531a71f26dc1ac593ce4eb23ece81db3a/xdr/asset.go#L9
* https://github.com/stellar/go/blob/7a3ca3a531a71f26dc1ac593ce4eb23ece81db3a/xdr/muxed_account.go#L7
* https://github.com/stellar/go/blob/7a3ca3a531a71f26dc1ac593ce4eb23ece81db3a/xdr/signer_key.go#L6

strkey/muxed_account.go
```diff
@@ -0,0 +1,91 @@
...
+func (m *MuxedAccount) ID() uint64 {
+	return m.id
+}
+
+// AccountID returns the muxed account G-address according with the SEP-23
+// definition for multiplexed accounts.
+func (m *MuxedAccount) AccountID() (string, error) {
+	if m.ed25519 == [32]byte{} {
+		return "", errors.New("muxed account has no ed25519 key")
+	}
```
leighmcculloch, line 47: 💡 I don't think you need to do this error check. In fact this probably makes it difficult to use the type to test some invalid cases in software we write.

leighmcculloch (approved):
👏🏻

strkey/muxed_account.go
marcelosalloum: Yeah, I wasn't sure about this one either. It sounds fine to remove this validation.

https://github.com/stellar/go/pull/4053/commits/3f3f08ad5e25e2cbb317c6401ab9a31d246f1b11

PR #4057, opened 2021-11-08
doc: use pull request title pattern for issues too

leighmcculloch:
### What

Prefix issues by their primarily affected service or area of the monorepo.

### Why

It's difficult to sometimes understand what the scope of an issue is because our issue titles assume the reader has some prior knowledge that the scope relates to the Go SDK, or Horizon, or something else. Sometimes we don't provide this context at all in the title, other times the context can be found in labels, and other times we do mention it in the title but inside the text.

We have a really clear pattern for communicating this in pull requests: we prefix titles with their area most affected. This results in pull requests having titles like `services/ticker: ...` or `ingest: ...`. This pattern can be applied to issues as well.

We have several teams working in this repository and so context is not always obvious to people in other teams.

The Go community also has a well established pattern of using this for issues at pull requests in the golang/go repo, and it shows up in the ecosystem too, and it is a helpful pattern in that much larger complex repo, so should scale well here too.

### Known limitations

N/A

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,7 +14,8 @@ Check out the [Stellar Contribution Guide](https://github.com/stellar/.github/bl
 ### Issues
 
 * Issues and PR titles start with:
-  * The package name most affected, ex. `services/horizon: fix...`.
+  * The package name most affected, ex. `ingest: fix...`.
+  * Or, for services and tools, the service most affected, ex. `services/horizon: fix...`, `services/ticker: add...`
   * Or, multiple package names separated by a comma when the fix addresses multiple packages worth noting, ex. `services/horizon, services/friendbot: fix...`.
   * Or, `all:` when changes are broad, ex. `all: update...`.
   * Or, `doc:` when changes are isolated to non-code documentation not limited to a single package.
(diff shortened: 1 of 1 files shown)


marcelosalloum (approved):
SGTM!

CONTRIBUTING.md
```diff
@@ -11,13 +11,19 @@ Check out the [Stellar Contribution Guide](https://github.com/stellar/.github/bl
...
+  * The package name most affected, ex. `ingest: fix...`.
+  * Or, for services and tools, the service most affected, ex. `services/horizon: fix...`, `services/ticker: add...`
   * Or, multiple package names separated by a comma when the fix addresses multiple packages worth noting, ex. `services/horizon, services/friendbot: fix...`.
   * Or, `all:` when changes are broad, ex. `all: update...`.
   * Or, `doc:` when changes are isolated to non-code documentation not limited to a single package.
+* Label issues with `bug` if they're clearly a bug.
+* Label issues with `feature request` if they're a feature request.
+
+### Pull Requests
+
```
acharb, line 26: 💭 If someone is creating a PR only they might not look through the `Issues` section first and see the titles rules there
```suggestion
* PR titles follow the same rules as described in the [Issues](#Issues) section above.
```

acharb (approved):
lgtm, had 1 💭

accordeiro (approved):
I love this idea!

ire-and-curses (approved):
Seems fine. No comments from Horizon team so let's do it. :+1:

PR #4058, opened 2021-11-09
services/friendbot: allow configuring friendbot with env vars and set some sensible defaults

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Allow configuring friendbot with environment variables in addition to via a .cfg file. Remove the default `./friendbot.cfg` value from the `--config` option, so that to use a file the user must specify it explicitly.

### Why

It is inconvenient to deploy friendbot in some environments because it is not configurable via environment variables. While in K8s environments it is extremely convenient to mount a cfg file, that is not the case in many other PaaS environments where environment variables are the most common approach to configuration and secret injection.

Environment variables have become common place because of their simplicity, and things like the [12factor](https://12factor.net) manifesto.

We should make it easy to deploy friendbot. It's a remarkably simple service and its deployment should be so too.

This arose out of my own attempt to deploy friendbot.

### Known limitations

This is in someway a risky change given that someone could have deployed friendbot with a .cfg file mounted in the default location. If they upgrade to a version of friendbot after this they'll see an error with a list of configurations they haven't set. Given that there are few, that we know of, deployments of friendbot, this is a very low risk. In any modern blue-green or canary deployment this failure to boot should be identified quickly without a production issue, so I think the risk is low. Also, given friendbot is only used in test environments, the risk of impact is significantly low.

I checked SDF's deployment to check if it explicitly specifies the location of the config file, and it does, so this change should have no impact on the SDF testnet deployment.


bartekn (approved):
LGTM! You can consider using [`ConfigOption`](https://pkg.go.dev/github.com/stellar/go/support/config#ConfigOption). There's no example in godoc but you can check Horizon for usage.

leighmcculloch:
> You can consider using [`ConfigOption`](https://pkg.go.dev/github.com/stellar/go/support/config#ConfigOption).

Ah of course. I can't believe I forgot this existed. I've used it in the recoverysigner and webauth applications too. I'll redo this using that. Thanks for pointing this out.

tsachiherman:
@leighmcculloch do you plan to keep working on this one ?

PR #4059, opened 2021-11-09
services/friendbot: remove disabled error messages that is dead code

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Remove the error messages for when friendbot is disabled.

### Why

Friendbot is never disabled. The field that this code checks is always set with a value, and so the error is never triggered. I think this code is old code that is probably from the days when friendbot was a part of Horizon and could be disabled within Horizon. Now the code is always enabled because for networks that don't have friendbot functionality we simply don't deploy friendbot.

### Known limitations

N/A


marcelosalloum (approved):
LGTM!

Issue #4010, opened 2021-10-18
Feature Request: add horizon db reingest fill-gaps command


tamirms:
### What problem does your feature solve?

When running the `horizon db reingest range` command with parallel workers the entire range to be ingested is divided among several threads. If the command crashes (via the user aborting or through an operational error) this will result in the range being partially ingested. However, because the range was being ingested concurrently, there is no guarantee that horizon's ingestion history will be left in a contiguous state.

For example, let's say we are reingesting from ledger 1 to ledger 1,000 with 10 workers.  After some time, the command is aborted. It is possible that all the workers finished their ingestion task except the worker which was ingesting the sub range from ledger 500 to 600.

Thus, there is a gap in history from ledger 500 to 600.

### What would you like to see?

We should implement a `horizon db reingest fill-gaps` command.

`horizon db reingest fill-gaps` can be called with no parameters in which case it queries for any existing gaps in the horizon db and then proceeds to ingest history to fill the gaps.

The command can also be called with a start and end ledger parameter. In which case, the command will only query for gaps within the provided range.

`horizon db reingest fill-gaps 1 1000` will fill any gaps occurring within the range 1-1000

We should also be able to configure the same parallel worker command line parameters that are used by `horizon db reingest range` so that `fill-gaps` can be also take advantage of concurrency.

### What alternatives are there?

There is a `horizon db detect-gaps` command which will examine the horizon db to find any gaps in history. So, horizon operators could manually call `horizon db detect-gaps` and then fill the gaps manually with several `horizon db reingest` commands. This workflow is doable for operators who maintain a single horizon instance. But it becomes cumbersome and error prone when you are managing many horizon instances.

PR #4060, opened 2021-11-09
services/horizon: Implement horizon db fill-gaps command

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Close https://github.com/stellar/go/issues/4010

`horizon db fill-gaps` can be called with no parameters in which case it queries for any existing gaps in the horizon db and then proceeds to ingest history to fill the gaps.

The command can also be called with a start and end ledger parameter. In which case, the command will only query for gaps within the provided range.

`horizon db fill-gaps 1 1000` will fill any gaps occurring within the range 1-1000

It is also possible to configure the same parallel worker command line parameters that are used by horizon db reingest range so that fill-gaps can be also take advantage of concurrency.

### Known limitations

[N/A]

### DIFF
--- a/services/horizon/internal/db2/history/ledger.go
+++ b/services/horizon/internal/db2/history/ledger.go
@@ -130,11 +130,11 @@ func (q *Q) InsertLedger(ctx context.Context,
 	return result.RowsAffected()
 }
 
-// GetLedgerGaps, obtains ingestion gaps in the history_ledgers table.
+// GetLedgerGaps obtains ingestion gaps in the history_ledgers table.
 // Returns the gaps and error.
 func (q *Q) GetLedgerGaps(ctx context.Context) ([]LedgerGap, error) {
-	var result []LedgerGap
-	err := q.SelectRaw(ctx, &result, `
+	var gaps []LedgerGap
+	query := `
     SELECT sequence + 1 AS gap_start,
 		next_number - 1 AS gap_end
 	FROM (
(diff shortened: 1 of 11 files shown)


services/horizon/internal/db2/history/ledger.go
```diff
@@ -130,20 +130,91 @@ func (q *Q) InsertLedger(ctx context.Context,
...
+			EndSequence:   min(end, oldestLedger-1),
+		})
+	}
+
+	gaps, err := q.GetLedgerGaps(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, gap := range gaps {
```
bartekn, line 197: I think this code depends on `gaps` being sorted. Are they always sorted?

services/horizon/internal/ingest/main.go
```diff
@@ -486,23 +486,52 @@ func (s *system) VerifyRange(fromLedger, toLedger uint32, verifyState bool) erro
...
+		}
+		if from == 0 {
+			return errors.Errorf("Invalid range: %v genesis ledger starts at 1", pair)
+		}
+		if i == 0 {
+			continue
+		}
+		prev := ledgerRanges[i-1]
+		if prev[1] >= from {
+			return errors.Errorf("ranges are not sorted prevRange %v  curRange %v", prev, pair)
```
bartekn, line 503:
```suggestion
			return errors.Errorf("ranges are not sorted prevRange %v curRange %v", prev, pair)
```

services/horizon/internal/ingest/parallel.go
```diff
@@ -85,9 +104,17 @@ func calculateParallelLedgerBatchSize(rangeSize uint32, batchSizeSuggestion uint
 	return (batchSize / historyCheckpointLedgerInterval) * historyCheckpointLedgerInterval
 }
 
-func (ps *ParallelSystems) ReingestRange(fromLedger, toLedger uint32, batchSizeSuggestion uint32) error {
+func totalRangeSize(ledgerRanges [][2]uint32) uint32 {
+	var sum uint32
+	for _, pair := range ledgerRanges {
+		sum += pair[1] - pair[0]
```
bartekn, line 110: Shouldn't we add +1 here? Like [1, 3] size is 3 (1, 2, 3) not 2.
```suggestion
		sum += pair[1] - pair[0] + 1
```

services/horizon/internal/db2/history/ledger.go
tamirms: I think the sql query should return them in sorted order but I can add an explicit sort

tamirms:

bartekn (approved):

PR #4063, opened 2021-11-10
xdr: update using stellar/xdrgen#68

2opremio:
It improves XDR marshaling performance.

See stellar/xdrgen#68 for details

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,6 +1,6 @@
 GIT
   remote: https://github.com/stellar/xdrgen.git
-  revision: 92a28e1f3ef3a2b7e9409a2baa2c5ed0751d3e1c
+  revision: 4677de9fcda28ce314415b3246211ec4657a0eab
   ref: master
   specs:
     xdrgen (0.1.1)
@@ -41,7 +41,7 @@ GEM
     faraday-net_http_persistent (1.2.0)
     faraday-patron (1.0.0)
     faraday-rack (1.0.0)
-    i18n (1.8.10)
+    i18n (1.8.11)
       concurrent-ruby (~> 1.0)
     memoist (0.11.0)
     method_source (1.0.0)
@@ -79,4 +79,4 @@ DEPENDENCIES
   xdrgen!
 
 BUNDLED WITH
-   2.2.22
+   2.2.29
--- a/Rakefile
+++ b/Rakefile
@@ -40,7 +40,6 @@ namespace :xdr do
     require "pathname"
     require "xdrgen"
     require 'fileutils'
-    FileUtils.rm_f("xdr/xdr_generated.go")
 
     compilation = Xdrgen::Compilation.new(
       LOCAL_XDR_PATHS,
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,8 @@ xdr/Stellar-overlay.x \
 xdr/Stellar-transaction.x \
 xdr/Stellar-types.x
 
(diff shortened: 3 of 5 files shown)


Rakefile
```diff
@@ -40,7 +40,6 @@ namespace :xdr do
     require "pathname"
     require "xdrgen"
     require 'fileutils'
-    FileUtils.rm_f("xdr/xdr_generated.go")
```
leighmcculloch, line 43: ❗ This looks unrelated and like an unnecessary change. Why is the change needed? I believe this is there to make sure that any xdr_generated.go we have at the end of a generate run is new and no silent error that doesn't result in a new file being written goes undetected.

Gemfile.lock
```diff
@@ -41,7 +41,7 @@ GEM
     faraday-net_http_persistent (1.2.0)
     faraday-patron (1.0.0)
     faraday-rack (1.0.0)
-    i18n (1.8.10)
+    i18n (1.8.11)
```
leighmcculloch, line 44: ❗ It looks like you probably ran `bundle update xdrgen` and it has updated another dependency that is unnecessary. Ruby's bundler will update any dependency it can. Can you change this back to 1.8.10? I think we can avoid updating other dependencies unless we see a need.

leighmcculloch (approved):
Two requests (❗), otherwise looks great. 🎉

bartekn (approved):
LGTM.

bartekn:
I was curious and run CPU profiler on Horizon 2.10.0 and master. Highlighted all `xdr.*` calls. `TransactionProcessor` is now faster then `OperationProcessor` on my machine! Great job @leighmcculloch and @2opremio!

`horizon-v2.10.0`
Screenshot 2021-11-10 at 23 46 11

`master`
Screenshot 2021-11-10 at 23 46 21

Rakefile
2opremio: It made development unpleasant because my IDE closed the buffer when the file was removed.

There is lo compelling reason to remove the file, since it's overwritten.

Gemfile.lock
2opremio: I run a conservative update, only on xdrgen.

leighmcculloch: Even on conservative updates bundler will update all transitive dependencies for the gem you requested be updated.

Rakefile
leighmcculloch: Ah interesting. We can solve both problems by replacing the `rm` with emptying the file. So the file never disappears, but it still behaves as before and is zeroed out?

2opremio: I still don't see a compelling reason to reset the file.

Gemfile.lock
2opremio: I don't see it as a very important side effect.

2opremio:
@bartekn is there a benchmark for that? If so, we should run it when optimizing Horizon. If not, I think we should create one.

2opremio:
I think we could have a benchmark for each transaction processor and one for ingestion in general.

bartekn:
I think benchmark makes sense but only with mocked DB to filter it out. I think we can do (in a separate PR) but we are constantly monitoring Horizon metrics so not sure if it's required.

2opremio:
It would be really helpful when working on profiling (not just for monitor how Horizon's performance is going). I think it would be useful even when running against a local DB.

PR #4064, opened 2021-11-11
xdr: update using xdrgen#71

2opremio:
This PR regenerates the xdr types using https://github.com/stellar/xdrgen/pull/71

There is a dramatic performance improvement. ~13x speedup and and 11x reduction of allocations.

```
goos: darwin
goarch: amd64
pkg: github.com/stellar/go/benchmarks
BenchmarkXDRUnmarshalWithReflection
BenchmarkXDRUnmarshalWithReflection-8                 	   70221	     17617 ns/op	    5528 B/op	     185 allocs/op
BenchmarkXDRUnmarshal
BenchmarkXDRUnmarshal-8                               	  762405	      1360 ns/op	    1344 B/op	      17 allocs/op
BenchmarkGXDRUnmarshal
BenchmarkGXDRUnmarshal-8                              	   39950	     28912 ns/op	   86736 B/op	     278 allocs/op
```

This performance can be later improved even further by using the same approach as in https://github.com/stellar/go/pull/4056

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,6 @@
 source 'https://rubygems.org'
 
-gem 'xdrgen', git:'https://github.com/stellar/xdrgen.git', ref: 'master'
+gem 'xdrgen', git:'https://github.com/stellar/xdrgen.git', ref: 'generate-decodefrom'
 gem 'pry'
 gem 'octokit'
 gem 'netrc'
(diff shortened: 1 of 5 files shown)


sreuland:
nice work, new hire here, just trying to follow, amazed to see the custom xdr compiler. are there 'before' benchmarks to compare with the table captured here on latest results?

bartekn (approved):

PR #4069, opened 2021-11-13
services/horizon: Use encoding buffer in state verifier

2opremio:
Profiling shows there are a lot of allocations due to buffer resizing.

### DIFF
--- a/services/horizon/internal/ingest/verify.go
+++ b/services/horizon/internal/ingest/verify.go
@@ -152,9 +152,7 @@ func (s *system) verifyState(verifyAgainstLatestCheckpoint bool) error {
 	}
 	defer stateReader.Close()
 
-	verifier := &verify.StateVerifier{
-		StateReader: stateReader,
-	}
+	verifier := verify.NewStateVerifier(stateReader, nil)
 
 	assetStats := processors.AssetStatSet{}
 	total := int64(0)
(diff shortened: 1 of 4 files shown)


tamirms (approved):
I think the encoding buffer could also be used by the checkpoint change reader which calls MarshalBinaryCompress() in two places:

https://github.com/stellar/go/blob/3e1e280c1108ab08747c6d17604bd48d74d6b757/ingest/checkpoint_change_reader.go#L370

https://github.com/stellar/go/blob/3e1e280c1108ab08747c6d17604bd48d74d6b757/ingest/checkpoint_change_reader.go#L424

There is also the XdrStream struct in the historyarchive package. We use that to extract xdr ledger entries out of the history archive snapshots. There are lots of decoding operations happening in XdrStream:

https://github.com/stellar/go/blob/a613f85a5654a7d9a6efd01678dd19af72acd5a1/historyarchive/xdrstream.go#L178

Would it be possible to use a decoding buffer there? I think that would speed up state verification and the  buildState step in the ingestion state machine.

2opremio:
Yep, I was already working on migrating the compress functions.

I will look into the XdrStream next. Thanks.

2opremio:
> I think the encoding buffer could also be used by the checkpoint change reader which calls MarshalBinaryCompress() in two places:

Done at https://github.com/stellar/go/pull/4071

> There is also the XdrStream struct in the historyarchive package. We use that to extract xdr ledger entries out of the history archive snapshots. There are lots of decoding operations happening in XdrStream:

I don't think it's worth optimizing this until https://github.com/stellar/xdrgen/pull/71 is merged

2opremio:
> There is also the XdrStream struct in the historyarchive package. We use that to extract xdr ledger entries out of the history archive snapshots. There are lots of decoding operations happening in XdrStream:

Addressed at https://github.com/stellar/go/pull/4075

PR #4071, opened 2021-11-13
xdr and horizon: Optimize compress-marshaling of ledger keys

2opremio:
15x CPU time improvement, all allocations were removed

Before:

```
goos: darwin
goarch: amd64
pkg: github.com/stellar/go/benchmarks
BenchmarkXDRMarshalCompress
BenchmarkXDRMarshalCompress-8   	  608031	      1970 ns/op	    2048 B/op	      59 allocs/op
PASS
```

After:

```
goos: darwin
goarch: amd64
pkg: github.com/stellar/go/benchmarks
BenchmarkXDRMarshalCompress
BenchmarkXDRMarshalCompress-8   	 8785383	       133.8 ns/op	       0 B/op	       0 allocs/op
PASS
```

~One of the remaning allocations is, in fact, caused by a call to PoolID.Encode which I think can be removed.~

### DIFF
--- a/xdr/account_id.go
+++ b/xdr/account_id.go
@@ -72,20 +72,25 @@ func (aid *AccountId) LedgerKey() (ret LedgerKey) {
 //
 // Warning, do not use UnmarshalBinary() on data encoded using this method!
 func (aid AccountId) MarshalBinaryCompress() ([]byte, error) {
-	m := []byte{byte(aid.Type)}
+	e := NewEncodingBuffer()
+	if err := e.accountIdCompressEncodeTo(aid); err != nil {
+		return nil, err
+	}
+	return e.xdrEncoderBuf.Bytes(), nil
+}
 
+func (e *EncodingBuffer) accountIdCompressEncodeTo(aid AccountId) error {
+	if err := e.xdrEncoderBuf.WriteByte(byte(aid.Type)); err != nil {
+		return err
+	}
 	switch aid.Type {
 	case PublicKeyTypePublicKeyTypeEd25519:
-		pk, err := aid.Ed25519.MarshalBinary()
-		if err != nil {
-			return nil, err
-		}
-		m = append(m, pk...)
+		// TODO: Encoding directly adds a padding (due to invoking EncodeFixedOpaque())
+		//       shouldn't we just write the bytes?
+		return aid.Ed25519.EncodeTo(e.encoder)
 	default:
 		panic("Unknown type")
 	}
-
-	return m, nil
 }
 
 func MustAddress(address string) AccountId {
--- a/xdr/account_id.go
+++ b/xdr/account_id.go
@@ -64,30 +64,14 @@ func (aid *AccountId) LedgerKey() (ret LedgerKey) {
 	return
 }
 
-// MarshalBinaryCompress marshals AccountId to []byte but unlike
-// MarshalBinary() it removes all unnecessary bytes, exploting the fact
-// that XDR is padding data to 4 bytes in union discriminants etc.
-// It's primary use is in ingest/io.StateReader that keep LedgerKeys in
-// memory so this function decrease memory requirements.
-//
-// Warning, do not use UnmarshalBinary() on data encoded using this method!
-func (aid AccountId) MarshalBinaryCompress() ([]byte, error) {
-	e := NewEncodingBuffer()
-	if err := e.accountIdCompressEncodeTo(aid); err != nil {
-		return nil, err
-	}
-	return e.xdrEncoderBuf.Bytes(), nil
-}
-
 func (e *EncodingBuffer) accountIdCompressEncodeTo(aid AccountId) error {
 	if err := e.xdrEncoderBuf.WriteByte(byte(aid.Type)); err != nil {
 		return err
 	}
 	switch aid.Type {
 	case PublicKeyTypePublicKeyTypeEd25519:
-		// TODO: Encoding directly adds a padding (due to invoking EncodeFixedOpaque())
-		//       shouldn't we just write the bytes?
-		return aid.Ed25519.EncodeTo(e.encoder)
+		_, err := e.xdrEncoderBuf.Write(aid.Ed25519[:])
+		return err
 	default:
 		panic("Unknown type")
 	}
--- a/xdr/asset.go
+++ b/xdr/asset.go
@@ -268,32 +268,17 @@ func (a Asset) StringCanonical() string {
 	return fmt.Sprintf("%s:%s", c, i)
 }
 
-// MarshalBinaryCompress marshals Asset to []byte but unlike
-// MarshalBinary() it removes all unnecessary bytes, exploting the fact
-// that XDR is padding data to 4 bytes in union discriminants etc.
-// It's primary use is in ingest/io.StateReader that keep LedgerKeys in
-// memory so this function decrease memory requirements.
-//
-// Warning, do not use UnmarshalBinary() on data encoded using this method!
-func (a Asset) MarshalBinaryCompress() ([]byte, error) {
-	e := NewEncodingBuffer()
-	if err := e.assetCompressEncodeTo(a); err != nil {
-		return nil, err
-	}
-	return e.xdrEncoderBuf.Bytes(), nil
-}
-
 func trimRightZeros(b []byte) []byte {
 	if len(b) == 0 {
 		return b
 	}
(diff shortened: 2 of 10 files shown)


xdr/ledger_key.go
```diff
@@ -132,62 +131,48 @@ func (key *LedgerKey) SetLiquidityPool(poolID PoolId) error {
...
+		return key.ClaimableBalance.BalanceId.EncodeTo(e.encoder)
 	case LedgerEntryTypeLiquidityPool:
-		cBalance, err := key.LiquidityPool.LiquidityPoolId.MarshalBinary()
-		if err != nil {
-			return nil, err
-		}
-		m = append(m, cBalance...)
+		// TODO: why are we encoding the full pool id (with padding)
+		//       here, in TrustLineAsset we just write the bytes directly?
+		return key.LiquidityPool.LiquidityPoolId.EncodeTo(e.encoder)
```
2opremio, line 171: reviewers: do you think there is a reason for this? 

this call actually causes an allocation (I think)

xdr/account_id.go
```diff
@@ -72,20 +72,25 @@ func (aid *AccountId) LedgerKey() (ret LedgerKey) {
...
 	switch aid.Type {
 	case PublicKeyTypePublicKeyTypeEd25519:
-		pk, err := aid.Ed25519.MarshalBinary()
-		if err != nil {
-			return nil, err
-		}
-		m = append(m, pk...)
+		// TODO: Encoding directly adds a padding (due to invoking EncodeFixedOpaque())
+		//       shouldn't we just write the bytes?
+		return aid.Ed25519.EncodeTo(e.encoder)
```
2opremio, line 90: reviewers: is there a reason for this?

xdr/asset.go
```diff
@@ -276,34 +276,48 @@ func (a Asset) StringCanonical() string {
...
-		code = []byte(strings.TrimRight(string(a.AlphaNum12.AssetCode[:]), "\x00"))
-		issuer, err = a.AlphaNum12.Issuer.MarshalBinary()
-		if err != nil {
-			panic(err)
+		code := trimRightZeros(a.AlphaNum12.AssetCode[:])
+		if _, err := e.xdrEncoderBuf.Write(code); err != nil {
+			return err
 		}
+		// TODO: this was panicking before, check with the team
+		return e.accountIdCompressEncodeTo(a.AlphaNum12.Issuer)
```
2opremio, line 315: Reviewers please advise

2opremio:
When writing this PR I started to wonder why don't we simply marshal the keys normally and hash them?

This would:
1. Safe a lot more space
2. Save us from maintaining this compression hack

In other words, do we really need this sort of compression if the result is opaque?

xdr/account_id.go
```diff
@@ -72,20 +72,25 @@ func (aid *AccountId) LedgerKey() (ret LedgerKey) {
 //
 // Warning, do not use UnmarshalBinary() on data encoded using this method!
 func (aid AccountId) MarshalBinaryCompress() ([]byte, error) {
```
2opremio, line 74: Reviewers: I kept this functions around because they are public and removing them would be an external breaking change. However, they are not used in the code base.

Given their specificity, I think I should probably remove them. Thoughts?

xdr/account_id.go
bartekn: I think it should be ok to append bytes directly.

xdr/asset.go
bartekn: Can you elaborate? What was panicking?

xdr/ledger_key.go
```diff
@@ -132,62 +131,48 @@ func (key *LedgerKey) SetLiquidityPool(poolID PoolId) error {
...
-		}
-		m = append(m, seller...)
-		offer, err := key.Offer.OfferId.MarshalBinary()
-		if err != nil {
-			return nil, err
+		if err := e.accountIdCompressEncodeTo(key.Offer.SellerId); err != nil {
+			return err
 		}
-		m = append(m, offer...)
+		return key.Offer.OfferId.EncodeTo(e.encoder)
```
bartekn, line 158: I was thinking about it when working on a different PR (I don't remember which one). Offer IDs are unique to network so technically `LedgerKey` for `Offer` doesn't need `SellerId`. So I think we could store just `OfferId` in compressed keys. I can confirm this with Stellar-Core team because it doesn't seem to be needed in main xdr files.

xdr/ledger_key.go
bartekn: `PoolID` is a sha256 hash so it's length is always 32 bytes (so multiple of 4 = no padding).

xdr/account_id.go
bartekn: I think we can remove them. This one specifically should be private actually.

xdr/asset.go
2opremio: Read above, sorry. There is a panic call

xdr/ledger_key.go
2opremio: Good point. However, we encode directly in asset trustline.

bartekn: Oh, so encoder is adding padding values? Then I think it should be fine to add LP ID bytes directly. But in general compressed keys don't need padding so if it's an issue maybe we could have a switch in encoder to disable padding?

2opremio: I don't think the encoder is adding padding values.

I just meant that in other places of the code (`(a TrustLineAsset) MarshalBinaryCompress()`) we don't invoke the encoder we just encode the bytes directly.

I just think we should do it consistently everywhere and I'll see to it.

xdr/asset.go
2opremio: 👍

xdr/ledger_key.go
2opremio: I fixed the allocation problem at stellar/xdrgen#73 , PTAL

xdr/ledger_key.go
2opremio: Really good catch. I will stop encoding the seller ID

2opremio:
@bartekn PTAL

xdr/trust_line_asset.go
```diff
@@ -42,24 +42,18 @@ func (a TrustLineAsset) MustExtract(typ interface{}, code interface{}, issuer in
...
-//
-// Warning, do not use UnmarshalBinary() on data encoded using this method!
-func (a TrustLineAsset) MarshalBinaryCompress() ([]byte, error) {
+func (e *EncodingBuffer) assetTrustlineCompressEncodeTo(a TrustLineAsset) error {
 	switch a.Type {
 	case AssetTypeAssetTypeNative,
 		AssetTypeAssetTypeCreditAlphanum4,
 		AssetTypeAssetTypeCreditAlphanum12:
-		return a.ToAsset().MarshalBinaryCompress()
+		return e.assetCompressEncodeTo(a.ToAsset())
```
bartekn, line 50: Looks like `assetCompressEncodeTo` is only used here so we could technically remove `a.ToAsset()` allocation by using `TrustLineAsset` object directly here.

bartekn (approved):

xdr/trust_line_asset.go
2opremio: Good idea! I will do that

2opremio:
```
BenchmarkXDRMarshalCompress
BenchmarkXDRMarshalCompress-8                         	 8939768	       135.8 ns/op	       0 B/op	       0 allocs/op
```
;)

xdr/ledger_key.go
```diff
@@ -120,74 +119,39 @@ func (key *LedgerKey) SetLiquidityPool(poolID PoolId) error {
...
-		account, err := key.Data.AccountId.MarshalBinaryCompress()
-		if err != nil {
-			return nil, err
+		if err := e.accountIdCompressEncodeTo(key.Data.AccountId); err != nil {
+			return err
 		}
-		m = append(m, account...)
-		dataName := []byte(strings.TrimRight(string(key.Data.DataName), "\x00"))
-		m = append(m, dataName...)
+		dataName := trimRightZeros([]byte(key.Data.DataName))
```
2opremio, line 143: I am surprised this casting to `[]byte` from string didn't cause an allocation. The escape analysis is probably getting better in Go (it wasn't nearly as good as I would had liked to)

PR #4075, opened 2021-11-16
xdr and historyarchive: Remove XDR decoder allocations from XdrStream

2opremio:
Also:
* Move Uint32Ptr into test file (it was only used for testing)
* Tidy up EncoderTo and DecoderFrom interfaces

### DIFF
--- a/historyarchive/json.go
+++ b/historyarchive/json.go
@@ -17,7 +17,7 @@ import (
 )
 
 func DumpXdrAsJson(args []string) error {
-	var tmp interface{}
+	var tmp xdr.DecoderFrom
 	var rdr io.ReadCloser
 	var err error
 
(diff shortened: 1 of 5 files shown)


2opremio:
@bartekn , @tamirms Please let me know if there are other places in Horizon which would benefit from a similar pattern.

tamirms (approved):

tamirms:
> Please let me know if there are other places in Horizon which would benefit from a similar pattern

I found 3 more places which I think would benefit from using decoders / encoders:

(1) https://github.com/stellar/go/blob/master/ingest/ledgerbackend/buffered_meta_pipe_reader.go#L96
This is where we read ledgers from the captive core pipe

(2) https://github.com/stellar/go/blob/master/ingest/ledgerbackend/remote_captive_core.go#L115
This is where we read ledgers from the http response that we receive from the remote captive core server

(3) https://github.com/stellar/go/blob/master/exp/services/captivecore/internal/api.go#L184
This is where we serialize the ledger into base 64 before in the captive core server

2opremio:
Great, I will take a look

2opremio:
I optimized the captive core pipe decoding at https://github.com/stellar/go/pull/4077

About remote captive core, I saw there was json marshaling/unmarshaling  involved which is done through reflection.

I don't think that an extra XDR decoder allocation is going to make a big difference before optimizing the json encoding. I may be wrong though so @tamirms feel free to take a look at it yourself (now I am curious).

Issue #4022, opened 2021-10-21
Profile Horizon


2opremio:
I don't think we have profiled the Time (CPU and Wallclock) and Memory performance of Horizon in a while.

We should try doing that and and improve any low hanging fruit we see. Particularly Horizon's ingestion pipeline.

For starters, this is a CPU profile I took from Horizon's staging pubnet deployment:
[pprof.stellar-horizon.samples.cpu.001.pb.gz](https://github.com/stellar/go/files/7391159/pprof.stellar-horizon.samples.cpu.001.pb.gz)

2opremio:
I took a quick look at the decoding object allocation and performance ... and I found that most (if not all) the intermediary buffers used by the [xdr decoding library](https://github.com/stellar/go-xdr/blob/master/xdr3/decode.go) end up allocated in the heap.

Here are the top-most object allocating functions of Horizon:

<img width="1078" alt="Screenshot 2021-11-02 at 17 18 22" src="https://user-images.githubusercontent.com/2362916/139904447-fce70dae-9d1c-477e-b596-b20493e0070b.png">

If we ignore the calls from the `reflect` package we can see that calls from `xdr.(*Decoder)` account for ~15% of the objects allocated.

Let's take a a peek at `DecodeInt()`:

```

pprof
stellar-horizon alloc_objects

      File: stellar-horizon
Type: alloc_objects
Time: Oct 22, 2021 at 1:30pm (CEST)
Active filters:
   focus=github\.com/stellar/go-xdr/xdr3\.\(\*Decoder\)\.DecodeInt
Showing nodes accounting for 8585334, 5.06% of 169782541 total
----------------------------------------------------------+-------------
      flat  flat%   sum%        cum   cum%   calls calls% + context 	 	 
----------------------------------------------------------+-------------
                                           7110754 82.82% |   github.com/stellar/go-xdr/xdr3.(*Decoder).decodeUnion github.com/stellar/go-xdr@v0.0.0-20201028102745-f80a23dac78a/xdr3/decode.go:530
                                            753674  8.78% |   github.com/stellar/go-xdr/xdr3.(*Decoder).DecodeBool github.com/stellar/go-xdr@v0.0.0-20201028102745-f80a23dac78a/xdr3/decode.go:175
                                            720906  8.40% |   github.com/stellar/go-xdr/xdr3.(*Decoder).decode github.com/stellar/go-xdr@v0.0.0-20201028102745-f80a23dac78a/xdr3/decode.go:801
[...]

PR #4086, opened 2021-11-17
support/db: Add benchmark for BatchInsertBuilder

2opremio:


support/db/batch_insert_builder_test.go
```diff
@@ -20,6 +21,44 @@ type invalidHungerRow struct {
...
+	defer db.Close()
+	sess := &Session{DB: db.Open()}
+	defer sess.DB.Close()
+	ctx := context.Background()
+	maxBatchSize := 1000
+	insertBuilder := &BatchInsertBuilder{
+		Table:        sess.GetTable("people"),
+		MaxBatchSize: maxBatchSize,
+	}
+	b.StartTimer()
```
tamirms, line 38: would it make sense to use b.ResetTimer() here instead of stoping and then starting?

sreuland (approved):
Hello, nice test coverage,  for relative newcomer into team, it wasn't apparent to understand driver for this, maybe brief mention in description on why it was needed, or issue# for context.

support/db/batch_insert_builder_test.go
2opremio: `auto-merge` merged before I could resolve this, sorry.

I thought @leighmcculloch changed the repo config so that all comments had to be resolved before auto-merged kicked in?

2opremio: @tamirms you are right, it's a better option. I will change it later on 👍

2opremio:
@sreuland You are totally right, I should had provided more context (there are PR guidelines for it which I didn't follow, sorry). The context is trying to optimize the ingestion system. See #4022 

Hopefully it will be more clear in an upcoming PR.

support/db/batch_insert_builder_test.go
leighmcculloch: Even though terraform does not support configuring the wait for conversations to be resolved, it appears it is being cleared on multiple repos, this one and the starlight repo. Maybe the terraform integration is clearing it on every deploy. I think this means we can't use the feature until https://github.com/integrations/terraform-provider-github/issues/868 is resolved.

It might be worth holding off on using auto-merge until this is resolved. The GitHub terraform provider usually ships new features quickly so we might not be waiting that long. 

 cc @stellar/ops-team

PR #4085, opened 2021-11-17
services/horizon: Remove unused indexes from history_transactions

2opremio:
Leftover from https://github.com/stellar/go/pull/4035

PR #4087, opened 2021-11-17
services/horizon: Remove by_hash index, replace with index on hash prefix

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This commit removes `by_hash` index and introduces a new field: `transaction_hash_prefix` (and index on it). This make the duration of `TransactionProcessor` comparable to other processors (around 3-4x speed improvement) but also keeps the queries by transaction hash fast (however fast `SELECT` queries require reingestion).

### Why

After analyzing the speed of queries sent by `TransactionProcessor` we understood that the majority of time is spent in updating indexes on string fields. In https://github.com/stellar/go/pull/4085 unused indexes were removed but there was one left that is actually used in `/transactions/{id}` endpoint: `by_hash` on `transaction_hash` field. Removing it makes the aforementioned endpoint extremely slow and, in reality, useless because of timeouts.

The solution in this commit creates a bigint field that represents a 16 characters (8 bytes) prefix of tx hash in numeric form (thus updating such index is fast). As a result all transactions are put into `math.MaxUint64` buckets. The buckets are used for index scan in `SELECT` statement and found rows are then filtered by `transaction_hash`. In practice the number of buckets is vastly larger than number of existing transactions in public network so we should see more than one transaction in a bucket very rarely or never.

### Known limitations

The problem with this solution is that while ingestion becomes faster the select queries are slow until the prefix column is populated by reingesiton. Also, while creating a new NULLable field is fast in Postgres, creating an index on it is actually very slow for a table with 2B rows (in stg it took around 6h to create such index).


services/horizon/internal/db2/history/transaction_batch_insert_builder.go
```diff
@@ -35,6 +35,19 @@ type transactionBatchInsertBuilder struct {
...
+	if len(hash) < 16 {
+		return 0, errors.New("invalid hash")
+	}
+
+	hexHashPrefix, err := strconv.ParseUint(hash[0:16], 16, 64)
+	if err != nil {
+		return 0, errors.Wrap(err, "error calculating hashHexHash")
+	}
+
+	return int64(hexHashPrefix), nil // overflow expected
```
2opremio, line 48: why not use simply use uint64?

2opremio: Ah, because postgres doesn't support it?

2opremio: It may be worth clarifying it.

bartekn: Correct, postgres doesn't support unsigned ints. I'll update the comment with clarifications.

services/horizon/internal/db2/schema/migrations/52_remove_by_hash_index.sql
```diff
@@ -0,0 +1,9 @@
+-- +migrate Up
+ALTER TABLE history_transactions ADD COLUMN transaction_hash_prefix bigint;
+CREATE INDEX transaction_hash_prefix on history_transactions (transaction_hash_prefix);
+DROP INDEX by_hash;
```
paulbellamy, line 4: Should we include:
```
UPDATE history_transactions SET transaction_hash_prefix = ('x' || lpad(transaction_hash, 16, '0'))::bit(64)::bigint where transaction_hash_prefix is null;
```
here to avoid forcing a reingestion? Might take a while to run?

services/horizon/internal/db2/schema/migrations/52_remove_by_hash_index.sql
```diff
@@ -0,0 +1,9 @@
+-- +migrate Up
+ALTER TABLE history_transactions ADD COLUMN transaction_hash_prefix bigint;
+CREATE INDEX transaction_hash_prefix on history_transactions (transaction_hash_prefix);
```
paulbellamy, line 3: Does `CREATE INDEX CONCURRENTLY` help this build faster? and/or adding a `WHERE transaction_hash_prefix IS NOT NULL`?

services/horizon/internal/db2/history/transaction_batch_insert_builder.go
paulbellamy: Could use numeric for the column, but bleh.

services/horizon/internal/db2/schema/migrations/52_remove_by_hash_index.sql
bartekn: I tried it in https://github.com/stellar/go/pull/4087/commits/fa49b48e62f05a78a7e56937b99e34c75a04dbdc and then removed it: I cancelled it after 10h in stg. I can try again but I think reingestion is better because you can actually track progress.

services/horizon/internal/db2/schema/migrations/52_remove_by_hash_index.sql
bartekn: I'm afraid that `CONCURRENTLY` will affect the performance a lot because of how big the table is. I'll try `WHERE transaction_hash_prefix IS NOT NULL`.

services/horizon/internal/db2/history/transaction.go
```diff
@@ -25,7 +37,22 @@ func (q *Q) TransactionByHash(ctx context.Context, dest interface{}, hash string
...
+	}
+
+	// Second: The query below will be extremely slow for DBs that are not
+	// reingested with transaction_hash_prefix field.
+	byHash = selectTransaction.
+		Where(
+			"(ht.transaction_hash_prefix IS NULL and ht.transaction_hash = ?)",
+			hash,
+		)
+	return q.Get(ctx, dest, byHash)
```
sreuland, line 55: would it be worthwhile to log out a debug/trace line here just for monitoring runtime profile, see how often it falls into here?

services/horizon/internal/db2/history/transaction_test.go
```diff
@@ -27,8 +28,19 @@ func TestTransactionQueries(t *testing.T) {
 	err := q.TransactionByHash(tt.Ctx, &tx, real)
 	tt.Assert.NoError(err)
 
-	fake := "not_real"
-	err = q.TransactionByHash(tt.Ctx, &tx, fake)
+	_, err = q.ExecRaw(tt.Ctx, "UPDATE history_transactions SET transaction_hash_prefix = NULL")
```
sreuland, line 31: nice test coverage, maybe this new verify warrants it's own test case method like 'TestNonPrefixedTransactionQueries' or maybe just a message on the assert adding detail like 'tx should be found with hash prefix' or 'tx should be found without hash prefix'

sreuland (commented):
Nice work! I'm green still, probably not qualified for approval, just observing.

2opremio:
I wonder if this would still be needed if we use https://github.com/stellar/go/pull/4094 to `COPY` into a temporary table without indexes and then merge it into the definite table.

bartekn:
> I wonder if this would still be needed if we use #4094 to COPY into a temporary table without indexes and then merge it into the definite table.

I'm not sure if performance improvement in #4094 actually works well with an index on string field. The benchmark runs on a table without an index.

bartekn:
@stellar/horizon-committers I run some tests in `stg` last week and unfortunately I wasn't able to migrate `history_transactions` table using an SQL query: after around 10h the DB run out of this space (because all rows need to be rewritten but old ones are still kept until vacuum run).

I started think about intermediate solution that will allow people with reingested DB to be able to use improved version but these with old rows still use the old index on hash field. I came up with this:

1. In `TransactionByHash` we first run the query: `select * from history_transactions where transaction_hash_prefix is null limit 1`.
  1. If the query above returns one row this means that there are some rows without prefix set. In such case we use the old query that's using `transaction_hash` field only.
  2. If there are no rows it means all rows have the prefix set so we can use optimized query.
2. At the same time during ingestion (because it always have RW access to DB) we run the same query (`select * from history_transactions where transaction_hash_prefix is null limit 1`).
3. If no rows are found we run `DROP INDEX IF EXISTS by_hash`.

By doing this we remove the performance issue in `TransactionProcessors` as soon as all rows are updated so we don't break performance of `/transactions/{id}` in a mean time. This is a bit tricky and changes the schema outside migrations so wanted to check with you first before implementing.

tamirms:
>At the same time during ingestion (because it always have RW access to DB) we run the same query (select * from history_transactions where transaction_hash_prefix is null limit 1).
If no rows are found we run DROP INDEX IF EXISTS by_hash.
Otherwise do nothing.

@bartekn I'm not sure if it's a good idea to have ingestion essentially perform part of the db migration automatically. perhaps we could have a separate horizon command to drop the index? we could also run the select query when horizon starts up and log a message about how the index should be dropped if no rows are found

bartekn:
Good point. I wanted it to be _automagic_ for users but doing this via CLI command is safer/clearer.

services/horizon/internal/db2/schema/migrations/52_remove_by_hash_index.sql
sreuland: maybe could move the UPDATE to be done on a smaller block of rows obtained from an outer select loop, and then checkpoint after each iteration to keep the commit log space down?

bartekn: I think the speed of this query is now less a problem than the space required. The problem is that `history_transactions` is one of the largest tables in full history table and it needs to be rewritten so it requires twice as much space. When I run this in `stg` we basically run out of memory. So I think I'll implement the idea in: https://github.com/stellar/go/pull/4087#issuecomment-975981439.

sreuland: ok, cool, I was thinking that the pg commit log gets huge when attempting one update stmt on all rows under one TX, so, by CHECKPOINT'ing on a smaller batch of rows, that would allow PG to flush the commit log storage areas often, potentially keeping mem/disk usage lower

bartekn: Oh I see what you mean. Looking at the charts it looks like Postgres is actually able to flush the log during the transaction. "Transaction Logs Disk Usage" peaked at 5GB and during that time "Free Storage Space" from from 1TB to 0. So it looks like the majority of data is actually updated rows.

bartekn:
@stellar/horizon-committers this is ready to review again. I implemented the idea here: https://github.com/stellar/go/pull/4087/#issuecomment-975981439 with @tamirms suggestion: we don't drop index automatically in ingestion. Instead, we print a message on startup that optimization is possible via `horizon db optimize-schema` command.

services/horizon/cmd/db.go
```diff
@@ -485,6 +485,47 @@ func runDBDetectGapsInRange(config horizon.Config, start, end uint32) ([]history
...
+		horizonSession, err := db.Open("postgres", config.DatabaseURL)
+		if err != nil {
+			return err
+		}
+		q := &history.Q{horizonSession}
+
+		// Should we finally set default to info in support/log?
+		hlog.DefaultLogger.SetLevel(hlog.InfoLevel)
+
+		hlog.Info("Checking if by_hash index can be removed...")
```
sreuland, line 507: maybe ln 507 - 523 could be refactored slightly, pull up the details into own function, and cmd here invokes standard interface for each optimization function available
`func(ctx Context, sess Session) error`

or hold-off, consider refactor when/if other optimization come up later.

services/horizon/internal/db2/history/transaction.go
```diff
@@ -11,21 +12,59 @@ import (
 	"github.com/stellar/go/xdr"
 )
 
+// TransactionsWithoutPrefixExist returns true if there is at least one
+// transaction in history_transactions table without transaction_hash_prefix
+// field set.
+func (q *Q) TransactionsWithoutPrefixExist(ctx context.Context) (bool, error) {
+	var id int64
```
sreuland, line 19: just wondering about potential performance option, does the sql query need to run forever? if row count > 0 and no null transaction_hash_prefix, could that first 'false' reading be safely cached here in memory and ensuing requests short circuit to that cache state for response and skip sql round trip, or could that state regress to where new rows still arrive with transaction_hash_prefix=null?

services/horizon/internal/db2/history/transaction.go
```diff
@@ -11,21 +12,59 @@ import (
...
+
+		byInnerHashString, args, err := byInnerHash.ToSql()
+		if err != nil {
+			return errors.Wrap(err, "could not get string for inner hash sql query")
+		}
+		union := byHash.Suffix("UNION ALL "+byInnerHashString, args...)
+
+		return q.Get(ctx, dest, union)
+	} else {
+		byHash := selectTransaction.Where("ht.transaction_hash = ?", hash)
```
tamirms, line 65: in this case don't we still need to union with the inner hash query?

services/horizon/internal/db2/history/transaction.go
```diff
@@ -11,21 +12,59 @@ import (
...
-		Where("ht.inner_transaction_hash = ?", hash)
-
-	byInnerHashString, args, err := byInnerHash.ToSql()
-	if err != nil {
-		return errors.Wrap(err, "could not get string for inner hash sql query")
+	// Check if all rows in `history_transactions` have prefix field set. If so
+	// let's use the index on `transaction_hash_prefix` because index on
+	// `transaction_hash` could be removed by Horizon admins for performance
+	// reasons (faster ingestion).
+	txWithoutPrefixExist, outerErr := q.TransactionsWithoutPrefixExist(ctx)
```
paulbellamy, line 36: Would this code be easier to read if it had fewer double-negatives?
```suggestion
	txPrefixIndexReady, outerErr := q.TransactionPrefixIndexReady(ctx)
```
(ready/available/ok/whatever)

Then below you could do:
```go
if txPrefixIndexReady {
  // All transactions have prefix field set -> use new index
  // ...
} else {
  // no index
}

services/horizon/internal/db2/schema/migrations/52_remove_by_hash_index.sql
paulbellamy: `concurrently` doesn't make it use more db cores, it just lets other txns happen while the index is being built. But probably safer without anyway.

I think the `WHERE transaction_hash_prefix IS NOT NULL` would hurt your `TransactionsWithoutPrefixExist` query, so also probably not worth it.

bartekn: > `concurrently` doesn't make it use more db cores, it just lets other txns happen while the index is being built. But probably safer without anyway.

Postgres [docs](https://www.postgresql.org/docs/12/sql-createindex.html#SQL-CREATEINDEX-CONCURRENTLY) say it actually can affect performance:

> PostgreSQL supports building indexes without locking out writes. This method is invoked by specifying the CONCURRENTLY option of CREATE INDEX. When this option is used, PostgreSQL must perform two scans of the table, and in addition it must wait for all existing transactions that could potentially modify or use the index to terminate. Thus this method requires more total work than a standard index build and takes significantly longer to complete. However, since it allows normal operations to continue while the index is built, this method is useful for adding new indexes in a production environment. **Of course, the extra CPU and I/O load imposed by the index creation might slow other operations.**

Issue #3705, opened 2021-06-17
CaptiveCoreBackend.isClosed can be confused with isPrepared


bartekn:
### What version are you using?

2.4.1

### What did you do?

While working on https://github.com/stellar/go/pull/3694 I noticed that `CaptiveCoreBackend.isClosed` is confused with `CaptiveCoreBackend.isPrepared`. When `CaptiveCoreBackend` is closed by `Close()` call it can't be reused because internal `ctx` is cancelled and `ledgerHashStore` is closed however we often check `isClosed()` and the print an error saying: `session is closed, call PrepareRange first`.

### What did you see instead?

I think what we should do is perform two checks:
* Check `isClosed()` - then prevent from using the `CaptiveCoreBackend` - `PrepareRange` and `GetLedger` should error.
* Check `isPrepared()` - if it's not return an error asking to prepare the range again.

PR #4088, opened 2021-11-18
ingest/ledgerbackend: Differentiate between isPrepared and isClosed in captive core

erika-sdf:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

`CaptiveCoreBackend` should be closed when the last ledger the core is prepared for has been consumed, or if it is explicitly closed. After the core is closed, and `GetLedger` is called on it, it returns the error `session is closed, call PrepareRange first`.

### Why

The caller should not be able to prepare a core after it is closed.

### Known limitations

N/A


ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -458,7 +459,11 @@ func (c *CaptiveStellarCore) GetLedger(ctx context.Context, sequence uint32) (xd
 	}
 
 	if c.isClosed() {
-		return xdr.LedgerCloseMeta{}, errors.New("session is closed, call PrepareRange first")
+		return xdr.LedgerCloseMeta{}, errors.New("session is closed")
+	}
+
+	if c.prepared == nil {
```
paulbellamy, line 465: Actually, now that `isClosed` is more specific, could we use `isPrepared` here?

erika-sdf, line 465: I'm not sure if the right thing to do here is to use `c.prepared` or `c.isPrepared(BoundedRange(sequence, sequence))`.

It seems like the range checks follow this line would yield a more accurate error (eg. `requested ledger ## is behind captive core stream` vs `session not prepared`).

ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -590,7 +595,10 @@ func (c *CaptiveStellarCore) GetLatestLedgerSequence(ctx context.Context) (uint3
 	defer c.stellarCoreLock.RUnlock()
 
 	if c.isClosed() {
-		return 0, errors.New("stellar-core must be opened to return latest available sequence")
+		return 0, errors.New("stellar-core must be open to return latest available sequence")
+	}
+	if c.prepared == nil {
```
erika-sdf, line 600: Similar question here. Originally, this condition was written as `c.isPrepared(UnboundedRange(0))`, but I think that will fail if `GetLatestLedgerSequence` is called with a CaptiveCore prepared for a smaller unbounded range (such as `UnboundedRange(64)`).

ingest/ledgerbackend/captive_core_backend.go
sreuland: that error detail if available, sounds valuable when reading the logs to get that narrow(er) context of the issue.

ingest/ledgerbackend/captive_core_backend_test.go
```diff
@@ -601,11 +602,17 @@ func TestCaptiveGetLedger(t *testing.T) {
 
 	// requires PrepareRange
 	_, err := captiveBackend.GetLedger(ctx, 64)
-	tt.EqualError(err, "session is closed, call PrepareRange first")
+	tt.EqualError(err, "session is not prepared, call PrepareRange first")
 
-	err = captiveBackend.PrepareRange(ctx, BoundedRange(65, 66))
+	ledgerRange := BoundedRange(65, 66)
+	tt.False(captiveBackend.isPrepared(ledgerRange))
```
sreuland, line 608: nice test coverage, is it possible to add a message to the assertion, briefly stating the application reason why this is a failure just for contextual understanding?

sreuland (commented):
Nice work, I'm probably not qualified to approve yet as still pretty new on the uptake of code.

ingest/CHANGELOG.md
```diff
@@ -5,7 +5,8 @@ All notable changes to this project will be documented in this file. This projec
 
 ## Unreleased
 
-* Let filewatcher use binary hash instead of timestap to detect core version update. [4050](https://github.com/stellar/go/pull/4050)
+* Let filewatcher use binary hash instead of timestamp to detect core version update [4050](https://github.com/stellar/go/pull/4050)
+* Differentiate between isPrepared and isClosed in captive core [4088](https://github.com/stellar/go/pull/4088)
```
tamirms, line 9: I don't think it's worth mentioning this in the changelog because this change is basically a code quality improvement. The changelog is more for calling out new features, bug fixes, or changes in horizon's behavior that are relevant to horizon operators

erika-sdf: Removed this PR's entry

ingest/ledgerbackend/captive_core_backend_test.go
erika-sdf: Added a message to this check

tamirms (approved):

ingest/ledgerbackend/captive_core_backend.go
bartekn: You should use `isPrepared`. Using `c.prepared` alone is not enough because it just means the backend is prepared for any range. As an example, let's say backend was prepared for bounded range 100-200 but someone is requesting 201.

ingest/ledgerbackend/captive_core_backend.go
bartekn: I think the usage of `c.prepared` here is correct because we just want to make sure backend is prepared, we don't need a boundary check. But we need to check if `c.stellarCoreRunner` is not nil because we call it later.

ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -590,7 +595,10 @@ func (c *CaptiveStellarCore) GetLatestLedgerSequence(ctx context.Context) (uint3
 	defer c.stellarCoreLock.RUnlock()
 
 	if c.isClosed() {
-		return 0, errors.New("stellar-core must be opened to return latest available sequence")
+		return 0, errors.New("stellar-core must be open to return latest available sequence")
```
bartekn, line 598: As in other comment. Error message needs to be changed here.

ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -600,11 +608,11 @@ func (c *CaptiveStellarCore) GetLatestLedgerSequence(ctx context.Context) (uint3
 }
 
 func (c *CaptiveStellarCore) isClosed() bool {
-	return c.prepared == nil || c.stellarCoreRunner == nil || c.stellarCoreRunner.context().Err() != nil
```
bartekn, line 603: I think we should check `c.closed` here only and because `isClosed` is only used internally we can remove it.

bartekn (commented):
I think the most important thing we need to fix as explained in https://github.com/stellar/go/issues/3705 (and also a comment for `CaptiveStellarCore.cancel`) is that once `Close()` is called is unusable but we don't enforce it leaving users with cryptic errors and/or not working backend.

We need to distinguish 3 states of the backend:
* `NOT PREPARED` - initial state, user need to call `PrepareRange` to do anything.
* `PREPARING/PREPARED` - this is de facto the same state because the backend is not thread-safe and is blocking when calling `PrepareRange` or `GetLedger`.
* `CLOSED` - backend closed.

ingest/ledgerbackend/captive_core_backend.go
paulbellamy: Ah, we can't use `isPrepared()` here, because this is called from `PrepareRange`, so the range may not be prepared yet. A better approach would be to have a separate `awaitRange()` function that `PrepareRange` could use, but that's a much bigger change here.

ingest/ledgerbackend/captive_core_backend.go
paulbellamy: But it might not be intentionally closed. `c.closed` only gets set when we call `Close()`, but it could also be closed due to a timeout or some other issue with the `c.stellarCoreRunner.context()`, no?

bartekn: Exactly and this is confusing. `closed` means that user called `Close()` and `CaptiveStellarCore` can no longer be used. However in case of `stellarCoreRunner` issues, crashes, etc. it just isn't prepared and can be started again.

erika-sdf: Based on this discussion, it sounds like you expect the errored state to be the same as the initial `NOT PREPARED` state. This is counter to the behavior in a couple of the tests that check for the core to be closed after the core gets an error getting a ledger and after the last ledger has been consumed. 

I've pulled out the second half of this check and prompt the user to call `PrepareRange` when there is a core error. I've also updated the tests to reflect that we now expect `stellar-core` to not be closed in these scenarios.

bartekn (approved):
LGTM! Just one comment: we don't need `coreHasError` helper.

ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -590,27 +611,40 @@ func (c *CaptiveStellarCore) GetLatestLedgerSequence(ctx context.Context) (uint3
...
+		return 0, errors.New("stellar-core has an error, call PrepareRange first")
 	}
-
 	if c.lastLedger == nil {
 		return c.nextExpectedSequence() - 1 + uint32(len(c.stellarCoreRunner.getMetaPipe())), nil
 	}
 	return *c.lastLedger, nil
 }
 
+func (c *CaptiveStellarCore) coreHasError() bool {
```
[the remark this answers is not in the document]
bartekn: OK, so I was wrong here. `c.stellarCoreRunner.context().Err() != nil` can also be true when Stellar-Core was closed in another way (crash, binary update). Fix in https://github.com/stellar/go/pull/4192).

PR #4092, opened 2021-11-22
services/horizon: Allow captive core to run with sqlite database

### LINKED ISSUE #4038: Feature Request: Allow captive core to run with sqlite database

If the size of the ledger continues to increase this will result in captive core consuming more memory since it maintains all of its data structures in memory. Stellar Core supports running with a sqlite database. So allowing captive core to maintain some of its state on disk would help reduce memory usage. But we should measure the performance impact of this change.

paulbellamy:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Set default captive core invocations to use the sqlite on-disk database rather than in-memory. 

This will change the existing Horizon run command with params for ingestion enabled and captive core enabled to use the on-disk database in core for ledger ingestion, such as this:  
`horizon --db-url <db_url> --ingest --captive-core-config-path <path_to_core_conf_file> --stellar-core-binary-path /usr/local/bin/stellar-core --history-archive-urls <archive_urls> --network-passphrase "<passphrase>" `

### Why

Help core use less memory, and be more efficient for large ledger handling.

Fixes #4038 

### Known limitations

`--in-memory` is required when we launch core with `--start-at-ledger`, so it can only use sqlite sometimes.

### TODO

- [x] Run on staging for a bit, and check the impact.
- [x] Decide if we want to add a horizon `--in-memory` flag to force the old behaviour.


ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -306,10 +319,10 @@ func (r *stellarCoreRunner) runFrom(from uint32, hash string) error {
 
 	r.cmd = r.createCmd(
 		"run",
-		"--in-memory",
 		"--start-at-ledger", fmt.Sprintf("%d", from),
 		"--start-at-hash", hash,
 		"--metadata-output-stream", r.getPipeName(),
+		"--in-memory",
```
2opremio, line 325: Isn't `createCmd()` responsible to adding `--in-memory`?

paulbellamy:
https://github.com/stellar/stellar-core/pull/3306 should help with disk usage leak when reingesting

ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -311,6 +311,19 @@ func (r *stellarCoreRunner) run() error {
 		return errors.Wrap(err, "error initializing core db")
 	}
 
+	// Do a quick catch-up to set the LCL in core to be our expected starting
+	// point.
+	//
+	// TODO: If we're re-using the directory from a previous run, we need to do
```
sreuland, line 317: doesn't the new-db command done just prior here, wipe the db down to genesis LCL first, seems like we always are starting from there which should be kinda give us a consistent state? 

Graydon did add the PR for getting the current LCL info from core, https://github.com/stellar/stellar-core/pull/3326, not sure if we want to refactor to use that later if this is still issue, seems like using that would add more complexity vs. edge case gain?

paulbellamy: AIUI, the `new-db` is idempotent if the db is already initialized, so in that case it won't do anything at all. But not sure on that.

sreuland: observations of `new-db` effects on db state from command line appear to differ, I've attached log of `catchup;offline-info;new-db;offline-info` and core reports genesis of '1' ledger LCL state is loaded after second new-db is run against the db that had LCL=300.

[new-db.log](https://github.com/stellar/go/files/7980477/new-db.log)

ingest/ledgerbackend/toml.go
```diff
@@ -405,6 +407,11 @@ func (c *CaptiveCoreToml) CatchupToml() (*CaptiveCoreToml, error) {
 }
 
 func (c *CaptiveCoreToml) setDefaults(params CaptiveCoreTomlParams) {
+	// TODO: Do we need to skip this if we are in --in-memory mode?
```
sreuland, line 410: this is moot, the stellar core runner no longer uses --in-memory when it invokes stellar-core, so should be able to assume this is always disk. based database, or maybe I have missed a nuance of this.

ingest/ledgerbackend/toml.go
```diff
@@ -440,6 +447,8 @@ func (c *CaptiveCoreToml) setDefaults(params CaptiveCoreTomlParams) {
 }
 
 func (c *CaptiveCoreToml) validate(params CaptiveCoreTomlParams) error {
+	// TODO: Any way to validate database here?
```
sreuland, line 450: maybe to some degree, it's valid to not have db file present, as `new-db` can create one, we could attempt to validate the STOREAGE_PATH/DATABASE equates to a valid o/s path at least? invoking something like `stellar-core offline-info` , appears to return exit code of 1 if no db file present, 0 if db present and > 1 if something else wrong,   running that would validate stellar-core bin, cfg file, and db file to some extent, or let that same thing happen when ingest attempts the run on first pass as-is.

ingest/ledgerbackend/stellar_core_runner.go
paulbellamy: Cool. Disregard me, then. 👍

ingest/ledgerbackend/toml.go
paulbellamy: I mean, if we add a `--in-memory` flag to horizon to force the old behaviour.

ingest/ledgerbackend/toml.go
paulbellamy: I mean, validate the params passed in.

ingest/ledgerbackend/toml.go
sreuland: ok, I added this in the [Later](https://stellarorg.atlassian.net/wiki/spaces/HOR/pages/4272980127/4038+-+Sqlite+for+captive-core#Later) section of confluence stream discussion for this. Will remove the TODO here.

ingest/ledgerbackend/toml.go
sreuland: yes, not much gain relative to  complexity of validation, maybe DATABASE being a valid o/s file name, no weird characters, but is stretch, I think we can remove this TODO or capture it in the [Later](https://stellarorg.atlassian.net/wiki/spaces/HOR/pages/4272980127/4038+-+Sqlite+for+captive-core#Later) if needed.

sreuland:
@paulbellamy , @ire-and-curses, here's performance comparison of Horizon and CaptiveCore ingestion between master(in-memory) and this branch(sqlite on disk):

Observed Performance Results:

test machine: MacBook Pro
cpu cores:  8 @ 600mhz - 3200mhz
ram: 16 GB

line-graphmem

line-graphdisk


**on master branch using captive core in-memory mode against pubnet:** 

- duration: 2 hours for horizon ingestion to catchup to current ledger from genesis
- notes: 
  *  Stellar-core took 8.5GB RAM within first 5 minutes and stayed there. 
  *  Horizon started at 100mb, reached about 2GB after 15 minutes, stayed there during ingestion, dropped to about 600MB RAM once caught up after 2 hours, when no more historical ledger replay.
  *  Captive core on-disk working directory went up to 7GB in first 5 minutes, and stayed there.
 
**on this branch using captive core sqlite on-disk against pubnet:** 

- duration: 2 hours for horizon ingestion to catchup to current ledger from genesis
- notes: 
  * Stellar-core started at 50mb, climbed up to about 150mb after 15 minutes, hovered around there.
  * Horizon started at 100mb, reached about 2GB after 15 minutes, stayed there during ingestion, dropped to about 600MB RAM once caught up after 2 hoiurs, when no more historical ledger replay.
  * Captive core on-disk working directory went gradually up from 1GB to 14GB in first 15mins, stayed there.

2opremio:
> 2 hours for horizon ingestion to catchup to current ledger from genesis

What does this exactly mean?

I am particularly interested in Horizon's ingestion throughput. And in the case of the SQLite variant, throughput vs disk IOPs.

sreuland:
> 2 hours for horizon ingestion to catchup to current ledger from genesis
> 
> What does this exactly mean?

it was the time during which Horizon was still ingesting the older ledgers from stellar-core meta stream, log output during this timeframe was primarily:

`INFO[2022-02-02T17:17:22.593-08:00] Processing ledger entry changes               pid=9610 processed_entries=50000 progress="0.41%" sequence=39462207 service=ingest source=historyArchive`

once that finished, a summary of ingestion stats was logged out:

```
INFO[2022-02-02T20:00:20.590-08:00] Processed state                               duration=2242.348303417 pid=9610 sequence=39462207 service=ingest stats_accounts_created=6522203 stats_accounts_removed=0 stats_accounts_updated=0 stats_claimable_balances_created=2004359 stats_claimable_balances_removed=0 stats_claimable_balances_updated=0 stats_data_created=54666 stats_data_removed=0 stats_data_updated=0 stats_liquidity_pools_created=2996 stats_liquidity_pools_removed=0 stats_liquidity_pools_updated=0 stats_offers_created=357107 stats_offers_removed=0 stats_offers_updated=0 stats_trust_lines_created=8138815 stats_trust_lines_removed=0 stats_trust_lines_updated=0
INFO[2022-02-02T20:00:21.591-08:00] Ingestion system state machine transition     current_state="buildFromCheckpoint(checkpointLedger=39462207)" next_state="resume(latestSuccessfullyProcessedLedger=39462207)" pid=9610 service=ingest
INFO[2022-02-02T20:00:21.592-08:00] Waiting for ledger to be available in the backend...  pid=9610 sequence=39462208 service=ingest
INFO[2022-02-02T20:00:21.592-08:00] Ledger returned from the backend              duration=6.375e-06 pid=9610 sequence=39462208 service=ingest
WARN[2022-02-02T20:00:21.598-08:00] Perf: 'MetaStream write' took 9789.231445 s   pid=9610 service=ingest subservice=stellar-core

```

the log timestamps show about 2.75 hours difference between when ingest service started and finished, at which point then the horizon log output cut over to processing current ledgers and 'Still Ingesting..' API errors went away.

sreuland:
Here are performance results for disk/mem observed on staging servers, they line up with what was seen locally, namely, memory usage goes down and disk usage goes up as captive core uses the on-disk sqlite instead:

[disk space used](https://grafana.stellar-ops.com/d/1OLed-uik/node-exporter-full?orgId=1&from=1644102254730&to=1644275054730&var-environment=stg&var-instance=horizon-pubnet-captivecore-001a.stg.stellar002&viewPanel=152)

[memory used](https://grafana.stellar-ops.com/d/1OLed-uik/node-exporter-full?orgId=1&from=1644102525323&to=1644275325324&var-environment=stg&var-instance=horizon-pubnet-captivecore-001a.stg.stellar002&viewPanel=135)

paulbellamy (commented):
LGTM, but I worked on it, so don't want to approve.

ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -278,15 +271,14 @@ func (c *CaptiveStellarCore) openOnlineReplaySubprocess(ctx context.Context, fro
 	return nil
 }
 
-// runFromParams receives a ledger sequence and calculates the required values to call stellar-core run with --start-ledger and --start-hash
-func (c *CaptiveStellarCore) runFromParams(ctx context.Context, from uint32) (runFrom uint32, ledgerHash string, err error) {
+// runFromParams receives a ledger sequence and calculates the required values to start stellar-core catchup from
+func (c *CaptiveStellarCore) runFromParams(ctx context.Context, from uint32) (runFrom uint32, err error) {
```
tamirms, line 275: nit: I think it would be better to remove the named return parameters and explicitly return the values for all return paths

ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -304,11 +307,26 @@ func (r *stellarCoreRunner) runFrom(from uint32, hash string) error {
 		return errors.New("runner already started")
 	}
 
+	if err := r.createCmd("new-db").Run(); err != nil {
+		return errors.Wrap(err, "error initializing core db")
+	}
+
+	// Do a quick catch-up to set the LCL in core to be our expected starting
+	// point.
+	//
```
tamirms, line 316: nit: blank line

services/horizon/internal/db2/history/ledger_test.go
```diff
@@ -61,10 +60,7 @@ func TestInsertLedger(t *testing.T) {
 	test.ResetHorizonDB(t, tt.HorizonDB)
 	q := &Q{tt.HorizonSession()}
 
-	ledgerHashStore := ledgerbackend.NewHorizonDBLedgerHashStore(tt.HorizonSession())
-	_, exists, err := ledgerHashStore.GetLedgerHash(tt.Ctx, 100)
-	tt.Assert.NoError(err)
-	tt.Assert.False(exists)
+	// TODO: Check the ledger doesn't exist in the db
```
tamirms, line 63: todo

services/horizon/internal/test/integration/integration.go
```diff
@@ -358,10 +358,6 @@ func (i *Test) StartHorizon() error {
 		HorizonURL: fmt.Sprintf("http://%s:%s", hostname, horizonPort),
 	}
 
-	if err = i.app.Ingestion().BuildGenesisState(); err != nil {
```
tamirms, line 361: why is this no longer necessary?

sreuland: observed during testing with stellar-core using on-disk sqlite, that `stellar-core run` command will not emit tx meta for ledger=2, regardless of any prior `new-db/catchup` commands that set LCL to it's earliest notion of genesis, instead it emits ledger=3 as the first ledger meta, which results in the test ingestion invocations that did this first getting hung since ingest fsm was looking for range starting with ledger=2, which only seems to get emitted by core when in-memory.

sreuland:
@tamirms , 
> will it still be possible to use captive core with all of its state kept in memory?

No, you are correct that this effort has changed the internal default run mode of captive core during ingestion to use the on disk sqlite, @paulbellamy noted the potential need for expose this as external horizon cmd flag to allow users to toggle this captive core mode, I suggested to table that for a future change request if it became needed. Do you think that needs to be available now? sounds like may need to add that in then as you both see that angle.

tamirms:
@sreuland I'm not sure that  we are configuring a default in this PR. Configuring a new default implies that the operator has the option to override the default but I don't think that's the case, right?

I am not opposed to changing the behavior so that from now on captive core *must* use a sqllite db since it seems like there is no degradation in ingestion throughput (e.g. time to ingest X ledgers) when using sqllite (assuming I understand your performance comparison correctly).

But, if we go with that route I think we need to add extra validation to the toml file so that a user provided toml cannot override the `DATABASE` field. We will need to add a bolded entry in the changelog so that operators are aware of the new performance characteristics of running horizon with captive core (in particular, now they need to make sure they have adequate disk space but they can forgo a bit of RAM).

Also, we should get approval from @jacekn  or someone else on the SRE team because they often have a different perspective when it comes to the rollout of significant features in horizon.

sreuland:
@tamirms , @jacekn , thanks for insights.

> @sreuland I'm not sure that we are configuring a default in this PR. Configuring a new default implies that the operator has the option to override the default but I don't think that's the case, right?

>Setting --in-memory to False by default should achieve that if I understand correctly? It will be very easy for anyone to set it to "true" if they want to enable it but at the same time people won't be surprised by it

correct, the ability for operator to configure memory or disk sqlite usage by captive core was not exposed as a new command line parameter rather this initial implementation had taken simpler approach to hardcode captive core invocation to use the disk sqlite and removed in-memory usage. But, per the feedback, I will add new horizon command line parameter `--captive-core-use-disk=false|true` with a default of `false` to provide the soft transition @jacekn recommended, so, by default, horizon will run captive core in with same in-memory behavior, and only if operator adds `--captive-core-use-disk=true` to horizon command line will it then trigger this new feature, will document the aspects of that usage in changelog.

@tamirms 
> if we go with that route I think we need to add extra validation to the toml file so that a user provided toml cannot override the DATABASE field. 

is it not ok to allow the operator to specify different file system path for where sqlite db file(stellar.db) is stored? the default file path location provided `sqlite3://stellar.db` results in the db file being created in the runtime current/working directory of horizon, operators may want to change that? What if add validation on the toml just to verify DATABASE is either absent(gets the default path) or if has value, it's a valid o/s path?

tamirms:
@sreuland 

> is it not ok to allow the operator to specify different file system path for where sqlite db file(stellar.db) is stored?

specifying a different path is ok. I was thinking if the operator provides a postgres url (see https://github.com/stellar/stellar-core/blob/master/docs/stellar-core_example.cfg#L40-L48 ) instead of a sqllite path. Will captive core work without any issues if a postgres DB is configured in the toml file?

sreuland:
we confirmed on further discussion in [slack thread](https://stellarfoundation.slack.com/archives/C0367MGFL/p1644865160327159?thread_ts=1644863613.125779&cid=C0367MGFL), to limit the DATABASE value for captive core config to just `sqlite://` db connection string, will add validation for that.

ingest/ledgerbackend/toml.go
```diff
@@ -404,6 +407,20 @@ func (c *CaptiveCoreToml) CatchupToml() (*CaptiveCoreToml, error) {
...
 }
 
+func (c *CaptiveCoreToml) ExternalLedgerStorageToml(fullStoragePath string) (*CaptiveCoreToml, error) {
+	newToml, err := c.clone()
+	if err != nil {
+		return nil, errors.Wrap(err, "could not clone toml")
+	}
+
+	if len(newToml.Database) == 0 {
+		newToml.Database = fmt.Sprintf("sqlite3://%v/stellar.db", fullStoragePath)
```
tamirms, line 417: is it necessary to include the absolute path? I think if you set the field to `sqlite3://stellar.db` it will write to the current working directory which will be in `fullStoragePath`

ingest/ledgerbackend/toml.go
```diff
@@ -404,6 +407,20 @@ func (c *CaptiveCoreToml) CatchupToml() (*CaptiveCoreToml, error) {
...
+	} else if !sqliteUrlRegEx.MatchString(newToml.Database) {
```
tamirms, line 418: I think this should be moved to inside the `validate()` function

ingest/ledgerbackend/toml.go
```diff
@@ -404,6 +407,20 @@ func (c *CaptiveCoreToml) CatchupToml() (*CaptiveCoreToml, error) {
 	return offline, nil
 }
 
+func (c *CaptiveCoreToml) ExternalLedgerStorageToml(fullStoragePath string) (*CaptiveCoreToml, error) {
```
tamirms, line 410: I don't think this function should be exported because it's only use in the ledgerbackend package

ingest/ledgerbackend/toml.go
```diff
@@ -30,6 +30,8 @@ var validQuality = map[string]bool{
 	"LOW":      true,
 }
 
+var sqliteUrlRegEx = regexp.MustCompile(`^sqlite3://\S+`)
```
tamirms, line 33: alternatively, we could use `strings.HasPrefix()` to check if it starts with `sqlite3://`

ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -124,6 +124,12 @@ type CaptiveCoreConfig struct {
 	// stored. We always append /captive-core to this directory, since we clean
 	// it up entirely on shutdown.
 	StoragePath string
+
+	// UseExternalStorageLedger, when true, instructs the core invocation to use an external db url
+	// for ledger states rather than in memory(RAM). The external db url is determined by the presence
+	// of DATABASE parameter in the captive-core-config-path or if absent, the db will default to sqlite
+	// and the db file will be stored at location derived from StoragePath parameter.
+	UseExternalStorageLedger bool
```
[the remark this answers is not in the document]
sreuland: I can change to that, the reason here was DATABASE captive core config param seemed like just a UNC path for captive core to interpret what the remote resource is, it's meaning in Horizon is mostly opaque and pass through, we did take the liberty of adding interpretation and validation in the captive core toml for DATABASE to be `sqlite` in short term for lower risk, as that came from chat with Jon, which he  mentioned other resource urls like postgres weren't precluded, but weren't really heavily used yet.

ingest/ledgerbackend/toml.go
tamirms: actually, I think we could get rid of this function if we add a `UseDB` field in CaptiveCoreTomlParams. Then we could set the default `Database` value in `setDefaults()`

ingest/ledgerbackend/toml.go
sreuland: the two could be different directories, run time(shell directory when horizon invoked) could be `/opt` but it may include `captive-core-storage-path=`/data/blah1` , in that case, would want located under `/data/blah1/captive-core/`

tamirms: I don't think that's possible because we explicity set the current working directory of the process to `fullStoreagePath`:

https://github.com/stellar/go/blob/21ca9cf2823139fe5a7ef7c8bc93d24b84877fe3/ingest/ledgerbackend/stellar_core_runner.go#L246

@Shaptic can confirm

ingest/ledgerbackend/toml.go
sreuland: `CaptiveCoreTomlParams` is for horizon flags that should be reflected into core's toml right, the DATABASE value isn't passed through horizon command line though, it just becomes a mandatory toml field when the `captive-core-use-db=true` flag is set, so there won't be a value for DATABASE in `CaptiveCoreTomlParams`, but, can follow this suggestion, and just set the default value of `sqlite://stellar.db` in `setDefaults()` when absent on toml.

ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -124,6 +124,12 @@ type CaptiveCoreConfig struct {
 	// stored. We always append /captive-core to this directory, since we clean
 	// it up entirely on shutdown.
 	StoragePath string
+
+	// UseExternalStorageLedger, when true, instructs the core invocation to use an external db url
```
tamirms, line 128: the godoc string should be updated with `UseDB` instead of `UseExternalStorageLedger`

ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -139,18 +141,18 @@ func newStellarCoreRunner(config CaptiveCoreConfig, mode stellarCoreRunnerMode)
...
 	if err != nil {
 		return "", err
 	}
 
 	return string(text), ioutil.WriteFile(location, text, 0644)
 }
 
-func generateConfig(captiveCoreToml *CaptiveCoreToml, mode stellarCoreRunnerMode) ([]byte, error) {
+func generateConfig(captiveCoreToml *CaptiveCoreToml, mode stellarCoreRunnerMode, config CaptiveCoreConfig) ([]byte, error) {
+	var err error
```
tamirms, line 154: we can move `var err error` back to line 153

ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -139,18 +141,18 @@ func newStellarCoreRunner(config CaptiveCoreConfig, mode stellarCoreRunnerMode)
...
+	text, err := generateConfig(captiveCoreToml, mode, config)
 	if err != nil {
 		return "", err
 	}
 
 	return string(text), ioutil.WriteFile(location, text, 0644)
 }
 
-func generateConfig(captiveCoreToml *CaptiveCoreToml, mode stellarCoreRunnerMode) ([]byte, error) {
+func generateConfig(captiveCoreToml *CaptiveCoreToml, mode stellarCoreRunnerMode, config CaptiveCoreConfig) ([]byte, error) {
```
tamirms, line 153: `config CaptiveCoreConfig` is no longer used. we can also remove it as a parameter from `writeConf()`

ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -304,13 +320,36 @@ func (r *stellarCoreRunner) runFrom(from uint32, hash string) error {
...
+			return errors.Wrap(err, "error initializing core db")
+		}
+		// Do a quick catch-up to set the LCL in core to be our expected starting
+		// point.
+		if from > 2 {
+			// Can't catch up to the genesis ledger.
+			if err := r.createCmd("catchup", fmt.Sprintf("%d/0", from-1)).Run(); err != nil {
+				return errors.Wrap(err, "error runing stellar-core catchup")
+			}
+		} else {
```
tamirms, line 334: I think it's better to do an `else if err := r.createCmd("catchup", "2/0").Run(); err != nil` here to avoid unnecessary nesting

ingest/ledgerbackend/toml.go
```diff
@@ -405,6 +406,11 @@ func (c *CaptiveCoreToml) CatchupToml() (*CaptiveCoreToml, error) {
 }
 
 func (c *CaptiveCoreToml) setDefaults(params CaptiveCoreTomlParams) {
+
+	if !c.tree.Has("DATABASE") {
```
tamirms, line 410: I think this needs to be `if params.UseDB && !c.tree.Has("DATABASE")` otherwise it will configure the toml `DATABASE` field to  `sqlite3://stellar.db` when using captive core with in-memory mode

ingest/ledgerbackend/stellar_core_runner.go
tamirms: I don't think my IDE warns about this I only noticed by looking at the diff

Issue #4051, opened 2021-11-06
Feature Request: Allow Efficient Access to Account "Actions" Where Account is Source


mootz12:
### What problem does your feature solve?

Ability to find account actions, where "account action" is defined as any modification to the ledger that the given account was the source of.

Currently, Horizon only allows us to find actions in which a given account was involved in. IE - it includes actions that affected it.

This allows for a few things:
1. Removes a risk / attack vector (detailed below) on any account where transaction/operation history is relevant
2. Allows the ability to quickly find all actions an account has committed, lowering query load against Horizon

#### Potential Solutions:
1. Add ability to filter by Operations by the `source_account` of the operations.

#### Attack Vector:

Consider a payroll app that has a specific account (Account 1) that does payroll for a company, and they store all information on chain to remove as much trust as possible. Users can run the app when they want to get paid, and it will pay them their salary for the time that has passed since they were last paid.

An attacker (say from the apps competition), could commit large amounts of 1 stroop payments to Account 1. When an employee tries to claim their pay next, the payroll app will have to wade through a TON of transactions / operations to find the last time Account 1 paid the employee, and worst case, might not be able to determine when the event occurred. 

### What would you like to see?

As new avenues open up for Stellar based applications to become more decentralized (like Stellar Turrets), allowing applications to safely find historical account actions will be vital to ensure these apps can be built securely and efficiently.

The method by which we find what actions an account caused isn't particularly important. We came up with one potential avenue that seemed like it would solve our use case:

#### Potential Solution In Detail

The solution would be to add a new DB query and endpoint where you could get Operations where the `source_account` was a given account. This likely would require `source_account` to be an indexed field.

Endpoint:
`/accounts/{account_id}/operations_as_source`

```java
// ForSourceAccount filters the operations collection to a specific source account
func (q *TransactionsQ) ForSourceAccount(ctx context.Context, aid string) *OperationsQ {
        // 1 - basic
	q.sql = q.sql.
		Where("hop.source_account = ?", aid)

        // 2 - include '' source_account operations
        q.sql = q.sql.Join(
		"history_transactions ht ON "+
			"ht.id = hop.transaction_id",
	).Where("hop.source_account = '' AND ht.account = ?", aid)
        
        // combine these results somehow
	return q
}
```

To my knowledge, any operations where the `source_account` field was not explicitly set defaults to '' in the `history_operations` table.

[...]

PR #316, opened 2018-02-16
Ingestion performance fixes

bartekn:
Ingestion code is making too many DB requests:
* It inserts each effect, operation, etc. in a separate query.
* It `GetCreateAccountID` is executed multiple times when checking row ID for a given `AccountID`.

Because of so many requests, ingesting a single transaction with 100 payment operations takes 200-500 ms. It can grow to 20 seconds per ledger causing delays in ingestion.

This PR introduces the following changes:
* It batches `INSERT`s to the database.
* It caches a result for `GetCreateAccountID` in memory. **EDIT** Cache no longer used. See comments below.

After implementing this changes ledger with 50 txs / 5000 operations is ingested in 3-5 seconds, ledger with 500 operations is ingested in 0.5 second.

However this is **not** ready to be merged. This needs to be addressed before merging:
* [x] What happens when account is merged? We need to invalidate cache. Fixed in 09fc13c.
* [x] For many effects/operations query length can exceed maximum number of params in Postgres. (`ERRO[0078] import session failed: exec failed: pq: got 100000 parameters but PostgreSQL only supports 65535 parameters  pid=15551`). Fixed in 98ddb01745c99229f98ed4c14bd5059fadf52e4e.

Other thoughts:
* ~We need to consider how quickly the cache can grow and what are the memory requirements.~ Cache no longer used.
* ~The cache makes sense when operations involve a small set of accounts. Otherwise, `GetCreateAccountID` may still be a bottleneck. Currently I can't find a better solution than DB schema redesign (so accounts are identified by `AccountID` instead of primary key in accounts table). Or maybe cache warmup on horizon init?~ Cache no longer used.
* Something else?

PR #4094, opened 2021-11-22
support/db: PoC: Use COPY instead of INSERT in BatchInsertBuilder

2opremio:
This is proof-of-concept PR which changes `BatchInsertBuilder` to use `COPY`instead of `INSERT`.

It brings a ~3.5x speedup.

Before:
```
goos: darwin
goarch: amd64
pkg: github.com/stellar/go/support/db
BenchmarkBatchInsertBuilder
BenchmarkBatchInsertBuilder-8   	      76	  14374755 ns/op
PASS
```

After:
```
goos: darwin
goarch: amd64
pkg: github.com/stellar/go/support/db
BenchmarkBatchInsertBuilder
BenchmarkBatchInsertBuilder-8   	     294	   3939888 ns/op
PASS
```

TODO:
- [ ] Allow `ON CONFLICT` insertions (this can be done by inserting on a temporary table. This can even speed up things further when indexes are present, since the temporary table doesn't require indexes. Related: #4087 and https://github.com/stellar/go/issues/4051).
- [ ] Safety implications of beginning a transaction within the batch insertion?


support/db/batch_insert_builder.go
```diff
@@ -58,20 +59,29 @@ func (b *BatchInsertBuilder) Row(ctx context.Context, row map[string]interface{}
...
+	_, err := b.stmt.ExecContext(ctx, rowSlice...)
+	return err
+}
 
-	// Call Exec when MaxBatchSize is reached.
-	if len(b.rows) == b.MaxBatchSize {
-		return b.Exec(ctx)
+func (b *BatchInsertBuilder) initStmt(ctx context.Context) error {
+	// TODO: could the transaction had been started before?
+	if err := b.Table.Session.Begin(); err != nil {
```
bartekn, line 68: Yes, in general `BatchInsertBuilder` is used inside a DB transaction of a single ledger ingestion, started before the first call to `Row()`. I explained some problems with `COPY` in https://github.com/stellar/go/pull/316#issuecomment-368990324.

bartekn (commented):
There are a couple of issues with using `COPY` especially connected to a DB transaction (linked in a comment). I think we can continue experimenting on this in a separate branch and decide later when we see all the pros and cons.

2opremio:
When using a temporary table (so that an insertion suffix can be used) we still get a ~3x speedup:

```
goos: darwin
goarch: amd64
pkg: github.com/stellar/go/support/db
BenchmarkBatchInsertBuilder
BenchmarkBatchInsertBuilder-8   	     216	   5379026 ns/op
PASS
```

2opremio:
I've had to move all the `COPY`ing to `Exec()` so that it happened all at once. Otherwise ingestion would intersperse other sql statements in the middle of the `COPY` leading to an error (`ERROR:  unexpected message type 0x50 during COPY from stdin`)

This reduces performance, but we are still at a ~2.7x improvement.

```
goos: darwin
goarch: amd64
pkg: github.com/stellar/go/support/db
BenchmarkBatchInsertBuilder
BenchmarkBatchInsertBuilder-8   	     204	   5994119 ns/op
PASS 
```

PR #4098, opened 2021-11-24
services/horizon: map client disconnects as Status 499 instead of 503 Problem

### LINKED ISSUE #3710: Improve HTTP metrics to distinguish client cancelled requests in 503 responses

2.4.1

### What did you do?

If client disconnects the `horizon_http_requests_duration_seconds_count` metric records this with `status=503` label. It's possible that it can affect alerting. We should be able to distinguish client initiated disconnects from actual 503 responses sent because of `ctx` being cancelled.

Maybe we should add another boolean label to `horizon_http_requests_duration_seconds_count` to be able to tell which 503 were a result of client disconnects.

sreuland:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Client Disconnects were being mapped by render/problem into ServiceUnavailable(503).
### Why
The problem was causing `horizon_http_requests_duration_seconds_count` metric keys to be generated with status=503 label which wasn't accurate, there was no server timeout on the request, rather just the opposite, the client disconnected from the socket. 

Closes #3710 

### Known limitations
Client socket disconnect is only detectable if load balancer/proxy being used does immediate close of upstream client connections or no LB and the client is connected directly to server. Detection of client socket disconnect relies on the golang http server detecting the socket close and in turn closing the Done channel on the http request context.

### DIFF
--- a/services/horizon/internal/actions/submit_transaction.go
+++ b/services/horizon/internal/actions/submit_transaction.go
@@ -78,7 +78,7 @@ func (handler SubmitTransactionHandler) response(r *http.Request, info envelopeI
 	}
 
 	if result.Err == txsub.ErrCanceled {
-		return nil, &hProblem.Timeout
+		return nil, &hProblem.ClientDisconnected
 	}
 
 	switch err := result.Err.(type) {
@@ -153,6 +153,6 @@ func (handler SubmitTransactionHandler) GetResource(w HeaderWriter, r *http.Requ
 	case result := <-submission:
 		return handler.response(r, info, result)
 	case <-r.Context().Done():
-		return nil, &hProblem.Timeout
+		return nil, &hProblem.ClientDisconnected
 	}
 }
--- a/services/horizon/internal/actions/submit_transaction.go
+++ b/services/horizon/internal/actions/submit_transaction.go
@@ -1,6 +1,7 @@
 package actions
 
 import (
+	"context"
 	"encoding/hex"
 	"mime"
 	"net/http"
@@ -16,8 +17,16 @@ import (
 	"github.com/stellar/go/xdr"
 )
 
+type NetworkSubmitter interface {
+	Submit(
+		ctx context.Context,
+		rawTx string,
+		envelope xdr.TransactionEnvelope,
+		hash string) <-chan txsub.Result
+}
+
 type SubmitTransactionHandler struct {
-	Submitter         *txsub.System
+	Submitter         NetworkSubmitter
 	NetworkPassphrase string
 	CoreStateGetter
 }
@@ -153,6 +162,9 @@ func (handler SubmitTransactionHandler) GetResource(w HeaderWriter, r *http.Requ
 	case result := <-submission:
 		return handler.response(r, info, result)
 	case <-r.Context().Done():
-		return nil, &hProblem.ClientDisconnected
+		if r.Context().Err() == context.Canceled {
+			return nil, hProblem.ClientDisconnected
+		}
+		return nil, hProblem.Timeout
 	}
 }
--- a/services/horizon/internal/httpx/server.go
+++ b/services/horizon/internal/httpx/server.go
@@ -56,7 +56,8 @@ func init() {
 	problem.RegisterError(sse.ErrRateLimited, hProblem.RateLimitExceeded)
 	problem.RegisterError(context.DeadlineExceeded, hProblem.Timeout)
 	problem.RegisterError(context.Canceled, hProblem.ClientDisconnected)
-	problem.RegisterError(db.ErrCancelled, hProblem.ServiceUnavailable)
+	problem.RegisterError(db.ErrCancelled, hProblem.ClientDisconnected)
+	problem.RegisterError(db.ErrTimeout, hProblem.ServiceUnavailable)
 	problem.RegisterError(db.ErrConflictWithRecovery, hProblem.ServiceUnavailable)
 	problem.RegisterError(db.ErrBadConnection, hProblem.ServiceUnavailable)
 }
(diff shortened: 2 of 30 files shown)


services/horizon/internal/test/http.go
```diff
@@ -10,20 +10,22 @@ import (
 )
 
 type RequestHelper interface {
-	Get(string, ...func(*http.Request)) *httptest.ResponseRecorder
-	Post(string, url.Values, ...func(*http.Request)) *httptest.ResponseRecorder
+	Get(string, ...func(*http.Request) *http.Request) *httptest.ResponseRecorder
+	Post(string, url.Values, ...func(*http.Request) *http.Request) *httptest.ResponseRecorder
```
bartekn, line 14: Could you elaborate why we need to change the definition here?

bartekn:
PR. Removed from "Horizon and SDKs" because there's a corresponding issue there (and this PR is linked).

services/horizon/internal/test/http.go
sreuland: returning a handle to the request  enables test clients to updating the Context in the http request object also as part of the 'modFn' composition. I used it here:
https://github.com/stellar/go/pull/4098/files#diff-34ad89baca4867eca4a36e2438e03844fe10f15d7163455815173f4936de5071R337

services/horizon/internal/actions/submit_transaction.go
```diff
@@ -153,6 +153,6 @@ func (handler SubmitTransactionHandler) GetResource(w HeaderWriter, r *http.Requ
 	case result := <-submission:
 		return handler.response(r, info, result)
 	case <-r.Context().Done():
-		return nil, &hProblem.Timeout
+		return nil, &hProblem.ClientDisconnected
```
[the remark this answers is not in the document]
sreuland: ah, yes, need to carve out the upstream disconnect from the downstream timeout, thanks, will do!

services/horizon/internal/actions/submit_transaction.go
```diff
@@ -78,7 +78,7 @@ func (handler SubmitTransactionHandler) response(r *http.Request, info envelopeI
 	}
 
 	if result.Err == txsub.ErrCanceled {
-		return nil, &hProblem.Timeout
+		return nil, &hProblem.ClientDisconnected
```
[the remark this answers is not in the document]
sreuland: yes, I tracked it to https://github.com/stellar/go/blob/master/services/horizon/internal/txsub/system.go#L230, which it derives from ctx.Done(), which that ctx is connected upstream to the client and downstream into the tx sub processing, It seems like I should do similar check there in system.go#L230 on ctx.Err() for Canceled vs other types and return txsub.errTimeout if other?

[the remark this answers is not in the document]
bartekn: Oh, you're right. I was thinking that `txsub` timeout is handled by `timeoutMiddleware` but it's actually maintaining it's own timer. So it looks your change is correct! 👍

services/horizon/internal/test/http.go
sreuland: I reworked the PR, removed the need for this change in test infra.

services/horizon/internal/actions/submit_transaction.go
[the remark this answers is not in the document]
sreuland: Ok, changed in [54b24c2e8653](https://github.com/stellar/go/pull/4098/commits/42ebd37a3c952a390f6d3997519d54b24c2e8653)

sreuland:
Hello @bartekn , I updated PR per feedback, should be ready for re-review, thanks!

services/horizon/internal/httpx/server.go
```diff
@@ -55,7 +55,7 @@ func init() {
 	problem.RegisterError(db2.ErrInvalidOrder, problem.BadRequest)
 	problem.RegisterError(sse.ErrRateLimited, hProblem.RateLimitExceeded)
 	problem.RegisterError(context.DeadlineExceeded, hProblem.Timeout)
-	problem.RegisterError(context.Canceled, hProblem.ServiceUnavailable)
+	problem.RegisterError(context.Canceled, hProblem.ClientDisconnected)
 	problem.RegisterError(db.ErrCancelled, hProblem.ServiceUnavailable)
```
bartekn, line 59: I think (we need to confirm this) that when context is cancelled while DB query is being executed it's returned as `db.ErrCancelled`. Could you check it and, if that's correct, can you also update the returned problem for `db.ErrCancelled` in line 59? It's also worth testing if the same error is returned in case of timeout. If that's the case then it will be too complicated to change so let's not add any changes to this PR then.

sreuland: Ok, yes, it looks like db.ErrCancelled is overloaded, i.e. it could represent upstream cancel or a downstream timeout, I can make adjust in session.go to break it out in same fashion done here with check on ctx.Err() and will need a new test for cancel vs. timeout. I found this one article pretty helpful is understanding the postgres timeout vs. cancel context nuances:
https://www.alexedwards.net/blog/how-to-manage-database-timeouts-and-cancellations-in-go

sreuland: Hello @bartekn , here's included distinction of the db conn timeout vs cancel:  [9a808356bff4cab47](https://github.com/stellar/go/pull/4098/commits/9a808356bff4cab475be9216cf764c663a689517)

bartekn (approved):
LGTM!

eliane345 (approved):
Desculpe _me

eliane345:
Desculpe...

PR #4102, opened 2021-11-29
exp/orderbook: Represent assets in orderbook graph as int32 instead of strings

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Previously we were representing the adjacency list for the graph as a mapping from asset string to the list of offers / pools which buy / sell that asset.

Now, every asset is assigned an integer id and the adjacency list is represented as an array where the integer id is used to index into the array. This new data structure is much more compact because the asset strings were very lengthy. The new data structure is also much faster because array indexing is significantly faster than looking up keys in a map.

### Why

New benchmark:

```
goos: darwin
goarch: amd64
BenchmarkVibrantPath
BenchmarkVibrantPath-12    	     100	  12064900 ns/op	 2811558 B/op	   67441 allocs/op
PASS
```

Old benchmark
```
goos: darwin
goarch: amd64
BenchmarkVibrantPath
BenchmarkVibrantPath-12    	      16	  74524903 ns/op	17888031 B/op	   65145 allocs/op
PASS
```

The new code reduced the latency from 74.5 ms per call to 12 ms per call. Also, it reduced the space from 17.8 mb per call to 2.8mb per call.

### Known limitations

[N/A]

### DIFF
--- a/exp/orderbook/graph.go
+++ b/exp/orderbook/graph.go
@@ -49,9 +49,19 @@ type OBGraph interface {
 // OrderBookGraph is an in-memory graph representation of all the offers in the
 // Stellar ledger.
 type OrderBookGraph struct {
-	assetStringToID map[string]int32
+	// idToAssetString maps an int32 asset id to its string representation.
+	// Every asset on the OrderBookGraph has an int32 id which indexes into idToAssetString.
+	// The asset integer ids are largely contiguous. When an asset is completely removed
+	// from the OrderBookGraph the integer id for that asset will be assigned to the next
+	// asset which is added to the OrderBookGraph.
 	idToAssetString []string
-	vacantIDs       []int32
+	// assetStringToID maps an asset string to its int32 id.
+	assetStringToID map[string]int32
+	// vacantIDs is a list of int32 asset ids which can be mapped to new assets.
+	// When a new asset is added to the OrderBookGraph we first check if there are
+	// any available vacantIDs, if so, we will assign the new asset to one of the vacantIDs.
+	// Otherwise, we will add a new entry to idToAssetString for the new asset.
+	vacantIDs []int32
 
 	// venuesForBuyingAsset maps an asset to all of its buying opportunities,
 	// which may be offers (sorted by price) or a liquidity pools.
@@ -200,12 +210,17 @@ func (graph *OrderBookGraph) getOrCreateAssetID(asset xdr.Asset) int32 {
 	if ok {
 		return id
 	}
+	// before creating a new int32 asset id we will try to use
+	// a vacant id so that we can plug any empty cells in the
+	// idToAssetString array.
 	if len(graph.vacantIDs) > 0 {
 		id = graph.vacantIDs[len(graph.vacantIDs)-1]
 		graph.vacantIDs = graph.vacantIDs[:len(graph.vacantIDs)-1]
 		graph.idToAssetString[id] = assetString
 	} else {
+		// idToAssetString never decreases in length unless we call graph.Clear()
 		id = int32(len(graph.idToAssetString))
+		// we assign id to asset
 		graph.idToAssetString = append(graph.idToAssetString, assetString)
 		graph.venuesForBuyingAsset = append(graph.venuesForBuyingAsset, nil)
 		graph.venuesForSellingAsset = append(graph.venuesForSellingAsset, nil)
@@ -221,6 +236,10 @@ func (graph *OrderBookGraph) maybeDeleteAsset(asset int32) {
 
 	if buyingEdgesEmpty && sellingEdgesEmpty {
 		delete(graph.assetStringToID, graph.idToAssetString[asset])
+		// When removing an asset we do not resize the idToAssetString array.
+		// Instead, we allow the cell occupied by the id to be empty.
+		// The next time we will add an asset to the graph we will allocate the
+		// id to the new asset.
 		graph.idToAssetString[asset] = ""
 		graph.vacantIDs = append(graph.vacantIDs, asset)
 	}
--- a/exp/orderbook/graph.go
+++ b/exp/orderbook/graph.go
@@ -28,9 +28,9 @@ const (
 // trading pair represents two assets that can be exchanged if an order is fulfilled
 type tradingPair struct {
 	// buyingAsset corresponds to offer.Buying.String() from an xdr.OfferEntry
-	buyingAsset string
+	buyingAsset int32
 	// sellingAsset corresponds to offer.Selling.String() from an xdr.OfferEntry
-	sellingAsset string
+	sellingAsset int32
 }
 
 // OBGraph is an interface for orderbook graphs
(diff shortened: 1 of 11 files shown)


exp/orderbook/graph.go
```diff
@@ -187,6 +194,38 @@ func (graph *OrderBookGraph) batch() *orderBookBatchedUpdates {
...
+	if len(graph.vacantIDs) > 0 {
+		id = graph.vacantIDs[len(graph.vacantIDs)-1]
+		graph.vacantIDs = graph.vacantIDs[:len(graph.vacantIDs)-1]
+		graph.idToAssetString[id] = assetString
+	} else {
+		id = int32(len(graph.idToAssetString))
+		graph.idToAssetString = append(graph.idToAssetString, assetString)
+		graph.venuesForBuyingAsset = append(graph.venuesForBuyingAsset, nil)
+		graph.venuesForSellingAsset = append(graph.venuesForSellingAsset, nil)
+	}
```
2opremio, line 212: I think this requires further clarification. Particularly as to why there are multiple ways to obtain the ID.

exp/orderbook/graph.go
```diff
@@ -49,12 +49,16 @@ type OBGraph interface {
 // OrderBookGraph is an in-memory graph representation of all the offers in the
 // Stellar ledger.
 type OrderBookGraph struct {
+	assetStringToID map[string]int32
+	idToAssetString []string
+	vacantIDs       []int32
```
2opremio, line 54: Please comment on what each of these is for.

exp/orderbook/graph.go
```diff
@@ -187,6 +194,38 @@ func (graph *OrderBookGraph) batch() *orderBookBatchedUpdates {
...
+	id, ok := graph.assetStringToID[assetString]
+	if ok {
+		return id
+	}
+	if len(graph.vacantIDs) > 0 {
+		id = graph.vacantIDs[len(graph.vacantIDs)-1]
+		graph.vacantIDs = graph.vacantIDs[:len(graph.vacantIDs)-1]
+		graph.idToAssetString[id] = assetString
+	} else {
+		id = int32(len(graph.idToAssetString))
```
2opremio, line 208: Can't this lead to id duplicates, since the length of `graph.idToAssetString` can decrease on asset deletion?

exp/orderbook/graph.go
2opremio: Also, do we really need the `vacantIDs` mechanism? It seems to complicate things and it's not obvious (to me at least) what gain it brings.

exp/orderbook/graph.go
tamirms: `idToAssetString` is an array of strings which never decreases. The array never shrinks. When we remove an asset the slot in  `idToAssetString` is left vacant so that the next asset we add to the graph can occupy it

exp/orderbook/graph.go
tamirms: the purpose of `vacantIDs` is to make sure the assets array does not waste any cells.

let's say we start out with an empty graph . the assets array will be empty in that case.
then we add the following assets:
0 -> 'usd'
1 -> 'eur'
2 -> 'chf'
3 -> 'sek'

now, we remove all offers and pools which have the chf asset so we can remove it. at this point the array looks like:

0 -> 'usd'
1 -> 'eur'
2 -> ''
3 -> 'sek'

the cell at index 2 is vacant and we can reuse it the next time we add a new asset, for example 'yen'


0 -> 'usd'
1 -> 'eur'
2 -> 'yen'
3 -> 'sek'

without the vacantIDs mechanism we will either have to add 'yen' to index 4 and forever let cell 2 to be empty, or we could try to reshuffle the mapping so there are no empty cells when remove 'chf' but that would be an expensive operation

exp/orderbook/graph.go
```diff
@@ -187,6 +204,47 @@ func (graph *OrderBookGraph) batch() *orderBookBatchedUpdates {
...
+		id = graph.vacantIDs[len(graph.vacantIDs)-1]
+		graph.vacantIDs = graph.vacantIDs[:len(graph.vacantIDs)-1]
+		graph.idToAssetString[id] = assetString
+	} else {
+		// idToAssetString never decreases in length unless we call graph.Clear()
+		id = int32(len(graph.idToAssetString))
+		// we assign id to asset
+		graph.idToAssetString = append(graph.idToAssetString, assetString)
+		graph.venuesForBuyingAsset = append(graph.venuesForBuyingAsset, nil)
+		graph.venuesForSellingAsset = append(graph.venuesForSellingAsset, nil)
```
bartekn, line 226: Shouldn't we clear `venuesForBuyingAsset` and `venuesForSellingAsset` when assigning to a vacant id? It seems we don't do that in `maybeDeleteAsset` either.

tamirms: in order to get included in the vacant id list it is a necessary condition that graph.venuesForBuyingAsset[asset] and graph.venuesForSellingAsset[asset] are empty:

```
func (graph *OrderBookGraph) maybeDeleteAsset(asset int32) {
	buyingEdgesEmpty := len(graph.venuesForBuyingAsset[asset]) == 0
	sellingEdgesEmpty := len(graph.venuesForSellingAsset[asset]) == 0

	if buyingEdgesEmpty && sellingEdgesEmpty {
		delete(graph.assetStringToID, graph.idToAssetString[asset])
		// When removing an asset we do not resize the idToAssetString array.
		// Instead, we allow the cell occupied by the id to be empty.
		// The next time we will add an asset to the graph we will allocate the
		// id to the new asset.
		graph.idToAssetString[asset] = ""
		graph.vacantIDs = append(graph.vacantIDs, asset)
	}
}
```

exp/orderbook/graph.go
2opremio: Ah, I see. Does wasting cells have a big performance impact? I presume we would be consuming a bit more memory, but maybe that's acceptable?

tamirms: I think there is a small performance impact because we iterate over all the cells in the search algorithm here:

https://github.com/stellar/go/blob/41b735e7aef4d07605b400e663716db97cc040a5/exp/orderbook/search.go#L141

For the empty cells we skip them through this check:

```
			currentAmount := bestAmount[currentAsset]
			if currentAmount == 0 {
				continue
			}
```

I think a few empty cells wouldn't matter but if we didn't fill in the empty cells I think the performance would slowly get worse over time as we accrue more and more empty cells until horizon restarts

2opremio (approved):

exp/orderbook/graph.go
```diff
@@ -187,6 +204,47 @@ func (graph *OrderBookGraph) batch() *orderBookBatchedUpdates {
...
 }
 
+func (graph *OrderBookGraph) getOrCreateAssetID(asset xdr.Asset) int32 {
+	assetString := asset.String()
+	id, ok := graph.assetStringToID[assetString]
+	if ok {
+		return id
+	}
+	// before creating a new int32 asset id we will try to use
+	// a vacant id so that we can plug any empty cells in the
```
sreuland, line 214: great design, just curious, would storing nil in idToAssetString equate to same result as maintaining separate vacancy state, i.e., iterate for `idToAssetString=nil` instead, perhaps for less code, but just wondering.

tamirms: yeah, that's true we could avoid having a `vacantIDs` list entirely if we scan through `idToAssetString` to find the first empty cell. in the worst case if there are no empty cells we have to scan through the entire array before realizing we have to append to the end. Having `vacantIDs` makes the operation of adding a new asset faster

sreuland (approved):
great insight and improvement.

PR #4104, opened 2021-11-29
exp/orderbook: Use uint128s for pool calculations to improve performance

2opremio:
Before:
```
goos: darwin
goarch: amd64
BenchmarkVibrantPath
BenchmarkVibrantPath-8                16          77022055 ns/op        18015408 B/op      70026 allocs/op
PASS
```

After:
```
goos: darwin
goarch: amd64
BenchmarkVibrantPath
BenchmarkVibrantPath-8                19          65228205 ns/op        17494446 B/op      43836 allocs/op
PASS
```

It improves CPU performance (by ~15%) and considerably reduces allocations (by ~40%)

### DIFF
--- a/exp/orderbook/pools.go
+++ b/exp/orderbook/pools.go
@@ -21,6 +21,7 @@ import (
 const (
 	tradeTypeDeposit     = iota // deposit into pool, what's the payout?
 	tradeTypeExpectation = iota // expect payout, what to deposit?
+	maxBips              = 10000
 )
 
 var (
@@ -28,6 +29,7 @@ var (
 	errBadPoolType   = errors.New("Unsupported liquidity pool: must be ConstantProduct")
 	errBadTradeType  = errors.New("Unknown pool exchange type requested")
 	errBadAmount     = errors.New("Exchange amount must be positive")
+	maxBips128       = uint128.From64(maxBips)
 )
 
 // makeTrade simulates execution of an exchange with a liquidity pool.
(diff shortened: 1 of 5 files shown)


tamirms (approved):
looks good to me but we should wait until @Shaptic reviews the pr

exp/orderbook/pools.go
```diff
@@ -91,32 +92,31 @@ func makeTrade(
...
-	maxBips := big.NewInt(10000)
-	f := new(big.Int).Sub(maxBips, F) // upscaled 1 - F
+	maxBips := uint128.From64(10000)
+	f := maxBips.Sub(F) // upscaled 1 - F
 
 	// right half: X + (1 - F)x
-	denom := X.Mul(X, maxBips).Add(X, new(big.Int).Mul(x, f))
-	if denom.Cmp(big.NewInt(0)) == 0 { // avoid div-by-zero panic
+	denom := X.Mul(maxBips).Add(x.Mul(f))
+	if denom.Cmp64(0) == 0 { // avoid div-by-zero panic
```
bartekn, line 109:
```suggestion
	if denom.IsZero() { // avoid div-by-zero panic
```

exp/orderbook/pools.go
```diff
@@ -91,32 +92,31 @@ func makeTrade(
...
 		return 0, false
 	}
 
 	// left half, a: (1 - F) Yx
-	numer := Y.Mul(Y, x).Mul(Y, f)
+	numer := Y.Mul(x).Mul(f)
 
 	// divide & check overflow
-	result := numer.Div(numer, denom)
+	result := numer.Div(denom)
```
bartekn, line 117: Some `Uint128` methods panic on overflow so I wonder if it can happen before here.

2opremio: `math/big` (AFAIU) doesn't overflow because it simply extends the precision as much as needed.

bartekn: Yeah, but `Uint128` can overflow.

2opremio: I added an overflow test. We are fine with using factors equaling `MaxInt64` but ... we cannot use `poolFeeBips` > 10000 (which causes an underflow). I think that probably isn't acceptable, but let's check with @Shaptic  first.

Shaptic (approved):
Nice lib find, porting LGTM :+1:

exp/orderbook/pools.go
```diff
@@ -91,32 +94,31 @@ func makeTrade(
 //
 // It returns false if the calculation overflows.
 func calculatePoolPayout(reserveA, reserveB, received xdr.Int64, feeBips xdr.Int32) (xdr.Int64, bool) {
```
tamirms, line 96: After replaying some production requests I found some cases where calculatePoolPayout panics with an overflow.
Here is one case which causes a panic:

calculatePoolPayout(92017260901926686, 9157376027422527, 4000000000000000000, 30)

PR #4105, opened 2021-11-29
xdr and exp/orderbook: Reduce path search allocations

2opremio:
This considerably reduces the allocations of path search by:

1. Allocating `pathNode`s through a slab (which piggyback's on `append()`s exponential allocation).
2. Preallocating some slices
3. Improving asset extraction to strings.
3. ~Optimizing strkey encoding. This is done in two ways:
    a. By creating `strkey.EncodingBuffer` which reuses the internal buffers between calls (much like https://github.com/stellar/go/pull/4056 ).
    b. By optimizing `strkey` encoding in general (there were a few things which were slow or caused uncessary allocartions, like using `fmt.Sprintf()` and using an array to obtain the `crc16`).~ This will be done at #4107 instead

Before:

```
goos: darwin
goarch: amd64
BenchmarkVibrantPath
BenchmarkVibrantPath-8               100          11918038 ns/op         2772506 B/op      65448 allocs/op
PASS
```

After:

```
goos: darwin
goarch: amd64
BenchmarkVibrantPath
BenchmarkVibrantPath-8   	     120	   9873280 ns/op	 2861917 B/op	   26235 allocs/op
PASS
```

This is an ~2.49x allocation reduction and a 20% speedup.

Most of the remaining allocations are removed at https://github.com/stellar/go/pull/4104

Depends on https://github.com/stellar/go/pull/4102


strkey/main.go
```diff
@@ -111,33 +111,76 @@ func MustDecode(expected VersionByte, src string) []byte {
 	return d
 }
 
-// Encode encodes the provided data to a StrKey, using the provided version
-// byte.
-func Encode(version VersionByte, src []byte) (string, error) {
+type Encoder struct {
```
2opremio, line 114: Maybe I should call this `EncodingBuffer` like at https://github.com/stellar/go/pull/4105 ?

exp/orderbook/graph.go
```diff
@@ -356,27 +425,29 @@ func (graph *OrderBookGraph) FindFixedPaths(
...
 	for _, destinationAsset := range destinationAssets {
-		destinationAssetString := destinationAsset.String()
-		target[destinationAssetString] = true
+		destinationAssetString := destinationAsset.StringWithEncoder(graph.strkeyEncoder)
+		target[graph.assetStringToID[destinationAssetString]] = true
 	}
 
+	// Initialize the capacity of paths to minimize allocations
+	// TODO: choose value wisely
+	pathsWithCapacity := make([]Path, 0, 64)
```
2opremio, line 436: @tamirms Please comment on the chosen length. Happy to remove this if it simply ends up being a magic constant.

exp/orderbook/search.go
```diff
@@ -78,49 +65,50 @@ type searchState interface {
...
 		path[i], path[opp] = path[opp], path[i]
 	}
 }
 
-func (e *pathNode) path() []xdr.Asset {
-	var result []xdr.Asset
+func (e *pathNode) path() []int32 {
+	// Initialize slice capacity to minimize allocations
+	// TODO: choose value wisely
+	result := make([]int32, 0, 8)
```
2opremio, line 111: @tamirms Please comment on the chosen length. Happy to remove this if it simply ends up being a magic constant.

exp/orderbook/search.go
```diff
@@ -177,11 +179,14 @@ func search(
...
+						// By piggybacking on slice appending (which uses exponential allocation)
+						// we avoid allocating each node individually, which is much slower and
+						// puts more pressure on the garbage collector.
+						slab = append(slab, pathNode{
 							asset: nextAsset,
 							prev:  pathToCurrentAsset,
-						}
-						updatedAssets = append(updatedAssets, nextAsset)
+						})
+						updatePath[nextAsset] = &slab[len(slab)-1]
```
tamirms, line 191: we shouldn't delete this line of code:
```
updatedAssets = append(updatedAssets, nextAsset)
```

once you add that the tests will pass

exp/orderbook/search.go
tamirms: we currently are using a max path length of 3 but that is configurable. 8 is the maximum supported by stellar core so I think it's a good default.

exp/orderbook/graph.go
tamirms: There are actually very few paths in the response so I don't think allocating the specific capacity will make the code much faster. I would prefer to remove this unless I'm wrong and the default capacity does speed up the algorithm

exp/orderbook/graph.go
```diff
@@ -348,10 +352,10 @@ func (graph *OrderBookGraph) FindPaths(
 	maxAssetsPerPath int,
 	includePools bool,
 ) ([]Path, uint32, error) {
-	destinationAssetString := destinationAsset.String()
+	destinationAssetString := destinationAsset.StringWithEncoder(graph.strkeyEncoder)
```
tamirms, line 355: is `graph.strkeyEncoder ` thread safe?  `FindPaths()` can be called concurrently

2opremio: Ops, no, it is not due to the scratch buffer. I will make it thread-safe and see if it still performs well.

strkey/main.go
2opremio: Done

exp/orderbook/graph.go
2opremio: Makes sense

exp/orderbook/search.go
2opremio: Great!

exp/orderbook/graph.go
2opremio: I added a small critical section in the `Find*Paths()` functions

PTAL

tamirms: @2opremio is there still a performance improvement with the critical section?

2opremio: Yes

2opremio: I think it's worth it because of the loop.

strkey/main.go
```diff
@@ -111,33 +111,78 @@ func MustDecode(expected VersionByte, src string) []byte {
...
+		panic(err)
+	}
+	return res
+}
+
+// Encode encodes the provided data to a StrKey, using the provided version
+// byte.
+func Encode(version VersionByte, src []byte) (string, error) {
+	encoder := NewEncodingBuffer()
+	return encoder.Encode(version, src)
```
leighmcculloch, line 185: ❓ The changes in this PR to strkey optimize for someone like Horizon who is using EncodingBuffer at the expense of someone calling strkey.Encode.

Not everyone can use EncodingBuffer. It would be difficult for us to use EncodingBuffer in starlight for example because EncodingBuffer cannot be used concurrently.

A quick allocs benchmark of `strkey.Encode` shows that:
- master (02dbef9662e93e9f85123575da2262e6784b6b8f) is 244 B/op and 6 allocs/op.
- this PR (3145ec2bc64c7ffeee4c30d008ba1d97356b0f45) is 656 B/op and 6 allocs/op.

<details>
<summary>Benchmarks</summary>

```go
func BenchmarkEncode_accountID(b *testing.B) {
	accountID := make([]byte, 32)
	for i := 0; i < b.N; i++ {
		_, _ = strkey.Encode(strkey.VersionByteAccountID, accountID)
	}
}
```

</details>

So the number of allocs are the same, but their size is 3x, almost 1KB to encode a single ~32byte strkey.

Improving strkey cost is of interest to me because at some point I noticed base32 encoding was a cost in Starlight. I think we can improve allocs within strkey.Encode for a greater set of use cases, including concurrent, by making some different changes. A quick sketch of the idea results in much less allocs and much less mem usage for all users of the function:
- PR #4107 is 64 B/op and 1 allocs/op.

leighmcculloch (commented):
I'm interested in strkey performance improvements because github.com/stellar/starlight uses strkey.Encode heavily and just looked at them. I left some thoughts, none-blocking if this is urgent. I defer to Horizon folks for the PR as a whole.

2opremio:
I removed `strkey.EncodingBuffer()` in favor of #4107

2opremio:
@tamirms PTAL

strkey/main.go
2opremio: Agreed. I reverted the encoding buffer approach in favor of #4107

strkey/main.go
```diff
@@ -111,33 +111,78 @@ func MustDecode(expected VersionByte, src string) []byte {
...
 		return "", err
 	}
 
 	// calculate and write checksum
-	checksum := crc16.Checksum(raw.Bytes())
-	if _, err := raw.Write(checksum); err != nil {
+	checksum := crc16.Checksum(e.buf.Bytes())
+	var crc [2]byte
+	binary.LittleEndian.PutUint16(crc[:], checksum)
+	if _, err := e.buf.Write(crc[:]); err != nil {
```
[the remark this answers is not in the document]
2opremio: Thanks. I reverted it for #4107 though

tamirms (approved):

Issue #1769, opened 2019-09-23
strkey: check key length validation methods


bartekn:
While checking Horizon diff I noticed that https://github.com/stellar/go/pull/1573 is not checking the keys properly. It checks if the values are properly encoded but it doesn't check the length of the keys (should be 32 bytes). Check js-sdk for comparison: https://github.com/stellar/js-stellar-base/blob/f9d03ce4ffe0989851d777abf2e934fdbf36b514/src/strkey.js#L119

PR #4107, opened 2021-11-29
strkey: reduce allocs in Encode

leighmcculloch:
### What
Replace bytes.Buffer with working with a byte array directly.

Also, contains the crc16 improvements by @2opremio that were in #4105.

### Why
Reduce the allocations in strkey.Encode from 6 to 1, and reduce the heap use from 244 B/op to 64 B/op.

In the past I have seen impact to performance of Starlight caused by the strkey.Encode functions. @2opremio also highlighted in #4105 there is an impact there.

The use of a bytes.Buffer in this code is rather unnecessary given we know the size of everything, and given that the size of strkeys are small encouraging their data onto the stack is fine.

It is worth noting that there is a functional impact by this change. Attempts to strkey.Encode a data value greater than the max payload size as defined by SEP-23 will now error. We've been wanting to add length checks (#1769) for sometime, so I think this is fine, and we can continue to improve those length checks in the future. 

#### Before

```
BenchmarkDecode_accountID-8      2576450               414.2 ns/op           130 B/op          3 allocs/op
BenchmarkEncode_accountID-8      3657649               319.6 ns/op           244 B/op          6 allocs/op
```

#### After

```
BenchmarkDecode_accountID-8      3563737               334.9 ns/op           128 B/op          2 allocs/op
BenchmarkEncode_accountID-8      7365306               165.4 ns/op            64 B/op          1 allocs/op
```

Co-authored-by: Alfonso Acosta <2362916+2opremio@users.noreply.github.com>


leighmcculloch:
I originally opened this as an example of an idea I wanted to share on #4105, but in an attempt to validate the idea, I think this is essentially complete and could be merged. What do folks think?

strkey/main.go
```diff
@@ -36,6 +36,13 @@ const (
 	VersionByteHashX = 23 << 3 // Base32-encodes to 'X...'
 )
 
+// maxPayloadSize is the maximum length of the payload for all versions.
+const maxPayloadSize = 43
+
+// encoding is the encoding to use when encoding and decoding the strkey to and
+// from strings.
+var encoding = base32.StdEncoding.WithPadding(base32.NoPadding)
```
2opremio, line 44: Is this thread-safe?

2opremio (approved):
Minor comment. Otherwise approved.

2opremio:
@tamirms do you think we should include this for the release? (it will reduce allocations in the path search algorithm).

If so, @leighmcculloch , could you change the base branch to `release-horizon-v2.12.0`. 

Thanks!

strkey/main.go
leighmcculloch: Yup. The type is a base32.Encoding. That's the same type as the base32.StdEncoding variable.

2opremio:
@leighmcculloch please go ahead and change the base branch if you don't mind. Thanks!

strkey/main.go
```diff
@@ -36,6 +36,13 @@ const (
 	VersionByteHashX = 23 << 3 // Base32-encodes to 'X...'
 )
 
+// maxPayloadSize is the maximum length of the payload for all versions.
+const maxPayloadSize = 43
```
bartekn, line 40: This is confusing. Max payload size is 40 for muxed (32+8) but later it's used like:
```go
const maxSize = 1 + maxPayloadSize + 2
```
So `maxSize` is 46 according to this. I guess we want to change `maxPayloadSize` to 40? Also worth changing `// test too long payload` test check to 41 bytes.

strkey/main.go
```diff
@@ -118,26 +125,32 @@ func Encode(version VersionByte, src []byte) (string, error) {
...
+		return "", fmt.Errorf("data exceeds maximum payload size for strkey")
 	}
 
-	result := base32.StdEncoding.WithPadding(base32.NoPadding).EncodeToString(raw.Bytes())
-	return result, nil
+	// pack
+	//  1 byte version
+	//  src bytes
+	//  2 byte crc16
+	const maxSize = 1 + maxPayloadSize + 2
```
bartekn, line 137: Probably a micro-optimization but can be moved outside this method.

strkey/main.go
```diff
@@ -118,26 +125,32 @@ func Encode(version VersionByte, src []byte) (string, error) {
...
+	rawArr := [maxSize]byte{}
+	size := 1 + len(src) + 2
+	raw := rawArr[:size]
+	raw[0] = byte(version)
+	copy(raw[1:], src)
+	crc := crc16.Checksum(raw[:1+len(src)])
```
bartekn, line 143: Looks like it's the same but more readable:
```suggestion
	crc := crc16.Checksum(src)
```

bartekn (approved):
LGTM! Just one comment about confusing const name.

strkey/main.go
leighmcculloch: You're right. I think I got confused, calculated the max strkey size, and used that here. Updated to 40 here: 844fd728.

strkey/main.go
leighmcculloch: It needs to calculate the checksum including the leading version byte. That's why it uses the raw string. I did try calculating the checksum in parts, first the version byte, then the src string, and I don't think the readability was better nor was there any performance benefit.

bartekn: I see, ok!

strkey/main.go
leighmcculloch: I moved it and the encoded max length to the package scope in c9884fc. Also, this had a run on benefit that I can now right a test to ensure that the encoded max length is correctly calculated and matches the encoded length that the encoder will produce 🎉. a9c29f5 Thanks!

leighmcculloch:
> @tamirms do you think we should include this for the release? (it will reduce allocations in the path search algorithm).
> 
> If so, @leighmcculloch , could you change the base branch to `release-horizon-v2.12.0`?

I've pushed this to master. Can you cherry-pick to the release if it is needed there?

PR #4108, opened 2021-11-30
Revert "exp/orderbook: Use uint128s for pool calculations to improve performance"

tamirms:
Reverts stellar/go#4104

After replaying some production requests I found some cases where calculatePoolPayout panics with an overflow.
Here is one case which causes a panic:

calculatePoolPayout(92017260901926686, 9157376027422527, 4000000000000000000, 30)

### DIFF
--- a/exp/orderbook/pools.go
+++ b/exp/orderbook/pools.go
@@ -2,8 +2,7 @@ package orderbook
 
 import (
 	"math"
-
-	"lukechampine.com/uint128"
+	"math/big"
 
 	"github.com/stellar/go/support/errors"
 	"github.com/stellar/go/xdr"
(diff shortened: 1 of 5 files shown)


paulbellamy (approved):
Could you add the points where it fails here, so when we go to fix it it's easy to reproduce?

tamirms:
To reproduce:

calculatePoolPayout(92017260901926686, 9157376027422527, 4000000000000000000, 30) will panic with an overflow on this line of code:

https://github.com/stellar/go/blob/b5d20587debc3ec2f7bd982aaf8a85a115e538d8/exp/orderbook/pools.go#L116

bartekn:
Does the previous code return some value for these params? If not (and just returns false on 2nd argument because of the overflow) we should recover from the panic and do the same thing?

tamirms:
@bartekn the previous code returns false on the 2nd argument because of the overflow

2opremio:
`unit128` has wraparound operations with which I think we can detect the overflow without panicking. I can take a look, but I don't think I can solve it before Friday.

PR #4109, opened 2021-11-30
exp/orderbook: Exclude arbitrage paths from path finding search

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

One of the changes to the path finding algorithm introduced by https://github.com/stellar/go/pull/4096 is that the new algorithm includes paths which take advantage of arbitrage opportunities. In this commit, I have removed that behavior so that no arbitrage paths will be included in the response.

### Why

The horizon v2.12.0 release contains substantial implementation changes to the path finding algorithm. Removing the differences in outputs between the old and new algorithm will make it easier to test. We can always include the arbitrage paths functionality in a future release.

Clients which always pick the first path from the path finding response and use it in a path payment operation might be frustrated when the first path is an arbitrage path which will have a lot of contention. You can imagine a situation where the client submits the path payment but it fails because the offers which lead to the arbitrage have already been consumed by another transaction.

### Known limitations

[N/A]

### DIFF
--- a/exp/orderbook/graph_test.go
+++ b/exp/orderbook/graph_test.go
@@ -1472,17 +1472,6 @@ func TestFindPaths(t *testing.T) {
 
 	expectedPaths := []Path{
 		{
-			// arbitrage usd then trade to xlm
-			SourceAmount: 2,
-			SourceAsset:  usdAsset.String(),
-			InteriorNodes: []string{
-				eurAsset.String(),
-				usdAsset.String(),
-			},
-			DestinationAsset:  nativeAsset.String(),
-			DestinationAmount: 20,
-		},
-		{
 			SourceAmount:      5,
 			SourceAsset:       usdAsset.String(),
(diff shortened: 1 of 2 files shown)


bartekn (approved):

sreuland (approved):
LGTM, has related test coverage. I'm kinda new to the area, so other reviews may be worthwhile.

Shaptic (approved):
LGTM, as long as my assumption that the removed tests fail with this change. (Maybe we should have keep a test that's _expected_ to fail?)

tamirms:
> LGTM, as long as my assumption that the removed tests fail with this change.

there are no removed test cases. I removed an arbitrage path from the expected response in the unit test. the unit test requires the response match the expected response otherwise the test will fail. in other words, the expected response contains no arbitrage paths but the graph in the test case actually has an arbitrage opportunity

PR #4113, opened 2021-12-01
exp/orderbook: Use uint128s for pool calculations to improve performance

2opremio:
Second attempt at #4104 by explicitly checking every possible overflow.

Before:

```
goos: darwin
goarch: amd64
BenchmarkVibrantPath
BenchmarkVibrantPath-8   	     100	  10375399 ns/op	 2758209 B/op	   64732 allocs/op
PASS
```

After:

```
goos: darwin
goarch: amd64
BenchmarkVibrantPath
BenchmarkVibrantPath-8   	     127	   9447364 ns/op	 2251733 B/op	   39260 allocs/op
PASS
```

### DIFF
--- a/exp/orderbook/pools.go
+++ b/exp/orderbook/pools.go
@@ -21,6 +21,7 @@ import (
 const (
 	tradeTypeDeposit     = iota // deposit into pool, what's the payout?
 	tradeTypeExpectation = iota // expect payout, what to deposit?
+	maxBips              = 10000
 )
 
 var (
@@ -28,6 +29,7 @@ var (
 	errBadPoolType   = errors.New("Unsupported liquidity pool: must be ConstantProduct")
 	errBadTradeType  = errors.New("Unknown pool exchange type requested")
 	errBadAmount     = errors.New("Exchange amount must be positive")
+	maxBips128       = uint128.From64(maxBips)
 )
 
 // makeTrade simulates execution of an exchange with a liquidity pool.
@@ -96,16 +98,16 @@ func calculatePoolPayout(reserveA, reserveB, received xdr.Int64, feeBips xdr.Int
 	F, x := uint128.From64(uint64(feeBips)), uint128.From64(uint64(received))
 
 	// would this deposit overflow the reserve?
-	if received > math.MaxInt64-reserveA {
+	// is feeBips within range?
+	if received > math.MaxInt64-reserveA || feeBips > maxBips {
 		return 0, false
 	}
 
 	// We do all of the math in bips, so it's all upscaled by this value.
-	maxBips := uint128.From64(10000)
-	f := maxBips.Sub(F) // upscaled 1 - F
+	f := maxBips128.Sub(F) // upscaled 1 - F
 
 	// right half: X + (1 - F)x
-	denom := X.Mul(maxBips).Add(x.Mul(f))
+	denom := X.Mul(maxBips128).Add(x.Mul(f))
 	if denom.IsZero() { // avoid div-by-zero panic
 		return 0, false
 	}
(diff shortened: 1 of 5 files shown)


tamirms:
I tried comparing the uint128 calculations with the old code which uses math/big ones:

```

func TestCalculatePoolExpectations(t *testing.T) {
	for i := 0; i < 1000000; i++ {
		reserveA := xdr.Int64(rand.Int63())
		reserveB := xdr.Int64(rand.Int63())
		disbursed := xdr.Int64(rand.Int63())

		result, ok := calculatePoolExpectationBig(reserveA, reserveB, disbursed, 30)
		result1, ok1 := calculatePoolExpectation(reserveA, reserveB, disbursed, 30)
		if assert.Equal(t, ok, ok1) {
			assert.Equal(t, result, result1)
		}
	}
}

func TestCalculatePoolPayout(t *testing.T) {
	for i := 0; i < 1000000; i++ {
		reserveA := xdr.Int64(rand.Int63())
		reserveB := xdr.Int64(rand.Int63())
		received := xdr.Int64(rand.Int63())

		result, ok := calculatePoolPayoutBig(reserveA, reserveB, received, 30)
		result1, ok1 := calculatePoolPayout(reserveA, reserveB, received, 30)
		if assert.Equal(t, ok, ok1) {
			assert.Equal(t, result, result1)
		}
	}
}
```

Unfortunately there were a lot of cases where the results differed because there were cases where the numerator overflowed but after dividing by the denominator the result was in the bounds of an int64.


But I did find https://github.com/holiman/uint256 . The library is mostly compatible with the big.Int API so it was easy to swap in.  When trying out the library I was able to get all the tests to pass and there was a decent performance improvement as well:

Before:
```
goos: darwin
goarch: amd64
BenchmarkVibrantPath
BenchmarkVibrantPath-12    	     100	  11879873 ns/op	 2794719 B/op	   66595 allocs/op
PASS
```

After:
```
goos: darwin
goarch: amd64
BenchmarkVibrantPath
BenchmarkVibrantPath-12    	     100	  10841301 ns/op	 2259508 B/op	   39745 allocs/op
PASS
```

tamirms (approved):
I have tested this by running through path finding requests on this branch and comparing the output with the result from the old code. Everything matches

exp/orderbook/pools.go
```diff
@@ -128,34 +128,36 @@ func calculatePoolPayout(reserveA, reserveB, received xdr.Int64, feeBips xdr.Int
...
-	result, rem := new(big.Int), new(big.Int)
-	result.DivMod(numer, denom, rem)
+	result, rem := new(uint256.Int), new(uint256.Int)
+	result.Div(numer, denom)
+	rem.Mod(numer, denom)
 
 	// hacky way to ceil(): if there's a remainder, add 1
-	if rem.Cmp(big.NewInt(0)) > 0 {
-		result.Add(result, big.NewInt(1))
+	if !rem.IsZero() {
```
2opremio, line 155: In some implementations, the reminder can be negative. But I think that only happens with negative integers. We should double-check to be sure since `IsZero()` maybe not be semantically equivalent.

exp/orderbook/pools.go
```diff
@@ -128,34 +128,36 @@ func calculatePoolPayout(reserveA, reserveB, received xdr.Int64, feeBips xdr.Int
...
 
 	// sanity check: disbursing shouldn't underflow the reserve
 	if disbursed >= reserveB {
 		return 0, false
 	}
 
 	// We do all of the math in bips, so it's all upscaled by this value.
-	maxBips := big.NewInt(10000)
-	f := new(big.Int).Sub(maxBips, F) // upscaled 1 - F
+	maxBips := uint256.NewInt(10000)
```
2opremio, line 140: It may be worth having a global var/constant for maxBips

exp/orderbook/pools.go
```diff
@@ -128,34 +128,36 @@ func calculatePoolPayout(reserveA, reserveB, received xdr.Int64, feeBips xdr.Int
...
 		return 0, false
 	}
 
 	numer := X.Mul(X, y).Mul(X, maxBips) // left half: Xy
 
-	result, rem := new(big.Int), new(big.Int)
-	result.DivMod(numer, denom, rem)
+	result, rem := new(uint256.Int), new(uint256.Int)
+	result.Div(numer, denom)
+	rem.Mod(numer, denom)
```
2opremio, line 152: I don't get why the library doesn't expose `DivMod` since that's how it implements `Div()` and `Mod()` internally:

https://github.com/holiman/uint256/blob/38e2c56bf6d3ee96b9d391351a612336b0b552a5/uint256.go#L485

exp/orderbook/pools.go
```diff
@@ -115,8 +115,8 @@ func calculatePoolPayout(reserveA, reserveB, received xdr.Int64, feeBips xdr.Int
 	// divide & check overflow
 	result := numer.Div(numer, denom)
 
-	i := xdr.Int64(result.Int64())
-	return i, result.IsInt64() && i > 0
+	val := xdr.Int64(result.Uint64())
```
2opremio, line 118: I wonder it it would be simpler and more performant to use `Uint64WithOverflow()`

exp/orderbook/pools.go
2opremio: I created a ticket at https://github.com/holiman/uint256/issues/103

exp/orderbook/pools.go
2opremio: (we would still need to check that the result is `<=` than `math.MaxInt64` though)

2opremio:
@tamirms https://github.com/holiman/uint256 seems to have overflow/underflow-checking alternatives for all the operations.

I wonder if we should be using those, since, unlike lukechampine.com/uint128, https://github.com/holiman/uint256 doesn't seem to panic on overflows.

PR #4125, opened 2021-12-03
exp/orderbook: Micro-optimization for paths search

2opremio:
Leftover from https://github.com/stellar/go/pull/4105

I just found it in a local branch and I realized I forgot to push it.

Very very minor stuff.

### DIFF
--- a/exp/orderbook/graph.go
+++ b/exp/orderbook/graph.go
@@ -423,7 +423,7 @@ func (graph *OrderBookGraph) FindFixedPaths(
 	maxAssetsPerPath int,
 	includePools bool,
 ) ([]Path, uint32, error) {
-	target := map[int32]bool{}
+	target := make(map[int32]bool, len(destinationAssets))
 	for _, destinationAsset := range destinationAssets {
 		destinationAssetString := destinationAsset.String()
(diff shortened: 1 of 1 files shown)


tamirms (approved):

exp/orderbook/graph.go
```diff
@@ -423,7 +423,7 @@ func (graph *OrderBookGraph) FindFixedPaths(
 	maxAssetsPerPath int,
 	includePools bool,
 ) ([]Path, uint32, error) {
-	target := map[int32]bool{}
+	target := make(map[int32]bool, len(destinationAssets))
 	for _, destinationAsset := range destinationAssets {
 		destinationAssetString := destinationAsset.String()
 		target[graph.assetStringToID[destinationAssetString]] = true
```
leighmcculloch, line 429: If this map is being used as a set where every entry has a value of true, there might also be a gain in converting the type to `map[int32]struct{}` since an empty strict consumes no memory while a book does. Not sure if the gain will be meaningful, but mentioning just in case.

leighmcculloch: Actually, for maps this might actually result in more allocations 🤔 TIL.

PR #4128, opened 2021-12-06
exp/orderbook:

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

There is a bug where if you query for path finding requests starting from or ending at a non-existent asset the path finding algorithm may return invalid paths. There is a mapping between an asset's string representation and its integer id. However, when accessing the map I did not check if the asset string actually exists. So, for non existent keys golang maps will default to the the 0 value.

For example if I search for paths from asset "DNE: GC4YD2I47LT64AFNN67QYG5VVZG32MJHNRD3DJ23FZSN33DIYHMK5QBX" to "XLM" and "DNE: GC4YD2I47LT64AFNN67QYG5VVZG32MJHNRD3DJ23FZSN33DIYHMK5QBX" does not exist, the code will actually search for paths from whichever asset has id 0 to "XLM".

This commit adds the proper check to first determine if the asset exists before doing the path finding search.

### Known limitations

[N/A]

### DIFF
--- a/exp/orderbook/graph.go
+++ b/exp/orderbook/graph.go
@@ -706,6 +706,7 @@ func (graph *OrderBookGraph) FindPaths(
 	}
 	destinationAssetID, ok := graph.assetStringToID[destinationAssetString]
 	if !ok || len(sourceAssetsMap) == 0 {
+		graph.lock.RUnlock()
 		return []Path{}, graph.lastLedger, nil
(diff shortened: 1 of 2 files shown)


bartekn (approved):
Great catch! LGTM!

PR #4134, opened 2021-12-08
all: use go run for go-bindata go generate commands

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### WHAT
Use go run for go-bindata rather than install go-bindata.

### WHY
There are many variations of go-bindata. The install script checks if a binary named go-bindata is present in the PATH. Because there are many variations of go-bindata a dev can have another installed that generates different files.

There's not really any need for us to install go-bindata and use it in the way we are. The go run command is now able to run a specific version of a tool without it being installed on the users PATH.

We use this pattern for goxdr generation, and the pattern works here too.

### Known limitations

N/A

### DIFF
--- a/exp/services/recoverysigner/internal/db/dbmigrate/dbmigrate.go
+++ b/exp/services/recoverysigner/internal/db/dbmigrate/dbmigrate.go
@@ -5,7 +5,7 @@ import (
 	migrate "github.com/rubenv/sql-migrate"
 )
 
-//go:generate go-bindata -nometadata -ignore .+\.(go|swp)$ -pkg dbmigrate -o dbmigrate_generated.go ./migrations
+//go:generate go run github.com/kevinburke/go-bindata/go-bindata@v3.18.0+incompatible -nometadata -ignore .+\.(go|swp)$ -pkg dbmigrate -o dbmigrate_generated.go ./migrations
 
 var migrationSource = &migrate.AssetMigrationSource{
 	Asset:    Asset,
--- a/exp/services/recoverysigner/docker/Dockerfile
+++ b/exp/services/recoverysigner/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.16 as build
+FROM golang:1.17 as build
 
 ADD . /src/recoverysigner
 WORKDIR /src/recoverysigner
--- a/exp/services/webauth/docker/Dockerfile
+++ b/exp/services/webauth/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.16 as build
+FROM golang:1.17 as build
 
 ADD . /src/webauth
 WORKDIR /src/webauth
--- a/gogenerate.sh
+++ b/gogenerate.sh
@@ -1,15 +1,6 @@
 #! /bin/bash
 set -e
 
-# Check if go-bindata is installed, if not install it.
-command -v go-bindata >/dev/null 2>&1 || (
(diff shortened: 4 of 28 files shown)


exp/services/recoverysigner/internal/db/dbmigrate/dbmigrate.go
```diff
@@ -5,7 +5,7 @@ import (
 	migrate "github.com/rubenv/sql-migrate"
 )
 
-//go:generate go-bindata -nometadata -ignore .+\.(go|swp)$ -pkg dbmigrate -o dbmigrate_generated.go ./migrations
+//go:generate go run github.com/kevinburke/go-bindata/go-bindata@v3.18.0+incompatible -nometadata -ignore .+\.(go|swp)$ -pkg dbmigrate -o dbmigrate_generated.go ./migrations
```
2opremio, line 8: Instead of specifying the version in every usage, I would use a `tools.go` file (see https://marcofranssen.nl/manage-go-tools-via-go-modules for instance)

2opremio: Much more context at https://github.com/golang/go/issues/25922

leighmcculloch: I've used that pattern before and it carries some complexity that isn't as intuitive or simple as a go run command, and I don't think that additional complexity is outsized by the benefit of shortening these comments by a few characters.

That tools.go pattern was proposed in 2018 before the go run command supported running specific versions as it does today. I think it will be interesting to see if it continues to be a norm.

In regards to the complexity, adding tools to the module expands the dependency graph of the module, even if they are in a 'tools' tag and not compiled or used at runtime. The dependency graph of our module impacts the dependency graph of any importer of our module, i.e. any importer of the Go SDK. This tool is used in development, not by the compiler, so I don't think we should introduce them into the dependency graph.

2opremio: I'll leave it up to you, but embedding the version in every invocation isn't great. I wonder if there is a way to have a different go module just for this, so that it doesn't expand the main module's graph.

leighmcculloch: I had a go of using a separate go.mod file a go.

I named it go.tools.mod initially and placed it in the root directory. Every call site had this long `go run -modfile=../../../../../../go.tools.mod`. Running `go mod tidy -modfile=go.tools.mod` added all the regular dependencies though and got rid of the dependency I needed.

I created a directory `tools-deps` and moved the mod file there, and added a tools.go, following the tools.go pattern. Every call site had this `go run -modfile=../../../../../../tools-deps/go.mod`.

It was all pretty fiddly, not intuitive to me, and much more fiddly than setting versions in a few places.

>but embedding the version in every invocation isn't great.

Assuming you're referring to the inconvenience, I agree, it's inconvenient to have the version in a few places rather than in a single place. I think it is less inconvenient to the multi-mod file tools.go setup though.

leighmcculloch: I took a look at how another repo does tools.go, and I'm curious to explore this more, but I think we should merge this as is for now as it gets benefits over what we have today, and tools.go if we add it can be an improvement on this.

2opremio (approved):

PR #4142, opened 2021-12-14
support/log: decouple logger and unit test concerns

### LINKED ISSUE #4121: intermittent CI test failures during 'go test race' condition checks on pull request(PR)
latest on master, currently based from v2.12.0

### What did you do?
submitted a PR, when the CI jobs ran to check the PR, one failed in `test_code_1_16` which executes:
```go test -race -coverprofile=coverage.txt -covermode=atomic ./...```

full output captured [test_code_1_16 (59649)](https://app.circleci.com/pipelines/github/stellar/go/10088/workflows/2fd1ab53-2eaa-43a6-9070-879525b0c637/jobs/59649)

if I run the same go test command locally, cannot reproduce the race errors, and upon retrying the failed CI job, it completed successfully several times afterwards. So, the occurrence is not easily replicated.

After some brief investigation on the race stack dumps provided, it appears as though two tests from unrelated/different test files are running at same time and colliding on a shared 'log' instance. 

test #1(middleware_test.go) was trying to finish and was attempting https://github.com/stellar/go/blob/master/support/log/entry.go#L211

test #2(actions_effects_test.go) was using a db session instance which tried to print a log Debugf statement https://github.com/stellar/go/blob/master/support/log/entry.go#L109

same passing result in CI runs as local runs for ```go test -race -coverprofile=coverage.txt -covermode=atomic ./...```

### What did you see instead?
race condition test failures on CI run. success/pass on local run.

### PR DESCRIPTION
**Pull request by sreuland**
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

investigation of intermittent failures during 'test race' condition checks on CircleCI jobs led to likely source being unit test hook/unhook state applied on default logger, if any unit tests happen to run in parallel, they would by default be using the same default logger and potentially collide on concurrent access to aspects of hooked state.

### Why

First alternative is to add mutex/thread sync around hooked state access in logger, but that seemed to add more complexity and potential brittleness to test infrastructure. Decided to take further step back first and looked at why was unit test and logger so tightly bound, since they should be unrelated. The reason is some unit tests were using the hooked logger to capture log output and then assert on the the logged output. That approach is likely opinionated/subjective, my initial reaction is that logs are code mechanics and don't represent or infer app functionality. 

The unit tests that were asserting on log outputs also were asserting on more concrete functional outputs of the system/unit, so I tried an approach here of removing the code that bound unit-test to logger which should remove the intermittent race condition, allow unit tests to be a bit more agnostic, idempotent for different runtime environments.

Closes #4121 

### Known limitations

If asserting on log outputs is a must, then will need to revert to different approach that applies mutex/thread sync into unit-test/log hooks.

### CODE REVIEW

File: services/horizon/internal/test/t.go [hunk 1]
```diff
@@ -26,17 +26,7 @@ func (t *T) CoreSession() *db.Session {
 // Finish finishes the test, logging any accumulated horizon logs to the logs
 // output
 func (t *T) Finish() {
-	logEntries := RestoreLogger()
 	operationfeestats.ResetState()
```
sreuland, line 29: looks like some functional concern that leaked up to general layer, maybe refactor in another pass and remove Finish().

support/log/entry.go
```diff
@@ -168,46 +166,3 @@ func (e *Entry) Print(args ...interface{}) {
 	e.entry.Print(args...)
 }
 
-// StartTest shifts this logger into "test" mode, ensuring that log lines will
-// be recorded (rather than outputted).  The returned function concludes the
-// test, switches the logger back into normal mode and returns a slice of all
-// raw logrus entries that were created during the test.
-func (e *Entry) StartTest(level logrus.Level) func() []logrus.Entry {
```
sreuland, line 175: this was crux of change, removing any unit test coupling with logger.

exp/services/recoverysigner/cmd/db_test.go
```diff
@@ -23,8 +22,6 @@ func TestDBCommand_Migrate_upDownAll(t *testing.T) {
 
 	// Migrate Up
 	{
-		logsGet := log.StartTest(logrus.InfoLevel)
```
leighmcculloch, line 26: @sreuland I don't think recoverysigner's use of the logger like this is a problem. If you look a few lines up at line 16 you'll see that the `log` here is created with `log.New()` and is therefore unique to this test. The log value is not shared with any other test, and therefore shouldn't be causing the problems described.

In some of our other apps, such as Horizon, that use a shared logger throughout the application and throughout the tests I see what you're saying, there would be issues, but that has more to do with those tests not injecting a unique logger per test, isn't it?

leighmcculloch: I think you can make these changes to Horizon, removing the use of the testing functionality, and you could remove the package level `StartTest` function, but you could keep the struct `StartTest` method, and leave the recoverysigner code in tact. This would remove the problems from Horizon, but would allow applications that are using a non-default non-global logger to continue to test log output.

exp/services/recoverysigner/cmd/db_test.go
```diff
@@ -43,23 +40,10 @@ func TestDBCommand_Migrate_upDownAll(t *testing.T) {
 			"20200320000002-create-auth-methods-audit.sql",
 		}
 		assert.Equal(t, wantIDs, ids)
```
sreuland, line 42: @leighmcculloch ,thanks for your feedback, wanted to get your input, is this assert on L45 providing same functional verification compared to the implied aspect of verification from log message asserts in L46-56 ?

leighmcculloch: It is, some of the tests in this file have assertions on non-logs. However, there are others that do not. These tests are testing user driven commands and their outputs that are intended to be displayed to the user, which is why they test the log output.

support/log/entry.go
leighmcculloch: Could we keep this function and just remove the package level version of the function? So that apps that use a unique logger for each test can keep their existing test functionality. I think the scope of this PR can be narrowed a little.

support/http/logging_middleware_test.go
```diff
@@ -28,7 +27,6 @@ func setContentMD5Middleware(next stdhttp.Handler) stdhttp.Handler {
 }
 
 func TestHTTPMiddleware(t *testing.T) {
-	done := log.DefaultLogger.StartTest(log.InfoLevel)
```
leighmcculloch, line 31: I think if you replace `log.DefaultLogger` with `log.New()`, and before calling `mux.Use.LoggingMiddleware` add a call of `mux.Use(SetLoggerMiddleware(...))`, this test can use an isolated logger and it won't experience the problems you're seeing, and you won't need a mutex.

i.e.:
```go
log := log.New() // 👈 Create a unique logger that won't be shared with other tests
done := log.StartTest(log.InfoLevel)

mux := chi.NewMux()

mux.Use(middleware.RequestID)
mux.Use(SetLoggerMiddleware(log)) // 👈 Specify the logger to use for requests that'll be used instead of the default
mux.Use(LoggingMiddleware)

// ...
```

exp/services/recoverysigner/cmd/db_test.go
sreuland: yes, I'll revert db_test.go and entry.StartTest, but this seems like logging was conflated into application semantics, such as the command is the UX, and would be tested from that approach, i.e. it accepts the command as input and creates the expected report response as output to user, how that report is rendered, i.e. to console, log, o/s file, etc is separate layer.

support/http/logging_middleware_test.go
sreuland: yes, I think will adjust PR to smaller footprint and take that approach, I saw the potential to remove the route and complexity of coupling unit test and logger for assertions and wanted to air that out in PR for this feedback, thanks!.

sreuland: the test appears to be asserting server url routing middleware and expected HTTP response codes, the logging assertions are inferring intermediate internal details, ultimately they equate to same functional assertion, was kinda what I'm wondering if worthwhile to carry the extra code, but bug fix here may not be place for that.

leighmcculloch: I don't think they are testing the same things.

The goal of the logging middleware, which is what is being tested here, is to log the requests.

There are two things I see being tested here:
1. The assertions on HTTP response codes is to ensure that the logging middleware has no side-effects on the response.
2. The assertions on the log output is testing the primary functionality provided by the logging middleware, logging the requests.

exp/services/recoverysigner/cmd/db_test.go
ire-and-curses: Interesting discussion. I do agree that in general testing log output is a bit odd. However I think it would be good to keep this PR focused on the race, which derives from Horizon's global logger issue, so I think I generally agree with Leigh here. I don't see too much benefit in being opinionated about the unit tests of the recoverysigner, if they aren't actually causing any problems.

support/http/logging_middleware_test.go
sreuland: @leighmcculloch , I went with using a new logger instance in the horizon test setup rather than letting it use default as you pointed out earlier and mostly reverted other aspects back, I think this should line up now, let me know if anything still concerns, thanks for your time/insight on review!

leighmcculloch: I realize one side-effect of changing this test so it no longer tests with the default logger is that there is no longer any test that exercises the functionality where `LoggingMiddleware` is used without `SetLoggerMiddleware`. I'm not sure if that is important to you. In newer apps we use SetLoggerMiddleware. If this test isn't causing problems I'd suggest leaving it as it was. If this test was causing problems, I'm fine with it, and just acknowledging we are losing a little test coverage.

leighmcculloch (approved):
One comment, see below, non-blocking.

support/http/logging_middleware_test.go
sreuland: sure, reverted, this test wasn't showing up directly in the Circle CI race condition checks, two other unit tests under services/horizon that both used default logger instance with `Log.Effect.StartTest()` were triggering 'go' race check related to the default logger during CI execution which runs tests for the whole services/horizon package space, this test is doing the same pattern with `Log.Effect.StartTest()`, it could be susceptible to triggering similar if other tests in same test run level happen to follow same convention.

services/horizon/internal/test/main.go
```diff
@@ -112,4 +87,27 @@ func StellarCoreDatabaseURL() string {
 	return tdb.StellarCoreURL()
 }
 
+// RestoreLogger restores the default horizon logger after it is overridden
+// using a call to `OverrideLogger`.  Panics if the default logger is not
+// presently overridden.
+func restoreLogger() []logrus.Entry {
+	if endLogTest == nil {
```
leighmcculloch, line 94: `endLogTest` looks like a package level variable, will there be racees on it when tests run in parallel?

sreuland: ah, thanks for pointing out, yes, moved func into the T 'per-test' session state struct.

services/horizon/internal/test/main.go
```diff
@@ -112,4 +88,11 @@ func StellarCoreDatabaseURL() string {
 	return tdb.StellarCoreURL()
 }
 
-var endLogTest func() []logrus.Entry = nil
+// retrieves entries from test logger instance
+func testLogs(t *T) []logrus.Entry {
```
leighmcculloch, line 92: 💡 Suggest making the function a function of T since its first parameter is always T anyway. It might be more intuitive to group helper functions for the type on the type. Although this makes no functional difference.
```suggestion
func (t *T) testLogs() []logrus.Entry {
```

leighmcculloch (approved):
One non-blocking suggestion, regardless, looks great to me 👏🏻. I'm no Horizon expert, so I defer to @ire-and-curses who had already commented.

tamirms (approved):

PR #4143, opened 2021-12-14
all: run builds and tests with go1.18rc1

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Run builds and tests with Go 1.18rc1.

### Why

See what breaks, make sure we're in a good place to switch when it releases in a couple months.

### Known limitations

N/A


2opremio:
Something odd is up, because we explicitly disable composite checks in `govet.sh` ( `go vet -all -composites=false ...`)

2opremio:
Could it be that now `go vet -vettool` also does the other go vet checks? (i.e. we need to run `go vet -composites=false -unreachable=false -tests=false -vettool=$(which shadow) ./...` instead of `go vet -vettool=$(which shadow) ./...`?

Another possibility is that `-vettool` is broken in the beta release.

leighmcculloch:
@2opremio This is very strange, I thought -vettool might be broken, I saw it now identify shadowed variable but identify composites locally, so I opened https://github.com/golang/go/issues/50195. But I have a new shell open and trying to reproduce it again while debugging and it isn't reproducing and it is working fine locally for me.

leighmcculloch:
I believe the issue was that `go get` no longer installs binaries in Go 1.18. Our `govet.sh` script was using `go get` to install shadow, then running `go vet -vettool=$(which shadow)`. Because shadow wasn't installed the result of `which shadow` would have been an empty string and it was like we were running `go vet` with no parameters. I've pushed 3674d36 that changes how we run the shadow binary to use `go run` so we are not fiddling around with installing.

staticcheck.sh
```diff
@@ -10,6 +16,10 @@ ls -d */ \
...
-
-# Check horizon for unused exported symbols (relying on the fact that it should be self-contained)
-$staticcheck -unused.whole-program -checks='U*' ./services/horizon/...
+# Whole program unused checks were removed from staticcheck in newer versions,
+# so this check is being sunset and will be removed once Go 1.18 is released and
+# a proper release of staticcheck is released that supports it.
+if [ "$version" = "2020.1.4" ]; then
+  # Check horizon for unused exported symbols (relying on the fact that it should be self-contained)
+  $staticcheck -unused.whole-program -checks='U*' ./services/horizon/...
+fi
```
leighmcculloch, line 25: FYI @stellar/horizon-committers. We'll be losing this check in the very near future because it was removed from staticcheck sometime ago. When Go 1.18 is released we'll upgrade our version of staticcheck and that version won't have the whole program unused check anymore. See https://github.com/dominikh/go-tools/commit/5cfc85b70e7b778eb76fd7338e538d7c9af21e4e.

.github/workflows/go.yml
```diff
@@ -11,14 +11,15 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        go: [1.17]
+        go: [1.17, 1.18.0-rc1]
```
acharb, line 14: out of curiosity do you know why `1.16.7` isn't included in this check job but is in the build and test jobs below?

acharb (approved):
lgtm! had 1 question, non-blocking

.github/workflows/go.yml
leighmcculloch: We support the last two versions of Go, but checks can change from one version to the next in subtle ways and there's little value in running the checks twice. What's more important is running tests and building against the previous version to be sure that we aren't using an API that isn't supported in the previous version, or to detect subtle API changes. Subtle API changes are so rare in Go that we probably will never see a error.

PR #4144, opened 2021-12-16
txnbuild: add ToXDR, Hash, HashHex to GenericTransaction

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add `ToXDR`, `Hash`, `HashHex` to `GenericTransaction`, copying the semantics of those same functions already available on `FeeBumpTransaction` and `Transaction`.

### Why

I'm writing some code that needs to get the hash of a transaction, irrespective of whether it is a fee bump or a standalone transaction. A `GenericTransaction` wraps a transaction envelope, and a transaction envelope always has a hash, so it seems appropriate to introduce this functionality here.

### Known limitations

N/A


txnbuild/transaction.go
```diff
@@ -603,6 +603,40 @@ func (t GenericTransaction) FeeBump() (*FeeBumpTransaction, bool) {
...
+	if err != nil {
+		return [32]byte{}, fmt.Errorf("hash GenericTransaction: %v", err)
+	}
+	return network.HashTransactionInEnvelope(envelope, networkStr)
+}
+
+// HashHex returns the network specific hash of this transaction
+// encoded as a hexadecimal string.
+func (t GenericTransaction) HashHex(network string) (string, error) {
+	h, err := t.Hash(network)
```
acharb, line 633: 💭  should we use the hashHex(...) helper method for consistency? since `Transaction` and `FeeBumpTransaction` both use it

acharb (commented):
makes sense to me, just had 1 💭

txnbuild/transaction.go
leighmcculloch: Good point. As I reflected on these functions I also realized that GenericTransaction is wrapping Transaction and FeeBumpTransaction who both have these functions already, therefore the hashing functions should probably just defer to calling the Hash and HashHex functions of the respective contained type, rather than reimplementing the logic or even using hashHex.

I pushed a change in d9ebf57. Wdyt?

acharb: yup, I think that's easier to understand too!

acharb (approved):
lgtm!

PR #4148, opened 2021-12-17
xdr: Add AbsBeforeEpoch to ClaimableBalance:Predicate API resource 

### LINKED ISSUE #4095: Feature Request: API - Add original integral value for ClaimableBalance:Claimant:Predicate:AbsBefore
Some SDK clients will not able to parse the custom/extended ISO 8601 date format for absBefore returned in API json responses for ClaimableBalance:Claimant:Predicate:AbsBefore per [the Horizon serverside API implementation changes done last year.](https://github.com/stellar/go/pull/3183/files) 

If the absBefore value happens to equate to a large date beyond 4 digit year range of 0000-9999, and clients try to apply their standard datetime parsing library on it, they will likely get error or invalid value from languages such as js/python/java, this is because they follow the ISO 8601 more closely, and aren't equipped to either handle the "=/-" prefix that is appended at beginning of the absBefore date string emitted by API, or they don't use large enough integer value to capture the year which can now roughly be 292277026596 into future and -292471206707 in past.

Add a new element, `AbsBeforeEpoch` to the API Resource 'ClaimableBalance:Claimant:Predicate', when a Predicate has `AbsBefore` value present, also include `AbsBeforeEpoch`, having the original integral value specified in the CreateClaimableBalance:Claimant:Predicate:AbsBefore source transaction. So, example of the updated API json model for 'ClaimableBalance:Claimant:Predicate' would resemble:

```
"claimants": [
    {
      "destination": "GBQECQVAS2FJ7DLCUXDASZAJQLWPXNTCR2DGBPKQDO3QS66TJXLRHFIK",
      "predicate": {
        "abs_before": "+39121901036-03-29T15:30:22Z",
        "abs_before_epoch": "1234567890982222222"
      }
    },
    {
      "destination": "GAKFBRS24U3PEN6XDMEX4JEV5NGBARS2ZFF5O4CF3JL464SQSD4MDCPF",
      "predicate": {
        "unconditional": true
      }
    }
  ]
```
Goal is to be additive, non-breaking change, denote `AbsBefore` as deprecated in the API docs for future removal. 

- Assess behavior of current SDKs: 
-- Go: works correctly, since it has the custom ser/deser code for absBefore
-- Java: throws exception as reported on [bug #354](https://github.com/stellar/java-stellar-sdk/issues/354) 
--  js: doesn't interpret absBefore, just pass through, if client code attempts to parse then would get issue.
-- python(StellarCN): same as js
- Update the API docs in any case to accurately inform clients about the details of the custom format of absBefore, have submitted [request #600](https://github.com/stellar/new-docs/issues/600). Clients choose how to handle/parse.
- Fix the Java SDK to handle the custom absBefore format since it's done automatically by the sdk during API json unmarshal phase, client has no control over it. Since issue already filed, have [PR #382](https://github.com/stellar/java-stellar-sdk/pull/382) to address.

sreuland:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

added a new optional field `AbsBeforeEpoch` to the API resource `ClaimableBalance:Predicate`, it is the Unix Epoch value represented by the same custom extended ISO date value in the `AbsBefore` field.

### Why

this extra info for Epoch will enable SDK's to use the epoch for absolute before value of claimable balance predicate rather than attempt to parse the ISO, which gets complicated when the date very large, beyond the ISO standard range of year 9999. Some SDK's like the java SDK have bug fix pending on this change, like [Java SDK #354](https://github.com/stellar/java-stellar-sdk/issues/354)

Closes #4095

### DIFF
--- a/xdr/json.go
+++ b/xdr/json.go
@@ -102,7 +102,7 @@ func (c claimPredicateJSON) toXDR() (ClaimPredicate, error) {
 		result.Type = ClaimPredicateTypeClaimPredicateBeforeRelativeTime
 		result.RelBefore = &relBefore
 	case c.AbsBefore != nil:
-		absBefore := Int64(c.AbsBefore.Epoch())
+		absBefore := Int64(c.AbsBefore.UTC().Unix())
 		result.Type = ClaimPredicateTypeClaimPredicateBeforeAbsoluteTime
 		result.AbsBefore = &absBefore
 	case c.Not != nil:
@@ -161,9 +161,10 @@ func (c ClaimPredicate) toJSON() (claimPredicateJSON, error) {
 		payload.Not = new(claimPredicateJSON)
 		*payload.Not, err = c.MustNotPredicate().toJSON()
 	case ClaimPredicateTypeClaimPredicateBeforeAbsoluteTime:
-		payload.AbsBefore = Newiso8601Time(int64(c.MustAbsBefore()))
+		absBeforeEpoch := int64(c.MustAbsBefore())
+		payload.AbsBefore = Newiso8601Time(absBeforeEpoch)
 		payload.AbsBeforeEpoch = new(int64)
-		*payload.AbsBeforeEpoch = payload.AbsBefore.Epoch()
+		*payload.AbsBeforeEpoch = absBeforeEpoch
 	case ClaimPredicateTypeClaimPredicateBeforeRelativeTime:
 		payload.RelBefore = new(int64)
 		*payload.RelBefore = int64(c.MustRelBefore())
--- a/xdr/json.go
+++ b/xdr/json.go
@@ -60,13 +60,22 @@ func (t *iso8601Time) UnmarshalJSON(b []byte) error {
 	return nil
 }
 
+func (t iso8601Time) Epoch() int64 {
+	return t.UTC().Unix()
+}
+
+func Newiso8601Time(epoch int64) *iso8601Time {
+	return &iso8601Time{time.Unix(epoch, 0).UTC()}
+}
+
 type claimPredicateJSON struct {
-	And           *[]claimPredicateJSON `json:"and,omitempty"`
-	Or            *[]claimPredicateJSON `json:"or,omitempty"`
-	Not           *claimPredicateJSON   `json:"not,omitempty"`
-	Unconditional bool                  `json:"unconditional,omitempty"`
-	AbsBefore     *iso8601Time          `json:"abs_before,omitempty"`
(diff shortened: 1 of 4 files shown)


xdr/json.go
```diff
@@ -152,8 +161,10 @@ func (c ClaimPredicate) toJSON() (claimPredicateJSON, error) {
 		payload.Not = new(claimPredicateJSON)
 		*payload.Not, err = c.MustNotPredicate().toJSON()
 	case ClaimPredicateTypeClaimPredicateBeforeAbsoluteTime:
-		payload.AbsBefore = new(iso8601Time)
-		*payload.AbsBefore = iso8601Time{time.Unix(int64(c.MustAbsBefore()), 0).UTC()}
+		absBeforeEpoch := int64(c.MustAbsBefore())
+		payload.AbsBefore = Newiso8601Time(absBeforeEpoch)
+		payload.AbsBeforeEpoch = new(int64)
```
tamirms, line 166: I think we can replace this line and the line below with:

```golang
payload.AbsBeforeEpoch = &absBeforeEpoch
```

sreuland: nice, I just learned more on that nuance in go,`escape',  it won't gc the local variable because it sees the scope lasts beyond the function, it'll promote to heap, thanks for mention on that! I could follow the same convention for relBefore on L170 probably as well.

xdr/json.go
```diff
@@ -60,13 +60,22 @@ func (t *iso8601Time) UnmarshalJSON(b []byte) error {
 	return nil
 }
 
+func (t iso8601Time) Epoch() int64 {
+	return t.UTC().Unix()
+}
+
+func Newiso8601Time(epoch int64) *iso8601Time {
```
tamirms, line 67: we should not export this function because `iso8601Time` is only meant to be used internally in this package. I recommend renaming the function to `newiso8601Time()`

xdr/json.go
```diff
@@ -60,13 +60,22 @@ func (t *iso8601Time) UnmarshalJSON(b []byte) error {
 	return nil
 }
 
+func (t iso8601Time) Epoch() int64 {
```
tamirms, line 63: this function is no longer used so I think it can be removed

tamirms (approved):

bartekn (approved):

eliane345 (approved):
Desculpe por transtornos causados

eliane345 (approved):
Desculpe o incômodo

PR #4149, opened 2021-12-19
services/horizon/internal/db2: Add column and index to improve trade_type filter performance

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add `trade_type` column which is an enum value that represents whether the trade occurs on the orderbook or against a liquidity pool. The `trade_type` column is populated during ingestion. This commit also adds an index on the `trade_type` column to make queries for liquidity pool trades fast.

### Why

Requests to the `/trades?trade_type=liquidity_pool` endpoint timeout because the query to fetch liquidity pool trades is too slow.
 
### Known limitations

It took 33 minutes on staging to apply the migration. Also, the size of the `htrd_by_trade_type` index is 3278 mb.

I considered removing the `trade_type` filter functionality from the trades endpoint but, within the past week, we have received horizon requests to the trades endpoint with  `trade_type` set to orderbook. So, it seems that we would be breaking clients by removing this functionality.


services/horizon/internal/db2/schema/migrations/52_add_trade_type_index.sql
```diff
@@ -0,0 +1,10 @@
+-- +migrate Up
+
+ALTER TABLE history_trades ADD trade_type smallint DEFAULT 1 CHECK(trade_type > 0);
+UPDATE history_trades SET trade_type = 2 WHERE base_liquidity_pool_id IS NOT NULL OR counter_liquidity_pool_id IS NOT NULL;
+CREATE INDEX htrd_by_trade_type ON history_trades USING BTREE(trade_type, history_operation_id, "order");
+
```
sreuland, line 6: would it be ok to drop the other liquidity pool id indexes on history_trades? not sure if they come into play on other queries, but they aren't used for deriving trade type query anymore, correct? `htrd_by_base_liquidity_pool_id` `htrd_by_counter_liquidity_pool_id`

sreuland (approved):
nice improvement.

services/horizon/internal/db2/schema/migrations/52_add_trade_type_index.sql
tamirms: there is an endpoint to fetch all trades occurring on a specific liquidity pool (e.g. https://horizon.stellar.org/liquidity_pools/001403bfa898ef65015bb585e07343811aca01e02cc2a6f126a6c842f6d5359c/trades ) . those indexes are used by that endpoint

bartekn:
I wonder if we can reuse existing fields in the index. 30m migration sounds like a long one (but I think we had similar ones in the past).

tamirms:
@bartekn I tried adding different indexes without adding a new trade type column, for example: 

```
CREATE INDEX htrd_base_liquidity_pool_id_with_order ON history_trades USING BTREE(base_liquidity_pool_id, history_operation_id, "order");
CREATE INDEX htrd_counter_liquidity_pool_id_with_order ON history_trades USING BTREE(counter_liquidity_pool_id, history_operation_id, "order");
```

But unfortunately those indexes were not effective in speeding up the query.

PR #4150, opened 2021-12-22
services/horizon: Update cursor when bumping ledger to latest

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Update cursor in Stellar-Core to latest ingested ledger when local ingestion is behind the cluster.

### Why

If there is a DB ingesting instance more than one ingesting instance: either Captive-Core or DB connected to another Stellar-Core, it's possible that the cursor will be updated very rarely. This will make the Stellar-Core DB size bloat as the old ledger will not be removed.

### Known limitations

[TODO or N/A]

### DIFF
--- a/services/horizon/internal/ingest/fsm.go
+++ b/services/horizon/internal/ingest/fsm.go
@@ -416,6 +416,14 @@ func (r resumeState) run(s *system) (transition, error) {
 		log.WithField("ingestLedger", ingestLedger).
 			WithField("lastIngestedLedger", lastIngestedLedger).
 			Info("bumping ingest ledger to next ledger after ingested ledger in db")
+
+		// Update cursor if there's more than one ingesting instance: either
+		// Captive-Core or DB ingestion connected to another Stellar-Core.
+		if err = s.updateCursor(lastIngestedLedger); err != nil {
+			// Don't return updateCursor error.
+			log.WithError(err).Warn("error updating stellar-core cursor")
+		}
+
 		return retryResume(resumeState{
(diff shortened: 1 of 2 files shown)


services/horizon/internal/ingest/fsm.go
```diff
@@ -416,6 +416,14 @@ func (r resumeState) run(s *system) (transition, error) {
 		log.WithField("ingestLedger", ingestLedger).
 			WithField("lastIngestedLedger", lastIngestedLedger).
 			Info("bumping ingest ledger to next ledger after ingested ledger in db")
+
+		// Update cursor if there's more than one ingesting instance: either
```
sreuland, line 420: `ingestLedger <= lastIngestedLedger` is a state that can only happen due to multiple ingestions instances? just wanted to understand the comment which refers to it, would a single ingestion instance ever get in here, i.e. this is a best effort attempt under any case?

sreuland (approved):
nice test coverage too!

tamirms (approved):

services/horizon/internal/ingest/fsm.go
bartekn: Correct, this can only happen if there is more than one ingesting instance. `ingestLedger` is the currently ingested ledger in this instance, `lastIngestedLedger` is the last ingested ledger in the cluster. I guess the variable names could be better.

eliane345 (approved):
Desculpe  pelo transtorno

PR #4151, opened 2021-12-22
services/horizon/../sse: Remove mutex from stream operations and count 'hello' event

### LINKED ISSUE #2284: Timeout error not properly rendered when streaming /account/{id}/transactions
### What did you do?

```
curl -v -H "Accept: text/event-stream" "https://horizon.stellar.org/accounts/GAFNBDOZBFUKHB4ZY5ANCQ25PM33YLTMWQKW25J7LXNZM3YSHRH3BELY/transactions?limit=200&cursor=119456526699307008&include_failed=true"
```

```
retry: 1000
event: open
data: "hello"

event: error
data: Your request timed out before completing.
```

### What did you see instead?

```
retry: 1000
event: open
data: "hello"

{
  "type": "https://stellar.org/horizon-errors/timeout",
  "title": "Timeout",
  "status": 504,
  "detail": "Your request timed out before completing.  Please try your request again. If you are submitting a transaction make sure you are sending exactly the same transaction (with the same sequence number)."
}
```

2opremio:
As far as I have seen, the stream object is never used in a multi-goroutine context.

Also, we were not properly counting the 'hello' event from the preamble.
Due to the error logic (which counts the sent events) this could cause
an error right after the 'hello' event to be incorrectly rendered as JSON.

I think this closes #2284 (although I couldn't reproduce as stated in the issue).


services/horizon/internal/render/sse/stream_test.go
```diff
@@ -18,20 +18,16 @@ import (
...
 // IsDone is safe to call concurrently and is exported.
 func (s *Stream) IsDone() bool {
-	s.mu.Lock()
-	defer s.mu.Unlock()
 	if s.limit == 0 {
 		return s.done
 	}
 
-	return s.done || s.sent >= s.limit
+	return s.done || s.eventsSent >= s.limit
```
sreuland, line 30: since preamble is now counted, does it possibly affect expectations here on limit comparisons

sreuland (approved):

services/horizon/internal/render/sse/stream.go
```diff
@@ -40,43 +37,37 @@ func NewStream(ctx context.Context, w http.ResponseWriter) *Stream {
 // hello message. This should be called before any method that writes to the client to ensure that the preamble
 // has been sent first.
 func (s *Stream) Init() {
-	s.initSync.Do(func() {
+	if s.eventsSent == 0 {
 		ok := WritePreamble(s.ctx, s.w)
-		if !ok {
+		if ok {
+			// The preamble includes sending a hello event
+			s.eventsSent++
```
bartekn, line 44: I think this is a breaking change. The stream will send up to `Stream.limit` events which is defined by the `limit` param of stream GET request. So with this change, the stream will close after `limit-1` objects sent.

2opremio: You mean that `Stream.limit` shouldn't take into account the `hello` event? Good catch.

2opremio: Interestingly, it seems `Stream.limit` is only used in tests, so I don't think it makes a difference.

2opremio: Is that a bug? From what you said, it seems it is.

2opremio: > So with this change, the stream will close after `limit-1` objects sent.

I have created https://github.com/stellar/go/pull/4175 but I think it's more important to fix the limit enforcement problem.

bartekn: We set it via `SetLimit` here:
https://github.com/stellar/go/blob/0582e145eabfee6a45c1b6d5368d3a0455603664/services/horizon/internal/render/sse/stream_handler.go#L35

2opremio: Yes, but we don't use it.

bartekn: I think we use it in `IsDone`, right?

bartekn: OK, you're right. We actually depend on the local `limit` variable in: https://github.com/stellar/go/blob/0582e145eabfee6a45c1b6d5368d3a0455603664/services/horizon/internal/render/sse/stream_handler.go#L35 Certainly needs a refactoring but it's not a bug.

PR #4152, opened 2021-12-22
services/horizon: Properly validate the history archive URLs

### LINKED ISSUE #4141: services/horizon: Improve error message when HISTORY_ARCHIVE_URLS is empty

Latest

### What did you do?

Start Horizon without `HISTORY_ARCHIVE_URLS` env.

Clear error message.

### What did you see instead?

```
level=fatal msg="error creating history archive: URL is empty"
```

Shaptic:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Checks that the history archive URLs contain no meaningful values.

Specifically, because of the way [`strings.Split`](https://pkg.go.dev/strings#Split) works, the returned array will *always* have at least one element. If the string is empty, the array will be `[""]`, so we need to check `urls[1] == ""` instead of checking the length.

### Why
Currently, the error is not clear; closes #4141.

### Known limitations
n/a


services/horizon/internal/flags.go
```diff
@@ -555,7 +555,7 @@ func ApplyFlags(config *Config, flags support.ConfigOptions, options ApplyOption
 
 		// config.HistoryArchiveURLs contains a single empty value when empty so using
 		// viper.GetString is easier.
-		if len(config.HistoryArchiveURLs) == 0 {
+		if len(config.HistoryArchiveURLs) == 1 && config.HistoryArchiveURLs[0] == "" {
```
sreuland, line 558: since it's not really a value comparison for empty string, doing `len(str) == 0` can be a bit more direct, but either way appears to be convention in go.

is there a unit test that hits this?

Shaptic: Yeah both length and content checks are idiomatic; I think we use `== ""` most of the time.

We've started work on testing CLI parameters (see #3193 and its references), but there's a lot more to be done there. In general no, we don't have unit tests for flag parsing afaik.

bartekn (approved):

PR #4155, opened 2022-01-01
horizon/ingest: refactor initial load of orderbook offers/pools to use functional streaming

### LINKED ISSUE #1622: Optimize Orderbook loading to use sql rows stream directly
Currently `GetAllOffers` is used to load all offers from a database into in-memory order book graph. Currently, there are around 33k offers in the public network however, if this number increases in the future this function may require more memory. Instead, we should stream offers from a DB using [`sql.Rows`](https://golang.org/pkg/database/sql/#Rows). We already use this method in several places in Horizon so check them for reference.

sreuland:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
at init time, entire orderbook was loaded into intermediate ram(slice) first from db resultset, then the slice was copied piece by piece into the orderbook memory, doubling the overall ram size needed by orderbook state for some initial amount of time.

### Why
performance improvement to load orderbook data more efficiently at startup, avoid the double ram usage, handle larger sizes as things scale up

Closes #1622 

### Known limitations

does not change final orderbook state, it is still stored in ram


services/horizon/internal/db2/history/liquidity_pools.go
```diff
@@ -85,7 +86,7 @@ func (lpar *LiquidityPoolAssetReserve) UnmarshalJSON(data []byte) error {
 type QLiquidityPools interface {
 	UpsertLiquidityPools(ctx context.Context, lps []LiquidityPool) error
 	GetLiquidityPoolsByID(ctx context.Context, poolIDs []string) ([]LiquidityPool, error)
-	GetAllLiquidityPools(ctx context.Context) ([]LiquidityPool, error)
+	StreamAllLiquidityPools(ctx context.Context, callback func(*LiquidityPool) error) error
```
tamirms, line 89: I don't see the reason for the callback to take a pointer to a LiquidityPool rather than the LiquidityPool itself. It seems all uses of `StreamAllLiquidityPools()` end up dereferencing the pointer.

sreuland: I was looking to avoid the implicit copy by-value aspect, the usage intent of the callback is read only, so there was no functional purpose for by-ref, if not meaningful for performance and seems clutter I'll remove?

tamirms: the performance improvement only occurs for large structs. I think we can allow the copy by value in this case

tamirms (approved):

services/horizon/internal/db2/history/liquidity_pools.go
```diff
@@ -186,13 +187,27 @@ func (q *Q) GetLiquidityPools(ctx context.Context, query LiquidityPoolsQuery) ([
...
+	if rows, err = q.Query(ctx, selectLiquidityPools.Where("deleted = ?", false)); err != nil {
+		return errors.Wrap(err, "could not run all liquidity pools select query")
 	}
 
-	return results, nil
+	defer rows.Close()
+	liquidityPool := LiquidityPool{}
+
+	for rows.Next() {
+		if err = rows.StructScan(&liquidityPool); err != nil {
```
bartekn, line 202: I remember one issue in our XDR decoder: an object reused for decoding was not cleared so sometimes pointer fields contained the old data. I think this _can_ be also the case here. Can we check? The fix would be to move `liquidityPool` variable inside the `for` loop.

bartekn: I found the PRs for reference: https://github.com/stellar/go-xdr/pull/13 https://github.com/stellar/go-xdr/pull/14.

sreuland: In this case, every field in the struct is set to a column value in results(selected from liquidity_pools table) per row iteration during StructScan, so, shouldn't need init between iterations.

PR #4158, opened 2022-01-03
horizon/ingest: Add method to ingest.Change that returns liquidity pool type

erika-sdf:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add method to `ingest.Change` that extracts liquidity pool type

### Why

This functionality is likely to be used elsewhere while processing liquidity pools.

### DIFF
--- a/ingest/change.go
+++ b/ingest/change.go
@@ -82,6 +82,34 @@ func (c *Change) LedgerEntryChangeType() xdr.LedgerEntryChangeType {
 	}
 }
 
+// getLiquidityPool gets the most recent state of the LiquidityPool that exists or existed.
+func (c *Change) getLiquidityPool() (*xdr.LiquidityPoolEntry, error) {
+	var pre, post *xdr.LiquidityPoolEntry
+	if c.Pre != nil {
+		pre = c.Pre.Data.LiquidityPool
+	}
+	if c.Post != nil {
+		post = c.Post.Data.LiquidityPool
+	}
+	if pre == nil && post == nil {
+		return &xdr.LiquidityPoolEntry{}, errors.New(
+			"both pre and post entries cannot be nil; this change may not include a liquidity pool")
+	}
+	if post != nil {
+		return post, nil
+	}
+	return pre, nil
+}
+
+// GetLiquidityPoolType returns the liquidity pool type.
+func (c *Change) GetLiquidityPoolType() (xdr.LiquidityPoolType, error) {
+	lp, err := c.getLiquidityPool()
+	if err != nil {
+		return xdr.LiquidityPoolType(0), err
+	}
+	return lp.Body.Type, nil
+}
+
 // AccountChangedExceptSigners returns true if account has changed WITHOUT
 // checking the signers (except master key weight!). In other words, if the only
 // change is connected to signers, this function will return false.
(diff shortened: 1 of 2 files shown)


ingest/change.go
```diff
@@ -82,6 +82,34 @@ func (c *Change) LedgerEntryChangeType() xdr.LedgerEntryChangeType {
...
+		return post, nil
+	}
+	return pre, nil
+}
+
+// GetLiquidityPoolType returns the liquidity pool type.
+func (c *Change) GetLiquidityPoolType() (xdr.LiquidityPoolType, error) {
+	lp, err := c.GetLiquidityPool()
+	if err != nil {
+		return xdr.LiquidityPoolType(0), err
```
sreuland, line 108: what is reason to return a default PoolType value if err present, would `nil` suffice since callers should ignore in that case? maybe mention the err expectation in the docs line.

ingest/change.go
```diff
@@ -82,6 +82,34 @@ func (c *Change) LedgerEntryChangeType() xdr.LedgerEntryChangeType {
 	}
 }
 
+// GetLiquidityPool gets the most recent state of the LiquidityPool that exists or existed.
+func (c *Change) GetLiquidityPool() (*xdr.LiquidityPoolEntry, error) {
```
sreuland, line 86: should this method be exported, don't see it used externally yet, but maybe that is anticipated later.

ingest/change.go
erika-sdf: The return type is not a pointer, so nil is not a valid return value.

ingest/change.go
erika-sdf: Yeah, I imagine it might be used externally. But since it's not being used at the moment, I will set it to unexported for now.

ingest/change.go
sreuland: oopsie, thanks!

sreuland (approved):

ingest/change.go
```diff
@@ -82,6 +82,34 @@ func (c *Change) LedgerEntryChangeType() xdr.LedgerEntryChangeType {
...
+		post = c.Post.Data.LiquidityPool
+	}
+	if pre == nil && post == nil {
+		return &xdr.LiquidityPoolEntry{}, errors.New(
+			"both pre and post entries cannot be nil; this change may not include a liquidity pool")
+	}
+	if post != nil {
+		return post, nil
+	}
+	return pre, nil
```
Shaptic, line 101: If post is present, we always return it over pre, right? We don't need to track pre and post separately, then. Something like this would simplify the conditional logic:
```suggestion
	var entry *xdr.LiquidityPoolEntry
	if c.Pre != nil {
		entry = c.Pre.Data.LiquidityPool
	}
	if c.Post != nil {
		entry = c.Post.Data.LiquidityPool
	}
	if entry == nil {
		return &xdr.LiquidityPoolEntry{}, errors.New("this change does not include a liquidity pool")
	}
	return entry, nil
```

Shaptic (approved):
Nice abstraction, small suggestion below :arrow_down:

ingest/change.go
erika-sdf: I like it, it's cleaner, and tests still pass.

PR #4159, opened 2022-01-04
services/horizon/internal/integration: Refactor integration tests

### LINKED ISSUE #3149: Remove Sanity test in integration tests
I think you're right. In general, when a new protocol version appears for testing we will want to run all existing integration tests to ensure there're no changes in behaviour. We will also write new tests testing new features. So it seems we always want to upgrade to the latest protocol version available in Core.

To sum it up:
* Maybe we should remove sanity test here? I added it when building integration tests engine but it seems it's not needed in the long term.
* Instead of picking the protocol version directly in `integration.Config` maybe we should just upgrade to the latest available version?

Going to merge the PR to fix the tests but I'll create a new issue to discuss this. Thanks for the comment!

_Originally posted by @bartekn in https://github.com/stellar/go/pull/3146#discussion_r509637383_

2opremio:
* Use the latest protocol version for all tests.
* Set the protocol to the latest version if not specified explicitly.
* Rename tests according to their content, stop using protocol numbers.
* Remove all the sanity tests.

Closes  #3149

### DIFF
--- a/services/horizon/internal/integration/claimable_balance_ops_test.go
+++ b/services/horizon/internal/integration/claimable_balance_ops_test.go
@@ -15,7 +15,7 @@ import (
 
 func TestClaimableBalanceCreationOperationsAndEffects(t *testing.T) {
 	tt := assert.New(t)
-	itest := integration.NewTest(t, protocol15Config)
+	itest := integration.NewTest(t, integration.Config{})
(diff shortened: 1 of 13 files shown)


Shaptic (approved):
Nice cleanup :+1:

PR #4161, opened 2022-01-04
services/horizon: Add more verbose logging for horizon start up

erika-sdf:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add more verbose logging to horizon start up. Include lines that signal db migration and captive core set up, which should be logged to stdout since it happens before logging is set up.

### Why

It would make debugging easier.

### DIFF
--- a/services/horizon/internal/flags.go
+++ b/services/horizon/internal/flags.go
@@ -509,6 +509,8 @@ func NewAppFromFlags(config *Config, flags support.ConfigOptions) (*App, error) 
 	if config.Ingest && !config.EnableCaptiveCoreIngestion && config.StellarCoreDatabaseURL == "" {
 		return nil, fmt.Errorf("flag --%s cannot be empty", StellarCoreDBURLFlagName)
 	}
+
+	log.Infof("Initializing horizon...")
 	app, err := NewApp(*config)
 	if err != nil {
 		return nil, fmt.Errorf("cannot initialize app: %s", err)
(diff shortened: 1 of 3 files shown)


jacekn (approved):
Thank you, it's a very nice improvement. It looks good to me but it would be good to get 2nd review from somebody with more go experience.

services/horizon/internal/flags.go
```diff
@@ -509,6 +509,8 @@ func NewAppFromFlags(config *Config, flags support.ConfigOptions) (*App, error)
 	if config.Ingest && !config.EnableCaptiveCoreIngestion && config.StellarCoreDatabaseURL == "" {
 		return nil, fmt.Errorf("flag --%s cannot be empty", StellarCoreDBURLFlagName)
 	}
+
+	log.Infof("Initializing horizon...")
```
sreuland, line 513: does this output print on console, wasn't sure on significance of using this or stdLog, the desired outcome is to be on console correct, assume they both output there, was just asking for my own edification on logging preferences.

sreuland (approved):

services/horizon/internal/flags.go
erika-sdf: Both `log` and `stdLog` calls print to console. The code that runs prior to the logger initialization uses the `stdLog` calls. I think  @jacekn 's feature request was specifically about seeing failures prior to the initialization, so the `stdLog` statements provide some context on what's happening.

services/horizon/internal/init.go
```diff
@@ -18,6 +18,7 @@ import (
 )
 
 func mustNewDBSession(subservice db.Subservice, databaseURL string, maxIdle, maxOpen int, registry *prometheus.Registry) db.SessionInterface {
+	log.Infof("Establishing database session at %s for %v", databaseURL, subservice)
```
bartekn, line 21: Sorry for a late review. I was OOO. Not sure if we should print `databaseURL` as it contains a sensitive info (password).

PR #4165, opened 2022-01-06
all: run shadow and staticcheck using go run

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Change how shadow and staticcheck run by using go run instead of installing them onto the path.

### Why

Same reasons as #4134.

The install script checks if a binaries named staticcheck and shadow are present in the PATH. Someone may have different versions installed and the script won't sure the checks run with the same versions that run on CI.

There's not really any need for us to install these binaries and use them in the way we are. The go run command is now able to run a specific version of a tool without it being installed on the users PATH.

We use this pattern for goxdr, and since #4134 also for go-bindata, and the pattern works here too.

Also, in Go 1.18 the go get command will no longer install binaries and these scripts are broken, so this prepares us for the Go 1.18 release to some degree.

### Known limitations

N/A


goshadow.sh
```diff
@@ -0,0 +1,4 @@
+#! /bin/bash
+set -e
+
+go run golang.org/x/tools/go/analysis/passes/shadow/cmd/shadow@latest "$@"
```
marcelosalloum, line 4: Is shadow always be executed along with `go vet`? If so, why not move this code inside `govet.sh`?

marcelosalloum (approved):
LGTM! I've only dropped one question below ⬇️

goshadow.sh
leighmcculloch: I tried before I made this file but unfortunately go vet will execute the given command as the command and can't handle command arguments. So I can make the call `go vet -vettool="go run golang.org/x/tools/go/analysis/passes/shadow/cmd/shadow@latest" ./...` because `go vet` will try and find a command `go run golang.org/x/tools/go/analysis/passes/shadow/cmd/shadow@latest` rather than run the `go` command with those arguments.

PR #4174, opened 2022-01-11
txnbuild: add AddSignatureDecorated to FeeBumpTransaction

leighmcculloch:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add `AddSignatureDecorated` to `FeeBumpTransaction`.

### Why

The function exists on `Transaction` but not on `FeeBumpTransaction`. I am writing some code that needs to be able to add decorated signatures to fee bump transactions in the same way we can add them to regular transactions.

### Known limitations

N/A

### DIFF
--- a/txnbuild/transaction.go
+++ b/txnbuild/transaction.go
@@ -519,6 +519,17 @@ func (t *FeeBumpTransaction) SignHashX(preimage []byte) (*FeeBumpTransaction, er
 	return t.clone(extendedSignatures), nil
 }
 
+// AddSignatureDecorated returns a new FeeBumpTransaction instance which extends the current instance
+// with an additional decorated signature(s).
+func (t *FeeBumpTransaction) AddSignatureDecorated(signature ...xdr.DecoratedSignature) (*FeeBumpTransaction, error) {
+	extendedSignatures, err := concatSignatureDecorated(t.envelope, t.Signatures(), signature)
+	if err != nil {
+		return nil, err
+	}
+
+	return t.clone(extendedSignatures), nil
+}
+
 // AddSignatureBase64 returns a new FeeBumpTransaction instance which extends the current instance
 // with an additional signature derived from the given base64-encoded signature.
 func (t *FeeBumpTransaction) AddSignatureBase64(network, publicKey, signature string) (*FeeBumpTransaction, error) {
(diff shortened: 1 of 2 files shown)


txnbuild/transaction.go
```diff
@@ -519,6 +519,17 @@ func (t *FeeBumpTransaction) SignHashX(preimage []byte) (*FeeBumpTransaction, er
 	return t.clone(extendedSignatures), nil
 }
 
+// AddSignatureDecorated returns a new FeeBumpTransaction instance which extends the current instance
+// with an additional decorated signature(s).
+func (t *FeeBumpTransaction) AddSignatureDecorated(signature ...xdr.DecoratedSignature) (*FeeBumpTransaction, error) {
+	extendedSignatures, err := concatSignatureDecorated(t.envelope, t.Signatures(), signature)
+	if err != nil {
+		return nil, err
```
nikhilsaraf, line 527: generally prefer wrapping errors so the context isn't lost

txnbuild/transaction_test.go
```diff
@@ -1638,6 +1638,51 @@ func TestAddSignatureDecorated(t *testing.T) {
 	}
 }
 
+func TestFeeBumpTransaction_AddSignatureDecorated(t *testing.T) {
+	kp0 := newKeypair0()
+	kp1 := newKeypair1()
+
+	tx, err := NewTransaction(TransactionParams{
```
nikhilsaraf, line 1645: this new transaction creation could go into a helper function since it seems to be used as the standard transaction across multiple tests

nikhilsaraf (approved):
LGTM! 🎉

txnbuild/transaction_test.go
leighmcculloch: It could, but then the tests would become more coupled together. I've found it is easier to maintain tests in the long run if they are mostly independent of each other. Helper functions can be helpful, but they will also narrow the inputs to tests, or balloon into big configurable things of their own. Also this way they act as an example of the public API as much as possible.

txnbuild/transaction.go
leighmcculloch: I often do that too, but the errors returned by there is a series of add signature functions of which this is just one variation and this is how they all work, with the errors being well defined within the concat functions.

PR #4178, opened 2022-01-12
services/horizon: Exclude trades with >10% rounding slippage from trade aggregations

### LINKED ISSUE #4160: services/horizon: Filter out "small" liquidity pool trades from /trade_aggregations

https://github.com/stellar/stellar-core/issues/3301

Small transactions on liquidity pools, can cause the price reported by /trade_aggregations to swing wildly (as in above).

DEX UIs use the output of /trade_aggregations to draw graphs, so the graphs can be misleading.

There are two proposed solutions:

a. Filter out "dust" liquidity pool trades when calculating trade aggregations.
  - Possibly using formula: `|unrounded - rounded| / (unrounded + rounded) < epsilon`, but how do we calculate the unrounded? Is that data stored on the trades/transactions?
    - @paulbellamy: afaict, we would have to add it, and reingest.
  - If we offer both filtered and unfiltered data, that doubles the db size of trade aggregations table (currently ~60gb)
  - This option seems more feasible.
 
b. Add "supply ratio" metrics to /trade_aggregations.
  - We would need to track it at each point in time, though. We could track it incrementally, but then a historical reingest could impact all following trade aggregations.
  - We would need an equivalent metric for orderbook trades as well.
    - @jonjove: Possibly, the best thing to do would be to consider the resulting bid or ask depending on the direction of the trade.
  - Adding supply ratio stats (open/close/high/low) should only add around 1.7gb to the db, so it’s better there, but might be more intensive to calculate.

paulbellamy:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Track rounding slippage on trades, and exclude any >10% from trade aggregations.

### Why

Fixes #4160

### Todo

- [x] Make the filter configurable
- [x] Check it won't affect any other trades
- [x] Check if we need an index on the new column


ingest/ledger_transaction.go
```diff
@@ -83,6 +83,16 @@ func (t *LedgerTransaction) GetChanges() ([]Change, error) {
 	return changes, nil
 }
 
+// GetOperation returns an operation by index.
+// TODO: Clean this up
+func (t *LedgerTransaction) GetOperation(index uint32) (xdr.Operation, bool, error) {
+	ops := t.Envelope.Operations()
+	if len(ops) == 0 || int(index) >= len(ops) {
+		return xdr.Operation{}, false, nil
+	}
```
2opremio, line 92: Shouldn't we let it panic?

services/horizon/internal/flags.go
```diff
@@ -491,6 +491,14 @@ func Flags() (*Config, support.ConfigOptions) {
 			Required:    false,
 			Usage:       "determines if Horizon instance is behind AWS load balances like ELB or ALB, in such case client IP in the logs will be replaced with the last IP in X-Forwarded-For header (cannot be used with --behind-cloudflare)",
 		},
+		&support.ConfigOption{
+			Name:        "rounding-slippage-filter",
+			ConfigKey:   &config.RoundingSlippageFilter,
+			OptType:     types.Float64,
+			FlagDefault: 10.0,
+			Required:    false,
+			Usage:       "excludes trades from /trade_aggregations if their rounding slippage is >x%",
```
2opremio, line 500: Is slippage a known term in finance/crypto? Otherwise I think we should clarify it further.

services/horizon/internal/ingest/processors/trades_processor.go
```diff
@@ -175,6 +178,75 @@ func (p *TradeProcessor) findOperationChange(tx ingest.LedgerTransaction, opidx
...
+		unrounded := new(big.Rat)
+		unrounded.Mul(X, y)
+		unrounded.Quo(unrounded, new(big.Rat).Sub(Y, y))
+		unrounded.Quo(unrounded, new(big.Rat).Sub(big.NewRat(1, 1), F))
+
+		rounded := x
+		value := new(big.Rat)
+		value.Sub(unrounded, rounded)
+		value.Abs(value)
+		value.Quo(value, rounded)
```
sydneynotthecity, line 240: Small nit, why did you choose `difference / rounded` instead of `difference / unrounded`? Are we treating the rounded value as the original number?

services/horizon/internal/flags.go
paulbellamy: It is, yes. Particularly w.r.t. AMMs. But it usually only affects large trades, or low-liquidity pools. The "rounding slippage" (affecting small trades) is a bit of a special case in stellar's AMM implementation (afaik).

services/horizon/internal/ingest/processors/trades_processor.go
```diff
@@ -175,6 +178,75 @@ func (p *TradeProcessor) findOperationChange(tx ingest.LedgerTransaction, opidx
 	return ingest.Change{}, errors.Errorf("could not find operation for key %v", key)
 }
 
+var zero db.NullRat
```
2opremio, line 181: Can't this be a const?

2opremio:
I wonder whether it would be feasible to compute the slippage on the fly instead of storing it in the DB.

services/horizon/internal/ingest/processors/trades_processor.go
paulbellamy: No const structs in go.

paulbellamy:
The issue with on-the-fly, is that you need the ledger meta, as it contains the pool reserves.

services/horizon/internal/db2/schema/migrations/53_add_trades_rounding_slippage.sql
```diff
@@ -0,0 +1,7 @@
+-- +migrate Up
+
+-- Add is_dust to trades table
+ALTER TABLE history_trades ADD rounding_slippage numeric(1000, 7);
```
tamirms, line 4: rounding slippage is supposed to be a float in between 0 and 100, right?

`numeric(1000, 7)` means the column can hold a float with 1000 significant digits with up to 7 digits on the right side of the decimal point.  that doesn't seem to align with a value that can be at most 100

paulbellamy: It can be > 100. 
```
requested = 0.02
rounded = 1
slippage = abs(requested - rounded) / requested = 49
```
So in this case, the rounding caused a 49x deviation from requested.
I am unsure if framing it the other way (deviation from rounded) would be more useful.

paulbellamy: tbh, I could have just left it as `numeric`, which defaults to 1000 significant digits, but I given we're putting a a rational in it, which could have repeating digits, I wanted to cap the precision.

services/horizon/internal/db2/history/trade_aggregation.go
```diff
@@ -248,6 +248,12 @@ func (q Q) RebuildTradeAggregationTimes(ctx context.Context, from, to strtime.Mi
 		"counter_amount",
 		"ARRAY[price_n, price_d] as price",
 	).From("history_trades").Where(
+		sq.Or{
+			// db rounding is stored as a multiplier. so 95% = 0.95
+			sq.Lt{"rounding_slippage": roundingSlippageFilter / 100.0},
```
tamirms, line 253: you could do get rid of the or condition with `COALESCE`.  `COALESCE(rounding_slippage, 0) < roundingSlippage`. I'm not sure if it's any more efficient though

services/horizon/internal/ingest/processors/trades_processor.go
```diff
@@ -175,6 +178,75 @@ func (p *TradeProcessor) findOperationChange(tx ingest.LedgerTransaction, opidx
...
+		X, Y = Y, X
+	}
+	x := big.NewRat(int64(trade.AmountSold()), 1)
+	y := big.NewRat(int64(trade.AmountBought()), 1)
+	F := big.NewRat(int64(pre.Params.Fee), 10000)
+	switch op.Body.Type {
+	case xdr.OperationTypePathPaymentStrictReceive:
+		// User specified the receive amount
+
+		// solve for the unrounded send amount
```
tamirms, line 228: is there anything in https://github.com/stellar/go/blob/master/exp/orderbook/pools.go that can be reused here ?

services/horizon/internal/ingest/processors/trades_processor_test.go
```diff
@@ -962,3 +966,161 @@ func TestTradeProcessor_ProcessTransaction_MuxedAccount(t *testing.T) {
...
+	s.Assert().NoError(err)
+
+	result, err := s.processor.roundingSlippage(tx, 0, trade)
+	s.Assert().NoError(err)
+	s.Assert().True(result.Valid)
+	s.Assert().Equal("42.1908930", result.Rat.FloatString(7))
+}
+
+// TODO: This is awful.
+func createTransactionForTrade(trade xdr.ClaimAtom, reserveA, reserveB int64) (ingest.LedgerTransaction, error) {
```
sreuland, line 1036: interesting, first thought was this might be nice wrapped up as Builder in main code for reuse, something like:
`LedgerTransaction.NewBuilder().withReserveA(3740000000).withReserveB(162020000000)`

 rather than buried in a test, this was the first instance of constructing a LedgerTransaction programatically?

ingest/ledger_transaction.go
```diff
@@ -83,6 +83,16 @@ func (t *LedgerTransaction) GetChanges() ([]Change, error) {
 	return changes, nil
 }
 
+// GetOperation returns an operation by index.
+// TODO: Clean this up
+func (t *LedgerTransaction) GetOperation(index uint32) (xdr.Operation, bool, error) {
+	ops := t.Envelope.Operations()
+	if len(ops) == 0 || int(index) >= len(ops) {
```
Shaptic, line 90: Doesn't the latter check encompass the former? That is, if there are no ops then any index would be invalid.

services/horizon/internal/ingest/processors/trades_processor.go
```diff
@@ -175,6 +178,76 @@ func (p *TradeProcessor) findOperationChange(tx ingest.LedgerTransaction, opidx
 	return ingest.Change{}, errors.Errorf("could not find operation for key %v", key)
 }
 
+var zero db.NullRat
```
Shaptic, line 181: am confused why this isn't part of `roundingSlippage`

services/horizon/internal/ingest/processors/trades_processor.go
```diff
@@ -175,6 +178,76 @@ func (p *TradeProcessor) findOperationChange(tx ingest.LedgerTransaction, opidx
...
+		// User specified the receive amount
+
+		// solve for the unrounded send amount
+		// unrounded := (X * y) / (Y - y) / (1 - F)
+		// TODO: Do this with fewer allocations
+		unrounded := new(big.Rat)
+		unrounded.Mul(X, y)
+		// TODO: Fix divide-by-0 if you ware buying all the reserves in the pool.
+		unrounded.Quo(unrounded, new(big.Rat).Sub(Y, y))
+		unrounded.Quo(unrounded, new(big.Rat).Sub(big.NewRat(1, 1), F))
```
Shaptic, line 235: Just echoing Tamir above: you should use the `uint256` module for perf at the very least, but ideally we could leverage the existing vetted code. It could probably be modified pretty easily to return an unrounded version alongside the rounded, since the rounding step is separate:

https://github.com/stellar/go/blob/73b2df17f33dbf6ebcbfeeb3609021a2a692bfdd/exp/orderbook/pools.go#L154-L157

shouldn't be a perf hit

services/horizon/internal/ingest/processors/trades_processor_test.go
```diff
@@ -962,3 +966,161 @@ func TestTradeProcessor_ProcessTransaction_MuxedAccount(t *testing.T) {
...
+	}
+}
+
+func TestTradeProcessor_RoundingSlippage_Big(t *testing.T) {
+	s := &TradeProcessorTestSuiteLedger{}
+	s.SetT(t)
+	s.SetupTest()
+	s.mockReadTradeTransactions(s.processor.ledger)
+
+	assetSold := xdr.MustNewCreditAsset("GRE", s.unmuxedSourceAccount.Address())
```
Shaptic, line 1012: high priority: change to `GET` to match latter `MAD` asset

services/horizon/internal/ingest/processors/trades_processor_test.go
```diff
@@ -962,3 +966,161 @@ func TestTradeProcessor_ProcessTransaction_MuxedAccount(t *testing.T) {
...
+	tx, err := createTransactionForTrade(trade, 3740000000, 162020000000)
+	s.Assert().NoError(err)
+
+	result, err := s.processor.roundingSlippage(tx, 0, trade)
+	s.Assert().NoError(err)
+	s.Assert().True(result.Valid)
+	s.Assert().Equal("42.1908930", result.Rat.FloatString(7))
+}
+
+// TODO: This is awful.
```
Shaptic, line 1035: perhaps something like https://github.com/stellar/go/blob/73b2df17f33dbf6ebcbfeeb3609021a2a692bfdd/exp/orderbook/graph_test.go#L2328 would help here?

services/horizon/internal/ingest/processors/trades_processor_test.go
paulbellamy: not `HAT`?

services/horizon/internal/ingest/processors/trades_processor.go
paulbellamy: idea was to save an allocation, but maybe meh.

ingest/ledger_transaction.go
paulbellamy: No, they could both be 0.

services/horizon/internal/ingest/processors/trades_processor_test.go
paulbellamy: Agree that a builder (or something) is  probably the way to go here. But, choosing to punt on that for now.

services/horizon/internal/ingest/processors/trades_processor.go
paulbellamy: This was a super good suggestion. I hadn't seen the `exp/orderbook` package.

bartekn:
The code tested in stg (commit 8f924e1781dfc0bccc7e8c7e1287944008db37f9 which isn't part of this PR but letting you know anyway) was crashing on ledger 39270855 with:
```
panic: division by zero
goroutine 121 [running]:
math/big.(*Rat).Quo(0xc0004eaf28, 0xc0004eb0a8, 0xc0004eafe8)
        math/big/rat.go:535 +0x136
github.com/stellar/go/services/horizon/internal/ingest/processors.(*TradeProcessor).roundingSlippage(0xbf2f077a6b1a7e7, {0x7a,>
        github.com/stellar/go/services/horizon/internal/ingest/processors/trades_processor.go:233 +0x7dd
github.com/stellar/go/services/horizon/internal/ingest/processors.(*TradeProcessor).extractTrades(_, {{0xca, 0x5f, 0x24, 0xc8,>
        github.com/stellar/go/services/horizon/internal/ingest/processors/trades_processor.go:375 +0x188d
github.com/stellar/go/services/horizon/internal/ingest/processors.(*TradeProcessor).ProcessTransaction(0xc012db65a0, {0x12a75b>
        github.com/stellar/go/services/horizon/internal/ingest/processors/trades_processor.go:53 +0x158
github.com/stellar/go/services/horizon/internal/ingest.groupTransactionProcessors.ProcessTransaction({{0xc0043c3950, 0x6d6d030>
        github.com/stellar/go/services/horizon/internal/ingest/group_processors.go:67 +0x133
github.com/stellar/go/services/horizon/internal/ingest/processors.StreamLedgerTransactions({0x12a75b0, 0xc000648640}, {0x128e5>
        github.com/stellar/go/services/horizon/internal/ingest/processors/change_processors.go:32 +0x245
github.com/stellar/go/services/horizon/internal/ingest.(*ProcessorRunner).RunTransactionProcessorsOnLedger(_, {_, _})
        github.com/stellar/go/services/horizon/internal/ingest/processor_runner.go:296 +0x328
github.com/stellar/go/services/horizon/internal/ingest.(*ProcessorRunner).RunAllProcessorsOnLedger(_, {_, _})
        github.com/stellar/go/services/horizon/internal/ingest/processor_runner.go:335 +0x406
github.com/stellar/go/services/horizon/internal/ingest.resumeState.run({0xce0d8f}, 0xc000032480)
        github.com/stellar/go/services/horizon/internal/ingest/fsm.go:471 +0x13fe
github.com/stellar/go/services/horizon/internal/ingest.(*system).runStateMachine(0xc000032480, {0x129c730, 0x19adaa0})
        github.com/stellar/go/services/horizon/internal/ingest/main.go:580 +0x364
github.com/stellar/go/services/horizon/internal/ingest.(*system).Run(0x0)
        github.com/stellar/go/services/horizon/internal/ingest/main.go:478 +0x45
github.com/stellar/go/services/horizon/internal.(*App).Serve.func1()
        github.com/stellar/go/services/horizon/internal/app.go:102 +0x35
created by github.com/stellar/go/services/horizon/internal.(*App).Serve
        github.com/stellar/go/services/horizon/internal/app.go:101 +0x291
```

ingest/ledger_transaction.go
Shaptic: Right, and you'd enter this conditional due to both the first clause (since len == 0) _and_ the second clause (since 0 >= 0); I think that makes the first clause redundant :thinking:

exp/orderbook/pools.go
```diff
@@ -84,65 +84,83 @@ func makeTrade(
...
+
+	if !rem.IsZero() {
+		// Recalculate with more precision
+		unrounded, rounded := new(uint256.Int), new(uint256.Int)
+		unrounded.Mul(numer, maxBips).Div(unrounded, denom)
+		rounded.Mul(result, maxBips)
+		S.Sub(unrounded, rounded)
+		S.Abs(S).Mul(S, maxBips)
+		S.Div(S, unrounded)
+	}
```
Shaptic, line 130: This might be a pretty significant performance hit; can you run the benchmarks and compare them against master? We might need to push this out to a separate function and have e.g. `CalculatePoolPayout(withSlippage = true/false)`

exp/orderbook/pools.go
```diff
@@ -84,65 +84,83 @@ func makeTrade(
...
 
 	val := xdr.Int64(result.Uint64())
-	return val, result.IsUint64() && val >= 0
+	roundingSlippageBips := xdr.Int64(S.Uint64())
+	ok := result.IsUint64() && val >= 0 && S.IsUint64() && roundingSlippageBips >= 0
```
Shaptic, line 134: Something that could simplify this check: can a value be a uint64 (that is, make `x.IsUint64() == true`) and yet be < 0? Intuition says no but testing would confirm

exp/orderbook/pools_test.go
```diff
@@ -271,17 +356,32 @@ func calculatePoolPayoutBig(reserveA, reserveB, received xdr.Int64, feeBips xdr.
...
+	if rem.Cmp(big.NewInt(0)) > 0 {
+		// Recalculate with more precision
+		unrounded, rounded := new(big.Int), new(big.Int)
+		unrounded.Mul(numer, maxBips).Div(unrounded, denom)
+		rounded.Mul(result, maxBips)
+		S.Sub(unrounded, rounded)
+		S.Abs(S).Mul(S, maxBips)
+		S.Div(S, unrounded)
+		S.Div(S, big.NewInt(100))
+	}
```
Shaptic, line 379: Same here re: possible perf hit

Shaptic (commented):
Added some minor notes, and an important one re: performance. Seeing the benchmark (exp/orderbook/graph_benchmark_test.go) would help alleviate my concern :+1:

exp/orderbook/pools.go
paulbellamy: There's a cast to `Int64` after the `IsUint64`. But we could probably just ask `IsInt64` instead, since that's what we actually care about.

Shaptic: good point, overlooked that :+1:

paulbellamy:
With orderbook dumped to ledger `39339519`...

master as of `4920b27104df0ec7631eae28b5295e220fb59be7`
```
$ go test ./exp/orderbook -bench '.*' -offers offers.dump -pools pools.dump
goos: darwin
goarch: arm64
BenchmarkVibrantPath-8                               160           7442436 ns/op         2868812 B/op         44 allocs/op
BenchmarkMultipleDestinationAssets-8                 100          10807999 ns/op         2871080 B/op         70 allocs/op
BenchmarkTestData-8                                    2         735662062 ns/op        286842320 B/op      2299 allocs/op
BenchmarkSingleLiquidityPoolExchange/deposit-8          1000000000               0.0000003 ns/op
BenchmarkSingleLiquidityPoolExchange/exchange-8         1000000000               0.0000001 ns/op
BenchmarkLiquidityPoolDeposits-8                        17120752                69.89 ns/op            0 B/op          0 allocs/op
BenchmarkLiquidityPoolExpectations-8                    71820445                16.11 ns/op            0 B/op          0 allocs/op
PASS
ok      github.com/stellar/go/exp/orderbook     32.008s
```

This branch as of `2af05facc58b59e955ae0e5fc59aa89aab190cea`
```
$ go test ./exp/orderbook -bench '.*' -offers offers.dump -pools pools.dump
goos: darwin
goarch: arm64
BenchmarkVibrantPath-8                               153           7868738 ns/op         2868812 B/op         44 allocs/op
BenchmarkMultipleDestinationAssets-8                 100          10674957 ns/op         2871083 B/op         70 allocs/op
BenchmarkTestData-8                                    2         745304916 ns/op        286842392 B/op      2299 allocs/op
BenchmarkSingleLiquidityPoolExchange/deposit-8          1000000000               0.0000004 ns/op
BenchmarkSingleLiquidityPoolExchange/exchange-8         1000000000               0.0000001 ns/op
BenchmarkLiquidityPoolDeposits-8                         7928089               152.8 ns/op             0 B/op          0 allocs/op
BenchmarkLiquidityPoolExpectations-8                    71751022                16.35 ns/op            0 B/op          0 allocs/op
PASS
ok      github.com/stellar/go/exp/orderbook     32.745s
```

So, the main difference is in BenchmarkLiquidityPoolDeposits. (Could be that the Expectations scenario isn't hitting the rounding slippage), which is ~2x slower. Not ideal. I'm not really sure how to write the math faster than it is, so maybe someone else knows, or the other option would be to only store the reserves, and calculate it as part of rebuilding trade aggregations. Which would be less efficient, but not in the hot path.

paulbellamy:
Ok, disabled the roundingSlippage when called through /paths, and tidied up some allocations. so benchmark is now the same as master.

paulbellamy:
Uncovered a bug in effect ingestion while testing this, so this is blocked on https://github.com/stellar/go/issues/4203

exp/orderbook/pools.go
```diff
@@ -84,65 +84,85 @@ func makeTrade(
...
+		unrounded, rounded := new(uint256.Int), new(uint256.Int)
+		unrounded.Mul(numer, maxBips).Div(unrounded, denom)
+		rounded.Mul(result, maxBips)
+		S.Sub(unrounded, rounded)
+		S.Abs(S).Mul(S, maxBips)
+		S.Div(S, unrounded)
+		S.Div(S, uint256.NewInt(100)) // Take off the excess 2 decimal places
+		roundingSlippageBips = xdr.Int64(S.Uint64())
+		ok = ok && S.IsUint64() && roundingSlippageBips >= 0
+	}
```
2opremio, line 134: Can you add a comment with the overall computation? It's hard to read when not using infix operators.

exp/orderbook/pools.go
```diff
@@ -84,65 +84,85 @@ func makeTrade(
...
 
-// calculatePoolPayout calculates the amount of `reserveB` disbursed from the
+// CalculatePoolPayout calculates the amount of `reserveB` disbursed from the
 // pool for a `received` amount of `reserveA` . From CAP-38:
 //
 //      y = floor[(1 - F) Yx / (X + x - Fx)]
 //
 // It returns false if the calculation overflows.
-func calculatePoolPayout(reserveA, reserveB, received xdr.Int64, feeBips xdr.Int32) (xdr.Int64, bool) {
+func CalculatePoolPayout(reserveA, reserveB, received xdr.Int64, feeBips xdr.Int32, calculateRoundingSlippage bool) (xdr.Int64, xdr.Int64, bool) {
```
2opremio, line 93: Instead of having a bool parameter, can we have two different variants of the function? It's easier to read and understand what's happening. It doesn't seem like all the uses of `CalculatePoolPayout()` consume all the returned parameters.

exp/orderbook/pools.go
```diff
@@ -152,12 +172,29 @@ func calculatePoolExpectation(
...
+			// Recalculate with more precision
+			unrounded, rounded := new(uint256.Int), new(uint256.Int)
+			unrounded.Mul(numer, maxBips).Div(unrounded, denom)
+			rounded.Mul(result, maxBips)
+			S.Sub(unrounded, rounded)
+			S.Abs(S).Mul(S, maxBips)
+			S.Div(S, unrounded)
+			S.Div(S, uint256.NewInt(100)) // Take off the excess 2 decimal places
+			roundingSlippageBips = xdr.Int64(S.Uint64())
+			ok = ok && S.IsUint64() && roundingSlippageBips >= 0
```
2opremio, line 191: Can you add a comment with the overall computation? It's hard to read when not using infix operators.

exp/orderbook/pools.go
```diff
@@ -84,65 +84,85 @@ func makeTrade(
...
 // put into a pool to get the `disbursed` amount of `reserveB`.
 //
 //      x = ceil[Xy / ((Y - y)(1 - F))]
 //
 // It returns false if the calculation overflows.
-func calculatePoolExpectation(
-	reserveA, reserveB, disbursed xdr.Int64, feeBips xdr.Int32,
-) (xdr.Int64, bool) {
+func CalculatePoolExpectation(
+	reserveA, reserveB, disbursed xdr.Int64, feeBips xdr.Int32, calculateRoundingSlippage bool,
```
2opremio, line 148: Can we have two variants of the function instead of passing a bool parameter?

exp/orderbook/pools_test.go
```diff
@@ -318,7 +419,19 @@ func calculatePoolExpectationBig(
 	// hacky way to ceil(): if there's a remainder, add 1
 	if rem.Cmp(big.NewInt(0)) > 0 {
 		result.Add(result, big.NewInt(1))
+
+		// Recalculate with more precision
```
2opremio, line 423: why is this necessary?

exp/orderbook/pools.go
paulbellamy: yeah. They have a bunch of the same precursor calculations, so the idea was to share those. Will have a go at refactoring it, though.

exp/orderbook/pools.go
```diff
@@ -84,65 +84,85 @@ func makeTrade(
...
+		S.Abs(S).Mul(S, maxBips)
+		S.Div(S, unrounded)
+		S.Div(S, uint256.NewInt(100)) // Take off the excess 2 decimal places
+		roundingSlippageBips = xdr.Int64(S.Uint64())
+		ok = ok && S.IsUint64() && roundingSlippageBips >= 0
+	}
 
 	val := xdr.Int64(result.Uint64())
-	return val, result.IsUint64() && val >= 0
+	ok = ok && result.IsUint64() && val >= 0
```
2opremio, line 137: Isn't the `val >= 0` check redundant? I would think so (the refactoring I am doing is easier without it)

paulbellamy: No. It might be a uint64, but not an int64.

exp/orderbook/pools.go
paulbellamy: Ok, splitting it into two functions, means we either have a bunch of duplicated precursor calculations, or we extract those into a third function. But, because the Go compiler sucks at inlining, extracting those into a third function slows down the `BenchmarkLiquidityPoolExpectations` benchmark by ~50%.

2opremio (approved):

support/config/config_option.go
```diff
@@ -141,6 +141,8 @@ func (co *ConfigOption) setSimpleValue() {
 			*(co.ConfigKey.(*uint)) = uint(viper.GetInt(co.Name))
 		case types.Uint32:
 			*(co.ConfigKey.(*uint32)) = uint32(viper.GetInt(co.Name))
+		case types.Float64:
```
bartekn, line 144: Leftover because the new config value was a `float64` at some point?

Issue #3268, opened 2020-12-07
Integration tests don't cleanup their Postgres databases


2opremio:
While writing https://github.com/stellar/go/pull/3267 I found out that the integration tests are not dropping their dedicated Horizon Postgres database.

I tried fixing it (look at the TODOs in the code from the PR above) but it turns out that Horizon's `App` leaves open sessions behind .... even after closing it (leading to `database  is being accessed by other users` errors).

This may indicate that Horizon is leaking DB connections, so we should look into it.

bartekn:
Possibly fixed?

2opremio:
I will test this out.

2opremio:
It's still broken, I get this error upon uncomenting the `postgres.Close()` call.

```
=== CONT  TestProtocol16Basics
    db.go:108: 
        	Error Trace:	db.go:108
        	            				db.go:135
        	            				db.go:53
        	            				integration.go:181
        	            				integration.go:170
        	            				once.go:68
        	            				once.go:59
        	            				integration.go:167
        	            				testing.go:901
        	            				testing.go:1007
        	            				testing.go:1187
        	Error:      	Received unexpected error:
        	            	pq: database "test_b9831a09ca15" is being accessed by other users
        	Test:       	TestProtocol16Basics
--- FAIL: TestProtocol16Basics (26.89s)
=== RUN   TestProtocol16Basics/Sanity
```

The test passes when the call is commented.

PR #4185, opened 2022-01-18
internal/integrations: db_test should drop test db instances when finished

sreuland:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
After integration test was ran, the postgres DB instances it created were not getting dropped/deleted at end of test run.

### Why

It's best to clean up test db instances, especially for local development, as the db server can start to have hundreds of abandoned test db instances that are left over.

Closes: #3268 

### Known limitations

This effort focused on getting the test db instances purged after integrations/test_db.go, other integration tests and unit tests that create test db instances are still leaving instances behind after runtime, likely requires revisiting each tests and look for similar db conn patterns, as those tend to block being able to drop the db at end of test.

### DIFF
--- a/services/horizon/internal/ingest/main.go
+++ b/services/horizon/internal/ingest/main.go
@@ -693,6 +693,7 @@ func (s *system) Shutdown() {
 	s.cancel()
 	// wait for ingestion state machine to terminate
 	s.wg.Wait()
+	s.historyQ.Close()
 	if err := s.ledgerBackend.Close(); err != nil {
 		log.WithError(err).Info("could not close ledger backend")
 	}
--- a/services/horizon/cmd/db.go
+++ b/services/horizon/cmd/db.go
@@ -374,17 +374,14 @@ var dbFillGapsCmd = &cobra.Command{
 }
 
 func runDBReingestRange(ledgerRanges []history.LedgerRange, reingestForce bool, parallelWorkers uint, config horizon.Config) error {
+	var err error
+
 	if reingestForce && parallelWorkers > 1 {
 		return errors.New("--force is incompatible with --parallel-workers > 1")
-	}
-	horizonSession, err := db.Open("postgres", config.DatabaseURL)
-	if err != nil {
-		return fmt.Errorf("cannot open Horizon DB: %v", err)
 	}
 
 	ingestConfig := ingest.Config{
 		NetworkPassphrase:           config.NetworkPassphrase,
-		HistorySession:              horizonSession,
 		HistoryArchiveURL:           config.HistoryArchiveURLs[0],
 		CheckpointFrequency:         config.CheckpointFrequency,
 		MaxReingestRetries:          int(retries),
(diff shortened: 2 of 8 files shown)


services/horizon/internal/ingest/main.go
```diff
@@ -693,6 +693,7 @@ func (s *system) Shutdown() {
 	s.cancel()
 	// wait for ingestion state machine to terminate
 	s.wg.Wait()
+	s.historyQ.Close()
```
bartekn, line 696: `Session.Close()` is calling `Session.DB.Close()` and `DB` can be shared with other `Session` instances outside ingestion package. I think we shouldn't call it here. `DB` should be closed when all Horizon subsystems are closed and this is done in `App.CloseDB()`.

sreuland: I stepped through this in debug, using PGAdmin to view open sessions on the test db instance as horizon app shuts down from a app.Close() invocation, and what I observed:

(1) waitForDone() https://github.com/sreuland/go/blob/3268_test_db_cleanup/services/horizon/internal/app.go#L170 runs first which then calls this ingest.Shutdown()
(2) App.CloseDB() https://github.com/sreuland/go/blob/3268_test_db_cleanup/services/horizon/internal/app.go#L157 runs second after waitForDone() finishes

horizon.waitForDone() and cmd.runDBReingestRange() are the only callers of Shutdown()

When horizon App shuts down, App.CloseDB() does not result in all the active sessions on the db instance being closed, there are 2 or 3 left open, when the historyQ.Close() is invoked from within ingest.Shutdown(), it results in those remaining sessions on db to be closed, maybe it's gc/memory thing since the DB session from horizon app is cloned and placed into a new history.Q struct instance which lives inside the ingest system struct scope, the sessions seem unique to their parent object references, which kind of contrasts the 'shared' aspect since, it appears as though each session instance has exclusive unique db sessions, and they won't be affected by the other, only when both of the sessions are closed, then the integration test cleanup routine that tries to drop the test db will work.

sreuland: @bartekn , did the session closing behavior i mentioned sound sensible? when ingest initiates shutdown it's from horizon waitForDone() which means overall app is going down anyway or from a `db reingest` command via runDBREingestRange(), the process is ending in both cases when ingest shutdown initiates the close on session which leads to DB. Without invoking the close on this ingest session instance, the underlying cloned session tends to lead to sessions left open on the db server, I haven't been able to remove them otherwise.

sreuland: @2opremio , wanted to get your opinion in this fix too, do you see value in taking this approach to delete test db instances, or is it just introducing unnecessary risk, etc. I was thinking to close this PR rather than lean on it anymore or adjust it if something more palpable is suggested. thanks for any insights!

bartekn: Sorry for a huge delay answering this. I missed the notification somehow. I prepared the code sample below that shows that closing a cloned session actually closes the underlying DB connection for both sessions:
```go
package main

import (
	"context"

	"github.com/stellar/go/support/db"
)

func main() {
	session, err := db.Open("postgres", "postgres://bartek@localhost:5432/test?sslmode=disable")
	if err != nil {
		panic(err)
	}

	c1 := session.Clone()
	c2 := session.Clone()

	_, err = c1.ExecRaw(context.Background(), "SELECT 1;")
	if err != nil {
		panic(err)
	}

	err = c1.Close()
	if err != nil {
		panic(err)
	}

	_, err = c2.ExecRaw(context.Background(), "SELECT 1;")
	if err != nil {
		panic(err)
	}
}
```
Result:
```
panic: exec failed: sql: database is closed

goroutine 1 [running]:
main.main()
	/Users/bartek/go/src/github.com/stellar/go/shawn_test/main.go:30 +0x132
exit status 2
```

The reason why the code works in this PR is that in `waitForDone` the ingestion system is shut down after webserver (`a.ingester.Shutdown()`). But if we ever move it before a webserver or add another service that will send DB connections in the background and will be used after `a.ingester.Shutdown()` this will generate errors and bugs (like background worker will not be able to commit it's job).

sreuland: Thanks for verifying the go db/session close behavior, very helpful to rule that out!
yes, we don't want implicit coupling like that either, ok, will look into more. w/o this DB close in ingest shutdown, we strangely get the opposite effect, which is when `App.CloseDB()` is executed, the sessions related to ingestion don't get closed on the db server side as observed in PGAdmin, that is, until the go o/s process is terminated at which point they are gone.

2opremio:
I _think_ that the problem this PR is trying to solve is causing the `Error starting ingestion read transaction: beginTx failed: pq: sorry, too many clients already` at integration tests (such as https://github.com/stellar/go/runs/6233228705 and https://github.com/stellar/go/runs/6267795798)

2opremio:
Ah, hold on, this is only solving the DB drop, but it doesn't make any improvements in connection reduction, does it?

services/horizon/internal/ingest/main.go
2opremio: > horizon.waitForDone() and cmd.runDBReingestRange() are the only callers of Shutdown()

All tests call it implitcitly on shutdown due to this:

```go
	// Register cleanup handlers (on panic and ctrl+c) so the containers are
	// stopped even if ingestion or testing fails.
	i.t.Cleanup(i.Shutdown)
```

2opremio (approved):

Issue #4170, opened 2022-01-10
Spike: ideas for ingestion filtering enhancements for scaling purposes


sreuland:
### What problem does your feature solve?
provide visibility of options for how horizon ingestion scales for short term historical retention as TPS grows 

### What would you like to see?
summarized tradeoffs on design options that address a single deployment of Horizon and how ingestion can be enhanced through filter mechanisms to enable deployments that avoid performance degradation and no run-away resource usage(ram/disk).

### What alternatives are there?
continue to use the current retention period mechanism:
`--history-retention-count <num_of_ledgers>`

|                |Pro                          |Con                        |
|----------------|-------------------------------|-----------------------------|
|Resource filter rules          | x            |x         |
|Temporal filter rules          |x          | x         |
|Account filter rules           | x          |x.         |
|Ledger type filter rules           | x          |x.         |

PR #4187, opened 2022-01-18
#4170: added skeleton of new asset filter

sreuland:


services/horizon/internal/ingest/filters/asset_filter.go
```diff
@@ -0,0 +1,67 @@
...
+	// if a liquidity pool references a filtered asset, then include all operation
+	// types within 'traverseOperationsList' that are related to the pool
+	resolveLiquidityPoolAsAsset bool
+
+	// true means generate effects for any operations referencing a filtered asset
+	traverseEffects bool
+
+	// list of 'Offers, Trades, Payments, TrustLine, Claimable Balance', include any
+	// of these operation types when they reference a filtered asset
+	traverseOperationsList []string
```
2opremio, line 30: I would keep it simple to start with and hardcode what you think is most reasonable.

sreuland: yes, I just added commit for filter params as json file also, to consider as light(er) weight middle ground between cli which is untenable as you mention and the db persistence for params.

services/horizon/internal/ingest/filters/asset_filter.go
```diff
@@ -0,0 +1,176 @@
...
+		traverseOperationsLookup: listToMap(filterParams.TraverseOperationsList),
+	}
+	return filter
+}
+
+func (f *AssetFilter) CurrentFilterParameters() AssetFilterParms {
+	return *f.filterParams
+}
+
+func (f *AssetFilter) FilterTransaction(ctx context.Context, transaction ingest.LedgerTransaction) (bool, error) {
```
sreuland, line 86: @2opremio , on the filter interface, I'm wondering, is the tx level a bit too course grained, this will result in dropping  operations in the tx that have no relevance to the filter? wondering if the the filter method interface should work a bit lower at the Operation level, like:

```
func FilterTransaction(ctx context.Context, transaction ingest.LedgerTransaction) ([]*xdr.Operation, error) 
```

2opremio: You are bring up a  really good point which, as you have problably seen, I incorporated to the design doc.

Since we are not sure what's the best granularity, I think it's better to keep it simple. Filtering whole transactions is arguably simpler.

However, if you have any examples in which full-transaction filtering doesn't make sense, please share them. That may justify going for finer granularity right away.

sreuland: Ok, it looks like like two effects from filtering at tx level emerge:

the first appears to be a net effect that would include additional operations that didn't match the filter intent, i.e. if an asset filter rule is for '3', and a TX comes in like this:
```
tx {
   opPayment: (Asset:1)
   opPayment: (Asset:2)
   opPathPayment: (SendAsset:2, DestAsset:3)
   opPathPayment: (SendAsset:4, DestAsset:5)
}
```

then the the asset filter would return `true` and the other ops with non-filtered assets would still get persisted, making the space saving aspect of the filter less effective. 

the second effect of filtering at tx level appears to be in case of asset filter could cause dropping op data that has nothing to do with filtered assets, i.e. if a tx has no ops that contained a whitelisted asset, then the asset filter returns `false` triggering the tx to be dropped, but it may have other non-filtered asset related ops data that should continue to be persisted , such as if asset filter rule is for '5', and this tx comes in:

```
tx {
   opPayment: (Asset:1)
   opPathPayment: (SendAsset:2, DestAsset:3)
   opMergeAcct: (Acct:1234)
   opSponsoring: (Acct:5555)
}
```

the asset filter returns `false`, and the other non-asset relevant ops like the acct ops, would get dropped. 

maybe for first pass on poc, we don't need to address this yet, just so we can start seeing some rough numbers or if we should consider getting the filter integrated at the ops level up front.

sreuland:
@2opremio , would like to get your opinion on a possible design change for accessing filters as a static singleton rather than constructing it in processor_runner.go, processors that want to be filter-aware, can invoke `filters.ProcessFilters(operation)` directly:
filter_ops_flow

2opremio:
> [edit] - nvm, I see processor_runner using a creational pattern per ledger for processors with`processor_runner:buildTransactionProcessor()`, and filters following the same pattern, so, looks consistent to follow that. The filters will at some point be parsing config parameters for rules during instantiation, may be some optimization on that later.

👍

services/horizon/internal/ingest/filters/asset_filter.go
2opremio: > then the the asset filter would return `true` and the other ops with non-filtered assets would still get persisted, making the space saving aspect of the filter less effective.

Good point. This is extensible to accounts as well.

From the usability standpoint, I think it really depends on how transactions are used and giving a consistent view from Horizon.

1. I think that in many cases transactions only make sense as a unit (e.g. sponsoring sandwich), even if you are only whitelisting certain accounts/assets in the transaction.
2. Also, only including certain operations from a transaction will result in information gaps when obtaining information about the transaction (effects, participants etc ... ). Admittedly, the same happens at the  transaction->ledger level, but end users don't produce ledgers, they submit full transactions. So I think it's less aggressive to partially omit ledger->transaction relationships than transaction->operation ones.

From the performance standpoint, it totally makes sense to have a finer granularity (resulting in larger storage savings). So we will have to see.

2opremio: > the second effect of filtering at tx level appears to be in case of asset filter could cause dropping op data that has nothing to do with filtered assets,

That's a really really good point. This doesn't really apply to accounts (since all operations are account-related).

I think that, as a user, I may be surprised about operations `opMergeAcct: (Acct:1234), opSponsoring: (Acct:5555)` being dropped.

The current asset filter implementation only makes sense for users who *only* care about asset operations. This may be too strict to be generally useful, I am not sure.

I don't really have a good solution for this, though. If users want to keep track of asset-unrelated operations while whitelisting assets, we will probably need operation-level filtering.

sreuland: > The current asset filter implementation only makes sense for users who only care about asset operations. This may be too strict to be generally useful, I am not sure.

> I don't really have a good solution for this, though. If users want to keep track of asset-unrelated operations while whitelisting assets, we will probably need operation-level filtering.

@jcx120 , any insights from product/user standpoint on this type of effect from asset filtering, pls let us know, thanks!

services/horizon/internal/db2/schema/migrations/53_whitelists.sql
```diff
@@ -1,9 +1,9 @@
 -- +migrate Up
 
-CREATE TABLE account_filter_whitelist (
-    account_id character varying(56) NOT NULL UNIQUE
+CREATE TABLE filter_config (
```
sreuland, line 3: @2opremio ,just thinking out loud on filter config aspects, such as generic filter config schema, example - https://github.com/stellar/go/pull/4187/files#diff-060057b3a0aa48e8d48274558827e83c7c4ded5edb0a416fc5b9eebaa32882d1

rather than exposing config from http API , which requires some security to restrict access, what if we used`horizon db` subcommand like `horizon db config filter` to display to console and `horizon db config filter load <json_file_path>`, which only a sysadmin will have access at that cli level. 

the user can define filer rules before ingestion starts, such as`horizon db init` and then `horizon db config filter load myfilters.json` could be done before`horizon serve ....` and afterwards to update per the filter config cacheing behavior ideas.

services/horizon/internal/httpx/router.go
```diff
@@ -130,6 +131,7 @@ func (r *Router) addMiddleware(config *RouterConfig,
 func (r *Router) addRoutes(config *RouterConfig, rateLimiter *throttled.HTTPRateLimiter, ledgerState *ledger.State) {
 	stateMiddleware := StateMiddleware{
 		HorizonSession: config.DBSession,
+		NoStateVerification: config.IngestDisableStateVerification,
```
sreuland, line 134: it looked like `DisableStateVerification` was intentionally left off the config, as i noticed `--ingest-disable-state-verification=true` had no effect from command line, was getting `Still ingesting` so, I added it in here, just to get it working in local runtime tests, but can remove this was not meant to be enabled.

services/horizon/internal/ingest/filters/asset_filter.go
2opremio: Can you make the file layout consistent with the account filtering? (i e. Either move account filtering to this directory or remove the directory)

services/horizon/internal/ingest/processor_runner.go
```diff
@@ -150,6 +151,21 @@ func (s *ProcessorRunner) buildTransactionProcessor(
 func (s *ProcessorRunner) buildTransactionFilterer() *groupTransactionFilterers {
 	return newGroupTransactionFilterers([]processors.LedgerTransactionFilterer{
 		processors.NewAccountFilter(s.historyQ),
+		filters.NewAssetFilterFromParams(&filters.AssetFilterParms{
+			// this is example asset filter config by list of assets that were
```
2opremio, line 155: Can you add a TODO about making this configurable?

2opremio (changes requested):

services/horizon/internal/ingest/filters/asset_filter.go
sreuland: yes, had to make processors.ParticipantsForTransaction public, [79c777311c04](https://github.com/stellar/go/pull/4187/commits/79c777311c04b71827482837dfc25819273ecc50)

services/horizon/internal/db2/history/account_filter_whitelist.go
```diff
@@ -48,6 +48,8 @@ func (q *Q) SetAccountFilterWhitelist(ctx context.Context, accountIDs []string)
 
 // RemoveFromAccountFilterWhitelist deletes rows from the account_filter_whitelist table.
 // Returns the number of rows affected and error.
+// TODO - is this needed, currently not referenced, static check?
```
2opremio, line 51: Let's remove this!

PR #4192, opened 2022-01-23
ingest/ledgerbackend: Restart Stellar-Core when it's context is cancelled

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

After refactoring in https://github.com/stellar/go/pull/4088 (and as a result of my wrong comment: https://github.com/stellar/go/pull/4088#discussion_r764394296) the `CaptiveCoreBackend.isPrepared` method returned `true` if `stellarCoreRunner` process was shutdown without calling `close()` - so in case of binary update but also in case of Stellar-Core crash.

This commit fixes this bug by checking if `stellarCoreRunner` context was cancelled (meaning Stellar-Core is closed or closing but not as a result of `close` call). I also removed `isClose` method because it was simply checking `close` variable. All test changes are only adding an extra `context()` call mocks because `isPrepared` now calls it.

### DIFF
--- a/ingest/ledgerbackend/captive_core_backend.go
+++ b/ingest/ledgerbackend/captive_core_backend.go
@@ -398,7 +398,7 @@ func (c *CaptiveStellarCore) IsPrepared(ctx context.Context, ledgerRange Range) 
 }
 
 func (c *CaptiveStellarCore) isPrepared(ledgerRange Range) bool {
-	if c.isClosed() {
+	if c.closed {
 		return false
 	}
 
@@ -462,7 +462,7 @@ func (c *CaptiveStellarCore) GetLedger(ctx context.Context, sequence uint32) (xd
 		return *c.cachedMeta, nil
 	}
 
-	if c.isClosed() {
+	if c.closed {
 		return xdr.LedgerCloseMeta{}, errors.New("stellar-core is no longer usable")
 	}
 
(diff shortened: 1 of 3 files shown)


ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -626,10 +624,6 @@ func (c *CaptiveStellarCore) GetLatestLedgerSequence(ctx context.Context) (uint3
 	return *c.lastLedger, nil
 }
 
-func (c *CaptiveStellarCore) isClosed() bool {
-	return c.closed
-}
```
Shaptic, line 631: Could we simplify this to include the above condition, instead? In other words,
```go
func (c *CaptiveStellarCore) isClosed() bool {
	return c.closed || c.stellarCoreRunner == nil || c.stellarCoreRunner.context().Err() != nil
}
```

Basically, is there a situation in which the instance would _not_ be closed and the runner wouldn't exist? Intuitively seems like no, but I'm not strong on this part of the codebase.

bartekn: Adding extra conditions to `isClosed` won't work. In https://github.com/stellar/go/pull/4088 we wanted to separate two cases:
* `CaptiveCoreBackend` closed - means no usable, you need to create a new instance.
* `stellarCoreRunner` closed - means it just needs to be restarted by calling `PrepareRange`.

Shaptic: Got it, thanks for clarifying! LGTM :+1:

Shaptic (approved):

PR #4193, opened 2022-01-24
friendbot: create new problem on each handler call

acharb:
<details>
  
### PR Structure

* [ x] This PR has reasonably narrow scope (if not, break it down into smaller PRs).
* [ ]x This PR avoids mixing refactoring changes with feature changes (split into two PRs
  otherwise).
* [ x] This PR's title starts with name of package that is most changed in the PR, ex.
  packages.

### Thoroughness

### Release planning

</details>

### What

fixes friendbot handler to register the account exists error on app start

### Why

Right now each request is registering the same error, which created an error when two requests try to access at the exact same time. See `fatal error: concurrent map writes` error at bottom of [kibana logs](https://kibana.stellar-ops.com/app/kibana#/discover?_g=(refreshInterval:(pause:!t,value:0),time:(from:'2022-01-23T13:25:27.059Z',mode:absolute,to:'2022-01-23T13:25:41.722Z'))&_a=(columns:!(message),filters:!(('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'logstash-*',key:application,negate:!f,params:(query:kube,type:phrase),type:phrase,value:kube),query:(match:(application:(query:kube,type:phrase)))),('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'logstash-*',key:k8s_cluster,negate:!f,params:(query:kube001-prd,type:phrase),type:phrase,value:kube001-prd),query:(match:(k8s_cluster:(query:kube001-prd,type:phrase)))),('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'logstash-*',key:kubernetes.namespace,negate:!f,params:(query:friendbot-prd,type:phrase),type:phrase,value:friendbot-prd),query:(match:(kubernetes.namespace:(query:friendbot-prd,type:phrase)))),('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'logstash-*',key:kubernetes.deployment.name,negate:!f,params:(query:friendbot,type:phrase),type:phrase,value:friendbot),query:(match:(kubernetes.deployment.name:(query:friendbot,type:phrase)))),('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'logstash-*',key:kubernetes.pod.name,negate:!f,params:(query:friendbot-6d669566f6-c644h,type:phrase),type:phrase,value:friendbot-6d669566f6-c644h),query:(match:(kubernetes.pod.name:(query:friendbot-6d669566f6-c644h,type:phrase))))),index:'logstash-*',interval:auto,query:(language:lucene,query:'NOT%20%22finished%20request%22%20AND%20NOT%20%22starting%20request%22%20AND%20NOT%20%22Selecting%20minion%22'),sort:!('@timestamp',desc)))

This now registers once to avoid that problem.

### Known limitations

n/a

### DIFF
--- a/services/friendbot/internal/friendbot_handler.go
+++ b/services/friendbot/internal/friendbot_handler.go
@@ -17,13 +17,15 @@ type FriendbotHandler struct {
 
 // Handle is a method that implements http.HandlerFunc
 func (handler *FriendbotHandler) Handle(w http.ResponseWriter, r *http.Request) {
+	// Register account doesn't exist error.
 	accountExistsProblem := problem.BadRequest
 	accountExistsProblem.Detail = ErrAccountExists.Error()
-	problem.RegisterError(ErrAccountExists, accountExistsProblem)
+	ps := problem.New(problem.DefaultServiceHost, problem.DefaultLogger, problem.LogAllErrors)
+	ps.RegisterError(ErrAccountExists, accountExistsProblem)
 
 	result, err := handler.doHandle(r)
 	if err != nil {
-		problem.Render(r.Context(), w, err)
+		ps.Render(r.Context(), w, err)
 		return
 	}
 
--- a/services/friendbot/internal/friendbot_handler.go
+++ b/services/friendbot/internal/friendbot_handler.go
@@ -17,15 +17,9 @@ type FriendbotHandler struct {
 
 // Handle is a method that implements http.HandlerFunc
 func (handler *FriendbotHandler) Handle(w http.ResponseWriter, r *http.Request) {
-	// Register account doesn't exist error.
-	accountExistsProblem := problem.BadRequest
-	accountExistsProblem.Detail = ErrAccountExists.Error()
-	ps := problem.New(problem.DefaultServiceHost, problem.DefaultLogger, problem.LogAllErrors)
-	ps.RegisterError(ErrAccountExists, accountExistsProblem)
-
 	result, err := handler.doHandle(r)
 	if err != nil {
-		ps.Render(r.Context(), w, err)
+		problem.Render(r.Context(), w, err)
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


services/friendbot/internal/friendbot_handler.go
```diff
@@ -17,13 +17,15 @@ type FriendbotHandler struct {
 
 // Handle is a method that implements http.HandlerFunc
 func (handler *FriendbotHandler) Handle(w http.ResponseWriter, r *http.Request) {
+	// Register account doesn't exist error.
 	accountExistsProblem := problem.BadRequest
 	accountExistsProblem.Detail = ErrAccountExists.Error()
-	problem.RegisterError(ErrAccountExists, accountExistsProblem)
+	ps := problem.New(problem.DefaultServiceHost, problem.DefaultLogger, problem.LogAllErrors)
+	ps.RegisterError(ErrAccountExists, accountExistsProblem)
```
leighmcculloch, line 24: ❗ I think we should register the error once on app start where errors are already being registered. The error is static and only needs registering once on app start rather than registering on every http request.
https://github.com/stellar/go/blob/8652df76c2551af1a2c1ab67a3b319a4cf80597c/services/friendbot/main.go#L100-L102

leighmcculloch (commented):
Thanks for jumping on this. One request (❗) below.

services/friendbot/internal/friendbot_handler.go
acharb: ah I didn't know that method existed. It felt weird at first to be using/modifying an instance from another package, but I agree this solution's better
https://github.com/stellar/go/pull/4193/commits/b0e7628a5aa26b49a5aa6d2a06ff7f21d8686447

leighmcculloch (approved):
:clap

services/friendbot/internal/friendbot_handler.go
leighmcculloch: Yeah fwiw I don't love the global / package level state that `problem` introduces across packages. It'd be better if we used problem.New and passed around a state to modify but here we are 😄.

PR #4204, opened 2022-01-31
services/horizon: Attempt state verification when bumping ledger sequence

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Try to verify state even when not ingesting current ledger.

### Why

In distributed ingestion, it's possible that a single node will keep ingesting for a longer period of time. If state verification is stuck for whatever reason on the ingesting instance the entire cluster will stop verifying state.

### Known limitations

[TODO or N/A]

### DIFF
--- a/services/horizon/internal/ingest/resume_state_test.go
+++ b/services/horizon/internal/ingest/resume_state_test.go
@@ -311,6 +311,9 @@ func (s *ResumeTestTestSuite) TestBumpIngestLedger() {
 		int32(101),
 	).Return(errors.New("my error")).Once()
 
+	// Skips state verification but ensures maybeVerifyState called
+	s.historyQ.On("GetExpStateInvalid", s.ctx).Return(true, nil).Once()
+
 	next, err := resumeState{latestSuccessfullyProcessedLedger: 99}.run(s.system)
 	s.Assert().NoError(err)
 	s.Assert().Equal(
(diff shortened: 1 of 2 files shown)


services/horizon/internal/ingest/resume_state_test.go
```diff
@@ -311,6 +311,9 @@ func (s *ResumeTestTestSuite) TestBumpIngestLedger() {
 		int32(101),
 	).Return(errors.New("my error")).Once()
 
+	// Skips state verification but ensures maybeVerifyState called
+	s.historyQ.On("GetExpStateInvalid", s.ctx).Return(true, nil).Once()
```
sreuland, line 315: just wondering if this were to return false, and allow the verification routine to run, would the test still pass also, not sure if that path is valuable also to trigger.

sreuland (approved):

services/horizon/internal/ingest/resume_state_test.go
bartekn: In case of `false` returned, we would need to mock more calls. The mocked call assertion here is just to ensure we call `maybeVerifyState`. Unfortunately, making this test more clear would require a larger refactoring (i.e. creating a struct responsible for running state verifier only).

PR #4210, opened 2022-02-02
Add ledgerexporter service which stores txmeta in GCS

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This is a prototype of a service which consumes tx meta from captive core and uploads individual ledgers as objects in a GCS bucket. Whenever a new ledger object is upload the service also updates the "latest"  object with the sequence number of the most recent ledger object in the GCS object.

Note we can have multiple instances of this service for redundancy in case one of the instances crashes. There is no coordination required by the multiple instances (GCS writes are atomic so it is safe if multiple instances write the same object at the same time).

### Why

Having tx meta is useful because it can save time when performing ingestion. It also provides an ingestion source which is easier to use than captive core. The problem with captive core is that you need to bundle the stellar-core binary and provide configuration for captive core. But reading tx meta from GCS we no longer need captive core as a dependency.

### Known limitations

* the service assumes there are no gaps (e.g. no ledger objects are missing and we have a contiguous range of ledger objects from genesis until the sequence recorded in the "latest" object)
* when starting from scratch it can will take a long time to populate all the ledger objects to span full history. we should support parallel exporters (similar to parallel ingestion).
* this is just a proof of concept so all the hardcoded configuration should be extracted out into parameters

### DIFF
--- a/exp/services/ledgerexporter/main.go
+++ b/exp/services/ledgerexporter/main.go
@@ -3,9 +3,11 @@ package main
 import (
 	"bytes"
 	"context"
+	"fmt"
 	"github.com/stellar/go/xdr"
 	"io"
 	"os"
+	"sort"
 	"strconv"
 	"time"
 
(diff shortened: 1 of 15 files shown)


Shaptic:
Can we close this PR in lieu of the Light Horizon mega-PR? #4352

PR #4218, opened 2022-02-07
services/ticker: ingest assets optimizations

stfung77:
Currently, the ticker `ingest assets` command is using around 4GB memory processing ~90k assets (~22k non-trash assets) on mainnet and takes around 25min to complete. The purpose of this PR is to reduce the memory footprint of this process and some speed optimizations.

**Observations:**
1) In parallelProcessAssets(), ticker was building a slice of non-trash assets and this consumed significant memory
2) Parallelism to process the assets was set to 10 with ~40% CPU utilization

**Changes:**
1) No longer maintain a slice of all non- trash "FinalAssets", created a assetQueue for producer workers to send
"FinalAssets". The consumer of this assetQueue writes the "FinalAsset" to database, allowing memory used by the "FinalAsset" to be freed. 
2) Waitgroup for parallelProcessAssets() is now based on parallelism and not the number of assets. This eliminated the need for "FinalAsset{IsTrash: true}" to be sent to a queue and waitgroup.Done() needing to be called for each asset. 
3) bump parallelism to 20, resulting in ~65% CPU utilization

**Results (tested locally, Macbook Pro - 16GB,  i9 2.3GHz):**
Before: ~3.9GB memory, ~25min to execute
After: ~175MB memory,  ~8min to execute


services/ticker/internal/actions_asset.go
```diff
@@ -19,29 +20,38 @@ func RefreshAssets(ctx context.Context, s *tickerdb.TickerSession, c *horizoncli
...
-			l.Error("Error inserting issuer:", dbIssuer, err)
-			continue
+	var wg sync.WaitGroup
+	parallelism := 20
+	assetQueue := make(chan *scraper.FinalAsset, parallelism)
+
+	go sc.ProcessAllAssets(0, parallelism, assetQueue)
+	wg.Add(1)
+
+	go func() {
```
leighmcculloch, line 30: 💡 This is just white space but I think will help with readability. On first read with the `wg.Add` close to the `ProcessAllAssets` I assumed that must have been the go routine you were tracking. On re-read I realize the go routine being tracked is the one further down. It's helpful to group related things together.
```suggestion
	go sc.ProcessAllAssets(0, parallelism, assetQueue)
	
	wg.Add(1)
	go func() {
```

For the same reason I think it would also be helpful to move the `var wg sync.WaitGroup` down too since the wait group is not used by the `ProcessAllAssets` call.

services/ticker/internal/scraper/asset_scraper.go
```diff
@@ -210,11 +210,15 @@ func makeFinalAsset(
 	t.LastChecked = now
 	t.AnchorAssetType = strings.ToLower(t.AnchorAssetType)
 
+	// TODO remove
+	// tSize := int(unsafe.Sizeof(*t))
+	// fmt.Println(tSize)
```
leighmcculloch, line 215: ☝🏻

services/ticker/internal/scraper/asset_scraper.go
```diff
@@ -234,23 +238,23 @@ func processAsset(asset hProtocol.AssetStat, shouldValidateTOML bool) (FinalAsse
 }
 
 // parallelProcessAssets filters the assets that don't match the shouldDiscardAsset criteria.
+// non-trash assets are sent to the assetQueue.
 // The TOML validation is performed in parallel to improve performance.
-func (c *ScraperConfig) parallelProcessAssets(assets []hProtocol.AssetStat, parallelism int) (cleanAssets []FinalAsset, numTrash int) {
-	queue := make(chan FinalAsset, parallelism)
+func (c *ScraperConfig) parallelProcessAssets(assets []hProtocol.AssetStat, parallelism int, assetQueue chan *FinalAsset) (numNonTrash int, numTrash int) {
```
leighmcculloch, line 243: 💡 It is helpful to the reader, and anyone maintaining the code, to specify the access required to the channel. It looks to me like this function writes to and closes the channel, therefore you can add the `<-` to make it explicit that's the case.
```suggestion
func (c *ScraperConfig) parallelProcessAssets(assets []hProtocol.AssetStat, parallelism int, assetQueue chan<- *FinalAsset) (numNonTrash int, numTrash int) {
```

services/ticker/internal/scraper/main.go
```diff
@@ -110,20 +110,20 @@ type OrderbookStats struct {
 	SpreadMidPoint     float64
 }
 
-// FetchAllAssets fetches assets from the Horizon public net. If limit = 0, will fetch all assets.
-func (c *ScraperConfig) FetchAllAssets(limit int, parallelism int) (assets []FinalAsset, err error) {
+// ProcessAllAssets fetches assets from the Horizon public net. If limit = 0, will fetch all assets.
+func (c *ScraperConfig) ProcessAllAssets(limit int, parallelism int, assetQueue chan *FinalAsset) (numNonTrash int, numTrash int) {
```
leighmcculloch, line 114: 💡 Same thing here:
```suggestion
func (c *ScraperConfig) ProcessAllAssets(limit int, parallelism int, assetQueue chan<- *FinalAsset) (numNonTrash int, numTrash int) {
```

services/ticker/internal/scraper/asset_scraper.go
```diff
@@ -149,13 +149,13 @@ func makeFinalAsset(
...
 	errors []error,
-) (t FinalAsset, err error) {
+) (t *FinalAsset, err error) {
 	amount, err := strconv.ParseFloat(asset.Amount, 64)
 	if err != nil {
 		return
 	}
 
-	t = FinalAsset{
+	t = &FinalAsset{
```
leighmcculloch, line 158: ❓ I noticed here and in a few places `FinalAsset` was changed to `*FinalAsset`. Why was that? I wouldn't have expected that to make a significant difference to the memory usage.

leighmcculloch (commented):
One question (❓) and a few suggestions (💡), otherwise this looks great.

services/ticker/internal/scraper/asset_scraper.go
stfung77: I was playing with the idea of pruning some fields but that wasnt needed. you're right though, if anything, this uses more memory with the pointer. I'll remove it.

leighmcculloch (approved):
🎉

PR #4228, opened 2022-02-13
services/horizon: Release DB connection in /paths when no longer needed

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Release DB connection to the pool as soon as it's no longer needed in `/paths/*`.

### Why

Horizon starts `REPEATABLE READ` transaction in `StateMiddleware` to ensure that queries return consistent (at the same ledger) data when sending multiple queries. However, the tx is held in `/paths/*` for entire request time, even though the path finding algorithm is not using a DB.

### Known limitations

[TODO or N/A]


paulbellamy (approved):

services/horizon/internal/actions/path.go
```diff
@@ -148,6 +148,18 @@ func (handler FindPathsHandler) GetResource(w HeaderWriter, r *http.Request) (in
 		}
 	}
 
+	// Rollback REPEATABLE READ transaction so that a DB connection is released
```
tamirms, line 151: @bartekn I think it would be cleaner to modify the StateMiddleware so that it can be configured without the repeatable read transactions.

tamirms: actually, nevermind. I just realized that approach would not work because I forgot that we have to do a db query to fetch assets belonging to an account. you can ignore this comment

tamirms (approved):

PR #4231, opened 2022-02-16
horizon: Make sure we test reingestion for all possible operations

### LINKED ISSUE #3660: Evaluate reingestion testing
#3659 could had probably been avoided through testing.

We should evaluate Horizon's testing of reingestion.

2opremio:
This change ensures that all existing operations are reingested and that all future operations are covered(by going through `xdr.OperationTypeToStringMap`)

Closes #3660

### DIFF
--- a/services/horizon/internal/integration/db_test.go
+++ b/services/horizon/internal/integration/db_test.go
@@ -25,14 +25,13 @@ import (
 	"github.com/stellar/go/txnbuild"
 )
 
-func generateLiquidityPoolOps(itest *integration.Test, tt *assert.Assertions) (lastLedger int32) {
-
+func submitLiquidityPoolOps(itest *integration.Test, tt *assert.Assertions) (submittedOperations []txnbuild.Operation, lastLedger int32) {
 	master := itest.Master()
 	keys, accounts := itest.CreateAccounts(2, "1000")
 	shareKeys, shareAccount := keys[0], accounts[0]
 	tradeKeys, tradeAccount := keys[1], accounts[1]
 
-	itest.MustSubmitMultiSigOperations(shareAccount, []*keypair.Full{shareKeys, master},
+	allOps := []txnbuild.Operation{
 		&txnbuild.ChangeTrust{
 			Line: txnbuild.ChangeTrustAssetWrapper{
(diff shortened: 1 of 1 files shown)


services/horizon/internal/integration/db_test.go
```diff
@@ -104,43 +104,333 @@ func generateLiquidityPoolOps(itest *integration.Test, tt *assert.Assertions) (l
...
+			DestAmount:  "10",
+			Path:        []txnbuild.Asset{txnbuild.NativeAsset{}},
+		},
+	}
+	txResp := itest.MustSubmitOperations(itest.MasterAccount(), itest.Master(), ops...)
+
+	return ops, txResp.Ledger
+}
+
+func submitSponsorshipOps(itest *integration.Test, tt *assert.Assertions) (submittedOperations []txnbuild.Operation, lastLedger int32) {
```
sreuland, line 164: just wondering, do you think these would be worthwhile as reusable test fixture methods in integration.go?

services/horizon/internal/integration/db_test.go
```diff
@@ -104,43 +104,333 @@ func generateLiquidityPoolOps(itest *integration.Test, tt *assert.Assertions) (l
...
+	}
+	// Inflation is not supported
+	delete(allOpTypes, xdr.OperationTypeInflation)
+
+	for _, op := range submittedOps {
+		opXDR, err := op.BuildXDR()
+		tt.NoError(err)
+		delete(allOpTypes, opXDR.Body.Type)
+	}
+	tt.Empty(allOpTypes)
```
sreuland, line 433: does the additional data factor into test verifications, is it based on the premise that the `Test*` methods would throw errors on their reingestion command invocations if they weren't able to process with all the ops data present?

2opremio: It's simply about future-proof coverage. The idea is that if a new operation is added to the XDR code, this test will fail until the operation is added to the test.

services/horizon/internal/integration/db_test.go
2opremio: I am note sure they would. Maybe!

sreuland (approved):

Issue #4197, opened 2022-01-27
services/horizon: Liquidity pools derived from order book stream does not match liquidity pool from ingestion


bartekn:
### What version are you using?

2.13.0 (issue occurs in 2.12.1 too)

### What did you do?

I noticed that:
```
liquidity pools derived from order book stream does not match liquidity pool from ingestion
```
error occurs around a few times per hour on average in production pubnet cluster. It's also happening in a cluster running the previous 2.12.1 version. It's not a big problem because in case of such thing the graph is rebuilt.

May be connected: https://github.com/stellar/go/issues/4195

bartekn:
Improved logging in https://github.com/stellar/go/pull/4211 because it's currently impossible to read it.

PR #4236, opened 2022-02-18
Fix StreamAllLiquidityPools and StreamAllOffers

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Fixes https://github.com/stellar/go/issues/4197

There is a bug in StreamAllLiquidityPools() and StreamAllOffers() which is caused by reusing the same struct when unmarshallng xdr assets which are contained with the liquidity pool and offer structs. See 
https://github.com/stellar/go/pull/4155#discussion_r781306047 for more details.

This bug introduces a false positive in the orderbook verification routine which prints a warning in the horizon logs whenever the in memory orderbook is inconsistent with the postgres liquidity pool and offers table.

I have also added some test cases to catch this regression. 

### Known limitations

[N/A]


services/horizon/internal/db2/history/liquidity_pools_test.go
```diff
@@ -90,6 +91,33 @@ func TestRemoveLiquidityPool(t *testing.T) {
 	tt.Assert.Equal(lp, lpObtained)
 }
 
+func TestStreamAllLiquidity(t *testing.T) {
```
sreuland, line 94: this is close to TestFindLiquidityPoolsByAssets coverage, but it sorts on PoolID as well, was that the issue with reusing the same struct, just wanted to understand as I did that prior change to use the same struct storage for what seemed like an optimization on allocation activity, as it looked like all fields of the struct get reset by sql row scan.

sreuland: ah, ok, I stepped through debug on this, and see where the inner AssetReserves(LiquidityPoolReserves) slice field of the LiquidityPool struct during streaming is not getting reset with the current row's asset_reserves json blob during row scan, it looks like row scan only populates LiquidityPool:AssetReserves the first time, when the fields value is an uninitialized slice, ouch.

sreuland (approved):

bartekn (approved):
LGTM! @paulbellamy please include in 2.15.0.

Issue #4247, opened 2022-02-26
clients/horizonclient: claimable_balance only takes one parameter


iateadonut:
### What version are you using?

go@v0.0.0-20220224214604-8f9a595f3bbc

### What did you do?

I was trying to get a list of claimable balances of a single asset on a single wallet.

```
balances, err := client.ClaimableBalances(horizonclient.ClaimableBalanceRequest{
	Claimant: "GDJCDWRAAB3LA6TT32OBOXV2UYVJEVPKMQXDLUWYH7FQ73MCTH3UNPT7",
	//Sponsor: "GDFCYDQOVJ2OEWPLEGIRQVAM3VTOQ6JDNLJTDZP5S5OGTEHM5CIWMYBH",
	Asset: "AQUA:GBNZILSTVQZ4R7IKQDGHYGY2QXL5QOFJYQMXPKWRRM5PAV7Y4M67AQUA",
})
```

### What did you expect to see?

I expected to see the same results I would see from this url:
https://horizon.stellar.org/claimable_balances/?claimant=GDJCDWRAAB3LA6TT32OBOXV2UYVJEVPKMQXDLUWYH7FQ73MCTH3UNPT7&asset=AQUA:GBNZILSTVQZ4R7IKQDGHYGY2QXL5QOFJYQMXPKWRRM5PAV7Y4M67AQUA&limit=200

### What did you see instead?

I got the error: "invalid request: too many parameters"

I think I can fix this in claimable_balance_request.go; I can make a PR, but want to make sure this isn't already being fixed already.

iateadonut:
PR: https://github.com/stellar/go/pull/4248

PR #4248, opened 2022-02-26
clients/horizonclient: fix multi-parameter url for claimable balance query

iateadonut:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

bug: https://github.com/stellar/go/issues/4247

Fix cbr.BuildURL() so that it can build a multi-parameter url if one of the url's is not ID.

### Why

Bug does not allow the proper claimable balance query to be made.

### Known limitations

N/A


daesu (approved):
lgtm

clients/horizonclient/claimable_balance_request.go
```diff
@@ -16,19 +17,24 @@ func (cbr ClaimableBalanceRequest) BuildURL() (endpoint string, err error) {
 	// Only one filter parameter is allowed, and you can't mix an ID query and
 	// filters.
 	nParams := countParams(cbr.Asset, cbr.Claimant, cbr.Sponsor, cbr.ID)
-	if nParams > 1 {
+	if cbr.ID != "" && nParams > 1 {
```
Shaptic, line 20: I'm not sure this is necessary? `countParams` returns 1 if the `cbr.ID` is set.

clients/horizonclient/claimable_balance_request.go
```diff
@@ -16,19 +17,24 @@ func (cbr ClaimableBalanceRequest) BuildURL() (endpoint string, err error) {
...
 	} else {
+		params := map[string]string{
+			"claimant": cbr.Claimant,
+			"sponsor":  cbr.Sponsor,
+			"asset":    cbr.Asset,
+			"cursor":   cbr.Cursor,
+		}
+		if (cbr.Limit) > 0 {
+			params["limit"] = strconv.FormatUint(uint64(cbr.Limit), 10)
+		}
```
Shaptic, line 35: Let's let the backend handle the defaults

clients/horizonclient/claimable_balance_request.go
```diff
@@ -16,19 +17,24 @@ func (cbr ClaimableBalanceRequest) BuildURL() (endpoint string, err error) {
...
 		queryParams := addQueryParams(
-			map[string]string{
-				"claimant": cbr.Claimant,
-				"sponsor":  cbr.Sponsor,
-				"asset":    cbr.Asset,
-			},
+			params,
 		)
```
Shaptic, line 38: Let's also keep the query-specific filters (claimant, etc.) separate from the "built-in" filters (cursor, limit). Something like:

```go
params := map[string]string{
    "claimant": cbr.Claimant,
    "sponsor":  cbr.Sponsor,
    "asset":    cbr.Asset,
}

queryParams := addQueryParams(params, cursor(cbr.Cursor), limit(cbr.Limit))
```

clients/horizonclient/claimable_balance_request_test.go
```diff
@@ -0,0 +1,34 @@
...
+	cbr = ClaimableBalanceRequest{
+		ID:       "1235",
+		Claimant: "CLAIMANTADDRESS",
+	}
+	_, err = cbr.BuildURL()
+	assert.EqualError(t, err, "invalid request: too many parameters")
+
+	//if you have two parameters, and neither of them are ID, it must use both in the URL
+	cbr = ClaimableBalanceRequest{
+		Claimant: "CLAIMENTADDRESS",
```
Shaptic, line 28: nitpick on a typo: claim**e**nt vs. claim**a**nt

Shaptic (commented):
Thank you for your pull request! I've added a few small comments below, mostly to align the implementation here with the way the other request builders work.

clients/horizonclient/claimable_balance_request.go
iateadonut: Yes it is necessary; you can have more than one parameter if the ID is not set.

clients/horizonclient/claimable_balance_request.go
iateadonut: Horizon still handles the defaults.  If the user gives a limit, then that limit is used.  The default is 10, but the user can set it up to 200.

This is a problem other users have had; here the user works around by doing an http request:
https://github.com/altugbakan/stellar-quest-go/blob/main/series4/quest6/solution.go#L53

clients/horizonclient/claimable_balance_request.go
iateadonut: done.

clients/horizonclient/claimable_balance_request.go
iateadonut: you may be referring to the second parameter of strconv.FormatUint(), which indicates the base of the number, not the default.
anyway, it's a moot point as we got rid of this.

Shaptic: Oh wow yep, completely misread this line of code: saw it as "if cbr.Limit <= 0" :facepalm:

clients/horizonclient/claimable_balance_request.go
Shaptic: Huh, so the comment is a lie then :thinking: can you rephrase it? Then :heavy_check_mark: from my perspective :+1:

Shaptic (approved):
Thank you again for the PR!

clients/horizonclient/claimable_balance_request.go
iateadonut: updated the comment to:
// You can't mix an ID query and filters.

and the error verbiage to:
return endpoint, errors.New("invalid request: you cannot have an ID query with filters")

Issue #4229, opened 2022-02-16
horizon: Completely migrate CI from Circle to GitHub actions


2opremio:
We are already building and testing the monorepo with GitHub actions (see https://github.com/stellar/go/blob/master/.github/workflows/go.yml ). However, there are a few things left:

- [x] Integration tests (with and without captive core).
- [x] Build, test and push the verify-range Docker image
- [x] When releasing:
   - [x] Build the release artifacts and upload them to the release
   - [ ] ~Publish the horizon docker image in Dockerhub (which is put on hold until manually approved)~ (we will relay on Jenkins for this, which pushes `stellar/stellar-horizon:{latest,<version>}` when stabilizing a debian package).
- [x] Publish the horizon docker image to DockerHub for every commit in master
- [x] Publish the state-diff Docker image to DockerHub
- [ ] Change the repo configuration to use the GitHub actions as a mandatory successful check
- [ ] Remove CircleCI configuration
- [ ] Fully enable release jobs and watch them closely during a release

PR #4250, opened 2022-02-28
all: migrate the rest of the CircleCI jobs to GitHub Actions

2opremio:
Addresses most of  https://github.com/stellar/go/issues/4229

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -33,9 +33,7 @@ jobs:
     steps:
     - uses: actions/checkout@v2
       with:
-        # For pull requests, make sure we checkout `ref/pull/<PR_number>/head`
-        # and not `ref/pull/<PR_number>/merge` (which doesn't exist for the PR creator and
-        # is what github.ref points to)
+        # For pull requests, build and test the PR head not a merge of the PR with the destination.
         ref: ${{ github.head_ref || github.ref }}
     - uses: ./.github/actions/setup-go
       with:
@@ -83,9 +81,7 @@ jobs:
     steps:
     - uses: actions/checkout@v2
       with:
-        # For pull requests, make sure we checkout `ref/pull/<PR_number>/head`
-        # and not `ref/pull/<PR_number>/merge` (which doesn't exist for the PR creator and
-        # is what github.ref points to)
+        # For pull requests, build and test the PR head not a merge of the PR with the destination.
         ref: ${{ github.head_ref || github.ref }}
         # We need to full history for git-restore-mtime to know what modification dates to use.
         # Otherwise, the Go test cache will fail (due to the modification time of fixtures changing).
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -32,6 +32,11 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
     - uses: actions/checkout@v2
+      with:
+        # For pull requests, make sure we checkout `ref/pull/<PR_number>/head`
+        # and not `ref/pull/<PR_number>/merge` (which doesn't exist for the PR creator and
+        # is what github.ref points to)
+        ref: ${{ github.head_ref || github.ref }}
     - uses: ./.github/actions/setup-go
       with:
         go-version: ${{ matrix.go }}
@@ -78,6 +83,10 @@ jobs:
     steps:
     - uses: actions/checkout@v2
       with:
+        # For pull requests, make sure we checkout `ref/pull/<PR_number>/head`
+        # and not `ref/pull/<PR_number>/merge` (which doesn't exist for the PR creator and
+        # is what github.ref points to)
+        ref: ${{ github.head_ref || github.ref }}
         # We need to full history for git-restore-mtime to know what modification dates to use.
         # Otherwise, the Go test cache will fail (due to the modification time of fixtures changing).
         fetch-depth: '0'
--- a/.github/workflows/horizon-release.yml
+++ b/.github/workflows/horizon-release.yml
@@ -30,7 +30,8 @@ jobs:
         run: ./gomod.sh
 
       - name: Build binaries
-        run: go run ./support/scripts/build_release_artifacts/main.go
+        # TODO: remove GITHUB_REF override before merging, it's just used for testing
+        run: GITHUB_REF=refs/tags/horizon-v20.3.1 go run ./support/scripts/build_release_artifacts/main.go
 
       - name: Upload binaries to release
         uses: svenstaro/upload-release-action@483c1e56f95e88835747b1c7c60581215016cbf2
--- a/.github/workflows/horizon-release.yml
+++ b/.github/workflows/horizon-release.yml
@@ -10,6 +10,7 @@ jobs:
 
   publish-artifacts:
     runs-on: ubuntu-latest
+    name: Upload artifacts to GitHub release
     steps:
       - name: Run deprecation tests
         shell: bash
@@ -44,6 +45,7 @@ jobs:
 
   push-horizon-image:
     runs-on: ubuntu-latest
+    name: Push stellar/horizon:{version,latest} to DockerHub
     steps:
       - uses: actions/checkout@v2
 
--- a/.github/workflows/horizon-release.yml
+++ b/.github/workflows/horizon-release.yml
@@ -41,31 +41,3 @@ jobs:
           file_glob: true
           overwrite: true
 
-  push-horizon-image:
-    runs-on: ubuntu-latest
-    name: Push stellar/horizon:{version,latest} to DockerHub
-    steps:
-      - uses: actions/checkout@v2
-
-      - name: Extract Horizon version
-        shell: bash
-        id: get_version
-        run: echo ::set-output name=VERSION::${GITHUB_REF_NAME#horizon-v}
-
-      - name: Login to DockerHub
-        uses: docker/login-action@bb984efc561711aaa26e433c32c3521176eae55b
-        with:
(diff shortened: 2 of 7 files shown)


leighmcculloch:
@2opremio I see your commits:
- [See why the current PR commit is missing](https://github.com/stellar/go/pull/4250/commits/05e96e38d70b048bee40833938d60ad3a72608df)
- [It turns out that GITHUB_SHA is not equivalente to CIRCLE_SHA1](https://github.com/stellar/go/pull/4250/commits/c89957ee4ca6aa2b9fd44223306ed77a62a026c1)

I recently went through this pain adding docker image publishing to stellar/quickstart and ran into the same thing. If you want to take a look at the workflows there, checkout:
- https://github.com/stellar/quickstart/blob/5d478fb/.github/workflows/Dockerfile.yml#L18
- https://github.com/stellar/quickstart/blob/5d478fb/.github/workflows/Dockerfile.dev.yml
- https://github.com/stellar/quickstart/blob/5d478fb/.github/actions/push/action.yml

GitHub PRs run Github Actions on a commit that has the master branch merged into the PR's latest commit. So `github.sha` and `GITHUB_SHA` are not the PR's latest commit, they are a merge commit that includes the master branch.

If you want to prevent this from happening, you need to do two things:

1. Use `github.event.pull_request.head.sha` in PRs wherever you need the sha, instead of `github.sha`. E.g.:
   ```
   env:
     HEAD_SHA: ${{ github.event.pull_request.head.sha || github.sha }}
   ```
   See this for an example: https://github.com/stellar/quickstart/blob/5d478fb/.github/workflows/Dockerfile.yml#L18

2. Checkout the last commit in the PR rather than checkout the default commit the action runs with:
   ```
       - uses: actions/checkout@v2
         with:
           ref: ${{ env.HEAD_SHA }}
   ```

Given you're making a workflow that's designed to run on tagged releases, none of this might matter.

leighmcculloch (commented):
🎉

2opremio:
> I recently went through this pain adding docker image publishing to stellar/quickstart and ran into the same thing

Yes, that's really annoying. It turns out that, for PRs, `github.ref` points to `refs/pull/<PR_number>/merge` instead of `refs/pull/<PR_number>/head` which is unexpected.

In fact, I don't think we should be checking out `refs/pull/<PR_number>/merge` which doesn't even exist for the PR creator and may not contain the same code being proposed (but the code after merging).

I think I will create a composite action to make up for this

.github/workflows/horizon-release.yml
```diff
@@ -0,0 +1,73 @@
...
+      - name: Extract Horizon version
+        shell: bash
+        id: get_version
+        run: echo ::set-output name=VERSION::${GITHUB_REF_NAME#horizon-v}
+
+      - name: Login to DockerHub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKER_USER }}
+          password: ${{ secrets.DOCKER_PASS }}
```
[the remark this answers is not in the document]
2opremio: We just used the names on CircleCI

.github/workflows/horizon-release.yml
```diff
@@ -0,0 +1,71 @@
...
+          go-version: 1.17
+
+      - name: Check dependencies
+        run: ./gomod.sh
+
+      - name: Build binaries
+        run: go run ./support/scripts/build_release_artifacts/main.go
+
+      - name: Upload binaries to release
+        uses: svenstaro/upload-release-action@v2
```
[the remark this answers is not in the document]
2opremio: I did it for this one. However, I think we should be consistent and also do it for the github ones. Why treat those especially?

.github/workflows/horizon-release.yml
```diff
@@ -0,0 +1,71 @@
...
+
+      - name: Build and push to DockerHub
+        uses: docker/build-push-action@v2
+        with:
+          # TODO: Commented out until we disable the CircleCI jobs
+          # push: true
+          # TODO(fons): pushing stellar/horizon:latest should be reviewed and manually approved.
+          # We might not want the latest tag here being pushed on every release.
+          # We might release a patch for older versions of horizon.
+          # Otherwise, if we are currently on e.g. v1.6.1 and need to release v1.5.3, v1.5.3 will become latest.
```
2opremio, line 69: I don't know how to solve this. There's https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment#required-reviewers , but we are not really releasing to an environment.

I can't see anything in GitHub actions similar to CircleCI's `type: approval`

leighmcculloch:
> In fact, I don't think we should be checking out `refs/pull/<PR_number>/merge` which doesn't even exist for the PR creator and may not contain the same code being proposed (but the code after merging).

The intent behind this is not unique to GitHub Actions, other build systems work similar, the intent is to test the code that would result from the PR being merged. In this repo this maybe matters less in that we require branches to be up-to-date with master before they can be merged. Making the change in this PR further locks us into that I think. I don't have a problem with that, but something to be mindful.

.github/workflows/go.yml
```diff
@@ -32,6 +32,11 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
     - uses: actions/checkout@v2
+      with:
+        # For pull requests, make sure we checkout `ref/pull/<PR_number>/head`
+        # and not `ref/pull/<PR_number>/merge` (which doesn't exist for the PR creator and
+        # is what github.ref points to)
```
leighmcculloch, line 38: Suggest giving a slightly less in the weeds description. When I read this I feel like it only makes sense to me because I have a lot of context already.
```suggestion
        # For pull requests, build and test the PR head not a merge of the PR with the destination.
```

leighmcculloch (commented):
Couple asks, looking great though.

.github/workflows/horizon-release.yml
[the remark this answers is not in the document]
2opremio: I don't see any high value in making them uniform either.

Since it was @jacekn who added the secrets, I will leave it up to him to rename it.

@jacekn let's do whatever you prefer, I am happy to adjust the code if you rename the secrets.

.github/workflows/horizon-release.yml
2opremio: I am confused. This is about preventing pushing to `stellar/horizon:latest` if the version being released is lower than the version which as already pointing `stellar/horizon:latest`

2opremio: But, now that I think about it, we could simply automate that.

.github/workflows/horizon-release.yml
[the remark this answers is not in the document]
leighmcculloch: Ah, @jacekn we already have these secrets in use on the stellar/quickstart repo at least. Given these credentials are going to be used on multiple repos, it might be worth defining them once as organization secrets that are enabled for use on these two repos and other repos we add, so that we have less secrets to track. (cc @brahman81 If you have thoughts.)

.github/workflows/horizon-release.yml
```diff
@@ -0,0 +1,71 @@
...
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Extract Horizon version
+        shell: bash
+        id: get_version
+        run: echo ::set-output name=VERSION::${GITHUB_REF_NAME#horizon-v}
+
+      - name: Login to DockerHub
+        uses: docker/login-action@v1
```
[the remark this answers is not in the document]
2opremio: > since any https://github.com/github action compromised probably has similar attack vectors to compromising the GitHub Action

I am not sure I agree with that, but sure!  I will pin them.

.github/workflows/horizon-release.yml
leighmcculloch: I think of something didn't need to be in Jenkins it is probably better here. cc @brahman81

.github/workflows/horizon-release.yml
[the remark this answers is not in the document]
2opremio: @jacekn friendly ping :)

.github/workflows/horizon-release.yml
2opremio: The decision was to favor the Jenkins images since they are built with debian packages

.github/workflows/horizon-release.yml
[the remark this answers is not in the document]
jacekn: @2opremio apologies for the delay, I added 2 new variables now: `DOCKERHUB_USERNAME` and `DOCKERHUB_TOKEN`

2opremio:
@leighmcculloch PTAL

leighmcculloch (approved):
🎉 The actions look good to me. I defer to @stellar/horizon-committers folks for the Horizon specifics.

.github/workflows/go.yml
```diff
@@ -32,6 +32,9 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
     - uses: actions/checkout@v2
+      with:
+        # For pull requests, build and test the PR head not a merge of the PR with the destination.
+        ref: ${{ github.head_ref || github.ref }}
```
leighmcculloch, line 37: @2opremio This change looks unnecessary now. It isn't used by the new actions. Do you need it still?

leighmcculloch (changes requested):

.github/workflows/go.yml
leighmcculloch: Oh, I see, this is for consistency with the horizon verify range push.

leighmcculloch (approved):

PR #4251, opened 2022-02-28
services/horizon: Add LedgerHashStore to Captive-Core config

bartekn:
`LedgerHashStore` was removed from Horizon's `CaptiveCoreConfig` in: https://github.com/stellar/go/commit/b7fde588a61abdcb52b13b1b802b3e827202013b. This change makes it impossible to `PrepareRange` using Captive-Core if there is any other ingestion instance running because `CaptiveCoreBackend` tries to load the hash of the ledger from archives when `LedgerHashStore` is not set. But ledgers are published every 5 minutes so they contain the hash of the latest ledger only for a brief amount of time (the moment when checkpoint is published).


ire-and-curses (approved):

paulbellamy:
If I'm understanding it right, it seems like there's a much more general issue that all captive-cores (ingesting to a single db) need to stay in sync at all times, because ingestion isn't idempotent...

PR #4254, opened 2022-03-02
all: fix ci ref_protected check for caching

leighmcculloch:
### What
Invert the ref_protected checked used to control whether test caching occurs.

### Why
Caching of tests should be disabled for protected branches, not enabled.

### DIFF
--- a/.github/actions/setup-go/action.yml
+++ b/.github/actions/setup-go/action.yml
@@ -42,7 +42,7 @@ runs:
   # rebuilds and cause test runs to skip tests that have no reason to rerun.
   # Don't run this for protected branches like master/main.
   - uses: actions/cache@v2
-    if: (!github.ref_protected) || github.ref != 'refs/heads/master' || github.ref != 'refs/heads/main'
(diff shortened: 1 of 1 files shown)


acharb (approved):
lgtm!

.github/actions/setup-go/action.yml
```diff
@@ -42,7 +42,7 @@ runs:
   # rebuilds and cause test runs to skip tests that have no reason to rerun.
   # Don't run this for protected branches like master/main.
   - uses: actions/cache@v2
-    if: github.ref_protected || github.ref != 'refs/heads/master'
+    if: (!github.ref_protected) && github.ref != 'refs/heads/master' && github.ref != 'refs/heads/main'
```
2opremio, line 45: Do we use the main branch? Otherwise that's unnecessary. I also don't think we need parentheses.

leighmcculloch: We don't, although we did discuss transitioning at some point. No harm in adding it though. Same with the parentheses, I can never remember the typical order of precedence.

2opremio: It adds noise. If we ever use main, then we will change it

PR #4266, opened 2022-03-09
all: [GitHub actions] Reset the module and build cache in master/protected

2opremio:
Before, we:

1. Didn't use the build cache in the master branch. This caused the cache
   to not be updated. This caused PRs to rerun all the tests.

2. Didn't reset the go module cache. This would eventually cause the cache
   to accumulate orphan dependencies when we remove them.

### DIFF
--- a/.github/actions/setup-go/action.yml
+++ b/.github/actions/setup-go/action.yml
@@ -50,6 +50,6 @@ runs:
 
   # Reset the cache for master/protected branches, to ensure they build and run the tests from zero
   # and that the module cache is cleaned (otherwise it accumulates orphan dependencies over time).
-  - if: "github.ref_protected || github.ref == 'refs/heads/master'"
+  - if: github.ref_protected
     shell: bash
     run: rm -r ~/.cache/go-build ~/go/pkg/mod
--- a/.github/actions/setup-go/action.yml
+++ b/.github/actions/setup-go/action.yml
@@ -36,7 +36,7 @@ runs:
     with:
       path: ~/go/pkg/mod
       key: ${{ env.PREFIX }}-go-mod-${{ hashFiles('**/go.sum') }}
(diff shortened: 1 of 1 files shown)


.github/actions/setup-go/action.yml
```diff
@@ -40,12 +40,16 @@ runs:
...
     with:
       path: ~/.cache/go-build
       key: ${{ env.PREFIX }}-go-build-${{ github.ref }}-${{ hashFiles('**', '!.git') }}
       restore-keys: |
         ${{ env.PREFIX }}-go-build-${{ github.ref }}
         ${{ env.PREFIX }}-go-build
+
+  # Reset the cache for master/protected branches, to ensure they build and run the tests from zero
+  # and that the module cache is cleaned (otherwise it accumulates orphan dependencies over time).
+  - if: "github.ref_protected || github.ref == 'refs/heads/master'"
```
leighmcculloch, line 53: 💡 I think we should simplify this if. I think you pointed out previously it was superfluous.
```suggestion
  - if: github.ref_protected
```

.github/actions/setup-go/action.yml
```diff
@@ -40,12 +40,16 @@ runs:
...
+    shell: bash
+    run: rm -r ~/.cache/go-build ~/go/pkg/mod
```
leighmcculloch, line 55: It's a shame that we still have to download the cache in the case that we don't want to use it, but this seems like the simplest approach. I opened an issue on actions/cache proposing there be a way to write a cache value without reading it: https://github.com/actions/cache/issues/761.

leighmcculloch (approved):
One suggestion (💡), otherwise LGTM.

.github/actions/setup-go/action.yml
2opremio: Yeah, I didn't see a way to only create the cache

PR #4269, opened 2022-03-10
Attempt to fix installation of captive core in Integration tests

2opremio:


2opremio:
It seems to have macgically fixed itself

PR #4270, opened 2022-03-11
all: Forgot to add sudo in #4266

2opremio:


.github/actions/setup-go/action.yml
```diff
@@ -53,4 +53,4 @@ runs:
   # and that the module cache is cleaned (otherwise it accumulates orphan dependencies over time).
   - if: github.ref_protected
     shell: bash
-    run: rm -r ~/.cache/go-build ~/go/pkg/mod
+    run: sudo rm -r ~/.cache/go-build ~/go/pkg/mod
```
leighmcculloch, line 56: It seems odd to me that sudo is required when these directories are written to the home directory and they're written by a regular user, not a super user. Or maybe it is the superuser as implied by https://github.com/actions/cache/issues/512. It would be interesting to see who the owning user and group is, and what permissions are set on the files by running `ls -lah` in the directory.

bartekn (approved):

leighmcculloch (approved):
👍

PR #4274, opened 2022-03-13
all: More go-setup github action fixes

2opremio:
This time I tested that everything works in a fork.

### DIFF
--- a/.github/actions/setup-go/action.yml
+++ b/.github/actions/setup-go/action.yml
@@ -46,11 +46,11 @@ runs:
       key: ${{ env.PREFIX }}-go-build-${{ github.ref }}-${{ hashFiles('**', '!.git') }}
       restore-keys: |
         ${{ env.PREFIX }}-go-build-${{ github.ref }}-
-        ${{ env.PREFIX }}-go-build-master-
+        ${{ env.PREFIX }}-go-build-refs/heads/master-
         ${{ env.PREFIX }}-go-build-
 
   # Reset the cache for master/protected branches, to ensure they build and run the tests from zero
   # and that the module cache is cleaned (otherwise it accumulates orphan dependencies over time).
   - if: github.ref_protected
     shell: bash
-    run: sudo rm -r ~/.cache/go-build ~/go/pkg/mod
+    run: sudo rm -rf ~/.cache/go-build ~/go/pkg/mod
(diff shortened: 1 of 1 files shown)


.github/actions/setup-go/action.yml
```diff
@@ -46,11 +46,11 @@ runs:
       key: ${{ env.PREFIX }}-go-build-${{ github.ref }}-${{ hashFiles('**', '!.git') }}
       restore-keys: |
         ${{ env.PREFIX }}-go-build-${{ github.ref }}-
-        ${{ env.PREFIX }}-go-build-master-
+        ${{ env.PREFIX }}-go-build-refs/heads/master-
```
leighmcculloch, line 49: Why is this additional restore key that explicitly targets master required? The restore key immediately below should already serve this purpose. GitHub Actions only returns caches that originated from the current branch or from the default branch, i.e. master in our case. A build on branch A can't pickup a cache from branch B. so there shouldn't be any value in specifically trying to target the master branch here. 

Details about the cache scope are here: https://github.com/actions/cache#cache-scopes

.github/actions/setup-go/action.yml
```diff
@@ -46,11 +46,11 @@ runs:
...
         ${{ env.PREFIX }}-go-build-
 
   # Reset the cache for master/protected branches, to ensure they build and run the tests from zero
   # and that the module cache is cleaned (otherwise it accumulates orphan dependencies over time).
   - if: github.ref_protected
     shell: bash
-    run: sudo rm -r ~/.cache/go-build ~/go/pkg/mod
+    run: sudo rm -rf ~/.cache/go-build ~/go/pkg/mod
```
leighmcculloch, line 56: Is the -f needed because there are nested directories or some other reason?

leighmcculloch (commented):
The `-f` lgtm, but the restore key change is less clear on why it's needed or what's going on with it.

The PR description doesn't state what problem is being addressed, and so I'm guessing why these changes are needed which makes reviewing more difficult. It'd be helpful if the PR description used the what-why template that shares knowledge and context, and makes reviewing easier and efficient.

.github/actions/setup-go/action.yml
2opremio: Because there can be a cache miss, and in that case, the directory is not created

.github/actions/setup-go/action.yml
2opremio: Ah, I didn't know that. I thought it could restore cache entries created from other branches.

2opremio: That's been addressed

leighmcculloch: Yeah I did too but just came across that tidbit last week. Seems pretty sad if we have lots of PRs destined for a Horizon release branch and they all get cache from master instead of from the destination release branch. If I get time tomorrow I'll open an issue about this being not great on the actions/cache repo.

leighmcculloch (approved):

Issue #4267, opened 2022-03-09
services/horizon/ingest/filtering: Backfill low level design


sreuland:
### What problem does your feature solve?
Design proposal for backfill behavior of filters.

### What would you like to see?
A design that defines user experience with filter rule changes and backfill behaviors in horizon. 
Low level design writeup for filtered data range behavior. 

Acceptance Criteria:
- Low level design that provides dev path for supporting:
  - As a Horizon user, I want to configure the range of ingested historical data that my filter rules will be effective within aka 'backfill' timeframe by using existing `--history-retention-count`
  - As a Horizon user, I want any transactions related to entities that are recognized as allowed by filter rules to be included in 'backfill' timeframe.

### Pre-requisites:
- This design is dependent on #4221 and #4222 being implemented first to establish the initial filtration framework within ingestion.
- This feature leverages the existing `--history-retention-count` for backfill range.

2opremio:
@bartekn thoughts?

sreuland:
> Can't we piggyback on the retention count cmdline flag?
yes, definitely, updated description to be clear about that usage.

sreuland:
@2opremio , @bartekn , wondering if could remove the `backfill_queue` aspect just to explore simplicity, the new backfill process invokes the ingestion engine regularly on scheduled intervals, triggering the latest filter rules to be executed by processors with range based on `--history-retention-count`. The end result would be idempotent, as ingestion processors just upsert to history, correct? 

the `backfill_queue` increases in complexity with the upcoming transitive rules, like 'all trustlines from account x' or 'sponsored accounts of account x', where the filters would need to compute effective list dynamically and then convey that into the queue, whereas if we just let the ingest processors run the filters, then by default effective scope of the filters is exercised by default through the existing ingest->processor->`filter.FilterTransaction()` path.

2opremio:
> The end result would be idempotent, as ingestion processors just upsert to history, correct?

Yes it would but I am not sure we can afford that (ingestion-wise).

BTW, we may also need to think about garbage collection (what if a filter is updated to not include certain accounts anymore, should we care about the data left behind?)

2opremio:
> Note that (1) needs to happen even when filtering isn't enabled, since, otherwise, backfilling won't work if filtering is enabled later on.

Actually, it doesn't need to be enabled when filtering is disabled, because then everything is reingested and thus, there is nothing to backfill there.

sreuland:
[...]

Issue #4224, opened 2022-02-09
Ingestion Filtering, advanced rule, include activity of assets issued by configured accounts


sreuland:
### What problem does your feature solve?
Capturing asset oriented tx/ops data during filtering which is normally one-hop away and not dynamically captured from normal whitelist account filter rule.

### What would you like to see?
As a service provider with my own Horizon installation, I want to configure a new filter rule to trigger tracking transactions and operations related to any assets issued by specific Accounts.

I want to configure which account ids trigger the ingestion filter of my Horizon installation to retain transaction and operational info related to any assets issued by the account.

I don't want my horizon historical database to contain more data related to other assets not related, so, I expect to see my horizon db size on disk still remain relatively small with a slower growth rate compared to no-filtering(i.e. data for entire ledger was saved)

Acceptance Criteria:

- Horizon provides an Asset Filter rule configuration service interface to update the list of asset issuer accounts that should be used for determining whether filter includes tx's/ops related to an issued asset of the account.
- All Horizon ingestion processes(current-forward/reingest) can apply this filtration rule and result in only persisting historical data that passes the filter.
- The asset filter for this new rule, effectively monitors the state of issued assets by the configured accounts, and filter detects when an account has issued a new asset, at which point it includes that new Asset in the normal Asset whitelist ruleset for Asset filter to now enforce.

### Pre-requisites:
This feature is dependent on #4221 and #4222 being implemented first to establish the initial filtration framework within ingestion.

PR #4277, opened 2022-03-14
services/horizon, clients/horizonclient: Allow filtering ingested transactions by account or asset.

### LINKED ISSUE #4221: Ingestion Filtering by Account
Horizon ingestion of full ledger space from core can take up massive amounts of disk space on db.

As a service provider using my own installation of Horizon, I want to track transactions and operations on my customer accounts that have been onboarded to my service. 

I want to configure which accounts trigger related transaction and operation info  to be ingested by my Horizon installation and persisted to my horizon historical database.

I don't want my horizon historical database to contain data related to other accounts, so, I expect to see my horizon db size on disk remain relatively small with a slower growth rate compared to no-filtering(i.e. data for entire ledger was saved)

Acceptance Criteria:

- Horizon provides an Account Filter rule configuration API endpoint on the secure Horizon Admin port to perform CRUD on the account whitelist rule model.
- Design the account whitelist rule model, likely a list/array of strings representing whitelisted account ids.
- All Horizon ingestion processes(current-forward/reingest) run filtration, apply the account whitelist rules and result in only persisting historical data that passes the filter.
- The account filter effectively applies latest config rules to current and ongoing ingestion activity which occurs after a filter rule is updated.
- The Horizon Account Filter feature can be configured as a toggle/switch, enabled or disabled.
- The Horizon Account Filter feature should be available as early/alpha drop builds to some willing early-adopter partners for trial usage. 
- Need to choose feature branch or mainline development(i.e. merged to master often).

### LINKED ISSUE #4222: Ingestion Filtering by Asset
Horizon ingestion of full ledger space from core can take up massive amounts of disk space on db.

As a service provider with my own Horizon installation, I want to track transactions and operations related to assets I issue. I want to perform reconciliation against my Horizon DB for assets I issue.

I want to configure which assets trigger related transaction and operational info to be ingested by my Horizon installation and persisted to my horizon historical database.

I don't want my horizon historical database to contain data related to other assets, so, I expect to see my horizon db size on disk remain relatively small with a slower growth rate compared to no-filtering(i.e. data for entire ledger was saved)

Acceptance Criteria:

- Horizon provides an Asset Filter rule configuration API endpoint on the secure Horizon Admin port to perfrom CRUD on the list of whitelisted canonical asset codes of `<asset_code>:<asset_isssuer>`
- The Asset filter rule configuration interface accepts a list of canonical asset codes `<asset_code>:<asset_isssuer>` to add or remove to whitelist.
- All Horizon ingestion processes(current-forward/reingest) run filtration and apply the whitelist asset rules and result in only persisting historical data that passes the filter.
- The asset filter effectively applies latest config rules to current and ongoing ingestion activity which occurs after a filter rule is updated.
- The Horizon Asset Filter feature can be configured as a toggle/switch, enabled or disabled.
- The Horizon Asset Filter feature should be available as early/alpha drop builds to some willing early-adopter partners for trial usage. 
- Need to choose feature branch or mainline development(i.e. merged to master often).

sreuland:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

* [] I've updated any docs ([developer docs](https://developers.stellar.org/api/), `.md`

### Release planning

</details>

### What

New Ingestion Filters Feature. Provide the ability to select which ledger transactions are accepted at ingestion time to be stored on horizon's historical database. 

Users manage filters through Admin API and the historical ingestion will check the filter config rules and only persist the ledger transactions that pass the filter rules. Initially, two filters and corresponding rules are possible:

  'whitelist by account id' ([4221](https://github.com/stellar/go/issues/4221))
  'whitelist by canonical asset id' ([4222](https://github.com/stellar/go/issues/4222))

The filters and their configuration are optional feature in Horizon and must be enabled with horizon command line parameter `admin-port=4200` and `enable-ingestion-filtering=true`. Once set, filter configurations and their rules are initially empty and the filters are disabled by default. To configure, refer to the [Admin API Docs ](services/horizon/internal/httpx/static/admin_oapi.yml)which are published on the Admin Port at http://localhost:<admin_port>/, follow details and examples for usage of `/ingestion/filters/account` and `/ingestion/filters/asset.`

### Why

[Horizon Ingestion Filtering - Product Feature Roadmap](https://stellarorg.atlassian.net/wiki/spaces/HOR/pages/4281696257/Ingestion+Filtering)

[Horizon Ingestion Filtering - Tech Design](https://stellarorg.atlassian.net/wiki/spaces/HOR/pages/4267671599/Ingestion+Filtering+for+Scale)

Closes #4222 
Closes #4221

### Known limitations

Filters accept or reject data at the Transactions level, not per operation. 
Filters are limited to whitelist rules for now, more advanced filter rules are yet to be implemented. 
* [Backfill Strategy](https://github.com/stellar/go/issues/4267)
* [Include activity of assets issued by a whitelisted account](https://github.com/stellar/go/issues/4224)
* [Include activity of accounts by trustline to a whitelisted asset](https://github.com/stellar/go/issues/4224)


services/horizon/internal/ingest/processors/change_processors.go
```diff
@@ -29,6 +34,20 @@ func StreamLedgerTransactions(
...
+			return errors.Wrapf(
+				err,
+				"could not filter transaction %v",
+				tx.Index,
+			)
+		}
+		if !include {
+			// TODO, may be worthwhile to emit new metrics here for 'Filtered Transaction Count' ?
+			log.Debugf("Filters did not find match, skipping tx hash %v", tx.Result.TransactionHash)
+			continue
```
sreuland, line 48: @bartekn @2opremio 

we discovered when filtering is enabled and triggered in ingestion for a tx that was submitted to the same horizon instance, it causes a side effect where a timeout now occurs on txsub pending submission synchronization [here](https://github.com/stellar/go/blob/filtering-feature/services/horizon/internal/txsub/system.go#L349) due to the absence of a history_transactions row when ingestion filtering drops it, as a result `OpenSubmissionList.Finish()` never gets invoked for the tx hash and it results in timeout err back to txsub client.

trying to think of alternatives:

1. insert sparse populated row in history_transactions during ingestion for any filtered tx's
2. invoke filters on each tx during[ txsub.Submit() here](https://github.com/stellar/go/blob/filtering-feature/services/horizon/internal/txsub/system.go#L160), if match, create a completed(done) response channel with populated Result object and return immediate, skip `OpenSubmissionsList.Add()`

2opremio: I don’t have a solution yet. But, does that mean that Horizon’s submission endpoint only works if Horizon is attached to a DB being ingested to?

bartekn: Yes, this is true that txsub only works when connected to a DB which is updated by ingesting instance. The alternative which we removed some time ago was to update it by using Stellar-Core DB but we really wanted to remove this dependency because of many reasons (all connected to why Captive-Core exists).

Regarding solution to filtering I have some ideas:
* Update `history_transactions` with a new `filtered` field. All `filtered=true` are not visible via the API but txsub can still access them. Then we have a background worker that removed all filtered txs every X minutes.
* Same as above but create a much simpler table with just tx hash and tx results (all txsub needs, correct?) and remove all the rows older than X minutes every once in a while.
* Add a new processor to ingesting pipeline that pushes all txs and their results to a queue which is later pulled by txsub instances. This requires some extra effort if we need multiple txsub instances, for example: instance A is keeping the HTTP connection but instance B pulled the given tx from the queue.

sreuland: Hello @bartekn , thanks for ideas, we were mulling about an idea that is very similar to your second bullet you mentioned, here is sequence diagram walking that design pattern a bit further into existing txsub system, the 'alt' blocks are the new suggestions to use a `txsub_requests` table:
txsubfilters

does this look close to what you were thinking?

2opremio: I was thinking that we can switch it around to minimize DB access:

1. Whenever a tx submission is made, the submission system creates an entry in the db (e.g. DB row from from a dedicated table, let's call it txsub_response) indicating it is waiting for a tx response. After that, it will start pooling for an answer for certain time.
2. Whenever the online ingestion system sees a transaction for which a response is waited for, it will inject the response in tx_response table.
4. The tx_responses rows is TTLd (to be defined how).

sreuland: @2opremio , ok, sounds good, captured that proposed flow in a bit more detail, to check if any snags:

txsub_response

looks like it should work, the `alt: NEW` sections highlight expected areas of change.

sreuland: I created branch off this one for dev work on new tx response, gonna start in here:
https://github.com/stellar/go/tree/filtering-feature-tx-response

2opremio: I implemented it in the current branch, but it still needs testing and fixing a few TODOs I left.

services/horizon/internal/db2/schema/migrations/55_filter_rules.sql
```diff
@@ -0,0 +1,29 @@
...
+    last_modified bigint NOT NULL
+);
+
+-- insert the default disabled state for each supported filter implementation
+INSERT INTO account_filter_rules VALUES (false, '{}', 0);
+INSERT INTO asset_filter_rules VALUES (false, '{}', 0);
+
+CREATE TABLE txsub_results (
+    transaction_hash varchar(64) NOT NULL UNIQUE,
+    tx_result        text, -- serialized history.Transaction
```
sreuland, line 21: what if we try using PG table inheritance to accomplish the same end result of embedding the history.Transaction record in txsub_results, with inheritance can reference the embedded column data directly in sql/results, rather than json marshaling, something like:

```
CREATE TABLE txsub_results (
    submitted_at timestamp NOT NULL DEFAULT NOW()
)  INHERITS (history_transactions)
```

2opremio: Note that the `ClosedAt` field from `history.Transaction` is taken from the ledgers table through a join.

Also, at a second thought, I am not sure whether replicating the schema plays to our advantage since we are only storing temporary data.

If we will need to book-keep the schema of `txsub_results` to mimic the one of `history_transactions` I think it may be better to simply keep a blob.

sreuland: Yes, the inheritance should automatically pull in changes from super/parent table

2opremio: Ah, then it's probably a better alternative

2opremio: Except I don't think we can use it, because when we initialize the rows we need everything to be null except the the hashes. However `history_transactions` has a lot of checks which disallows it (including `NOT NULL`).

clients/horizonclient/client.go
```diff
@@ -841,5 +844,74 @@ func (c *Client) PrevLiquidityPoolsPage(page hProtocol.LiquidityPoolsPage) (lp h
...
+	}
+	req, err := http.NewRequest(http.MethodPut, url, buf)
+	if err != nil {
+		return errors.Wrap(err, "error creating HTTP request")
+	}
+	req.Header.Add("Content-Type", "application/json")
+	return c.sendHTTPRequest(req, nil)
+}
+
+func (c *Client) AdminSetIngestionAccountFilter(filter hProtocol.AccountFilterConfig) error {
```
bartekn, line 898: NIT: I think we should have a separate struct for admin calls, like `AdminClient`.

clients/horizonclient/internal.go
```diff
@@ -32,7 +32,10 @@ func decodeResponse(resp *http.Response, object interface{}, hc *Client) (err er
 		}
 		return horizonError
 	}
-
+	if object == nil {
```
bartekn, line 35: Looks like we should be able to move it to the top of this method.

services/horizon/internal/actions/filter_rules.go
```diff
@@ -0,0 +1,170 @@
...
+		problem.Render(r.Context(), w, err)
+		return
+	}
+
+	config, err := historyQ.GetAssetFilterConfig(r.Context())
+
+	if historyQ.NoRows(err) {
+		err = problem.NotFound
+	}
+
```
bartekn, line 28: NIT
```suggestion
```

`sql.ErrNoRows` in Horizon is a registered error so `problem.Render` should handle it.

services/horizon/internal/actions/filter_rules.go
```diff
@@ -0,0 +1,170 @@
...
+		problem.Render(r.Context(), w, err)
+	}
+}
+
+func (handler FilterConfigHandler) assetFilterResource(r *http.Request) (hProtocol.AssetFilterConfig, error) {
+	var filterRequest hProtocol.AssetFilterConfig
+	dec := json.NewDecoder(r.Body)
+	if err := dec.Decode(&filterRequest); err != nil {
+		p := problem.BadRequest
+		p.Extras = map[string]interface{}{
```
bartekn, line 135: NIT: Use `problem.NewProblemWithInvalidField`.

services/horizon/internal/actions/submit_transaction.go
```diff
@@ -50,6 +47,13 @@ func extractEnvelopeInfo(raw string, passphrase string) (envelopeInfo, error) {
 		return result, err
 	}
 	result.hash = hex.EncodeToString(hash[:])
+	if result.parsed.IsFeeBump() {
```
bartekn, line 50: Any reason for adding innerhash here?? I think that after this change the txsub system may return innertx result for a fee bump transaction.

services/horizon/internal/txsub/system.go
```diff
@@ -339,20 +347,7 @@ func (sys *System) Tick(ctx context.Context) {
...
-		}
-
-		// In Tick we only check txs in a queue so those which did not have results before Tick
-		// so we check for them in the last 5 mins of ledgers: 60.
-		sinceLedgerSeq := int32(latestLedger) - 60
-		if sinceLedgerSeq < 0 {
-			sinceLedgerSeq = 0
-		}
-
-		txs, err := db.TransactionsByHashesSinceLedger(ctx, pending, uint32(sinceLedgerSeq))
```
bartekn, line 355: I think we shouldn't remove this code. This change affects the current behaviour in which `POST /transactions` returns the result even for very old transactions. After removing this snippet it won't return it even for unfiltered Horizon deployments.

services/horizon/internal/ingest/processor_runner.go
```diff
@@ -150,11 +150,23 @@ func (s *ProcessorRunner) buildTransactionProcessor(
...
+	var f []processors.LedgerTransactionFilterer
+	if s.config.EnableIngestionFiltering {
+		f = append(f, s.filters.GetFilters(s.historyQ, s.ctx)...)
 	}
 
-	return newGroupTransactionFilterers(s.filters.GetFilters(s.historyQ, s.ctx))
+	return newGroupTransactionFilterers(f)
+}
+
+func (s *ProcessorRunner) buildUnfilteredProcessor(ledger xdr.LedgerHeaderHistoryEntry) *groupTransactionProcessors {
```
sreuland, line 161: thanks for the cleanup on unfiltered tx metrics timing provided by GroupTransactionProcessor, there were many ways to accomplish metrics and I was getting a bit in the weeds, but this looks optimal, and the other optimization to avoid unfiltered tx when Reingest mode looks good.

2opremio: No problem!

services/horizon/internal/actions/submit_transaction.go
2opremio: Because you need to be able to look results up by inner hash.

2opremio: For reference, see how it's done in the history transactions table.

Or are you suggesting we should fill in the inner hash when setting the result?

jcx120:
A few questions on usability for this first cut of the feature MVP (@sreuland):

1. When whitelisting by an Asset, will the data include Accounts (fully formed, beyond the account ID) that hold the asset? (and conversely, if filtering by Account, will it include data for Assets held by whitelist Accounts). Example case: we have a customer that only cares about 1 asset (and customers that interact with it), will they have to whitelist explicitly both the 1 asset and all accounts
2. In the above case where (1 Asset whitelisted), and they have to manually whitelist a list of Accounts that hold it (and maintain that on a daily basis), what is the lag between adding the account to the whitelist and time that it is reflected in the database? IE: we should instruct users to expect an X hour lag between when they add a new entity to the whitelist and DB being populated, and therefore they should "onboard" accounts x+y time before it is operational.

2opremio:
1. whitelisting an asset implies that if the asset appears in a transaction (in any way). The transaction will be considered for ingestion (filtered in). Any operations in filtered-in transactions involving a given account will be filtered it.

(the same principle extends to whitelisted accounts)

2. Right now there is a REST API for configuring the whitelists. Its effect is immediate. However, note that backfilling isn't implemented and thus configuration changes only impact transactions yet to be ingested.

2opremio:
Regarding (2) @jacekn is indicating us that operators may prefer to use config files. We may implement this (in addition or replacing the REST API). However, that comes with some complications:

1. How to make changes in the config file be accounted for inmediately:
   * Restarting Horizon is probably too big of a requirement
   * Sending SIGHUP (which is how these sort of thing is typically implemented) comes with its own complications.

2. Distribution of the config file across Horizon instances. There may be multiple Horizon ingestion instances. Making sure that the configuration file is:
   * identical in all ingestion instances
   * changes are atomic
   may be difficult or impossible.

clients/horizonclient/client.go
sreuland: yes, pulling this up further, will move the `Admin*` methods over to similar new `AdminClientInterface`

jacekn:
> Regarding (2) @jacekn is indicating us that operators may prefer to use config files. We may implement this (in addition or replacing the REST API). However, that comes with some complications:
> 
>     1. How to make changes in the config file be accounted for inmediately:
>        
>        * Restarting Horizon is probably too big of a requirement
>        * Sending SIGHUP (which is how these sort of thing is typically implemented) comes with its own complications.

What are the complications of handling SIGHUP (or SIGUSR1)? Signals are commonly used by daemons and support for them is very good across different distros.
For k8s workloads I saw some projects add support for a special HTTP "config reload" endpoint that we could run on the admin port if needed

>     2. Distribution of the config file across Horizon instances. There may be multiple Horizon ingestion instances. Making sure that the configuration file is:
>        
>        * identical in all ingestion instances

This problem needs to be solved for non-filtering config options too. For example operators need to ensure `INGEST_FAILED_TRANSACTIONS` or `ALLOW_EMPTY_LEDGER_DATA_RESPONSES` are consistent.
There are configuration management tools that can help with that for larger environments.

>        * changes are atomic

Good point. What's the worst case if config is slightly out of sync for a ledger or two? There are workarounds that operators could implement to ensure we can't have a drift but maybe a short drift is acceptable?

>        may be difficult or impossible to implement.
>        Otherwise we may end up with an ingestion split-brain. So I am more and more inclined towards just keeping the REST API.

The REST API has some downsides too. One of them is that config management tools may not have native support for it. Without that support it may be difficult to describe "config as a code".

clients/horizonclient/client.go
sreuland: pushed change for breaking out Admin from Client in horizonclient:

https://github.com/stellar/go/pull/4277/commits/90a3352da1a145049940eb07e8082a035cb6822f

sreuland:
@jcx120 @jacekn @2opremio , Per the API/FlatFile config discussion, added that as design feature on the filtering product spec:

https://stellarorg.atlassian.net/wiki/spaces/HOR/pages/4281696257/Ingestion+Filtering#Config

the points here are good, maybe we can continue and finalize specs on the product sheet for visibility. We were having a [similar comment chat here](https://stellarorg.atlassian.net/wiki/spaces/HOR/pages/4281696257/Ingestion+Filtering?focusedCommentId=4314529806#comment-4314529806) on sidebar of this notion of clustered horizon and  flat file config, one naive approach is to document this as only configure one instance in a horizon cluster to have a config file, that instance digests the file(and any changes to it) and propagates it to db table, all other instances re-fetch the latest values from the table already per a cache implementation here(they check last updated time of the their in-memory config rules and the db and reload if needed), so they will pull the latest changes as soon as pushed.

sreuland:
@jcx120,  @2opremio per those questions on asset and account whitelist outcomes, yeah, @2opremio summed it up, I think that is where some early beta feedback might indicate if the current whitelist behavior accomplished here will suffice for the expectations @jcx120 mentioned, i.e. if assetA is whitelisted, then the only account history that is kept are operations referencing assetA. so the account history is just an incidental side effect, any operations from accountX on non asset related operations or asset ops but with other assetB/C/D, will not be kept in that case, if customer needs that depth on account also, then yes, they need to define accountX as whitelisted too for the current filter capabilities. we've e[armarked more advanced transitive-hop filter rules](https://stellarorg.atlassian.net/wiki/spaces/HOR/pages/4281696257/Ingestion+Filtering) for consideration.

services/horizon/internal/actions/filter_rules.go
sreuland: good tip, thx.

services/horizon/internal/actions/filter_rules.go
sreuland: nice utility, using it now.

services/horizon/internal/ingest/filters/account.go
```diff
@@ -0,0 +1,63 @@
...
+	RefreshAccountFilter(filterConfig *history.AccountFilterConfig) error
+}
+
+func NewAccountFilter() AccountFilter {
+	return &accountFilter{
+		whitelistedAccountsSet: map[string]struct{}{},
+	}
+}
+
+// TODO:(fons) this code should probably be generic for all filters
```
sreuland, line 28: This is kind of tip of the iceberg for filter configs, each filter can have it's own independent rule schema, we've chosen to vertically isolate these models through to the db where each filter has it's own table aligned to the model,  rather than coerce/transport as a common/generic blob at other layers, so, I don't think it's much gain for generalizing above that here, the filters package hardcodes reference to each available filter implementation that is loadable in `filters.GetFilters()` , and has specific routine per config data fetch for each filter .

services/horizon/internal/ingest/filters/asset.go
```diff
@@ -0,0 +1,134 @@
...
+			if pool, ok := operation.Body.ChangeTrustOp.Line.GetLiquidityPool(); ok {
+				if f.assetMatchedFilter(&pool.ConstantProduct.AssetA) || f.assetMatchedFilter(&pool.ConstantProduct.AssetB) {
+					allowed = true
+				}
+			} else {
+				asset := operation.Body.ChangeTrustOp.Line.ToAsset()
+				allowed = f.assetMatchedFilter(&asset)
+			}
+		case xdr.OperationTypeClaimClaimableBalance:
+			// TODO, try to get asset for claimable balance id
```
sreuland, line 74: @bartekn , I was looking for efficient way to derive the canonical asset id given just a claimable balance id here, do you have recommendations, it looks like have to do db round trip(s) right?

services/horizon/internal/actions/submit_transaction.go
bartekn: OK, I see where it's coming from now. Thanks!

services/horizon/internal/txsub/system.go
bartekn: Ping on this. Again I think it will change the behaviour of txsub even if someone isn't filtering.

services/horizon/internal/ingest/filters/asset.go
bartekn: It seems like the only way but it will not work properly when reingesting.

services/horizon/internal/ingest/filters/main.go
```diff
@@ -0,0 +1,74 @@
...
+
+	"github.com/stellar/go/services/horizon/internal/db2/history"
+	"github.com/stellar/go/services/horizon/internal/ingest/processors"
+	"github.com/stellar/go/support/log"
+)
+
+var (
+
+	// the filter config cache will be checked against latest from db at most once per each of this interval,
+	FilterConfigCheckIntervalSeconds int64 = 10
```
bartekn, line 15: Nit:
```suggestion
	FilterConfigCheckInterval time.Duration = 10 * time.Second
```

services/horizon/internal/ingest/filters/main.go
```diff
@@ -0,0 +1,74 @@
...
+		assetFilter:   NewAssetFilter(),
+		accountFilter: NewAccountFilter(),
+	}
+}
+
+// Provide list of the active filters. Optimize performance by caching the list, only
+// rebuild the list on expiration time interval. Method is NOT thread-safe.
+func (f *filtersCache) GetFilters(filterQ history.QFilter, ctx context.Context) []processors.LedgerTransactionFilterer {
+	// only attempt to refresh filter config cache state at configured interval limit
+	if time.Now().Unix() < (f.lastFilterConfigCheckUnixEpoch + FilterConfigCheckIntervalSeconds) {
```
bartekn, line 45: If you change `FilterConfigCheckIntervalSeconds` to `time.Duration` this condition will be easier to read.

bartekn (approved):
Looks good! The only blocker from me is: https://github.com/stellar/go/pull/4277/files#r856086643. Added some NIT comments if you want to include them.

services/horizon/internal/txsub/system.go
sreuland: @bartekn , I added it back in, and a bit more test coverage to assert on this outcome. Is the 5 minute lag aspect mentioned in docs, or just been baked in from early on?

https://github.com/stellar/go/pull/4277/commits/b37dc444dcc093cb71cf683413d0be8c6d68ef64

services/horizon/internal/ingest/filters/asset.go
sreuland: I think will just remove these inclusion cases then, i.e. will not attempt to match up on asset codes for these claimable balance ops, maybe if it comes up later can spend more time investing into this edge.

sreuland: https://github.com/stellar/go/pull/4277/commits/d476a4e864c77a96b29184090afacb6a1b1917c8

clients/horizonclient/admin_client.go
```diff
@@ -0,0 +1,112 @@
...
+
+	if resp, err := c.HTTP.Do(req.WithContext(ctx)); err != nil {
+		return err
+	} else {
+		return decodeResponse(resp, a, req.URL.String(), c.clock)
+	}
+}
+
+func (c *AdminClient) getIngestionFiltersURL(name string) (string, error) {
+	baseURL, err := url.Parse("http://localhost")
```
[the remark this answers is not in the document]
sreuland: included overrided settings on AdminClient - https://github.com/stellar/go/pull/4277/commits/b693b95795c3fe8a6d95ef53db16cbe643d552bd

protocols/horizon/main.go
```diff
@@ -843,3 +843,55 @@ type LiquidityPoolReserve struct {
...
+		return err
+	}
+
+	if config.Whitelist == nil {
+		return errors.New("missing required whitelist")
+	}
+
+	if config.Enabled == nil {
+		return errors.New("missing required enabled")
+	}
```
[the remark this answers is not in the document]
sreuland: this is also used on inbound http PUT payloads from remote clients:

services/horizon/internal/actions/filter_rules.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+	return filterRequest, nil
+}
+
+func (handler FilterConfigHandler) accountFilterResource(r *http.Request) (hProtocol.AccountFilterConfig, error) {
+	var filterRequest hProtocol.AccountFilterConfig
+	dec := json.NewDecoder(r.Body)
+	if err := dec.Decode(&filterRequest); err != nil {
+		p := problem.BadRequest
+		p.Extras = map[string]interface{}{
```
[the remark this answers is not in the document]
sreuland: got it thanks,

services/horizon/internal/db2/history/txsub_result.go
```diff
@@ -0,0 +1,160 @@
...
+	union := byHash.Suffix("UNION ALL "+byInnerHashString, args...)
+	var result []string
+	err = q.Select(ctx, &result, union)
+	if err != nil {
+		return nil, err
+	}
+
+	txs := make([]Transaction, len(result))
+	for i := 0; i < len(result); i++ {
+		err = json.Unmarshal([]byte(result[i]), &txs[i])
```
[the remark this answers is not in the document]
sreuland: nice catch, thanks!

https://github.com/stellar/go/pull/4277/commits/54ddeac51b3ea9ae7dadb944bb7421fc28811cfa

services/horizon/internal/ingest/filters/account.go
2opremio: Makes sense. Can you remove the TODO then? Thanks!

sreuland: yes, done, thanks.

services/horizon/internal/ingest/filters/asset.go
```diff
@@ -0,0 +1,134 @@
...
+			if f.assetMatchedFilter(&operation.Body.PaymentOp.Asset) {
+				allowed = true
+			}
+		case xdr.OperationTypePathPaymentStrictReceive:
+			if f.assetMatchedFilter(&operation.Body.PathPaymentStrictReceiveOp.DestAsset) || f.assetMatchedFilter(&operation.Body.PathPaymentStrictReceiveOp.SendAsset) {
+				allowed = true
+			}
+		case xdr.OperationTypePathPaymentStrictSend:
+			if f.assetMatchedFilter(&operation.Body.PathPaymentStrictSendOp.DestAsset) || f.assetMatchedFilter(&operation.Body.PathPaymentStrictSendOp.SendAsset) {
+				allowed = true
```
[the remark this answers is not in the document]
sreuland: yes, good suggestion,  earlier, there was some common filtering rules being applied after this initial check, however that has since been removed and can follow this, thanks.

https://github.com/stellar/go/pull/4277/commits/05e72117b57808e35e6f57f034a052a8381c0bf8

services/horizon/internal/ingest/filters/main.go
sreuland: yes, followed that suggestion, thx.

https://github.com/stellar/go/pull/4277/commits/05e72117b57808e35e6f57f034a052a8381c0bf8

services/horizon/internal/ingest/filters/asset.go
```diff
@@ -0,0 +1,124 @@
...
+			}
+		case xdr.OperationTypePathPaymentStrictReceive:
+			if f.assetMatchedFilter(&operation.Body.PathPaymentStrictReceiveOp.DestAsset) || f.assetMatchedFilter(&operation.Body.PathPaymentStrictReceiveOp.SendAsset) {
+				return true, nil
+			}
+		case xdr.OperationTypePathPaymentStrictSend:
+			if f.assetMatchedFilter(&operation.Body.PathPaymentStrictSendOp.DestAsset) || f.assetMatchedFilter(&operation.Body.PathPaymentStrictSendOp.SendAsset) {
+				return true, nil
+			}
+		}
```
Shaptic, line 106: Possible idea for restructuring this: It feels a little brittle to need to have these complex conditionals for each operation. What about something like:

```go

candidateAssets := []xdr.Asset{}

for _, operation := range tx.Tx.Operations {
		switch operation.Body.Type {
		case xdr.OperationTypeChangeTrust:
			if pool, ok := operation.Body.ChangeTrustOp.Line.GetLiquidityPool(); ok {
				candidateAssets = append(candidateAssets, pool.ConstantProduct.AssetA, pool.ConstantProduct.AssetB)
			} else {
				candidateAssets = append(candidateAssets, operation.Body.ChangeTrustOp.Line.ToAsset())
			}
		// other cases...
	}
}

for _, asset := range candidateAssets {
	if f.assetMatchedFilter(asset) {
		return true, nil
	}
}

return false, nil
```

Shaptic (commented):
There's a lot to digest here so this is a bit of a superficial review, but hopefully will have more thoughts/suggestions/questions after a closer look.

clients/horizonclient/main.go
```diff
@@ -149,11 +149,34 @@ type Client struct {
...
+	AdminHost string
+
+	// HTTP client to make requests with
+	HTTP HTTP
+
+	// max client wait time for response
+	horizonTimeout time.Duration
+
+	// clock is a Clock returning the current time.
+	clock *clock.Clock
```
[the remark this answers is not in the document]
sreuland: yes, I pulled up AdminClient from Client, followed it's pattern of usage, I think can remove this, it looks like `setCurrentServerTime` asks for it, but all it does is a `clock.Now()` seems like it doesn't need a clock parameter to accomplish system time?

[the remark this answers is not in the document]
Shaptic: Ah that makes sense - yeah imo if we're just doing `.Now()`s then unless there's a real need for a specific clock, `time.Now()` should achieve the same goal?

protocols/horizon/main.go
```diff
@@ -843,3 +843,55 @@ type LiquidityPoolReserve struct {
 	Asset  string `json:"asset"`
 	Amount string `json:"amount"`
 }
+
+type AssetFilterConfig struct {
+	Whitelist    []string `json:"whitelist"`
+	Enabled      *bool    `json:"enabled"`
```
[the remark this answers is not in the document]
sreuland: avoid default zero value(false) if absent on json unmarshal, allows the PUT action handler to detect presence.

services/horizon/internal/ingest/filters/account.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+func NewAccountFilter() AccountFilter {
+	return &accountFilter{
+		whitelistedAccountsSet: map[string]struct{}{},
+	}
+}
+
+func (filter *accountFilter) RefreshAccountFilter(filterConfig *history.AccountFilterConfig) error {
+	// only need to re-initialize the filter config state(rules) if it's cached version(in  memory)
```
Shaptic, line 29: super nit :smile: 
```suggestion
	// only need to re-initialize the filter config state(rules) if its cached version(in  memory)
```

services/horizon/internal/db2/schema/migrations/55_filter_rules.sql
```diff
@@ -0,0 +1,30 @@
...
+    whitelist varchar[] NOT NULL,
+    last_modified bigint NOT NULL
+);
+
+-- insert the default disabled state for each supported filter implementation
+INSERT INTO account_filter_rules VALUES (false, '{}', 0);
+INSERT INTO asset_filter_rules VALUES (false, '{}', 0);
+
+CREATE TABLE txsub_results (
+    transaction_hash       varchar(64) NOT NULL UNIQUE,
```
Shaptic, line 20: this is related to https://github.com/stellar/go/pull/4277#discussion_r861310208: we might get into trouble if this assumption ever breaks, even by accident

Shaptic (commented):
more thoughts, overall I think the API is super clean and it will be really easy to extend w/ other filters if/when the time comes!

services/horizon/internal/ingest/filters/account.go
```diff
@@ -0,0 +1,61 @@
...
+import (
+	"context"
+
+	"github.com/stellar/go/ingest"
+	"github.com/stellar/go/services/horizon/internal/db2/history"
+	"github.com/stellar/go/services/horizon/internal/ingest/processors"
+)
+
+type accountFilter struct {
+	whitelistedAccountsSet map[string]struct{}
```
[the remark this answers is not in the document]
2opremio: For good or for bad I think using `struct{}` maps in Go has become idiomatic.

I admit I resisted getting used to it. This will probably get resolved when we start incorporating generics, since this is the perfect usecase.

Thus, I see little value in using interim abstractions.

clients/horizonclient/admin_client.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+func (c *AdminClient) sendHTTPRequest(req *http.Request, a interface{}) error {
+	if c.HTTP == nil {
+		c.HTTP = http.DefaultClient
+	}
+
+	if c.horizonTimeout == 0 {
+		c.horizonTimeout = HorizonTimeout
+	}
```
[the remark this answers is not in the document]
sreuland: yes, refactoring that now

services/horizon/internal/ingest/filters/account.go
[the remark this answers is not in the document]
Shaptic: Fair enough, seems like we have a few candidates for a "add generics" refactor pass on the codebase (min/max also come to mind).

clients/horizonclient/admin_client.go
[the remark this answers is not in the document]
sreuland: https://github.com/stellar/go/pull/4277/commits/40155508e859045cdc28511750532c6fefc5e565

clients/horizonclient/main.go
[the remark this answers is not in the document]
sreuland: looking into further, the `clock.Now()` is relative to the underlying source in the clock, if the clock is zero valued, then it defaults to local system time for Now(), but I found a [test that artificially sets the clock source in Client to something else](https://github.com/stellar/go/blob/master/clients/horizonclient/main_test.go#L1451), so, I will not disturb this in Client, I have removed it from AdminClient, since there is no usage of it yet.

clients/horizonclient/client.go
```diff
@@ -270,6 +270,9 @@ func (c *Client) setDefaultClient() {
 // fixHorizonURL strips all slashes(/) at the end of HorizonURL if any, then adds a single slash
 func (c *Client) fixHorizonURL() string {
 	c.fixHorizonURLOnce.Do(func() {
+		// TODO: we shouldn't happily edit data provided by the user,
```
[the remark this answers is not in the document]
2opremio: It's unrelated code, but I found it so ugly that I added a TODO

clients/horizonclient/admin_client.go
```diff
@@ -0,0 +1,116 @@
...
+	adminHost := baseURL.Hostname()
+	if len(c.AdminHost) > 0 {
+		adminHost = c.AdminHost
+	}
+	baseURL.Host = fmt.Sprintf("%s:%d", adminHost, adminPort)
+	return baseURL.String(), nil
+}
+
+func (c *AdminClient) AdminGetIngestionAssetFilter() (hProtocol.AssetFilterConfig, error) {
```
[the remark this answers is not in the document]
sreuland: https://github.com/stellar/go/pull/4277/commits/5e8a912bc3d1507b6158317a4d0513981f0003f7

services/horizon/internal/db2/history/txsub_result.go
```diff
@@ -0,0 +1,160 @@
...
+		return Transaction{}, err
+	}
+	switch len(transactions) {
+	case 0:
+		return Transaction{}, sql.ErrNoRows
+	case 1:
+		return transactions[0], nil
+	default:
+		return Transaction{}, fmt.Errorf("unexpected result size > 1 (%d)", len(transactions))
+	}
```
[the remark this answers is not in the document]
sreuland: duplicate hashes aren't allowed in the txsub_results table due to UNIQUE constraint, so this multi-row response case is not possible, I will remove.

services/horizon/internal/db2/history/transaction.go
```diff
@@ -28,8 +28,6 @@ func (q *Q) TransactionByHash(ctx context.Context, dest interface{}, hash string
 	return q.Get(ctx, dest, union)
 }
 
-// TransactionsByHashesSinceLedger fetches transactions from the `history_transactions`
-// table which match the given hash since the given ledger sequence (for perf reasons).
```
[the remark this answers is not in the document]
sreuland: oversight, at one point we had removed this method, when restored, missed this, moved back in. 

https://github.com/stellar/go/pull/4277/commits/0ae77dce7086db19434b9a3d8d4e1ace5e44c22c

services/horizon/internal/db2/history/txsub_result.go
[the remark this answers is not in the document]
sreuland: removed here https://github.com/stellar/go/pull/4277/commits/0ae77dce7086db19434b9a3d8d4e1ace5e44c22c

services/horizon/internal/db2/history/txsub_result.go
```diff
@@ -0,0 +1,163 @@
...
+			return totalRowsAffected, err
+		}
+		totalRowsAffected += affected
+		transactions = transactions[batchSize:]
+	}
+	return totalRowsAffected, nil
+}
+
+func (q *Q) setTxSubmissionResults(ctx context.Context, transactions []ingest.LedgerTransaction, sequence uint32, ledgerClosetime time.Time) (int64, error) {
+	caseStmt := sq.Case(txSubResultHashColumnName)
```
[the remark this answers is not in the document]
sreuland: https://github.com/stellar/go/pull/4277/commits/e80e5f32056f8a8c5fbfbb08096d1d47fa035731

services/horizon/internal/ingest/filters/asset.go
sreuland: I'll refactor the ChangeTrust to it's own method. I was using guard clause pattern here, ideally to reduce mental effort on following, you can see when it's over, rather than flow into a second 'post' loop. i'm not partial to either, if you feel it's better with the post loop, I'll do that, but wanted to explain the intention of current way.

sreuland: I tuned this up a bit more after adding the V0 in also - https://github.com/stellar/go/pull/4277/commits/7b77d7ba00a7aecc6ae08cde92ee0b40be097777

services/horizon/internal/ingest/filters/asset.go
```diff
@@ -0,0 +1,124 @@
...
+func (f *assetFilter) FilterTransaction(ctx context.Context, transaction ingest.LedgerTransaction) (bool, error) {
+	// filtering is disabled if the whitelist is empty for now as that is the only filter rule
+	if len(f.canonicalAssetsLookup) < 1 || !f.enabled {
+		return true, nil
+	}
+
+	tx, v1Exists := transaction.Envelope.GetV1()
+	if !v1Exists {
+		return true, nil
+	}
```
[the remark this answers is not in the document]
sreuland: yes, good point, added support, this will help reingest older ranges where V0's could be, is possible to submit a V0 now on current protocol level? thanks!: 

https://github.com/stellar/go/pull/4277/commits/7b77d7ba00a7aecc6ae08cde92ee0b40be097777

services/horizon/internal/httpx/static/admin_oapi.yml
```diff
@@ -0,0 +1,203 @@
+openapi: 3.0.3
```
sreuland, line 1: Screen Shot 2022-05-06 at 4 14 58 PM

services/horizon/internal/db2/schema/migrations/55_filter_rules.sql
sreuland: Ok, worst case, just the tx's with the same hash would be hosed, but then purged off this table after a couple minutes anyways, and we remove this if dups are legit, @2opremio , is this acceptable?

Shaptic: well I'm not sure exactly what happens if this constraint is violated, but I would imagine ingestion halts because it will repeatedly error on this ledger. since we don't have this constraint on history_transactions it will be safer to also omit it here

services/horizon/internal/ingest/filters/asset.go
Shaptic: Oh I wasn't concerned so much w/ `ChangeTrust` portion, just used it as an example. 

I see what you mean about "you can see when it's over." From my perspective, it was more like "Are any of these `case`s doing something other than checking that assets match?" which meant I had to carefully read them all. 

Grouping the check at the bottom makes it clearer to me that the switch will not terminate and there is common logic in all of the branches. No strong opinion, though, just likewise want to explain motivation behind review comment.

services/horizon/internal/db2/schema/migrations/55_filter_rules.sql
2opremio: I don't see how this constraint can be violated. If two different transactions have the same hash we have much bigger problems than this constraint.

services/horizon/internal/flags.go
```diff
@@ -199,6 +199,14 @@ func Flags() (*Config, support.ConfigOptions) {
...
 			ConfigKey:   &config.EnableCaptiveCoreIngestion,
 		},
+		&support.ConfigOption{
+			Name:        "enable-ingestion-filtering",
+			OptType:     types.Bool,
+			FlagDefault: true,
+			Required:    false,
+			Usage:       "causes Horizon to enable Ingestion filtering and the ingestion admin HTTP endpoint at /ingestion/filter",
+			ConfigKey:   &config.EnableIngestionFiltering,
+		},
```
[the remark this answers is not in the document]
sreuland: I think we meant to have filtering feature disabled by default, so will change `FlagDefault: false` here, @2opremio , let me know if any objections to that?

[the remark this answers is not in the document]
2opremio: No objections.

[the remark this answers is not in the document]
2opremio: In fact I think it should be false by default, yeah

services/horizon/internal/db2/schema/migrations/55_filter_rules.sql
Shaptic: relevant thread, just for visibility: https://stellarfoundation.slack.com/archives/G01FUHTUQ8Z/p1647892985378699

sreuland:
> more thoughts, overall I think the API is super clean and it will be really easy to extend w/ other filters if/when the time comes!

@Shaptic , thanks for your review on this monster, I have incorporated your feedback.

Shaptic (approved):
:+1: hope testing goes well!

services/horizon/internal/db2/history/txsub_result.go
```diff
@@ -0,0 +1,163 @@
...
+	err = q.Select(ctx, &result, union)
+	if err != nil {
+		return nil, err
+	}
+
+	txs := make([]Transaction, len(result))
+	for i := 0; i < len(result); i++ {
+		err = json.Unmarshal([]byte(result[i]), &txs[i])
+		if err != nil {
+			return nil, err
```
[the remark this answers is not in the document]
sreuland: I think that would be inconsistent with expected convention? when responding with error, typically that means value responses are not valid, return the zero value, and caller shouldn't attempt to use values in this case.

2opremio:
@sreuland Thanks a lot, really good work!!! :)

Issue #4275, opened 2022-03-14
services/horizon: Support new CAP-21 account fields


Shaptic:
Breakdown and Rollout: [Horizon & Protocol 19](https://docs.google.com/document/d/1qZ6o14JBOZpyDG6rNEd8Bj-sCzpeZgyx7Vtl1SMKI28). See #4261 for the master issue.

----------------

### Background
[CAP-21](https://stellar.org/protocol/cap-21) adds a new account extension which adds two additional fields to an account entry. These fields represent when the account's sequence number was increased, both as a **time** and as a **ledger number**. This tracks the last time an account was either:
* the source account for an executed transaction, or
* the source account for a BumpSequence operation (regardless of whether it modified the seqno of the account).

Relevant XDR:

```xdr
struct AccountEntryExtensionV3
{
    // We can use this to add more fields, or because it is first, to
    // change AccountEntryExtensionV3 into a union.
    ExtensionPoint ext;

    // Ledger number at which `seqNum` took on its present value.
    uint32 seqLedger;

    // Time at which `seqNum` took on its present value.
    TimePoint seqTime;
};
```

### Affected Endpoints
* `GET /accounts`
* `GET /accounts/:id`

### API Changes

```diff
--- a/protocols/horizon/main.go
+++ b/protocols/horizon/main.go
@@ -43,6 +43,8 @@ type Account struct {
        ID                       string        `json:"id"`
        AccountID                string        `json:"account_id"`
        Sequence                 string        `json:"sequence"`
+       SequenceLedger           uint32        `json:"sequence_ledger,omitempty"`
+       SequenceTime             *time.Time    `json:"sequence_time,omitempty"`
        SubentryCount            int32         `json:"subentry_count"`
        InflationDestination     string        `json:"inflation_destination,omitempty"`
        HomeDomain               string        `json:"home_domain,omitempty"`
```

### Behavior
The new fields are empty/omitted until an account does something, then they’re filled on ledger ingestion. They should not be back-filled, as behavior for fields being zero/null is well-defined:

> If an account does not have an `AccountEntryExtensionV3` because it hasn't been upgraded yet, then it behaves as if `seqLedger` and `seqTime` are both 0.

erika-sdf:
Is it useful and possible to have the `SequenceLedger` field be defined as a `uint32` instead of a `int32`? I suggest this change for consistency since it looks like ledger numbers are `uint32` elsewhere (including in the proposed XDR for this issue).

```
[...]

PR #4279, opened 2022-03-14
all: Add Protocol 19 XDR and update StrKey to support Signed Payloads

### LINKED ISSUE #4278: txnbuild: Update StrKey implementation to support CAP-40 signed payloads

Shaptic:
This updates XDR definitions for Protocol 19, taken from [sisuresh/stellar-core#cap21-40-2](https://github.com/sisuresh/stellar-core/commits/cap21-40-2), and incorporates them into the codebase. This involves:

 - updating transaction conditions wherever they're built directly as XDR,
 - updating claimable balance predicate types (Int64 -> TimePoint or Duration)
 - integrating CAP-40 (signed payloads) into strkey (based on stellar/stellar-protocol#1014)
 - extending an `account_signers` column to support longer signer keys 
 - removing a state verifier test until #4275 is done

Closes #4278.


strkey/main.go
```diff
@@ -34,13 +34,19 @@ const (
...
-// maxPayloadSize is the maximum length of the payload for all versions.
-const maxPayloadSize = 40
+// maxPayloadSize is the maximum length of the payload for all versions. The
+// largest strkey is a signed payload: 32-byte public key + 4-byte payload
+// length + 64-byte payload
+const maxPayloadSize = 100
 
 // maxRawSize is the maximum length of a strkey in its raw form not encoded.
-const maxRawSize = 1 + maxPayloadSize + 2
+const maxRawSize = 1 + /* version byte */ maxPayloadSize + 2 /* checksum */
```
sreuland, line 49: might be good to capture these as protocol constants also for visibility:
```
const checksumSize = 2
const versionSize =1
```

strkey/signed_payload.go
```diff
@@ -0,0 +1,94 @@
...
+
+// Encode turns a signed payload structure into its StrKey equivalent.
+func (sp *SignedPayload) Encode() (string, error) {
+	signerBytes, err := Decode(VersionByteAccountID, sp.Signer)
+	if err != nil {
+		return "", errors.Wrap(err, "failed to encode signed payload")
+	}
+
+	pad := [4]byte{0, 0, 0, 0}
+	// Encoding has four parts:
```
sreuland, line 35: minor, could pull up some of hardcoding to constants, just for readability :
```
const publicKeyLength = 32
const payloadLength = 4
```

xdr/preconditions.go
```diff
@@ -0,0 +1,29 @@
+package xdr
+
+// UpgradePrecondition converts the old-style optional union into a new
+// (post-CAP-21) xdr.Preconditions structure. This lets you avoid mucking around
+// with types and transparently convert from nil/TimeBounds to the appropriate
+// Preconditions struct.
+func UpgradePrecondition(timebounds *TimeBounds) Preconditions {
```
sreuland, line 7: nice, :+1:

xdr/signer_key.go
```diff
@@ -101,10 +118,32 @@ func (skey *SignerKey) SetAddress(address string) error {
...
+		pubkey, innerErr := strkey.Decode(strkey.VersionByteAccountID, sp.Signer)
+		if innerErr != nil {
+			return innerErr
+		}
+
+		var signer Uint256
+		copy(signer[:], pubkey)
+		*skey, innerErr = NewSignerKey(keytype, SignerKeyEd25519SignedPayload{
+			Ed25519: signer,
+			Payload: sp.Payload, // do we need a buffer copy here?
```
sreuland, line 142: NewSignerKey would retain reference from current closure here that `sp` was created in, looks like additional copy shouldn't be necessary?

xdr/Stellar-overlay.x
```diff
@@ -93,7 +98,9 @@ enum MessageType
     HELLO = 13,
 
     SURVEY_REQUEST = 14,
-    SURVEY_RESPONSE = 15
+    SURVEY_RESPONSE = 15,
+
+    SEND_MORE = 16
```
sreuland, line 103: is `SEND_MORE` related to cap21 or cap40? asking to know in terms of java sdk, which has stellar-overlay.x in there, assuming will need to include this also.

sreuland (approved):
nice work @Shaptic !

services/horizon/internal/ingest/verify_test.go
```diff
@@ -35,9 +35,13 @@ func genAccount(tt *test.T, gen randxdr.Generator) xdr.LedgerEntryChange {
 			{randxdr.FieldMatches(regexp.MustCompile("created\\.data\\.account\\.signers.*weight")), randxdr.SetPositiveNum32},
 			{randxdr.FieldEquals("created.data.account.ext.v1.liabilities.selling"), randxdr.SetPositiveNum64},
 			{randxdr.FieldEquals("created.data.account.ext.v1.liabilities.buying"), randxdr.SetPositiveNum64},
+			{randxdr.FieldEquals("created.data.account.ext.v1.ext.v"), randxdr.SetU32(2)},
```
tamirms, line 38: I don't think we need this line

xdr/Stellar-overlay.x
Shaptic: it's not, but this came from stellar/stellar-core's current src/xdr on master... might be some in-progress work sneaking in? cc @sisuresh

leighmcculloch (commented):
👏🏻

strkey/signed_payload.go
```diff
@@ -0,0 +1,94 @@
...
+	//  - N-byte payload (where N <= 64)
+	//  - M-byte padding (where 0 <= M <= 4)
+	baseLength := 32 + 4 + len(sp.Payload)
+	padding := getNextMultiple(baseLength, 4) - baseLength
+	src := make([]byte, baseLength+padding)
+
+	copy(src[:32], signerBytes)                                     // signer
+	binary.BigEndian.PutUint32(src[32:36], uint32(len(sp.Payload))) // length
+	copy(src[36:], sp.Payload)                                      // payload
+	copy(src[36+len(sp.Payload):], pad[:padding])                   // padding
```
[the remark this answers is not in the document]
leighmcculloch: Note that the `xdr` package is dependent on the `strkey` package, and so you won't actually be able to call the xdr package from here because of that. We also don't really put any specific strkey encoding functions typically in the strkey package, it's more like the raw base. But I do like the idea of having a package with functions for more nicely encoding/decoding each type of strkey. Maybe we could make a new package `strkeys` as a subpkg of `strkey`, where `strkeys` is dependent on `xdr` and `strkey` and exposes the things you're adding here.

services/horizon/internal/db2/history/transaction_batch_insert_builder.go
```diff
@@ -135,10 +134,6 @@ func formatTimeBounds(transaction ingest.LedgerTransaction) TimeBounds {
 	}
 
 	maxTime := timeBounds.MaxTime
-	if maxTime > math.MaxInt64 {
```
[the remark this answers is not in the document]
Shaptic: Yep, this is from before that conversation iirc

services/horizon/internal/ingest/verify_test.go
Shaptic: Good point!

strkey/signed_payload.go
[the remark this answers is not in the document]
Shaptic: I'm a little confused: if I can't use `stellar/go/xdr` in strkey, then I can't marshal the type directly. I can, however, do a better job marshalling the opaque buffer! And do some cleaner byte packing. Done in 0ef0948f and b945315f; is that what you meant?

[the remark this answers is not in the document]
leighmcculloch: I was suggesting you move this new code into a `strkeys` package, so that the implementations of different strkeys is independent from the `strkey` base package. As a bonus this would allow you to import `xdr` into `strkeys`.

strkey/signed_payload_test.go
```diff
@@ -0,0 +1,103 @@
...
+		// Valid signed payload with an ed25519 public key and a 29-byte payload
+		// which becomes zero padded.
+		{
+			"0102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d",
+			"GA7QYNF7SOWQ3GLR2BGMZEHXAVIRZA4KVWLTJJFC7MGXUA74P7UJVSGZ",
+			"PA7QYNF7SOWQ3GLR2BGMZEHXAVIRZA4KVWLTJJFC7MGXUA74P7UJUAAAAAOQCAQDAQCQMBYIBEFAWDANBYHRAEISCMKBKFQXDAMRUGY4DUAAAAFGBU",
+		},
+	}
+
+	for _, testCase := range testCases {
```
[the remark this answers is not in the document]
Shaptic: Good call; done in 4196d39c.

strkey/signed_payload.go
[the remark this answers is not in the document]
Shaptic: Let's keep a refactor like that out of this PR. I've added an issue to track this: #4293.

xdr/preconditions.go
```diff
@@ -0,0 +1,29 @@
...
+	cond := Preconditions{Type: PreconditionTypePrecondNone}
+	if timebounds != nil {
+		cond.Type = PreconditionTypePrecondTime
+		cond.TimeBounds = timebounds
+	}
+	return cond
+}
+
+func GetTimebounds(tx *Transaction) *TimeBounds {
```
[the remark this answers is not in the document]
Shaptic: We didn't have an `xdr/`transaction.go, heh, but done in ebdecc000be38d3a0671437e125d105723083f08.

strkey/signed_payload.go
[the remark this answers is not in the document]
leighmcculloch: This wouldn't be a refactor of anything that already exists in the strkey package, I'm commenting only on the logic being added by this PR and suggesting you place it somewhere else so you can use the xdr module instead of writing the encoding/decoding yourself.

xdr/preconditions.go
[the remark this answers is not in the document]
Shaptic: Done in ebdecc000be38d3a0671437e125d105723083f08

strkey/signed_payload.go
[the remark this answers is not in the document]
Shaptic: The code is much cleaner after bd12a577 and aligns more with how `strkey/muxed_account.go` does it :+1:

strkey/signed_payload.go
Shaptic: incorporated in latest push :+1:

xdr/signer_key.go
```diff
@@ -40,6 +41,18 @@ func (skey *SignerKey) GetAddress() (string, error) {
...
+		sp := skey.MustEd25519SignedPayload()
+		publicKey, err := strkey.Encode(strkey.VersionByteAccountID, sp.Ed25519[:])
+		if err != nil {
+			return "", errors.Wrap(err, "failed to decode payload signer")
+		}
+		spWrapper, err := strkey.MakeSignedPayload(publicKey, sp.Payload)
+		if err != nil {
+			return "", errors.Wrap(err, "failed to create signed payload")
+		}
+		return spWrapper.Encode()
```
leighmcculloch, line 54: 💭 Something we might want to come back to and improve is these lines of code encode a G strkey, then decode it, then re-encode it in a P strkey. Strkey's functions have been optimized, but they have to allocate unfortunately. Let's not worry about this now, but I'm thinking a `strkey.NewEd25519SignedPayload` that accepts a `ed25519 byte[32]`. I can add that when testing Starlight if I find Starlights benchmarks are adversely affected.

Technically you could also just call the following too which wouldn't do so much encoding/decoding.
```suggestion
		raw, err = sp.MarshalBinary()
		if err != nil {
			return "", err
		}
```

xdr/transaction.go
```diff
@@ -0,0 +1,15 @@
...
+// Preconditions.
+func (tx *Transaction) TimeBounds() *TimeBounds {
+	switch tx.Cond.Type {
+	case PreconditionTypePrecondNone:
+		return nil
+	case PreconditionTypePrecondTime:
+		return tx.Cond.TimeBounds
+	case PreconditionTypePrecondV2:
+		return tx.Cond.V2.TimeBounds
+	}
```
leighmcculloch, line 13: ❗ We should have a panic in here for unrecognized XDR:
```suggestion
		return tx.Cond.V2.TimeBounds
 	default:
		panic("unsupported transaction type: " + e.Type.String())
	}
```

leighmcculloch (commented):
Nice, looking great. There's a couple things I think we should change or add (💡❗). Otherwise this looks great.

Once this is merged I'll experiment with the new strkey API a bit and see how well/poorly it performs in benchmarks to see if we need to prefer the full XDR encoding or not.

services/horizon/internal/ingest/processors/signers_processor.go
```diff
@@ -138,7 +139,8 @@ func (p *SignersProcessor) Commit(ctx context.Context) error {
 					sponsor,
 				)
 				if err != nil {
-					return errors.Wrap(err, "Error inserting a signer")
+					return errors.Wrap(err,
+						fmt.Sprintf("Error inserting a signer (%s)", signer))
```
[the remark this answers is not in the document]
Shaptic: awesome, didn't know about this one :+1:

xdr/preconditions.go
```diff
@@ -0,0 +1,12 @@
+package xdr
+
+// NewPreconditionWithTimebounds constructs the simplest possible
+// `Preconditions` instance given the (possibly empty) timebounds.
+func NewPreconditionWithTimeBounds(timebounds *TimeBounds) Preconditions {
```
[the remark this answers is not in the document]
Shaptic: nice catch - I must've used the singular because there was only one precondition, heh

strkey/signed_payload.go
```diff
@@ -0,0 +1,71 @@
...
+const maxPayloadLen = 64
+
+func MakeSignedPayload(signerPublicKey string, payload []byte) (*SignedPayload, error) {
+	if len(payload) > maxPayloadLen {
+		return nil, errors.Errorf("payload length %d exceeds max %d",
+			len(payload), maxPayloadLen)
+	}
+
+	src := make([]byte, len(payload))
+	copy(src, payload)
```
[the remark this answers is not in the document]
Shaptic: :thinking: okay, I'll just note that it's a ref in the docstring

txnbuild/helpers_test.go
```diff
@@ -109,7 +109,7 @@ func convertToV0(tx *Transaction) {
 			SourceAccountEd25519: *tx.envelope.SourceAccount().Ed25519,
 			Fee:                  xdr.Uint32(tx.envelope.Fee()),
 			SeqNum:               xdr.SequenceNumber(tx.envelope.SeqNum()),
-			TimeBounds:           tx.envelope.V1.Tx.TimeBounds,
+			TimeBounds:           tx.envelope.V1.Tx.Cond.TimeBounds,
```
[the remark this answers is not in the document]
Shaptic: good call!

xdr/signer_key.go
Shaptic: Yeah I wasn't happy about the back-and-forth there, but wow it's so much simpler to just marshal it directly.. nice, thanks!

strkey/signed_payload.go
```diff
@@ -0,0 +1,85 @@
...
+	return &SignedPayload{Signer: signerPublicKey, Payload: src}, nil
+}
+
+// Encode turns a signed payload structure into its StrKey equivalent.
+func (sp *SignedPayload) Encode() (string, error) {
+	signerBytes, err := Decode(VersionByteAccountID, sp.Signer)
```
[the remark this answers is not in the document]
Shaptic: Yep, this is what led to my Q on Discord about whether or not we could count on this. In either case, wouldn't this be on the inner `Decode` to grok out?

strkey/signed_payload.go
```diff
@@ -0,0 +1,71 @@
...
+import (
+	"bytes"
+
+	xdr "github.com/stellar/go-xdr/xdr3"
+	"github.com/stellar/go/support/errors"
+)
+
+type SignedPayload struct {
+	Signer  string
+	Payload []byte
```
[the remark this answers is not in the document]
Shaptic: Do you think there's value in doing this prior to having a reason for doing it? :thinking: Done in 1e35b188 but curious what you see happening here

strkey/main.go
```diff
@@ -34,13 +34,19 @@ const (
...
+
+	//VersionByteSignedPayload is the version byte used for encoding "signed
+	//payload" (CAP-40) signer keys.
+	VersionByteSignedPayload = 15 << 3 // Base-32 encodes to 'P'
 )
 
-// maxPayloadSize is the maximum length of the payload for all versions.
-const maxPayloadSize = 40
+// maxPayloadSize is the maximum length of the payload for all versions. The
+// largest strkey is a signed payload: 32-byte public key + 4-byte payload
```
sreuland, line 44: can the signed payload be potentially bigger, up to the 40 byte 'multiplexed' version of key, it looks like the extra 8 bytes for 'memo' from that was included in length calculations in [sep-23/cap-40 adjustment here](https://github.com/stellar/stellar-protocol/pull/1014/files#diff-f47bbb2b1ad34e5c977897686a4b9c88f97f5c0f244fe6a8c71fbd683bf12794R79)

Shaptic: muxed addresses and signed payloads are incompatible: the signer cannot be muxed. In other words, (3) xor (4) in those instructions.

(cc @leighmcculloch to correct me if I'm wrong)

services/horizon/internal/db2/schema/migrations/54_bigger_account_signers.sql
```diff
@@ -0,0 +1,10 @@
+-- +migrate Up
+
+-- CAP-40 signed payload strkeys can be 165 characters long, see
```
sreuland, line 3: should max length of encoded strkey also include the version byte and crc, it would add 3, for total of 168 chars?

Shaptic: no: the version byte and crc are part of the binary, and `maxEncodedSize` is the base32-ification of the binary, so it already includes it

strkey/signed_payload.go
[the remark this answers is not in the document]
leighmcculloch: Changing a field from exported to not-exported is a breaking API change and so harder to walk back later given that the strkey package is essentially part of the txnbuild/horizonclient SDKs. It's trivial to make these exported in the future if we decide that the ergonomics of this aren't great. But so many of our types already work this way, such as `Transaction`, `MuxedAccount`, etc.

strkey/main.go
leighmcculloch: Yup, signed payloads only support public keys, not muxed public keys. In the future if we have different length public keys this number will potentially change.

strkey/signed_payload.go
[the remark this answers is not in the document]
leighmcculloch: Yes, I think the inner Decode would have to be updated to return the `alg` 3 bits. Not something that exists today because realistically the 3 bits are always 0.

xdr/signer_key.go
```diff
@@ -17,34 +18,42 @@ func (skey *SignerKey) Address() string {
...
+		if err != nil {
+			return "", errors.Wrap(err, "failed to marshal signed payload")
+		}
+		buf.Write(buffer[:])
 	default:
 		return "", fmt.Errorf("unknown signer key type: %v", skey.Type)
 	}
 
-	return strkey.Encode(vb, raw)
+	return strkey.Encode(vb, buf.Bytes())
```
leighmcculloch, line 56: Could we walk this change to use the buffer back to moving the make and encode into each branch? Growing buffers introduce some additional allocations into these which is a shame given we know the size of all these ahead of time.

leighmcculloch (approved):
One ask, otherwise looks great. 🎉 

Let's make sure we checkin on the strkey <-> xdr interface and relationship before this gets release in an official release.

services/horizon/internal/db2/schema/migrations/54_bigger_account_signers.sql
sreuland: ok, I thought the potential max binary length comes out to 103?
```
32(public key) + 4(signed payload length) + 64(signed payload) + 1(version byte) + 2(crc) = 103 bytes = 168 base32 encoded bytes?
```

xdr/signer_key.go
Shaptic: The thing is, we don't know the size of the signed payload buffer, so it must grow. However, I was _significantly_ over-complicating that lol.. see ef43250.

services/horizon/internal/db2/schema/migrations/54_bigger_account_signers.sql
Shaptic: According to the code, `maxRawSize` (103 bytes) becomes (`maxEncodedSize`) 165 after base32, not 168:

```go
// maxEncodedSize is the maximum length of a strkey when base32 encoded.
const maxEncodedSize = (maxRawSize*8 + 4) / 5 // (8n+4)/5 is the EncodedLen for no padding
```

xdr/signer_key.go
leighmcculloch: ef4325083c76f72c3ef4904da0a9ba1100cb6ee0 overwrites the `raw`, but then there's a `raw` that's been allocated unnecessarily.

I'd change the definition of `raw` to be:
```
raw := []byte(nil)
```

Then do the `raw = make([]byte, 32)` in each branch of the switch statement. In the signed payload branch you an do what you're doing here, just setting it.

services/horizon/internal/db2/schema/migrations/54_bigger_account_signers.sql
sreuland: doh! my bad, I see, no padding, thanks for clarification!

PR #4280, opened 2022-03-14
xdr: add instructions for generating xdr

leighmcculloch:
### What

Add a readme for the xdr package describing how to generate xdr.

### Why

I frequently forget and have to figure it out each time. This might help someone else too.

The examples use docker so that it is a quick copy/paste for most people. We use docker frequently so most of us have docker. There's a good chance that less of us have ruby setup in a good state locally.

I also added the ref to the Rakefile because I also forget how to specify a different source location for downloading XDR and if the ref was always specified it would be easier to just edit it.

### Known limitations

N/A

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -29,7 +29,7 @@ namespace :xdr do
 
     HAYASHI_XDR.each do |src|
       local_path = "xdr/" + File.basename(src)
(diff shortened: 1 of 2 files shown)


Shaptic (approved):
Can confirm this worked :+1:

PR #4284, opened 2022-03-15
all: remove go1.16 add go1.18

leighmcculloch:
### What

Remove testing and official support for Go 1.16, and add support for Go 1.18. Update all docker images to use Go 1.18, except Horizon.

### Why

Go 1.18 was released today and Go 1.16 is now no longer supported for security fixes.

### Known limitations

I have not updated the Horizon docker images as that should be coordinated by @stellar/horizon-committers.

Please note that this commit will cease to run Horizon whole program checks. See this for more details:
https://github.com/stellar/go/blob/3d935af3d9e78bec5a729473cac3434689cd4186/staticcheck.sh#L19-L25

Depends on: https://github.com/docker-library/official-images/pull/12045

### DIFF
--- a/exp/services/recoverysigner/docker/Dockerfile
+++ b/exp/services/recoverysigner/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.17 as build
+FROM golang:1.18 as build
 
 ADD . /src/recoverysigner
 WORKDIR /src/recoverysigner
--- a/go.mod
+++ b/go.mod
@@ -3,44 +3,29 @@ module github.com/stellar/go
 go 1.17
 
 require (
-	cloud.google.com/go/firestore v1.5.0 // indirect
 	firebase.google.com/go v3.12.0+incompatible
 	github.com/BurntSushi/toml v0.3.1
 	github.com/Masterminds/squirrel v1.5.0
 	github.com/Microsoft/go-winio v0.4.14
 	github.com/adjust/goautoneg v0.0.0-20150426214442-d788f35a0315
-	github.com/ajg/form v0.0.0-20160822230020-523a5da1a92f // indirect
 	github.com/asaskevich/govalidator v0.0.0-20210307081110-f21760c49a8d
 	github.com/aws/aws-sdk-go v1.39.5
 	github.com/elazarl/go-bindata-assetfs v1.0.0
-	github.com/fatih/structs v1.0.0 // indirect
-	github.com/gavv/monotime v0.0.0-20161010190848-47d58efa6955 // indirect
 	github.com/getsentry/raven-go v0.0.0-20160805001729-c9d3cc542ad1
 	github.com/go-chi/chi v4.0.3+incompatible
 	github.com/go-errors/errors v0.0.0-20150906023321-a41850380601
-	github.com/gobuffalo/packr v1.12.1 // indirect
 	github.com/golang-jwt/jwt v3.2.1+incompatible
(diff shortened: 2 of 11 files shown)


exp/services/recoverysigner/docker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.17 as build
+FROM golang:1.18 as build
```
leighmcculloch, line 1: The 1.18 image hasn't been released yet, but it's on its way, so I'm going to keep this here and merge this when it's out which will hopefully be today or tomorrow.

go.mod
```diff
@@ -76,11 +105,24 @@ require (
...
+	golang.org/x/net v0.0.0-20211112202133-69e39bad7dc2 // indirect
+	golang.org/x/oauth2 v0.0.0-20210628180205-a41e5a781914 // indirect
 	golang.org/x/sys v0.0.0-20210630005230-0f9fa26af87c // indirect
-	google.golang.org/api v0.50.0
-	gopkg.in/gavv/httpexpect.v1 v1.0.0-20170111145843-40724cf1e4a0
+	golang.org/x/term v0.0.0-20201126162022-7de9c90e9dd1 // indirect
+	golang.org/x/text v0.3.6 // indirect
+	golang.org/x/tools v0.1.4 // indirect
+	golang.org/x/xerrors v0.0.0-20200804184101-5ec99f83aff1 // indirect
+	google.golang.org/appengine v1.6.7 // indirect
```
paulbellamy, line 119: Seems like this has pulled in a lot of new stuff, which is weird?

go.mod
```diff
@@ -1,73 +1,102 @@
 module github.com/stellar/go
 
-go 1.16
+go 1.17
```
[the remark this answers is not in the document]
leighmcculloch: This is the minimum version we support which is now 1.17 because we support them last two versions since the last two versions get security releases.

go.mod
leighmcculloch: Go 1.18 came with some changes both fixes and changed behavior to how go.mod files are rendered and the contents that get included. The most important thing is that the go.sum and go.list files don't change. Those files capture definitively hashes for our dependencies and what dependencies we have in our graph, respectively.

leighmcculloch: I do think it is odd appengine is being added here, because that isn't included in our go.list, although it is included in our go.sum for some reason. The go.list tells us what is used to build everything in the repo so I thought it would be included if it was in our go.mod.

The friendly folks on Gophers Slack ([conversation here](https://gophers.slack.com/archives/C0275J6CG1M/p1647631070243549)) helped me understand that the `go list` command we use to generate `go.list` is affected by build tags, so the contents can change per operating system, architecture, and environment (such as is it running in appengine), where-as the go.mod and go.sum ignore build tags. This might change in Go 1.19 if golang/go#42504 is implemented. So this oddity is fine.

paulbellamy: ah, so it was always there, just explicit now. Cool.

```
$ go mod why google.golang.org/appengine
# google.golang.org/appengine
github.com/stellar/go/exp/services/recoverysigner/internal/serve
firebase.google.com/go/auth
google.golang.org/appengine
```
part of firebase, looks like. 👍

paulbellamy:
For the whole-program checks, is there something else we could add to the github actions to compensate? or good enough without. Wdyt?

paulbellamy (approved):

PR #4292, opened 2022-03-17
all: Fix Horizon integration tests

2opremio:
They are failing with:

```
The following packages have unmet dependencies:
 stellar-core : Depends: libc++1-8 but it is not going to be installed
E: Unable to correct problems, you have held broken packages.
```

Most likely due to a problem in the azure deb Ubuntu mirror used by GitHub

Workaround for https://github.com/actions/virtual-environments/issues/5245

### DIFF
--- a/.github/workflows/horizon.yml
+++ b/.github/workflows/horizon.yml
@@ -58,8 +58,8 @@ jobs:
       name: Install and enable Captive Core
       run: |
         # Workaround for https://github.com/actions/virtual-environments/issues/5245 , remove when fixed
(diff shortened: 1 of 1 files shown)


2opremio:
Tests passed when rerunning, so I don’t think this fixes it.

Shaptic (approved):
Hope it works! :heart_eyes: :raised_hands:

2opremio:
Now some integration tests time out, jeez

2opremio:
I will deal with the timeout separately, merging.

Issue #4276, opened 2022-03-14
services/horizon: Support new CAP-21 transaction conditions


Shaptic:
Breakdown and Rollout: [Horizon & Protocol 19](https://docs.google.com/document/d/1qZ6o14JBOZpyDG6rNEd8Bj-sCzpeZgyx7Vtl1SMKI28). See #4261 for the master issue.

----------------

### Background
[CAP-21](https://stellar.org/protocol/cap-21) adds a new set of transaction preconditions:

  * Time bounds (`timeBounds`): as before, you can define a min/max time range
  * Ledger bounds (`ledgerBounds`): a min/max ledger range
  * Sequence num (`minSeqNum`): a min sequence number, meaning the transaction is only valid if the source account’s sequence number is >= than this
  * Duration (`minSeqAge`): a ledger time delta, meaning the transaction is only valid if a certain amount of time has passed since account’s last sequence number change (expanded upon later)
  * Sequence-ledger gap (`minSeqLedgerGap`): a ledger number delta past the account’s last sequence number change (similar to `minSeqAge` but a ledger count)

Relevant XDR:

```xdr
typedef int64 Duration;

struct LedgerBounds {
    uint32 minLedger;
    uint32 maxLedger;
};

struct PreconditionsV2 {
    TimeBounds *timeBounds;

    // Transaciton only valid for ledger numbers n such that
    // minLedger <= n < maxLedger
    LedgerBounds *ledgerBounds;

    // If NULL, only valid when sourceAccount's sequence number
    // is seqNum - 1.  Otherwise, valid when sourceAccount's
    // sequence number n satisfies minSeqNum <= n < tx.seqNum.
    // Note that after execution the account's sequence number
    // is always raised to tx.seqNum, and a transaction is not
    // valid if tx.seqNum is too high to ensure replay protection.
    SequenceNumber *minSeqNum;

    // For the transaction to be valid, the current ledger time must
    // be at least minSeqAge greater than sourceAccount's seqTime.
    Duration minSeqAge;

    // For the transaction to be valid, the current ledger number
    // must be at least minSeqLedgerGap greater than sourceAccount's
    // seqLedger.
    uint32 minSeqLedgerGap;

    // For the transaction to be valid, there must be a signature
    // corresponding to every Signer in this array, even if the
    // signature is not otherwise required by the sourceAccount or
    // operations.
    SignerKey extraSigners<2>;
};
```

### Affected Endpoints
 * `GET /transactions`
[...]

PR #4297, opened 2022-03-21
services/horizon: Support new CAP-21 transaction conditions

paulbellamy:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Support new Protocol 19 Transaction preconditions.

Since all previous preconditions are subsets of the new `xdr.PreconditionsV2`, this PR replaces `xdr.TransactionEnvolope.TimeBounds()` with the more general `xdr.TransactionEnvelope.Preoconditions()`, which normalizes old transactions, and returns a `xdr.PreconditionsV2`.

I also renamed the `Signers` field to `ExtraSigners`, so that it matches the json field naming. Though I'm not strongly attached to this if people oppose.

Stores the preconditions in the DB as part of the `transaction_processor.go`.

Renders them in the API via the `resourceadapter`.

### Why

Fixes https://github.com/stellar/go/issues/4276

### Known limitations

Depends on https://github.com/stellar/go/pull/4279


leighmcculloch (commented):
👏🏻 Couple asks/issues (❗) and a suggestion (💡).

services/horizon/internal/db2/schema/migrations/55_protocol_19.sql
```diff
@@ -0,0 +1,13 @@
+-- +migrate Up
+ALTER TABLE history_transactions ADD ledger_bounds                   int8range;
+ALTER TABLE history_transactions ADD min_account_sequence            integer;
+ALTER TABLE history_transactions ADD min_account_sequence_age        integer;
+ALTER TABLE history_transactions ADD min_account_sequence_ledger_gap integer;
+ALTER TABLE history_transactions ADD extra_signers                   character varying(165)[];
```
Shaptic, line 6: Couple of questions:
 * what're your thoughts on making this a separate table e.g. `history_transaction_preconditions`? feels icky to bloat `history_transactions` for a feature that seems likely be used pretty rarely :thinking: 
 * can we rename the migration to be more descriptive?
 * will these columns be nullable by default? i.e. we don't have to add some constraints here?

services/horizon/internal/resourceadapter/transaction.go
```diff
@@ -57,6 +57,60 @@ func PopulateTransaction(
 	if !row.TimeBounds.Null {
 		dest.ValidBefore = timeString(dest, row.TimeBounds.Upper)
 		dest.ValidAfter = timeString(dest, row.TimeBounds.Lower)
```
Shaptic, line 59: Maybe add an `// Action needed` comment here for when we can drop these duplicate fields?

services/horizon/internal/resourceadapter/transaction.go
```diff
@@ -57,6 +57,60 @@ func PopulateTransaction(
...
+		if dest.Preconditions == nil {
+			dest.Preconditions = &protocol.TransactionPreconditions{}
+		}
+		dest.Preconditions.MinAccountSequenceLedgerGap = uint32(row.MinAccountSequenceLedgerGap.Int64)
+	}
+
+	if len(row.ExtraSigners) > 0 {
+		if dest.Preconditions == nil {
+			dest.Preconditions = &protocol.TransactionPreconditions{}
+		}
```
Shaptic, line 112: Thoughts on outsourcing this to a `HasPreconditions(row history.Transaction) boolean` helper and putting this repeated block once at the top?

services/horizon/internal/db2/history/transaction_ledger_bounds.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/stellar/go/support/errors"
+	"github.com/stellar/go/xdr"
+)
+
+// LedgerBounds represents the ledger bounds of a Stellar transaction
+type LedgerBounds struct {
+	Null      bool
+	MaxLedger null.Int
+	MinLedger null.Int
+}
```
Shaptic, line 18: I take it we don't have a way to force these to be `uint32`s? I ask because I see a lot of `...MaxLedger.Int64()` elsewhere in the code

services/horizon/internal/resourceadapter/transaction.go
```diff
@@ -57,6 +57,60 @@ func PopulateTransaction(
 	if !row.TimeBounds.Null {
 		dest.ValidBefore = timeString(dest, row.TimeBounds.Upper)
 		dest.ValidAfter = timeString(dest, row.TimeBounds.Lower)
+
+		if dest.Preconditions == nil {
+			dest.Preconditions = &protocol.TransactionPreconditions{}
+		}
+		if dest.Preconditions.Timebounds == nil {
```
Shaptic, line 64: supernit: can we be consistent and say `TimeBounds` and `LedgerBounds` everywhere? (capital `B`) Even though I always read `timebounds` as one word like you have here, we should just stick to the existing convention

services/horizon/internal/resourceadapter/transaction.go
```diff
@@ -57,6 +57,60 @@ func PopulateTransaction(
...
+
+	if !row.LedgerBounds.Null {
+		if dest.Preconditions == nil {
+			dest.Preconditions = &protocol.TransactionPreconditions{}
+		}
+		if dest.Preconditions.Ledgerbounds == nil {
+			dest.Preconditions.Ledgerbounds = &protocol.TransactionPreconditionsLedgerbounds{}
+		}
+		if row.LedgerBounds.MinLedger.Valid {
+			minLedger := uint32(row.LedgerBounds.MinLedger.Int64)
```
Shaptic, line 79: This won't throw if it's not a `uint32`, and even though it definitely shouldn't be, do we have a place to enforce that invariant?

xdr/preconditions.go
```diff
@@ -10,3 +10,24 @@ func NewPreconditionsWithTimeBounds(timebounds *TimeBounds) Preconditions {
 	}
 	return cond
 }
+
+// GetPreconditions normalizes and returns the preconditions from the transaction.
```
Shaptic, line 14: not sure if blind but this seems unused? if that's the intent (sdk helper) then maybe `GetNormalizedPreconditions()` is a better name by pure virtue of the fact that the thing that gets returned might not match the thing that's actually in the transaction

xdr/transaction.go
```diff
@@ -14,3 +14,78 @@ func (tx *Transaction) TimeBounds() *TimeBounds {
...
+	case PreconditionTypePrecondNone:
+		return nil
+	case PreconditionTypePrecondTime:
+		return nil
+	case PreconditionTypePrecondV2:
+		return tx.Cond.V2.LedgerBounds
+	default:
+		panic("unsupported transaction type: " + tx.Cond.Type.String())
+	}
+}
```
Shaptic, line 31: We could invert the logic here: drop the `switch` and only say `if tx.Cond.Type == PreconditionTypePrecondV2 && tx.Cond.V2 != nil { return tx.Cond.V2.LedgerBounds }`, instead. I guess we lose the panic case but it removes a **lot** of no-op code from these helpers

xdr/transaction_envelope.go
```diff
@@ -103,6 +103,86 @@ func (e TransactionEnvelope) TimeBounds() *TimeBounds {
...
+	case EnvelopeTypeEnvelopeTypeTxV0:
+		return nil
+	default:
+		panic("unsupported transaction type: " + e.Type.String())
+	}
+}
+
+// MinSeqNum returns the min seq num set in the transaction envelope
+// Note for fee bump transactions, TimeBounds() returns the value
+// of the inner transaction
```
Shaptic, line 124: Copy paste error in a few places? You could say, generically, `Note that for fee bump transactions, this will return the field on the *inner* transaction, if it exists.`

Shaptic (commented):
Coming along nicely :clap: added a potpourri of thoughts :arrow_down:

xdr/preconditions.go
leighmcculloch: If this isn't used, I think we should drop it all together. Normalized preconditions are already accessible from the top level with functions like `.TimeBounds()`, `.LedgerBounds()`. The internals of how the data is structured inside the XDR is unimportant to most applications, and when I prototyped CAP-21 a need for a function like this didn't show up.

xdr/transaction.go
leighmcculloch: +1 to the panic case being important.

It's convenient that all these functions look the same. They're easy to check and understand. Less lines of code doesn't always help readability and clarity. Looks great imo!

xdr/transaction.go
```diff
@@ -14,3 +14,78 @@ func (tx *Transaction) TimeBounds() *TimeBounds {
...
+func (tx *Transaction) LedgerBounds() *LedgerBounds {
+	switch tx.Cond.Type {
+	case PreconditionTypePrecondNone:
+		return nil
+	case PreconditionTypePrecondTime:
+		return nil
+	case PreconditionTypePrecondV2:
+		return tx.Cond.V2.LedgerBounds
+	default:
+		panic("unsupported transaction type: " + tx.Cond.Type.String())
```
leighmcculloch, line 29: Looks like a copy-paste typo here, the type is preconditions, not transaction:
```suggestion
		panic("unsupported preconditions type: " + tx.Cond.Type.String())
```

services/horizon/internal/db2/schema/migrations/55_protocol_19.sql
leighmcculloch, line 6: Is there a reason we prefer `varchar` over `text`? In Postgres text is generally preferred. There is no performance gain of using varchar. See https://wiki.postgresql.org/wiki/Don%27t_Do_This#Don.27t_use_varchar.28n.29_by_default.

Also altering varchar length might (I don't remember, I suspect) require a lock on the entire table which would make future migrations expensive each time we add a longer signer key.

services/horizon/internal/resourceadapter/transaction.go
paulbellamy: The existing convention was actually both upper and lowercase, so I matched that. But I agree more consistency would be good.

services/horizon/internal/db2/schema/migrations/55_protocol_19.sql
paulbellamy: I picked varchar for consistency with what we do everywhere else we store signer keys.

The lock for resizing is probably a massive issue on https://github.com/stellar/go/pull/4279/files#diff-11d5fd25b957ff7f9d9b3728324a2bb58234ae077393090618dad1d96c407dfaR6, then.

services/horizon/internal/db2/schema/migrations/55_protocol_19.sql
```diff
@@ -0,0 +1,13 @@
+-- +migrate Up
+ALTER TABLE history_transactions ADD ledger_bounds                   int8range;
+ALTER TABLE history_transactions ADD min_account_sequence            integer;
+ALTER TABLE history_transactions ADD min_account_sequence_age        integer;
+ALTER TABLE history_transactions ADD min_account_sequence_ledger_gap integer;
```
paulbellamy, line 5: Note: Check integer is big enough for uint32

paulbellamy: From https://www.postgresql.org/docs/9.5/datatype-numeric.html#DATATYPE-INT, `integer` is 4 bytes, so. all good.

services/horizon/internal/db2/schema/migrations/55_protocol_19.sql
paulbellamy: - I'd want to move the time_bounds field over to that new table, which might be a big migration?
- sure. I assumed we'd merge them all together into a protocol_19 migration pre-release, anyway.
- Yes. If we added `NOT NULL` they would not be nullable. But would take forever to add (unless we moved to a separate table)

xdr/transaction.go
paulbellamy: Left the switch, but grouped the nil cases.

services/horizon/internal/db2/schema/migrations/55_protocol_19.sql
paulbellamy: Thinking through if there's a quick way to do the migration. Maybe it could be like:

```sql
create table history_transactions_preconditions (
  transaction_id text primary key,
  time_bounds int8range
  --- the rest ---
);

-- 2. create the preconditions for existing transactions.
-- This is probably the slow step due to the select from...
insert into history_transactions_preconditions (
  select
    id as transaction_id,
    time_bounds
  from history_transactions
);

-- 3. drop the old time_bounds column.
alter table history_transactions drop column time_bounds;
```

By putting the `transaction_id` column on the preconditions side, we don't have to modify the `history_transactions` table.

paulbellamy: ^ cc @Shaptic

leighmcculloch: The typical pattern I've used for col changes are below. Since there are always multiple versions of the app deployed talking to the same db, you can't drop or stop populating columns that an older version might be using. Strictly speaking you should always space these releases apart to adhere to any rollback policy. i.e. If you say you support rollbacks to the last 2 releases, then all the migrations and releases should be at least 3 releases apart.

---
migration 1
- add new col

release 1
- write to new and old col
- read old col
---
migration 2
- backfill new col with old col

release 2
- read new col
---
release 3
- stop writing to old col
---
migration 4
- drop old col
---

services/horizon/internal/resourceadapter/transaction.go
Shaptic: seems good :+1:

services/horizon/internal/db2/schema/migrations/55_protocol_19.sql
Shaptic: yep I did the same thing for consistency.. we should figure out why as it's before my time. asking internally

Shaptic: > if we added `NOT NULL`

we _should_ let them null, that's what I wanted to check :+1: 

> I'd want to move the time_bounds field over to that new table, which might be a big migration?

Since we're already duplicating it in the API response, we might as well carry that through to the DB :joy: and Leigh is right, we can kick off a proper migration in phases and by Horizon v3.0.0 we'd have it all in a separate table

leighmcculloch: Preconditions are a concept of the XDR introduced purely to group these new fields together, and purely because we could sneakily extend the transaction using the pointer discriminant on the existing time bounds. The fact the XDR has "preconditions" does not necessarily need to be a concept in the systems above. Transactions just have a few new attributes. Don't feel compelled to group them like that in other systems.

Do you always have to join to the other table? Then it may be better to keep them inside the same table if there is no natural need for a multi-to-one relationship.

Also, null columns only take up 1-bit of space on disk in Postgres. That 1-bit is stored in the null bitmap. Each row has a null bitmap. If the row already has a nullable column, that means adding the new columns will likely take up no new space, because each new column just uses a new bit in the existing null bitmap. I'm not sure what the alignment of the null bitmap is, but whether it gets bigger or not will be a factor of whether there are any bit slots left available. i.e. Adding these columns might increase table size, or might not.

It is true that really wide tables can be hard to manage for other reasons. Most of these fields are small though, except extrasigners. Maybe just extra signers should be separate and the rest should be bundled into transactions?

services/horizon/internal/db2/schema/migrations/55_protocol_19.sql
Shaptic: Should still convert correctly, yeah? 😅

services/horizon/internal/db2/schema/migrations/55_protocol_19.sql
paulbellamy: extrasigners is max [2]varchar(165), so 330 bytes. Not sure what the point where wide columns get unwieldy is, but I'd expect it is larger than that... 🤔 

We'd need to join the new table for listing txns, or showing txn by hash.

paulbellamy: Having extrasigners as a separate table *could* lead to some deduplication, though. assuming most of the values are addresses, and not payloads.

Shaptic: These are some interesting points. My suggestion re: a separate table the concern was two-fold and you covered both points: keeping conditions as a separate concern and overall table "width" (i.e. new columns that are usually empty). Since it seems like it doesn't really matter as far as table size blowup goes, and since migrations to a separate table would be annoying, maybe not worth it? But will defer to Paul.

leighmcculloch: We may add additional extra signers in the future. It's set to a length of 2 now only because that's the bare minimum we need today. If a protocol building on Stellar says they need more, I don't see why we wouldn't increase it. Given the additional size extra signers add, it could be worth adding an additional table? But I agree all the others don't seem to drive the need for a separate 'preconditions' table.

services/horizon/internal/resourceadapter/transaction.go
paulbellamy: it comes from the xdr, which is a uint32

protocols/horizon/main.go
```diff
@@ -499,34 +499,57 @@ type Transaction struct {
...
+}
+
+type TransactionPreconditionsTimebounds struct {
+	MinTime string `json:"min_time,omitempty"`
+	MaxTime string `json:"max_time,omitempty"`
+}
+
+type TransactionPreconditionsLedgerbounds struct {
+	MinLedger *uint32 `json:"min_ledger,omitempty"`
+	MaxLedger *uint32 `json:"max_ledger,omitempty"`
```
Shaptic, line 552: Wait, can these be `nil`? They can be zero, yes, but if the structure is there then both of these fields must also be present. So neither pointers nor omitempty, I think.

services/horizon/internal/db2/history/transaction_batch_insert_builder_test.go
```diff
@@ -180,3 +182,98 @@ func TestTransactionToMap_SourceMuxedAndFeeSourceUnmuxed(t *testing.T) {
...
+		Ed25519: &xdr.Uint256{3, 2, 1},
+	}
+	minSeqNum := xdr.SequenceNumber(24)
+	signerKey := xdr.SignerKey{
+		Type:    xdr.SignerKeyTypeSignerKeyTypeEd25519,
+		Ed25519: source.Ed25519,
+	}
+	tx := ingest.LedgerTransaction{
+		Index: 1,
+		Envelope: xdr.TransactionEnvelope{
```
Shaptic, line 198: I'm pretty sure you can just marshal/b32ify this to get the sample xdr you were asking for, btw!

services/horizon/internal/db2/history/transaction_ledger_bounds.go
```diff
@@ -0,0 +1,92 @@
...
+	if ledgerBounds == nil {
+		return LedgerBounds{Null: true}
+	}
+
+	// TODO: Is this right?
+	if ledgerBounds.MaxLedger == 0 {
+		return LedgerBounds{
+			MinLedger: null.IntFrom(int64(ledgerBounds.MinLedger)),
+		}
+	}
```
Shaptic, line 85: you can leave the 0 upper bound

services/horizon/internal/db2/history/transaction_ledger_bounds.go
```diff
@@ -0,0 +1,92 @@
...
+	return nil
+}
+
+// Value implements the database/sql/driver Valuer interface.
+func (t LedgerBounds) Value() (driver.Value, error) {
+	if t.Null {
+		return nil, nil
+	}
+
+	if !t.MaxLedger.Valid {
```
Shaptic, line 68: if you leave the 0 upper bound for `MaxLedger = 0`, you can check it here for the right formatting, e.g. 
```suggestion
	if !t.MaxLedger.Valid || t.MaxLedger == 0 {
```

services/horizon/internal/db2/history/transaction_ledger_bounds.go
Shaptic: in fact, neither should ever be `nil` (mentioned this above), just double-check that you handle `{minLedger: n, maxLedger: 0}` correctly

services/horizon/internal/db2/schema/migrations/54_bigger_account_signers.sql
```diff
@@ -1,9 +1,10 @@
 -- +migrate Up
 
 -- CAP-40 signed payload strkeys can be 165 characters long, see
--- strkey/main.go:maxEncodedSize
+-- strkey/main.go:maxEncodedSize. But we use text here, so we don't need to
+-- adjust it *ever again*.
```
Shaptic, line 5: :tada:

services/horizon/internal/resourceadapter/transaction.go
```diff
@@ -54,9 +54,51 @@ func PopulateTransaction(
...
-		dest.ValidAfter = timeString(dest, row.TimeBounds.Lower)
+		// Action needed in release: horizon-v3.0.0: remove ValidBefore and ValidAfter
+		dest.ValidBefore = timeString(row.TimeBounds.Upper)
+		dest.ValidAfter = timeString(row.TimeBounds.Lower)
+
+		if dest.Preconditions.Timebounds == nil {
+			dest.Preconditions.Timebounds = &protocol.TransactionPreconditionsTimebounds{}
+		}
+		dest.Preconditions.Timebounds.MaxTime = timeString(row.TimeBounds.Upper)
+		dest.Preconditions.Timebounds.MinTime = timeString(row.TimeBounds.Lower)
```
Shaptic, line 71: wait, erm, weird question and might be a knowledge gap. why not always create the structure?

```suggestion
		dest.Preconditions.Timebounds = &protocol.TransactionPreconditionsTimebounds{
			MaxTime: timeString(row.TimeBounds.Upper),
			MinTime: timeString(row.TimeBounds.Lower),
		}
```

services/horizon/internal/resourceadapter/transaction.go
```diff
@@ -54,9 +54,51 @@ func PopulateTransaction(
...
+
+	if row.MinAccountSequenceAge.Valid {
+		dest.Preconditions.MinAccountSequenceAge = fmt.Sprint(row.MinAccountSequenceAge.Int64)
+	}
+
+	if row.MinAccountSequenceLedgerGap.Valid {
+		dest.Preconditions.MinAccountSequenceLedgerGap = uint32(row.MinAccountSequenceLedgerGap.Int64)
+	}
+
+	if len(row.ExtraSigners) > 0 {
```
Shaptic, line 100: can skip check the condition, right? if it's empty it'll stay empty

xdr/transaction_envelope_test.go
```diff
@@ -85,6 +85,66 @@ func createTx() TransactionEnvelope {
 	}
 }
 
+func createV2Tx() TransactionEnvelope {
```
Shaptic, line 88: nit: technically it's a `CondV2` transaction, not a V2 transaction :eyes: 
```suggestion
func createCondV2Tx() TransactionEnvelope {
```

Shaptic (commented):
Handful of small notes, but one bigger thing: while `Preconditions.LedgerBounds` can be `nil`, I don't think its members `Min/MaxLedger` can be `nil`. That's an oversight in the design doc on my part, sorry.

services/horizon/internal/resourceadapter/transaction.go
paulbellamy: Without the check it panics on empty.

services/horizon/internal/resourceadapter/transaction.go
```diff
@@ -54,9 +54,49 @@ func PopulateTransaction(
...
+	if !row.LedgerBounds.Null {
+		if dest.Preconditions.Ledgerbounds == nil {
+			dest.Preconditions.Ledgerbounds = &protocol.TransactionPreconditionsLedgerbounds{}
+		}
+		if row.LedgerBounds.MinLedger.Valid {
+			dest.Preconditions.Ledgerbounds.MinLedger = uint32(row.LedgerBounds.MinLedger.Int64)
+		}
+		if row.LedgerBounds.MaxLedger.Valid {
+			dest.Preconditions.Ledgerbounds.MaxLedger = uint32(row.LedgerBounds.MaxLedger.Int64)
+		}
```
Shaptic, line 83: Both of these are always true now since they can never be not set, right?

Shaptic (approved):
One tiny note, but otherwise LGTM! :tada:

services/horizon/internal/resourceadapter/transaction.go
paulbellamy: Eh, they could technically be nil in the db.

PR #4301, opened 2022-03-22
services/horizon: Update txsub queue to account for new CAP-21 preconditions

### LINKED ISSUE #4283: services/horizon: Update txsub queue to account for new CAP-21 preconditions
The rules that make a transaction’s sequence number valid have changed considerably. As such, the transaction submission queue will need changes. Horizon will have a harder time “helping” submitters by ordering their transactions, waiting for particular sequence numbers, etc. However, this _only_ happens in the case in which transactions have the `minSeqNum` condition set, which is not expected to be a large portion.

Today Horizon orders transactions by sequence, then submits the lowest one in that queue if it equals the next expected sequence on the account. With CAP-21 loosening the validity conditions, Horizon needs to allow gaps in sequence numbers (rather than strict increments), though it should still order them in ascending order.

The necessary changes are minor: if the next transaction in the queue has a `minSeqNum` condition set, we also check if:

    account.seqNum < tx.seqNum && account.seqNum >= tx.minSeqNum - 1

2opremio:
Fixes #4283


services/horizon/internal/txsub/sequence/account_tx_submission_queue.go
```diff
@@ -63,35 +73,34 @@ func (q *AccountTxSubmissionQueue) Push(sequence uint64) <-chan error {
...
-
-		if hseq < q.nextSequence {
-			ch <- ErrBadSequence
-			close(ch)
-		} else if hseq == q.nextSequence {
-			ch <- nil
-			close(ch)
+		if q.lastSeenAccountSequence > tx.MaxAccSeqNum {
+			// The transaction and account sequences will never match
+			removeWithErr(ErrBadSequence)
```
tamirms, line 96: do we need to break out of the for loop?

services/horizon/internal/txsub/sequence/account_tx_submission_queue.go
```diff
@@ -63,35 +73,34 @@ func (q *AccountTxSubmissionQueue) Push(sequence uint64) <-chan error {
...
+		} else if q.lastSeenAccountSequence >= tx.MinAccSeqNum {
+			// within range, ready to submit!
+			removeWithErr(nil)
```
tamirms, line 99: same question as above, do we need to break out of the for loop?

services/horizon/internal/txsub/sequence/account_tx_submission_queue.go
tamirms: nevermind, I misread the code

services/horizon/internal/txsub/sequence/account_tx_submission_queue.go
```diff
@@ -63,35 +73,34 @@ func (q *AccountTxSubmissionQueue) Push(sequence uint64) <-chan error {
...
+		q.lastSeenAccountSequence = sequence
 	}
 
 	wasChanged := false
 
-	for {
-		if q.Size() == 0 {
-			break
-		}
+	// We need to traverse the full queue (ordered by MaxAccSeqNum)
```
tamirms, line 82: does the order of the elements in the queue actually matter? It seems like the logic will be correct even if the queue is out of order

tamirms: if the order does not matter anymore then I think it would be better to just use an unsorted slice instead of a heap because the code would be cleaner (we could get rid of `pqueue`)

services/horizon/internal/txsub/sequence/account_tx_submission_queue.go
```diff
@@ -51,9 +51,19 @@ func (q *AccountTxSubmissionQueue) Size() int {
 // 2. Load the current sequence number for the source account from the DB
 // 3. Call NotifyLastAccountSequence() with the result from step 2 to trigger the submission if
 //		possible
-func (q *AccountTxSubmissionQueue) Push(sequence uint64) <-chan error {
+func (q *AccountTxSubmissionQueue) Push(sequence uint64, minSeqNum *uint64) <-chan error {
 	ch := make(chan error, 1)
-	heap.Push(&q.queue, item{sequence, ch})
+	// From CAP 19: If minSeqNum is nil, the tx is only valid when sourceAccount's sequence number is seqNum - 1.
```
Shaptic, line 56: 21, not 19!

services/horizon/internal/txsub/sequence/account_tx_submission_queue.go
Shaptic: I think if we track the lowest `minSeq` in the queue separately, we can avoid traversals. Something like:

 * on every push(), track the smallest `minSeq` and the pqueue `item` that it corresponds to
 * on every notify(), if it exists and is <= the pqueue head, find it and execute it
 * if you executed, then you can drop all of the earlier items in the pqueue since they're invalid now

Then the queue can still just be sorted by `tx.minSeq`.

2opremio: @tamirms I would love not to traverse the full priority queue but we need to track both minseq and maxseq separately, see https://github.com/stellar/go/issues/4283#issuecomment-1074503302

suggestions are welcome.

Processing order matters, since you want to process txs which result in the the smallest account sequence first (to maximizes the opportunity of successfully submitting the remaining transactions)

2opremio: CC @leighmcculloch

services/horizon/internal/txsub/sequence/account_tx_submission_queue.go
```diff
@@ -63,35 +73,34 @@ func (q *AccountTxSubmissionQueue) Push(sequence uint64) <-chan error {
...
-		ch, hseq := q.head()
-		// if the next queued transaction has a sequence higher than the account's
-		// current sequence, stop removing entries
-		if hseq > q.nextSequence {
-			break
+		removeWithErr := func(err error) {
+			tx.Chan <- err
+			close(tx.Chan)
+			wasChanged = true
+			heap.Remove(&q.queue, i)
```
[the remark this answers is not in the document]
2opremio: Removing an element will reorder the slice yes. I was assuming that the slice was sorted incrementally.

my code assumes that the underlying slice is index-ordered. If that’s not the case then it’s wrong and I should rethink my implementation.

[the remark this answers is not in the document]
2opremio: > Maybe a slice that is sorted on insert would be better?

probably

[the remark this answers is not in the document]
2opremio: > In fact by removing the use of pop, it seems like we aren't really using the heap as a heap. Maybe a slice that is sorted on insert would be better?

yeah, probably

services/horizon/internal/txsub/sequence/account_tx_submission_queue.go
2opremio: @Shaptic I don’t think it’s enough to track only the smallest minseq, you need to track them all if you want to signal the submission of all possible transactions in one pass (like we do for seq, now maxseq)

2opremio: > if you executed, then you can drop all of the earlier items in the pqueue since they're invalid now

that’s a nice optimization, but you can’t do that until you get a confirmation about the transaction being successful (since, if it isn’t you could still submit those transactions)

Shaptic: Good point @2opremio :thinking: then if you end up using the smallest minSeq on `notify()`, you can traverse to find the next-smallest minSeq and repeat.

leighmcculloch: We should weigh up the size of these queues with how performant sensitive we need to be. There's a separate queue per source account, and with a 1000 transaction limit per ledger, these things aren't likely to be very big? i.e. Does it matter if we traverse the entire slice?

Shaptic: Also re: confirmation, the queue doesn't do retries, right? Then you can just submit all of the transactions `<= smallestMinSeq` and let the network figure out which ones it wants to let succeed.

tamirms: @2opremio thanks, https://github.com/stellar/go/issues/4283#issuecomment-1074503302 was very helpful. I think I understand the requirements now.

I suggest you change `pqueue.Less()` so that it has items with the least `MinAccSeqNum` at the top of the heap. Then you can pop elements from the heap into a slice which you will later sort by `MaxAccSeqNum` so that we can trigger the tx submission in order of `MaxAccSeqNum`:

```golang
   var readyToSubmit []item
   for q.Size() > 0 {
      tx := q.queue[0]
      if tx.MinAccSeqNum > q.lastSeenAccountSequence {
         break
      }
      heap.Pop(&q.queue)
      readyToSubmit = append(readyToSubmit, tx)
   }

   sort.Slice(readyToSubmit, func (i, j int) bool {
      return readyToSubmit[i].MaxAccSeqNum < readyToSubmit[j].MaxAccSeqNum
   })

   for _, tx := range readyToSubmit {
      if tx.MaxAccSeqNum < q.lastSeenAccountSequence {
         tx.Chan <- ErrBadSequence
         close(tx.Chan)
      } else {
         tx.Chan <- nil
         close(tx.Chan)
      }
   }

   if len(readyToSubmit) > 0 {
      q.lastActiveAt = time.Now()
      return
   }

   if time.Since(q.lastActiveAt) > q.timeout {
      for _, tx := range q.queue {
         tx.Chan <- ErrBadSequence
         close(tx.Chan)
      }
      q.queue = []item{}
   }
}
```

tamirms: alternatively, we can get rid of the heap and keep the queue unsorted. This will require us to iterate through all the items in the queue every time `NotifyLastAccountSequence()` is called. In the case where all the items in the queue are less than or equal to `q.lastSeenAccountSequence` this will actually be faster than popping every element from the heap. I think the size of the queue should be pretty small anyways:


```golang
   var readyToSubmit []item
   // q.queue is an unsorted slice
   for i := len(q.queue) -1; i >= 0; i-- {
      tx := q.queue[i]
      if tx.MinAccSeqNum > q.lastSeenAccountSequence {
         continue
      }
      readyToSubmit = append(readyToSubmit, tx)
      // remove tx from the queue
      if i < len(q.queue) - 1 {
         q.queue[i], q.queue[len(q.queue) -1] = q.queue[len(q.queue) -1], q.queue[i]
      }
      q.queue = q.queue[:len(q.queue)-1]
   }

   sort.Slice(readyToSubmit, func (i, j int) bool {
      return readyToSubmit[i].MaxAccSeqNum < readyToSubmit[j].MaxAccSeqNum
   })

   for _, tx := range readyToSubmit {
      if tx.MaxAccSeqNum < q.lastSeenAccountSequence {
         tx.Chan <- ErrBadSequence
         close(tx.Chan)
      } else {
         tx.Chan <- nil
         close(tx.Chan)
      }
   }

   if len(readyToSubmit) > 0 {
      q.lastActiveAt = time.Now()
      return
   }

   if time.Since(q.lastActiveAt) > q.timeout {
      for _, tx := range q.queue {
         tx.Chan <- ErrBadSequence
         close(tx.Chan)
      }
      q.queue = []item{}
   }
```

2opremio:
I have replaced the heap with a sorted slice.

I could keep a linked list (or a second slice) sorted by the minSeq instead of iterating the full slice every time but I don't think it's worth it.

services/horizon/internal/txsub/sequence/account_tx_submission_queue.go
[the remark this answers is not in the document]
2opremio: I did just that.

services/horizon/internal/txsub/sequence/account_tx_submission_queue.go
2opremio: I went for a sorted slice (by MaxAccSeqNum) which we fully iterate over (in order to reach all the possible minseqs at the end). Please take a look.

> I suggest you change `pqueue.Less()` so that it has items with the least `MinAccSeqNum` at the top of the heap. Then you can pop elements from the heap into a slice which you will later sort by `MaxAccSeqNum` so that we can trigger the tx submission in order of `MaxAccSeqNum`:

This would work, but seems a bit more elaborate than using a single sorted slice (the latest solution I proposed) and I am not sure it has better performance

> alternatively, we can get rid of the heap and keep the queue unsorted

I like this solution much better and I think it's even simpler than what I proposed. Let's go for that

2opremio: I saw a problem with your approach, if you wait to check `tx.MaxAccSeqNum < q.lastSeenAccountSequence`  until after it's put in  `readyToSubmit` you won't exhaustively notify of `ErrBadSequence` to all transactions (i.e. there can be transactions with `tx.MaxAccSeqNum < q.lastSeenAccountSequence` not moved to `readyToSubmit`)

I will fix that.

2opremio: OK, I've implemented it. @tamirms PTAL

2opremio:
And, after @tamirms 's suggestion, I replaced it by an unsorted slice which we sort as needed on the fly.

services/horizon/internal/txsub/sequence/account_tx_submission_queue.go
```diff
@@ -63,99 +76,58 @@ func (q *AccountTxSubmissionQueue) Push(sequence uint64) <-chan error {
...
-			close(ch)
-		} else if hseq == q.nextSequence {
-			ch <- nil
-			close(ch)
-		}
+	// To maximize successful submission opportunity, submit transactions by the account sequence
+	// which would result from a successful submission (i.e. maxAccSeqNum+1)
+	sort.Slice(txsToSubmit, func(i, j int) bool {
+		return txsToSubmit[i].maxAccSeqNum < txsToSubmit[j].maxAccSeqNum
+	})
```
leighmcculloch, line 112: 💡 This is going to sort the slice every time an account's sequence number changes. Would Horizon do less work if it pqueue inserted each item in sorted order? Sorting on insert is relatively trivial using the [`sort.Search`](https://pkg.go.dev/sort#Search) function to find the insert point. There's an example here: https://play.golang.org/p/4RkVgEpKsWq (from [so](https://stackoverflow.com/a/55460931)).

I'm not sure which would be more performant, and regardless it might not matter.

leighmcculloch (commented):
🎉 The rewrite using a simple slice is easy to understand, and the code is really clear top to bottom. So wonderful.

I left two comments, one suggestion that is not critical but mentioning in case it was not considered (💡), and the other a missing hook up that I'm not sure is in or out of scope of this PR (🔌).

services/horizon/internal/txsub/sequence/account_tx_submission_queue.go
```diff
@@ -63,99 +76,58 @@ func (q *AccountTxSubmissionQueue) Push(sequence uint64) <-chan error {
 // This function is monotonic... calling it with a sequence number lower than
 // the latest seen sequence number is a noop.
 func (q *AccountTxSubmissionQueue) NotifyLastAccountSequence(sequence uint64) {
-	if q.nextSequence <= sequence {
-		q.nextSequence = sequence + 1
+	if q.lastSeenAccountSequence <= sequence {
```
Shaptic, line 79: silly nitpick, but you could make this `<` instead of `<=` here lol

services/horizon/internal/txsub/sequence/account_tx_submission_queue.go
```diff
@@ -63,99 +76,58 @@ func (q *AccountTxSubmissionQueue) Push(sequence uint64) <-chan error {
...
-		if q.Size() == 0 {
-			break
+	queueWasChanged := false
+
+	txsToSubmit := make([]txToSubmit, 0, len(q.transactions))
+	// Extract transactions ready to submit and notify those which are un-submittable.
+	for i := 0; i < len(q.transactions); {
+		candidate := q.transactions[i]
+		removeCandidateFromQueue := false
+		if q.lastSeenAccountSequence > candidate.maxAccSeqNum {
```
Shaptic, line 90: I _think_ this can be `>=`? Since if the sequence number has been "seen" that means it has been "consumed."

services/horizon/internal/txsub/sequence/account_tx_submission_queue.go
```diff
@@ -51,9 +54,19 @@ func (q *AccountTxSubmissionQueue) Size() int {
...
+	if minSeqNum != nil {
+		effectiveMinSeqNum = *minSeqNum
+	}
 	ch := make(chan error, 1)
-	heap.Push(&q.queue, item{sequence, ch})
+	q.transactions = append(q.transactions, txToSubmit{
+		minAccSeqNum:   effectiveMinSeqNum,
+		maxAccSeqNum:   sequence - 1,
+		notifyBackChan: ch,
+	})
```
Shaptic, line 69: It's possible to submit a tx where `*minSeqNum >= sequence` here (of course it would never succeed, but someone could still submit it...). Can you double-check that this wouldn't blow up the queue somehow?

Shaptic (approved):
This approach is really clean :clap: minor notes below and something to test :arrow_down:

tamirms (approved):

services/horizon/internal/txsub/sequence/account_tx_submission_queue.go
2opremio: It depends on the sorting algorithm. I would say it’s an early optimization unless submission queues are performance critical (which I doubt).

services/horizon/internal/txsub/system.go
```diff
@@ -132,11 +132,11 @@ func (sys *System) Submit(
 
 	// queue the submission and get the channel that will emit when
 	// submission is valid
-	seq := sys.SubmissionQueue.Push(sourceAddress, uint64(envelope.SeqNum()))
+	seq := sys.SubmissionQueue.Push(sourceAddress, uint64(envelope.SeqNum()), nil)
```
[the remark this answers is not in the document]
2opremio: I do think it’s out of scope

services/horizon/internal/txsub/sequence/account_tx_submission_queue.go
2opremio: No, if it’s equal it can still be submitted. If I didn’t understand your point (which may be the case) can you elaborate further?

services/horizon/internal/txsub/sequence/account_tx_submission_queue.go
2opremio: true

services/horizon/internal/txsub/sequence/account_tx_submission_queue.go
2opremio: We later on do check `maxSeq` but .... we do assume that `maxSeq` is always equal or higher that `minSeq` and I guess that doesn't necessarily need to be the case.

Good catch.

services/horizon/internal/txsub/system.go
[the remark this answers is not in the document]
2opremio: Ah, sorry, I misunderstood!

I implemented `MinSeqNum()` and plugged it in.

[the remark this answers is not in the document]
2opremio: PTAL

tamirms (approved):

PR #4302, opened 2022-03-22
xdr, keypair: Add helpers to create CAP-40 decorated signatures

Shaptic:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Add a new method to the `Keypair` interface: 

```golang
func SignDecoratedPayload(input []byte) (xdr.DecoratedSignature, error)
```

which creates a valid CAP-40 signature hint, combining the signer key hint with the payload. This method leverages a new abstraction layer, `xdr/decorated_signature.go` which creates the two types of signature hints in the protocol.

(Shoutout to @leighmcculloch for the PoC and test cases :100:) 

### Why
Supplements #4279.


xdr/decorated_signature.go
```diff
@@ -0,0 +1,44 @@
...
+		}
+	}
+
+	return DecoratedSignature{
+		Hint:      SignatureHint(hint),
+		Signature: Signature(sig),
+	}
+}
+
+func iMax(a, b int) int {
```
Shaptic, line 39: todo: add some generics to our monorepo :sob:

xdr/decorated_signature.go
```diff
@@ -0,0 +1,44 @@
+package xdr
+
+// NewDecoratedSignature constructs a decorated signature structure directly
+// from the given signature and key hint. Note that the key hint should
+// correspond to the key that created the signature, but this helper cannot
+// ensure that.
```
Shaptic, line 6: @leighmcculloch I borrowed your abstraction from the PoC because it just came out cleaner than sticking it into `Keypair`, but note the comment here: there's now a way to make invalid decorated signatures since `sig` and `keyHint` aren't necessarily related.

xdr/decorated_signature_test.go
```diff
@@ -0,0 +1,41 @@
...
+		t.Run(fmt.Sprintf("%d-byte", len(testCase.payload)), func(t *testing.T) {
+			decoSig := xdr.NewDecoratedSignature(signature, keyHint)
+			assert.EqualValues(t, keyHint, decoSig.Hint)
+			assert.EqualValues(t, signature, decoSig.Signature)
+
+			decoSig = xdr.NewDecoratedSignatureForPayload(
+				signature, keyHint, testCase.payload)
+			assert.EqualValues(t, testCase.expectedHint, decoSig.Hint)
+			assert.EqualValues(t, signature, decoSig.Signature)
+		})
```
leighmcculloch, line 39: ❗ This test is confusing because it is testing two independent functions together, where `expectedHint` applies only to one and not the other. In fact, it looks like `NewDecoratedSignature` gets tested twice inside the for loop with the same inputs. Also, the intent with `t.Run` is usually to separate the results of one unit test from another, but here two unit tests have been bundled inside a single `t.Run`. This seems like it is unnecessarily coupling. Can you test the two New functions separately?

xdr/decorated_signature.go
leighmcculloch: > there's now a way to make invalid decorated signatures since `sig` and `keyHint` aren't necessarily related.

There's always been a way to create decorated signatures that take any value. The fields of the type are public and can be set to anything at anytime.

Whether a decorated signature is invalid or not is difficult to pin down. Decorated signatures represent many different key types. I don't think the introduction of this New function introduces any problems.

leighmcculloch (commented):
A few requests (❗) and a suggestion (💡).

xdr/decorated_signature.go
```diff
@@ -0,0 +1,44 @@
...
+	sig []byte, keyHint [4]byte, payload []byte,
+) DecoratedSignature {
+	hint := [4]byte{}
+	j := iMax(0, len(payload)-4)
+
+	for i := 0; i < len(keyHint); i++ {
+		hint[i] = keyHint[i]
+		if j < len(payload) {
+			hint[i] ^= payload[j]
+			j++
```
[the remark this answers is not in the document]
Shaptic: Yeah I think I just stared at and iterated on this code for too long and it turned into a monstrosity.. I was trying to find a way to essentially emulate Python's slicing behavior which is so much cleaner, e.g. `payload[-4:]` covers all cases... I combined both approaches in 1b155d1b.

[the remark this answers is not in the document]
leighmcculloch: Looks great. Very clear, and elegant too. 👍

xdr/decorated_signature_test.go
Shaptic: Yep, the second set of tests was an afterthought and I didn't split them correctly. Fixed in latest!

keypair/full.go
```diff
@@ -109,10 +109,18 @@ func (kp *Full) SignDecorated(input []byte) (xdr.DecoratedSignature, error) {
...
-	return xdr.DecoratedSignature{
-		Hint:      xdr.SignatureHint(kp.Hint()),
-		Signature: xdr.Signature(sig),
-	}, nil
+	return xdr.NewDecoratedSignature(sig, kp.Hint()), nil
+}
+
+// SignDecoratedPayload returns a decorated signature using this key's hint and
+// the last four bytes of both the input and the signature on that input.
+func (kp *Full) SignDecoratedPayload(input []byte) (xdr.DecoratedSignature, error) {
```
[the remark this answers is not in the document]
Shaptic: Seems reasonable, done!

xdr/decorated_signature.go
Shaptic: > There's always been a way to create decorated signatures that take any value. The fields of the type are public and can be set to anything at anytime. 

I mean yeah, that's obviously true, but we're talking about SDK helpers that no longer "check your work" so to speak. But you're right that this is a mild and ambiguous distinction.

xdr/decorated_signature.go
```diff
@@ -0,0 +1,38 @@
+package xdr
+
+// NewDecoratedSignature constructs a decorated signature structure directly
+// from the given signature and key hint. Note that the key hint should
+// correspond to the key that created the signature, but this helper cannot
+// ensure that.
+func NewDecoratedSignature(sig []byte, keyHint [4]byte) DecoratedSignature {
+	return DecoratedSignature{
+		Hint:      SignatureHint(keyHint),
```
leighmcculloch, line 9: ❗  This is subtle, but the signer may not be a key.
```suggestion
// NewDecoratedSignature constructs a decorated signature structure directly
// from the given signature and hint. Note that the hint should
// correspond to the signer that created the signature, but this helper cannot
// ensure that.
func NewDecoratedSignature(sig []byte, hint [4]byte) DecoratedSignature {
	return DecoratedSignature{
		Hint:      SignatureHint(hint),
```

keypair/full.go
```diff
@@ -109,10 +109,18 @@ func (kp *Full) SignDecorated(input []byte) (xdr.DecoratedSignature, error) {
...
 
-	return xdr.DecoratedSignature{
-		Hint:      xdr.SignatureHint(kp.Hint()),
-		Signature: xdr.Signature(sig),
-	}, nil
+	return xdr.NewDecoratedSignature(sig, kp.Hint()), nil
+}
+
+// SignPayloadDecorated returns a decorated signature using this key's hint and
+// the last four bytes of both the input and the signature on that input.
```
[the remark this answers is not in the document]
Shaptic: Good catch, thanks

xdr/decorated_signature.go
Shaptic: Very subtle, thanks!

xdr/decorated_signature.go
```diff
@@ -0,0 +1,38 @@
...
+// bytes of the input that got signed. Note that the signature should be the
+// signature of the payload via the key being hinted, but this construction
+// method cannot ensure that.
+func NewDecoratedSignatureForPayload(
+	sig []byte, keyHint [4]byte, payload []byte,
+) DecoratedSignature {
+	hint := [4]byte{}
+	// copy the last four bytes of the payload into the hint
+	if len(payload) >= len(hint) {
+		copy(hint[:], payload[len(payload)-4:])
```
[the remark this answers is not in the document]
Shaptic: True!

xdr/decorated_signature_test.go
```diff
@@ -0,0 +1,50 @@
...
+		t.Run(
+			fmt.Sprintf("%d-byte decorated sig for payload", len(testCase.payload)),
+			func(t *testing.T) {
+				decoSig := xdr.NewDecoratedSignatureForPayload(
+					signature, keyHint, testCase.payload)
+				assert.EqualValues(t, testCase.expectedPayloadHint, decoSig.Hint)
+				assert.EqualValues(t, signature, decoSig.Signature)
+			})
+	}
+}
```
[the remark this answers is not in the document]
Shaptic: Ah, I see what you're getting at. Wdyt about d37806d3? Keeps the extensible table of tests (and adds a length-zero test), but completely separates it from the `NewDecoratedSignature` test, and minimizes the scattered "magic byte arrays".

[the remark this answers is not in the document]
Shaptic: (I recommend viewing the [whole file]https://github.com/stellar/go/blob/d37806d3dc2c795266dc535168a97c9c6da31e5b/xdr/decorated_signature_test.go) rather than the commit diff.)

leighmcculloch (approved):

Keleigh123:
Hi, i'm looking for a method to sign a prebuilt XDR in golang and im finding difficulty as most of the examples are outdated. Can someone please assist me

PR #4303, opened 2022-03-23
txnbuild: Add support for new CAP-21 preconditions.

### LINKED ISSUE #4282: txnbuild: Extend transaction building to allow CAP-21 preconditions
Go-specific issue for the parent, #4281.

Shaptic:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Adds support for new CAP-21 preconditions. This starts with a very simple API change:

```diff
@@ -212,7 +212,7 @@ type Transaction struct {
     sourceAccount SimpleAccount
     operations    []Operation
     memo          Memo
-    timebounds    Timebounds
+    preconditions Preconditions
 }

@@ -799,23 +799,23 @@ type TransactionParams struct {
     BaseFee       int64
     Memo          Memo
-    Timebounds    Timebounds
+    Preconditions Preconditions
 }
```

and evolves into all of the changes necessary to make that happen: the new `Preconditions` structure (creation, validation, XDR encoding, etc.) and coalescing between V1 conditions (timebounds only) and V2 conditions (CAP-21, incl. timebounds).

### Why
Closes #4282.


txnbuild/preconditions.go
```diff
@@ -0,0 +1,153 @@
...
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (cond *Preconditions) ValidateSigners() bool {
+	return len(cond.ExtraSigners) <= 2
```
sreuland, line 99: is this validation rule worth maintaining on client side, or let it get submitted and have core return txsub errors since it is source of truth on this aspect?

Shaptic: this will already blow up client-side because the XDR won't encode (not stellar core validation):

```xdr
SignerKey extraSigners<2>;
```

this was just being a little friendlier about it, since it's not fixed to be 2 w/in `Preconditions`. no strong opinion on keeping it, though.

txnbuild/preconditions.go
```diff
@@ -0,0 +1,153 @@
...
+func NewPreconditionsFromTimebounds(minTime, maxTime int64) Preconditions {
+	tb := NewTimebounds(minTime, maxTime)
+	return NewPreconditions(&tb)
+}
+
+// SetTimebounds enables the timebound precondition.
+//
+// Note that timebounds are a *required* precondition, but they're passed here
+// by pointer in order to align with `TransactionParams.Timebound`.
+func (cond *Preconditions) SetTimebounds(timebounds *Timebounds) error {
```
leighmcculloch, line 51: Is aligning with `TransactionParams.Timebound` important enough to make the API a bit wonky here? The consequence of this not aligning is unclear.

txnbuild/preconditions.go
```diff
@@ -0,0 +1,153 @@
...
+	if err := timebounds.Validate(); err != nil {
+		return err
+	}
+
+	if cond.timebounds != nil {
+		// only fail if they differ
+		if cond.timebounds.MinTime != timebounds.MinTime ||
+			cond.timebounds.MaxTime != timebounds.MaxTime {
+			return errors.New("timebounds set twice")
+		}
```
leighmcculloch, line 65: I think a runtime error like this is not great. How about we either:
1. Make a breaking change remove TimeBounds from the TransactionParams so it has to be set here. The upgrade path will be relatively simple.
2. Don't put time bounds inside Preconditions. Move it in a breaking API change.
3. Put it in both places and put a comment saying it is an error to set both TimeBounds and Preconditions. i.e. If you're using the new Preconditions type, you should probably just be using it and not mixing it with the old TimeBounds.

txnbuild/preconditions.go
```diff
@@ -0,0 +1,153 @@
...
+	return nil
+}
+
+func (cond *Preconditions) ValidateSigners() bool {
+	return len(cond.ExtraSigners) <= 2
+}
+
+// HasV2Conditions determines whether or not this has conditions on top of
+// the (required) timebound precondition.
+func (cond *Preconditions) HasV2Conditions() bool {
```
leighmcculloch, line 104: Can this function be not exported? Users who need to know, are probably working at the level of needing to understand the XDR and can work with the XDR package. Also, this seems like an implementation detail of decisions the BuildXDR function is making.

txnbuild/preconditions_test.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+	tb = NewTimebounds(27, 42)
+	pc = Preconditions{
+		timebounds:                 &tb,
+		Ledgerbounds:               &Ledgerbounds{27, 42},
+		MinSequenceNumber:          &seqNum,
+		MinSequenceNumberAge:       xdr.Duration(27),
+		MinSequenceNumberLedgerGap: 42,
+		ExtraSigners:               Signers[:1],
+	}
```
leighmcculloch, line 44: Imx it is quite difficult to maintain tests that use a lot of shared inputs, and it improves test readability if you put these inputs into the test, so that each test reads top-to-bottom like an application using the SDK would read.

txnbuild/preconditions_test.go
```diff
@@ -0,0 +1,173 @@
...
+				newCond, newPc := clone(xdrCond, pc)
+				newCond.V2.MinSeqNum = nil
+				newPc.MinSequenceNumber = nil
+				return newCond, newPc
+			},
+		},
+	}
+	for _, testCase := range preconditionModifiers {
+		t.Run(testCase.Name, func(t *testing.T) {
+			xdrCond, precond := testCase.Modifier()
```
leighmcculloch, line 112: Is this test modifying global variables? This feels unintuitive and difficult to tell what state each input is in. I think this table driven test would be much easier to understand if the values being tested on were specified in each row of the table.

leighmcculloch: Ok, I see now, it isn't modifying global vars, it's cloning global vars? Although another test further down is indeed modifying global vars. I think this code is too clever at a steep cost of readability.

txnbuild/transaction.go
```diff
@@ -770,11 +770,11 @@ func transactionFromParsedXDR(xdrEnv xdr.TransactionEnvelope) (*GenericTransacti
 		},
 		operations: nil,
 		memo:       nil,
-		timebounds: Timebounds{},
 	}
 
 	if timeBounds := xdrEnv.TimeBounds(); timeBounds != nil {
-		newTx.simple.timebounds = NewTimebounds(int64(timeBounds.MinTime), int64(timeBounds.MaxTime))
+		newTx.simple.preconditions = NewPreconditionsFromTimebounds(
+			int64(timeBounds.MinTime), int64(timeBounds.MaxTime))
```
leighmcculloch, line 777: Will this code also need to set the other preconditions too?

txnbuild/transaction.go
```diff
@@ -794,26 +794,26 @@ func transactionFromParsedXDR(xdrEnv xdr.TransactionEnvelope) (*GenericTransacti
...
-	BaseFee              int64
-	Memo                 Memo
-	Timebounds           Timebounds
+	SourceAccount           Account
+	IncrementSequenceNum    bool
+	Operations              []Operation
+	BaseFee                 int64
+	Memo                    Memo
+	Timebounds              Timebounds
+	AdditionalPreconditions Preconditions
```
leighmcculloch, line 806: Suggested dropping the prefix. It doesn't add much value, especially since it is basically replacing Timebounds.
```suggestion
	Preconditions Preconditions
```

txnbuild/transaction.go
```diff
@@ -823,15 +823,26 @@ func NewTransaction(params TransactionParams) (*Transaction, error) {
 		return nil, errors.Wrap(err, "could not obtain account sequence")
 	}
 
+	// Because both V1 and V2 preconditions allow timebounds, and we don't want
+	// to introduce a breaking change by dropping
+	// `TransactionParams.Timebounds`, nor require users to set up the
+	// `AdditionalPreconditions`, we need to coalesce the two values here.
+	if params.Timebounds != (Timebounds{}) {
+		err = params.AdditionalPreconditions.SetTimebounds(&params.Timebounds)
```
leighmcculloch, line 831: This could cause unexpected behavior in applications that reuse a Preconditions object across multiple transactions. Since SetTimebounds errors when the value passed in is different, setting a new top-level time bounds on a second tx, and reusing the Preconditions would cause an error in a situation that would be surprising. Can we do this in a way where we aren't modifying the users passed in Preconditions?

leighmcculloch (commented):
👏🏻. A few comments inline.

txnbuild/preconditions.go
sreuland: I think #1, and if timebounds=nil here, instead of err, could it default to 'empty/no preconditions' =  (timebounds.minTime=0 and timebounds.maxTime=0) = INFINITE_TIMEOUT? the java sdk tx builder is shaping up to that...

leighmcculloch: I think I'm learning for 1 too. You don't strictly need a time bounds, but it is a best practice to always use one, which is why we error on one not being set explicitly.

txnbuild/preconditions.go
sreuland: interesting, on the java sdk side, I checked the generated java from xdrgen, [PreconditionsV2.java](https://github.com/stellar/java-stellar-sdk/pull/413/files#diff-1c2812de0a0b421d83904a598bacc4351a3c80d9701f4d7260ca9cf1b4f7ad00), it doesn't capture the `SignerKey extraSigners<2>` size limit like it is on go with `xdrmaxsize`, I may need to hardcode that as a constant or just document that it's by default 2 and submission will be final validation

txnbuild/preconditions.go
Shaptic: I definitely would prefer (1), but even though the upgrade path is _simple_ (changing `Timebounds: xxx` to `Preconditions: NewPreconditions(xxx)`), it's massively _impactful_ (there are a TON of places this would need to be updated, even internally... that's why this convoluted approach came about :thinking: 

(2) and (3) have their own problem: now the XDR can't be built from the `Precondition` structure, so the abstraction layer doesn't even make sense to maintain.

Thoughts on that in the new light?

txnbuild/preconditions_test.go
Shaptic: Huh, it's really interesting that you say that, because imx it's the exact opposite. I look at tests that appear to be nearly identical and it takes time to decipher where the deltas are. With the structure I have here, it's very apparent exactly how the structures are being modified to test each use case :thinking:

txnbuild/preconditions.go
sreuland: if TimeBounds goes into PreconditionsV2, I think a follow-on question is can we have a best practice that client sdks can follow for consistent validation of PreconditionsV2 going forward? Eager validation attempts at client likely become stale and incur more maintenance to keep in sync with server over time, wondering if no validation on client side is best approach, such that even an empty PreconditionsV2 is valid, which effectively means infinite TimeBounds(0,0) correct?

txnbuild/preconditions_test.go
Shaptic: There are something like 8! combinations of preconditions, right? This tests only a tiny subset of them. Adding a test for a new combination under this structure is only a few lines of code, whereas keeping each test completely in its own block, variables and all, would require copy-pasting entire pages of code and _then_ modifying a few lines.

I do agree that it was sloppy to keep the "base" structures as global variables, and I've improved that in df7b4c27, but I honestly can't see how readability would improve if I broke out the subtests into individual tests, each with nearly-identical `xdr.Preconditions` and `Preconditions` values.

txnbuild/transaction.go
Shaptic: The issue is related to https://github.com/stellar/go/pull/4303#discussion_r834497518. I've decided to do the breaking change in 468e7de4 so that we can see the impact. It cleans up the code related to this PR quite a bit, but at cost to a wider scope of changes.

txnbuild/preconditions.go
Shaptic: Two things here:

We can definitely rely on the XDR encoder and/or Core to do all validation necessary, but we can optionally be helpful at the SDK layer. Yes it's more maintenance, but it's also SDK ergonomics. The Laboratory, for example, does a **ton** of extra validation that usually happens at the XDR level in order to keep error messages user friendly. As long as we keep it optional (i.e. downstream SDKs don't _have_ to replicate it), I'm okay with including it.

Separately, though, we absolutely should continue to require timebounds. Infinite timebounds should be explicit, exactly as they are today. We need to actively discourage them, even though the protocol technically allows it. Timebounds are an important part of [error handling](https://developers.stellar.org/docs/tutorials/handling-errors/), and technically transactions can quite literally live in the ether forever during high network load if they have infinite timebounds.

txnbuild/transaction.go
Shaptic: It would, great catch thank you. And it reminds me that I need more transaction xdr encode/decode tests.

Shaptic:
@leighmcculloch @sreuland I made the breaking change to move `TransactionParams.Timebounds` -> `TransactionParams.Preconditions.Timebounds` in 468e7de. Like I mentioned, it's simple but broad, so let me know what you guys think.

txnbuild/timebounds.go
```diff
@@ -48,10 +48,6 @@ func (tb *Timebounds) Validate() error {
 	return nil
 }
 
-func (tb *Timebounds) IsEmpty() bool {
-	return tb == nil || *tb == Timebounds{}
-}
-
```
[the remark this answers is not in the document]
Shaptic: Huh.. I wrote this and then removed it.. why is it showing up in the diff?

Shaptic:
I've also forked off an alternative, non-breaking method. This one keeps `Preconditions.timebounds` private, but makes it a non-pointer. The branch diff is here: [horizon-protocol-19...Shaptic:txnbuild-tx-conds-alternative](https://github.com/stellar/go/compare/horizon-protocol-19...Shaptic:txnbuild-tx-conds-alternative?expand=1), but the main change I think we all care about is [here](https://github.com/stellar/go/compare/horizon-protocol-19...Shaptic:txnbuild-tx-conds-alternative?expand=1#diff-bd94368e661d629d7679f3b3eb9da1f18e91ca7fc31a3f418f53109677df631aR826-R843) (no more coalescing or pointer voodoo).

txnbuild/preconditions.go
Shaptic: That's interesting - seems like a limitation of the generator? :thinking: definitely think it's worth a better error message from the SDK if possible

txnbuild/preconditions.go
```diff
@@ -0,0 +1,120 @@
...
+		err = cond.Ledgerbounds.Validate()
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (cond *Preconditions) ValidateSigners() bool {
```
leighmcculloch, line 65: 💡 Can we unexport this? It feels like an implementation detail to the `Validate` function. We don't have validate methods for every field either. It doesn't look like it delivers much value being a function on its own. Any reason you can't just check the length inside the main Validate method?

txnbuild/preconditions.go
```diff
@@ -0,0 +1,120 @@
...
+
+// NewPreconditions creates a set of preconditions with timebounds enabled
+func NewPreconditions(timebounds Timebounds) Preconditions {
+	cond := Preconditions{Timebounds: timebounds}
+	return cond
+}
+
+func NewPreconditionsWithTimebounds(minTime, maxTime int64) Preconditions {
+	return NewPreconditions(NewTimebounds(minTime, maxTime))
+}
```
leighmcculloch, line 41: ❗ I don't think we need these New functions. NewPreconditions is a little odd it accepts a time bounds, especially when there's another function NewPreconditionsWithTimebounds that doesn't accept a Timebounds. Also, it is easier imo to construct the struct manually than use these, and it makes more code look the same.

For example, taking this piece of code from one of the tests and modifying it, it is more consistent than using a New function:
```go
cond := Preconditions{
	Timebounds: NewTimeout(300),
	Ledgerbounds: &Ledgerbounds{0, 1},
	MinSequenceNumber: nil,
	MinSequenceNumberAge: xdr.Duration(10),
	MinSequenceNumberLedgerGap: 2,
```

Also, one liners are still trivial:
```go
cons := Preconditions{Timebounds: NewTimeout(300)}
```

txnbuild/preconditions_test.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+// TestPreconditionsValidation ensures that validation fails when necessary.
+func TestPreconditionsValidation(t *testing.T) {
+	pc := NewPreconditionsWithTimebounds(27, 42)
+	pc.ExtraSigners = Signers
+
+	t.Run("too many signers", func(t *testing.T) {
+		pc.ExtraSigners = Signers
+		assert.Error(t, pc.Validate())
```
leighmcculloch, line 135: ❗ This test looks in complete, as it only tests signers. It also looks like you're setting ExtraSigners twice.

txnbuild/preconditions_test.go
```diff
@@ -0,0 +1,178 @@
...
+	for _, testCase := range preconditionModifiers {
+		t.Run(testCase.Name, func(t *testing.T) {
+			xdrPrecond, precond := testCase.Modifier()
+			assert.NoError(t, precond.Validate())
+
+			expectedBytes, err := xdrPrecond.MarshalBinary()
+			assert.NoError(t, err)
+
+			actualBytes, err := precond.BuildXDR().MarshalBinary()
+			assert.NoError(t, err)
```
leighmcculloch, line 121: ❗ It would be worth round tripping as well unless you plan to test that elsewhere. Assumably any encoded should be rehydratable to an identical value via parsing a transaction.

txnbuild/preconditions_test.go
```diff
@@ -0,0 +1,178 @@
...
+
+	clone := xdr.Preconditions{}
+	if err = clone.UnmarshalBinary(binary); err != nil {
+		panic(err)
+	}
+
+	return clone
+}
+
+func clonePreconditions(precond Preconditions) Preconditions {
```
leighmcculloch, line 157: 💡 The way this set of tests work by cloning things is particularly complicated. How about just writing out each input verbosely? It would be much easier to understand and maintain.

The inspiration for why I'm suggesting this is that DRY tests are often harder to maintain:
- https://thoughtbot.com/blog/the-case-for-wet-tests
- https://evhaus.medium.com/using-dry-wet-damp-code-6ab3c8245bb2

leighmcculloch (commented):
🎉 I like the breaking change moving TimeBounds. While it is a breaking change, I think the alternatives all come with significant tradeoffs.

I left some requests (❗), and also some suggestions (💡).

txnbuild/preconditions.go
Shaptic: I agree that the naming sucks, tbh, but there's an important subtlety in preventing direct creation: timebounds are **required**. This is technically covered by the [small change](https://github.com/stellar/go/pull/4303/files#diff-f1fd167b9ec077c8e6969a5afea23de8fc45a9480c0f6521bccffb138be86b62R31) to Timebounds.Validate(), but it's a hard detail to rely on. How do we reconcile this? Or do we just not care?

leighmcculloch: It looks like the case is already covered as you say. To this end I think we could also make TimeBounds a pointer in the Preconditions struct, and error if it is nil. I'm not sure if someone was trying to avoid pointers in TransactionsParams or ease of copying the value without shared components, but regardless there are other pointers (in the form of slices and interfaces) anyway, so that seems unimportant.

txnbuild/preconditions.go
sreuland: > @leighmcculloch @sreuland I made the breaking change to move `TransactionParams.Timebounds` -> `TransactionParams.Preconditions.Timebounds` in [468e7de](https://github.com/stellar/go/commit/468e7de48982a96da3cebae4a27e726f86d1bd40). Like I mentioned, it's simple but broad, so let me know what you guys think.

yep, following the same refactor on the java sdk side, will have that up soon for look

txnbuild/preconditions.go
Shaptic: Making `TimeBounds` a pointer is an ergonomics issue: You can't do `Preconditions{Timebounds: &NewTimebounds(x, y)}` anymore since you can't `&` a return value. I think the error case is sufficient, too: it should be very apparent what went wrong.

txnbuild/transaction.go
Shaptic: Done in d54243ad!

txnbuild/preconditions.go
sreuland: ok, sounds applicable for best practices in any client sdk using the xdr, so, sdk's check PreconditionsV2.TimeBounds is set to something by caller and leverage xdr encoding/decoding builtins as minimum baseline for validation on PreconditionsV2, thanks!

txnbuild/preconditions_test.go
leighmcculloch: It's a tradeoff for sure. Even if DRY tests look good on day one, imx they are harder to maintain over time, because the way they are structured are making assumptions about the tests we'll have to write in the future or the ways we'll need to modify these tests, requiring test refactors, etc. My comment was a suggestion, not a request.

If you are going to keep the cloning, might I suggest replacing it with a helper that creates the original structure. This way you don't need to write cloning code. (Cloning code is unavoidably brittle code.)

Shaptic: That's a good point about test evolution. Cross that bridge when we get to it, then? :laughing: Either way, just want to reiterate that I really appreciate the feedback and suggestions, as it makes me challenge my own assumptions and get a better understanding of the trade-offs I may be making w/o realizing.

You're on the money about cloning: I hated that aspect of it. I opted to make "test fixtures" with a helper function in 84c882e8c9bd52ef93954392619d2e686f503882; I think it's much less brittle now overall!

Shaptic:
Last round of feedback, I hope, @leighmcculloch and @stellar/horizon-committers? :pray: 

To make reviews easier, here are the files with relevant (i.e. not just adapting to the breaking `TransactionParams` delta) changes:

 * new files: txnbuild/ledgerbounds.go, txnbuild/preconditions.go, txnbuild/preconditions_test.go
 * most of txnbuild/transaction.go
 * the top of txnbuild/transaction_test.go
 * txnbuild/timebounds.go
 * xdr/transaction_envelope.go
 * txnbuild/CHANGELOG.md

txnbuild/ledgerbounds.go
```diff
@@ -0,0 +1,23 @@
+package txnbuild
+
+import "github.com/stellar/go/support/errors"
+
+// Ledgerbounds represent a transaction precondition that controls the ledger
+// range for which a transaction is valid. Setting MaxLedger = 0 indicates there
+// is no maximum ledger.
+type Ledgerbounds struct {
```
leighmcculloch, line 8: Typo: Ledgerbounds => LedgerBounds

Shaptic: Consistency with `txnbuild.Timebounds`? :thinking:

leighmcculloch: You know, we could safely fix the Timebounds typo as part of this protocol release. By doing these things:
- We're already moving the Timebounds field, so simply name the new field TimeBounds.
- We can rename txnbuild.Timebounds to txnbuild.TimeBounds.
- We can use a type alias to specify that txnbuild.Timebounds is the same type as txnbuild.TimeBounds.

Shaptic: I'm good with that! We still have inconsistencies (), and we can't rename the constructors w/o more breaking changes (e.g. `NewTimebounds`), but it's still progress :construction_worker_man:. Both renames done in latest push :+1:

txnbuild/timebounds.go
```diff
@@ -9,26 +9,28 @@ import (
...
-type Timebounds struct {
+// Create a TimeBounds struct using one of NewTimebounds(), NewTimeout(), or NewInfiniteTimeout().
+type TimeBounds struct {
 	MinTime  int64
 	MaxTime  int64
 	wasBuilt bool
 }
 
-// Validate for Timebounds sanity-checks the configured Timebound limits, and confirms the object was built
+// type Timebounds = TimeBounds
```
leighmcculloch, line 27: Why is this commented? 
```suggestion
type Timebounds = TimeBounds
```

Shaptic: I... I don't know :laughing: I did a ctrl+f and then un-commented it, but that disappeared somehow

leighmcculloch (approved):

txnbuild/preconditions.go
```diff
@@ -0,0 +1,143 @@
+package txnbuild
+
+import (
+	"github.com/stellar/go/support/errors"
+	"github.com/stellar/go/xdr"
+)
+
+// Preconditions is a container for all transaction preconditions.
+type Preconditions struct {
```
sreuland, line 9: what is the benefit for using this abstraction model rather than using xdr.PreconditionsV2 direct? This and TimeBounds and LedgerBounds appear to be 1-1 proxy of their xdr counterparts?

Shaptic: Hm, didn't see this comment for some reason. They're 1-to-1 now, but the txnbuild package in general avoids exposing XDR to SDK consumers.

sreuland: ok, see that de-coupling at top level, sounds good, there are nested references from here into xdr types such as `SignerKey` and `Duration`, and was looking at it from bottom up, thanks!

PR #4307, opened 2022-03-25
txnbuild: Complete rename, avoid using XDR types in `TransactionParams`.

Shaptic:
This renames two fields on `TransactionParams`, which should've been part of #4303: 
  - `TransactionParams.Timebounds` -> `TransactionParams.TimeBounds` and
  - `TransactionParams.Ledgerbounds` -> `TransactionParams.LedgerBounds`

Besides that, it drops all XDR types within `TransactionParams`:
  - `TransactionParams.ExtraSigners` is now `[]string` instead of `[]xdr.SignerKey` and
  - `TransactionParams.Duration` is now `int64` instead of `xdr.Duration`


txnbuild/preconditions.go
```diff
@@ -123,21 +132,33 @@ func (cond *Preconditions) FromXDR(precondXdr xdr.Preconditions) {
...
+					return err
+				}
+				cond.ExtraSigners[i] = signer
+			}
+		}
 
 	case xdr.PreconditionTypePrecondNone:
-	default: // panic?
+	default:
+		return errors.New("unsupported precondition type: " + precondXdr.Type.String())
```
leighmcculloch, line 150: +1

leighmcculloch (approved):

Issue #4263, opened 2022-03-08
Allow limiting number of path finding cycles per second


bartekn:
It should be able to limit a number of cycles a single path finding instance/Horizon instance can execute per second. This should allow to cap the CPU usage so that it doesn't consume 100% under heavy load connected to path finding. There should also be a simple stress-testing tool that will help checking what the limit should be set to on a specific instance type.

tamirms:
I would prefer if we implemented a limit on the number of concurrent calls to the path finding function. Any attempts to call the function beyond the limit will be rejected with an error. I think this would be easier to implement.

sreuland:
we are really pushing on the abilities of rules in lb/proxy rate limits based on anonymous client access, seems like it will hit a wall soon, may be worthwhile in parallel to start looking at authorized API access options on Horizon to address the scale of growing traffic in longer term. The concept has come up in [prior slack ](https://stellarfoundation.slack.com/archives/G01FUHTUQ8Z/p1646850924741439)here.

Two options could be:
 * move the high traffic endpoints behind AWS API Gateway and configure it with a [custom Lambda Authorizer](https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-use-lambda-authorizer.html) and check for stellar-account signed JWT in http `Authorization` header and crosschecks with whitelist of accepted account ids from somewhere tbd 
 
 * Implement the same JWT authorization  routine as Horizon http middleware filter, it checks `Authorization` header for stellar-account signed JWT andcrosschecks with whitelist of accepted account ids from somewhere tbd

PR #4310, opened 2022-03-28
services/horizon: Add a rate limit for path finding requests.

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Fixes https://github.com/stellar/go/issues/4263
Add a per second rate limit for path finding requests backed by https://pkg.go.dev/golang.org/x/time/rate .

### Why

Unlimited path finding requests can max out the CPU and starve other horizon requests.

### Known limitations

[N/A]


marcinx:
It looks like this is a platform wide setting. Would one single party be able to consume all path finding requests and starve out everyone else? Would it make sense to make this setting to be applied per IP in the future to prevent that (to a certain degree, considering IP rotation etc)?

bartekn (approved):

tamirms:
@marcinx 

> Would one single party be able to consume all path finding requests and starve out everyone else? 

Yes, that is possible.  Horizon does have ip based rate limiting using the `--per-hour-rate-limit` configuration variable but that rate limit includes all horizon requests, not just path finding.

IMO having more fine grained ip based rate limits specifically for path finding requests would be best implemented at the load balancer level (either using a reverse proxy like nginx or, a cloud based load balancer like AWS ALB or cloudflare).

Issue #4305, opened 2022-03-25
Horizon: Protocol 19 Integration Tests


Shaptic:
Breakdown and Rollout: [Horizon & Protocol 19](https://docs.google.com/document/d/1qZ6o14JBOZpyDG6rNEd8Bj-sCzpeZgyx7Vtl1SMKI28). See #4261 for the master issue.

----------------------

Now that Protocol 19 feature work is mostly complete in the [horizon-protocol-19](https://github.com/stellar/go/tree/horizon-protocol-19) branch and the Go SDK is up-to-date, we can turn our focus to stability. We need end-to-end integration tests for the new features:

- [x] Generating full account extensions for state verifier tests (#4304)
- [x] Add stellar-core builds that include Protocol 19 work (#4312)
- [x] Test new CAP-21 account fields (#4322)
- [x] Test new transaction preconditions (#4320)
- [x] Test transaction submission queue (#4327)
- [x] Add state verifier tests for all of the above
- [x] Test Starlight (Leigh)

jcx120:
Are there tests we should include for regressing re-ingestion and regular ingestion to keep up with ledger?

Shaptic:
@jcx120 added an item re: state verifier. I think we get regression testing "automatically" by running our test suite without an actual Protocol 19 build of Stellar Core (aka what's been happening until #4306 gets merged), but moving forward yes, we need to ensure each new feature has the relevant ingestion testing (I think it did in their individual PRs).

2opremio:
> Add state verifier tests for all of the above

@Shaptic how do you suggest we do this? Changing TestStateVerifier to submit transactions with preconditions? Or are you just referring to the new Account fields?

sreuland:
@2opremio , [PR #4332](https://github.com/stellar/go/pull/4332) is up for new integration test coverage on those remaining transaction preconditions.

Shaptic:
@2opremio You're right that account extensions were supposed to have been tested! And preconditions aren't part of state, so we don't need to test those.

Unfortunately, the account extensions actually _weren't_ tested after #4304 because this line:
```go
{randxdr.FieldEquals("created.data.account.ext.v1.ext.v2.ext.v"), randxdr.SetU32(0)},
```
kept the extension disabled and I didn't realize that during the review :facepalm: 

This is resolved by #4337.

PR #4322, opened 2022-04-06
horizon: Add integration test for new protocol 19 account fields

2opremio:
Also, improve the protocol check of tests

Addresses part of https://github.com/stellar/go/issues/4305

### DIFF
--- a/services/horizon/internal/integration/transaction_preconditions_test.go
+++ b/services/horizon/internal/integration/transaction_preconditions_test.go
@@ -1,9 +1,11 @@
 package integration
 
 import (
+	"strconv"
 	"testing"
 	"time"
 
+	sdk "github.com/stellar/go/clients/horizonclient"
 	"github.com/stellar/go/keypair"
 	"github.com/stellar/go/services/horizon/internal/test/integration"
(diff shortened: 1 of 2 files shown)


services/horizon/internal/integration/transaction_preconditions_test.go
```diff
@@ -86,3 +88,29 @@ func buildTXParams(master *keypair.Full, masterAccount txnbuild.Account, sourceA
 		},
 	}
 }
+
+func TestTransactionPreconditionsAccountFields(t *testing.T) {
```
sreuland, line 92: this asserts the case of an existing account that would already be on the ledger(history) with no v3 ext, should we do another test case for creating a new account with some v3 values and retrieving that one also?

sreuland (approved):

services/horizon/internal/integration/transaction_preconditions_test.go
2opremio: I am confused. AFAIU this test checks the new v3 ext fields, but maybe I am missing something. Can you elaborate?

PR #4327, opened 2022-04-08
horizon: Add transaction submission test for Protocol 19

2opremio:
Addresses part of #4305


Shaptic (approved):
Very nice :clap: Do you think we need more complicated sets of sequences?

PR #4337, opened 2022-04-13
services/horizon: Use `bigint` over `timestamp` for the new `accounts.sequence_time` column.

Shaptic:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Three things:
 - Generating V3 account extensions is _actually_ enabled now, after my mistake in overlooking #4304 
 - The Protocol 19 database migrations have been merged into a single file
 - The `accounts.sequence_time timestamp` column is now a `bigint`

### Why

(Copied from 7915eb234f3a3a0452247d28c5cdbde37f723178):

This is necessary because PostgreSQL puts "reasonable" limitations on how dates can be represented. Because the state verifier test will generate random int64s for the V3 account extension field introduced in [CAP-21](https://stellar.org/protocol/cap-21), this will blow up on inserting into the DB with errors like:

    pq: date/time field value out of range: "120911623444-02-01 21:35:20Z"

because PostgreSQL [only allows](https://stackoverflow.com/a/36446977) years up to 5874897.

Since this field is not user-controlled and comes directly from Stellar Core, we will never see this in practice.

However, it's a more durable fix to switch to `bigint` and stop the extra layer of "interpretation" of this value, treating it like it is: a 64-bit integer that happens to represent a UNIX timestamp.

### Known limitations
I'm not sure what would happen if `randxdr.SetPositiveNum64` generated a value exceeding `int64`, but this would be a transient error since, again, this **cannot** happen in practice.


erika-sdf (approved):
LGTM. Thanks for tracking this down.

PR #4339, opened 2022-04-14
Change `min_account_sequence_age` from `bigint` to string

Shaptic:
### What
This changes the column from a `bignum` to a `varchar(20)` to accommodate unsigned 64-bit duration values.

### Why
The XDR uses `xdr.Duration` for this field which maps to `xdr.Uint64`. Since `bignum` is signed, and this field is user-controlled, we can't use that data type for fear of overflow. We should instead use a `text`, or a `varchar(20)` where 20 is the length of the longest uint64 value possible.

### Known limitations
n/a


leighmcculloch:
> Since `bignum` is signed, and this field is user-controlled, we can't use that data type for fear of overflow.

On the flip side using a string increases the possible set of invalid values, and you can still put a negative number in a string. 

Would it be simple to check overflow by checking if the number is >= 0 and <= 9223372036854775807.

Shaptic:
@sreuland I spent a while poking around and there's one existing place where there's a user-controlled `xdr.Uint64` value that we process: transaction timebounds. (There's another place: claimable balance time predicates, but we don't store the predicates in the database.)

I thought this would lead to a similar bug (i.e. postgres blowing up if you pass too-big of timebounds), but the way that the existing codebase handles this is with the following code: https://github.com/stellar/go/blob/ae4b37a2ada56c57f1034bd8bb7a3c29367d4566/services/horizon/internal/db2/history/transaction_batch_insert_builder.go#L137-L140

In other words, it just... caps it at `MaxInt64`.

Technically, this is a reinterpretation of the ingested ledger data. But setting fields to these values is nonsensical enough as it is... I'd be fine with reworking the code to account for that, but technically using a string lets us stay true to it.

sreuland:
> Technically, this is a reinterpretation of the ingested ledger data. But setting fields to these values is nonsensical enough as it is... I'd be fine with reworking the code to account for that, but technically using a string lets us stay true to it.

@Shaptic, agreed on preserving data, the approach taken here of capturing all bits into opaque storage is good, allows for interpretations afterwards, users may be storing meaningful constants or whatever huge values as the type allows, and if they see different values later from horizon API, that would be worse.

PR #4342, opened 2022-04-14
services/horizon: Cap the minSeqAge precondition at MaxInt64

Shaptic:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
This caps the minSeqAge precondition field (of type `xdr.Duration = Uint64`) value to be an int64.

This is an alternative to #4339.

### Why
This ensures negative values don't make it into the database. We do the same thing with timebounds currently.

### Known limitations
You can't use the values in the interval `(MaxInt64, MaxUint64]`, but those are absurd anyway.

### DIFF
--- a/services/horizon/internal/db2/history/transaction_batch_insert_builder.go
+++ b/services/horizon/internal/db2/history/transaction_batch_insert_builder.go
@@ -241,7 +241,7 @@ func formatDuration(d *xdr.Duration) null.Int {
 	}
 
 	if *d > math.MaxInt64 {
-		*d = xdr.Duration(math.MaxInt64)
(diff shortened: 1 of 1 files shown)


sreuland:
I thought the approach taken on [#4339](https://github.com/stellar/go/pull/4339) to preserve this data bit for bit as char blob was more compelling design, in principle, horizon stays neutral, clients get to interpret.  Maybe avoid some edge cases, If clients chose to stuff a huge constant into minAge on a tx which has logical meaning in their application, and then they get something different from horizon, maybe that is worse?

Shaptic:
Closing in lieu of #4339.

PR #4349, opened 2022-04-21
services/horizon: Pass through nil ExtraSigners to avoid nil pointer deref

Shaptic:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
When the `ExtraSigners` array is `nil`, parse it as `nil`.

### Why
Otherwise, it's parsed into an empty array. Then, when we build the transaction row for the response JSON:

https://github.com/stellar/go/blob/18ab1d80e5e0d4e17e0d21ce5be652618705948d/services/horizon/internal/resourceadapter/transaction.go#L98-L100

there's a nil pointer dereference because `dest.Preconditions == nil`, because `HasPreconditions()` failed, because it checked that `len(row.ExtraSigners) > 0` which is `false`, and yet since `row.ExtraSigners != nil`, so it will enter this conditional and :boom: 

### Known limitations
n/a

Thank you @overcat for the bug report!


services/horizon/internal/db2/history/transaction_batch_insert_builder.go
```diff
@@ -249,6 +249,10 @@ func formatUint32(u *xdr.Uint32) null.Int {
 }
 
 func formatSigners(s []xdr.SignerKey) pq.StringArray {
+	if s == nil {
+		return nil
+	}
+
```
leighmcculloch, line 255: Note that this should have no material effect on the XDR because a nil var length array in XDR is a zero length array. There is no distinction about them, unless we've optionalized the array with a pointer in the XDR, which we didn't for extraSigners.
```xdr
SignerKey extraSigners<2>;
```

leighmcculloch (commented):
It's not really clear to me how this fixes the problem, or where exactly the problem is. Where does the nil ptr deref occur?

leighmcculloch:
Ah, I think the issue is these lines:
https://github.com/stellar/go/blob/18ab1d80e5e0d4e17e0d21ce5be652618705948d/services/horizon/internal/resourceadapter/transaction.go#L58-L60

The `HasV2Preconditions` returns `false`, and so we never set `dest.Preconditions` with a value, but then later try to use the value here, but `dest.Preconditions` is `nil`:
https://github.com/stellar/go/blob/18ab1d80e5e0d4e17e0d21ce5be652618705948d/services/horizon/internal/resourceadapter/transaction.go#L98-L100

leighmcculloch:
Regardless, I think the fix is good, and will use less db space I assume to store `NULL` instead of an empty signers?

leighmcculloch (approved):

services/horizon/internal/integration/transaction_preconditions_test.go
```diff
@@ -334,3 +337,64 @@ func TestTransactionPreconditionsAccountV3Fields(t *testing.T) {
...
+
+	// taken from txnbuild.marshallBinary
+	var txBytes bytes.Buffer
+	envelope.V1.Signatures = []xdr.DecoratedSignature{sig}
+	_, err = xdr.Marshal(&txBytes, envelope)
+	tt.NoError(err)
+	b64 := base64.StdEncoding.EncodeToString(txBytes.Bytes())
+
+	_, err = itest.Client().SubmitTransactionXDR(b64)
+	tt.NoError(err)
```
sreuland, line 399: may be worthwhile in the test to also fetch the TX from API to verify the NONE precond gets serialized through db/api/json stack correctly too?

`itest.Client().TransactionDetail(txResp.Hash)`

sreuland:
> The HasPreconditions returns false, and so we never set dest.Preconditions with a value, but then later try to use the value here, but dest.Preconditions is nil

that looks like the root cause,  the if statement could be removed and `dest.Preconditions` can be set to empty pre-condtions struct all the time.

Shaptic:
@leighmcculloch @sreuland yeah, that's right. Originally, `row.ExtraSigners != nil` always since it will do `make([]string, 0)` in the `nil` case, but `HasPreconditions()` checks if `len(row.ExtraSigners) > 0` which is different. These conditions need to match for this not to blow up. There are a few options:

1. Always initialize Preconditions, as Shawn suggested. We can instead do a check at the *end* of the function and remove the field if it shouldn't have been used. This is more defensive against nil dereferences.
2. Add the `nil` passthrough (the original delta)
3. Check `len(row.ExtraSigners) > 0` in both `HasPreconditions` and the subsequent setter, since `len(nil)` is valid and is always `0`.
4. All of these, for robustness.

I've added (1) and (3) in 4096c086.

services/horizon/internal/integration/transaction_preconditions_test.go
Shaptic: Good idea!

sreuland (approved):

services/horizon/internal/resourceadapter/transaction.go
```diff
@@ -136,6 +133,11 @@ func PopulateTransaction(
 	dest.Links.Succeeds = lb.Linkf("/transactions?order=desc&cursor=%s", dest.PT)
 	dest.Links.Precedes = lb.Linkf("/transactions?order=asc&cursor=%s", dest.PT)
 
+	// If we didn't need the structure, drop it.
+	if !row.HasPreconditions() {
+		dest.Preconditions = nil
```
sreuland, line 138: is this necessary, is it because an empty Preconditions is considered not valid json/resource view?

Shaptic: It's more like: if we never used it, get rid of it. But only get rid of it after the fact, just in case some weird bug sneaks that would've caused a nil deref.

PR #4357, opened 2022-04-28
services/horizon: Properly allow nullable Protocol 19 account fields

Shaptic:
### What
Properly parses null sequence time/ledger values from database account entries into an account resource.

### Why
The columns are `NULL` by default because they cannot and will not be set for accounts that haven't done anything prior to Protocol 19. This means that they will explode when loaded into the `AccountEntry` structure because that structure requires non-null uint32 and uint64 fields. This 

### Known limitations
The alternative is `DEFAULT 0` on the column (see 782fec322a308e52b4753b5faa2d14a4cc3f510a), but then the migration will take a (possibly much) longer time.


services/horizon/internal/ingest/verify.go
```diff
@@ -422,13 +434,7 @@ func addAccountsToStateVerifier(ctx context.Context, verifier *verify.StateVerif
...
 							NumSponsoring:       xdr.Uint32(row.NumSponsoring),
 							SignerSponsoringIDs: signerSponsoringIDs,
-							Ext: xdr.AccountEntryExtensionV2Ext{
-								V: 3,
-								V3: &xdr.AccountEntryExtensionV3{
-									SeqLedger: xdr.Uint32(row.SequenceLedger),
-									SeqTime:   xdr.TimePoint(row.SequenceTime),
-								},
-							},
+							Ext:                 v3extension,
```
leighmcculloch, line 437: +1 We should maintain a NULL for any accounts without these values. We shouldn't translate it to 0 I think which would misrepresent the state of the ledger, even if that is how they get treated. Also agree, the backfill would be expensive, and also NULLable columns are cheap where-as setting lots of 0s will take up more space.

leighmcculloch (commented):
It looks like no tests are being updated as part of this change. Are these areas of code untested? Can you add tests that confirm these cases are handled?

The change looks good though.

Shaptic:
@leighmcculloch yeah, test cases are updated now :+1:. The bug itself was not directly testable because of the gap between how we describe columns in the Go code (e.g. a `uint32`) vs. how they're actually in the database (actually a `*uint32`). But the tests to consider account upgrades, now.

leighmcculloch (approved):

services/horizon/internal/ingest/verify.go
```diff
@@ -394,6 +394,18 @@ func addAccountsToStateVerifier(ctx context.Context, verifier *verify.StateVerif
...
+		// V3 extension, so we need to check whether or not this extension needs
+		// to be filled out.
+		v3extension := xdr.AccountEntryExtensionV2Ext{V: 0}
+		if row.SequenceLedger.Valid && row.SequenceTime.Valid {
+			v3extension.V = 3
+			v3extension.V3 = &xdr.AccountEntryExtensionV3{
+				SeqLedger: xdr.Uint32(row.SequenceLedger.Int64),
+				SeqTime:   xdr.TimePoint(row.SequenceTime.Int64),
+			}
+		}
```
bartekn, line 407: This should be done in `xdr.LedgerEntry.Normalize` method.

bartekn (approved):

bartekn:
Also, it seems that `TestProtocol19Upgrade/TestTransactionAccountV3Upgrade` test fails.

services/horizon/internal/ingest/verify.go
Shaptic: I see what you're hinting at, but I'm not sure what you mean. It states that the purpose of the function is "to compare two ledger entries that are the same but for one of them extensions are not set." That means we should create a zero-valued V3 extension there. However, if we're creating an `AccountEntry` in the purest sense of the word, i.e. one that accurately represents what's in the database (i.e. in `addAccountsToStateVerifier`), we should not create a V3 extension yet.

Of course, the difference between V1/V2 extensions and the V3 extension is that we backfilled the former, but we aren't backfilling this new one. In that sense, maybe we should *not* "normalize" the V3 extension, as a ledger entry that does have a V3 extension and one that doesn't are more "strictly" distinct..

Maybe I'm misunderstanding your comment? Hopefully my thoughts makes sense, though..

PR #4360, opened 2022-04-29
services/horizon: Improve transaction precondition `omitempty` behavior

Shaptic:
### What
Two new omissions:
 1. omit minimum account sequence age when it's `"0"`
 2. omit `max_ledger` when it's `0`

### Why
1. Because we use a string for the field (to work around languages that don't natively support uint64 values) in the JSON response, but the field is a non-pointer in the XDR, it will always be set and will be 0 by default. Since it's a duration, 0 is effectively "not set", and so it should be omitted from the resulting precondition.
2. Because `0` is "no upper bound," so it's effectively "not set."

### Known limitations
(1) feels a little hacky. Caveat: if someone explicitly sets it to 0 when building their tx, this is equivalent to not setting it, so they may be confused about its omission, but this is working as intended.

### DIFF
--- a/protocols/horizon/main.go
+++ b/protocols/horizon/main.go
@@ -552,7 +552,7 @@ type TransactionPreconditionsTimebounds struct {
 
 type TransactionPreconditionsLedgerbounds struct {
 	MinLedger uint32 `json:"min_ledger"`
-	MaxLedger uint32 `json:"max_ledger"`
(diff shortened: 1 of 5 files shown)


leighmcculloch (approved):
Looks good, but I don't see any test change or new test?

leighmcculloch (approved):

PR #4361, opened 2022-04-30
services/horizon: Use UNIX timestamps instead of RFC3339 strings for timebounds.

Shaptic:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
The timebounds w/in the new preconditions field are now UNIX timestamps.

### Why
We all agreed that time-related fields going forward should be UNIX timestamps in seconds.

This one slipped through the cracks, because it comes from the existing ValidBefore/After timebound fields, but though they are the same _time_, they should not be the same _value_.

### Known limitations
This will break SDKs that are parsing strings, but the community SDK blast specifies that it's a UNIX timestamp. Furthermore, Horizon isn't released yet, so nobody should even know that they're strings right now.

### DIFF
--- a/services/horizon/CHANGELOG.md
+++ b/services/horizon/CHANGELOG.md
@@ -11,6 +11,8 @@ This is the final release after the [release candidate](v2.17.0-release-candidat
   * `min_account_sequence_age` when it's `"0"`, as this is the default value when the condition is not set
   * `preconditions.ledgerbounds.max_ledger` when it's set to 0 (this means that there is no upper bound)
 
+- Timebounds within the `preconditions` object are a string representing UNIX timestamps in seconds rather than formatted date-times (which was a bug) ([4361](https://github.com/stellar/go/pull/4361)).
+
 ## V2.17.0 Release Candidate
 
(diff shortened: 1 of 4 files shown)


services/horizon/CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ This is the final release after the [release candidate](v2.17.0-release-candidat
   * `min_account_sequence_age` when it's `"0"`, as this is the default value when the condition is not set
   * `preconditions.ledgerbounds.max_ledger` when it's set to 0 (this means that there is no upper bound)
 
+- Timebounds within the `preconditions` object are a string representing UNIX timestamps in seconds rather than formatted date-times (which was a bug) ([4361](https://github.com/stellar/go/pull/4361)).
```
leighmcculloch, line 14:
```suggestion
- Timebounds within the `preconditions` object are a string containing int64 UNIX timestamps in seconds rather than formatted date-times (which was a bug) ([4361](https://github.com/stellar/go/pull/4361)).
```

bartekn (approved):

PR #4365, opened 2022-04-30
tools/horizon-cmp: Improve panic error message

Shaptic:
### What
This adds the requested URL to a panic message.

### Why
If `horizon-cmp` encounters an unexpected HTTP status code (like a 500), it will just spit out `panic: 500`. This isn't useful for debugging or reproducing the error, since you have no idea what the requested URL was.


sreuland (approved):
a reason for problem, super helpful!

bartekn (approved):

PR #4366, opened 2022-05-02
xdrgen: remove gemfile and rakefile to just use docker for the xdrgen

paulbellamy:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Remove `Gemfile*`, and `Rakefile`, and replace with a docker image in `Makefile`

### Why

Inspired by: https://github.com/stellar/rs-stellar-xdr/blob/main/Makefile#L11-L26

Reduces some code duplication, and "ruby-ness"

### Known limitations

Doesn't have a rake command to copy over xdr files from stellar-core anymore. Could add a make rule for that if we feel strongly about it.


Makefile
```diff
@@ -32,7 +32,24 @@ gxdr/xdr_generated.go: $(XDRS)
 	go run github.com/xdrpp/goxdr/cmd/goxdr -p gxdr -enum-comments -o $@ $(XDRS)
 	go fmt $@
 
-xdr/xdr_generated.go: $(XDRS) Rakefile Gemfile.lock
-	bundle exec rake xdr:generate
+define xdrheader
+//lint:file-ignore S1005 The issue should be fixed in xdrgen. Unfortunately, there's no way to ignore a single file in staticcheck.
+//lint:file-ignore U1000 fmtTest is not needed anywhere, should be removed in xdrgen.
```
leighmcculloch, line 37: 💡 I think we should add these to the frontmatter of the go generator in xdrgen. It'll make this simpler, and it'll make it easier to use xdrgen with other xdr. If you do this, it will simplify this Makefile target I think.

Makefile
```diff
@@ -32,7 +32,24 @@ gxdr/xdr_generated.go: $(XDRS)
...
+endef
+export xdrheader
+
+xdr/xdr_generated.go: $(XDRS)
+	docker run --platform linux/amd64 -it --rm -v $$PWD:/wd -w /wd ruby /bin/bash -c '\
```
leighmcculloch, line 42: You shouldn't need the platform specification here. I use the arm64 ruby image to do this and it works fine.
```suggestion
	docker run -it --rm -v $$PWD:/wd -w /wd ruby /bin/bash -c '\
```

Makefile
```diff
@@ -32,7 +32,24 @@ gxdr/xdr_generated.go: $(XDRS)
...
+		gem install specific_install && \
```
leighmcculloch, line 43: Could you lock this gem install to a specific version to ensure consistency.
```suggestion
		gem install specific_install -v 0.3.7 && \
```

Fun fact, this still doesn't guarantee a consistent install of dependencies, and there's no way with `gem` to make it consistent. (See https://twitter.com/___leigh___/status/1520997579772678144.) It doesn't matter for specific_install though because it has no dependencies.

Makefile
leighmcculloch: If you open a PR on xdrgen for this, assign me and I'll approve it.

Makefile
```diff
@@ -32,7 +32,24 @@ gxdr/xdr_generated.go: $(XDRS)
...
 
-xdr/xdr_generated.go: $(XDRS) Rakefile Gemfile.lock
-	bundle exec rake xdr:generate
+define xdrheader
+//lint:file-ignore S1005 The issue should be fixed in xdrgen. Unfortunately, there's no way to ignore a single file in staticcheck.
+//lint:file-ignore U1000 fmtTest is not needed anywhere, should be removed in xdrgen.
+endef
+export xdrheader
+
+xdr/xdr_generated.go: $(XDRS)
```
leighmcculloch, line 41: We need a target that downloads the XDRs. The Rakefile does that. This could be some simple curl commands.

Makefile
paulbellamy: Ah, hmm. I got a weird error locally.. let me try again.

Makefile
leighmcculloch: You could copy this: https://github.com/stellar/rs-stellar-xdr/pull/34

Makefile
paulbellamy: https://github.com/stellar/xdrgen/pull/82

leighmcculloch (approved):

Makefile
```diff
@@ -32,7 +32,23 @@ gxdr/xdr_generated.go: $(XDRS)
...
 
-xdr/xdr_generated.go: $(XDRS) Rakefile Gemfile.lock
-	bundle exec rake xdr:generate
+xdr/%.x:
+	curl -Lsf -o $@ https://raw.githubusercontent.com/stellar/stellar-core/master/src/$@
+
+xdr/xdr_generated.go: $(XDRS)
+	docker run -it --rm -v $$PWD:/wd -w /wd ruby /bin/bash -c '\
+		gem install specific_install -v 0.3.7 && \
+		gem specific_install https://github.com/stellar/xdrgen.git -b master && \
```
leighmcculloch, line 41: One more thing, we should probably lock this to a commit of xdrgen.

Makefile
```diff
@@ -32,7 +32,23 @@ gxdr/xdr_generated.go: $(XDRS)
 	go run github.com/xdrpp/goxdr/cmd/goxdr -p gxdr -enum-comments -o $@ $(XDRS)
 	go fmt $@
 
-xdr/xdr_generated.go: $(XDRS) Rakefile Gemfile.lock
-	bundle exec rake xdr:generate
+xdr/%.x:
```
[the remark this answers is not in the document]
paulbellamy: It is like a "resolver". So when `xdr/xdr_generated.go` requires them as dependencies, the files will be downloaded.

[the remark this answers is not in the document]
paulbellamy: this is what the `xdr-update` rule uses to update them. Delete them all, and build `xdr/xdr_generated.go`. That implicitly downloads them.

PR #4367, opened 2022-05-03
services/horizon/internal/integration: Clean up executing every integration test

2opremio:
We were not calling `Shutdown()` appropriately calling shutdown after executing integration tests.

I think this is (at least partly) the cause of recent integration test problems like https://github.com/stellar/go/runs/6233228705

### DIFF
--- a/services/horizon/internal/integration/claimable_balance_ops_test.go
+++ b/services/horizon/internal/integration/claimable_balance_ops_test.go
@@ -16,6 +16,7 @@ import (
 func TestClaimableBalanceCreationOperationsAndEffects(t *testing.T) {
 	tt := assert.New(t)
 	itest := integration.NewTest(t, integration.Config{})
+	defer itest.Shutdown()
 	master := itest.Master()
 
 	t.Run("Successful", func(t *testing.T) {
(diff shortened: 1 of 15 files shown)


2opremio:
In the end I am not sure this will help with the integration test runs, but ... it's a good practice:

@bartekn 's reasoning:

> I’m not sure it’s caused by a lack of Shutdown(). the error comes from postgres server and each test runs an isolated postgres in docker. so I think it must be something isolated to TestTxsub test. I noticed something else. once 20 test accounts are created the test starts sending requests below:
> ```
> time="2022-04-29T18:45:37.141Z" level=info msg="Finished request" app_name=undefined app_version=undefined bytes=322 client_name=go-stellar-sdk client_version=2.1.0 duration=0.005064157 host="localhost:8000" ip="[::1]" ip_port="[::1]:49474" method=GET path=/accounts/GBZXN7PIRZGNMHGA7MUUUF4GWPY5AYPV6LY4UV2GL6VJGIQRXFDNMADI/data/config.memo_required pid=5821 referer=undefined req=fv-az206-0/Br5gxfz9Ua-000960 route="/accounts/{account_id}/data/{key}" status=404 streaming=false x_forwarder_for=
> ```
> looks like it comes from horizonclient and is requested on each request to POST /transactions. each call to /accounts/{account_id}/data/{key} starts a DB transaction so requires a DB transaction lock.

As suggested by @bartekn the solution is probably to simply limit the number of connections to 100 during integration tests.

2opremio:
Closing since this PR is useless (we implicitly shut the test down by using `i.t.Cleanup(i.Shutdown)`)

PR #4369, opened 2022-05-04
exp/lighthorizon: Prototyping support for XDR serialization

Shaptic:
### What
Make checkpoint indices use XDR for serialization rather than a custom format.

Generate the XDR by going to `/exp/lighthorizon/index` then running `make`.

### Why
Standardization with the rest of our codebase, easier to grok binary format.

### Known limitations
n/a, tests pass:

```bash
$ go test -v ./index/...
ok  	github.com/stellar/go/exp/lighthorizon/index	2.409s
```


exp/lighthorizon/index/main.go
```diff
@@ -18,44 +18,16 @@ type CheckpointIndex struct {
...
 	}
 
 	return &CheckpointIndex{
-		bitmap:          bitmap,
-		firstCheckpoint: uint32(firstCheckpoint),
-		lastCheckpoint:  uint32(lastCheckpoint),
+		bitmap:          xdrCheckpoint.Bitmap,
+		firstCheckpoint: uint32(xdrCheckpoint.FirstCheckpoint),
+		lastCheckpoint:  uint32(xdrCheckpoint.LastCheckpoint),
 	}, nil
```
Shaptic, line 31: Better would of course to make xdr.CheckpointIndex part of `CheckpointIndex`, but since I'm just prototyping this is less changes until we land on the "right" pattern for this.

exp/lighthorizon/index/Gemfile
```diff
@@ -0,0 +1,7 @@
+source 'https://rubygems.org'
+
+gem 'xdrgen', git:'https://github.com/stellar/xdrgen.git', ref: 'master'
+gem 'pry'
+gem 'octokit'
+gem 'netrc'
+gem 'rake'
```
leighmcculloch, line 7: Recommend looking at master for the new way we're building xdr without Gemfile, etc. @paulbellamy just committed it recently.

exp/lighthorizon/index/Rakefile
```diff
@@ -0,0 +1,53 @@
...
+      output_dir: "xdr",
+      namespace:  "xdr",
+      language:   :go
+    )
+    compilation.compile
+
+    xdr_generated = IO.read("xdr/xdr_generated.go")
+    IO.write("xdr/xdr_generated.go", <<~EOS)
+      //lint:file-ignore S1005 The issue should be fixed in xdrgen. Unfortunately, there's no way to ignore a single file in staticcheck.
+      //lint:file-ignore U1000 fmtTest is not needed anywhere, should be removed in xdrgen.
```
leighmcculloch, line 47: You don't need these comments. Xdrgen adds them.

paulbellamy (approved):
This is obviously brilliant, but let's use the new xdrgen makefile setup from https://github.com/stellar/go/pull/4366, please 🙏

exp/lighthorizon/index/Rakefile
Shaptic: Yep, remnants

exp/lighthorizon/index/Gemfile
Shaptic: Great call! Didn't see that PR.

Shaptic:
good call @leighmcculloch, looks like the `lighthorizon` branch is pretty out of date

PR #4373, opened 2022-05-05
exp/lighthorizon: Modify trie tests to use `assert` over raw comparisons.

Shaptic:
### What
Use `testify/assert` library in `TrieIndex` tests. 

### Why 
This makes the test cases cleaner by having less reliance on custom conditionals & error messages. It's a preliminary step to moving the `TrieIndex` serialization to XDR so that it's clear that existing tests are unchanged (and pass).

Things still work:
```
exp/lighthorizon> go test -failfast ./index/...    
ok  	github.com/stellar/go/exp/lighthorizon/index	13.964s
```

### DIFF
--- a/exp/lighthorizon/index/trie_test.go
+++ b/exp/lighthorizon/index/trie_test.go
@@ -5,8 +5,8 @@ import (
 	"encoding/binary"
 	"encoding/hex"
 	"encoding/json"
-	"fmt"
 	"math/rand"
+	"strconv"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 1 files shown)


exp/lighthorizon/index/trie_test.go
```diff
@@ -101,36 +99,30 @@ func TestTrieIndexSuffixes(t *testing.T) {
...
-			value, ok := read.Get([]byte(key))
-			if !ok {
-				t.Errorf("Key not found: %s", key)
-			} else {
-				ledger := binary.BigEndian.Uint32(value)
-				if ledger != expected {
-					t.Errorf("Key %s found: %v, expected: %v", key, ledger, expected)
+			for key, expected := range inserts {
+				value, ok := read.Get([]byte(key))
+				if !assert.Truef(t, ok, "Key not found: %s", key) {
```
paulbellamy, line 118:
```suggestion
				require.Truef(t, ok, "Key not found: %s", key)
```

paulbellamy: `require` is generally closer to what you mean than `assert`

paulbellamy (approved):

exp/lighthorizon/index/trie_test.go
Shaptic: Really good call! Honestly was too lazy to look into the difference and now thinking we should use `require` everywhere we currently use `assert` :facepalm:

PR #4379, opened 2022-05-11
horizon: build docker image with --no-cache flag

jacekn:
### What

Add `--no-cache` flag to horizon docker build.

### Why

We provide stellar-horizon package version as parameter but we don't do that for core.
This means we may end up using cached layer for package installs and this may result
in older core being installed

### Testing

The change was tested locally using:
```
VERSION=2.17.0-267 make docker-build
```


leighmcculloch:
I think this change will unfairly punish folks using the image. I would rather see the stellar-core version become a parameter like the horizon version. Doing so would mean the image is more deterministic.

For the specific 19.0.1 case I think you should add in a command to clear the cache in the Jenkins pipeline file. It won't require tagging a new Horizon to do it. Or, @stellar/ops-team could ssh into the Jenkins workers and simply clear the docker cache.

Shaptic:
That may probably a better holistic approach, but that's a huge break in how this image worked prior. Every [tag](https://hub.docker.com/r/stellar/stellar-horizon/tags) is tied to a specific build of Horizon and the "latest" version of Core at the time it's built. It's basically a snapshot of the world at the time of release. I think it's okay to update that snapshot with a new picture of the world since the old picture was found to be broken. Having yet another place to forget to bump versions is frustrating.

Anyway, I'm not entirely opposed, but if we opt for that moving forward, we'd need to tag images with both Horizon and Core versions, so instead of [2.17.0-267](https://hub.docker.com/layers/stellar-horizon/stellar/stellar-horizon/2.17.0-267/images/sha256-9f1e8667db5ee539efb7a76e02a5b9633ba5ea337a71558a27450984098b1081?context=explore) it would be something like `2.17.0-267~19.0.1-[xxx]`.

Shaptic:
This goes back to "What is the intent of the stellar/stellar-horizon image?"

To me, the intent seems pretty clear: "This is stable Horizon `vX.Y.Z` packaged with a working stable Core." If that means replacing the image because that Core version _wasn't_ working, that's completely fine. There's no tangible benefit to having broken images in the registry, and I can easily imagine us dealing with people downloading broken images and it being hard to track down why their Horizon isn't working.

If the intent is different, i.e. "This is Horizon vX.Y.Z packaged with Core vA.B.C," then fine, but that seems... less useful.

jacekn:
To make docker builds fast by default but uncached when needed we could add optional environment variable with support in the Makefile.
Maybe something like this:
```
DOCKER_OPTS="--no-cache" make docker-build
```

I think that addresses @leighmcculloch concern about impacting wider audience while also giving us easy way to have fresh, uncached builds

leighmcculloch:
@jacekn Yup. Alternatively, clear the layer cache on the worker before building, which requires no changes here.

satyamz (approved):
LGTM!

2opremio:
Regarding tags, @Shaptic you have a point, but in practice that means that if there is a problem with core (like there has been now) users won't really know whether they are using the right version or not.

Sure, users can always pull, but that brings confusion and problems (due to the mutable tags).

If we don't want to add the full core version, we can add a docker image version which we append whenever there is an update to the same Horizon version.

jacekn:
> @jacekn Yup. Alternatively, clear the layer cache on the worker before building, which requires no changes here.

I pushed new commit to add optional docker flags. It's very low overhead and default behaviour will remain the same. Also maybe in the future we'll use it for some other purpose so I think there is no harm in having it here.

@Shaptic @leighmcculloch @2opremio your points about rebuilding on core updates and immutable tags are very good for sure. I think it's worth discussing further but maybe separately from this PR? WDYT?

Shaptic:
Agreed @jacekn - I think having add'l docker args is useful regardless of the conclusion we come to there.

Do we want related PRs that will expose this in Jenkins, etc. before merging this?

Shaptic:
Agreed @jacekn - I think having add'l docker args is useful regardless of the conclusion we come to there.

Do we want related PRs that will expose this in Jenkins, etc. before merging this?

Shaptic (approved):

PR #4385, opened 2022-05-19
services/horizon: Optimize claimable balances query to limit records earlier

Shaptic:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Move from `WITH (...) SELECT ... LIMIT x` to `WITH (... LIMIT x) SELECT ...`.

### Why
Limiting records sooner -> faster queries.

### Known limitations
We should probably do this "properly" and avoid the redundant `WITH` statement altogether, but this is the smallest possible delta that results in a much faster query.


Shaptic:
@sydneynotthecity my perspective was basically "smallest change for biggest gain to minimize regression risk", so even though getting rid of the CTE would satisfy my desire for simpler code, it's possible it might have a different effect than we expect on different CB queries? since we've been scoped on a specific kind. I don't know enough to argue either way.

sydneynotthecity:
@Shaptic Definitely agree with your approach and think it's the right change for the circumstances.

bartekn (approved):
LGTM, the reason for subquery is probably explained in the comment above the changes:
```go
	// we need to use WITH syntax to force the query planner to use the right
	// indexes, otherwise when the limit is small, it will use an index scan
	// which will be very slow once we have millions of records
```
We probably need to check carefully if removing the subquery doesn't make the planner use wrong indexes, etc.

sydneynotthecity:
> 
> We probably need to check carefully if removing the subquery doesn't make the planner use wrong indexes, etc.
>

I looked into this a little bit in the [doc](https://docs.google.com/document/d/12lu10tspqk6aIyv_Zw57qoLOS-1jRaqjdcM8PfDRElk/edit?usp=sharing) and moving the `LIMIT` clause is what actually impacts which index is used during query planning. Regardless of whether we use a CTE, moving the `LIMIT` up causes the query planner to use `claimable_balances_by_last_modified_ledger_and_id` instead of `claimable_balances_by_asset`. 

My opinion is that we should trust the query planner to use whatever index will return the results the fastest given moving this clause reduces the return time from 9639.470 ms to 3.845 ms.

@2opremio might be worth running some non AQUA claimable balance requests during the alb-replay to ensure we're not impacting non long-running traffic.

sreuland (approved):

Shaptic:
Repointed target branch to [`horizon-db-optimizations`](https://github.com/stellar/go/tree/horizon-db-optimizations), will merge once tests pass :+1:

PR #4390, opened 2022-05-20
support/db, services/horizon/internal: Configure postgres client connection timeouts for read only db

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Postgres provides several configuration options for clients:

https://www.postgresql.org/docs/9.6/runtime-config-client.html

In this commit I have configured the statement_timeout and idle_in_transaction_session_timeout for the horizon read only connections. We should have not queries or read only transactions running longer than the http request timeout.

### Why

Although we do have context cancellation which should cancel the postgres query if the http request is cancelled, I thought it would still be good to configure a statement timeout. The context cancellation is implemented by having the client library send a request to postgres to cancel the query. However, if postgres is at maximum CPU utilization it might not be available to receive the cancellation request. We should not have that problem using a statement timeout because the client connection is initialized with that timeout setting and postgres will abort the query automatically after the elapsed timeout period.  

### Known limitations

Technically, we don't need to implement code changes to configure statement_timeout and idle_in_transaction_session_timeout because we can configure those parameters directly in the db_url configured in puppet.

However, I thought it would be a good idea to implement it in code so we could have the values configured by default. If you prefer this change to be implemented in puppet I can close this PR.

### DIFF
--- a/services/horizon/internal/init.go
+++ b/services/horizon/internal/init.go
@@ -18,9 +18,9 @@ import (
 	"github.com/stellar/go/support/log"
 )
 
-func mustNewDBSession(subservice db.Subservice, databaseURL string, maxIdle, maxOpen int, registry *prometheus.Registry) db.SessionInterface {
+func mustNewDBSession(subservice db.Subservice, databaseURL string, maxIdle, maxOpen int, registry *prometheus.Registry, clientConfigs ...db.ClientConfig) db.SessionInterface {
 	log.Infof("Establishing database session for %v", subservice)
-	session, err := db.Open("postgres", databaseURL)
+	session, err := db.Open("postgres", databaseURL, clientConfigs...)
 	if err != nil {
 		log.Fatalf("cannot open %v DB: %v", subservice, err)
 	}
@@ -47,21 +47,36 @@ func mustInitHorizonDB(app *App) {
 	}
 
 	if app.config.RoDatabaseURL == "" {
+		var clientConfigs []db.ClientConfig
+		if !app.config.Ingest {
+			// if we are not ingesting then we don't expect to have long db queries / transactions
+			clientConfigs = append(
+				clientConfigs,
+				db.StatementTimeout(app.config.ConnectionTimeout),
+				db.IdleTransactionTimeout(app.config.ConnectionTimeout),
+			)
+		}
 		app.historyQ = &history.Q{mustNewDBSession(
 			db.HistorySubservice,
 			app.config.DatabaseURL,
 			maxIdle,
 			maxOpen,
 			app.prometheusRegistry,
+			clientConfigs...,
 		)}
 	} else {
 		// If RO set, use it for all DB queries
+		roClientConfigs := []db.ClientConfig{
+			db.StatementTimeout(app.config.ConnectionTimeout),
+			db.IdleTransactionTimeout(app.config.ConnectionTimeout),
+		}
 		app.historyQ = &history.Q{mustNewDBSession(
 			db.HistorySubservice,
 			app.config.RoDatabaseURL,
 			maxIdle,
 			maxOpen,
 			app.prometheusRegistry,
+			roClientConfigs...,
 		)}
 
 		app.primaryHistoryQ = &history.Q{mustNewDBSession(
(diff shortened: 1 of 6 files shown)


support/db/main.go
```diff
@@ -163,8 +169,46 @@ func pingDB(db *sqlx.DB) error {
...
+func IdleTransactionTimeout(timeout time.Duration) ClientConfig {
+	return ClientConfig{
+		Key:   "idle_in_transaction_session_timeout",
+		Value: strconv.FormatInt(timeout.Milliseconds(), 10),
+	}
+}
+
+func augmentDSN(dsn string, clientConfigs []ClientConfig) string {
+	parsed, err := url.Parse(dsn)
+	if err != nil || parsed.Scheme == "" {
```
2opremio, line 193: Is parsing expected to fail in some cases? when? shouldn't we just fail if that's the case?

(I think a commend is in order)

sreuland: good point, returning error here would maintain most closely the current/expected behavior, rather than new outcome of attempting fix, less code to introduce here too?

services/horizon/internal/init.go
```diff
@@ -47,21 +47,36 @@ func mustInitHorizonDB(app *App) {
 	}
 
 	if app.config.RoDatabaseURL == "" {
+		var clientConfigs []db.ClientConfig
+		if !app.config.Ingest {
+			// if we are not ingesting then we don't expect to have long db queries / transactions
+			clientConfigs = append(
+				clientConfigs,
+				db.StatementTimeout(app.config.ConnectionTimeout),
```
sreuland, line 55: nice addition, should be interesting to see how often this StatementTimeout kicks in!

support/db/main_test.go
```diff
@@ -27,3 +28,26 @@ func TestGetTable(t *testing.T) {
...
+		StatementTimeout(4 * time.Millisecond),
+	}
+	for _, testCase := range []struct {
+		input    string
+		expected string
+	}{
+		{"postgresql://localhost", "postgresql://localhost?idle_in_transaction_session_timeout=2000&statement_timeout=4"},
+		{"postgresql://localhost", "postgresql://localhost?idle_in_transaction_session_timeout=2000&statement_timeout=4"},
+		{"postgresql://localhost", "postgresql://localhost?idle_in_transaction_session_timeout=2000&statement_timeout=4"},
+		{"user=bob password=secret", "user=bob password=secret idle_in_transaction_session_timeout=2000 statement_timeout=4"},
```
sreuland, line 44: minor for consideration, in case where external environments try to set something, maybe a test to confirm overwrite:
```
{"user=bob password=secret statement_timeout=1", "user=bob password=secret idle_in_transaction_session_timeout=2000 statement_timeout=4"}
```

support/db/session_test.go
```diff
@@ -129,3 +129,34 @@ func TestSession(t *testing.T) {
...
+	sess, err := Open(db.Dialect, db.DSN, IdleTransactionTimeout(50*time.Millisecond))
+	assert.NoError(err)
+	defer sess.Close()
+
+	assert.NoError(sess.Begin())
+	<-time.After(100 * time.Millisecond)
+
+	var count int
+	err = sess.GetRaw(context.Background(), &count, "SELECT COUNT(*) FROM people")
+	assert.ErrorIs(err, ErrBadConnection)
```
sreuland, line 161: interesting, a force closed tx, invalidates the whole connection, really good test to show that effect!

sreuland (approved):
nice work, I think the approach done here to embed the settings into horizon runtime is worthwhile, because deployments benefit automatically rather than having to manually config environment to get this protection, which doesn't restrict timing any further, since follows existing connection timeout threshold.

support/db/main_test.go
tamirms: I added a test case for this but I decided that we shouldn't override `statement_timeout` and `idle_in_transaction_session_timeout` if the connection string already configures those parameters because it would probably be surprising for the operator

support/db/main.go
tamirms: I added some comments. Basically, I check that we're given a url connection string (e.g. `postgres://localhost`) and if that's not the case then I assume it must be a space separated connection string. It's possible that the connection string is completely invalid but I don't attempt to validate it here and instead I will let it fail when sql.Open() is called

support/db/main_test.go
sreuland: yes, agreed, that would be expected convention, thx!

PR #4397, opened 2022-05-23
Implement StreamAllOffers with batches

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

The query to get all rows in the offers table currently takes more than 10 seconds to execute.

### Why

Given that the horizon request timeout is 10 seconds, we need to reimplement StreamAllOffers to use batching.

### Known limitations

[N/A]

### DIFF
--- a/services/horizon/internal/db2/history/offers.go
+++ b/services/horizon/internal/db2/history/offers.go
@@ -2,12 +2,15 @@ package history
 
 import (
 	"context"
+	"math"
 
 	sq "github.com/Masterminds/squirrel"
 	"github.com/jmoiron/sqlx"
 
 	"github.com/stellar/go/support/errors"
 )
+
+const offersBatchSize = 50000
 
 // QOffers defines offer related queries.
 type QOffers interface {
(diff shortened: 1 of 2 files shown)


services/horizon/internal/db2/history/offers.go
```diff
@@ -83,28 +86,45 @@ func (q *Q) GetOffers(ctx context.Context, query OffersQuery) ([]Offer, error) {
 
 // StreamAllOffers loads all non deleted offers
 func (q *Q) StreamAllOffers(ctx context.Context, callback func(Offer) error) error {
+	lastID := int64(math.MinInt64)
```
bartekn, line 89: Zero seems to be good given offer IDs start at 1?
```suggestion
	lastID := int64(0)
```

services/horizon/internal/db2/history/offers.go
```diff
@@ -83,28 +86,45 @@ func (q *Q) GetOffers(ctx context.Context, query OffersQuery) ([]Offer, error) {
...
+func (q *Q) streamAllOffersBatch(ctx context.Context, lastId int64, limit uint64, callback func(Offer) error) (int64, error) {
 	var rows *sqlx.Rows
 	var err error
 
-	if rows, err = q.Query(ctx, selectOffers.Where("deleted = ?", false)); err != nil {
-		return errors.Wrap(err, "could not run all offers select query")
+	rows, err = q.Query(ctx, selectOffers.
+		Where("deleted = ?", false).
+		Where("offer_id > ? ", lastId).
+		OrderBy("offer_id asc").Limit(limit))
```
bartekn, line 109: This may not work properly outside `REPEATABLE READ` transaction. We probably should ensure if this is called inside a DB tx.

services/horizon/internal/db2/history/offers.go
```diff
@@ -2,13 +2,16 @@ package history
...
 	"context"
+	"math"
 
 	sq "github.com/Masterminds/squirrel"
 	"github.com/jmoiron/sqlx"
 
 	"github.com/stellar/go/support/errors"
 )
 
+const offersBatchSize = 50000
```
sreuland, line 13: this is somewhat dependent on the compute resources of the db right, do you think it's worthwhile to be more conservative with a smaller batch size to keep the query under 10s, or does it them impact throughput here more-so?

sreuland (approved):
got it, more executions of same sql, but filtered to a smaller result set, should keep it under 10s each, not completely deterministic right, could be somewhat vulnerable to low compute resources on db, but if following suggested hw configs, should be more than sufficient.

PR #4398, opened 2022-05-24
/db/history: update the claimable balance SQL query

sreuland:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Change the claimable balance SQL Query generation to position LIMIT clause dependent on presence of cursor parameters as that determines how many effective rows and the indexes are chosen by db query planner.

### Why

the query planner isn't picking the right indexes to use on claimable balance query due to LIMIT clause

### DIFF
--- a/services/horizon/internal/db2/history/claimable_balances.go
+++ b/services/horizon/internal/db2/history/claimable_balances.go
@@ -56,32 +56,37 @@ func (cbq ClaimableBalancesQuery) Cursor() (int64, string, error) {
 // ApplyCursor applies cursor to the given sql. For performance reason the limit
 // is not applied here. This allows us to hint the planner later to use the right
 // indexes.
-func (cbq ClaimableBalancesQuery) ApplyCursor(sql sq.SelectBuilder) (sq.SelectBuilder, error) {
+func (cbq ClaimableBalancesQuery) ApplyCursor(sql sq.SelectBuilder) (sq.SelectBuilder, bool, error) {
+	var hasPagedLimit bool
 	p := cbq.PageQuery
 	l, r, err := cbq.Cursor()
 	if err != nil {
-		return sql, err
+		return sql, hasPagedLimit, err
+	}
+	if l > 0 && r != "" {
+		hasPagedLimit = true
 	}
 
 	switch p.Order {
 	case db2.OrderAscending:
-		if l > 0 && r != "" {
+		if hasPagedLimit {
 			sql = sql.
 				Where(sq.Expr("(cb.last_modified_ledger, cb.id) > (?, ?)", l, r))
+
 		}
 		sql = sql.OrderBy("cb.last_modified_ledger asc, cb.id asc")
 	case db2.OrderDescending:
-		if l > 0 && r != "" {
+		if hasPagedLimit {
 			sql = sql.
 				Where(sq.Expr("(cb.last_modified_ledger, cb.id) < (?, ?)", l, r))
 		}
 
 		sql = sql.OrderBy("cb.last_modified_ledger desc, cb.id desc")
 	default:
-		return sql, errors.Errorf("invalid order: %s", p.Order)
+		return sql, hasPagedLimit, errors.Errorf("invalid order: %s", p.Order)
 	}
 
-	return sql, nil
+	return sql, hasPagedLimit, nil
 }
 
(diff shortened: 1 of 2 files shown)


sydneynotthecity (commented):
After performing several more benchmarking experiments, I think we actually need to place the limit clause based on whether or not the `claimants` filter is present. 

Regardless of pagination, if the user is filtering on `claimants`, the limit clause should be outside the subselect query:
```
") select " + claimableBalancesSelectStatement + " from cb LIMIT ?"
```

If the `claimants` filter is **not** present, the `LIMIT` clause should always reside in the subselect:
```
"LIMIT ?) select " + claimableBalancesSelectStatement + " from cb"
```

I tested with the following combinations:

* filtering only `asset`
* filtering `asset` with pagination
* filtering only `claimants`
* filtering `claimants` with pagination 
* filtering  only `sponsor`
* only pagination
* filtering `asset` **and** `claimants` with pagination

the `assets` filter sees great performance improvement by always having the `LIMIT` clause within the subselect. Other combinations only had negligible improvement. Conversely, the `claimants` filter requires the `LIMIT` clause to always exist outside the subselect for optimal performance.

sreuland:
@2opremio , I can adjust this to assemble the sql query based on @sydneynotthecity 's profiling, should I proceed on that or hold off?

2opremio:
Oh crap, I deleted the branch by mistake. Let me recreate it.

2opremio:
Done! No, we shouldn't hold off!

sreuland:
@sydneynotthecity @2opremio , I reworked that claimant balance query for LIMIT placement per the profiling results and recommendations.

services/horizon/internal/db2/history/claimable_balances.go
```diff
@@ -197,11 +202,17 @@ func (q *Q) FindClaimableBalanceByID(ctx context.Context, balanceID string) (Cla
 // GetClaimableBalances finds all claimable balances where accountID is one of the claimants
 func (q *Q) GetClaimableBalances(ctx context.Context, query ClaimableBalancesQuery) ([]ClaimableBalance, error) {
 	sql, err := query.ApplyCursor(selectClaimableBalances)
+	// we need to use WITH syntax and correct LIMIT placement to force the query planner to use the right
+	// indexes, otherwise when the limit is small, it will use an index scan
+	// which will be very slow once we have millions of records
+	limitClausePlacement := "LIMIT ?) select " + claimableBalancesSelectStatement + " from cb"
```
sydneynotthecity, line 208: Do we need to worry about adding a space before `LIMIT`? Or does the original statement insert the necessary space? Like `" LIMIT ?) select "` instead

sydneynotthecity (approved):
One question regarding whether we need to add a space to the SQL syntax, otherwise looks good!

services/horizon/internal/db2/history/claimable_balances.go
sreuland: Hello @sydneynotthecity , yes, looks like the squirrel `Suffix` takes care of spacing on concatenation, i dumped the final emitted sql under both conditions and looks ok with spacing:

```
# limit on the outer
WITH cb AS ( SELECT cb.id, cb.claimants, cb.asset, cb.amount, cb.sponsor, cb.last_modified_ledger, cb.flags FROM claimable_balances cb WHERE (cb.last_modified_ledger, cb.id) > (?, ?) AND cb.claimants @> '[{"destination": "GC3C4AKRBQLHOJ45U4XG35ESVWRDECWO5XLDGYADO6DPR3L7KIDVUMML"}]' ORDER BY cb.last_modified_ledger asc, cb.id asc ) select cb.id, cb.claimants, cb.asset, cb.amount, cb.sponsor, cb.last_modified_ledger, cb.flags from cb LIMIT ? 

# limit on the inner
WITH cb AS ( SELECT cb.id, cb.claimants, cb.asset, cb.amount, cb.sponsor, cb.last_modified_ledger, cb.flags FROM claimable_balances cb ORDER BY cb.last_modified_ledger asc, cb.id asc LIMIT ?) select cb.id, cb.claimants, cb.asset, cb.amount, cb.sponsor, cb.last_modified_ledger, cb.flags from cb 
```

thanks for double checking on that!

PR #4401, opened 2022-05-24
Run tests against Postgres 14

2opremio:
To see if there are any incompatibilities before trying to see the potential performance improvements.


2opremio:
Interestingly, integration tests work just fine, but unit tests get stuck.

PR #4405, opened 2022-05-24
exp/services/ledgerexporter: Extend tool to support lower ledger bound.

Shaptic:
Passing `-start-ledger <value>` makes it easier to export small ledger ranges for local testing, since a full export of any real network can take hours or even days.


exp/services/ledgerexporter/main.go
```diff
@@ -63,27 +59,39 @@ func main() {
...
-	if err := core.PrepareRange(context.Background(), ledgerbackend.UnboundedRange(latestLedger)); err != nil {
-		logger.WithError(err).Fatalf("could not prepare unbounded range %v", nextLedger)
+	var ledgerRange ledgerbackend.Range
+	if *startingLedger == 0 {
+		ledgerRange = ledgerbackend.UnboundedRange(latestLedger)
+	} else if *startingLedger != 0 && latestLedger == 2 {
+		ledgerRange = ledgerbackend.UnboundedRange(uint32(*startingLedger))
+		latestLedger = uint32(*startingLedger)
+	} else {
+		ledgerRange = ledgerbackend.BoundedRange(uint32(*startingLedger), latestLedger)
```
tamirms, line 74: seems like there is an implicit assumption that starting ledger is less than the latest ledger. should we validate that explicitly?

Shaptic: Good point, cleaned up the checks in 36fb0970 and made this explicit!

tamirms (approved):

PR #4409, opened 2022-05-25
protocols/horizon: Change Transaction.AccountSequence to int64

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Change `protocols/horizon.Transaction.AccountSequence` to `int64` from `string`.

### Why

There shouldn't be any extra parsing required in Go since `AccountSequence` returned by Horizon will be always a valid int64 value. We can make sure the value is rendered as a string to support JS using `string` tag.

### Known limitations

[TODO or N/A]


services/horizon/internal/resourceadapter/operations_test.go
```diff
@@ -14,11 +14,12 @@ import (
 )
 
 func TestNewOperationAllTypesCovered(t *testing.T) {
+	tx := &history.Transaction{TransactionWithoutLedger: history.TransactionWithoutLedger{AccountSequence: "1"}}
```
sreuland, line 17: would it help for consistency of AccountSequence type to also have 'TransactionWithoutLedget.AccountSequence' type int64 also?

sreuland (approved):
LGTM, thanks!

protocols/horizon/main.go
```diff
@@ -511,7 +511,7 @@ type Transaction struct {
 	Account           string    `json:"source_account"`
 	AccountMuxed      string    `json:"account_muxed,omitempty"`
 	AccountMuxedID    uint64    `json:"account_muxed_id,omitempty,string"`
-	AccountSequence   string    `json:"source_account_sequence"`
+	AccountSequence   int64     `json:"source_account_sequence,string"`
```
Shaptic, line 514: Is this a breaking change to the Go SDK?

Also, we have some other fields like this (that are `string`s but could be `int64`s). We should probably do a pass to find them all (this came up briefly during P19).

bartekn: > Is this a breaking change to the Go SDK?

Yes, this is a breaking change. However this will probably allow Go SDK users to remove some conversion code.

> Also, we have some other fields like this (that are strings but could be int64s). We should probably do a pass to find them all (this came up briefly during P19).

Sounds good. I changed this one because it simplifies the code in Starbridge. I did a quick pass and couldn't find anything else but if there are other instances of this we should definitely change them to ints.

Issue #4403, opened 2022-05-24
lighthorizon: Enable parallel creation of indices via map-reduce


Shaptic:
See parent issue for context: https://github.com/stellar/go/issues/4317

MapReduce (`/index/cmd/batch`): Indices are massive and take a long time to build, so we need a clean way to be able to build these using MapReduce, e.g. via AWS Batch jobs. The work has been started but has not been updated with the latest implementation from the single-process version. Ideally, these should use the same "core" code for index creation and just add abstractions for map/reduce parallelism vs. direct execution.

Shaptic:
Closed by #4423 and #4424 (with #4422 as a supporting PR).

PR #4410, opened 2022-05-25
exp/lighthorizon: Refactor single-process index builder.

Shaptic:
### What
This refactors the single-process index builder: (see 90d8ffa1c5512e467e5368acdf71485b878702c0):
 - allow worker count to be a command line parameter
 - split work by checkpoints rather than ledgers
 - move actual index insertion work to helpers
 - move progress bar into helpers
 - simplify participants code, payments vs. all
 - move index building into its own object (`IndexBuilder`)

This also chunks out work to the workers using ledger ranges that correspond to checkpoints.

### Why
Code cleanup and optimization: this minimizes flushes to the filesystem/network by isolating it to checkpoint ranges, which makes flushes more meaningful and is a better discretization.

The refactor into `IndexBuilder` is prep work for #4403, where an important part of that is using a unified codebase for both builders.

### Known limitations
tests are lacking but this is still in poc stages

### DIFF
--- a/exp/lighthorizon/index/builder.go
+++ b/exp/lighthorizon/index/builder.go
@@ -1,0 +1,294 @@
+package index
+
+import (
+	"context"
+	"fmt"
+	"io"
+
+	"github.com/stellar/go/historyarchive"
+	"github.com/stellar/go/ingest"
+	"github.com/stellar/go/ingest/ledgerbackend"
+	"github.com/stellar/go/support/log"
+	"github.com/stellar/go/toid"
+	"github.com/stellar/go/xdr"
+)
+
+// Module is a way to process data and store it into an index.
+type Module func(
+	idx Store,
+	ledger xdr.LedgerCloseMeta,
+	checkpoint uint32,
+	transaction ingest.LedgerTransaction,
+) error
+
+// IndexBuilder contains everything needed to build indices from ledger ranges.
+type IndexBuilder struct {
+	store             Store
+	history           ledgerbackend.HistoryArchiveBackend
+	networkPassphrase string
+
+	modules []Module
+}
+
+func NewIndexBuilder(
+	indexStore Store,
+	backend ledgerbackend.HistoryArchiveBackend,
+	networkPassphrase string,
+) *IndexBuilder {
+	return &IndexBuilder{
+		store:             indexStore,
+		history:           backend,
+		networkPassphrase: networkPassphrase,
+	}
+}
+
+// RegisterModule adds a module to process every given ledger. It is not
+// threadsafe and all calls should be made *before* any calls to `Build`.
+func (builder *IndexBuilder) RegisterModule(module Module) {
+	builder.modules = append(builder.modules, module)
+}
+
+// RunModules executes all of the registered modules on the given ledger.
+func (builder *IndexBuilder) RunModules(
+	ledger xdr.LedgerCloseMeta,
+	checkpoint uint32,
+	tx ingest.LedgerTransaction,
+) error {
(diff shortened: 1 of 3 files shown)


exp/lighthorizon/index/builder.go
```diff
@@ -0,0 +1,294 @@
...
+}
+
+// RegisterModule adds a module to process every given ledger. It is not
+// threadsafe and all calls should be made *before* any calls to `Build`.
+func (builder *IndexBuilder) RegisterModule(module Module) {
+	builder.modules = append(builder.modules, module)
+}
+
+// RunModules executes all of the registered modules on the given ledger.
+func (builder *IndexBuilder) RunModules(
```
paulbellamy, line 52: If this were `func (builder *IndexBuilder) RunModules(idx Store, ledger xdr.LedgerCloseMeta, checkpoint uint32, transaction ingest.LedgerTransaction) error`, it would itself be a module. 🤔 Not sure if that helps at all.

exp/lighthorizon/index/builder.go
```diff
@@ -0,0 +1,294 @@
...
+
+func (builder *IndexBuilder) Build(ctx context.Context, ledgerRange historyarchive.Range) error {
+	for ledgerSeq := ledgerRange.Low; ledgerSeq <= ledgerRange.High; ledgerSeq++ {
+		ledger, err := builder.history.GetLedger(ctx, ledgerSeq)
+		if err != nil {
+			log.WithField("error", err).Errorf("error getting ledger %d", ledgerSeq)
+			return err
+		}
+
+		checkpoint := ledgerSeq / 64
```
paulbellamy, line 74: do we need a +1 here? as ledger 2/64 would be checkpoint 0, no?

exp/lighthorizon/index/cmd/single/main.go
```diff
@@ -187,142 +142,62 @@ func main() {
...
+	log.Infof("Processed %d ledgers via %d workers", processed, parallel)
+	log.Infof("Uploading indices to %s", *targetUrl)
 	if err := indexStore.Flush(); err != nil {
 		panic(err)
 	}
 }
 
-func participantsForOperations(transaction ingest.LedgerTransaction, onlyPayments bool) ([]string, error) {
-	var participants []string
+func postProgress(prefix string, done, total uint64, startTime time.Time) {
```
paulbellamy, line 161:
```suggestion
func printProgress(prefix string, done, total uint64, startTime time.Time) {
```
`printProgress` maybe? `post` says, "http request" to me.

paulbellamy (approved):
A question, but lgtm.

exp/lighthorizon/index/builder.go
Shaptic: heh that's kinda cute, but should be easy to refactor if it becomes useful

exp/lighthorizon/index/builder.go
Shaptic: functionality-wise, I tried to change as little as possible, so this comes from the original code :thinking:

exp/lighthorizon/index/cmd/single/main.go
Shaptic: yeah, good call

exp/lighthorizon/index/builder.go
Shaptic: You're right though! Doing this means an account that's active in the first checkpoint is indistinguishable from an account that's not active at all.

Shaptic: Fixed in cde0ba1f.

PR #4411, opened 2022-05-25
horizon: Highlight that Ingestion Filtering is experimental

2opremio:


jacekn (approved):
Thank you for clarifying this!

PR #4422, opened 2022-06-03
exp/lighthorizon: Add thread-safe support for reading account list from `FileBackend`.

Shaptic:
### What
Three key changes:
* actually read the account list when using a filesystem backend
* using `O_APPEND` on the file to support concurrent writes
* ensure that the _read_ list is a unique set of accounts

### Why
Concurrent writes make `FileBackend` a viable option for MapReduce-based index building.

### Known Limitations
The written list of accounts may contain duplicates due to concurrent writes.

### DIFF
--- a/exp/lighthorizon/index/file.go
+++ b/exp/lighthorizon/index/file.go
@@ -8,10 +8,8 @@ import (
 	"os"
 	"path/filepath"
 
-	xdr3 "github.com/stellar/go-xdr/xdr3"
 	"github.com/stellar/go/support/errors"
 	"github.com/stellar/go/support/log"
-	"github.com/stellar/go/xdr"
 )
 
 type FileBackend struct {
@@ -52,21 +50,11 @@ func (s *FileBackend) FlushAccounts(accounts []string) error {
 
 	defer f.Close()
 
+	// We write one account at a time because writes that occur within a single
+	// `write()` syscall are thread-safe. A larger write might be split into
+	// many calls and thus get interleaved, so we play it safe.
 	for _, account := range accounts {
-		muxed := xdr.MuxedAccount{}
-		if err := muxed.SetAddress(account); err != nil {
-			return errors.Wrapf(err, "failed to encode %s", account)
-		}
-
-		raw, err := muxed.MarshalBinary()
-		if err != nil {
-			return errors.Wrapf(err, "failed to marshal %s", account)
-		}
-
-		_, err = f.Write(raw)
-		if err != nil {
-			return errors.Wrapf(err, "failed to write to %s", path)
-		}
+		f.Write([]byte(account + "\n"))
 	}
 
 	return nil
@@ -156,16 +144,6 @@ func (s *FileBackend) ReadAccounts() ([]string, error) {
 	path := filepath.Join(s.dir, "accounts")
 	log.Debugf("Opening accounts list at %s", path)
 
-	// We read the file in chunks with guarantees that we will always read on an
-	// account boundary:
-	//
-	//   Accounts w/o IDs are always 36 bytes (4-byte type, 32-byte pubkey)
-	//   Muxed accounts with IDs are always 44 bytes (36 + 8-byte ID)
-	//
-	// Thus, if we read 36*44=1584 bytes at a time, we are guaranteed to have a
-	// complete set of accounts (no partial ones).
-	const chunkSize = 36 * 44 * 4 // times 4 to make it a reasonable buffer
-
 	f, err := os.Open(path)
 	if os.IsNotExist(err) {
 		return nil, err
@@ -174,7 +152,7 @@ func (s *FileBackend) ReadAccounts() ([]string, error) {
 	}
 
 	// We ballpark the capacity assuming all of the values being G-addresses.
-	preallocationSize := chunkSize / 36
+	preallocationSize := 56 * 100 // default to 100 lines
 	info, err := os.Stat(path)
 	if err == nil { // we can still safely continue w/ errors
 		// Note that this will never be too large, but may be too small.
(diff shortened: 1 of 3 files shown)


exp/lighthorizon/index/file.go
```diff
@@ -122,7 +153,51 @@ func (s *FileBackend) Read(account string) (map[string]*CheckpointIndex, error)
...
-	panic("TODO")
+	path := filepath.Join(s.dir, "accounts")
+	log.Debugf("Opening accounts list at %s", path)
+
+	// The entirety of pubnet accounts in a single file will consume ~400MB
+	// (according to Hubble + napkin math). That should never be done on a
+	// single machine anyway. Thus, if this file is insurmountably large to fit
+	// into memory, you should be splitting the work better. That means we can
+	// safely read it all in one go.
+	buffer, err := os.ReadFile(path)
```
sreuland, line 164: could memory mapped file(mmap-go) be used rather than os.Read, just wondering if that transparently allows the big files, 400m+, without changing anything downstream here, on 32bit it would work up to about 2g, on 64bit it would never max out.

Shaptic: Good idea! Reading in chunks is complicated, though, because you suddenly need to handle binary boundaries between accounts. In other words, if you read `N` bytes at a time, you might only partially read an account into memory.

Shaptic: Wait... :thought_balloon: muxed accounts are always either exactly 4(type)+32(pubkey)=36 or 4+32+8(muxed id)=44 bytes. So if we read in chunks of 36*44=1584, we are guaranteed to read on a boundary! Right? :thinking:

sreuland: where'd the 44 constant originate from? was thinking you could get the byte size of `path` file first with `Fstat()`, that should be fast, no i/o overhead, then can calc how many buffers of x-size(1584?) are present in the mmap to iterate?

Shaptic: The problem with replacing the buffering w/ a buffered reader is that we can't detect EOFs because `muxed.DecodeFrom` returns a string error that hides the underlying `xdr3.UnmarshalError` :(

How about b4d7f484?

exp/lighthorizon/index/file.go
```diff
@@ -37,8 +41,35 @@ func (s *FileBackend) Flush(indexes map[string]map[string]*CheckpointIndex) erro
 func (s *FileBackend) FlushAccounts(accounts []string) error {
 	path := filepath.Join(s.dir, "accounts")
 
-	accountsString := strings.Join(accounts, "\n")
-	return os.WriteFile(path, []byte(accountsString), fs.ModeDir|0755)
+	f, err := os.OpenFile(path, os.O_CREATE|
+		os.O_APPEND| // crucial! since we might flush from various sources
```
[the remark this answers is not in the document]
Shaptic: That's true, but how else do we coordinate parallel (and partial) flushes? If you decide to flush indices to the filesystem every 512 accounts, say, then there is no way to guarantee a lack of dupes nor that you're done writing.

I added some code to return a unique list in 06df7c12, but I'm not sure how to prevent it from being generated in the first place. We could do something dumb like periodically read/unique-ify/flush it, or maybe store the list in memory at all times? But unsure. Ideas?

exp/lighthorizon/index/file.go
sreuland: i see, the file structure is non-delimited, variable length records, which complicates batch parsing, would it help if the records were serialized in the file as fixed length of max, maybe with a temporary xdr struct that pads to 44? then could be deterministic on how many accounts present, and leverage the buffered reader for less code?

Shaptic: I don't think we should pad the structs since that means custom XDR types. However, I realized I can solve my above limitation by using `.Peek(1)` on a buffered reader: 6a1de696 makes the read loop much simpler! I don't think resizing `accounts` is problematic as there should be minimal growth with current pre-allocation strategy.

exp/lighthorizon/index/file.go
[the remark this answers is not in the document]
Shaptic: Tried it out (and reverted it) in 34b9ba2c if you wanna have a look at that.

bartekn, line 45: Does `O_APPEND` is thread safe? Ie. given two go routines writing to a file: is the write a blocking call? I can't see any info in godoc claiming it's true.

> Namely, if each address is plopped in with a trailing newline, you need a newline at the end of the file, otherwise the next append will squish two addresses together on the same line. But, if there's a newline there, and there are concurrent writes, there may be a bunch of newlines creating weird gaps in the file. So when reading, you'd need to filter out the crap lines. Using the raw XDR binary solves all of this, since you just read chunk by chunk and turn that into xdr.MuxedAccounts (since there may be, and are, M-addresses as participants).

Not sure I understand this. Why can't we just write `account` + `\n` together atomically to prevent having empty lines in a file?

exp/lighthorizon/index/file.go
```diff
@@ -122,7 +141,52 @@ func (s *FileBackend) Read(account string) (map[string]*CheckpointIndex, error)
...
+			return accounts, errors.Wrapf(err, "failed to read %s", path)
+		}
+
+		accounts = append(accounts, line[:len(line)-1]) // trim newline
+	}
+
+	// The account list is very unlikely to be unique (especially if it was made
+	// w/ parallel flushes), so let's ensure that that's the case.
+	count := 0
+	accountMap := make(map[string]struct{}, len(accounts))
```
sreuland, line 180: could the intermediate full size array of account strings be avoided by just put'ing to this map on ln 174 instead then iterate the de-dup'd map keys into string slice based on map len for a bit less thrash?

sreuland (approved):
nice work vetting this, looks like ideal solution.

exp/lighthorizon/index/file.go
Shaptic: Absolutely brilliant point - done in 91468ecc.

PR #4423, opened 2022-06-03
exp/lighthorizon: Unify map-reduce and single-process index builders

Shaptic:
### What
A handful of refactors:
* The main thing: `./index/cmd/single` and `./index/cmd/batch/map` now leverage the same index building code (`BuildIndices`)
* Extends the map-reduce builder to take the txmeta source / index destination URLs from the environment rather than hard-coding S3 paths (because splitting that out from the giga-PR was difficult).
* Infer checkpoints from `ledger.LedgerSequence()` rather than passing it as a parameter
* Add a new `ProcessAccountsWithoutBackend` module for the Map job

### Why
One codebase to rule them all :+1: 

This gets us per-ledger index building "for free" for the map-reduce version.

### Known limitations
n/a

### DIFF
--- a/exp/lighthorizon/index/builder.go
+++ b/exp/lighthorizon/index/builder.go
@@ -4,27 +4,166 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"math"
+	"sync/atomic"
+	"time"
 
 	"github.com/stellar/go/historyarchive"
 	"github.com/stellar/go/ingest"
 	"github.com/stellar/go/ingest/ledgerbackend"
+	"github.com/stellar/go/support/errors"
 	"github.com/stellar/go/support/log"
 	"github.com/stellar/go/toid"
 	"github.com/stellar/go/xdr"
+	"golang.org/x/sync/errgroup"
 )
 
-// Module is a way to process data and store it into an index.
+func BuildIndices(
+	ctx context.Context,
+	sourceUrl string, // where is raw txmeta coming from?
+	targetUrl string, // where should the resulting indices go?
+	networkPassphrase string,
+	startLedger, endLedger uint32,
+	modules []string,
+	workerCount int,
+) error {
+	indexStore, err := Connect(targetUrl)
+	if err != nil {
+		return err
+	}
+
+	// Simple file os access
+	source, err := historyarchive.ConnectBackend(
+		sourceUrl,
+		historyarchive.ConnectOptions{
+			Context:           ctx,
+			NetworkPassphrase: networkPassphrase,
+		},
+	)
+	if err != nil {
+		return err
+	}
+
+	ledgerBackend := ledgerbackend.NewHistoryArchiveBackend(source)
+	defer ledgerBackend.Close()
+
+	if endLedger == 0 {
+		latest, err := ledgerBackend.GetLatestLedgerSequence(ctx)
+		if err != nil {
+			return err
+		}
+		endLedger = latest
+	}
+
+	ledgerCount := 1 + (endLedger - startLedger) // +1 because endLedger is inclusive
+	parallel := max(1, workerCount)
+
+	startTime := time.Now()
+	log.Infof("Creating indices for ledger range: %d through %d (%d ledgers)",
+		startLedger, endLedger, ledgerCount)
+	log.Infof("Using %d workers", parallel)
+
+	// Create a bunch of workers that process ledgers a checkpoint range at a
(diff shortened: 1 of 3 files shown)


paulbellamy (approved):

exp/lighthorizon/index/builder.go
```diff
@@ -148,6 +289,48 @@ func ProcessAccounts(
 
 	return nil
 }
+
+func ProcessAccountsWithoutBackend(
```
paulbellamy, line 293: We probably want a `ProcessTransactionsWithoutBackend` as well, no?

paulbellamy (approved):

exp/lighthorizon/index/builder.go
```diff
@@ -4,35 +4,174 @@ import (
...
+		startLedger, endLedger, ledgerCount)
+	log.Infof("Using %d workers", parallel)
+
+	// Create a bunch of workers that process ledgers a checkpoint range at a
+	// time (better than a ledger at a time to minimize flushes).
+	wg, ctx := errgroup.WithContext(ctx)
+	ch := make(chan historyarchive.Range, parallel)
+
+	indexBuilder := NewIndexBuilder(indexStore, ledgerBackend, networkPassphrase)
+	for _, part := range modules {
```
sreuland, line 72: is it worthwhile to use stronger type like enum to represent module types as they increase in number?

Shaptic: I don't think so, because these are supposed to be plug-and-play. Neither their execution order nor enum value would ever matter.

exp/lighthorizon/index/builder.go
Shaptic: There was no code doing that, so I didn't add it :man_shrugging: This new module was ripped straight from the existing variant, and you can see [here](https://github.com/stellar/go/pull/4423/files#diff-6ec60418fbf34172627dc88e3b6d59e46cc20aaa04aa708686fef9c1887844eeL145-L148) that transactions were added with a backend in the original version.

exp/lighthorizon/index/builder.go
```diff
@@ -4,35 +4,174 @@ import (
...
+		default:
+			return fmt.Errorf("Unknown module: %s", part)
+		}
+	}
+
+	// Submit the work to the channels, breaking up the range into individual
+	// checkpoint ranges.
+	go func() {
+		// Recall: A ledger X is a checkpoint ledger iff (X + 1) % 64 == 0
+		nextCheckpoint := (((startLedger / 64) * 64) + 63)
```
bartekn, line 89: Nit: Ideal use case for [`CheckpointManager`](https://pkg.go.dev/github.com/stellar/go@v0.0.0-20220531233648-5c7b70ce9b62/historyarchive#CheckpointManager).

exp/lighthorizon/index/builder.go
```diff
@@ -4,35 +4,174 @@ import (
...
+		wg.Go(func() error {
+			for ledgerRange := range ch {
+				count := (ledgerRange.High - ledgerRange.Low) + 1
+				nprocessed := atomic.AddUint64(&processed, uint64(count))
+
+				log.Debugf("Working on checkpoint range [%d, %d]",
+					ledgerRange.Low, ledgerRange.High)
+
+				// Assertion for testing
+				if ledgerRange.High != endLedger && (ledgerRange.High+1)%64 != 0 {
```
bartekn, line 116: Nit: [`CheckpointManager.IsCheckpoint`](https://pkg.go.dev/github.com/stellar/go@v0.0.0-20220531233648-5c7b70ce9b62/historyarchive#CheckpointManager.IsCheckpoint)

exp/lighthorizon/index/builder.go
```diff
@@ -292,3 +478,47 @@ func getLedgerKeyParticipants(ledgerKey xdr.LedgerKey) []string {
...
+		panic(fmt.Errorf("error for %s: done > total (%d > %d)",
+			prefix, done, total))
+	}
+
+	progress := float64(done) / float64(total)
+	elapsed := time.Since(startTime)
+
+	// Approximate based on how many ledgers are left and how long this much
+	// progress took, e.g. if 4/10 took 2s then 6/10 will "take" 3s (though this
+	// assumes consistent ledger load).
```
bartekn, line 495: Maybe we could improve the estimation by counting ops instead of ledgers?

PR #4424, opened 2022-06-03
exp/lighthorizon: Refactor and repair the reduce job

Shaptic:
### What
This refactors and repairs the reduce job in the following way (:flying_saucer: = refactor, :hammer_and_wrench: = fix):
* :flying_saucer: use environmental variables for every parameter, including index sources and final target
* :flying_saucer: helper functions: merging indices, partitioning work based on hashes
* :flying_saucer: much more logging!
* :flying_saucer: move thread-safe set into its own abstraction
* :hammer_and_wrench: make reduce back-end agnostic rather than tied to S3
* :hammer_and_wrench: generalize non-existence errors (`os.IsNotExist(err)` over `err == ErrNotExist`)
* :hammer_and_wrench: make transaction flushing thread-safe (for parallel reduces)
* :hammer_and_wrench: flush accounts list as part of `Flush()`
* :hammer_and_wrench: split work across multiple channels rather than just one

### Why
Cleaning up the codebase, more logging, easier to follow, etc.

The :hammer_and_wrench: fixes are mostly related to concurrency:
* if flushes aren't thread-safe, we get deadlocks across reduce workers
* if the account list is flushed separately via `FlushAccounts()`, the written file ends up completely empty because a prior `Flush()` cleared indices from memory
* if the work comes from a single channel, accounts can get skipped overall because they aren't put back on the queue if they're skipped by a single worker

### Known limitations
This is part of a larger change, so it's dependent on other PRs (#4423, #4422).

For commit history and finer deltas, see #4420.

### DIFF
--- a/exp/lighthorizon/index/cmd/batch/reduce/main.go
+++ b/exp/lighthorizon/index/cmd/batch/reduce/main.go
@@ -5,245 +5,293 @@ import (
 	"fmt"
 	"hash/fnv"
 	"os"
+	"path/filepath"
 	"strconv"
 	"sync"
 
-	"github.com/aws/aws-sdk-go/aws"
 	"github.com/stellar/go/exp/lighthorizon/index"
+	"github.com/stellar/go/support/errors"
 	"github.com/stellar/go/support/log"
 )
 
-var (
-	// Should we use runtime.NumCPU() for a reasonable default?
-	parallel = uint32(16)
+const (
+	ACCOUNT_FLUSH_FREQUENCY = 200
+	// arbitrary default, should we use runtime.NumCPU()?
+	DEFAULT_WORKER_COUNT = 2
 )
+
+type ReduceConfig struct {
+	JobIndex        uint32
+	MapJobCount     uint32
+	ReduceJobCount  uint32
+	IndexTarget     string
+	IndexRootSource string
+
+	Workers uint32
+}
+
+func ReduceConfigFromEnvironment() (*ReduceConfig, error) {
+	const (
+		mapJobsEnv          = "MAP_JOB_COUNT"
+		reduceJobsEnv       = "REDUCE_JOB_COUNT"
+		workerCountEnv      = "WORKER_COUNT"
+		jobIndexEnv         = "AWS_BATCH_JOB_ARRAY_INDEX"
+		mappedIndicesUrlEnv = "INDEX_SOURCE_ROOT"
+		finalIndexUrlEnv    = "INDEX_TARGET"
+	)
+
+	jobIndex, err := strconv.ParseUint(os.Getenv(jobIndexEnv), 10, 32)
+	if err != nil {
+		return nil, errors.Wrap(err, "invalid parameter "+jobIndexEnv)
+	}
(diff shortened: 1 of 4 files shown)


exp/lighthorizon/index/cmd/batch/reduce/set.go
```diff
@@ -0,0 +1,29 @@
+package main
+
+import "sync"
+
+// SafeStringSet is a simple thread-safe set.
+type SafeStringSet struct {
+    lock sync.Mutex
```
paulbellamy, line 7: why not?
```suggestion
    lock sync.RWMutex
```

paulbellamy (approved):

exp/lighthorizon/index/cmd/batch/reduce/main.go
```diff
@@ -5,250 +5,354 @@ import (
...
+				defer wg.Done()
+				logger := logger.
+					WithField("worker", routineIndex).
+					WithField("total", len(accounts))
+				logger.Info("Started worker")
+
+				var accountsProcessed, accountsSkipped uint64
+				for account := range workQueues[routineIndex] {
+					logger.Infof("Account: %s", account)
+					if (accountsProcessed+accountsSkipped)%97 == 0 {
```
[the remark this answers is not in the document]
Shaptic: just an arbitrary prime number to give "interesting" progress updates rather than e.g. every 100

exp/lighthorizon/index/cmd/batch/reduce/set.go
Shaptic: Well I kept it the same from the existing code, but no particular reason it can't be a RW, I guess :thinking:

Shaptic: Done in e0a0436 :+1:

exp/lighthorizon/index/cmd/batch/reduce/main.go
```diff
@@ -5,250 +5,354 @@ import (
...
 	"github.com/stellar/go/exp/lighthorizon/index"
+	"github.com/stellar/go/support/errors"
 	"github.com/stellar/go/support/log"
 )
 
-var (
-	// Should we use runtime.NumCPU() for a reasonable default?
-	parallel = uint32(16)
+const (
+	ACCOUNT_FLUSH_FREQUENCY = 200
```
bartekn, line 18: Nit: the convention for `const`s is camel-case.

exp/lighthorizon/index/cmd/batch/reduce/main.go
```diff
@@ -5,250 +5,354 @@ import (
...
+
+	return nil
+}
+
+func (cfg *ReduceConfig) shouldProcessAccount(account string, routineIndex uint32) bool {
+	hash := fnv.New64a()
+
+	// Docs state (https://pkg.go.dev/hash#Hash) that Write will never error.
+	hash.Write([]byte(account))
+	digest := uint32(hash.Sum64()) // discard top 32 bits
```
bartekn, line 311: If only 32 bits are needed maybe `New32a` can be used here?

PR #4431, opened 2022-06-13
exp/lighthorizon: enforce the limit from request on the response size 

### LINKED ISSUE #4430: exp/lighthorizon: Correctly paginate responses to respect query limits
page limit parameter on requests should determine the response results sizee
contributes to #4317 development tasks, specifically to implement correct paging behavior for `limit`

A request to serve `?limit=N` of a resource should actually serve N of that resource, rather than reading N ledgers and serving `0 <= M <= N` instances matching the requested resource (start in ./archive/main.go).

--------

### Details

In `./archive/wrapper.go`, we see two methods: `GetTransactions()` and `GetOperations()`. For both of these cases, the `limit` parameter is reached via the following (abridged) code:

```golang
txns = append(txns, common.Transaction{TransactionEnvelope: &tx.Envelope, /* etc. */})
if int64(len(txns)) == limit {
	return txns, nil
}
```

where `tx` comes from `reader.Read()`, an `ingest.LedgerTransactionReader` instance built from `ledger`. `ledger` comes from `Wrapper.GetLedger(ledgerSequence)` (where the sequence is derived from the cursor), and this corresponds to a lookup to a txmeta dump (see `./main.go`).

**Key point:** We want to see indices utilized when finding ledgers to look up, rather than just using them as a starting point for the cursor (see `actions/transaction.go`: all the index store is used for is advancing `cursor` to the next active ledger). We should move index usage to within the `GetTransactions()` method and advance to the next active ledger, pulling transactions until the `limit` is fulfilled.

sreuland:
check the page limit parameter on tx and ops listing requests, it should determine the response results size.
Closes #4430 

Confirmed by live testing, exported and indexed small ledger range,

START_LEDGER= 1538223
STOP_LEDGER= 1538323

`/transactions` processes the cursor and limit query string parameters, if no limit defaults to 10

request by cursor: l=1538223, t=1, o=1
 
`GET localhost:8080/transactions?cursor=6606617478959105&limit=2`

```
{
  "_links": {
    "self": {
      "href": "http://localhost:8080/transactions?cursor=\u0026limit=0\u0026order="
    },
    "next": {
      "href": "http://localhost:8080/transactions?cursor=6606621773930497\u0026limit=0\u0026order="
    },
    "prev": {
      "href": "http://localhost:8080/transactions?cursor=6606621773926401\u0026limit=0\u0026order=asc"
    }
  },
  "_embedded": {
    "records": [
      {
        "memo": "xdr.Memo{Type: xdr.MemoTypeMemoNone}",
        "_links": {
          "self": {
            "href": "http://localhost:8080/transactions/544469b76cd90978345a4734a0ce69a9d0ddb4a6595a7afc503225a77826722a"
          },
          "account": {
            "href": ""
          },
          "ledger": {
            "href": ""
          },
          "operations": {
            "href": ""
          },
          "effects": {
            "href": "http://localhost:8080/transactions/544469b76cd90978345a4734a0ce69a9d0ddb4a6595a7afc503225a77826722a/effects"
          },
          "precedes": {
            "href": "http://localhost:8080/effects?order=asc\u0026cursor=6606621773926401"
          },
          "succeeds": {
            "href": "http://localhost:8080/effects?order=desc\u0026cursor=6606621773926401"
          },
          "transaction": {
            "href": ""
          }
        },
        "id": "544469b76cd90978345a4734a0ce69a9d0ddb4a6595a7afc503225a77826722a",
        "paging_token": "6606621773926401",
        "successful": true,
        "hash": "544469b76cd90978345a4734a0ce69a9d0ddb4a6595a7afc503225a77826722a",
        "ledger": 1538224,
        "created_at": "2022-06-17T23:29:42Z",
        "source_account": "GBGTCH47BOEEKLPHHMR2GOK6KQFGL3O7Q53FIZTJ7S7YEDWYJ5IUDJDJ",
        "source_account_sequence": "",
        "fee_account": "",
        "fee_charged": "84000",
        "max_fee": "0",
        "operation_count": 28,
        "envelope_xdr": "AAAAAgAAAABNMR-fC4hFLec7I6M5XlQKZe3fh3ZUZmn8v4IO2E9RQQABSCAADOFXAABwTwAAAAAAAAAAAAAAHAAAAAAAAAADAAAAAUVVUlYAAAAAL3ZBkx0hzBH3SxRQmbsxJvS0lyUOUUEqMkdUbvATQ00AAAABVVNEVgAAAAAvdkGTHSHMEfdLFFCZuzEm9LSXJQ5RQSoyR1Ru8BNDTQAAAAAAAAAAAAAAAQCYloAAAAAAAAZ-OwAAAAAAAAADAAAAAUVVUlYAAAAAL3ZBkx0hzBH3SxRQmbsxJvS0lyUOUUEqMkdUbvATQ00AAAABVVNEVgAAAAAvdkGTHSHMEfdLFFCZuzEm9LSXJQ5RQSoyR1Ru8BNDTQAAAAAAAAAAAAAAAQCYloAAAAAAAAZ-PgAAAAAAAAADAAAAAUVVUlYAAAAAL3ZBkx0hzBH3SxRQmbsxJvS0lyUOUUEqMkdUbvATQ00AAAABVVNEVgAAAAAvdkGTHSHMEfdLFFCZuzEm9LSXJQ5RQSoyR1Ru8BNDTQAAAAAAAAAAAAAAAQCYloAAAAAAAAZ-OAAAAAAAAAADAAAAAUVVUlYAAAAAL3ZBkx0hzBH3SxRQmbsxJvS0lyUOUUEqMkdUbvATQ00AAAABVVNEVgAAAAAvdkGTHSHMEfdLFFCZuzEm9LSXJQ5RQSoyR1Ru8BNDTQAAAAAAAAAAAAAAAQCYloAAAAAAAAZ-OgAAAAAAAAADAAAAAUVVUlYAAAAAL3ZBkx0hzBH3SxRQmbsxJvS0lyUOUUEqMkdUbvATQ00AAAABVVNEVgAAAAAvdkGTHSHMEfdLFFCZuzEm9LSXJQ5RQSoyR1Ru8BNDTQAAAAAAAAAAAAAAAQCYloAAAAAAAAZ-PQAAAAAAAAADAAAAAUVVUlYAAAAAL3ZBkx0hzBH3SxRQmbsxJvS0lyUOUUEqMkdUbvATQ00AAAABVVNEVgAAAAAvdkGTHSHMEfdLFFCZuzEm9LSXJQ5RQSoyR1Ru8BNDTQAAAAAAAAAAAAAAAQCYloAAAAAAAAZ-OQAAAAAAAAADAAAAAUVVUlYAAAAAL3ZBkx0hzBH3SxRQmbsxJvS0lyUOUUEqMkdUbvATQ00AAAABVVNEVgAAAAAvdkGTHSHMEfdLFFCZuzEm9LSXJQ5RQSoyR1Ru8BNDTQAAAAAAAAAAAAAAAQCYloAAAAAAAAZ-PAAAAAAAAAAMAAAAAVVTRFYAAAAAL3ZBkx0hzBH3SxRQmbsxJvS0lyUOUUEqMkdUbvATQ00AAAABRVVSVgAAAAAvdkGTHSHMEfdLFFCZuzEm9LSXJQ5RQSoyR1Ru8BNDTQAAAAAAAAAAAAAAAQCYloAAAAAAAAZ-QwAAAAAAAAAMAAAAAVVTRFYAAAAAL3ZBkx0hzBH3SxRQmbsxJvS0lyUOUUEqMkdUbvATQ00AAAABRVVSVgAAAAAvdkGTHSHMEfdLFFCZuzEm9LSXJQ5RQSoyR1Ru8BNDTQAAAAAAAAAAAAAAAQCYloAAAAAAAAZ-QQAAAAAAAAAMAAAAAVVTRFYAAAAAL3ZBkx0hzBH3SxRQmbsxJvS0lyUOUUEqMkdUbvATQ00AAAABRVVSVgAAAAAvdkGTHSHMEfdLFFCZuzEm9LSXJQ5RQSoyR1Ru8BNDTQAAAAAAAAAAAAAAAQCYloAAAAAAAAZ-PwAAAAAAAAAMAAAAAVVTRFYAAAAAL3ZBkx0hzBH3SxRQmbsxJvS0lyUOUUEqMkdUbvATQ00AAAABRVVSVgAAAAAvdkGTHSHMEfdLFFCZuzEm9LSXJQ5RQSoyR1Ru8BNDTQAAAAAAAAAAAAAAAQCYloAAAAAAAAZ-QgAAAAAAAAAMAAAAAVVTRFYAAAAAL3ZBkx0hzBH3SxRQmbsxJvS0lyUOUUEqMkdUbvATQ00AAAABRVVSVgAAAAAvdkGTHSHMEfdLFFCZuzEm9LSXJQ5RQSoyR1Ru8BNDTQAAAAAAAAAAAAAAAQCYloAAAAAAAAZ-QAAAAAAAAAAMAAAAAVVTRFYAAAAAL3ZBkx0hzBH3SxRQmbsxJvS0lyUOUUEqMkdUbvATQ00AAAABRVVSVgAAAAAvdkGTHSHMEfdLFFCZuzEm9LSXJQ5RQSoyR1Ru8BNDTQAAAAAAAAAAAAAAAQCYloAAAAAAAAZ-RAAAAAAAAAAMAAAAAVVTRFYAAAAAL3ZBkx0hzBH3SxRQmbsxJvS0lyUOUUEqMkdUbvATQ00AAAABRVVSVgAAAAAvdkGTHSHMEfdLFFCZuzEm9LSXJQ5RQSoyR1Ru8BNDTQAAAAAAAAAAAAAAAQCYloAAAAAAAAZ-RQAAAAAAAAADAAAAAUVVUlYAAAAAL3ZBkx0hzBH3SxRQmbsxJvS0lyUOUUEqMkdUbvATQ00AAAABVVNEVgAAAAAvdkGTHSHMEfdLFFCZuzEm9LSXJQ5RQSoyR1Ru8BNDTQAAAAAEDnlMQh53yD1RDVAAAAAAAAAAAAAAAAAAAAADAAAAAUVVUlYAAAAAL3ZBkx0hzBH3SxRQmbsxJvS0lyUOUUEqMkdUbvATQ00AAAABVVNEVgAAAAAvdkGTHSHMEfdLFFCZuzEm9LSXJQ5RQSoyR1Ru8BNDTQAAAAAIHPJwL02EVCvo-8wAAAAAAAAAAAAAAAAAAAADAAAAAUVVUlYAAAAAL3ZBkx0hzBH3SxRQmbsxJvS0lyUOUUEqMkdUbvATQ00AAAABVVNEVgAAAAAvdkGTHSHMEfdLFFCZuzEm9LSXJQ5RQSoyR1Ru8BNDTQAAAAAMK2uoSl41QEUaEEAAAAAAAAAAAAAAAAAAAAADAAAAAUVVUlYAAAAAL3ZBkx0hzBH3SxRQmbsxJvS0lyUOUUEqMkdUbvATQ00AAAABVVNEVgAAAAAvdkGTHSHMEfdLFFCZuzEm9LSXJQ5RQSoyR1Ru8BNDTQAAAAAQOeVELled8CsaWcAAAAAAAAAAAAAAAAAAAAADAAAAAUVVUlYAAAAAL3ZBkx0hzBH3SxRQmbsxJvS0lyUOUUEqMkdUbvATQ00AAAABVVNEVgAAAAAvdkGTHSHMEfdLFFCZuzEm9LSXJQ5RQSoyR1Ru8BNDTQAAAAAUSF58KBZa5CVSZpAAAAAAAAAAAAAAAAAAAAADAAAAAUVVUlYAAAAAL3ZBkx0hzBH3SxRQmbsxJvS0lyUOUUEqMkdUbvATQ00AAAABVVNEVgAAAAAvdkGTHSHMEfdLFFCZuzEm9LSXJQ5RQSoyR1Ru8BNDTQAAAAAYVte0C3eGiAqvtQQAAAAAAAAAAAAAAAAAAAADAAAAAUVVUlYAAAAAL3ZBkx0hzBH3SxRQmbsxJvS0lyUOUUEqMkdUbvATQ00AAAABVVNEVgAAAAAvdkGTHSHMEfdLFFCZuzEm9LSXJQ5RQSoyR1Ru8BNDTQAAAAAcZVDsOkaLRDZcvfAAAAAAAAAAAAAAAAAAAAAMAAAAAVVTRFYAAAAAL3ZBkx0hzBH3SxRQmbsxJvS0lyUOUUEqMkdUbvATQ00AAAABRVVSVgAAAAAvdkGTHSHMEfdLFFCZuzEm9LSXJQ5RQSoyR1Ru8BNDTQAAAAAcZVDsZMg_SGDvpLAAAAAAAAAAAAAAAAAAAAAMAAAAAVVTRFYAAAAAL3ZBkx0hzBH3SxRQmbsxJvS0lyUOUUEqMkdUbvATQ00AAAABRVVSVgAAAAAvdkGTHSHMEfdLFFCZuzEm9LSXJQ5RQSoyR1Ru8BNDTQAAAAAYVte0Gte2_BnYMCQAAAAAAAAAAAAAAAAAAAAMAAAAAVVTRFYAAAAAL3ZBkx0hzBH3SxRQmbsxJvS0lyUOUUEqMkdUbvATQ00AAAABRVVSVgAAAAAvdkGTHSHMEfdLFFCZuzEm9LSXJQ5RQSoyR1Ru8BNDTQAAAAAUSF58NIsz0DKkMfwAAAAAAAAAAAAAAAAAAAAMAAAAAVVTRFYAAAAAL3ZBkx0hzBH3SxRQmbsxJvS0lyUOUUEqMkdUbvATQ00AAAABRVVSVgAAAAAvdkGTHSHMEfdLFFCZuzEm9LSXJQ5RQSoyR1Ru8BNDTQAAAAAQOeVEbwopqGsg3aAAAAAAAAAAAAAAAAAAAAAMAAAAAVVTRFYAAAAAL3ZBkx0hzBH3SxRQmbsxJvS0lyUOUUEqMkdUbvATQ00AAAABRVVSVgAAAAAvdkGTHSHMEfdLFFCZuzEm9LSXJQ5RQSoyR1Ru8BNDTQAAAAAMK2uoe_bRSHe4OVAAAAAAAAAAAAAAAAAAAAAMAAAAAVVTRFYAAAAAL3ZBkx0hzBH3SxRQmbsxJvS0lyUOUUEqMkdUbvATQ00AAAABRVVSVgAAAAAvdkGTHSHMEfdLFFCZuzEm9LSXJQ5RQSoyR1Ru8BNDTQAAAAAIHPJwaZQbUGYRTZAAAAAAAAAAAAAAAAAAAAAMAAAAAVVTRFYAAAAAL3ZBkx0hzBH3SxRQmbsxJvS0lyUOUUEqMkdUbvATQ00AAAABRVVSVgAAAAAvdkGTHSHMEfdLFFCZuzEm9LSXJQ5RQSoyR1Ru8BNDTQAAAAAEDnlMCU7cgAkB-ngAAAAAAAAAAAAAAAAAAAAB2E9RQQAAAEDOHy2IQJTNEz9kgUbHYolo_qjWo5s1ZkHOo70AQM2DRVWSNb4Py043HLbgZ6B5WdgkAs3gzoQ9sYk77KVZ5CsG",
        "result_xdr": "AAAAAAABSCAAAAAAAAAAHAAAAAAAAAADAAAAAAAAAAAAAAACAAAAAAAAAAMAAAAAAAAAAAAAAAIAAAAAAAAAAwAAAAAAAAAAAAAAAgAAAAAAAAADAAAAAAAAAAAAAAACAAAAAAAAAAMAAAAAAAAAAAAAAAIAAAAAAAAAAwAAAAAAAAAAAAAAAgAAAAAAAAADAAAAAAAAAAAAAAACAAAAAAAAAAwAAAAAAAAAAAAAAAIAAAAAAAAADAAAAAAAAAAAAAAAAgAAAAAAAAAMAAAAAAAAAAAAAAACAAAAAAAAAAwAAAAAAAAAAAAAAAIAAAAAAAAADAAAAAAAAAAAAAAAAgAAAAAAAAAMAAAAAAAAAAAAAAACAAAAAAAAAAwAAAAAAAAAAAAAAAIAAAAAAAAAAwAAAAAAAAAAAAAAAAAAAABNMR-fC4hFLec7I6M5XlQKZe3fh3ZUZmn8v4IO2E9RQQAAAAAABn5GAAAAAUVVUlYAAAAAL3ZBkx0hzBH3SxRQmbsxJvS0lyUOUUEqMkdUbvATQ00AAAABVVNEVgAAAAAvdkGTHSHMEfdLFFCZuzEm9LSXJQ5RQSoyR1Ru8BNDTQAAAAAEDnlMQh53yD1RDVAAAAAAAAAAAAAAAAAAAAADAAAAAAAAAAAAAAAAAAAAAE0xH58LiEUt5zsjozleVApl7d-HdlRmafy_gg7YT1FBAAAAAAAGfkcAAAABRVVSVgAAAAAvdkGTHSHMEfdLFFCZuzEm9LSXJQ5RQSoyR1Ru8BNDTQAAAAFVU0RWAAAAAC92QZMdIcwR90sUUJm7MSb0tJclDlFBKjJHVG7wE0NNAAAAAAgc8nAvTYRUK-j7zAAAAAAAAAAAAAAAAAAAAAMAAAAAAAAAAAAAAAAAAAAATTEfnwuIRS3nOyOjOV5UCmXt34d2VGZp_L-CDthPUUEAAAAAAAZ-SAAAAAFFVVJWAAAAAC92QZMdIcwR90sUUJm7MSb0tJclDlFBKjJHVG7wE0NNAAAAAVVTRFYAAAAAL3ZBkx0hzBH3SxRQmbsxJvS0lyUOUUEqMkdUbvATQ00AAAAADCtrqEpeNUBFGhBAAAAAAAAAAAAAAAAAAAAAAwAAAAAAAAAAAAAAAAAAAABNMR-fC4hFLec7I6M5XlQKZe3fh3ZUZmn8v4IO2E9RQQAAAAAABn5JAAAAAUVVUlYAAAAAL3ZBkx0hzBH3SxRQmbsxJvS0lyUOUUEqMkdUbvATQ00AAAABVVNEVgAAAAAvdkGTHSHMEfdLFFCZuzEm9LSXJQ5RQSoyR1Ru8BNDTQAAAAAQOeVELled8CsaWcAAAAAAAAAAAAAAAAAAAAADAAAAAAAAAAAAAAAAAAAAAE0xH58LiEUt5zsjozleVApl7d-HdlRmafy_gg7YT1FBAAAAAAAGfkoAAAABRVVSVgAAAAAvdkGTHSHMEfdLFFCZuzEm9LSXJQ5RQSoyR1Ru8BNDTQAAAAFVU0RWAAAAAC92QZMdIcwR90sUUJm7MSb0tJclDlFBKjJHVG7wE0NNAAAAABRIXnwoFlrkJVJmkAAAAAAAAAAAAAAAAAAAAAMAAAAAAAAAAAAAAAAAAAAATTEfnwuIRS3nOyOjOV5UCmXt34d2VGZp_L-CDthPUUEAAAAAAAZ-SwAAAAFFVVJWAAAAAC92QZMdIcwR90sUUJm7MSb0tJclDlFBKjJHVG7wE0NNAAAAAVVTRFYAAAAAL3ZBkx0hzBH3SxRQmbsxJvS0lyUOUUEqMkdUbvATQ00AAAAAGFbXtAt3hogKr7UEAAAAAAAAAAAAAAAAAAAAAwAAAAAAAAAAAAAAAAAAAABNMR-fC4hFLec7I6M5XlQKZe3fh3ZUZmn8v4IO2E9RQQAAAAAABn5MAAAAAUVVUlYAAAAAL3ZBkx0hzBH3SxRQmbsxJvS0lyUOUUEqMkdUbvATQ00AAAABVVNEVgAAAAAvdkGTHSHMEfdLFFCZuzEm9LSXJQ5RQSoyR1Ru8BNDTQAAAAAcZVDsOkaLRDZcvfAAAAAAAAAAAAAAAAAAAAAMAAAAAAAAAAAAAAAAAAAAAE0xH58LiEUt5zsjozleVApl7d-HdlRmafy_gg7YT1FBAAAAAAAGfk0AAAABVVNEVgAAAAAvdkGTHSHMEfdLFFCZuzEm9LSXJQ5RQSoyR1Ru8BNDTQAAAAFFVVJWAAAAAC92QZMdIcwR90sUUJm7MSb0tJclDlFBKjJHVG7wE0NNAAAAAB2FvShg76SwZMg_SAAAAAAAAAAAAAAAAAAAAAwAAAAAAAAAAAAAAAAAAAAATTEfnwuIRS3nOyOjOV5UCmXt34d2VGZp_L-CDthPUUEAAAAAAAZ-TgAAAAFVU0RWAAAAAC92QZMdIcwR90sUUJm7MSb0tJclDlFBKjJHVG7wE0NNAAAAAUVVUlYAAAAAL3ZBkx0hzBH3SxRQmbsxJvS0lyUOUUEqMkdUbvATQ00AAAAAGUd8mxnYMCQa17b8AAAAAAAAAAAAAAAAAAAADAAAAAAAAAAAAAAAAAAAAABNMR-fC4hFLec7I6M5XlQKZe3fh3ZUZmn8v4IO2E9RQQAAAAAABn5PAAAAAVVTRFYAAAAAL3ZBkx0hzBH3SxRQmbsxJvS0lyUOUUEqMkdUbvATQ00AAAABRVVSVgAAAAAvdkGTHSHMEfdLFFCZuzEm9LSXJQ5RQSoyR1Ru8BNDTQAAAAAVC2xdMqQx_DSLM9AAAAAAAAAAAAAAAAAAAAAMAAAAAAAAAAAAAAAAAAAAAE0xH58LiEUt5zsjozleVApl7d-HdlRmafy_gg7YT1FBAAAAAAAGflAAAAABVVNEVgAAAAAvdkGTHSHMEfdLFFCZuzEm9LSXJQ5RQSoyR1Ru8BNDTQAAAAFFVVJWAAAAAC92QZMdIcwR90sUUJm7MSb0tJclDlFBKjJHVG7wE0NNAAAAABDRjn1rIN2gbwopqAAAAAAAAAAAAAAAAAAAAAwAAAAAAAAAAAAAAAAAAAAATTEfnwuIRS3nOyOjOV5UCmXt34d2VGZp_L-CDthPUUEAAAAAAAZ-UQAAAAFVU0RWAAAAAC92QZMdIcwR90sUUJm7MSb0tJclDlFBKjJHVG7wE0NNAAAAAUVVUlYAAAAAL3ZBkx0hzBH3SxRQmbsxJvS0lyUOUUEqMkdUbvATQ00AAAAADJnfsXe4OVB79tFIAAAAAAAAAAAAAAAAAAAADAAAAAAAAAAAAAAAAAAAAABNMR-fC4hFLec7I6M5XlQKZe3fh3ZUZmn8v4IO2E9RQQAAAAAABn5SAAAAAVVTRFYAAAAAL3ZBkx0hzBH3SxRQmbsxJvS0lyUOUUEqMkdUbvATQ00AAAABRVVSVgAAAAAvdkGTHSHMEfdLFFCZuzEm9LSXJQ5RQSoyR1Ru8BNDTQAAAAAIZGP3ZhFNkGmUG1AAAAAAAAAAAAAAAAAAAAAMAAAAAAAAAAAAAAAAAAAAAE0xH58LiEUt5zsjozleVApl7d-HdlRmafy_gg7YT1FBAAAAAAAGflMAAAABVVNEVgAAAAAvdkGTHSHMEfdLFFCZuzEm9LSXJQ5RQSoyR1Ru8BNDTQAAAAFFVVJWAAAAAC92QZMdIcwR90sUUJm7MSb0tJclDlFBKjJHVG7wE0NNAAAAAAQxGOcJAfp4CU7cgAAAAAAAAAAAAAAAAA==",
        "result_meta_xdr": "",
        "fee_meta_xdr": "",
        "memo_type": "MemoTypeMemoNone",
        "signatures": [
          "2E9RQQAAAEDOHy2IQJTNEz9kgUbHYolo_qjWo5s1ZkHOo70AQM2DRVWSNb4Py043HLbgZ6B5WdgkAs3gzoQ9sYk77KVZ5CsG"
        ]
      },
      {
        "memo": "xdr.MemoText(\"psp:1405\")",
        "_links": {
          "self": {
            "href": "http://localhost:8080/transactions/5fef21d5ef75ecf18d65a160cfab17dca8dbf6dbc4e2fd66a510719ad8dddb09"
          },
          "account": {
            "href": ""
          },
          "ledger": {
            "href": ""
          },
          "operations": {
            "href": ""
          },
          "effects": {
            "href": "http://localhost:8080/transactions/5fef21d5ef75ecf18d65a160cfab17dca8dbf6dbc4e2fd66a510719ad8dddb09/effects"
          },
          "precedes": {
            "href": "http://localhost:8080/effects?order=asc\u0026cursor=6606621773930497"
          },
          "succeeds": {
            "href": "http://localhost:8080/effects?order=desc\u0026cursor=6606621773930497"
          },
          "transaction": {
            "href": ""
          }
        },
        "id": "5fef21d5ef75ecf18d65a160cfab17dca8dbf6dbc4e2fd66a510719ad8dddb09",
        "paging_token": "6606621773930497",
        "successful": false,
        "hash": "5fef21d5ef75ecf18d65a160cfab17dca8dbf6dbc4e2fd66a510719ad8dddb09",
        "ledger": 1538224,
        "created_at": "2022-06-17T23:29:42Z",
        "source_account": "GCFJN22UG6IZHXKDVAJWAVEQ3NERGCRCURR2FHARNRBNLYFEQZGML4PW",
        "source_account_sequence": "",
        "fee_account": "",
        "fee_charged": "3000",
        "max_fee": "0",
        "operation_count": 1,
        "envelope_xdr": "AAAAAgAAAACKlutUN5GT3UOoE2BUkNtJEwoipGOinBFsQtXgpIZMxQAAJxAAE05oAAHUKAAAAAEAAAAAAAAAAAAAAABirQ6AAAAAAQAAAAhwc3A6MTQwNQAAAAEAAAAAAAAAAQAAAADpPdN37FA9KVcJfmMBuD8pPcaT5jqlrMeYEOTP36Zo2AAAAAJBVE1ZUgAAAAAAAAAAAAAAZ8rWY3iaDnWNtfpvLpNaCEbKdDjrd2gQODOuKpmj1vMAAAAAGHAagAAAAAAAAAABpIZMxQAAAEDNJwYToiBR6bzElRL4ORJdXXZYO9cE3-ishQLC_fWGrPGhWrW7_UkPJWvxWdQDJBjVOHuA1Jjc94NSe91hSwEL",
        "result_xdr": "AAAAAAAAC7j_____AAAAAQAAAAAAAAAB____-gAAAAA=",
        "result_meta_xdr": "",
        "fee_meta_xdr": "",
        "memo_type": "MemoTypeMemoText",
        "signatures": [
          "pIZMxQAAAEDNJwYToiBR6bzElRL4ORJdXXZYO9cE3-ishQLC_fWGrPGhWrW7_UkPJWvxWdQDJBjVOHuA1Jjc94NSe91hSwEL"
        ]
      }
    ]
  }
}

```


exp/lighthorizon/actions/static/api_docs.yml
```diff
@@ -0,0 +1,243 @@
+openapi: 3.0.3
```
sreuland, line 1: should we establish something for capturing docs on the API, put this oapi/yml in for consideration, the content can be migrated to other formats later as desired, intent is just to provide a staring place, close to the code, to consolidate for time being. the docs are published on root path of api server, if you paste the content from that into editor.swagger.io, you'll get the visual rendering.

Screen Shot 2022-06-20 at 3 06 30 PM

Shaptic:
Just getting around to reviewing this - sorry for the delay. Are you sure the noted test actually respects the limit? I see 

```js
        "ledger": 1538224
```

in both records, meaning it never had to look up or ingest another.

exp/lighthorizon/actions/transaction.go
```diff
@@ -22,89 +22,87 @@ func Transactions(archiveWrapper archive.Wrapper, indexStore index.Store) func(h
...
 			if err != nil {
-				fmt.Fprintf(w, "Error: %v", err)
+				sendErrorResponse(w, http.StatusBadRequest, "Invalid transaction id request parameter")
 				return
 			}
 			if len(b) != 32 {
-				fmt.Fprintf(w, "Error: invalid hash")
+				sendErrorResponse(w, http.StatusBadRequest, "Invalid transaction id request parameter")
 				return
 			}
```
Shaptic, line 71: these should be easy to combine now, e.g. `if err != nil || len(b) != 32`. if we want to keep them separate, it might be nice to include more details (e.g. "invalid hash: " + err vs. "invalid hash length")

exp/lighthorizon/actions/main.go
```diff
@@ -0,0 +1,102 @@
...
+
+	if limitRequested != "" {
+		paginate.Limit, err = strconv.ParseInt(limitRequested, 10, 64)
+		if err != nil {
+			return Pagination{}, err
+		}
+	}
+
+	if orderRequested != "" && orderRequested == string(OrderDesc) {
+		paginate.Order = OrderDesc
```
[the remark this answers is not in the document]
sreuland: how do you feel about the minimal approach first, having the api do some opinionated error proofing rather than more chatter? will put in the error response, but wanted to get a read on that style aspect first?

exp/lighthorizon/actions/static/api_docs.yml
sreuland: yes, added mention of this as Phase 1 task, thanks!

sreuland:
> Just getting around to reviewing this - sorry for the delay. Are you sure the noted test actually respects the limit? I see
> 
> ```js
>         "ledger": 1538224
> ```
> 
> in both records, meaning it never had to look up or ingest another. So it may be that it works w/ multiple transactions w/in the same ledger, but not for multiple.

great catch, I think have fixed with [this commit](https://github.com/stellar/go/pull/4431/commits/a5a8b83b0e69f6023ba453d2748b62ebd2270bbe). `archive.GetTransactions` wasn't incrementing tx order num all the time, resulted in skipping all tx's in the first ledger encoded in cursor.

exp/lighthorizon/actions/main.go
```diff
@@ -0,0 +1,102 @@
...
+	var cursorRequested, limitRequested, orderRequested string
+	var err error
+	paginate := Pagination{
+		Order: OrderAsc,
+	}
+
+	if cursorRequested, err = RequestUnaryParam(r, "cursor"); err != nil {
+		return Pagination{}, err
+	}
+
```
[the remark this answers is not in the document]
sreuland: I was following guard check pattern, no prob going with this, will update, thx.

exp/lighthorizon/actions/operation.go
```diff
@@ -26,102 +24,81 @@ func Operations(archiveWrapper archive.Wrapper, indexStore index.Store) func(htt
...
+			log.Error(err)
+			sendErrorResponse(w, http.StatusInternalServerError, "")
+			return
+		}
+
 		if account != "" {
 			// Skip the cursor ahead to the next active checkpoint for this account
-			checkpoint, err := indexStore.NextActive(account, "all_all", uint32(toid.Parse(cursor).LedgerSequence/64))
+			var checkpoint uint32
+			checkpoint, err = indexStore.NextActive(account, "all_all", uint32(toid.Parse(paginate.Cursor).LedgerSequence/64))
```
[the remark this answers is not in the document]
sreuland: got it, thx, we should probably consider refactor, pulling these up into type-safe constants maybe later

exp/lighthorizon/actions/transaction.go
sreuland: yes, good suggestion, added more detail on each error case, to help the caller out - https://github.com/stellar/go/pull/4431/commits/9e401e5b49b89c8d28655be296349a836a4dcc0e

exp/lighthorizon/actions/main.go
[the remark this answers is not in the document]
Shaptic: I have no strong opinion, agreed that we should keep things minimal to keep eyes on the prize - aka working mvp

exp/lighthorizon/archive/main.go
```diff
@@ -69,6 +70,7 @@ func (a *Wrapper) GetOperations(cursor int64, limit int64) ([]common.Operation,
 
 			for operationOrder := range tx.Envelope.Operations() {
 				currID := toid.New(int32(ledgerSequence), transactionOrder+1, int32(operationOrder+1)).ToInt64()
+				transactionOrder++
```
Shaptic, line 73: Wait, I'm not entirely sure this is correct. cc @stellar/horizon-committers - does anyone know how TOIDs work? I get operation order but I don't understand transaction order. Is it essentially an index into the list of transactions in a ledger?

Shaptic (commented):
Seems like we're close! Would be **really** good to get some test cases as a follow-up so that we can analyze exactly what behavior is expected/reproducible/working, maybe in a follow-up PR.

exp/lighthorizon/archive/main.go
sreuland: that was a mistake on my part, i incremented it in wrong place, I [pushed a correction ](https://github.com/stellar/go/pull/4431/commits/a6d26c478cf65a6fcc75f377c9b61a293edd5d93)to only increment the tx order on the outside of ops loop.

On you point about not using the index, it appears like multiple optimal paths for the queries, when just cursor and no other query params, the results list is just a sequential listing of ops or txs from that start point, and can resolve that with just toid and the ledger backend reader, there's no value to using index there right, however, when the account query parameter is included at least as here on ops, then index.nextActive needs to be used rather than sequential toid iteration. 

I think I'm working on this now for [#4433](https://github.com/stellar/go/issues/4433), the net will be refactor the account param off  `/operations?account=` and over to `/accounts/:id/operations?cursor=&limit=` and will be changing `archive.GetOperations` to leverage index.nextActive in the way you're recommending.

sreuland: @Shaptic , I added [tests on archive op/tx retreival ](https://github.com/stellar/go/pull/4431/files#diff-9f1cb5c5eda48dde3ea6ed96d954478f841d51a81f4f4c592efc64de157b9120), asserts what data is on the index/archive side and what's expected on the results side, may help detect if there is a gap in current impl vs. design intent.

2opremio: According to [SEP 35](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0035.md) both the transaction and operation order start at 1:

> Transaction Application Order: the order that the transaction was applied within the ledger where it was validated. This takes up 20 bits of the id, from bits 32 to 51, inclusive. The application order value starts at 1. The maximum supported number of transactions per operation is 1,048,575.
>
> Operation Index: the index of the operation within its parent transaction. This takes up 12 bits of the id, from bits 52 to 63, inclusive. The operation index value starts at 1. The maximum supported number of operations per transaction is 4095.

Thus, I think that the code is correct.

However, I would suggest bookeeping an `toid.ID` instead of maintaining the three counters separately.

exp/lighthorizon/archive/ingest_archive.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	tx.Index = ingestLedgerTransaction.Index
+	tx.Envelope = ingestLedgerTransaction.Envelope
+	tx.Result = ingestLedgerTransaction.Result
+	tx.FeeChanges = ingestLedgerTransaction.FeeChanges
+	tx.UnsafeMeta = ingestLedgerTransaction.UnsafeMeta
+
+	return tx, nil
+}
```
Shaptic, line 46: Sorry, this might be a misunderstanding on my part, but what is the benefit of wrapping ingestion into a similar structure like this for light horizon? I can't see what it offers that the `ingest` package doesn't already (which is meant for public consumption as-is). Basically - why this over what was present before?

exp/lighthorizon/index/cmd/batch/reduce/main.go
```diff
@@ -240,7 +240,7 @@ func mergeAllIndices(finalIndexStore index.Store, config *ReduceConfig) error {
 					}
 				}
 
-				logger.Infof("Final account flush.")
+				jobLogger.Infof("Final account flush.")
```
Shaptic, line 243: thanks for fixing all of the warnings :pray:

exp/lighthorizon/archive/ingest_archive.go
sreuland: the objective was to see about LightHorizon(LH) programing to its own domain model, its notion of `LedgerTransaction` could diverge from `ingest` or vice versa at any point perhaps or no? LH has defined the new `Archive` interface which refers to TX's so the model ends up being referenced all around in LH. This allows`ingest`'s version of model to be used, but encapsulated to here, the rest of LH does not know `ingest` exists, is that worthwhile as a building block? If the abstraction just adds more complexity than it's worth, I'll yank this out np, lmk, thanks!

Shaptic: thought about this some more, this is a good question... 

>  its notion of `LedgerTransaction` could diverge from `ingest` or vice versa at any point perhaps or no?

I think it's a safe bet that we'll need to do this - @paulbellamy already mentioned how we should add versioning to the meta files, which would be data on top of the `xdr.LedgerCloseMeta` -> `ingest.LedgerTransaction` ingestion layer. But the changes will be purely additive.

I'm not psyched about the naming ("adaption" and "adaptation") but that's not a blocker by any means hah; we can solve that later. (It's one of the [two hard things](https://martinfowler.com/bliki/TwoHardThings.html) in CS, anyway.)

Shaptic (approved):
:shipit:

PR #4435, opened 2022-06-15
exp/lighthorizon: Merge on-disk indices with in-memory ones on load.

Shaptic:
### What
Merges newly-loaded on-disk indices with in-memory indices.

This also adds a test for single-process index building.

### Why
You can lose in-memory changes to unrelated indices when loading a non-existing one from disk.

### Known limitations
n/a


exp/lighthorizon/index/cmd/single/main_test.go
```diff
@@ -0,0 +1,200 @@
...
+	"github.com/stellar/go/network"
+	"github.com/stellar/go/support/log"
+	"github.com/stellar/go/toid"
+	"github.com/stretchr/testify/require"
+
+	"github.com/stellar/go/exp/lighthorizon/index"
+)
+
+const (
+	txmetaSource = "s3://horizon-indices-testnet/txmeta/"
```
sreuland, line 24: not suggesting change, just wondering, would it have been possible to store the files for the 100 ledgers obtained from s3 locally under test folder instead, to avoid dependency on external service, i.e. more unit-like, rather than integration.

exp/lighthorizon/index/store.go
```diff
@@ -205,7 +205,26 @@ func (s *store) getCreateIndex(account, id string) (*types.CheckpointIndex, erro
 		ind = &types.CheckpointIndex{}
 		accountIndexes[id] = ind
 	}
-	s.indexes[account] = accountIndexes
+
+	if memoryAccountIndices := s.indexes[account]; memoryAccountIndices != nil {
```
sreuland, line 209: any concerns on this affecting performance since doing this while still in the same mutex, would it help to kick this operation off on a separate async go routine?

exp/lighthorizon/index/builder.go
```diff
@@ -109,12 +116,12 @@ func BuildIndices(
...
 				nprocessed := atomic.AddUint64(&processed, uint64(count))
 
-				log.Debugf("Working on checkpoint range [%d, %d]",
+				L.Debugf("Working on checkpoint range [%d, %d]",
 					ledgerRange.Low, ledgerRange.High)
 
 				// Assertion for testing
 				if ledgerRange.High != endLedger && (ledgerRange.High+1)%64 != 0 {
-					log.Fatalf("Upper ledger isn't a checkpoint: %v", ledgerRange)
+					L.Fatalf("Upper ledger isn't a checkpoint: %v", ledgerRange)
```
bartekn, line 124: Maybe it should be a returned error. `Fatalf` calls `panic` so it will kill all the remaining workers during their job.

bartekn (approved):

exp/lighthorizon/index/cmd/single/main_test.go
Shaptic: yeah :100: agreed, I did this locally before vacation to validate but hadn't pushed yet

exp/lighthorizon/index/builder.go
Shaptic: Oh this is actually an assertion I used for testing a previous PR - it should never happen because the "work-creation" code will always break it up into checkpoint chunks. So this code can actually be removed.

exp/lighthorizon/index/store.go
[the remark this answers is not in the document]
Shaptic: Not exactly - the `if` above is:

```golang
ind, ok = accountIndexes[id]
	if !ok {
		// Not found anywhere, make a new one.
		ind = &types.CheckpointIndex{}
		accountIndexes[id] = ind
	}
```

but `accountIndexes[id]` is finding a _specific named index_ within the map for the account, while `memoryAccountIndices := s.indexes[account]` is finding _all_ indices for the account. Because of the above `found, err := s.backend.Read(account)`, if the _specific_ index was refreshed, this will affect _all_ indices in memory. So it's separate checks: "is there an index w/ this name for this account?" and "do we already have indices in memory?"

exp/lighthorizon/index/cmd/single/main_test.go
Shaptic: done!

exp/lighthorizon/index/store.go
Shaptic: great insight! this led me to realize that we can avoid the loop entirely and just merge the specific named index: https://github.com/stellar/go/pull/4435/commits/e23e460cccbc638664fa64043cf46e42ff210492. this will have much less performance impact than before. not sure if it's still a bottleneck, but don't want to prematurely optimize now until/if we see that it's a problem!

PR #4440, opened 2022-06-24
exp/lighthorizon: Add test for batch index building map job

Shaptic:
### What
Besides introducing the end-to-end test case (`map.sh` and `map_test.go`), this makes two changes:

 - it enables controlling the worker count in a map job via the `WORKER_COUNT` envvar
 - it introduces `Store.ClearMemory(bool)` to control whether or not indices should be cleared from memory after they're flushed to disk

### Why
More tests = more stability.

Both of the changes are important for testing.

The former change lets us ensure that map jobs work with multiple workers.

The latter actually fixes a bug: when using the "accounts_unbacked" module, the indices are never read from disk to save bandwidth and time. This means that doing a `Flush()` will cause things in memory to be lost when they shouldn't be. Thus, we should disable clearing memory when using this module.

### Known limitations
n/a


exp/lighthorizon/index/builder.go
```diff
@@ -99,18 +95,14 @@ func BuildIndices(
...
-		ledger := startLedger
-		nextLedger := min(endLedger, ledger+(nextCheckpoint-startLedger))
-		for ledger <= endLedger {
-			chunk := historyarchive.Range{Low: ledger, High: nextLedger}
-			L.Debugf("Submitted [%d, %d] for work", chunk.Low, chunk.High)
-			ch <- chunk
+		checkpoints := historyarchive.NewCheckpointManager(0)
+		for ledger := range ledgerRange.GenerateCheckpoints(checkpoints) {
+			chunk := checkpoints.GetCheckpointRange(ledger)
+			chunk.High = min(chunk.High, ledgerRange.High)
```
sreuland, line 101: super minor, maybe the high/low logic could be encapsulated in `GetCheckpointRange(ledger, ledgerRange)` rather than here

exp/lighthorizon/index/cmd/batch/map/main.go
```diff
@@ -24,27 +25,12 @@ const (
...
-		return nil, err
-	}
-
-	return NewBatchConfig()
-}
-
 func NewBatchConfig() (*BatchConfig, error) {
-	targetUrl := os.Getenv(indexTargetUrlEnv)
-	if targetUrl == "" {
+	indexTargetRootUrl := os.Getenv(indexTargetUrlEnv)
```
sreuland, line 32: is there a reason why separate env and BatchConfig struct for config, could the env variables be consolidated into the BatchConfig, i.e. it's the caller's job, decouple this layer from anything outside of BatchConfig

exp/lighthorizon/index/cmd/map.sh
```diff
@@ -0,0 +1,95 @@
+#!/bin/bash
```
sreuland, line 1: was sh script used for expediency, to capture the functionality up front, could be refactored to potentially more o/s agnostic go routine later?

exp/lighthorizon/index/cmd/map_test.go
```diff
@@ -0,0 +1,170 @@
...
+
+	// Make sure everything got marked as expected in at least one index.
+	for _, item := range expected {
+		require.Containsf(t, foundCheckpoints, item,
+			"failed to find %d for %s (found %v)",
+			int(item), account, foundCheckpoints)
+	}
+}
+
+func keys[T any](dict map[string]T) []string {
```
sreuland, line 164: nice, go generics, horizon.yml includes go 1.17 in matrix, is that why GHA is complaining on this

exp/lighthorizon/index/cmd/single_test.go
```diff
@@ -0,0 +1,276 @@
...
+/**
+ * There are three parts to testing this correctly:
+ *  - test that single-process indexing works
+ *  - test that single-process w/ multi-worker works
+ *  - test map-reduce against the single-process results
+ *
+ * Therefore, if any of these fail, the subsequent ones are unreliable.
+ */
+
+func TestSingleProcess(tt *testing.T) {
```
sreuland, line 36: great coverage!

sreuland (approved):
great test coverage.

exp/lighthorizon/index/cmd/map.sh
Shaptic: Mostly expediency, but also I wasn't sure how else to build the binary and launch a bunch of parallel processes. I considered a script e.g. Python but thought that'd actually make it heavier rather than lighter since you'd need a Python container to run it in CI.

exp/lighthorizon/index/cmd/map_test.go
Shaptic: yeah probably, forgot to drop this until we drop 1.17 support :roll_eyes:

exp/lighthorizon/index/builder.go
Shaptic: that's a good point, but that function is part of the `historyarchive` package so I don't want to change its internals! there might be side effects to other parts of the code that use it, too

exp/lighthorizon/index/cmd/single_test.go
```diff
@@ -0,0 +1,276 @@
...
+// CreateBaselineIndices will connect to a dump of ledger txmeta for the given
+// ledger range and build two maps from scratch (i.e. without using the indexer)
+// by ingesting them manually:
+//
+//  - a map of tx hashes to TOIDs
+//  - a map of accounts to a list of checkpoints they were active in
+//
+// These should be used as a baseline comparison of the indexer, ensuring that
+// all of the data is identical.
+func CreateBaselineIndices(
```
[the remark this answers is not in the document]
Shaptic: good call, renamed to `IndexLedgerRange`

PR #4442, opened 2022-06-30
exp/lighthorizon/index: More testing for batch indexing and off-by-one bugfix.

Shaptic:
### What
This adds a test, `TestReduce`, that ensures that running parallel reduce jobs on previously-map'd jobs results in a complete and valid set of indices. It also extends the existing `TestMap` test to ensure that TOIDs are indexed properly.

This test uncovered an off-by-one bug that segues into an overflow bug, summarized by this (pseudo)diff:

```diff
-	for i := byte(0x00); i < 0xff; i++ {
+	for ii := int(0x00); ii <= 0xff; ii++ {
+		i := byte(ii) // can't loop over byte range bc overflow
```

### Why
Testing is important and bugs are bad.

### Known limitations
This introduces another Bash script to compile & run binaries in parallel, which means dependence on *nix-based systems.

### DIFF
--- a/exp/lighthorizon/index/cmd/batch/reduce/main.go
+++ b/exp/lighthorizon/index/cmd/batch/reduce/main.go
@@ -253,19 +253,20 @@ func mergeAllIndices(finalIndexStore index.Store, config *ReduceConfig) error {
 					WithField("indexed", transactionsProcessed).
 					WithField("skipped", transactionsSkipped)
 
-				for i := byte(0x00); i < 0xff; i++ {
+				for i := int(0x00); i <= 0xff; i++ {
+					b := byte(i) // can't loop over range bc overflow
 					if i%97 == 0 {
 						logger.Infof("%d transactions processed (%d skipped)",
 							transactionsProcessed, transactionsSkipped)
 					}
 
-					if !config.shouldProcessTx(i, routineIndex) {
+					if !config.shouldProcessTx(b, routineIndex) {
 						transactionsSkipped++
 						continue
 					}
 					transactionsProcessed++
 
-					prefix := hex.EncodeToString([]byte{i})
+					prefix := hex.EncodeToString([]byte{b})
 
 					for k := uint32(0); k < config.MapJobCount; k++ {
 						url := filepath.Join(config.IndexRootSource, fmt.Sprintf("job_%d", k))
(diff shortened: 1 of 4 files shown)


exp/lighthorizon/index/cmd/batch/reduce/main.go
```diff
@@ -159,66 +159,70 @@ func mergeAllIndices(finalIndexStore index.Store, config *ReduceConfig) error {
...
+					WithField("subservice", "accounts")
+				accountLog.Info("Started worker")
 
 				var accountsProcessed, accountsSkipped uint64
 				for account := range workQueues[routineIndex] {
-					logger.Infof("Account: %s", account)
+					accountLog.
+						WithField("total", len(accounts)).
+						WithField("indexed", accountsProcessed).
+						WithField("skipped", accountsSkipped)
```
bartekn, line 172: I can be wrong but I think you need to assign it back to `accountLog`. The returned logger contains the new fields but it won't update the object itself.

exp/lighthorizon/index/cmd/batch/reduce/main.go
```diff
@@ -228,71 +232,82 @@ func mergeAllIndices(finalIndexStore index.Store, config *ReduceConfig) error {
...
-					if i%97 == 0 {
-						logger.Infof("%d transactions processed (%d skipped)",
-							transactionsProcessed, transactionsSkipped)
+				txLog := jobLogger.
+					WithField("worker", routineIndex).
+					WithField("subservice", "transactions")
+
+				var prefixesProcessed, prefixesSkipped uint64
+				for i := int(0x00); i <= 0xff; i++ {
+					b := byte(i) // can't loop over range bc overflow
```
bartekn, line 261: Wow, I was staring at this for 10 minutes trying to understand what's going on here. Nice one.

bartekn (approved):

exp/lighthorizon/index/cmd/batch/reduce/main.go
Shaptic: Yeahh.. took a lot more than 10 minutes to figure out why the indices were incorrect :laughing: 

For anyone else curious:
 - if you do `for b := byte(0x00); b < 0xff; b++`, then you have an off-by-one because you don't include `0xff`
 - however, if you decide to do `for b := byte(0x00); b <= 0xff; b++` (that is, change `<` to `<=`), then you introduce an overflow error because `(0xff)++` becomes `0x00`, so the loop never terminates

exp/lighthorizon/index/cmd/batch/reduce/main.go
Shaptic: good catch!

Issue #4426, opened 2022-06-06
exp/services/ledgerexporter: Extend tool to work on ledger ranges in parallel.


Shaptic:
We need to extend the `ledgerexporter` tool to export ledgers in parallel (e.g. doable on AWS Batch-able), because single-process exporting is absolutely time-prohibitive for large (and dense) ledger ranges.

2opremio:
I started implementing a single-process parallel version using goroutines, but .... considering the current memory requirements of Captive Core I am thinking that's probably a bad idea.

Then, I thought about simply adding a new `--end-ledger` flag, which we could use to split out subranges across different processes, but that poses another challenge: There is race condition on the `/latest` path. I don't think we can get around using some sort of distributed lock.

Additionally, I am thinking whether it's OK to update the `/latest` path before all the prior ledgers are successfully exported (e.g. update `/latest` when ledger 35 is exported but ledger 34 hasn't finished yet). Depending on the answer to that we may want to choose different parallelization strategies.

2opremio:
@bartekn any thoughts about this? Using a distributed lock is going to be a bit of a pain in the ass. If we need to do so I think I would rather use Kubernetes instead of AWS Batch (were we can use things like https://github.com/werf/lockgate )

2opremio:
Actually, we could use a lock in the archive backend itself. For S3 we have [https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-lock.html](object locks)

2opremio:
Yes, it doesn't make sense to parallelize exporting online ledgers. However, how do you suggest we handle the update of the `/latest` path?

1. The network is at ledger 50000
2. The ledger archive is at, say ledger 10000, (as pointed out by the `/latest` path)
3. We want to use e.g. 4 workers (10k ledgers each), each in its own process
   * worker A: range 10001 to 11000
   * worker B: range 11001 to 12000
   * ....

How does `/latest` get updated? If we (wrongly) decided that each process updates `/latest` on its own, we have a race condition (e.g. if A finishes the processing last, `/latest` will end up with value 11000, which is incorrect).


Maybe we could cut corners and keep it simple e.g:

* we disable updating the `/latest` endpoint through a flag (which we supply to the exporter workers when running in parallel)
* we allow allow having gaps in the archive (so that we don't have to worry about only updating `latest` when all the prior ledgers have been exported)

thoughts?

Shaptic:
[...]

PR #4443, opened 2022-06-30
Horizon: Dockerize ledgerexport to run in AWS Batch

2opremio:
This change:

1. Creates docker image (stellar/horizon-ledgerexporter) which works in a similar fashion to stellar/horizon-verify-rage
   and is tested and pushed as part of the Horizon GitHub workflow.
2. Refactors  the way in which ledger ranges are prepared and stops using the target's /latest value implicitly (see --continue flag below). 
3. Adds three more parameters to ledgerexporter
   * --end-ledger: which indicates at what ledger to stop the export
   * --continue: which indicates the export should continue from the last previously exported ledger (i.e. /latest path in the target)
   * --write-latest-path: which indicates whether to update the /latest path of the target

Latest path writing is disabled in the container by default in order to avoid race-conditions between parallel jobs

Closes https://github.com/stellar/go/issues/4426


services/horizon/docker/ledgerexporter/README.md
```diff
@@ -0,0 +1,44 @@
+# `stellar/expingest-verify-range`
```
Shaptic, line 1: reminder to fix this

.github/workflows/horizon.yml
```diff
@@ -119,3 +120,32 @@ jobs:
...
+  ledgerexporter:
+    name: Test (and push) ledgerexporter image
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          # For pull requests, build and test the PR head not a merge of the PR with the destination.
+          ref:  ${{ github.event.pull_request.head.sha || github.ref }}
+
+      - name: Build and test the Verify Range Docker image
```
Shaptic, line 134:
```suggestion
      - name: Build and test the Ledger Exporter Docker image
```

services/horizon/docker/ledgerexporter/start
```diff
@@ -0,0 +1,49 @@
+#! /usr/bin/env bash
+set -e
+
+FROM="${FROM:=2}"
+TO="${TO:=4294967295}"
```
Shaptic, line 5: Are we sure about these defaults :sweat_smile:

services/horizon/docker/ledgerexporter/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+RUN apt-get clean
+
+ADD captive-core-pubnet.cfg /
+
+ADD start /
+RUN ["chmod", "+x", "start"]
+
+COPY --from=builder /stellar-go/ledgerexporter ./
+
+ENTRYPOINT ["/start"]
```
Shaptic, line 37: Do you need to link a volume to export to?

exp/services/ledgerexporter/main.go
```diff
@@ -37,6 +40,11 @@ func main() {
 	if *captiveCoreTomlPath == "" {
 		logger.Fatal("Missing -captive-core-toml-path flag")
 	}
+	startLedger := uint32(*startingLedger)
+	endLedger := uint32(*endingLedger)
+	if endLedger < startLedger {
```
Shaptic, line 45: What about unbounded exports?

services/horizon/docker/ledgerexporter/Dockerfile
2opremio: What do you mean? In case you export to files?

exp/services/ledgerexporter/main.go
2opremio: I reworked the ranges, PTAL

services/horizon/docker/ledgerexporter/start
2opremio: I reworked them, thanks

services/horizon/docker/ledgerexporter/README.md
2opremio: I fixed it here, thanks

services/horizon/docker/ledgerexporter/Dockerfile
sreuland: maybe meant external volume mounts, the default ephemeral storage space allocated to root f/s in docker vm is only about 30G, saw that limit come up on verify-range in AWS batch jobs, and just put up a [PR #4444](https://github.com/stellar/go/pull/4444) on allocating some top level f/s paths and mounting external volumes to them from AWS batch job ec2 launch templates to avoid that limit.

exp/services/ledgerexporter/main.go
```diff
@@ -35,7 +39,7 @@ func main() {
 		HistoryArchiveURLs: strings.Split(*historyArchiveUrls, ","),
 	}
 	if *captiveCoreTomlPath == "" {
-		logger.Fatal("Missing -captive-core-toml-path flag")
+		logger.Fatal("Missing --captive-core-toml-path flag")
```
Shaptic, line 42: nit: this was correct - flags are prefixed with a single `-` in Go for some reason :shrug:

exp/services/ledgerexporter/main.go
```diff
@@ -62,34 +66,37 @@ func main() {
...
+	rightInterval := ""
+	if endLedger == 0 {
+		rightInterval = "∞)"
+		ledgerRange = ledgerbackend.UnboundedRange(lowerBound)
+	} else {
+		rightInterval = fmt.Sprintf("%d]", endLedger)
+		ledgerRange = ledgerbackend.BoundedRange(lowerBound, endLedger)
 	}
 
+	logger.Infof("preparing to export [%d,%s", lowerBound, rightInterval)
```
Shaptic, line 95: I think `ledgerbackend.Range` has a `String()` method :)

exp/services/ledgerexporter/main.go
Shaptic: same goes for other `Fatal(...)` calls

exp/services/ledgerexporter/main.go
```diff
@@ -25,7 +26,10 @@ func main() {
 	networkPassphrase := flag.String("network-passphrase", network.TestNetworkPassphrase, "network passphrase")
 	historyArchiveUrls := flag.String("history-archive-urls", "https://history.stellar.org/prd/core-testnet/core_testnet_001", "comma-separated list of history archive urls to read from")
 	captiveCoreTomlPath := flag.String("captive-core-toml-path", os.Getenv("CAPTIVE_CORE_TOML_PATH"), "path to load captive core toml file from")
-	startingLedger := flag.Int("start-ledger", 0, "ledger to start export from")
+	startingLedger := flag.Uint("start-ledger", 0, "ledger to start export from")
```
Shaptic, line 29: Would 2 be a better default?

exp/services/ledgerexporter/main.go
```diff
@@ -62,34 +66,37 @@ func main() {
...
+			logger.Fatalf("--start-ledger and --continue cannot both be set")
 		}
-		ledgerRange = ledgerbackend.BoundedRange(startLedger, latestLedger)
+		startLedger = readLatestLedger(target)
+	}
+	var ledgerRange ledgerbackend.Range
+	lowerBound := uint32(2)
+	if startLedger > lowerBound {
+		lowerBound = startLedger
+	}
```
Shaptic, line 85: Is there a situation in which this conditional would **not** be true? I think it always executes

exp/services/ledgerexporter/main.go
```diff
@@ -62,34 +66,37 @@ func main() {
...
+	logger.Infof("preparing to export [%d,%s", lowerBound, rightInterval)
 	err = core.PrepareRange(context.Background(), ledgerRange)
-	logFatalIf(err, "could not prepare unbounded range %v", latestLedger)
-
-	logger.Infof("Unpacking ledger range [%d, %d]", *startingLedger, latestLedger)
+	logFatalIf(err, "could not prepare range")
 
-	nextLedger := latestLedger
-	for {
+	for nextLedger := lowerBound; nextLedger < endLedger; {
```
Shaptic, line 99: intentionally not using an inclusive (`<=`) range here?

services/horizon/docker/ledgerexporter/Dockerfile
```diff
@@ -0,0 +1,37 @@
+FROM golang:1.18 AS builder
+
+ARG GIT_COMMIT
+# TODO: update when merged into stellar/go
+WORKDIR /
+RUN git clone https://github.com/2opremio/go-2.git stellar-go
```
Shaptic, line 6: reminder to fix this todo!

services/horizon/docker/ledgerexporter/README.md
```diff
@@ -0,0 +1,42 @@
...
+| 1                           | 127    | 191  |
+| 2                           | 191    | 255  |
+| 3                           | 255    | 319  |
+
+## Tips when using AWS Batch
+
+* In "Job definition" set vCPUs to 2 and Memory to 4096. This represents the `c5.large` instances Horizon should be using.
+* In "Compute environments":
+    * Set instance type to "c5.large".
+    * Set "Maximum vCPUs" to 2x the number of instances you want to start (because "c5.large" has 2 vCPUs). Ex. 10 vCPUs = 5 x "c5.large" instances.
```
Shaptic, line 38: Are these HW specs accurate or copied from verify-range? Just sanity checking.

services/horizon/docker/ledgerexporter/README.md
```diff
@@ -0,0 +1,42 @@
...
+* Use spot instances! It's much cheaper and speed of testing will be the same in 99% of cases.
+* You need to publish the image if there are any changes in `Dockerfile` or one of the scripts.
+* When batch processing is over check if instances have been terminated. Sometimes AWS doesn't terminate them.
+* Make sure the job timeout is set to a larger value if you verify larger ranges. Default is just 100 seconds.
```
Shaptic, line 42:
```suggestion
* Make sure the job timeout is set to a larger value if you export larger ranges. Default is just 100 seconds.
```

and the default might need to be larger :thinking:

services/horizon/docker/ledgerexporter/start
```diff
@@ -0,0 +1,50 @@
...
+	#
+	# Ex: BATCH_START_LEDGER=63, BATCH_SIZE=64 will create the following ranges:
+	# AWS_BATCH_JOB_ARRAY_INDEX=0: [63, 127]
+	# AWS_BATCH_JOB_ARRAY_INDEX=1: [127, 191]
+	# AWS_BATCH_JOB_ARRAY_INDEX=2: [191, 255]
+	# AWS_BATCH_JOB_ARRAY_INDEX=3: [255, 319]
+	# ...
+
+	if [ $BATCH_START_LEDGER -eq 1 ]; then
+		export START=`expr "$BATCH_SIZE" \* "$AWS_BATCH_JOB_ARRAY_INDEX" + "$BATCH_START_LEDGER" - "2"`
```
Shaptic, line 31: I haven't seen this syntax before :thinking: could we use arithmetic expressions? e.g.

```suggestion
		export START=$(($BATCH_SIZE * $AWS_BATCH_JOB_ARRAY_INDEX + $BATCH_START_LEDGER - 2))
```

services/horizon/docker/ledgerexporter/start
```diff
@@ -0,0 +1,50 @@
...
+		export END=`expr "$BATCH_SIZE" \* "$AWS_BATCH_JOB_ARRAY_INDEX" + "$BATCH_START_LEDGER" + "$BATCH_SIZE"`
+	fi
+fi
+
+echo "START: $START END: $END"
+
+export TRACY_NO_INVARIANT_CHECK=1
+/ledgerexporter --target $ARCHIVE_TARGET \
+                --captive-core-toml-path /captive-core-pubnet.cfg \
+                --history-archive-urls 'https://history.stellar.org/prd/core-live/core_live_001' --network-passphrase 'Public Global Stellar Network ; September 2015' \
```
Shaptic, line 47: Do we need another version for testnet?

exp/services/ledgerexporter/main.go
2opremio: whenever startledger is > 2

exp/services/ledgerexporter/main.go
2opremio: good point, I will fix that

services/horizon/docker/ledgerexporter/Dockerfile
2opremio: I thought that I could only remove it when merged, but since we pull the pr branches I think it should be fine. I will try

services/horizon/docker/ledgerexporter/README.md
2opremio: Copied from verify-range

services/horizon/docker/ledgerexporter/README.md
2opremio: Any specific suggestions?

services/horizon/docker/ledgerexporter/start
2opremio: I just copied it from verify-range, I think it's fine

services/horizon/docker/ledgerexporter/start
2opremio: we could make it configurable, do you have a need for testnet right away? otherwise let's change it later

services/horizon/docker/ledgerexporter/start
```diff
@@ -0,0 +1,50 @@
...
+	else
+		export START=`expr "$BATCH_SIZE" \* "$AWS_BATCH_JOB_ARRAY_INDEX" + "$BATCH_START_LEDGER"`
+		export END=`expr "$BATCH_SIZE" \* "$AWS_BATCH_JOB_ARRAY_INDEX" + "$BATCH_START_LEDGER" + "$BATCH_SIZE"`
+	fi
+fi
+
+echo "START: $START END: $END"
+
+export TRACY_NO_INVARIANT_CHECK=1
+/ledgerexporter --target $ARCHIVE_TARGET \
```
sreuland, line 45: does ledgerexporter invoke captive core `catchup` or `run` with `--in-memory`, probably want to ensure it's not, so it will default to using on-disk sqlite3 db instead, that came up as a [needed change in the verify-range PR](https://github.com/stellar/go/pull/4444/files#diff-965bedcd3ef6d7f1cae4e0fa42ba6cf5057d4b059af1ec1efbe01f2c2024ab13R74), as captive core was running out of memory in the aws batch jobs when horizon ingest didn't have captive-core-use-db=true, also noticed problems with running out of ephemeral disk space on root volume in the docker vm which is limited to like 30G, now that captive core's stellar.db is on disk and likely grows over 30GB after catchup, in which case allocating some f/s paths in the vm to hold stellar.db and cc working directory, can be mounted to larger external volumes.

services/horizon/docker/ledgerexporter/start
Shaptic: seems good in that case :+1:

exp/services/ledgerexporter/main.go
Shaptic: but the lowest possible start ledger is 2 anyway, so this block is the same as just saying `lowerBound = startLedger`, right?

Shaptic (approved):
LGTM! Would just check on what Shawn said above.

services/horizon/docker/ledgerexporter/start
Shaptic: good point, no immediate need

exp/services/ledgerexporter/main.go
2opremio: I don't think it does, buut ... I could add this there.

services/horizon/docker/ledgerexporter/README.md
2opremio: but ... 100 seconds is larger that 1 minute

PR #4444, opened 2022-06-30
Trying cc use disk db on verify-range

sreuland:
testing branch if cc use-db will fix verify-range ci job

### DIFF
--- a/services/horizon/docker/verify-range/Dockerfile
+++ b/services/horizon/docker/verify-range/Dockerfile
@@ -12,6 +12,8 @@ RUN ["chmod", "+x", "dependencies"]
 RUN /dependencies
 
 ADD stellar-core.cfg /
+ADD captive-core-pubnet.cfg /
+RUN mkdir -p /cc
 
 ADD start /
 RUN ["chmod", "+x", "start"]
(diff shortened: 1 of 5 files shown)


services/horizon/docker/verify-range/Dockerfile
```diff
@@ -12,6 +12,8 @@ RUN ["chmod", "+x", "dependencies"]
 RUN /dependencies
 
 ADD stellar-core.cfg /
```
Shaptic, line 14: can drop this now, right?

sreuland: atm, just testing out changes to use 'on disk' cc mode on CI, to see if verify-range gets resolved, but it doesn't seem to be affecting it, it's wedged on these frame header errors.

from master or any branch on go repo, it seems to be easy to replicate it locally from top folder:
```
$ docker build -f services/horizon/docker/verify-range/Dockerfile -t stellar/horizon-verify-range services/horizon/docker/verify-range/
$ docker run -e FROM=10000063 -e TO=10000127 stellar/horizon-verify-range
```

Issue #4317, opened 2022-04-04
Horizon Lite - MVP Epic


jcx120:
# Epic: Ingestion Lite 
Also known as "Horizon Lite", "On-the-Fly Ingestion", and "The Constant DB solution".

**Working Branch**: https://github.com/stellar/go/pull/4352, [`stellar/go#lighthorizon`](https://github.com/stellar/go/tree/lighthorizon).

-------------------

## Acceptance Criteria
The MVP for this feature is an end-to-end deployment of Horizon Lite for a key endpoint:

  - [x] `GET /accounts/:id/transactions`

This is considered a high-impact historical endpoint that accounts for ~30% of all historical requests, based on a previous [request analysis](https://docs.google.com/document/d/10tXpCSsZFptmJ0T8F-DRDAx83TWLE-7vdYcNjA1sTb4/edit#heading=h.nc29x2g0f2dm).

-------------------

## Active Task Breakdown

The current Horizon Lite deployment is on Kubernetes:

    https://lighthorizon-pubnet.prototypes.kube001.services.stellar-ops.com/

There is a bit of important feature work remaining (besides the backlog), but the primary focus is two parallel threads:

 - we need to have at least *some* per-ledger indices up to see how they compare
 - we need to experiment with what we have in order to come up with numbers

### Remaining Feature Work

- [x] #4468

### Deployment & Experimentation

- [x] #4567
- [x] #4526
- [x] #4514
- [x] #4493

When we're ready to "get real" on SDF's infrastructure: 
- [x] stellar/ops#1818

-------------------
**Next stage of Epic:** #4571

Issue #4396, opened 2022-05-23
horizon: `history_claimable_balances` is not cleared out by the reaper.


2opremio:
We should:

1. Make sure the reaper takes care of `history_claimable_balances`
2. Make sure all the other history tables are included.
3. Add a test that ensure all the `history_` tables in the DB are cleared out (in order prevent this problem from happening to tables added in the future).

2opremio:
maybe @sydneynotthecity may have a better querying suggestion?

sydneynotthecity:
Do you know how expensive of a query cost is too expensive? 

Another option is that we could [create a trigger](https://stackoverflow.com/questions/13785855/how-to-log-data-change-in-postgresql) that tracked any record deletions in the `history_*` tables and logged them to an events logging table. We could set it up in a way where the deleted id was logged to the events table. Then, we could delete the orphaned entries based on a join to this events table, which should be significantly cheaper than joining on the original tables. 

Once the records were cleared out, we could either update an indicator in the event tables for `is_deleted = TRUE` or just wipe the records from the events table so that the table only retains orphaned records to be deleted.

sydneynotthecity:
@2opremio the other point is the `history_transaction_claimable_balances` and `history_operation_claimable_balances` were missing indices for `claimable_balance_id` which is what was causing joins/searches on `claimable_balance_id` to be super slow. @sreuland added the indices as a test in #4455 and it looks like it has greatly improved query performance. 

It appears that the equivalent `liquidity_pools` tables are also missing indices on `liquidity_pool_id` only. My guess is that if we added those as well, we should see enough query performance improvement to be able to identify orphaned records by query, like you originally proposed.

2opremio:
let's wait for the indices to be added and retake this after that.

bartekn:
https://github.com/stellar/go/pull/4518 clears orphans from `history_claimable_balances` and `history_liquidity_pools`. The last two remaining tables (`history_accounts` and `history_assets`) require indexes in other tables like `horizon_operation_participants`. @ire-and-curses agreed that new indexes should be add in one of the future releases after further tests.

PR #4450, opened 2022-07-06
lighthorizon: Prepend version to ledger files

2opremio:
Make ledger files versionable by prepending a version header

Part of https://github.com/stellar/go/issues/4317


exp/services/ledgerexporter/main.go
```diff
@@ -139,10 +139,10 @@ func readLatestLedger(backend historyarchive.ArchiveBackend) uint32 {
 func writeLedger(backend historyarchive.ArchiveBackend, ledger xdr.LedgerCloseMeta) error {
 	blob, err := ledger.MarshalBinary()
 	logFatalIf(err, "could not serialize ledger %v", ledger.LedgerSequence())
-
+	var versionHeader = [...]byte{ledgerbackend.CurrentLedgerFileVersion}
 	return backend.PutFile(
 		"ledgers/"+strconv.FormatUint(uint64(ledger.LedgerSequence()), 10),
-		io.NopCloser(bytes.NewReader(blob)),
+		io.NopCloser(io.MultiReader(bytes.NewReader(versionHeader[:]), bytes.NewReader(blob))),
```
Shaptic, line 145: This is brittle if we decide to later expand what's stored on disk alongside unpacked txmeta. We should probably define our own XDR and do proper serialization, e.g. in the `./index/xdr/LightHorizon-types.x` file adding

```xdr
struct TransactionMeta {
    int version;
    LedgerCloseMeta meta; // from the root xdr/ folder
}
```

or something.

I can take this work on as it's a little more substantial if you're busy with #4396 now

2opremio: Adding a data structure (which won't be used by the code) won't make (de)serialization more robust.

Put into other words, how would the struct make serialization different?

If you are referring to documentation, maybe we could improve that, yeah.

ingest/ledgerbackend/history_archive_backend.go
```diff
@@ -59,6 +62,16 @@ func (b *HistoryArchiveBackend) GetLedger(ctx context.Context, sequence uint32)
...
+	// read and check version header
+	var version [1]byte
+	l, err := r.Read(version[:])
+	if err != nil {
+		return xdr.LedgerCloseMeta{}, err
+	}
+	if l != 1 || version[0] != CurrentLedgerFileVersion {
+		return xdr.LedgerCloseMeta{}, fmt.Errorf("unexpected ledger header version number (0x%x)", version[0])
+	}
+
```
Shaptic, line 74: Wait, just realized that this is in the public `ingest` package. Won't this break Horizon? And anyone else using the `HistoryArchiveBackend` today?

I think if we want to do this, we need to have our own way to read from the 

```
latest
ledgers/
ledgers/2
ledgers/3
```

directory structure rather than hacking in `HistoryArchiveBackend` for convenience like we did in the PoC.

exp/services/ledgerexporter/main.go
Shaptic: Well right now there's an implicit "one byte then a LedgerCloseMeta struct" binary format. Yeah, we could document it, but as it expands, we'll have to write custom serialization code (here, it's still relatively simple w/ `io.MultiReader`, but it will get complex later I'm sure) instead of a single `MarshalBinary()` call on an XDR structure.

For example, what if, in version 2, we decide to have `byte || network passphrase || LedgerCloseMeta`? Well we'd probably need something more like `byte || len(network_passphrase) || network_passphrase || LedgerCloseMeta` to do that. And now we have a bunch of custom (de)serialization code.

ingest/ledgerbackend/history_archive_backend.go
sreuland: looks like bumping into limits of package re-use from existing repo(classic horizon), it was convenient for PoC, but looks like it is starting to lead into more effort/complexity to continue using for lighthorizon requirements, is this where we start to bite the bullet, decouple from dependencies to classic and copy/fork what we need under exp/lighthorizon, would equate to less refactoring pain later when we decide to cut over to new repo?

sreuland: was looking at our current lighthorizon domain model, it defines it's own model of an `archive.Archive` interface, and initially one implementation of it `archive.NewIngestArchive` which is just thin wrapper around dependency to mono-repo's `ingest` pkg, maybe could switch that to reference a local copy,`exp\lighthorizon\archive\ingest` which then includes extra version byte mods on top instead.

exp/services/ledgerexporter/main.go
2opremio: We can structure it more as/if the format changes.

ingest/ledgerbackend/history_archive_backend.go
2opremio: Wasn't `HistoryArchiveBackend` added just for horizon light? If so, I don't see a problem with adding the byte there.

Regardless ... this falls out of the context of this PR.

exp/services/ledgerexporter/main.go
paulbellamy: Would an xdr union make more sense? e.g. Preconditions. The concern with this issue was really just making sure we don't get stuck if we need to change/append the format later. If there's a way to upgrade it to a union later without getting stuck, let's do that and ignore this for now, otherwise I'd say we should use xdr. In other places in lighthorizon we moved away from custom serialization in favour of xdr, so that all our file read/writes are more "standard".

ingest/ledgerbackend/history_archive_backend.go
Shaptic: Oh that's a good point

2opremio:
@sreuland @Shaptic @paulbellamy PTAL

paulbellamy (approved):
lgtm. with a nit on xdr organization

Makefile
```diff
@@ -6,7 +6,8 @@ xdr/Stellar-ledger-entries.x \
 xdr/Stellar-ledger.x \
 xdr/Stellar-overlay.x \
 xdr/Stellar-transaction.x \
-xdr/Stellar-types.x
+xdr/Stellar-types.x \
+xdr/Stellar-lighthorizon.x
```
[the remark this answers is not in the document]
2opremio: I don't think _(b)_ can be easily done since `SerializedLedgerCloseMeta` depends on the `xdr`  package (`LedgerCloseMeta`) but .... I can try _(a)_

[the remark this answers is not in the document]
Shaptic: Yeah (a) makes sense, just means we need to keep XDR in sync/copied if we move to another repo.

Shaptic (approved):
XDR union was a great call, LGTM :+1:

PR #4453, opened 2022-07-09
/exp/lighthorizon: new endpoints for tx and ops paged listing by account id

### LINKED ISSUE #4433: exp/lighthorizon: add endpoints for paged list of entities(tx/op) by account ID
getting a paged list of tx's and op's that belong to an account id

two new action/endpoints
`GET /accounts/:id/transactions?cursor=&limit=`  
`GET /accounts/:id/operations?cursor=&limit=`

leverages indexes for looking up all tx's related to the account id.

sreuland:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

added new endpoints:
/accounts/{account_id}/transactions?cursor=&limit=
/accounts/{account_id}/operations?cursor=&limit=

### Why

these endpoints leverage the indexes for fast lookup of tx and ops related to an account id.
closes #4433


exp/lighthorizon/services/main.go
```diff
@@ -0,0 +1,359 @@
...
+			for {
+				tx, readErr := reader.Read()
+				if readErr != nil {
+					if readErr == io.EOF {
+						break
+					}
+					return nil, readErr
+				}
+
+				transactionOrder++
```
Shaptic, line 164: You can use `tx.Index` instead of tracking `transactionOrder` separately

exp/lighthorizon/services/main.go
```diff
@@ -0,0 +1,359 @@
...
+					txs = append(txs, common.Transaction{
+						TransactionEnvelope: &tx.Envelope,
+						TransactionResult:   &tx.Result.Result,
+						LedgerHeader:        &ledger.V0.LedgerHeader.Header,
+						TxIndex:             int32(transactionOrder),
+					})
+					if int64(len(txs)) == limit {
+						return txs, nil
+					}
+				}
```
Shaptic, line 180: Do we need to return a set here? Since we only care about a single value, it's just as "expensive" to loop through the slice of participants and match against `accountId`, isn't it?

exp/lighthorizon/services/main.go
```diff
@@ -0,0 +1,359 @@
...
+
+			if appending {
+				txns = append(txns, common.Transaction{
+					TransactionEnvelope: &tx.Envelope,
+					TransactionResult:   &tx.Result.Result,
+					// TODO: Use a method to get the header
+					LedgerHeader: &ledger.V0.LedgerHeader.Header,
+					TxIndex:      int32(transactionOrder),
+				})
+			}
```
Shaptic, line 321: Same thing here re: `tx.Index`. Something like this would be more reliable:

```suggestion
			newTx := common.Transaction{
				TransactionEnvelope: &tx.Envelope,
				TransactionResult:   &tx.Result.Result,
				// TODO: Use a method to get the header
				LedgerHeader: &ledger.V0.LedgerHeader.Header,
				TxIndex:      tx.Index,
			}

			if txToid := newTx.TOID(); txToid >= cursor {
				appending = true
				if txToid == cursor {
					continue
				}
			}
			
			if appending {
				txns = append(txns, newTx)
			}
```

Shaptic (commented):
some initial thoughts, mostly LGTM though :+1: you can probably drop the /operations stuff since that shouldn't be part of the MVP, maybe stash it elsewhere and we can come back to it? To keep the code lighter

exp/lighthorizon/services/main.go
```diff
@@ -0,0 +1,359 @@
...
+	GetOperationsByAccount(ctx context.Context, cursor int64, limit int64, accountId string) ([]common.Operation, error)
+	GetOperations(ctx context.Context, cursor int64, limit int64) ([]common.Operation, error)
+}
+
+type TransactionRepository interface {
+	GetTransactionsByAccount(ctx context.Context, cursor int64, limit int64, accountId string) ([]common.Transaction, error)
+	GetTransactions(ctx context.Context, cursor int64, limit int64) ([]common.Transaction, error)
+}
+
+func (os *OperationsService) GetOperationsByAccount(ctx context.Context, cursor int64, limit int64, accountId string) ([]common.Operation, error) {
```
Shaptic, line 46: It seems to me that all of these handlers follow a very similar pattern:

- skip to the starting cursor
- read a ledger
- process it
- skip to the next cursor

The only difference from handler to handler is how to skip ahead (which depends on the handler's index) and how to process it (but even here, it's some loop over the transactions in a ledger).

As we keep expanding the endpoints, if we follow the current pattern here, it seems like we're going to have more and more code duplication.

Maybe we need to consolidate the common points of a handler so that each part (e.g. transactions or operations) can focus on just its unique parts. Something like,

```go
type Endpoint interface {
	GetArchive() Archive
	GetNextCursor(current int64) int64
	ProcessLedger(ledger xdr.SerializedLedgerCloseMeta) error
	ProcessTransaction(tx ingest.LedgerTransaction) error
}

type TransactionHandler struct {
	RequestParams 

	archive Archive
	store index.Store
	accountId string // optional

	results []common.Transaction
}

func (handler *TransactionHandler) GetNextCursor(cursor int64) int64 {
	if len(handler.results) >= handler.RequestParams.limit {
		return 0 // signifies that we should stop
	}
	
	// excuse the pseudocode
	ledger, txOrder, opOrder := toid.Parse(cursor)

	if handler.accountId != "" {
		return handler.store.NextActive(ledger / 64) * 64
	}
	return ledger + 1
}

func (handler *TransactionHandler) ProcessTransaction(tx ingest.Transaction) error {
	participants := handler.GetArchive().GetTransactionParticipants(tx)
	_, found := participants[accountId]
	if !found {
		return io.EOF
	}
	
	handler.results = append(handler.results, common.Transaction{
		TransactionEnvelope: &tx.Envelope,
		TransactionResult:   &tx.Result.Result,
		LedgerHeader:        &ledger.V0.LedgerHeader.Header,
		TxIndex:             int32(transactionOrder),
	})
	
	if int64(len(handler.results)) >= handler.RequestParams.limit {
		return io.EOF
	}
}
// etc.
```

with the unified code doing something like

```go
func EndpointProcessor(ctx context.Context, cursor, limit int64, ep Endpoint) error {
	cursor = ep.GetNextCursor(cursor)
	for cursor != 0 {
		ledgerSeq, _, _ := toid.Parse(cursor)

		// excuse more pseudocode and no error checks
		ledger = os.Archive.GetLedger(ctx, ledgerSeq)
		reader = os.Archive.NewLedgerTransactionReaderFromLedgerCloseMeta(ledger)

		for {
			if ctx.Err() != nil {
				return nil, ctx.Err()
			}

			tx, err := reader.Read()
			if err == io.EOF {
				break
			} else if err != nil {
				return err
			}

			err = ep.ProcessTransaction(tx)
			if err != nil {
				return err
			}
		}

		cursor = ep.GetNextCursor(cursor)
	}

	return nil
}
```

Shaptic: Do my thoughts make sense?

Basically, we want it to be as easy and flexible as possible to write new endpoints that have new ways of processing a transaction (in fact I bet a lot of this code will be quite similar to the existing ingestion processors code in old Horizon), but it seems error-prone and clunky to repeat the "update cursor, read ledgers, loop through transactions" boilerplate every time.

exp/lighthorizon/services/main.go
sreuland: yes, that would be more lift on the caller side here to handle, or this way encapsulated it in the data access layer, the archive method, I don't have strong pref, can adjust.

exp/lighthorizon/services/main.go
sreuland: yes, definitely, I had an urge to try some amount of patterning here to avoid the dup, but then decided on getting functionality out first, but this screams for some refactoring going forward, can address this tech debt now rather than later, your ideas sound good, I can start there.

exp/lighthorizon/main.go
```diff
@@ -41,13 +43,29 @@ func main() {
...
+			Archive:    ingestArchive,
+			Passphrase: *networkPassphrase,
+			IndexStore: indexStore,
+		},
+		Operations: services.OperationsService{
+			Archive:    ingestArchive,
+			Passphrase: *networkPassphrase,
+			IndexStore: indexStore,
+		},
+	}
```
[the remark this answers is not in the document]
sreuland: I think so, can change to pass pointer of the one `indexStore` instance to services rather than a copy of struct? I took a conservative path here first on how services get dependencies, maybe need specific config'd instances at some point, but this can be consolidated place that LightHorizon knows how to compose it, ideally the services are decoupled, not aware of this.

exp/lighthorizon/services/main.go
Shaptic: and the `.TOID()` method on `common.Transaction` to avoid tinkering w/ the off-by-ones

exp/lighthorizon/services/main.go
Shaptic: no strong pref either, just wasn't sure of the benefit of building out the set but you're right that it adds caller lift. it might introduce code dupe in the future too. lets keep it as-is then

exp/lighthorizon/main.go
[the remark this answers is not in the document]
Shaptic: `Store` is an interface so `indexStore Store` should get passed around by reference anyway? Agreed around decoupling, it's a good design principle. Let's leave it as-is for now because thinking about it more, some handlers might prioritize some indices over others :thinking: separate caches, etc. could be useful in the future

exp/lighthorizon/services/main.go
Shaptic: :+1: for functionality first, but I definitely think now is a good time to address the tech debt because it will give us a meaningful insight on how easy/hard it will be to extend the MVP to more endpoints in the future

exp/lighthorizon/main.go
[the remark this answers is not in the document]
sreuland: oh, just realized, it's already passing pointer to same instance of `indexStore` to both services, as it's an interface ref.

exp/lighthorizon/services/main.go
sreuland: yes, re-oriented to that reference, thanks.

exp/lighthorizon/services/main.go
sreuland: yes, I refactored to more functional composition here - https://github.com/stellar/go/pull/4453/commits/6de7faa6b3d5f3739ee6e638c728a0be74f2b25e

exp/lighthorizon/actions/accounts.go
```diff
@@ -0,0 +1,129 @@
...
+		sendErrorResponse(w, http.StatusBadRequest, string(invalidPagingParameters))
+		return "", pagination{}
+	}
+
+	if paginate.Cursor < 1 {
+		paginate.Cursor = toid.New(1, 1, 1).ToInt64()
+	}
+
+	if paginate.Limit < 1 {
+		paginate.Limit = 10
```
bartekn, line 41: This should probably result in `400 Bad Request`. If there's a bug in the client code it's better to return an error than change it to a correct value. At the same time we could set the value to default value if it's not set in query params.

exp/lighthorizon/actions/accounts.go
```diff
@@ -0,0 +1,129 @@
...
+	urlAccountId = "account_id"
+)
+
+func accountRequestParams(w http.ResponseWriter, r *http.Request) (string, pagination) {
+	var accountId string
+	var accountErr bool
+
+	if accountId, accountErr = getURLParam(r, urlAccountId); !accountErr {
+		sendErrorResponse(w, http.StatusBadRequest, "")
+		return "", pagination{}
```
bartekn, line 27: Nit: Maybe we could reuse existing Horizon helpers here? We could return error from `problems` package that gets automatically rendered. I understand this is still MVP but using existing code will actually allow us to write less code.

bartekn (commented):
Two comments, probably not required to be part of this PR.

exp/lighthorizon/actions/accounts.go
sreuland: I initially started copying over some of the problems infra package and some of `Make` and `New` helpers, but it ended up being a fair amount of code with deps, and wasn't sure how much we want light to carry over or be from scratch, I can comment in some tech debt for that to be considered on next iterations?

exp/lighthorizon/services/main.go
```diff
@@ -0,0 +1,192 @@
...
+				}
+
+				if _, found := participants[accountId]; found {
+					if finished, callBackErr := callback(tx, &ledger.V0.LedgerHeader.Header); callBackErr != nil {
+						return callBackErr
+					} else {
+						if finished {
+							return nil
+						}
+					}
```
Shaptic, line 149: nit: can use `else if`
```suggestion
					} else if finished {
						return nil
					}
```

exp/lighthorizon/services/main.go
```diff
@@ -0,0 +1,192 @@
...
+
+	if err := searchTxByAccount(ctx, cursor, accountId, ts.Config, txsCallback); err != nil {
+		return nil, err
+	}
+	return txs, nil
+}
+
+func searchTxByAccount(ctx context.Context, cursor int64, accountId string, config Config, callback searchCallback) error {
+	// Skip the cursor ahead to the next active checkpoint for this account
+	nextCheckpoint, err := getAccountNextCheckpointCursor(accountId, cursor, config.IndexStore)
```
Shaptic, line 104: We need to be a little careful here because this couples the fact that account indices are per-checkpoint to this code. We may need to make account indices per-**ledger** in the future if we decide the latency is too high w/ per-checkpoint, so this would need to be reworked. Ideally we could isolate the "get next ledger to read" part of the code away from this.

exp/lighthorizon/services/main.go
```diff
@@ -0,0 +1,192 @@
...
+			ledgerSequence++
+		}
+
+		nextCheckpoint, err = getAccountNextCheckpointCursor(accountId, nextCheckpoint, config.IndexStore)
+		if err != nil {
+			if err == io.EOF {
+				return nil
+			}
+			return err
+		}
```
Shaptic, line 165: nit / personal preference: one less line of code :joy: 
```suggestion
		if err == io.EOF {
			return nil
		} else if err != nil {
			return err
		}
```

exp/lighthorizon/services/main.go
```diff
@@ -0,0 +1,192 @@
...
+			if opParticipantErr != nil {
+				return false, opParticipantErr
+			}
+			if _, foundInOp := opParticipants[accountId]; foundInOp {
+				ops = append(ops, common.Operation{
+					TransactionEnvelope: &tx.Envelope,
+					TransactionResult:   &tx.Result.Result,
+					LedgerHeader:        ledgerHeader,
+					TxIndex:             int32(tx.Index),
+					OpIndex:             int32(operationOrder + 1),
```
Shaptic, line 63: I think this is 0-indexed just like transactions, because the `.TOID()` method does `+1` to both this and the `TxIndex` field.

exp/lighthorizon/services/main.go
```diff
@@ -0,0 +1,192 @@
...
+		txs = append(txs, common.Transaction{
+			TransactionEnvelope: &tx.Envelope,
+			TransactionResult:   &tx.Result.Result,
+			LedgerHeader:        ledgerHeader,
+			TxIndex:             int32(tx.Index),
+		})
+		if int64(len(txs)) == limit {
+			return true, nil
+		}
+		return false, nil
```
Shaptic, line 93: suggestion: could combine these :man_shrugging: 
```suggestion
		return (int64(len(txs)) >= limit), nil
```

exp/lighthorizon/services/main.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+
+	if err := searchTxByAccount(ctx, cursor, accountId, ts.Config, txsCallback); err != nil {
+		return nil, err
+	}
+	return txs, nil
```
Shaptic, line 99: Should we return the `txs` anyway and let the caller decide whether they care about the error value?
```suggestion
	return txs, searchTxByAccount(ctx, cursor, accountId, ts.Config, txsCallback)
```

exp/lighthorizon/services/main.go
```diff
@@ -0,0 +1,192 @@
...
+		}
+		return err
+	}
+	log.Debugf("Searching index by account %v starting at checkpoint cursor %v", accountId, nextCheckpoint)
+
+	for {
+		startingCheckPointLedger := cursorLedger(nextCheckpoint)
+		ledgerSequence := startingCheckPointLedger
+
+		for (ledgerSequence - startingCheckPointLedger) < 64 {
```
Shaptic, line 117: why not index-based loop?
```suggestion
		for ledgerSeq := startingCheckPointLedger; ledgerSeq <= startingCheckPointLedger + 64; ledgerSeq++ {
```
seems easier to follow the condition

exp/lighthorizon/services/main.go
```diff
@@ -0,0 +1,192 @@
...
+			ledger, ledgerErr := config.Archive.GetLedger(ctx, ledgerSequence)
+			if ledgerErr != nil {
+				return errors.Wrapf(ledgerErr, "ledger export state is out of sync, missing ledger %v from checkpoint %v", ledgerSequence, ledgerSequence/64)
+			}
+
+			reader, readerErr := config.Archive.NewLedgerTransactionReaderFromLedgerCloseMeta(config.Passphrase, ledger)
+			if readerErr != nil {
+				return readerErr
```
Shaptic, line 125: nit: error verbosity?
```suggestion
				return errors.Wrapf(readerErr, "failed reading ledger %d", ledgerSequence)
```

exp/lighthorizon/services/main.go
```diff
@@ -0,0 +1,192 @@
...
+	var checkpoint uint32
+	checkpoint, err := store.NextActive(accountId, "all/all", uint32(toid.Parse(cursor).LedgerSequence/64))
+	if err != nil {
+		return 0, err
+	}
+	ledger := int32(checkpoint * 64)
+	if ledger < 0 {
+		// Check we don't overflow going from uint32 -> int32
+		return 0, fmt.Errorf("overflowed ledger number")
+	}
```
Shaptic, line 179: this shouldn't happen for a loooong time, but great defensive check :+1:

exp/lighthorizon/services/main_test.go
```diff
@@ -0,0 +1,221 @@
...
+	"github.com/stellar/go/exp/lighthorizon/archive"
+	"github.com/stellar/go/exp/lighthorizon/index"
+	"github.com/stellar/go/xdr"
+	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
+)
+
+func TestItGetsTransactionsByAccount(tt *testing.T) {
+	// l=1586111, t=1, o=1
+	// cursor = 6812294872829953, checkpoint=24782
```
Shaptic, line 17: Maybe we can do end-to-end testing on the existing fixtures, too?

exp/lighthorizon/actions/accounts.go
sreuland: updated in  https://github.com/stellar/go/pull/4453/commits/91fcd88715a810a201fd5b3ff6bcc517d1c88ee3

exp/lighthorizon/services/main.go
sreuland: https://github.com/stellar/go/pull/4453/commits/91fcd88715a810a201fd5b3ff6bcc517d1c88ee3

exp/lighthorizon/services/main.go
sreuland: is that a standard convention for non-recoverable err handling, the data is in incomplete/unknown state, doesn't see worth the complexity for callers to navigate that?

Shaptic: I don't know if it's a convention necessarily (in fact, I doubt it is), I just have a personal preference of a "here's what I did before the error happened" policy for long-running processes like this. Imo if there's an error, you can still grant the caller agency about whether or not they want to deal with recovering incomplete data.

Shaptic: Like I said, though, it's personal preference, so feel free to leave it as-is :)

exp/lighthorizon/actions/accounts.go
Shaptic: We should definitely make an issue for doing proper error messages. I'm pretty sure we 500 right now for most errors (e.g. 404s) rather than returning a problem or even using the right error code.

exp/lighthorizon/archive/ingest_archive.go
```diff
@@ -25,6 +26,42 @@ func (ingestArchive) NewLedgerTransactionReaderFromLedgerCloseMeta(networkPassph
...
+	participants, err := index.GetTransactionParticipants(a.ingestTx(transaction))
+	if err != nil {
+		return nil, err
+	}
+	set := make(map[string]struct{})
+	exists := struct{}{}
+	for _, participant := range participants {
+		set[participant] = exists
+	}
+	return set, nil
```
Shaptic, line 39: You can use `SafeStringSet` if you'd like! It's thread-safe, though.

Either way we have sets all over the place... we should probably write an abstraction eventually. _makes note to self_

exp/lighthorizon/index/modules.go
```diff
@@ -105,133 +105,137 @@ func GetPaymentParticipants(transaction ingest.LedgerTransaction) ([]string, err
...
 		}
 
-		// Requires meta
-		// sponsor, err := operation.getSponsor()
-		// if err != nil {
-		// 	return nil, err
-		// }
-		// if sponsor != nil {
-		// 	otherParticipants = append(otherParticipants, *sponsor)
-		// }
```
Shaptic, line 229: Where did this go?

exp/lighthorizon/services/main.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+// this deals in ledgers but adapts to the index model, which is currently keyed by checkpoint for now
+func getAccountNextLedgerCursor(accountId string, cursor int64, store index.Store) (uint64, error) {
+	nextLedger := toid.Parse(cursor).LedgerSequence + 1
+	cursorCheckpoint := uint32(nextLedger / 64)
+	queryCheckpoint := cursorCheckpoint
+	if queryCheckpoint > 0 {
+		queryCheckpoint = queryCheckpoint - 1
+	}
```
Shaptic, line 167: Can you explain this bit of code?

exp/lighthorizon/services/main.go
```diff
@@ -0,0 +1,183 @@
...
+	nextCheckpoint, err := store.NextActive(accountId, "all/all", queryCheckpoint)
```
Shaptic, line 168: What about for `/accounts/:id/payments`? We'll need a different index.

Shaptic, line 168: Should we make this stateful-ish? We don't need to query the index every time: once you get an initial checkpoint, you can just `++` until you need to query for the next checkpoint.

Maybe the logic can be:

```
ledger = toid.Parse(cursor)
if !IsCheckpoint(ledger):
  return ledger + 1
else:
  return NextActive(ledger + 1)
```

Would that work? Might need some edge-case handling around getting the first cursor, not sure... Point is we should avoid index queries whenever possible.

exp/lighthorizon/actions/accounts.go
sreuland: on this pass, I believe we have at least captured the core semantics between 5's for server problem and 4's for client problem, which can guide into later iteration for more verbose, will TODO for pulling horizon `problems`

exp/lighthorizon/services/main.go
sreuland: that's a good point, I cleaned this up to be zero based on func interfaces and doc'd it - https://github.com/stellar/go/pull/4453/commits/5ffcb321a9996bd090d75750f73e0fbfd3983331

exp/lighthorizon/services/main.go
sreuland: it derives the checkpoint number from incoming requested cursor for the checkpoint arg value to pass to index.Next, the response will be either the same or greater checkpoint than the current requested cursor, iterating ledgers at that point from either +1 seq of requested cursor or first seq of newer checkpoint respectively, to provide the ledger based interface.

exp/lighthorizon/services/main.go
sreuland: yes, sounds like another ticket to track on that scope for another iteration?

sreuland: yes, I agree, I saw this angle coming up as I re-wrote this to make it ledger based, wanted to get that done first, let me think over your suggestion and I'll look for way to get this in now, thanks!

exp/lighthorizon/archive/main.go
```diff
@@ -36,148 +30,39 @@ type LedgerTransactionReader interface {
...
+	// returns error or map with keys of participant account id's and value of empty struct
+	GetTransactionParticipants(transaction LedgerTransaction) (map[string]struct{}, error)
+
+	// GetOperationParticipants - get set of all participants(accounts) in a operation
+	//
+	// transaction - the ledger transaction
+	// operation   - the operation within this transaction
+	// opIndex     - the 0 based index of the operation within the transaction
+	//
+	// returns error or map with keys of participant account id's and value of empty struct
```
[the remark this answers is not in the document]
sreuland: yes, I'll clean that up, that's about a decade of javadocs habit there.

exp/lighthorizon/services/main.go
Shaptic: absolutely - that endpoint should be its own ticket. I was mostly highlighting that we shouldn't hardcode the index here

Shaptic: cool, sounds good!

exp/lighthorizon/archive/main.go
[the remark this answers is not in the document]
Shaptic: definitely felt a little javaesque to me :wink:

exp/lighthorizon/services/main.go
sreuland: doh! got it - https://github.com/stellar/go/pull/4453/commits/a6b87ec67c1490863b6edaa550a53b6b8fc7526e

exp/lighthorizon/index/modules.go
sreuland: looked like dead code left in, so dropped it,  can always get it back from history right?

exp/lighthorizon/actions/accounts.go
sreuland: yes, created https://github.com/stellar/go/issues/4472

exp/lighthorizon/archive/main.go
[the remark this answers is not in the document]
sreuland: https://github.com/stellar/go/pull/4453/commits/fa45ee286b9ea878645e2ed567d0c8499502c0dc

exp/lighthorizon/services/main.go
sreuland: ok, was able to wedge in that perf optimization, thanks - https://github.com/stellar/go/pull/4453/commits/53fde551059fd3b3033a88fd4a89fc60452aa5ac

exp/lighthorizon/services/main_test.go
sreuland: are you suggesting dedicated integration tests, i.e. `/exp/lighthorizon/integration` ? 
beyond these mock/unit tests, I'm running locally paired with small index store built up from testnet using that script tool, can we spin up a ticket to track on integration tests?

exp/lighthorizon/services/main.go
```diff
@@ -0,0 +1,187 @@
...
+		return 0, err
+	}
+
+	// return the first ledger of the next checkpoint that had account activity after cursor
+	return uint64(indexNextCheckpoint * 64), nil
+}
+
+func isCheckpoint(ledger int32) bool {
+	return ledger%64 == 0
+}
```
Shaptic, line 187: Recommend using `historyarchive.CheckpointManager` for all checkpoint math!

Shaptic (approved):
awesome, I think it's ready!! :clap:

Issue #4451, opened 2022-07-06
Add metrics to Horizon Lite


Shaptic:
Epic: #4317

-----------------

As part of Phase 3, we need metrics to effectively compare it to Sunset Horizon. This issue covers a few metrics:

  * [ ] How long do indices take to build?
  * [ ] How long do indices take to update in real time?
  * [x] How old is a request, and, how does it correlate with how long servicing it takes?
  * [x] How long different parts of the on-the-fly ingestion take: 
    - [x] index lookup, 
    - [x] ledger downloads, 
    - [x] ingestion itself, etc.
  * [ ] #4528

Some of these should be exposed as Prometheus metrics on the new `/metrics` endpoint while others are part of the logs.

sreuland:
I'm working on these metrics initially:

http request - Gauge
http request duration m/s - Gauge
number of of ledgers traversed in http request - Gauge

Shaptic:
I'm closing this in favor of leaving the one remaining issue open, #4528

PR #4456, opened 2022-07-11
exp/lighthorizon: Add basic scaffolding for metrics.

Shaptic:
### What
Two major changes:
  - adding a `ConnectWithConfig` helper that enables passing a config to `index.Store`
  - adding a preliminary metric to demonstrate scaffolding: the size of the index working set

### Why
Metrics are good! Plus #4451.

I'd like to get :heavy_check_mark: on the way I've set up metrics before investing too much time into the metrics themselves.

### Known limitations
n/a

### DIFF
--- a/exp/lighthorizon/index/builder.go
+++ b/exp/lighthorizon/index/builder.go
@@ -103,8 +103,8 @@ func BuildIndices(
 		checkpoints := historyarchive.NewCheckpointManager(0)
 		for ledger := range ledgerRange.GenerateCheckpoints(checkpoints) {
 			chunk := checkpoints.GetCheckpointRange(ledger)
-			chunk.High = min(chunk.High, ledgerRange.High)
-			chunk.Low = max(chunk.Low, ledgerRange.Low)
+			chunk.High = min(chunk.High, ledgerRange.High) // don't exceed upper bound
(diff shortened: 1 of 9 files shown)


exp/lighthorizon/index/builder.go
```diff
@@ -136,19 +144,19 @@ func BuildIndices(
...
+	printProgress("Reading ledgers", processed, uint64(ledgerCount), startTime)
 
 	L.Infof("Processed %d ledgers via %d workers", processed, parallel)
 	L.Infof("Uploading indices to %s", targetUrl)
 	if err := indexStore.Flush(); err != nil {
 		return indexBuilder, errors.Wrap(err, "flushing indices failed")
 	}
 
+	// Assertion for testing
+	if processed != uint64(ledgerCount) {
```
sreuland, line 156: are these counts meaningful to the caller, i.e. rather than pass judgement in here, let the caller get them and choose severity, etc?

Shaptic: yeah as comment notes this was mostly for debugging while I tested the builder. left it in there as a warning because they should always be equal, but we can drop them later or turn it into an error/return value! shouldn't differ while we continue on the MVP, though, so still a useful log

PR #4457, opened 2022-07-14
exp/lighthorizon: Add an on-disk cache for frequently accessed ledgers.

Shaptic:
### What
Two changes:
 - Update `historyarchive.FsCacheBackend` to use an off-the-shelf package ([`hashicorp/golang-lru`](https://pkg.go.dev/github.com/hashicorp/golang-lru#Cache)) over a custom solution.
 - Use the above to wrap the ledger backend

### Why
This nearly eliminates request latency if the ledgers are cached locally.

### Known limitations
We should think about a best practice for cache eviction when we upgrade the exported meta, though maybe this just means operators delete their caches.

### DIFF
--- a/exp/lighthorizon/archive/ingest_archive.go
+++ b/exp/lighthorizon/archive/ingest_archive.go
@@ -2,6 +2,7 @@ package archive
 
 import (
 	"context"
+	"fmt"
 	"net/url"
 
 	"github.com/stellar/go/exp/lighthorizon/index"
@@ -14,25 +15,32 @@ import (
 	"github.com/stellar/go/xdr"
 )
 
-const (
-	maxLedgersToCache = (60 * 60 * 24) / 6 // 1 day of ledgers @ 6s each
-)
-
 type ArchiveConfig struct {
 	SourceUrl         string
 	NetworkPassphrase string
 	CacheDir          string
+	CacheSize         int
 }
 
 func NewIngestArchive(config ArchiveConfig) (Archive, error) {
-	// If the source URL is an S3 url and it has a region specified, we should
-	// try to extract it.
+	if config.CacheSize <= 0 {
+		return nil, fmt.Errorf("invalid cache size: %d", config.CacheSize)
+	}
+
 	parsed, err := url.Parse(config.SourceUrl)
 	if err != nil {
 		return nil, errors.Wrapf(err, "%s is not a valid URL", config.SourceUrl)
 	}
+
 	region := ""
-	if parsed.Scheme == "s3" {
+	needsCache := true
+	switch parsed.Scheme {
+	case "file":
+		// We should only avoid a cache if the ledgers are already local.
+		needsCache = false
+
+	case "s3":
+		// We need to extract the region if it's specified.
 		region = parsed.Query().Get("region")
 	}
 
@@ -50,17 +58,22 @@ func NewIngestArchive(config ArchiveConfig) (Archive, error) {
 		return nil, err
 	}
 
-	cache, err := historyarchive.MakeFsCacheBackend(source, config.CacheDir, maxLedgersToCache)
-	if err != nil { // warn but continue w/o cache
-		log.WithField("path", config.CacheDir).
-			WithError(err).
-			Warnf("Failed to create cached ledger backend")
-		cache = source
-	} else {
-		log.WithField("path", config.CacheDir).Infof("On-disk cache configured")
+	if needsCache {
+		cache, err := historyarchive.MakeFsCacheBackend(source,
+			config.CacheDir, uint(config.CacheSize))
+
+		if err != nil { // warn but continue w/o cache
+			log.WithField("path", config.CacheDir).
+				WithError(err).
+				Warnf("Failed to create cached ledger backend")
+		} else {
+			log.WithField("path", config.CacheDir).
+				Infof("On-disk cache configured")
+			source = cache
+		}
 	}
 
-	ledgerBackend := ledgerbackend.NewHistoryArchiveBackend(cache)
+	ledgerBackend := ledgerbackend.NewHistoryArchiveBackend(source)
 	return ingestArchive{ledgerBackend}, nil
 }
 
--- a/exp/lighthorizon/archive/ingest_archive.go
+++ b/exp/lighthorizon/archive/ingest_archive.go
@@ -2,13 +2,20 @@ package archive
 
 import (
 	"context"
+	"net/url"
 
 	"github.com/stellar/go/exp/lighthorizon/index"
 	"github.com/stellar/go/ingest"
 	"github.com/stellar/go/ingest/ledgerbackend"
+	"github.com/stellar/go/support/errors"
+	"github.com/stellar/go/support/log"
 
 	"github.com/stellar/go/historyarchive"
 	"github.com/stellar/go/xdr"
+)
+
+const (
+	maxLedgersToCache = (60 * 60 * 24) / 6 // 1 day of ledgers @ 6s each
 )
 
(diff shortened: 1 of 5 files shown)


Shaptic:
The [split diff view](https://github.com/stellar/go/pull/4457/files?diff=split&w=1) is recommended to review the LRU code.

exp/lighthorizon/archive/ingest_archive.go
```diff
@@ -2,15 +2,68 @@ package archive
...
+			Context:           context.Background(),
+			NetworkPassphrase: config.NetworkPassphrase,
+			S3Region:          region,
+		},
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	cache, err := historyarchive.MakeFsCacheBackend(source, config.CacheDir, maxLedgersToCache)
```
sreuland, line 53: is it worthwhile to make the fs cache optional with a config switch in `ArchiveConfig`

exp/lighthorizon/archive/ingest_archive.go
```diff
@@ -2,15 +2,68 @@ package archive
...
+type ArchiveConfig struct {
+	SourceUrl         string
+	NetworkPassphrase string
+	CacheDir          string
+}
+
+func NewIngestArchive(config ArchiveConfig) (Archive, error) {
+	// If the source URL is an S3 url and it has a region specified, we should
+	// try to extract it.
+	parsed, err := url.Parse(config.SourceUrl)
```
sreuland, line 30: this S3 parse/config logic seems a bit more of a concern for the archive to know about, could it be encapsulated in `historyarchive.ConnectBackend()` ?

exp/lighthorizon/main.go
```diff
@@ -21,10 +21,11 @@ func main() {
 	sourceUrl := flag.String("source", "gcs://horizon-archive-poc", "history archive url to read txmeta files")
 	indexesUrl := flag.String("indexes", "file://indexes", "url of the indexes")
 	networkPassphrase := flag.String("network-passphrase", network.TestNetworkPassphrase, "network passphrase")
+	cacheDir := flag.String("ledger-cache", "", "path to cache frequently-used ledgers")
```
sreuland, line 24: might help to mention the 'empty' behavior in the description, i.e. caching is enabled|disabled in that case, and if former, it will use o/s temp directory.

historyarchive/fs_cache.go
```diff
@@ -5,222 +5,211 @@
...
+		}
+
+		b.log.WithField("key", filepath).
+			Debug("retrieving file from remote backend")
+
+		// Since it's not on-disk, pull it from the remote backend, shove it
+		// into the cache, and write it to disk.
+		remote, err := b.ArchiveBackend.GetFile(filepath)
+		if err != nil {
+			return remote, err
```
sreuland, line 86: would `remote` be in valid state, should it return nil if err happened to avoid callers getting too involved with handling the response?

historyarchive/fs_cache.go
```diff
@@ -5,222 +5,211 @@
...
 	}
+
+	return b.ArchiveBackend.PutFile(filepath, in)
 }
 
 func (b *FsCacheBackend) Close() error {
+	// This means nothing for the cached side of things unless we start storing
+	// file handles in the cache.
+	//
+	// TODO: Actually, should we purge the cache here?
```
sreuland, line 167: purge the in-mem lru sounds good for this life cycle event, and leaves the disk in whatever state right?

historyarchive/fs_cache.go
```diff
@@ -5,222 +5,211 @@
...
-type lruCache []*lruCacheItem
-
-type lruCacheItem struct {
-	path       string
-	lastUsedAt time.Time
-	index      int
-}
-
-func (c lruCache) Len() int { return len(c) }
+func (b *FsCacheBackend) createLocal(filepath string) (*os.File, error) {
```
sreuland, line 180: should this be idempotent, if a file already exists at `filepath` for whatever reason, let this silently delete it and return with new empty file.

historyarchive/fs_cache.go
Shaptic: I think of this as a direct proxy to the wrapped backend: it returns exactly what `b.Archive.GetFile` would return. So if `b.ArchiveBackend.GetFile` returns `nil` on an error, this will likewise return `nil`. If it returns some partial result or non-fatal `err`, this does that as well.

historyarchive/fs_cache.go
Shaptic: right, that's the follow-up question: if we purge cache, do we also purge disk?

historyarchive/fs_cache.go
Shaptic: I think this already happens implicitly with [os.Create](https://pkg.go.dev/os#Create):

> If the file already exists, it is truncated. 

Similarly, `os.MkdirAll` returns nil if the path already exists.

So we're good, right?

exp/lighthorizon/archive/ingest_archive.go
Shaptic: I honestly think a cache shouldn't be optional if you're using a remote source of ledgers. Is that too opinionated? Wdyt?

historyarchive/fs_cache.go
sreuland: nice, yes, I shoulda read the docs!

historyarchive/fs_cache.go
sreuland: ok, maybe that's worth a brief doc line on the method. I assume a caller is going to take standard convention on err handling and short-circuit on err first, etc.

exp/lighthorizon/archive/ingest_archive.go
sreuland: yes, convention over config is good, if it were a local source of ledgers, does it result in un-necessarily perf impact? if so, then could consider choosing FsCache dynamically based on `sourceUrl starts with file://`. but, that could be later iteration too if it becomes noticeable.

Shaptic: yeah that's what I realized from this review comment as well - 73e87899

exp/lighthorizon/archive/ingest_archive.go
Shaptic: I somewhat agree, but this also follows the way `index.ConnectWithConfig()` works. I'm hesitant to modify `historyarchive` package since it's heavily used elsewhere so there may be some bad side-effects to any changes there, even additive ones.

exp/lighthorizon/main.go
Shaptic: done 73e87899

historyarchive/fs_cache.go
Shaptic: fair - aim is to be as pure of a wrapper as possible. added docstrings here - 0bda3f1a

historyarchive/fs_cache.go
Shaptic: nope, was a great thing to verify!

historyarchive/fs_cache.go
sreuland: I'd vote for leaving disk as-is, there's a similar pattern being followed recently for captive core restart in horizon, where preserve the `on-disk` storage path of captive core's ledger state between restarts - [#4471](https://github.com/stellar/go/pull/4471)

sreuland (approved):
nice design and code.

historyarchive/fs_cache.go
Shaptic: fair enough! done in ec93f7d6

PR #4465, opened 2022-07-19
exp/lighthorizon: Set a default number of workers.

Shaptic:
### What
This sets a default worker count for the index store if it's unspecified.

### Why
The default value for an integer in Golang is 0, which means no writer workers get created when flushing to the filesystem if you don't specify a quantity, which means nothing gets written to disk :facepalm: 

### Known limitations
It's likely that the CPU-based default value is not the best choice for every case. 

Would just setting it to 1 be better?

### DIFF
--- a/exp/lighthorizon/index/connect.go
+++ b/exp/lighthorizon/index/connect.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net/url"
 	"path/filepath"
+	"runtime"
 
 	"github.com/aws/aws-sdk-go/aws"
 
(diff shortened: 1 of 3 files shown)


exp/lighthorizon/index/connect.go
```diff
@@ -15,6 +16,10 @@ func Connect(backendUrl string) (Store, error) {
 }
 
 func ConnectWithConfig(config StoreConfig) (Store, error) {
+	if config.Workers <= 0 {
+		config.Workers = uint32(runtime.NumCPU()) - 1
```
2opremio, line 20: What if there's a single CPU?

Shaptic: LOL great point :facepalm: I'm just going to default it to 1

sreuland (approved):
nice, will this resolve the hanging tests that have been observed? maybe is related to @2opremio 's observation about race condition, I was observing the timeout locally on mac with latest from `lighthorizon` plus the fix for the shell script environment issue that happened on mac, which I committed a fix for in the [txs by account pr](https://github.com/stellar/go/pull/4453/commits/be847d19f54f4c54b36e8a3da610c9c7cb367a98)

PR #4470, opened 2022-07-20
exp/lighthorizon: Fix the single-process index builder data race.

Shaptic:
### What
Synchronizes the parallel flush job distributor goroutine.

### Why
There's a write/read data race that's breaking tests.

### Known limitations
n/a


sreuland (approved):
looks agile to me, we'll get to the bottom of it, thanks!

PR #4471, opened 2022-07-21
ingest: Reuse Stellar-Core on-disk DB in online mode

### LINKED ISSUE #4454: /services/horizon/ingest: captive core on-disk ingestion, optimize catchup times
Long ingestion catchup times from captive core on-disk mode for pubnet observed during horizon start up.

_These times were observed on the following environment:_
stellar-core 19.0.1
horizon 2.17.1 with an already full ingested history db of pubnet
CPU - 3 cores
32GB ram
500GB disk, 3K IOPS

**on-disk pubnet(1h15min for catchup, 1h25min total for latest ledger sync)** 
```
20:30:46 stellar-core new-db
20:30:46 stellar-core catchup 41746401/0
21:43:39 stellar-core run
21:45:52 horizon reports "Range prepared duration=4506.688054431 from=41746403"
... ledgers emitted on txmeta, horizon is logging ingested ledgers to history
21:55:10 stellar-core run reports "LM_CATCHING_UP_STATE -> LM_SYNCED_STATE"
```
[on-disk logs](https://github.com/stellar/go/files/9106799/on-diskcc.log.gz)

**in-memory pubnet(2min total for catchup and latest ledger sync)**
```
20:18:27 stellar-core run --in-memory --start-at-ledger 41746276 --start-at-hash xyz
20:20:44 horizon reports "Range prepared" duration=137.273946913 from=41746277"
... ledgers emitted on txmeta, horizon is logging ingested ledgers to history
20:20:54 stellar-core run reports "LM_CATCHING_UP_STATE -> LM_SYNCED_STATE"
```
[in-mem logs](https://github.com/stellar/go/files/9106801/in-memcc.log.gz)

_These on-disk times were observed on environment with faster SSD disk for comparison:_
stellar-core 19.1.0
horizon 2.18.1 with an empty history db
CPU - 8 cores
16GB ram
500GB disk, 2G/s IOPS(avg rate read/write)

**on-disk pubnet with faster disk(35min for catchup)** 
```
14:09:59 stellar-core new-db
14:09:59 stellar-core catchup 41746749/0
14:45:12 stellar-core run
14:46:44 horizon reports "Range prepared" duration=2206.337728458 from=41746751"
... ledgers emitted on txmeta, horizon is repeatedly logging 
Processing ledger entry changes pid=4073 processed_entries=XXX progress="XX.XX%"
...
let it run for another 45min, stayed in this pattern of log output, didn't see LM_SYNCED_STATE emitted, stopped process at this point."
```
[ssd on-disk logs](https://github.com/stellar/go/files/9113609/stellar-horizon.log.gz)

**summary observations:**
* The `stellar-core run --in-memory` invocation takes a couple minutes at most for catch up and emitting latest ledgers on tx meta output. 
* The `on disk` mode to perform catchup was observed taking longer time, 35mins to over an hour. 
* The catchup times for on-disk vary and are dependent on the IOPS of the disk that stellar-core's stellar.db is located, fastest observed catchup time so far was 35min on SSD disk with 2Gb/s IOPS.
 

identify solutions to address the on-disk ingestion catch up time. 
* operator has a new optional horizon command line flag which can be set to force horizon to tell captive core to use an existing on-disk core db rather than init'ing empty at startup, could result in faster startup times as sync is only needed from youngest ledger in the db. Needs to be validated if a populated db from prior core process can be re-used from a new core process or does it ignore/wipe it or fail. 
* other ideas

use older in-memory configuration for captive core, but that is running into limitations due to the 30+Gb of RAM required, on-disk mode of captive core is preferred.

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This commit changes the behaviour of `stellarCoreRunner` when using an on-disk DB in online mode to check if existing storage dir contains the DB in a state that allows Captive Core to start without rebuilding Stellar-Core state. In short, it checks (by using `stellar-core offline-info` command) if the LCL of Stellar-Core matches the requested ledger in `startFrom`.

Close #4454.

### Why

While applying state from buckets was relatively fast in memory mode of Captive Core it can be extremely slow when using disk. This change allows reusing existing state in most cases.

### Known limitations

[TODO or N/A]

### DIFF
--- a/ingest/ledgerbackend/stellar_core_runner.go
+++ b/ingest/ledgerbackend/stellar_core_runner.go
@@ -360,17 +360,15 @@ func (r *stellarCoreRunner) runFrom(from uint32, hash string) error {
 		}
 
 		if removeStorageDir {
-			err := os.RemoveAll(r.storagePath)
-			if err != nil {
+			if err = os.RemoveAll(r.storagePath); err != nil {
 				return errors.Wrap(err, "error removing existing storage-dir contents")
 			}
 
-			err = createCheckDirectory(r.storagePath)
-			if err != nil {
+			if err = createCheckDirectory(r.storagePath); err != nil {
 				return err
 			}
 
-			if _, err := r.writeConf(); err != nil {
+			if _, err = r.writeConf(); err != nil {
 				return errors.Wrap(err, "error writing configuration")
 			}
 
@@ -381,10 +379,10 @@ func (r *stellarCoreRunner) runFrom(from uint32, hash string) error {
 			// Do a quick catch-up to set the LCL in core to be our expected starting
 			// point.
 			if from > 2 {
-				if err := r.createCmd("catchup", fmt.Sprintf("%d/0", from-1)).Run(); err != nil {
+				if err = r.createCmd("catchup", fmt.Sprintf("%d/0", from-1)).Run(); err != nil {
 					return errors.Wrap(err, "error runing stellar-core catchup")
 				}
-			} else if err := r.createCmd("catchup", "2/0").Run(); err != nil {
+			} else if err = r.createCmd("catchup", "2/0").Run(); err != nil {
 				return errors.Wrap(err, "error runing stellar-core catchup")
 			}
 		}
--- a/ingest/ledgerbackend/stellar_core_runner.go
+++ b/ingest/ledgerbackend/stellar_core_runner.go
@@ -3,6 +3,7 @@ package ledgerbackend
 import (
 	"bufio"
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -18,6 +19,7 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
+	"github.com/stellar/go/protocols/stellarcore"
 	"github.com/stellar/go/support/log"
 )
 
@@ -66,6 +68,7 @@ type stellarCoreRunner struct {
 	processExitError error
 
 	storagePath string
+	toml        *CaptiveCoreToml
 	useDB       bool
 	nonce       string
 
@@ -102,18 +105,9 @@ func newStellarCoreRunner(config CaptiveCoreConfig, mode stellarCoreRunnerMode) 
 		fullStoragePath = path.Join(config.StoragePath, "captive-core")
 	}
 
-	info, err := os.Stat(fullStoragePath)
-	if os.IsNotExist(err) {
-		innerErr := os.MkdirAll(fullStoragePath, os.FileMode(int(0755))) // rwx|rx|rx
-		if innerErr != nil {
-			return nil, errors.Wrap(innerErr, fmt.Sprintf(
-				"failed to create storage directory (%s)", fullStoragePath))
-		}
-	} else if !info.IsDir() {
-		return nil, errors.New(fmt.Sprintf("%s is not a directory", fullStoragePath))
-	} else if err != nil {
-		return nil, errors.Wrap(err, fmt.Sprintf(
-			"error accessing storage directory (%s)", fullStoragePath))
+	err := createCheckDirectory(fullStoragePath)
+	if err != nil {
+		return nil, err
 	}
 
 	ctx, cancel := context.WithCancel(config.Context)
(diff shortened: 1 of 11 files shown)


ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -318,17 +347,44 @@ func (r *stellarCoreRunner) runFrom(from uint32, hash string) error {
...
-		if err := r.createCmd("new-db").Run(); err != nil {
-			return errors.Wrap(err, "error initializing core db")
+		// Check if on-disk core DB exists and what's the LCL there. If not what
+		// we need remove storage dir and start from scratch.
+		removeStorageDir := false
+		info, err := r.offlineInfo()
+		if err != nil {
+			r.log.Infof("Error running offline-info: %v, removing existing storage-dir contents", err)
+			removeStorageDir = true
+		} else if uint32(info.Info.Ledger.Num) != from {
```
sreuland, line 357: do you know how core maintains `info.Info.Ledger.Num` , i.e. does it only bump it when it knows the meta record for that sequence was read off the pipe? wondering if `info.Info.Ledger.Num` will tend to be farther ahead than `from` which represents the last sequence that horizon read off the pipe(and serialized to history), if it does drift asynchronously from meta pipe reader activity(horizon), then this condition won't get hit much, right, result being it ends up in same routine of `new-db/catchup`?

ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -318,17 +347,44 @@ func (r *stellarCoreRunner) runFrom(from uint32, hash string) error {
 	}
 
 	if r.useDB {
-		if err := r.createCmd("new-db").Run(); err != nil {
-			return errors.Wrap(err, "error initializing core db")
+		// Check if on-disk core DB exists and what's the LCL there. If not what
+		// we need remove storage dir and start from scratch.
+		removeStorageDir := false
```
sreuland, line 352: might be worthwhile to add a unit test in `stellar_core_runner_test.go` to assert this new outcome?

bartekn: Just a quick update on this: I'm working on refactoring `stellarCoreRunner` to allow writing better unit tests. I'll have a new commit ready by the end of today.

ingest/ledgerbackend/stellar_core_runner.go
bartekn: To my best knowledge and some experimenting it seems that Stellar-Core only closes the ledger once it's read from meta pipe. This leaves us with two cases:
* Horizon is catching up (after restart or state build) - it this case `bufferedLedgerMetaReader` can read ledgers from meta pipe upfront which will make the Horizon to be behind. In this case, when Horizon is stopped with ledgers in the buffer the solution in this PR will not work because the ledger sequences in will not match on restart. We can try removing `bufferedLedgerMetaReader` in _online_ mode but I'm not sure about performance of this change. We can explore it in a separate PR.
* Horizon is ingesting latest ledgers - in this case the `bufferedLedgerMetaReader` will contain up to one ledger but if Horizon is shutdown gracefully it will process this ledger before shutting down.

ingest/ledgerbackend/stellar_core_runner.go
bartekn: Actually while refactoring I changed some other parts of `stellarCoreRunner` that seemed inconsistent. Would you mind 👍 this PR (if there is nothing else that requires changes) and I'll open another PR with refactoring and tests?

ingest/ledgerbackend/stellar_core_runner.go
sreuland: ok, that's interesting, meaning there's only one ledger of data present in that pipe at any time, sounds like core writer blocks until it's empty, which is the signal that prior ledger was read, but, this at least recovers from any out-of-sync case and worst outcome is it does the same as current day of full removal first and init first.

sreuland (approved):
nice solution with minimal coding!

ingest/ledgerbackend/stellar_core_runner.go
bartekn: @sreuland follow up PR: https://github.com/stellar/go/pull/4480

PR #4476, opened 2022-07-25
exp/lighthorizon: Add duration metrics for on-the-fly ingestion elements.

Shaptic:
### What
Add basic aggregate metrics for request fulfillment:

 - how long did ledger _downloads_ take, on average?
 - how long did ledger _processing_ take, on average?
 - how long did index lookups take, on average?
 - how many ledgers were needed?
 - how long did the entire request take, in total?

### Why
#4451, these will be crucial for doing performance analysis, identifying optimization opportunities, and making comparisons to classic Horizon.

### Known limitations
There are some alternatives:
 - logging each of these metrics on its own line
 - should these be part of Prometheus, instead?

### DIFF
--- a/exp/lighthorizon/main.go
+++ b/exp/lighthorizon/main.go
@@ -30,6 +30,7 @@ if left empty, uses a temporary directory`)
 
 	L := log.WithField("service", "horizon-lite")
 	L.SetLevel(log.InfoLevel)
+	L.SetLevel(log.DebugLevel)
 	L.Info("Starting lighthorizon!")
 
 	registry := prometheus.NewRegistry()
--- a/exp/lighthorizon/main.go
+++ b/exp/lighthorizon/main.go
@@ -30,7 +30,6 @@ if left empty, uses a temporary directory`)
 
 	L := log.WithField("service", "horizon-lite")
 	L.SetLevel(log.InfoLevel)
-	L.SetLevel(log.DebugLevel)
 	L.Info("Starting lighthorizon!")
 
 	registry := prometheus.NewRegistry()
--- a/exp/lighthorizon/services/main.go
+++ b/exp/lighthorizon/services/main.go
@@ -78,7 +78,7 @@ type TransactionsService struct {
 }
 
 type OperationsService struct {
-	OperationsRepository,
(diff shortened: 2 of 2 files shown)


exp/lighthorizon/main.go
```diff
@@ -30,6 +30,7 @@ if left empty, uses a temporary directory`)
 
 	L := log.WithField("service", "horizon-lite")
 	L.SetLevel(log.InfoLevel)
+	L.SetLevel(log.DebugLevel)
```
tamirms, line 33: should we delete either the line above or this line? it doesn't make sense to set the log level to info and then to set it to debug right afterwards

exp/lighthorizon/services/main.go
```diff
@@ -259,13 +293,27 @@ func searchTxByAccount(ctx context.Context, cursor int64, accountId string, conf
 			}
 		}
 
+		incrementAverage(&avgProcessDuration, time.Since(start), count)
+
+		start = time.Now()
 		cursor, err = cursorMgr.Advance()
```
tamirms, line 299: should we check err before  fetching the ledger?

tamirms: nevermind, I misunderstood getLedgerFromCursor

exp/lighthorizon/services/main.go
```diff
@@ -259,13 +293,27 @@ func searchTxByAccount(ctx context.Context, cursor int64, accountId string, conf
...
+		nextLedger = getLedgerFromCursor(cursor)
+		incrementAverage(&avgIndexFetchDuration, time.Since(start), count)
```
tamirms, line 301: to be consistent with the other metrics it seems like we should only include durations where err is nil

exp/lighthorizon/main.go
Shaptic: Yeah this probably slipped into a commit while I was testing things locally

exp/lighthorizon/services/main.go
Shaptic: good call, done in 47047f7d!

tamirms (approved):

PR #4480, opened 2022-07-26
ingest/ledgerbackend: Refactor stellarCoreRunner

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This commit refactors `stellarCoreRunner` to allow writing unit tests that does not call fs/os functions and adds unit tests for new functionality added in 4850d22. It also changes `newStellarCoreRunner` by removing `mode` argument. The value will be set internally based on method calls.

### Why

Code and test infrastructure improvements.

### Known limitations

[TODO or N/A]


ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -51,18 +53,101 @@ type pipe struct {
 	File io.Closer
 }
 
+type isDir interface {
```
sreuland, line 56: just suggestion as stellar_core_runner.go is getting large, could move these interfaces and method implementations up to ingest/ledgerbackend/main.go?

ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -84,71 +169,62 @@ func createRandomHexString(n int) string {
...
+		//    https://github.com/stellar/go/issues/3631
+		return path.Join(r.storagePath, "captive-core")
 	}
-
-	return runner, nil
 }
 
-func createCheckDirectory(fullStoragePath string) error {
-	info, err := os.Stat(fullStoragePath)
+func (r *stellarCoreRunner) createCheckDirectory() error {
```
sreuland, line 215: super minor, method name doesn't match to its purpose at least was my interpretation, maybe something like `establishStorageDirectory` if that sounds reasonable.

ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -84,71 +169,62 @@ func createRandomHexString(n int) string {
...
+		// On Windows, first we ALWAYS append something to the base storage path,
+		// because we will delete the directory entirely when Horizon stops. We also
+		// add a random suffix in order to ensure that there aren't naming
+		// conflicts.
+		// This is done because it's impossible to send SIGINT on Windows so
+		// buckets can become corrupted.
+		// We also want to use random directories in offline mode (reingestion)
+		// because it's possible it's running multiple Stellar-Cores on a single
+		// machine.
+		return path.Join(r.storagePath, "captive-core-"+createRandomHexString(8))
```
sreuland, line 205: are the comments, ln 196-204, still relevant for windows o/s, it looks like this would prevent windows o/s deployments from re-using the same storage path in `runFrom` correct?

ingest/ledgerbackend/stellar_core_runner.go
```diff
@@ -557,15 +667,15 @@ func (r *stellarCoreRunner) close() error {
 		r.pipe.Reader.Close()
 	}
 
-	if runtime.GOOS == "windows" ||
+	if r.mode != 0 && (runtime.GOOS == "windows" ||
```
sreuland, line 670: similar to other comment above, looks like intentionally won't preserve storage path across restarts for windows o/s.

ingest/ledgerbackend/stellar_core_runner_test.go
```diff
@@ -2,107 +2,272 @@ package ledgerbackend
...
+	).Return(cmdMock)
+	scMock.On("removeAll", mock.Anything).Return(nil)
+	runner.systemCaller = scMock
 
-	// Directory cleaned up on shutdown when reingesting to save space
-	_, err = os.Stat(tempDir)
-	assert.Error(t, err)
-	assert.Contains(t, err.Error(), "no such file or directory")
+	assert.NoError(t, runner.catchup(100, 200))
+	assert.NoError(t, runner.close())
```
sreuland, line 52: nice coverage!

ingest/ledgerbackend/stellar_core_runner_test.go
```diff
@@ -2,107 +2,272 @@ package ledgerbackend
...
+		"100",
+		"--start-at-hash",
+		"hash",
+		"--metadata-output-stream",
+		"fd:3",
+	).Return(cmdMock)
+	runner.systemCaller = scMock
 
+	assert.NoError(t, runner.runFrom(100, "hash"))
 	assert.NoError(t, runner.close())
```
sreuland, line 94: maybe assert on storage path persistence , `assert.DirExists(t, runner.StoragePath)`

sreuland, line 94: maybe assert on storage path persistence , assert.DirExists(t, runner.StoragePath)

ingest/ledgerbackend/stellar_core_runner_test.go
```diff
@@ -2,107 +2,272 @@ package ledgerbackend
...
+		mock.Anything,
+		"run",
+		"--metadata-output-stream",
+		"fd:3",
+	).Return(cmdMock)
+	// removeAll not called
+	runner.systemCaller = scMock
+
+	assert.NoError(t, runner.runFrom(100, "hash"))
 	assert.NoError(t, runner.close())
```
sreuland, line 199: maybe assert on storage path persistence after close, `assert.DirExists(t, runner.StoragePath)`

ingest/ledgerbackend/stellar_core_runner.go
bartekn: Yes, Windows is unable to shutdown internal process gracefully so in most cases sqlite DB is just broken after shutdown.

ingest/ledgerbackend/stellar_core_runner_test.go
bartekn: Not sure if we should do that because after calling `close` the `stellarCoreRunner` instance shouldn't be reused. So we don't care about it's internal state anymore.

bartekn:
@sreuland PTAL.

ingest/ledgerbackend/stellar_core_runner_test.go
sreuland: the intent is to assert core's `stellar.db`which resides on disk's `runner.StoragePath` is not deleted after shutdown of runner correct?

ingest/ledgerbackend/stellar_core_runner.go
sreuland: ah, ok, should the entry for this change be added to Changelog.MD now, to mention this additional aspect that windows is not supported o/s for this optimization of re-using core's state db?

sreuland (approved):
great test coverage!

ingest/ledgerbackend/stellar_core_runner_test.go
bartekn: OK, I misread the assertion as checking if `runner.StoragePath` is empty. Anyway, we don't need to do that because we ensure that mocked `systemCaller.removeAll` is not called. `assert.DirExists` checks if there's a real file in real filesystem. All of the filesystem calls in this test are mocked so no files are created during the test.

PR #4482, opened 2022-07-27
exp/lighthorizon/index: Drop building indices for successful transactions.

Shaptic:
### What
For the MVP described in #4317, we don't need indices for successful transactions. We only need them for all transactions and payments, should we decide to expand to that endpoint.

This also slightly unifies the code & its tests to do checkpoint number math consistently.

### Why
This will reduce the index sizes for the MVP.


bartekn (approved):
LGTM!

PR #4488, opened 2022-07-28
Refactor access to meta archive

### LINKED ISSUE #4460: exp/lighthorizon: Refactor codebase to avoid using `HistoryArchiveBackend` as a filesystem interface
Remove kludges around using `historyarchive.ArchiveBackend` as a filesystem interface.

Currently, we're using `historyarchive.ArchiveBackend` as a way to access raw ledgers from the filesystem. This is pretty hacky since these aren't raw `LedgerCloseMeta`s but rather a custom format on top of them (see #4450), so it doesn't make sense as a `LedgerBackend` anyway, plus we want to be able to use the version and maybe read other config data from the directory.

There's a little bit more context [here](https://github.com/stellar/go/pull/4450#discussion_r916104923); the general point is that we shouldn't hack in the ledgerbackend interface for Horizon Lite.

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Refactor `historyarchive` and `ledgerbackend` to allow better access to the new meta archives:
* Created `metaarchive` package that connects to the new meta archives (and allows accessing `xdr.SerializedLedgerCloseMeta`).
* Extracted `ArchiveBackend` to the new `support/storage` package as it contains only storage related methods. New package is used in both `historyarchive` and `metaarchive`.

Close #4460.

### Why

Improve code separation, allow access to `metaarchive` specific objects.

### Known limitations

[TODO or N/A]

### DIFF
--- a/historyarchive/archive_test.go
+++ b/historyarchive/archive_test.go
@@ -17,6 +17,7 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/stellar/go/support/storage"
 	"github.com/stellar/go/xdr"
 	"github.com/stretchr/testify/assert"
 )
@@ -35,11 +36,17 @@ func GetTestS3Archive() *Archive {
 	if env_region := os.Getenv("ARCHIVIST_TEST_S3_REGION"); env_region != "" {
 		region = env_region
 	}
-	return MustConnect(bucket, ConnectOptions{S3Region: region, CheckpointFrequency: 64})
+	return MustConnect(
+		bucket,
+		ArchiveOptions{
+			CheckpointFrequency: 64,
+			ConnectOptions:      storage.ConnectOptions{S3Region: region},
+		},
+	)
 }
 
 func GetTestMockArchive() *Archive {
-	return MustConnect("mock://test", ConnectOptions{CheckpointFrequency: 64})
+	return MustConnect("mock://test", ArchiveOptions{CheckpointFrequency: 64})
 }
 
 var tmpdirs []string
@@ -54,7 +61,7 @@ func GetTestFileArchive() *Archive {
 	} else {
 		tmpdirs = append(tmpdirs, d)
 	}
-	return MustConnect("file://"+d, ConnectOptions{CheckpointFrequency: 64})
+	return MustConnect("file://"+d, ArchiveOptions{CheckpointFrequency: 64})
 }
 
 func cleanup() {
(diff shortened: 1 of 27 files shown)


historyarchive/archive_test.go
```diff
@@ -35,11 +36,17 @@ func GetTestS3Archive() *Archive {
 	if env_region := os.Getenv("ARCHIVIST_TEST_S3_REGION"); env_region != "" {
 		region = env_region
 	}
-	return MustConnect(bucket, ConnectOptions{S3Region: region, CheckpointFrequency: 64})
+	return MustConnect(
+		bucket,
+		ArchiveOptions{
+			CheckpointFrequency: 64,
```
Shaptic, line 42: I know this is somewhat unrelated code, but maybe we can use the `historyarchive.DefaultCheckpointFrequency` here?

ingest/ledgerbackend/history_archive_backend.go
```diff
@@ -53,22 +33,19 @@ func (b *HistoryArchiveBackend) IsPrepared(ctx context.Context, ledgerRange Rang
...
+	output, isV0 := serializedLedger.GetV0()
 	if !isV0 {
-		return xdr.LedgerCloseMeta{}, fmt.Errorf("unexpected serialized ledger version number (0x%x)", ledger.V)
+		return xdr.LedgerCloseMeta{}, fmt.Errorf("unexpected serialized ledger version number (0x%x)", serializedLedger.V)
 	}
 	return output, nil
 }
+
+func (b *HistoryArchiveBackend) Close() error {
+	// Noop
```
Shaptic, line 49: is there a chance we need to close the meta archive's storage backend?

Shaptic (commented):
I really like this refactor - separating out the txmeta archives should be really helpful going forward as we expand the functionality there. Any limitations you thought of or encountered while doing this?

ingest/ledgerbackend/history_archive_backend.go
bartekn: We don't now and I don't think we'll need to close it in the future. `HistoryArchiveBackend.Close` is here because `HistoryArchiveBackend` must implement `LedgerBackend` interface.

Shaptic (approved):
I'm a fan of this refactor! :shipit:

PR #4490, opened 2022-07-28
services/horizon/docker/ledgerexporter: deploy ledgerexporter image as service

sreuland:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Enable the ledgerexporter image to be deployed/run as service in container orchestration, enable more env settings for configurability between testnet/pubnet. Create deployment on kubernetes that emits to S3 bucket.

### Why

Will enable latest txmeta from network to be constantly exported to s3 buckets.

### DIFF
--- a/exp/services/ledgerexporter/main.go
+++ b/exp/services/ledgerexporter/main.go
@@ -72,18 +72,23 @@ func main() {
 	// Build the appropriate range for the given backend state.
 	startLedger := uint32(*startingLedger)
 	endLedger := uint32(*endingLedger)
+
+	logger.Infof("processing requested range of -start-ledger=%v, -end-ledger=%v", startLedger, endLedger)
+	if *continueFromLatestLedger {
+		if startLedger != 0 {
+			logger.Fatalf("-start-ledger and -continue cannot both be set")
+		}
+		startLedger = readLatestLedger(target)
+		logger.Infof("continue flag was enabled, next ledger found was %v", startLedger)
+	}
+
 	if startLedger < 2 {
 		logger.Fatalf("-start-ledger must be >= 2")
 	}
 	if endLedger != 0 && endLedger < startLedger {
 		logger.Fatalf("-end-ledger must be >= -start-ledger")
 	}
-	if *continueFromLatestLedger {
-		if startLedger != 0 {
-			logger.Fatalf("-start-ledger and -continue cannot both be set")
-		}
-		startLedger = readLatestLedger(target)
-	}
+
 	var ledgerRange ledgerbackend.Range
 	if endLedger == 0 {
 		ledgerRange = ledgerbackend.UnboundedRange(startLedger)
--- a/exp/services/ledgerexporter/main.go
+++ b/exp/services/ledgerexporter/main.go
@@ -29,6 +29,7 @@ func main() {
 	continueFromLatestLedger := flag.Bool("continue", false, "start export from the last exported ledger (as indicated in the target's /latest path)")
 	endingLedger := flag.Uint("end-ledger", 0, "ledger at which to stop the export (must be a closed ledger), 0 means no ending")
 	writeLatestPath := flag.Bool("write-latest-path", true, "update the value of the /latest path on the target")
+	captiveCoreUseDb := flag.Bool("captive-core-use-db", true, "configure captive core to store database on disk in working directory rather than in memory")
 	flag.Parse()
 
 	logger.SetLevel(supportlog.InfoLevel)
@@ -51,6 +52,7 @@ func main() {
 		CheckpointFrequency: 64,
 		Log:                 logger.WithField("subservice", "stellar-core"),
 		Toml:                captiveCoreToml,
+		UseDB:               *captiveCoreUseDb,
 	}
 	core, err := ledgerbackend.NewCaptive(captiveConfig)
 	logFatalIf(err, "Could not create captive core instance")
@@ -91,7 +93,7 @@ func main() {
 	err = core.PrepareRange(context.Background(), ledgerRange)
 	logFatalIf(err, "could not prepare range")
 
-	for nextLedger := startLedger; nextLedger <= endLedger; {
+	for nextLedger := startLedger; endLedger < 1 || nextLedger <= endLedger; {
 		ledger, err := core.GetLedger(context.Background(), nextLedger)
 		if err != nil {
 			logger.WithError(err).Warnf("could not fetch ledger %v, retrying", nextLedger)
--- a/exp/services/ledgerexporter/main.go
+++ b/exp/services/ledgerexporter/main.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/aws/aws-sdk-go/service/s3"
 	"github.com/stellar/go/historyarchive"
 	"github.com/stellar/go/ingest/ledgerbackend"
 	"github.com/stellar/go/network"
@@ -61,7 +62,8 @@ func main() {
 	target, err := historyarchive.ConnectBackend(
 		*targetUrl,
 		storage.ConnectOptions{
-			Context: context.Background(),
+			Context:    context.Background(),
+			S3WriteACL: s3.ObjectCannedACLBucketOwnerFullControl,
 		},
 	)
 	logFatalIf(err, "Could not connect to target")
--- a/historyarchive/s3_archive.go
+++ b/historyarchive/s3_archive.go
@@ -121,7 +121,6 @@ func (b *S3ArchiveBackend) PutFile(pth string, in io.ReadCloser) error {
 	params := &s3.PutObjectInput{
 		Bucket: aws.String(b.bucket),
 		Key:    aws.String(key),
-		ACL:    aws.String(s3.ObjectCannedACLPublicRead),
 		Body:   bytes.NewReader(buf.Bytes()),
 	}
 	req, _ := b.svc.PutObjectRequest(params)
--- a/services/horizon/docker/ledgerexporter/ledgerexporter.yml
+++ b/services/horizon/docker/ledgerexporter/ledgerexporter.yml
@@ -18,11 +18,11 @@ data:
   # when using core 'on disk', the earliest ledger to get streamed out after catchup to 2, is 3
   # whereas on in-memory it streas out 2, adjusted here, otherwise horizon ingest will abort
   # and stop process with error that ledger 3 is not <= expected ledger of 2.    
-  START: "3"
+  #START: "3"
   END: "0"
 
   # can only have CONTINUE or START set, not both.
-  #CONTINUE: "true"
+  CONTINUE: "true"
   WRITE_LATEST_PATH: "true"
   CAPTIVE_CORE_USE_DB: "true"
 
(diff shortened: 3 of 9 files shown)


exp/services/ledgerexporter/main.go
```diff
@@ -91,7 +93,7 @@ func main() {
 	err = core.PrepareRange(context.Background(), ledgerRange)
 	logFatalIf(err, "could not prepare range")
 
-	for nextLedger := startLedger; nextLedger <= endLedger; {
+	for nextLedger := startLedger; endLedger < 1 || nextLedger <= endLedger; {
```
sreuland, line 96: during testing with END=0, ran into this as it was stopping before generation anything

historyarchive/s3_archive.go
```diff
@@ -121,7 +121,6 @@ func (b *S3ArchiveBackend) PutFile(pth string, in io.ReadCloser) error {
 	params := &s3.PutObjectInput{
 		Bucket: aws.String(b.bucket),
 		Key:    aws.String(key),
-		ACL:    aws.String(s3.ObjectCannedACLPublicRead),
```
sreuland, line 124: Current txmeta bucket uses policy for auth, and it requires credentials from an IAM principal(initially a user in the AWS account, need to create a non-user service account) but when using policy with `bucket owner enforced`, it can trigger auth error if explicit ACLs are set by client when accessing the bucket, and this ACL was causing `AccessControlListNotSupported `, I can either change the txmeta bucket to be ACL based and revert this, or if this is only pertinent to specific bucket(s), could keep this and tick the `Everyone` for `Read` ACL on those buckets to accomplish the same. Noticed this was the only usage of s3 ACLs, I see other s3 usages that don't send the ACL.

services/horizon/docker/ledgerexporter/ledgerexporter.yml
```diff
@@ -0,0 +1,92 @@
...
+            cpu:      250m
+            memory:   500m
+        volumeMounts:
+        - mountPath: /cc
+          name: tempfs-volume
+      dnsPolicy: ClusterFirst
+      volumes:
+      - name: tempfs-volume
+        emptyDir: 
+          medium: Memory
```
sreuland, line 89: captive core's on-disk db(stellar.db) will be on this tempfs RAM drive

services/horizon/docker/ledgerexporter/ledgerexporter.yml
```diff
@@ -0,0 +1,92 @@
...
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  labels:
+    app: ledgerexporter
+  name: ledgerexporter-pubnet-secret
+type: Opaque
+data:
+  AWS_REGION: <base64 encoded value here>
```
sreuland, line 37: AWS credentials loaded into the cluster as secrets, which deployment pulls into ledgerexporter container as env variables.

services/horizon/docker/ledgerexporter/ledgerexporter.yml
```diff
@@ -0,0 +1,92 @@
...
+data:
+  START: "2"
+  END: "0"
+  # can only have CONTINUE or START set, not both.
+  #CONTINUE: "true"
+  WRITE_LATEST_PATH: "true"
+  CAPTIVE_CORE_USE_DB: "true"
+  HISTORY_ARCHIVE_URLS: "https://history.stellar.org/prd/core-live/core_live_001,https://history.stellar.org/prd/core-live/core_live_002,https://history.stellar.org/prd/core-live/core_live_003"
+  NETWORK_PASSPHRASE: "Public Global Stellar Network ; September 2015"
+  ARCHIVE_TARGET: "s3://horizon-ledgermeta-pubnet"
```
sreuland, line 27: i've config'd the [horizon-ledgermeta-pubnet](https://s3.console.aws.amazon.com/s3/buckets/horizon-ledgermeta-pubnet?region=us-east-1&tab=permissions) bucket in same AWS account as Batch, with `bucket owner enforced` with `ACLs disabled` and a inline policy that defines allowed/disallowed statements, following [aws recommendations ](https://docs.aws.amazon.com/AmazonS3/latest/userguide/about-object-ownership.html).

sreuland:
The deployment is active on the dev k8s cluster:
`$ kubectl describe deployment ledgerexporter -n horizon-dev`

logs here:
`$ kubectl logs -l app=ledgerexporter -f -n horizon-dev`

exp/services/ledgerexporter/main.go
2opremio: Good catch

services/horizon/docker/ledgerexporter/ledgerexporter.yml
2opremio: You mean you changed the existing bucket settings?

Unfortunately the S3-writing code (HistoryArchive) code assumes ACLs are enabled (because it writes them). That's why they where enabled.

I am not against the change but then we should also change the writing code.

services/horizon/docker/ledgerexporter/start
```diff
@@ -1,17 +1,25 @@
...
+CAPTIVE_CORE_USE_DB="${CAPTIVE_CORE_USE_DB:=true}"
 
 if [ -z "$ARCHIVE_TARGET" ]; then
   echo "error: undefined ARCHIVE_TARGET env variable"
   exit 1
 fi
 
+if [ "$NETWORK_PASSPHRASE" = "Test SDF Network ; September 2015" ]; then
+  CAPTIVE_CORE_CONFIG="/captive-core-testnet.cfg"
+fi
```
2opremio, line 21: This is a bit brittle. It may be better to add an env variable to specify the network explicitly.

support/storage/s3.go
```diff
@@ -150,7 +150,6 @@ func (b *S3Storage) PutFile(pth string, in io.ReadCloser) error {
 	params := &s3.PutObjectInput{
 		Bucket: aws.String(b.bucket),
 		Key:    aws.String(key),
-		ACL:    aws.String(s3.ObjectCannedACLPublicRead),
```
2opremio, line 153: Ahhh, I see you changed it here. Is this horizon-light-only code? Otherwise we would need to check whether it breaks something.

services/horizon/docker/ledgerexporter/ledgerexporter.yml
sreuland: I haven't changed any existing bucket permissions, but this looks fairly constrained on ACL usage though, only one place in s3 writing code sets the object ACL to include `public read` during put [which I removed and left a comment on one potential to migrate off that.](https://github.com/stellar/go/pull/4490#discussion_r933663812). There are several other places where s3 object writing(puts) are done with s3uploadmanager and those don't specify ACLs.

So, the net seems to be in identifying how many existing buckets which could have been written to by this routine in `historyarchive/s3_archive.go`. With just removal of ACL here, the net effect is that put still works to the existing buckets but the objects won't have public read until the bucket permissions are updated to have ACL's disabled and a policy added with statement for Allow `Everyone/Public Read`.

support/storage/s3.go
sreuland: yes, [I was wondering the same on usage](https://github.com/stellar/go/pull/4490#discussion_r934826078), it seems like it may be contained to just the AWS Batch account S3?

services/horizon/docker/ledgerexporter/start
sreuland: yes, cleaned that up with some more docs on settings for pub, test, standalone, https://github.com/stellar/go/pull/4490/commits/2c8589cd50fabb832623bc3242a5c98d3820ef04

support/storage/s3.go
Shaptic: Does the `git blame` on this line have any meaningful history behind it? I'm pretty sure this storage backend is used by Horizon (though maybe not by SDF's deployments), but I'm not sure.

services/horizon/docker/ledgerexporter/ledgerexporter.yml
sreuland: scratch that, after further testing on cluster, found the tempfs ram was not suitable/stable for captive core on-disk db usage, the size of this volume is derived from the underlying host k8s node RAM, and is limited by default to half, so, it's variable, anytime the pod restarts it could end up running on a diff node with different RAM and I saw a few OOMKilled due to this, switched volume to use a PV/PVC.

support/storage/s3.go
sreuland: Yeah, that [ACL has been there since 2017](https://github.com/stellar/go/commit/f9aab491f80d6d28f7ec3a50aba02c4b8ebffd50#diff-96a84ec806044eb47844f0d1e0f5eb4a14c232abfc965be72c2595a620cb41fcR53), so, could be used in various ways, @bartekn , do you have insight on types of usages of `S3Storage.PutFile`, like which SDF S3 buckets has it been used or is it used by customers for their own S3 stuff? 

I think I'm going to revert this line, leave that ACL in to avoid unraveling, and instead add `ConnectOptions.DisableS3ACL` and `S3Storage.disableS3ACL` , ledgerexporter can then set that to true when invoking `ConnectBackend`, lmk if any thoughts either way, thx!

sreuland: ok, I went with that change and added config option to override the S3 File ACL rule applied - https://github.com/stellar/go/pull/4490/commits/452b20c6bca6634f202004eb03634d9d9c6bd649

bucket owner policy works with `ObjectCannedACLBucketOwnerFullControl`

services/horizon/docker/ledgerexporter/ledgerexporter.yml
sreuland: I updated approach and added ACL config option on s3 - https://github.com/stellar/go/pull/4490/commits/452b20c6bca6634f202004eb03634d9d9c6bd649, this way client can work with buckets in either permissions config.

historyarchive/s3_archive.go
sreuland: I went with alternative approach and added ACL config option  - https://github.com/stellar/go/commit/452b20c6bca6634f202004eb03634d9d9c6bd649, this way client can work with buckets in either permissions config.

services/horizon/docker/ledgerexporter/ledgerexporter.yml
sreuland: @2opremio , I added more to permissions on `horizon-ledgermeta-prodnet-test` and `horizon-index`, added policy to grant write to ec2(Batch access) and an IAM user(k8s access) and a PublicRead rule.

I noticed slight diff on their main config, horizon-ledgermeta-prodnet-test had `ACLs enabled/bucket owner preferred`, and horizon-index  had `ACLs disabled/bucket owner enforced`. The policy applies on top of either.

Shaptic (approved):
LGTM! :shipit: :+1:

support/storage/s3.go
2opremio: Good! Will the code still work if we remove ACL support from the bucket?

sreuland: It will if client is the bucket owner, and for any other prinicipals that should have access, adding an inline policy on the bucket with allow statements should provide write and read rules, I added such a policy doc to `horizon-ledgermeta-prodnet-test`, I noticed `horizon-index-prodnet-test` had ACLs disabled with owner enforced already, but it doesn't have policy set yet, we can update that for Public(read) and other principals like k8s, which would need read for web and write for single indexer(watching for latest).

PR #4491, opened 2022-07-28
exp/lighthorizon/index: Parse network passphrase from the env.

Shaptic:
### What
Besides some minor logging and memory usage improvements, this adds the `NETWORK_PASSPHRASE` env which can be set to either `pubnet` or `testnet` for the public and test network passphrases, respectively, or a custom passphrase.

### Why
This is important for getting the map jobs on AWS Batch.

### Known limitations
n/a

### DIFF
--- a/exp/lighthorizon/index/builder.go
+++ b/exp/lighthorizon/index/builder.go
@@ -99,8 +99,8 @@ func BuildIndices(
 
 	// Submit the work to the channels, breaking up the range into individual
 	// checkpoint ranges.
+	checkpoints := historyarchive.NewCheckpointManager(0)
 	go func() {
-		checkpoints := historyarchive.NewCheckpointManager(0)
 		for ledger := range ledgerRange.GenerateCheckpoints(checkpoints) {
 			chunk := checkpoints.GetCheckpointRange(ledger)
 			chunk.High = min(chunk.High, ledgerRange.High) // don't exceed upper bound
@@ -112,7 +112,6 @@ func BuildIndices(
 		close(ch)
 	}()
 
-	chkProcessed := uint64(0)
 	processed := uint64(0)
 	for i := 0; i < parallel; i++ {
 		wg.Go(func() error {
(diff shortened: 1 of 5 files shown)


exp/lighthorizon/index/builder.go
```diff
@@ -128,12 +129,14 @@ func BuildIndices(
...
 				if nprocessed%97 == 0 {
-					printProgress("Reading ledgers", nprocessed, uint64(ledgerCount), startTime)
+					PrintProgress("Reading ledgers", nprocessed, uint64(ledgerCount), startTime)
 				}
 
-				// Upload indices once per checkpoint to save memory
-				if err := indexStore.Flush(); err != nil {
-					return errors.Wrap(err, "flushing indices failed")
+				// Upload indices once every 10 checkpoints to save memory
+				if atomic.AddUint64(&chkProcessed, uint64(1))%10 == 0 {
```
sreuland, line 136: rather than new counter, just curious, could the same congruence be found from the existing counter like
`if  nprocessed%640 == 0`

exp/lighthorizon/index/cmd/batch/map/main.go
```diff
@@ -65,12 +68,28 @@ func NewBatchConfig() (*BatchConfig, error) {
...
+	switch networkPassphrase {
+	case "":
+		log.Warnf("%s not specified, defaulting to 'pubnet'", networkPassphraseEnv)
+		fallthrough
+	case "pubnet":
+		networkPassphrase = network.PublicNetworkPassphrase
+	case "testnet":
+		networkPassphrase = network.TestNetworkPassphrase
+	default:
+		return nil, fmt.Errorf("%s must be 'pubnet' or 'testnet', got '%s'",
```
sreuland, line 81: should this be open to allowing config for standalone network passphrases? i.e. remove`default`.  would there be some potential type of use case where people want to build indexing on their own private networks.

sreuland (approved):
looks good, I left a couple questions for consideration.

2opremio (approved):

exp/lighthorizon/index/cmd/batch/map/main.go
Shaptic: Good call! Changed in ce19bd97.

exp/lighthorizon/index/builder.go
Shaptic: Great suggestion! Done in aab29686.

PR #4492, opened 2022-08-01
exp/lighthorizon: Add response age prometheus metrics

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

As part of https://github.com/stellar/go/issues/4451 , this PR adds a metric to keep track of the age of the data contained in the service response. The response age is defined as the difference between the current time and the ledger close time of the oldest item in the service response.

This PR also adds the HTTP route for the request duration metric which is something I forgot to do in https://github.com/stellar/go/pull/4486 .

### Why

So we can know how far back in time we need to go for the requests served by horizon lite.

### Known limitations

We have preset the buckets to be:

5 minutes
1 hour
1 day
1 week
30 days
90 days
180 days
1 year

### DIFF
--- a/exp/lighthorizon/main.go
+++ b/exp/lighthorizon/main.go
@@ -54,10 +54,10 @@ if left empty, uses a temporary directory`)
 	defer ingestArchive.Close()
 
 	Config := services.Config{
-		Archive:              ingestArchive,
-		Passphrase:           *networkPassphrase,
-		IndexStore:           indexStore,
-		ResponseAgeHistogram: services.NewResponseAgeHistogramMetric(registry),
+		Archive:    ingestArchive,
+		Passphrase: *networkPassphrase,
+		IndexStore: indexStore,
+		Metrics:    services.NewMetrics(registry),
 	}
 
 	lightHorizon := services.LightHorizon{
--- a/exp/lighthorizon/main.go
+++ b/exp/lighthorizon/main.go
@@ -5,6 +5,7 @@ import (
 	"net/http"
 
 	"github.com/prometheus/client_golang/prometheus"
+
 	"github.com/stellar/go/exp/lighthorizon/archive"
 	"github.com/stellar/go/exp/lighthorizon/index"
 	"github.com/stellar/go/exp/lighthorizon/services"
@@ -53,9 +54,10 @@ if left empty, uses a temporary directory`)
 	defer ingestArchive.Close()
 
 	Config := services.Config{
-		Archive:    ingestArchive,
-		Passphrase: *networkPassphrase,
-		IndexStore: indexStore,
+		Archive:              ingestArchive,
+		Passphrase:           *networkPassphrase,
+		IndexStore:           indexStore,
+		ResponseAgeHistogram: services.NewResponseAgeHistogramMetric(registry),
 	}
 
 	lightHorizon := services.LightHorizon{
--- a/exp/lighthorizon/services/main.go
+++ b/exp/lighthorizon/services/main.go
@@ -3,6 +3,7 @@ package services
 import (
 	"context"
 	"io"
+	"strconv"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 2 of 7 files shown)


exp/lighthorizon/services/main.go
```diff
@@ -89,12 +133,37 @@ func (os *OperationsService) GetOperationsByAccount(ctx context.Context,
...
+		}).Observe(age)
+	}
 	return ops, nil
 }
 
+func transactionsResponseAgeSeconds(txs []common.Transaction) float64 {
+	if len(txs) == 0 {
+		return -1
+	}
+	lastCloseTime := time.Unix(int64(txs[len(txs)-1].LedgerHeader.ScpValue.CloseTime), 0).UTC()
```
Shaptic, line 158: What if `order=desc`? Then it should be `txs[0]`, right? I'm not sure the best way to handle this. Maybe we only do the sort "on the way out" in handling the response? Then this code can stay the same.

exp/lighthorizon/services/main.go
```diff
@@ -114,8 +183,21 @@ func (ts *TransactionsService) GetTransactionsByAccount(ctx context.Context,
...
+		}
 		return nil, err
 	}
+
+	if age := transactionsResponseAgeSeconds(txs); age >= 0 {
+		ts.Config.ResponseAgeHistogram.With(prometheus.Labels{
+			"request":    "GetTransactionsByAccount",
+			"successful": "true",
+		}).Observe(age)
+	}
```
Shaptic, line 200: Since this code is almost identical, we could do something mildly clever with `defer` like

```suggestion
	var searchErr error
	
	defer func() {
		success := "true"
		if searchErr != nil {
			success = "false"
		}
		if age := transactionsResponseAgeSeconds(txs); age >= 0 {
			ts.Config.ResponseAgeHistogram.With(prometheus.Labels{
				"request":    "GetTransactionsByAccount",
				"successful": success,
			}).Observe(age)
		}
	}()

	return txs, searchTxByAccount(ctx, cursor, accountId, ts.Config, txsCallback)
```

Thoughts? Too convoluted?

exp/lighthorizon/services/main.go
```diff
@@ -57,6 +88,19 @@ type TransactionRepository interface {
 // processing (e.g. when a limit is reached) and any error that occurred.
 type searchCallback func(archive.LedgerTransaction, *xdr.LedgerHeader) (finished bool, err error)
 
+func operationsResponseAgeSeconds(ops []common.Operation) float64 {
+	if len(ops) == 0 {
+		return -1
+	}
+	lastCloseTime := time.Unix(int64(ops[len(ops)-1].LedgerHeader.ScpValue.CloseTime), 0).UTC()
+	now := time.Now().UTC()
+	if now.Before(lastCloseTime) {
```
Shaptic, line 97: Nice edge case!

exp/lighthorizon/main.go
```diff
@@ -53,9 +54,10 @@ if left empty, uses a temporary directory`)
 	defer ingestArchive.Close()
 
 	Config := services.Config{
-		Archive:    ingestArchive,
-		Passphrase: *networkPassphrase,
-		IndexStore: indexStore,
+		Archive:              ingestArchive,
+		Passphrase:           *networkPassphrase,
+		IndexStore:           indexStore,
+		ResponseAgeHistogram: services.NewResponseAgeHistogramMetric(registry),
```
Shaptic, line 60: Maybe overengineering too early, but what about a `Config { Metrics: Metrics { ResponseAgeHistogram: ... }` pattern? Then the `Config` doesn't need to change every time? Thoughts?

Shaptic (commented):
Some initial thoughts. Should we gather feedback on the bucket sizes or just wing it and iterate later if necessary?

tamirms:
@Shaptic I think we can always change the buckets if we want more granularity in certain ranges.

exp/lighthorizon/services/main.go
Shaptic: it's not being used yet, but we will eventually need "parameter parity" with current Horizon. you're right, though, that we can't exactly handle the case until it's present :laughing:

exp/lighthorizon/services/main.go
```diff
@@ -132,30 +147,30 @@ func (os *OperationsService) GetOperationsByAccount(ctx context.Context,
...
-		return nil, err
-	}
-
+	err := searchTxByAccount(ctx, cursor, accountId, os.Config, opsCallback)
 	if age := operationsResponseAgeSeconds(ops); age >= 0 {
-		os.Config.ResponseAgeHistogram.With(prometheus.Labels{
+		os.Config.Metrics.ResponseAgeHistogram.With(prometheus.Labels{
 			"request":    "GetOperationsByAccount",
-			"successful": "true",
+			"successful": strconv.FormatBool(err == nil),
```
Shaptic, line 154: wow, way better than my approach

Shaptic (approved):

PR #4495, opened 2022-08-01
exp/lighthorizon/index: Allow accounts to be indexed by ledger.

Shaptic:
### What
This adds `Module`s to build indices by ledger in addition to by checkpoint for account indices.

It also extends the map job to have a `MODULES` environmental variable to control which modules are built. To build without indices, you would pass `MODULES="accounts_by_ledger_unbacked"`.

It also removes building transaction indices by default.

### Why
We need to evaluate the difference in size / latency / etc. in using by-checkpoint or by-ledger indices for accounts.

### Known limitations
n/a

### DIFF
--- a/exp/lighthorizon/index/modules.go
+++ b/exp/lighthorizon/index/modules.go
@@ -13,6 +13,13 @@ var (
 	checkpointManager = historyarchive.NewCheckpointManager(0)
 )
 
+const (
+	ByCheckpoint = iota
+	ByLedger     = iota
+)
+
+type AccountIndexMode int
+
 func ProcessTransaction(
 	indexStore Store,
 	ledger xdr.LedgerCloseMeta,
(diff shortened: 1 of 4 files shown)


2opremio (approved):

exp/lighthorizon/index/modules.go
```diff
@@ -94,26 +124,20 @@ func ProcessAccountsWithoutBackend(
...
-	// 	err = indexStore.AddParticipantsToIndexesNoBackend(checkpoint, "successful/payments", paymentsParticipants)
-	// 	if err != nil {
-	// 		return err
-	// 	}
-	// }
-
 	return nil
 }
 
+// GetCheckpointNumber returns the next checkpoint NUMBER (NOT the checkpoint
```
sreuland, line 135: do you think there's value in having this in `historyarchive.CheckpointManager`, inlined here is good as well because it avoids dependency into mono repo, we're already using CheckpointManager elsewhere in lighthorizon.

sreuland (approved):
module design is paying off! nice work.

exp/lighthorizon/index/modules.go
Shaptic: Interesting idea! I'm not entirely sure -- this detail is specific to the way indexes are built. We might change the way it's represented in the future, e.g. what if we decide a different way to "bucket" them? I guess that's a really opaque notion... I'm for moving it as it stands today. I'll open a separate PR!

PR #4496, opened 2022-08-02
tools/goreplay-middleware: Add goreplay middleware

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Adds middleware for `goreplay` which checks if the mirrored response matches the original response.

### Why

`goreplay` [middleware](https://github.com/buger/goreplay/wiki/Middleware) gives access to request and responses of original and mirrored targets. This allows us to replicate `horizon-cmp` functionality but on a larger scale (ex. `horizon-cmp`  sends requests to public load balancers so normal rate limiting applies). 

### Known limitations

Currently logs mismatched response bodied to `stderr`. In the future, we can send files to S3 and build some diff checker infrastructure on top of it.

### DIFF
--- a/tools/goreplay-middleware/main.go
+++ b/tools/goreplay-middleware/main.go
@@ -1,0 +1,84 @@
+// Based on https://github.com/buger/goreplay/blob/master/examples/middleware/token_modifier.go
+package main
+
+import (
+	"bufio"
+	"bytes"
+	"encoding/hex"
+	"os"
+	"time"
+)
+
+// maxPerSecond defines how many requests should be checked at max per second
+const maxPerSecond = 1
+
+const (
+	requestType          byte = '1'
+	originalResponseType byte = '2'
+	replayedResponseType byte = '3'
+)
+
+var lastCheck = time.Now()
+var reqsCheckedPerSeq = 0
+var pendingRequests map[string]*Request
+
+func main() {
+	scanner := bufio.NewScanner(os.Stdin)
+
+	for scanner.Scan() {
+		encoded := scanner.Bytes()
+		buf := make([]byte, len(encoded)/2)
+		hex.Decode(buf, encoded)
+
+		process(buf)
+	}
+}
+
+func process(buf []byte) {
+	// First byte indicate payload type:
+	payloadType := buf[0]
+	headerSize := bytes.IndexByte(buf, '\n') + 1
+	header := buf[:headerSize-1]
+
+	// Header contains space separated values of: request type, request id, and request start time (or round-trip time for responses)
+	meta := bytes.Split(header, []byte(" "))
+	// For each request you should receive 3 payloads (request, response, replayed response) with same request id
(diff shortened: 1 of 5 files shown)


Shaptic:
I'm concerned about using `goreplay` to mirror production traffic to the k8s cluster that @sreuland is deploying Horizon Lite to. Will the hardware be able to handle it? Can we otherwise tweak the replay settings to e.g. replay 10% of requests or at least filter them on a particular endpoint? And does the main prod that's doing the mirroring care at all about the performance of the servers it's mirroring to? For example if Lite takes 30s to fulfill a mirrored request, does the initiator care at all?

bartekn:
> Can we otherwise tweak the replay settings to e.g. replay 10% of requests or at least filter them on a particular endpoint?

Both things can be done via CLI flags to goreplay command:
- [Request filtering](https://github.com/buger/goreplay/wiki/Request-filtering)
- [Rate limiting](https://github.com/buger/goreplay/wiki/Rate-limiting)

> And does the main prod that's doing the mirroring care at all about the performance of the servers it's mirroring to? For example if Lite takes 30s to fulfill a mirrored request, does the initiator care at all?

No, prod doesn't care about mirroring and it doesn't affect it at all: [How Traffic Mirroring works](https://docs.aws.amazon.com/vpc/latest/mirroring/traffic-mirroring-how-it-works.html).

tools/goreplay-middleware/main.go
```diff
@@ -0,0 +1,84 @@
...
+		}
+	case replayedResponseType:
+		if req, ok := pendingRequests[reqID]; ok {
+			delete(pendingRequests, reqID)
+			req.MirroredResponse = payload
+
+			if !req.ResponseEquals() {
+				// TODO improve the message to at least print the requested path
+				// TODO in the future publish the results to S3 for easier processing
+				os.Stderr.WriteString("MISMATCH " + req.SerializeBase64() + "\n")
```
sreuland, line 80: can these replay comparison results potentially intermix with normal horizon app log output to console, making it harder to read overall? if so, maybe it's worth optional configurable target(defaulted to console err) for goreplay results output early on?

bartekn: This messages land in completely different log: goreplay log. It's separate from Horizon log. Ideally the diffs are sent to external server but for testing purposes I wanted to start with something simple.

sreuland:
looks good, would it be possible to show a small diagram of how the replay process works for context? such as the flow from prod(source), target(mirror server), goreplay.log , and the jenkins `stellar-goreplay-service-action` job. I was trying to understand where 'rate limit' parameter from the jenkins job would flow into here, or I may have mis-understood the context.

bartekn:
@sreuland I added a short comment explaining how middleware works. @stellar/horizon-committers I think this is ready for review because it helped a lot during Horizon v2.20.0 testing. PTAL.

Issue #4468, opened 2022-07-20
exp/lighthorizon: Fetch ledgers in parallel whenever possible


Shaptic:
Parent epic: #4317.

-----------------

**TL;DR:** Concurrent `GetLedger()` downloads should be implemented wherever possible.

The handful of Horizon Lite endpoints currently process requests in a single-threaded pipeline: get the ledger, extract the information we care about, add it to the response, repeat until done.

Ledger downloads are a big portion of overall response latency, yet these downloads can be done in parallel while ledgers are read/processed. Caching helps with reducing the amount of downloads, but concurrent downloads ensure network bandwidth is saturated as much as possible and minimize processing "downtime".

There is a big open question (though instinct says "yes") about whether or not this is a worthwhile optimization. Since pubnet ledgers are fairly large, doing processing on them while the network does nothing, even though we know we'll need more downloads later, seems like it would be worth it.

This is especially important for checkpoint-based indexing, since we know for a fact that we need to look through at least 64 ledgers.

PR #4500, opened 2022-08-02
exp/lighthorizon/services: Fetch ledgers in parallel to them being processed.

Shaptic:
### What
This modifies the ledger download+processing code to download subsets of the entire checkpoint range in parallel. Preliminary testing suggests that this can **massively** reduce request latency.

Before:

> "GET http://localhost:8080/accounts/GBBJPGXTNNYBQGAIGONBFB6P7SUNO75GJEQ4Y7FVGMB2GPS3DJQVXSEP/transactions?cursor=179115963198013441&limit=10 HTTP/1.1" from 127.0.0.1:38834 - 200 199360B in **55.8414775s**

After:

> "GET http://localhost:8080/accounts/GBBJPGXTNNYBQGAIGONBFB6P7SUNO75GJEQ4Y7FVGMB2GPS3DJQVXSEP/transactions?cursor=179115963198013441&limit=10 HTTP/1.1" from 127.0.0.1:54188 - 200 199360B in **16.043121887s**

(Note that the indices are local and no on-disk cache was used.)

### Why
We can reduce latency by doing ledger downloads in parallel to processing. See #4468.

### Known limitations
Obviously, neither of these represent acceptable levels of latency if we're talking about _parity_ with classic Horizon. However, it's still a massive improvement over the previous code if we decide that high latency is acceptable for deeply historical requests.

### DIFF
--- a/exp/lighthorizon/services/cursor.go
+++ b/exp/lighthorizon/services/cursor.go
@@ -10,6 +10,7 @@ import (
 type CursorManager interface {
 	Begin(cursor int64) (int64, error)
 	Advance() (int64, error)
+	Skip(count int) (int64, error)
 }
 
 type AccountActivityCursorManager struct {
@@ -88,6 +89,17 @@ func (c *AccountActivityCursorManager) Advance() (int64, error) {
 	}
 
 	return c.lastCursor.ToInt64(), nil
+}
+
+func (c *AccountActivityCursorManager) Skip(count int) (cursor int64, err error) {
+	for i := 1; i <= count; i++ {
+		cursor, err = c.Advance()
+		if err != nil {
+			return
+		}
+	}
+
+	return
 }
 
 var _ CursorManager = (*AccountActivityCursorManager)(nil) // ensure conformity to the interface
--- a/exp/lighthorizon/services/cursor.go
+++ b/exp/lighthorizon/services/cursor.go
@@ -10,7 +10,7 @@ import (
 type CursorManager interface {
 	Begin(cursor int64) (int64, error)
 	Advance() (int64, error)
-	Skip(count int) (int64, error)
+	Skip(count uint) (int64, error)
 }
 
 type AccountActivityCursorManager struct {
@@ -91,15 +91,17 @@ func (c *AccountActivityCursorManager) Advance() (int64, error) {
 	return c.lastCursor.ToInt64(), nil
 }
 
-func (c *AccountActivityCursorManager) Skip(count int) (cursor int64, err error) {
-	for i := 1; i <= count; i++ {
+func (c *AccountActivityCursorManager) Skip(count uint) (int64, error) {
+	var err error
+	cursor := c.lastCursor.ToInt64()
+	for i := uint(1); i <= count; i++ {
 		cursor, err = c.Advance()
 		if err != nil {
-			return
+			return cursor, err
 		}
 	}
 
-	return
+	return cursor, nil
 }
 
 var _ CursorManager = (*AccountActivityCursorManager)(nil) // ensure conformity to the interface
--- a/exp/lighthorizon/services/main.go
+++ b/exp/lighthorizon/services/main.go
@@ -205,7 +205,8 @@ func incrementAverage(prevAverage *time.Duration, latest time.Duration, newCount
 // `downloadWorkerCount` of them in parallel.
 //
 // It's the caller's responsibility to ensure that all of the goroutines in the
-// returned group have completed.
+// returned group have completed. In contrast, this function closes the output
+// channel when all work has been submitted (or the context errors).
 //
 // FIXME: Should this be a part of archive.Archive?
 func downloadLedgers(
--- a/exp/lighthorizon/services/main.go
+++ b/exp/lighthorizon/services/main.go
@@ -3,22 +3,23 @@ package services
 import (
 	"context"
 	"io"
+	"sync"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
+	"golang.org/x/sync/errgroup"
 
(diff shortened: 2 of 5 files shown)


exp/lighthorizon/services/main.go
```diff
@@ -386,13 +389,15 @@ func downloadLedgers(
...
-					Debugf("Downloaded ledger %d", ledgerSeq)
 				if err != nil {
 					return err
 				}
+				log.WithField("duration", time.Since(start)).
+					Debugf("Downloaded ledger %d", ledgerSeq)
 
+				cond.L.Lock()
 				ledgerFeed[ledgerSeq-ledgerRange.Low] = &txmeta
+				cond.L.Unlock()
```
Shaptic, line 400: Note that the writes here are safe even without the lock/unlock, but it's possible that the publisher gets a partial read on the array index from an unrelated `Signal()` on a _different_ (i.e. later) index.

exp/lighthorizon/services/main.go
```diff
@@ -4,23 +4,24 @@ import (
...
-
-	"github.com/stellar/go/support/errors"
 	"github.com/stellar/go/support/log"
+	"github.com/stellar/go/xdr"
 )
 
 const (
 	allTransactionsIndex = "all/all"
 	allPaymentsIndex     = "all/payments"
+	maxParallelDownloads = 8
```
Shaptic, line 24: This is purely empirical. If this becomes viable, we'll probably want to make this a CLI option.

exp/lighthorizon/services/main_test.go
```diff
@@ -117,10 +117,10 @@ func mockArchiveAndIndex(ctx context.Context, passphrase string) (archive.Archiv
...
 
 	mockArchive.
-		On("GetLedger", ctx, uint32(1586112)).Return(expectedLedger1, nil).
-		On("GetLedger", ctx, uint32(1586113)).Return(expectedLedger2, nil).
-		On("GetLedger", ctx, uint32(1586114)).Return(expectedLedger3, nil).
-		On("GetLedger", ctx, mock.Anything).Return(xdr.LedgerCloseMeta{}, nil)
+		On("GetLedger", mock.Anything, uint32(1586112)).Return(expectedLedger1, nil).
+		On("GetLedger", mock.Anything, uint32(1586113)).Return(expectedLedger2, nil).
+		On("GetLedger", mock.Anything, uint32(1586114)).Return(expectedLedger3, nil).
+		On("GetLedger", mock.Anything, mock.Anything).Return(xdr.LedgerCloseMeta{}, nil)
```
Shaptic, line 123: Does anyone know how to do a "any context.Context instance" equivalent here? I tried `mock.AnythingOfType` but that wasn't it.

sreuland: what was the syntax, looks like `mock.AnythingOfType("context.Context")` shoulda worked?

exp/lighthorizon/services/cursor.go
```diff
@@ -10,6 +10,7 @@ import (
 type CursorManager interface {
 	Begin(cursor int64) (int64, error)
 	Advance() (int64, error)
+	Skip(count uint) (int64, error)
```
sreuland, line 13: not suggestion change, just a question on interface ergo, do you think `Advance(skip uint)` would be less moving parts for same new functionality?

exp/lighthorizon/services/main.go
```diff
@@ -269,3 +288,120 @@ func searchTxByAccount(ctx context.Context, cursor int64, accountId string, conf
...
+// downloadLedgers allows parallel downloads of ledgers via an archive. Give it
+// a ledger range and a channel to which to output ledgers, and it will feed
+// ledgers to it *in sequential order* while downloading up to
+// `downloadWorkerCount` of them in parallel.
+//
+// It's the caller's responsibility to ensure that all of the goroutines in the
+// returned group have completed. In contrast, this function closes the output
+// channel when all work has been submitted (or the context errors).
+//
+// FIXME: Should this be a part of archive.Archive?
```
sreuland, line 301: +1, to refactor this ledger loading optimization to be encapsulated in Archive instead, i.e. this caller method would largely be unchanged, still uses `Archive.GetLedger(ctx, nextLedger)`, the new perf optimizations for loading get relo'd within there ?

sreuland (approved):
looks really good, existing tests exercise this right? I think is worthwhile to follow your FIXME intuition now rather than wait, and refactor the logic into `Archive.GetLedger` , it may incur different unit test coverage..I think tests on this service mocked out Archive.

exp/lighthorizon/services/cursor.go
Shaptic: interesting idea, I think it's pretty ergo. seeing `Advance(0)` becomes a little unintuitive, but I'll try it out

exp/lighthorizon/services/main.go
Shaptic: I started working through refactoring this and realized that we actually probably don't need most of the `Archive` interface anymore after Bartek's work in #4488 which gives us a clean `MetaArchive` abstraction, which I think was a big point of what we were trying to accomplish with `Archive`.

I'll try to throw up a PR tomorrow to clean that up.

exp/lighthorizon/services/cursor.go
sreuland: ah, subtle, rather than `skip` semantics, it could be `Advance(nextOrdinalPosition uint)`  so `Advance(0)` would be no-op ?

Shaptic:
Abandoning in lieu of two separate PRs related to the comment about hiding this behind `Archive.GetLedger()`.

Issue #4493, opened 2022-08-01
Dev deployment of Horizon Light 


sreuland:
### What problem does your feature solve?
Access to a common/shared instance of Horizon Light Web server pointed at pubnet indexes and hosted on cloud infra. 
Provides a demoable instance.

### What would you like to see?
A public url that can be accessed from anywhere that has the HorizonLight web service behind it.

On Later iteration, add in CI pipeline to automate updating the HorizonLight instance when git branch changes.

### What alternatives are there?
do demos from local workstations running HorizonLight web server.

PR #4501, opened 2022-08-03
Horizon light: Dockerize indexer to run in AWS Batch

2opremio:
This PR creates a docker image (stellar/horizon-indexer) which works in a similar fashion to stellar/horizon-verify-rage
and is tested and pushed as part of the Horizon GitHub workflow.


services/horizon/docker/indexer/Dockerfile
```diff
@@ -0,0 +1,27 @@
+FROM golang:1.18 AS builder
```
sreuland, line 1: what is the basis for locating this in the services/horizon/docker path, I'm currently developing in same space for dockering up light images and was thinking of the docker/k8s files being closer to lighthorizon like `/exp/lighthorizon/build` or something?

2opremio: I think you have a point. Can we move them in a separate PR? (Since we also need to move the ledgerexporter files)

sreuland: yes, I can do that in my upcoming PR for deployment of this on k8s, here's [compare of my dev branch I have in progress that uses exp/lighthorizon/builld ](https://github.com/stellar/go/compare/lighthorizon...sreuland:go:4493_deploy_lighthorizon_web) 

once you merge this as-is, I'll pull that into my 4493 dev, and refactor this and ledger exporter over to exp/lighthorizon/build, sound good?

sreuland (approved):

exp/lighthorizon/index/types/bitmap.go
```diff
@@ -8,8 +8,6 @@ import (
 	"github.com/stellar/go/xdr"
 )
 
-const CheckpointIndexVersion = 1
```
Shaptic, line 11: Is this not being used? Must've missed that early on. Should we modify the serialization code to include this in the on-disk format in this PR? It will make it much easier to upgrade down the line. Or maybe it doesn't matter because the range for the MVP is small anyway.

services/horizon/docker/indexer/README.md
```diff
@@ -0,0 +1,42 @@
+# `stellar/horizon-ledgerexporter`
```
Shaptic, line 1: This needs to be updated, right?

services/horizon/docker/indexer/start
```diff
@@ -0,0 +1,47 @@
+#! /usr/bin/env bash
+set -e
+
+START="${START:=0}"
+END="${END:=0}"
+CONTINUE="${CONTINUE:=false}"
+# Writing to /latest is disabled by default to avoid race conditions between parallel container runs
+WRITE_LATEST_PATH="${WRITE_LATEST_PATH:=false}"
```
Shaptic, line 8: Same here, right? I'm not sure how the parameters are getting passed through.

Shaptic (changes requested):

services/horizon/docker/indexer/start
sreuland: @Shaptic @2opremio , I wonder if it may be easier on us If we hold on this PR, and I take the net intent of it for index and roll it into the [dev branch for #4493  i have in in process](https://github.com/stellar/go/compare/lighthorizon...sreuland:go:4493_deploy_lighthorizon_web), which is dockering index and web processes from horizonlight and defining k8s for them? was thinking to also move services/horizon/docker/ledgerexporter into there as well.

Shaptic: We definitely need a map-reduce-specific Docker image for index building so that we can build indices on AWS Batch :arrow_right: S3.

The "real-time" or "watcher" indexer can be a separate image and deployed on k8s like you're saying.

Maybe we can just scope this one to the map/reduce part and ignore `single`?

sreuland: ah, yes, that break-out helps, so, there are two index deployment modes, `map-reduce` and `real-time` and we can align docker images against that mode.

For `map-reduce` , it needs to run the two index commands `map` and `reduce` in some arrangement of o/s processes in one container or across multiple depends on the orchestration environment, in either of those cases it could be a single docker image having both index `map` and `reduce` installed and env vars determine which is executed?

and then for 'real-time' we build another image with just the index 'single' cmd installed and env variables to config it's exec. sounds good, I can focus on dev for the `real-time` in #4493, and then won't overlap/collide with scope of this PR for the 'map-reduce'. 

thanks for clarification!

2opremio: The start file was wrongly copied. It's not used.

services/horizon/docker/indexer/README.md
2opremio: True!

services/horizon/docker/indexer/start
2opremio: @sreuland why can't you reuse the indexer image?

exp/lighthorizon/index/types/bitmap.go
2opremio: It was unused, yeah. I think we should include the version on the XDR like we did for txmeta.

If we want something to show by EOW we should fix this later.

services/horizon/docker/indexer/start
sreuland: no tech reason, there seemed to be some separation/encapsulation above at orchestration layer, ie.a docker image per higher level mode/function which provides ENTRYPOINT through a `start` script that adapts env variables and knows how to invoke the lower level index commands, rather than the orchestration layer needing to know details on how to invoking each index cmd directly.

sreuland: @2opremio , I tried this out in the [PR for the web docker](https://github.com/stellar/go/pull/4508), I put all the docker for lighthorizon under `exp/lighthorizon/build`, I think `exp/lighthorizon/build/build.sh index-batch stellar` would produce the same docker image with bins as here, moved ledgerexporter into there also. lmk if it looks like a good direction, maybe we can combine efforts.

.github/workflows/horizon.yml
```diff
@@ -168,11 +168,11 @@ jobs:
           for I in $(seq -f "%.0f" 10000064 10000127); do echo $I; curl --fail -s https://horizon-ledgermeta-prodnet-test.s3.amazonaws.com/ledgers/$I -o ledgermeta/ledgers/$I ; done
           
           # run map job
-          docker run -e NETWORK_PASSPHRASE='pubnet' -e AWS_BATCH_JOB_ARRAY_INDEX=0 -e BATCH_SIZE=64 -e FIRST_CHECKPOINT=10000064 \
+          docker run -e NETWORK_PASSPHRASE='pubnet' -e JOB_INDEX_ENV=AWS_BATCH_JOB_ARRAY_INDEX -e AWS_BATCH_JOB_ARRAY_INDEX=0 -e BATCH_SIZE=64 -e FIRST_CHECKPOINT=10000064 \
                      -e WORKER_COUNT=1 -v $PWD/ledgermeta/:/ledgermeta -e TXMETA_SOURCE=file:///ledgermeta -v $PWD/index:/index -e INDEX_TARGET=file:///index stellar/horizon-indexer map
 
           # run reduce job
-          docker run -e NETWORK_PASSPHRASE='pubnet' -e AWS_BATCH_JOB_ARRAY_INDEX=0 -e MAP_JOB_COUNT=1 -e REDUCE_JOB_COUNT=1 \
+          docker run -e NETWORK_PASSPHRASE='pubnet' -e JOB_INDEX_ENV=AWS_BATCH_JOB_ARRAY_INDEX -e AWS_BATCH_JOB_ARRAY_INDEX=0 -e MAP_JOB_COUNT=1 -e REDUCE_JOB_COUNT=1 \
```
Shaptic, line 175: Why not `JOB_INDEX=$AWS_BATCH_JOB_ARRAY_INDEX`? Then it's just passing the value, rather than the name of a name to look up heh.

2opremio: That assumes that the batch system we use will replace env variables when setting them, which I don't even know is the case for batch

sreuland:
@2opremio , since #4519 merged which refactored the docker images under exp/lighthorizon/build, I grabbed the net changes from this PR based on current lighthorizon branch and captured them in new PR #4543, I think that PR can be used  and deprecate the need to fix conflicts&merge this PR, lmk what you think, thanks!

2opremio:
@sreuland sounds good! Thanks for the refactoring work!!!

sreuland:
confirmed with @2opremio, closing this PR in lieu of PR https://github.com/stellar/go/pull/4543 which carries the net changes from this on top of newly refactored docker image setup under lighthorizon.

PR #4507, opened 2022-08-04
exp/lighhorizon: Make indexing s3 bucket configurable

### LINKED ISSUE #4504: exp/lighthorizon: Make s3 bucket of indexer backend configurable
Right now the bucket name is hardcoded to `const BUCKET = "horizon-index"`

2opremio:
Closes #4504 

Also, 

* Renames Url to URL
* stops the abuse of `StoreConfig.URL` which I don't think should be overwritten


Shaptic (approved):
Nice, cleaner :+1:

PR #4509, opened 2022-08-05
exp/lighthorizon: Add JSON content type to responses.

Shaptic:
### What
Adds JSON `Content-Type` header to responses to render correctly.

### Why
self-explanatory

### DIFF
--- a/exp/lighthorizon/actions/main.go
+++ b/exp/lighthorizon/actions/main.go
@@ -61,6 +61,8 @@ func sendPageResponse(w http.ResponseWriter, page hal.Page) {
 	if err != nil {
 		log.Error(err)
 		sendErrorResponse(w, http.StatusInternalServerError, "")
+	} else {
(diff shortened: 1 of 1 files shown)


exp/lighthorizon/actions/main.go
```diff
@@ -61,6 +61,8 @@ func sendPageResponse(w http.ResponseWriter, page hal.Page) {
 	if err != nil {
 		log.Error(err)
 		sendErrorResponse(w, http.StatusInternalServerError, "")
+	} else {
+		w.Header().Set("Content-Type", "application/openapi+yaml")
```
bartekn, line 65: Shouldn't it be the same as Horizon, so `application/hal+json; charset=utf-8`?

Shaptic: WOW how did I mess this up...? copy-paste is evil

bartekn (approved):

PR #4510, opened 2022-08-05
exp/lighthorizon, xdr: Rename `CheckpointIndex` to better reflect its capabilty.

Shaptic:
### What
Rename `{types,xdr}.CheckpointIndex` -> `{types,xdr}.BitmapIndex`.  
Rename `NextActive` -> `NextActiveBit`.

### Why
This is a better reflection of the capability of the index: it can operate on any value, not necessarily just checkpoints (this is further evidenced by using it to index by ledger).

### Known limitations
n/a

### DIFF
--- a/exp/lighthorizon/index/types/bitmap.go
+++ b/exp/lighthorizon/index/types/bitmap.go
@@ -8,7 +8,7 @@ import (
 	"github.com/stellar/go/xdr"
 )
 
-const CheckpointIndexVersion = 1
+const BitmapIndexVersion = 1
 
 type BitmapIndex struct {
 	mutex    sync.RWMutex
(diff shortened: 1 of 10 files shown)


exp/lighthorizon/index/types/bitmap.go
```diff
@@ -29,83 +29,83 @@ func NewCheckpointIndex(b []byte) (*CheckpointIndex, error) {
 	return NewCheckpointIndexFromXDR(xdrCheckpoint), nil
 }
 
-func NewCheckpointIndexFromXDR(index xdr.CheckpointIndex) *CheckpointIndex {
-	return &CheckpointIndex{
-		bitmap:          index.Bitmap[:],
-		firstCheckpoint: uint32(index.FirstCheckpoint),
-		lastCheckpoint:  uint32(index.LastCheckpoint),
+func NewCheckpointIndexFromXDR(index xdr.BitmapIndex) *BitmapIndex {
```
sreuland, line 32: to be consistent on naming, would this be better as `NewBitmapIndexFromXDR` ?

exp/lighthorizon/index/types/bitmap.go
```diff
@@ -151,55 +151,55 @@ func (i *CheckpointIndex) iterate(f func(checkpoint uint32)) error {
...
 		}
-		err = i.setActive(checkpoint)
+		err = i.setActive(index)
 	})
 
 	return err
 }
 
-// NextActive returns the next checkpoint (inclusive) where this index is
+// NextActiveBit returns the next checkpoint (inclusive) where this index is
```
sreuland, line 169: adjust docs a bit per rename, like:
```
Returns the next active index location (inclusive). 
If index is active at 'index' then it returns same 'index'.
```

sreuland (approved):
looks good, left minor suggestions.

exp/lighthorizon/index/types/bitmap.go
Shaptic: good catch! ctrl+f isn't perfect, heh. done in cf37fe77.

exp/lighthorizon/index/types/bitmap.go
Shaptic: yep, great call! done in cf37fe77.

PR #4513, opened 2022-08-05
exp/lighthorizon: *Correctly* set `Content-Type`, plus JSONify errors

Shaptic:
### What
This sets the content-type at the right time and also makes error messages proper JSON.

### Why
The headers can't be set after `Write()` is called on `w`.


sreuland (approved):
looks good, at some point later, we should start to back fill in some http level tests to validate json, marshaling.

Shaptic:
@sreuland yess absolutely, the tech debt is building up :sob:

PR #4518, opened 2022-08-07
services/horizon: Reap history object tables when ingestion is idle

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

### Why

### Known limitations

[TODO or N/A]


services/horizon/internal/db2/history/main.go
```diff
@@ -825,6 +826,145 @@ func (q *Q) CloneIngestionQ() IngestionQ {
...
+	}
+
+	for i := range historyTables {
+		_, err = fmt.Fprintf(&sb, "c%d = 0 and ", i)
+		if err != nil {
+			return "", err
+		}
+	}
+
+	_, err = sb.WriteString("1=1);")
```
Shaptic, line 960: Can you explain why this last clause is necessary? :confused:

services/horizon/internal/db2/history/main.go
```diff
@@ -825,6 +826,145 @@ func (q *Q) CloneIngestionQ() IngestionQ {
...
+// 		(select count(*) from history_transaction_claimable_balances
+// 		 where history_claimable_balance_id = hcb.id) as c2,
+//		1 as cx,
+//      from history_claimable_balances hcb order by id limit 100 offset 1000)
+//  as sub where c1 = 0 and c2 = 0 and 1=1);
+//
+// In short it checks the 100 rows omiting 1000 row of history_claimable_balances
+// and counts occurences of each row in corresponding history tables.
+// If there are no history rows for a given id, the row in
+// history_claimable_balances is removed.
```
Shaptic, line 920: This is an interesting way to determine "age." It creates a dependence between the history tables, right? Is there a reason we don't rely on ledger close time, instead? I guess probably because `history_claimable_balances` doesn't have that row :disappointed:

services/horizon/internal/ingest/main.go
```diff
@@ -629,6 +641,7 @@ func (s *system) maybeVerifyState(lastIngestedLedger uint32) {
 	stateInvalid, err := s.historyQ.GetExpStateInvalid(s.ctx)
 	if err != nil && !isCancelledError(err) {
 		log.WithField("err", err).Error("Error getting state invalid value")
+		return
```
Shaptic, line 644: This is new behavior, right? Just want to make sure it's intended. Otherwise I guess state verify would run even if there was an error getting state?

services/horizon/internal/db2/history/main.go
bartekn: I'm using `strings.Builder` to build the query. There are two loops: first loop adds the count of occurrences of each row in lookup tables and the second loop adds `where` conditions that only matches the rows whose all counters are zero. In the first loop, each _piece_ of query ends with `,` and in the second loop each condition ends with `and`. `select` statement syntax disallows dangling `and` and `,`. So I had two options:
* Add some special handling of last slice element to not add `and` and `,` in such case. Or remove them after adding them (not possible with `strings.Builder` because it only allows appends).
* Add some neutral condition to dangling `and` (i.e. which always resolves to `true` so for example `1=1`) and some field to fields list which I will simply ignore (`1 as cx`).

I decided to go with the second option because it's less coding.

services/horizon/internal/db2/history/main.go
bartekn: It's not really about age. The rows are sorted by `id` which is just a sequence integer value assigned to specific ledger object (like claimable balance). The `limit ... offset` listing here is just to ensure we iterate over entire table in multiple cycles.

services/horizon/internal/ingest/main.go
bartekn: Yes, I forgot about it. It's just a missing `return` I notices (if there's an error, this method should not continue). I will extract it to a separate PR.

bartekn: https://github.com/stellar/go/pull/4523

bartekn:
I'm going to merge this without 👍 to start testing a new release. Please add comments and I'll open another PR with review fixes/requests.

services/horizon/internal/db2/history/main.go
Shaptic: Brilliant, I like it!

PR #4522, opened 2022-08-08
exp/lighthorizon: Add a suite of tools to manage the on-disk ledger cache.

Shaptic:
### What
Adds a set of tools to manage the on-disk cache:

 - `cache build`: caches a ledger range on-disk w/o running Horizon Lite
 - `cache purge`: purges ledger ranges from the disk
 - `cache show`: shows which ledger ranges are cached on-disk

Output of `cache --help`, for more details:

```
Manages the on-disk cache of ledgers.

Usage:
  cache [flags]
  cache [command]

Examples:

cache build --start 1234 --count 1000 s3://txmeta /tmp/example
cache purge /tmp/example 1234 1300
cache show /tmp/example

Available Commands:
  build       
  purge       
  show        

Use "cache [command] --help" for more information about a command.
```

### Why
Caching is important for performance, but by default, Horizon Lite only caches ledgers that get _used_. It's very useful to be able to cache ledgers _before_ they are needed, for example to have a cache of the last day of ledgers before you run Horizon Lite.

This can be viewed as a parallel to `horizon db reingest range` et. al.

### Known limitations
The arg parsing code isn't ideal - any cobra-fu would be appreciated!

### DIFF
--- a/exp/lighthorizon/tools/cache.go
+++ b/exp/lighthorizon/tools/cache.go
@@ -1,0 +1,273 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/spf13/cobra"
+	"github.com/stellar/go/metaarchive"
+	"github.com/stellar/go/support/log"
+	"github.com/stellar/go/support/storage"
+)
+
+var (
+	s3Region = "us-east-1"
+)
+
+const (
+	defaultCacheCount = (60 * 60 * 24) / 5 // ~24hrs worth of ledgers
+)
+
+func main() {
+	log.SetLevel(log.InfoLevel)
+
+	cmd := &cobra.Command{
+		Use:  "cache",
+		Long: "Manages the on-disk cache of ledgers.",
+		Example: `
+cache build --start 1234 --count 1000 s3://txmeta /tmp/example
+cache purge /tmp/example 1234 1300
+cache show /tmp/example`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			// require a subcommand for now - eventually this will live under
+			// the lighthorizon command
+			return cmd.Help()
+		},
+	}
+	purge := &cobra.Command{
+		Use:  "purge [flags] path <start> <end>",
+		Long: "Purges individual ledgers (or ranges) from the cache, or the entire cache.",
+		Example: `
+purge /tmp/example              # empty the whole cache
+purge /tmp/example 1000         # purge one ledger
+purge /tmp/example 1000 1005    # purge a ledger range`,
+		RunE: func(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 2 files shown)


exp/lighthorizon/tools/cache.go
```diff
@@ -0,0 +1,273 @@
...
+				cmd.Println("--count should be a positive integer")
+				return cmd.Usage()
+			}
+			s3Region, _ = cmd.Flags().GetString("region")
+			repair, _ := cmd.Flags().GetBool("repair")
+			return BuildCache(args[0], args[1], start, count, repair)
+		},
+	}
+
+	build.Flags().Bool("repair", false, "attempt to purge the cache and retry ledgers that error")
```
sreuland, line 122: I don't think you need to have region param, can pass empty right, the aws sdk will pull it and creds from a sequence of o/s env sources/variables for profiles, ~/.aws files etc

sreuland:
very nice addition, I'm  wondering for purposes of service deployment of the web, where this user tool couldn't be used b/c the cache folder is private within the deployment, maybe we can add similar via some new flags like `preload-cache`?

exp/lighthorizon/tools/cache.go
Shaptic: oooh yes, can use e.g. `AWS_REGION ./cache build [...]`. great call! didn't realize this. e80f671d.

Shaptic:
@sreuland why do you say it wouldn't be usable by the deployment? Couldn't it run `cache build [...]` as a step before running `lighthorizon [...]` forever?

sreuland:
> @sreuland why do you say it wouldn't be usable by the deployment? Couldn't it run `cache build [...]` as a step before running `lighthorizon [...]` forever?

oh yeah, the web could automate internally the usage of `BuildCache()` or maybe higher up at the cmd, kicking it off from separate go routine in a loop. thinking of that, is it worth encapsulating that infinite loop behavior within `BuildCache(... , continue bool)` in here? 

Per the web process as client, maybe we would need to include a toggle flag like `preload-cache=true|false` or just do it always?

Shaptic:
Sorry let me clarify, I think we're mixing concepts here @sreuland: this PR introduces a separate command that you can run from the command-line, e.g. 

```
cache build --start <ledger> s3://horizon-txmeta /tmp/cache
```

which you can then use when running light horizon, e.g. 

```
lighthorizon -ledger-cache /tmp/cache  [... other parameters ...]
```

I'm wondering why that would be a problematic flow for the k8s deployment. There would be a step to build the cache (which is a no-op if the ledgers are already cached), then another step to run the server.

sreuland:
> 
> I'm wondering why that would be a problematic flow for the k8s deployment. There would be a step to build the cache (which is a no-op if the ledgers are already cached), then another step to run the server.

@Shaptic , thanks for huddle on slack, understand now the ways to use the cache tooling here from web.

sreuland (approved):

PR #4525, opened 2022-08-09
services/horizon: Fix ReapLookupTables query

bartekn:
This commit fixes the query added in ee063a7. The previous query was using `limit ... offset ...` which becomes slow for larger `offset` values. This is happening because (from docs[1]):
> The rows skipped by an OFFSET clause still have to be computed inside the server; therefore a large OFFSET might be inefficient.

I rewrote the query to use `id` field in tables in `where id > ...`. Before each run, the new `id` offset is fetched from the table that is later returned to be used in the next cycle.

I also realized that rewriting `count(*)` in subqueries to `1` with `limit 1` further improves the query performance. This is because instead of counting all rows in parent tables it just check if there are _any_ rows in tables (what we need to determine if the row is orphaned or not). This allows running the reaper also for `history_accounts` and `history_assets` tables.

[1] https://www.postgresql.org/docs/current/queries-limit.html


services/horizon/internal/ingest/main.go
```diff
@@ -721,6 +721,8 @@ func (s *system) maybeReapLookupTables(lastIngestedLedger uint32) {
 	}
 
 	s.reapOffsets = newOffsets
+	// Remove before merging
+	log.WithField("offsets", newOffsets).Info("New offsets")
```
2opremio, line 725: removal reminder

services/horizon/internal/db2/history/main.go
```diff
@@ -935,7 +957,7 @@ func constructReapLookupTablesQuery(table string, historyTables []tableObjectFie
 	for i, historyTable := range historyTables {
 		_, err = fmt.Fprintf(
 			&sb,
-			`(select count(*) from %s where %s = hcb.id) as c%d, `,
+			`(select 1 from %s where %s = hcb.id limit 1) as c%d, `,
```
Shaptic, line 960: very nice optimization :+1:

services/horizon/internal/db2/history/main_test.go
```diff
@@ -102,11 +102,11 @@ func TestConstructReapLookupTablesQuery(t *testing.T) {
...
-			"(select count(*) from history_trades where counter_account_id = hcb.id) as c3, "+
-			"(select count(*) from history_transaction_participants where history_account_id = hcb.id) as c4, "+
-			"1 as cx from history_accounts hcb order by id limit 10 offset 0) as sub "+
-			"where c0 = 0 and c1 = 0 and c2 = 0 and c3 = 0 and c4 = 0 and 1=1);", query)
+			"(select id, (select 1 from history_effects where history_account_id = hcb.id limit 1) as c0, "+
+			"(select 1 from history_operation_participants where history_account_id = hcb.id limit 1) as c1, "+
+			"(select 1 from history_trades where base_account_id = hcb.id limit 1) as c2, "+
+			"(select 1 from history_trades where counter_account_id = hcb.id limit 1) as c3, "+
+			"(select 1 from history_transaction_participants where history_account_id = hcb.id limit 1) as c4, "+
+			"1 as cx from history_accounts hcb where id >= 0 order by id limit 10) as sub "+
```
Shaptic, line 110: are negative ids possible?

Shaptic (commented):
I have a general question about the way reaping works, as that isn't part of the codebase I have had much exposure to yet.

The docstring on `ReapLookupTables` states that it

> removes rows from lookup tables which aren't used (orphaned), i.e. history entries for them were reaped

To me, the bigger picture of reaping entries has to do with ledger advancement, right? i.e. you set `HISTORY_RETENTION_COUNT=x` means data from `x` ledgers get kept around. If every table either (a) tracks the ledger a row applies to or (b) can somehow be joined to identify said ledger, isn't reaping then just a matter of `DELETE FROM table WHERE ledgerSeq < latestIngestedLedgerSeq;` for every table?

In a related vein, it'd be nice if this PR clarified what `offsets` means within the docstring of `ReapLookupTables`!

bartekn:
> To me, the bigger picture of reaping entries has to do with ledger advancement, right? i.e. you set HISTORY_RETENTION_COUNT=x means data from x ledgers get kept around. If every table either (a) tracks the ledger a row applies to or (b) can somehow be joined to identify said ledger, isn't reaping then just a matter of DELETE FROM table WHERE ledgerSeq < latestIngestedLedgerSeq; for every table?

This is exactly what `services/horizon/internal/reap` does: it removes all rows from history tables which contains transactions, operations, effects, etc. before a specified ledger (current sequence — retention count). However, we also have lookup tables that denormalize history tables to save space. They look like `[id, name]` where `name` is account ID in `history_accounts` or claimable balance ID in `history_claimable_balances`. And then when used in history tables like `history_operation_participants` we can use an integer `id` instead of the long account ID.

The obvious problem with lookup tables is that a given ID can be used in multiple ledger ranges (depends on activity of an account or claimable balance). So we can't just simply remove rows from lookup tables while removing rows from _normal_ history tables. Second, lookup tables are actively used by ingestion so if reaper removes a lookup row after ingestion loaded it it can cause inconsistencies in the DB.

services/horizon/internal/db2/history/main_test.go
bartekn: No, the `0` in `id >=0` condition is the `offset` param and it's there just to ensure we move forward when iterating over the tables in batches. So for next cycle it will be `id >=1000` if the ID after 1000 rows (batch size) is equal 1000.

paulbellamy (approved):

services/horizon/internal/db2/history/main_test.go
paulbellamy: Nit: Could probably clean this up with `WITH` clauses, and named sub-queries?

bartekn:
@stellar/horizon-committers I added two more things, PTAL:
* [8c2e40c](https://github.com/stellar/go/pull/4525/commits/8c2e40cccd0d79cf515cdc1fa3b7b9f8823c677e) Logging deleted rows counter (only when any rows removed).
* [0317d5a](https://github.com/stellar/go/pull/4525/commits/0317d5a4902295e3bc25728f2b4dac7be51803da) Activates lookup reaping feature only when `--history-retention-count` is set.

Shaptic (approved):

PR #4530, opened 2022-08-10
all: Add a generic 'complete' job for PR pass criteria flexibility.

Shaptic:
### What
This adds a `go-complete` (prefixed in case we want to add e.g. a `horizon-complete` in the future) job that encompasses all of the Go-related checks into one job.

### Why
It makes it easier to control which jobs are "required" for a PR to be merged. Additional context from @leighmcculloch is [here](https://github.com/stellar/terraform/pull/892#discussion_r942737090).

### Known limitations
n/a

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -6,6 +6,13 @@ on:
   pull_request:
 
 jobs:
+  go-complete:
+    if: always()
+    needs: [check, build, test]
+    runs-on: ubuntu-latest
+    steps:
+    - if: contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled')
+      run: exit 1
 
   check:
     strategy:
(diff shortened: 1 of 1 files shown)


leighmcculloch (approved):

.github/workflows/go.yml
```diff
@@ -6,6 +6,13 @@ on:
   pull_request:
 
 jobs:
+  go-complete:
```
leighmcculloch, line 9: I would name this simply `complete`, so that if there are other workflows with a `complete` job, they will all be required.

Shaptic: Ohhhh very interesting - my impression was that we couldn't have conflicting names for some reason. Brilliant, fixed.

leighmcculloch (approved):

PR #4531, opened 2022-08-10
exp/lighthorizon: Properly transform transactions into JSON.

### LINKED ISSUE #4527: exp/lighthorizon: Properly serialize transaction records to JSON.
Horizon Lite Epic: #4317.

----------------

We need the transaction response objects in Light Horizon to mirror the existing ones in order to effectively perform comparisons between them.

This involves modifying `./adapters/transaction.go` to mirror the way `services/horizon/internal/resourceadapter/transaction.go::PopulateTransaction` translates the data into JSON.

Shaptic:
### What
Properly transforms every field in a transaction record from XDR to JSON.

### Why
Completeness and parity w/ Horizon Classic, closes #4527.

### Known limitations
There's still a piece missing:

```golang
// TODO: Figure out what this means? Maybe @tamirms knows.
// if transactionHash != row.TransactionHash {
// 	dest.Signatures = dest.InnerTransaction.Signatures
// }
```

Any thoughts on this @tamirms? It comes from Horizon's code. I'm not sure what this check is, well, checking for...


exp/lighthorizon/adapters/transaction_test.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/stellar/go/toid"
+	"github.com/stellar/go/xdr"
+)
+
+// TestTransactionAdapter confirms that the adapter correctly serializes a
+// transaction to JSON by actually pulling a transaction from the
+// known-to-be-true horizon.stellar.org, turning it into an "ingested"
+// transaction, and serializing it.
+//
+// TODO: Instead of making an unreliable network request, we should just create
```
sreuland, line 26: yes, would saving the json response from this request to a file, and then having the test load the file instead of fetch url suffice?

Shaptic: I actually was thinking about something else (storing raw an envelope XDR and hard-coding a bunch of fields), but this is way better! Done in latest push.

exp/lighthorizon/adapters/fixtures/transactions.json
```diff
@@ -0,0 +1,67 @@
+{
```
tamirms, line 1: I think typically we put these files in a testdata directory

https://eli.thegreenplace.net/2022/file-driven-testing-in-go/

tamirms (approved):

PR #4532, opened 2022-08-10
all: Add some generics and incorporate them into the codebase.

Shaptic:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Adds and incorporates a generic `Min`/`Max` rather than customs ones per-type and per-package.

### Why
blessed generics

### Known limitations
There might be other places we're doing this inline that are much harder to find, e.g.

```go
if a < b {
    a = b
}
```

one-offs that didn't warrant a dedicated `min` for that package.


support/generics/math.go
```diff
@@ -0,0 +1,21 @@
+package generics
```
leighmcculloch, line 1: `generics` is a reasonable generic name. Given that this package contains functions that work with ordered types, how about we named it `ordered`?

Shaptic: The idea was to expand it over time to include other helpers (as you can tell from my ambitious branch naming). Min/Max are the most obvious offenders, but now that I've added and incorporated a `Set[T any]` container, would you reconsider the rename? :wink:

Shaptic: I also plan to add both

```golang
func mapKeys[T any, U any](m map[T]U) []T      // and
func mapValues[T any, U any](m map[T]U) []U
```

once this gets merged into the `lighthorizon` branch.

leighmcculloch: I'd put the map functions into a `collections/map` package, and the `Set` type into a `collections/set` package. i.e.
```
ordered
collections/set
collections/map
```

leighmcculloch: This is a small detail and I defer to you. It's great we can do these things now 🎉.

leighmcculloch (approved):
👏🏻

support/generics/math.go
Shaptic: I guess my goal was to keep everything in its own subfolder (with no _explicit_ reason besides it being a 1.18+ feature and the general satisfaction of keeping a PR in a single folder), but I do like these names! Done in 4d58a361.

PR #4537, opened 2022-08-11
Incorporate generics into Light Horizon code

Shaptic:
### What
Incorporate `set.Set[T]` into the codebase and introduce a new `Keys` for `map`s.

### Why
Forcing users to use `map[string]struct{}` as a set structure in 2022 is absurd.

### Known limitations
None, generics are amazing.

### DIFF
--- a/exp/lighthorizon/index/backend/file.go
+++ b/exp/lighthorizon/index/backend/file.go
@@ -10,6 +10,7 @@ import (
 
 	types "github.com/stellar/go/exp/lighthorizon/index/types"
 
+	"github.com/stellar/go/support/collections/set"
 	"github.com/stellar/go/support/errors"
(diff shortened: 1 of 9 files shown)


exp/lighthorizon/index/backend/file.go
```diff
@@ -173,8 +174,8 @@ func (s *FileBackend) ReadAccounts() ([]string, error) {
 
 		// The account list is very unlikely to be unique (especially if it was made
 		// w/ parallel flushes), so let's ensure that that's the case.
-		if _, ok := accountMap[account]; !ok {
-			accountMap[account] = struct{}{}
+		if !accountMap.Contains(account) {
```
sreuland, line 177: nice, fluent.

sreuland (approved):

PR #4542, opened 2022-08-13
exp/lighthorizon/actions: use standard Problem model on API error responses

sreuland:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Apply the standard Problem json model on API error responses. 

### Why

to be consistent with the same Problem json model that classic horizon applies on API error responses for cmp tool usage.


exp/lighthorizon/actions/accounts_test.go
```diff
@@ -0,0 +1,151 @@
...
+	pathParams := make(map[string]string)
+	pathParams["account_id"] = "G1234"
+	request := makeRequest(
+		t,
+		map[string]string{},
+		pathParams,
+	)
+
+	mockOperationService := &services.MockOperationService{}
+	mockTransactionService := &services.MockTransactionService{}
```
Shaptic, line 63: By using mocking here, are we omitting testing of the actual behavior in relation to transaction querying? e.g. I don't see tests for 404s on a missing account. And a lot of these errors actually happen before `TransactionService.GetTransactionsByAccount()` gets invoked, right? Parameter validation, etc. happens beforehand. I think we'd get better coverage if we mock exactly what needs to be mocked (e.g. 500s).

sreuland: I think that is what is mocked here, have told the data access function to throw an error `not good` which represents unrecoverable/unexpected err from code, equates to server-side 500 response. 

Haven't added tests on 404's because the app layer hasn't determined that yet, it's returning empty list responses when the index reports no `NextActive()`, maybe because absence of the account is not as deterministic here since we don't have current state, just what exists after some range is indexed? maybe we just need to be consistent and choose whether to return 404's or empty list, I think with a stateless cursor API, the empty list response may feel more natural. since client would get data for each cursor fetch iteration, until the last cursor fetch would give empty list instead of 404.

exp/lighthorizon/actions/problem.go
```diff
@@ -0,0 +1,93 @@
+package actions
```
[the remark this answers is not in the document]
sreuland: yes, to avoid dependency on classic horizon package

exp/lighthorizon/actions/accounts_test.go
```diff
@@ -0,0 +1,151 @@
...
+	assert.Equal(t, http.StatusBadRequest, resp.StatusCode)
+
+	raw, err := ioutil.ReadAll(resp.Body)
+	assert.NoError(t, err)
+
+	var problem problem.P
+	err = json.Unmarshal(raw, &problem)
+	assert.NoError(t, err)
+	assert.Equal(t, "Bad Request", problem.Title)
+	assert.Equal(t, "invalid request parameters", problem.Extras["invalid_field"])
```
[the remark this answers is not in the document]
sreuland: adjusted - https://github.com/stellar/go/pull/4542/commits/8c9f81112b1416f24b3b3db001e6132dbbec88e1

exp/lighthorizon/actions/problem.go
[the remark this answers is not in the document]
Shaptic: I see it now, makes sense. Can you add a comment here about where it comes from? (`services/horizon/internal/render/problem/problem.go`) to make it easier to relate them in the future and see divergences

exp/lighthorizon/actions/accounts_test.go
Shaptic: Sorry, I put this comment in the wrong place. I meant for it to apply to the test in `TestTxByAccountMissingParamError`, but I don't think it applies there anymore, either, thinking about it more. Since the mocked code (`TransactionService`) never gets reached before hitting the 400 error like I said, it doesn't matter if it's mocked out or not.

Re: 404s, we can still distinguish between `account_id` not being in the index (which would be a 404) vs. `NextActive()` returning EOF (which would be an empty tx list). I see what you mean about a partial view of the world, but it still counts as a 404 imo: if the ledger range you indexed doesn't have any activity for an account, a 404 is a valid description of the account.

exp/lighthorizon/actions/helpers_test.go
```diff
@@ -0,0 +1,33 @@
+package actions
```
Shaptic, line 1: file naming: maybe `test_helpers.go`? since there are no tests here. I expected to see tests for helpers heh. tiny nitpick though so feel free to ignore.

exp/lighthorizon/actions/helpers_test.go
```diff
@@ -0,0 +1,33 @@
...
+
+func makeRequest(
+	t *testing.T,
+	queryParams map[string]string,
+	routeParams map[string]string,
+) *http.Request {
+	request, err := http.NewRequest("GET", "/", nil)
+	if err != nil {
+		t.Fatalf("unexpected error %v", err)
+	}
```
Shaptic, line 20: might as well use `require` here for brevity
```suggestion
	request, err := http.NewRequest("GET", "/", nil)
	require.NoError(t, err)
```

exp/lighthorizon/http_test.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/stellar/go/exp/lighthorizon/services"
+	"github.com/stellar/go/support/render/problem"
+)
+
+func TestUnknownUrl(t *testing.T) {
+	recorder := httptest.NewRecorder()
+	request, err := http.NewRequest("GET", "/unknown", nil)
+	if err != nil {
+		t.Fatalf("unexpected error %v", err)
+	}
```
Shaptic, line 22: similar comment here about `require`, in fact you can probably use `require` over `assert` in all of the tests (unless you're sure you want to continue the test in case of an error)
```suggestion
	request, err := http.NewRequest("GET", "/unknown", nil)
	require.NoError(t, err)
```

(`require` will basically `t.Fatal` if the assertion doesn't pass)

Shaptic (approved):
LGTM!

PR #4519, opened 2022-08-07
exp/lighthorizon: add horizon web docker/k8s deployment

sreuland:
<details>
  <summary>PR Checklist</summary>
  
### PR Structure

* [x] This PR has reasonably narrow scope (if not, break it down into smaller PRs).
* [ ] This PR avoids mixing refactoring changes with feature changes (split into two PRs
  otherwise).
* [ ] This PR's title starts with name of package that is most changed in the PR, ex.
  `services/friendbot`, or `all` or `doc` if the changes are broad or impact many
  packages.

### Thoroughness

* [ ] This PR adds tests for the most critical parts of the new functionality or fixes.
* [ ] I've updated any docs ([developer docs](https://developers.stellar.org/api/), `.md`
  files, etc... affected by this change). Take a look in the `docs` folder for a given service,
  like [this one](https://github.com/stellar/go/tree/master/services/horizon/internal/docs).

### Release planning

* [ ] I've updated the relevant CHANGELOG ([here](services/horizon/CHANGELOG.md) for Horizon) if
  needed with deprecations, added features, breaking changes, and DB schema changes.
* [ ] I've decided if this PR requires a new major/minor version according to
  [semver](https://semver.org/), or if it's mainly a patch change. The PR is targeted at the next
  release branch if it's not a patch change.
</details>

### What

Added docker and k8s deployment wrapper for the lighthorizon web process. 
Refactor, consolidated  docker images belonging to lighthorizon processes to be under e`xp/lighthorizon/build`
documented [k8s MVP deployment](https://stellarorg.atlassian.net/wiki/spaces/HOR/pages/4393730346/k8s+deployment)

Closes #4493 

### Why

Enable deployment of lighthorizon web process onto infra/container orchestration layers.

PR #4543, opened 2022-08-13
exp/lighthorizon/build/index-batch: merge map/reduce updates to latest on feature branch

sreuland:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This is merging the net changes to map/reduce from #4501  to the latest on main feature branch, lighthorizon, which had changes merged from #4519  which refactored/consolidated related docker images.

### Why

remove merge conflicts so net changes to map/reduce can be be committed, this PR represents the net changes needed from #4501 and deprecates need to merge that PR.


exp/lighthorizon/build/k8s/lighthorizon_batch_map_job.yml
```diff
@@ -0,0 +1,43 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: 'batch-map-job'
+spec:
+  completions: 52
+  parallelism: 10
```
Shaptic, line 7: What does this mean?

exp/lighthorizon/build/k8s/lighthorizon_batch_map_job.yml
```diff
@@ -0,0 +1,43 @@
...
+            name: <reference to secret name here if needed for source/target>
+        env:
+        - name: RUN_MODE
+          value: "map" 
+        - name: BATCH_SIZE
+          value: "10048"   
+        - name: FIRST_CHECKPOINT
+          value: "41426080"
+        - name: WORKER_COUNT
+          value: "8"
```
Shaptic, line 27: 8 workers but 4 CPUs? :thinking:

exp/lighthorizon/build/k8s/lighthorizon_batch_map_job.yml
```diff
@@ -0,0 +1,43 @@
...
+        envFrom:
+        - secretRef:
+            name: <reference to secret name here if needed for source/target>
+        env:
+        - name: RUN_MODE
+          value: "map" 
+        - name: BATCH_SIZE
+          value: "10048"   
+        - name: FIRST_CHECKPOINT
+          value: "41426080"
```
Shaptic, line 25: Arbitrary?

exp/lighthorizon/index/cmd/map.sh
```diff
@@ -68,7 +68,7 @@ for (( i=0; i < $BATCH_COUNT; i++ ))
 do
     echo -n "Creating map job $i... "
 
-    NETWORK_PASSPHRASE='testnet' MODULES='accounts_unbacked,transactions' \
+    NETWORK_PASSPHRASE='testnet' JOB_INDEX_ENV='AWS_BATCH_JOB_ARRAY_INDEX' MODULES='accounts_unbacked,transactions' \
```
Shaptic, line 71: Same comment here as on the other PR - why do we need the name of the name of the value rather than just passing the value itself? e.g. `JOB_INDEX=$AWS_BATCH_JOB_ARRAY_INDEX`?

exp/lighthorizon/index/backend/file.go
```diff
@@ -25,6 +25,12 @@ func NewFileBackend(dir string, parallel uint32) (*FileBackend, error) {
 		parallel = 1
 	}
 
+	err := os.MkdirAll(dir, fs.ModeDir|0755)
+	if err != nil {
+		log.Errorf("Unable to mkdir %s, %v", dir, err)
+		return nil, err
+	}
+
```
Shaptic, line 33: This is a behavior change. I'm for it, but can we use this as a chance to add a docstring? e.g.

```
NewFileBackend connects to indices stored at `dir`, creating the directory if one doesn't 
exist, and uses `parallel` to control how many workers to use when flushing to disk.
```

Shaptic (approved):
One small suggestion (re: job index naming), and some comments around defaults, but otherwise :+1:

exp/lighthorizon/build/k8s/lighthorizon_batch_map_job.yml
sreuland: k8s job will only run 10 pods concurrently, and job will continue to start a new pod when one exits, until 52 pods have been launched, each pod getting assigned  a unique `JOB_COMPLETION_INDEX=0..51`

exp/lighthorizon/build/k8s/lighthorizon_batch_map_job.yml
sreuland: yes, a little bit of time slicing, these can be suggestive, for where this is aiming to deploy on dev cluster, there aren't huge amount of resources available, this also ties into the `parallelism: 10` which is multiplier of cpu/ram limits here for job total.

exp/lighthorizon/build/k8s/lighthorizon_batch_map_job.yml
sreuland: yes, @2opremio mentioned running for roughly the last month to start with to verify the job stream, this is around July 1st.

exp/lighthorizon/index/cmd/map.sh
sreuland: hmm, good point, this would be a good simplification, might squeeze it in.

sreuland: oh, I think I see why, under different orchestration environments like k8s, aws, that layer injects the job index number as a container env variable, the name can be anything, AWS_BATCH_JOB_ARRAY_INDEX or JOB_COMPLETION_INDEX for k8s, and there is no shell layer like here for interpolation, the only way the app code which got launched in the container deployed on orchestration layer can realistically reference/find the value is to be given the known deployment env var key name for job index instead and app code then performs the 2nd look up against that in env space to get the value.

exp/lighthorizon/build/k8s/lighthorizon_batch_map_job.yml
Shaptic: very cool, thanks!

.github/workflows/horizon.yml
```diff
@@ -134,14 +134,27 @@ jobs:
...
+          mkdir $PWD/index
+
+          ./exp/lighthorizon/build/build.sh all stellar latest false
+          docker run -e ARCHIVE_TARGET=file:///ledgerexport\
                      -e START=5\
-                     -e END=50\
+                     -e END=150\
                      -e NETWORK_PASSPHRASE="Test SDF Network ; September 2015"\
                      -e CAPTIVE_CORE_CONFIG="/captive-core-testnet.cfg"\
                      -e HISTORY_ARCHIVE_URLS="https://history.stellar.org/prd/core-testnet/core_testnet_001,https://history.stellar.org/prd/core-testnet/core_testnet_002"\
```
2opremio, line 146: I'm not sure it's a good idea to test on testnet (since it can be reset)

exp/lighthorizon/build/k8s/lighthorizon_batch_map_job.yml
2opremio: > 8 workers but 4 CPUs? 🤔

I think the process is mostly IO-bound so I think it's fine to have more workers than CPUs

.github/workflows/horizon.yml
2opremio: Also, there's a missing space before the `horizon-light:` job

exp/lighthorizon/build/index-batch/README.md
```diff
@@ -0,0 +1,7 @@
+# `stellar/horizon-indexer`
```
2opremio, line 1: Isn't the image called `stellar/lighthorizon-index-batch`?

.github/workflows/horizon.yml
sreuland: yes, intent was to avoid the long catchup seen on pubnet ranges, trade-off was a potential window of a couple minutes of time when newly reset testnet hasn't gen'd at least 150 ledgers yet during which this CI job step would fail if was invoked at same time. 

I changed it on a small follow-up [PR #4552](https://github.com/stellar/go/pull/4552/files#diff-734d2e7c3b283e8b975c1e9e58292168bc7040248b3931d7610c8eee19591c1c) to reference same early range but on pubnet, maybe that suffices, need to check how long catchup runs on that, if it's quick, then maybe suffices?

exp/lighthorizon/build/index-batch/README.md
sreuland: yes, thanks for catch, fixed on small follow-up  [pr #4552](https://github.com/stellar/go/pull/4552/files#diff-81d03a4e1934725ea7701357fdd0f191180a0f452c8eb78a8f4abfc1325ffc4f).

PR #4544, opened 2022-08-15
Merge release-horizon-v2.20.0 into master

bartekn:
Contains two commits without reviews fixing the build so PTAL.

PR #4545, opened 2022-08-16
.github: Add building binaries to check workflow

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add `build_release_artifacts` script to `check` workflow.

### Why

We don't check if binaries are successfully built before release action. But it's possible that something can be broken, i.e. the windows build was broken during the last Horizon release. It's better to ensure everything is working in every PR.

### Known limitations

[TODO or N/A]

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -48,6 +48,7 @@ jobs:
     - uses: ./.github/actions/setup-go
       with:
         go-version: ${{ matrix.go }}
+    - run: go build ./...
     - name: Build binaries
       run: go run ./support/scripts/build_release_artifacts/main.go
 
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -33,8 +33,6 @@ jobs:
     - run: ./gogenerate.sh
     - run: ./govet.sh
     - run: ./staticcheck.sh
-    - name: Build binaries
-      run: go run ./support/scripts/build_release_artifacts/main.go
 
   build:
     strategy:
@@ -50,7 +48,8 @@ jobs:
     - uses: ./.github/actions/setup-go
       with:
         go-version: ${{ matrix.go }}
-    - run: go build ./...
+    - name: Build binaries
+      run: go run ./support/scripts/build_release_artifacts/main.go
 
   test:
     strategy:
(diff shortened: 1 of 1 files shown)


bartekn:
Tests will pass after https://github.com/stellar/go/pull/4544 is merged into master.

.github/workflows/go.yml
```diff
@@ -33,6 +33,8 @@ jobs:
     - run: ./gogenerate.sh
     - run: ./govet.sh
     - run: ./staticcheck.sh
+    - name: Build binaries
+      run: go run ./support/scripts/build_release_artifacts/main.go
```
leighmcculloch, line 37: What's the motivation for having the artifact build occur in the `check` job?

Does build binaries build code? If so I think it should be outside the `check` because check is primarily doing non-build related things like linting and is intended to be a very fast job. It could go into the `build` job which I think makes sense because the build command will have already built all the artifacts and presumably the artifacts command will be reusing those cached artifacts.

leighmcculloch (commented):
One comment, and it looks like a couple checks are failing.

.github/workflows/go.yml
bartekn: OK, I missed the `build` job we already have. I'm replaced the `go build ./...` with `build_release_artifacts` scripts because it builds for all platforms.

leighmcculloch (approved):

.github/workflows/go.yml
```diff
@@ -48,7 +48,8 @@ jobs:
     - uses: ./.github/actions/setup-go
       with:
         go-version: ${{ matrix.go }}
-    - run: go build ./...
```
leighmcculloch, line 51: Wait, we need `go build ./...` still.

leighmcculloch (changes requested):

.github/workflows/go.yml
leighmcculloch: `go build ./...` builds all packages in the entire module. The `go run` command that has replaced it doesn't do that. We still need to build all packages, otherwise we won't know about compile errors in packages not referenced by an artifact.

leighmcculloch (approved):

PR #4547, opened 2022-08-16
services/horizon: Fix check_release_hash in Golang 1.18+

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Ensure the git working dir in `check_release_hash` image is clean.

### Why

Starting from the Go 1.18 [release notes](https://tip.golang.org/doc/go1.18):
> The go command now embeds version control information in binaries. It includes the currently checked-out revision, commit time, and a flag indicating whether edited or untracked files are present. Version control information is embedded if the go command is invoked in a directory within a Git, Mercurial, Fossil, or Bazaar repository, and the main package and its containing main module are in the same repository. This information may be omitted using the flag -buildvcs=false.

Go binaries now embed git state information in `vcs` map. One of the fields: `vcs.modified` is set to `true` if working dir is not clear, `false` otherwise. This can break builds reproducibility if Horizon is built in a dirty dir.

### Known limitations

[TODO or N/A]

### DIFF
--- a/services/horizon/internal/scripts/check_release_hash/check.sh
+++ b/services/horizon/internal/scripts/check_release_hash/check.sh
@@ -20,6 +20,14 @@ unzip $TAG-windows-amd64.zip
 
 cd -
 
+# Since Go 1.18 vcs (git) info is added to the binary. One of the values is:
+# vcs.modified which determines if git working dir is clean. We need to
+# specifically add the files below to .gitignore to make git ignore them.
+touch ~/.gitignore
+echo -e "check.sh\n" >> ~/.gitignore
+echo -e "released/\n" >> ~/.gitignore
+git config --global core.excludesFile '~/.gitignore'
+
 git pull origin --tags
 git checkout $TAG
 # -keep: artifact directories are not removed after packaging
--- a/services/horizon/internal/scripts/check_release_hash/Dockerfile
+++ b/services/horizon/internal/scripts/check_release_hash/Dockerfile
@@ -1,5 +1,5 @@
 # Change to Go version used in CI or rebuild with --build-arg.
-ARG GO_IMAGE=golang:1.18
+ARG GO_IMAGE=golang:1.19
 FROM $GO_IMAGE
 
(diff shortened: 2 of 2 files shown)


services/horizon/internal/scripts/check_release_hash/check.sh
```diff
@@ -20,6 +20,14 @@ unzip $TAG-windows-amd64.zip
 
 cd -
 
+# Since Go 1.18 vcs (git) info is added to the binary. One of the values is:
+# vcs.modified which determines if git working dir is clean. We need to
+# specifically add the files below to .gitignore to make git ignore them.
+touch ~/.gitignore
+echo -e "check.sh\n" >> ~/.gitignore
+echo -e "released/\n" >> ~/.gitignore
```
leighmcculloch, line 28: How come these files can't be added to the projects gitignore?

bartekn: Both are created inside this docker image. Not sure if it makes sense to add them `stellar/go`'s `.gitignore` if the files that we ignore here can't be created outside the container.

leighmcculloch (approved):

services/horizon/internal/scripts/check_release_hash/check.sh
Shaptic: Doesn't `echo` add a newline by default? And `>>` will create if it doesn't exist, so no need to `touch`. I think this can be simplified to

```bash
echo "check.sh" >> ~/.gitignore
echo "released/" >> ~/.gitignore
```

Or, combining the two:

```bash
echo -e "check.sh\nreleased/" >> ~/.gitignore

leighmcculloch: Or write the files to `/tmp` instead.

bartekn: Yeah, good points. Let's leave it as it, because it's NIT but whoever will edit this code in the future can simplify it.

PR #4551, opened 2022-08-24
services/horizon: Skip integration tests when no changes

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Skip integration test if they have already passed for the `[hash of horizon binary, hash of horizon integration tests files, p19 core docker image name, p18 core docker image name]` set. The solution leverages `actions/cache@v3` action to store the hash between workflow runs.

### Why

Integration tests usually take more than 25 minutes and will take more if more tests are added in the future. Very often PR in `stellar/go` are not connected to Horizon or change docs. In such case there is no reason to run integration tests.

### Known limitations

I added extended `integration` job just before tests are executed and I use:
```yaml
if: ${{ steps.horizon_binary_tests_hash.outputs.cache-hit != 'true' }}
````
condition in the last run which starts integration tests. This means all of the steps before that like downloading and installing Stellar-Core are executed. The reason for this is that it's impossible to stop Github actions workflow at any job with success result. The run must either fail or all consecutive runs must contain `if` case.

### DIFF
--- a/.github/workflows/horizon.yml
+++ b/.github/workflows/horizon.yml
@@ -100,7 +100,7 @@ jobs:
       uses: actions/cache@v3
       with:
         path: ./horizon
-        key: horizon-hash-${{ hashFiles('./horizon') }}-${{ hashFiles('./services/horizon/internal/integration/**') }}
+        key: horizon-hash-${{ hashFiles('./horizon') }}-${{ hashFiles('./services/horizon/**/*_test.go') }}
 
     - if: ${{ steps.horizon_binary_tests_hash.outputs.cache-hit != 'true' }}
       run: go test -race -timeout 25m -v ./services/horizon/internal/integration/...
--- a/.github/workflows/horizon.yml
+++ b/.github/workflows/horizon.yml
@@ -94,13 +94,14 @@ jobs:
     - name: Build Horizon reproducible build
       run: |
         go build -v -trimpath -buildvcs=false ./services/horizon
+        touch empty
 
     - name: Save Horizon binary and integration tests source hash to cache
       id: horizon_binary_tests_hash
       uses: actions/cache@v3
       with:
-        path: ./horizon
-        key: horizon-hash-${{ hashFiles('./horizon') }}-${{ hashFiles('./services/horizon/internal/integration/**') }}-${{ env.PROTOCOL_19_CORE_DOCKER_IMG }}-${{ env.PROTOCOL_18_CORE_DOCKER_IMG }}
+        path: ./empty
+        key: horizon-hash-${{ hashFiles('./horizon') }}-${{ hashFiles('./clients/horizonclient/**') }}-${{ hashFiles('./protocols/horizon/**') }}-${{ hashFiles('./services/horizon/internal/integration/**') }}-${{ env.PROTOCOL_19_CORE_DOCKER_IMG }}-${{ env.PROTOCOL_18_CORE_DOCKER_IMG }}
 
     - if: ${{ steps.horizon_binary_tests_hash.outputs.cache-hit != 'true' }}
       run: go test -race -timeout 25m -v ./services/horizon/internal/integration/...
--- a/.github/workflows/horizon.yml
+++ b/.github/workflows/horizon.yml
@@ -100,7 +100,7 @@ jobs:
       uses: actions/cache@v3
       with:
         path: ./horizon
-        key: horizon-hash-${{ hashFiles('./horizon') }}-${{ hashFiles('/usr/bin/stellar-core') }}-${{ hashFiles('./services/horizon/**/*_test.go') }}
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by leighmcculloch — APPROVED** [review]
Wow this is cool. 👏🏻 🎉

File: .github/workflows/horizon.yml [hunk 1]
```diff
@@ -91,7 +91,19 @@ jobs:
...
+        go build -v -trimpath -buildvcs=false ./services/horizon
+
+    - name: Save Horizon binary and integration tests source hash to cache
+      id: horizon_binary_tests_hash
+      uses: actions/cache@v3
+      with:
+        path: ./horizon
+        key: horizon-hash-${{ hashFiles('./horizon') }}-${{ hashFiles('./services/horizon/internal/integration/**') }}-${{ env.PROTOCOL_19_CORE_DOCKER_IMG }}-${{ env.PROTOCOL_18_CORE_DOCKER_IMG }}
+
+    - if: ${{ steps.horizon_binary_tests_hash.outputs.cache-hit != 'true' }}
```
leighmcculloch, line 105: I suspect you can do the following, but not big deal.
```suggestion
    - if: !steps.horizon_binary_tests_hash.outputs.cache-hit
```

.github/workflows/horizon.yml
```diff
@@ -91,7 +91,19 @@ jobs:
...
+    - name: Build Horizon reproducible build
+      run: |
+        go build -v -trimpath -buildvcs=false ./services/horizon
+
+    - name: Save Horizon binary and integration tests source hash to cache
+      id: horizon_binary_tests_hash
+      uses: actions/cache@v3
+      with:
+        path: ./horizon
+        key: horizon-hash-${{ hashFiles('./horizon') }}-${{ hashFiles('./services/horizon/internal/integration/**') }}-${{ env.PROTOCOL_19_CORE_DOCKER_IMG }}-${{ env.PROTOCOL_18_CORE_DOCKER_IMG }}
```
leighmcculloch, line 103: What about files outside horizon that change? Such as the SDK?

leighmcculloch (commented):
Hmm, one question.

.github/workflows/horizon.yml
leighmcculloch: This cache line won't just save the horizon file, it will restore it if there's a cache hit. The cache hit is based only on files inside the horizon repo, but the horizon binary could change if the SDK, or dependencies change.

bartekn: Good points! In https://github.com/stellar/go/pull/4551/commits/3acdafc95d045fefed66e0abe5b377801faea4e4:
* I added `clients/horizonclient` hash and `protocols/horizon` hash to cache key.
* Instead of saving horizon binary to cache, I'm now storing empty file. We only care if the key exists or not.

.github/workflows/horizon.yml
bartekn: I found it in the docs and I'm not sure if you GH will auto convert true string to bool so I'll just leave it as it is.

.github/workflows/horizon.yml
leighmcculloch: What about keypair, txnbuild?

It would be ideal if we didn't have to hardcode paths, and instead could use the hash of the binary. Is there someway we could use the hash of the binary instead? Use the hash of the binary as the cache key?

bartekn: I run `go version -m horizon | grep stellar/go` and I got:
```
	path	github.com/stellar/go/services/horizon
	mod	github.com/stellar/go	(devel)	
	dep	github.com/stellar/go-xdr	v0.0.0-20211103144802-8017fc4bdfee	h1:fbVs0xmXpBvVS4GBeiRmAE3Le70ofAqFMch1GTiq
```
so Horizon includes `stellar/go` but pinned to a specific commit/version. This won't change the hash if we update any of `client/**` or `protocols/horizon/**` files in PR branch. Unless I'm missing a nice solution for this it seems our best shot is to include hashes of the packages that affect integration tests in cache key.

leighmcculloch: Oh I just realized you're already doing what I was suggesting, you're already using the hash of the binary, which is sufficient I think. There's no need to use the hash of anything else?

```
go build -v -trimpath -buildvcs=false ./services/horizon
```

```
key: horizon-hash-${{ hashFiles('./horizon') }}
```

leighmcculloch: Oh I get it, you need the integration directory because of fixtures in there.

leighmcculloch (approved):
Looks great. I think I misunderstood with comments above, looks good 👍🏻.

.github/workflows/horizon.yml
bartekn: Yes, we need integration directory but also `clients/horizonclient` and `txnbuild` (added in eb39b8ca) because we use them in integration tests.

Shaptic (approved):
Nice!

PR #4552, opened 2022-08-28
exp/lighthorizon/cmd: index batch reduce tweaks

sreuland:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

fixing minor bug noticed during index batch reduce, s3 urls referencing prior map job output had invalid formatting, also pulled in some more pr feedback updates.

### Why

get the load of indexes to s3 moving forward


.github/workflows/horizon.yml
```diff
@@ -141,18 +142,18 @@ jobs:
...
-                     -e HISTORY_ARCHIVE_URLS="https://history.stellar.org/prd/core-testnet/core_testnet_001,https://history.stellar.org/prd/core-testnet/core_testnet_002"\
+                     -e NETWORK_PASSPHRASE="Public Global Stellar Network ; September 2015"\
+                     -e CAPTIVE_CORE_CONFIG="/captive-core-pubnet.cfg"\
+                     -e HISTORY_ARCHIVE_URLS="https://history.stellar.org/prd/core-live/core_live_001"\
                      -v $PWD/ledgerexport:/ledgerexport\
                      stellar/lighthorizon-ledgerexporter
           
           # run map job
-          docker run -e NETWORK_PASSPHRASE='testnet' -e JOB_INDEX_ENV=AWS_BATCH_JOB_ARRAY_INDEX -e AWS_BATCH_JOB_ARRAY_INDEX=0 -e BATCH_SIZE=64 -e FIRST_CHECKPOINT=64 \
+          docker run -e NETWORK_PASSPHRASE='prodnet' -e JOB_INDEX_ENV=AWS_BATCH_JOB_ARRAY_INDEX -e AWS_BATCH_JOB_ARRAY_INDEX=0 -e BATCH_SIZE=64 -e FIRST_CHECKPOINT=64 \
```
Shaptic, line 152: it's `pubnet`

exp/lighthorizon/index/connect.go
```diff
@@ -24,6 +24,11 @@ func ConnectWithConfig(config StoreConfig) (Store, error) {
 	}
 	switch parsed.Scheme {
 	case "s3":
+		s3Url := fmt.Sprintf("%s/%s", config.URL, config.URLSubPath)
```
Shaptic, line 27: why not `filepath.Join` here like you did below?

exp/lighthorizon/index/store.go
```diff
@@ -38,8 +38,11 @@ type Store interface {
 
 type StoreConfig struct {
 	// init time config
-	URL     string
-	Workers uint32
+	// the base url for the store resource
+	URL string
+	// optional url path to append to the base url to realize the complete url
+	URLSubPath string
```
Shaptic, line 44: Can you explain the rationale for putting this into the store itself? It feels odd to me that the index store itself should care about something like a sub-URL when the caller can just do a `Join(basePath, subPath)` prior to calling `index.ConnectWithConfig`. That is, above in [this diff](https://github.com/stellar/go/pull/4552/files#diff-cd102d7ba72f83c517485cae749ada0814bf7daa7a04cd6f67eb60421bf7cabdR116-R120), why not do the string join prior to the connect as it was before?

sreuland: `filepath.Join("s3://testing", "job_0")` will mangle the proto scheme slashes, the result from this example is `s3:/testing/job_0` which aws sdk didn't like, so, moved the url path composition down to the point at which real store is determined from scheme already and build URL paths in that known context, rather than caller duplicating the same logic to check scheme first to identify store type to know how to compose url path.

.github/workflows/horizon.yml
sreuland: thanks, fixed it.

exp/lighthorizon/index/store.go
Shaptic: I see :thinking: still a little weird to me as I think it's still the caller's responsibility to build sane paths, but I can understand the rationale for hiding it, as well.

Shaptic (approved):

Issue #4526, opened 2022-08-09
Horizon Lite: Automate ledger cache management for the web service.


sreuland:
A crucial part of getting Horizon Lite to an acceptable level of performance is caching ledgers locally to avoid latency spikes from downloading ledgers.

There's a toolkit to "pre-load" the cache in addition to the existing behavior of caching ledgers on use, e.g.:

```
exp/lighthorizon> ./lighthorizon cache build --start <ledger>
```

This same process should be done in parallel on startup of the web service if a parameter is provided, and it should subsequently be integrated into the k8s deployment.

### Acceptance Criteria

- [ ] An operator can easily control how much index data is pre-cached locally based on number of ledgers as a web server startup parameter. Enable parallel cache warm up, web server should be responsive as early as possible, using the cache immediately, of which it's state can be changing as it's warmed up from async population process which does not interfere with concurrent web read access attempts to the index.

### Sub tasks
- [x] enable index to support concurrent read/write operations safely, @Shaptic  - https://github.com/stellar/go/pull/4575 
- [ ] add `--preload-cache=<count>` flag support to web server startup params, k8s manifest, @sreuland

sreuland:
@Shaptic , wanted to confirm on scope, this would be one new startup parameter to the horizon light web server, like `--preload-cache=true|false` and the behavior is that web server will launch an async parallel routine that executes the `./tools cache build --start <ledger>` , and can it just pass 0 for `--start` to default to earliest ledger in the source index?

Shaptic:
That's a great question :thinking: I don't think starting at the earliest available ledger is a good idea. Based on our research, _recent_ ledgers make much more sense to cache up-front. Perhaps we can do something like `--preload-cache=<count>` which does a `cache build --start $(latest - count) --count <count>` (where `latest` is a pseudo-placeholder for the latest ledger in the index [or meta? this is hard to determine from the index] store).

As for async/parallel, my perspective was more that it happens **before** the web server launches, i.e. if you want to deploy the webserver, it first loads X ledgers into the cache, then starts the webserver. So it's almost part of the provisioning/installation/setup step of deployment. I'm concerned about synchronization and race conditions issues if it's done in parallel to serving requests.

paulbellamy:
[...]

PR #4575, opened 2022-09-02
support/storage: Make the on-disk cache thread-safe.

Shaptic:
### What
When caching files, add a `<name>.lock` file that exists until the file is fully written. When retrieving files, check that they don't have a corresponding .lock file before reading them from the cache.

### Why
This adds thread safety to the on-disk cache and is necessary for both #4468 and #4526.

### Known limitations
n/a

### DIFF
--- a/support/storage/ondisk_cache.go
+++ b/support/storage/ondisk_cache.go
@@ -65,7 +65,10 @@ func (b *OnDiskCache) GetFile(filepath string) (io.ReadCloser, error) {
 	L := b.log.WithField("key", filepath)
 	localPath := path.Join(b.dir, filepath)
 
-	if _, ok := b.lru.Get(localPath); !ok {
+	// If the lockfile exists, we should defer to the remote source.
+	_, err := os.Stat(nameLockfile(localPath))
+
+	if _, ok := b.lru.Get(localPath); !ok || err == nil {
 		// If it doesn't exist in the cache, it might still exist on the disk if
 		// we've restarted from an existing directory.
 		local, err := os.Open(localPath)
@@ -94,7 +97,9 @@ func (b *OnDiskCache) GetFile(filepath string) (io.ReadCloser, error) {
 			return remote, nil
 		}
 
-		return teeReadCloser(remote, local), nil
+		return teeReadCloser(remote, local, func() error {
+			return os.Remove(nameLockfile(localPath))
+		}), nil
 	}
 
 	// The cache claims it exists, so just give it a read and send it.
@@ -143,7 +148,8 @@ func (b *OnDiskCache) Size(filepath string) (int64, error) {
 		}
 
 		L.WithError(err).Debug("retrieving size of cached ledger failed")
-		b.lru.Remove(localPath) // stale cache?
+		b.lru.Remove(localPath)            // stale cache?
+		os.Remove(nameLockfile(localPath)) // errors don't matter
 	}
 
 	return b.Storage.Size(filepath)
@@ -162,7 +168,9 @@ func (b *OnDiskCache) PutFile(filepath string, in io.ReadCloser) error {
 		L.WithError(err).Error("failed to put file locally")
 	} else {
 		// tee upload data into our local file
-		in = teeReadCloser(in, local)
+		in = teeReadCloser(in, local, func() error {
+			return os.Remove(nameLockfile(path.Join(b.dir, filepath)))
+		})
 	}
 
 	return b.Storage.PutFile(filepath, in)
@@ -202,9 +210,17 @@ func (b *OnDiskCache) createLocal(filepath string) (*os.File, error) {
 	if err != nil {
 		return nil, err
 	}
+	_, err = os.Create(nameLockfile(localPath))
+	if err != nil {
+		return nil, err
+	}
 
 	b.lru.Add(localPath, struct{}{}) // just use the cache as an array
 	return local, nil
+}
+
+func nameLockfile(file string) string {
+	return file + ".lock"
 }
 
 // The below is a helper interface so that we can use io.TeeReader to write
@@ -219,12 +235,13 @@ func (t trc) Close() error {
 	return t.close()
 }
 
-func teeReadCloser(r io.ReadCloser, w io.WriteCloser) io.ReadCloser {
+func teeReadCloser(r io.ReadCloser, w io.WriteCloser, onClose func() error) io.ReadCloser {
 	return trc{
 		Reader: io.TeeReader(r, w),
 		close: func() error {
 			r.Close()
-			return w.Close()
+			w.Close()
+			return onClose()
 		},
 	}
 }
--- a/support/storage/ondisk_cache.go
+++ b/support/storage/ondisk_cache.go
@@ -66,9 +66,9 @@ func (b *OnDiskCache) GetFile(filepath string) (io.ReadCloser, error) {
 	localPath := path.Join(b.dir, filepath)
 
 	// If the lockfile exists, we should defer to the remote source.
-	_, err := os.Stat(nameLockfile(localPath))
+	_, statErr := os.Stat(nameLockfile(localPath))
 
-	if _, ok := b.lru.Get(localPath); !ok || err == nil {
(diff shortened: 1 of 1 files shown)


support/storage/ondisk_cache.go
```diff
@@ -65,7 +65,10 @@ func (b *OnDiskCache) GetFile(filepath string) (io.ReadCloser, error) {
...
 	localPath := path.Join(b.dir, filepath)
 
-	if _, ok := b.lru.Get(localPath); !ok {
+	// If the lockfile exists, we should defer to the remote source.
+	_, statErr := os.Stat(nameLockfile(localPath))
+
+	if _, ok := b.lru.Get(localPath); !ok || statErr == nil {
 		// If it doesn't exist in the cache, it might still exist on the disk if
 		// we've restarted from an existing directory.
 		local, err := os.Open(localPath)
```
sreuland, line 74: should this attempt to access local cache file be guarded by stateErr != nil? these local disk cache files could be getting re-written by the async preload process concurrently at same time?

support/storage/ondisk_cache.go
```diff
@@ -202,11 +210,19 @@ func (b *OnDiskCache) createLocal(filepath string) (*os.File, error) {
...
+	_, err = os.Create(nameLockfile(localPath))
+	if err != nil {
+		return nil, err
+	}
 
 	b.lru.Add(localPath, struct{}{}) // just use the cache as an array
 	return local, nil
 }
 
+func nameLockfile(file string) string {
```
sreuland, line 222: rather than a lock file per each cache file, wdyt about courser lock scope and intra-process for preload exec, it would cover loading of the entire cache as one atomic thing, and letting the caller coordinate it by pass the lock:

```
func MakeOnDiskCache(upstream Storage, dir string, maxFiles uint, preloadSemaphore *bool) (Storage, error) 
```

then could replace `os.Stat(nameLockfile(localPath))` with `atomic.LoadBool(preloadSemaphore)`

was mentioning intra-process for the preload, because it doesn't seem to add value for for externalizing preload as another o/s process(requiring file lock) rather, the preload can be a goroutine launched in the web server process at startup and passes the semaphore from that to here. works in bare-metal or containerized environments with less config(i.e. no need to define a second container in k8s deployment pod to run the preload)

support/storage/ondisk_cache.go
Shaptic: `b.lru.Get` will only check the _in-memory_ LRU cache that will tell us if `localPath` might be on disk, so it's a safe check. It doesn't touch the on-disk file.

support/storage/ondisk_cache.go
Shaptic: So I'm a little hazy on the details on what you're proposing here, but let me offer up a justification for the fine-grained locking I did here.

First off, this part is intentionally separate from the add'l code that will be necessary for the preload task, because I wanted to (a) keep things decoupled and (b) this needs thread safety for #4468 (parallel fetches) as well. I'm seeing issues in my draft PR with reads interfering with parallel fetches because of the caching layer (i.e. parallel fetches are still writing to the cache, but the `GetLedger()` notices a cache hit so it tries to read). That's why the locking is fine-grained on a per-file basis: parallel downloads means some files are safe while multiple may be unsafe.

I agree that the preload itself can be a goroutine, but that's distinct from needing the cache to be thread-safe for simultaneous cache read/write. Maybe the semaphore aspect can come in that next PR and act as a layer on top of the per-file locking here?

sreuland: @Shaptic , got it, the cache thread safety design needs granularity that works at both extremes, i.e. whole cache for preload and concurrently per cache key/entry for web requests. I see how the file lock mechanism encapsulated inside here will work for both those extremes. 

I played around with a reader/writer pattern with mutexes as alternative to per-file locks for concurrency. In this approach, both the web requests and preload threads sync across concurrent read locks and exclusive write locks on each cache key/entry from the same singleton instance of `OnDiskCache`.  

I drew it out [in this sequence diagram](https://www.websequencediagrams.com/?lz=dGl0bGUgbGlnaHQgd2ViIHNlcnZlciAtIHJlYWRlci93cml0ZXIgc3luYyBvbiBjYWNoZQoKcGFydGljaXBhbnQgb3BlcmF0b3IACA1tYWluABkNY2xpZW50ACwNd2ViAGgGAEINcHJlbG9hZABWDUFyY2hpdmUAag1UcmFuc2FjdGlvblNlcnZpYwARDk9uRGlza0MAgSoFAIEiDGxydQCBMQ1maWxlc3lzdGVtAIFIDXJlbW90ZVN0b3JhZ2UKCgCBXAgtPm1haW46IHBhc3MgY2xpIGFyZ3Moc291cmNlLGluZGV4ZXMsbGVkZ2VyLQCCJgUsIAACDC1zaXplKQptYWluLT4AgRcLOiBNYWtlAIEoCwBLCAAxDQA3FQpvcHQAgxAGIACDJgVzIGZyb20Agj4IZXIAZgcAgk8HOiBCdWlsZABeBiBvAIIaCiwgc3RhcnQsIGNvdW50LCByZXBhaXI9dHJ1ZSkKYWN0aXZhdGUAgxMJb3B0IGl0ZXJhdGUgYWxsIACBdQUAgkUGIGluIHJlcXVlc3RlZCByYW5nZSAKAINNBy0-AINCBzogR2V0TACCIgUoeCkKAINYBwCCBA9QdXRGaWxlKHhfRmlsZXBhdGgsAIUQBUxvY2s9bmlsAIENCwCDYQwAg24LLT4AhT8GX211dGV4OiBnZXQgbG9jayBmb3IAhVkHABUGWwBaCl0AMw4AhXkGACwVAA8MACcabHJ1OiBwdXQoZgCBPgcpIACBFQ4AhHcKOiByZQCGagUAhCcNLwCBdgoAgH8cdW4AgVUFAG8mAIIJDgA0BwCBeBlkZQCCVBVlbmQAg08KAIYRBmZpbmlzaGVkAIc8CQAwCwCHUAhlbmQAhSkMcmVhZCBhbmQAhTENbXVsdGlwbGUAiQYMdGhyZWFkcwCGNgcAhEAJTmV3SW5nZXN0AIgdByguLi4AhU0LAIZlCACIYwk6AIVjBgCJYAx3aXRoAIY6CWFwcGVkIGEAiGoHAIkwBgAyDUdFVCAvYWMAhiMFcy97SUR9AIYWCgCJSAoAiVMJLT4AiRcSAIV6BQCJNQtzQnlBAEsGCgCJRBIAhh0YAIcMCQCKFQgAhjAWR2UAhj4IAIR4CQCGBzEAhm4IID0AhjgLcm9tAIYqGgCJCwphbHJlYWR5IGhhcyBhbiBlbnRyeSwganVzdCB1cwCDeQZlcgCHEQUAhiUTZ2UAhhImZ2UAi0MGAIUrOwCHBw9BdACJIgcAUgVlbgCJbgYAimIGZG9lcyBub3QgaGF2ZQCBSgsAgUgIAIVFCgCBRRMAiUwXAIgWCACJWQopAHEgAIcAFwCKRAkAhGQUAI0VBgCHRgwAjlMIAIRlFACGKgt0AIUjCwCIBQwAhV0VAI9mBgAmDwCHcwUKCgo&s=default)

the net effect on sync behavior should be the same, it's just a matter of how much thrash do we think is avoided by doing memory locks vs. file system locks, lmk if you think it's worth exploring this. 

the `OnDiskCache` interface here stays the same, but, if it's used, on that follow-on PR I do as part of #4526 would be to have the web server main create one instance of `OnDiskCache` and pass it to Archive for all web request threads and same instance passed to preload `BuildCache` thread, rather than letting each of them create their own instance.

Shaptic: The benefit of file-based locks is that there's no need for read locks - we get that "for free" from the OS (concurrent reads of a file is always safe), and similarly there's no need for a singleton because it's all on disk and visible to everyone. 

And also the biggest benefit is that the work is already done :laughing: imo we should punt exploring memory-based synchronization till later unless we see issues with this. The latency is going to be dominated by ledger and index downloads, anyway.

sreuland (approved):

PR #4577, opened 2022-09-06
all: Bump Go to 1.18.6 and 1.19.1

bartekn:
Ref: https://groups.google.com/g/golang-announce/c/x49AQzIVX-s/m/0tgO0pjiBQAJ

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -18,7 +18,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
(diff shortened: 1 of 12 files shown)


paulbellamy (approved):
Approved, but there's also a docker image `Dockerfile.dev` which is on `golang:1.18`. Would be good to update that as well so they stay in sync.

bartekn:
@paulbellamy done in 4f4faaf1.

leighmcculloch (approved):

PR #4579, opened 2022-09-08
exp/lighthorizon: Incorporate tool subcommands into the webserver.

Shaptic:
### What
This refactors the web server to properly use `cobra.Command` to manage itself and its subcommands.

We now have:

```
Horizon Lite command suite             

Usage:
  lighthorizon <subcommand> [flags]
  lighthorizon [command]

Available Commands:
  cache       
  index       
  serve       

Use "lighthorizon [command] --help" for more information about a command.
```

where `lighthorizon serve` mimics the behavior we had before. Flags now have POSIX-compliant (i.e. `--`) instead of Go's default (i.e. `-`) prefixes to commands.

This also adds a root response with some details about the configuration and moves the OpenAPI doc to `/api`.

### Why
This consolidates all tooling under a single binary and makes its usage parallel regular Horizon better.

The root response is nice because accessing the web server triggers a file download by default (at least for me on Firefox).

### Known limitations
n/a


exp/lighthorizon/main.go
```diff
@@ -1,81 +1,137 @@
...
 	"github.com/stellar/go/exp/lighthorizon/services"
+	"github.com/stellar/go/exp/lighthorizon/tools"
 
 	"github.com/stellar/go/network"
 	"github.com/stellar/go/support/log"
 )
 
 const (
-	defaultCacheSize = (60 * 60 * 24) / 6 // 1 day of ledgers @ 6s each
+	HorizonLiteVersion = "0.0.1-alpha"
```
sreuland, line 22: nice, I was thinking about how to trap the git commit during docker build also, may be something for considering to add later, can also help with troubleshooting, i.e. confirming runtime version,etc.

exp/lighthorizon/http_test.go
```diff
@@ -2,7 +2,7 @@ package main
```
sreuland, line 2: just for consideration, could add a test here to confirm new `RootResponse` on `GET /`

exp/lighthorizon/main.go
```diff
@@ -1,81 +1,137 @@
...
 
-	Config := services.Config{
-		Archive:    ingestArchive,
-		Passphrase: *networkPassphrase,
-		IndexStore: indexStore,
-		Metrics:    services.NewMetrics(registry),
-	}
+	serve := &cobra.Command{
+		Use: "serve <txmeta source> <index source>",
+		Long: `Starts the Horizon Lite server, which only serves the following endpoint:
```
sreuland, line 39: this Long description will likely get stale soon, what if instead it defers to generic like:

`Launches web service and binds it to port 8080 on all local interfaces of host machine.  Refer to http://localhost:8080/api for available api endpoints:`

We could enhance api_docs.yml, to highlight this state of service, i.e. it's in MVP mode and only certain endpoints, etc?

exp/lighthorizon/tools/main.go
```diff
@@ -19,7 +19,7 @@ func main() {
 		},
 	}
 
-	cmd = addCacheCommands(cmd)
```
sreuland, line 22: this whole main can be deleted now, since these subs were added to lighthorizon?

exp/lighthorizon/actions/root.go
```diff
@@ -0,0 +1,29 @@
...
+)
+
+type RootResponse struct {
+	Version      string `json:"version"`
+	LedgerSource string `json:"ledger_source"`
+	IndexSource  string `json:"index_source"`
+	LatestLedger uint32 `json:"latest_indexed_ledger"`
+}
+
+func Root(config RootResponse) func(http.ResponseWriter, *http.Request) {
```
sreuland, line 18: just curious, is it considered slightly more idiomatic/standard to represent as struct method:
`func (config *RootResponse) Handle func(http.ResponseWriter, *http.Request)`

sreuland (approved):
nice refactor

exp/lighthorizon/main.go
Shaptic: yeah definitely, I just didn't want to over-complicate it for this PR. we do something like that in Horizon so we can probably borrow the code from there

exp/lighthorizon/tools/main.go
Shaptic: good point :facepalm:

exp/lighthorizon/http_test.go
Shaptic: great call! :+1:

exp/lighthorizon/main.go
Shaptic: Definitely a good idea :+1: we'll need a proper pass of the OpenAPI spec once we're out of MVP anyway

exp/lighthorizon/actions/root.go
Shaptic: Honestly? I have no idea. I was following the pattern for the other actions.

PR #4581, opened 2022-09-09
services/ticker: convert trades pk to BigInt

accordeiro:
### What

This migrates the Ticker's `trades`  table PK to BigInt so it can store more entries, and also makes sure the ids will cycle once they get to the Int64 ceiling (as the Ticker only cares about the last 7-day window of data and everything else can be cleaned up).

### Why

The volume of trades has significantly increased. Even with frequent clean ups, the sequential id will likely hit the ceiling of Int32.

### Known limitations

This is a knowingly suboptimal approach, as it will re-write the entire `trades` table while locking it. For a big `trades` table that isn't frequently cleaned up, this migration can take a long time.

Since the Ticker service is self-recoverable (i.e. it can fully ingest all data it needs to operate in a reasonable amount of time), you can either: 
a) Clean up unused old entries using the `ticker clean trades --keep-days=<NUM_DAYS>` command; or
b) Truncate the `trades` table and re-ingest during a scheduled maintenance window


leighmcculloch (approved):
👍🏻 The migrations looks fine, assuming the table isn't huge. Anyone who runs the migration probably wants to truncate the table first or reduce it using the process noted in the description.

services/ticker/internal/tickerdb/helpers.go
```diff
@@ -136,3 +138,15 @@ func (s *TickerSession) performUpsertQuery(ctx context.Context, dbStruct interfa
 	_, err := s.ExecRaw(ctx, qs, dbValues...)
 	return err
 }
+
+func CheckDBVersion(t *testing.T) *dbtest.DB {
```
leighmcculloch, line 142: 💡 Might be worth naming this something like `Open` given its primary function is the open the database connection. It also means in tests that there will be a corresponding `Open` and `Close()` which is a common pattern.

leighmcculloch (approved):
👍🏻 Looks good.

One suggestion (💡), but regardless looks good.

services/ticker/internal/tickerdb/helpers.go
accordeiro: Good catch -- renaming now

accordeiro: Resolved on [cf40e20](https://github.com/stellar/go/pull/4581/commits/cf40e20e080f0c075de8d15874aa1db6da9580a5)

PR #4582, opened 2022-09-10
exp/lighthorizon/index/cmd: Fix index single watch, slow down the retry on not-found ledgers 

sreuland:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Noticed the index single cmd with `watch` was err'ing on getting latest ledger from txmeta source, and then exiting execution, the s3 'not found' was not being converted to the expected os `ErrNotFound` to allow it to re-attempt after backoff, instead it treated it as severe err instead.

### Why

the single `watch` needs to continue iteration attempts to get latest ledger tx meta, not trigger the entire single process to exit.

Closes https://github.com/stellar/go/issues/4583

### DIFF
--- a/exp/lighthorizon/index/builder.go
+++ b/exp/lighthorizon/index/builder.go
@@ -294,13 +294,13 @@ func (builder *IndexBuilder) Watch(ctx context.Context) error {
 		//
 		// [1]: https://stellarfoundation.slack.com/archives/C02B04RMK/p1654903342555669
 
-		// We sleep with linear backoff starting with 3s. Ledgers get posted
+		// We sleep with linear backoff starting with 6s. Ledgers get posted
 		// every 5-7s on average, but to be extra careful, let's give it a full
 		// minute before we give up entirely.
 		timedCtx, cancel := context.WithTimeout(ctx, 60*time.Second)
 		defer cancel()
 
-		sleepTime := (3 * time.Second)
+		sleepTime := (6 * time.Second)
 	outer:
 		for {
 			time.Sleep(sleepTime)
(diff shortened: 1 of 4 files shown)


support/storage/s3.go
```diff
@@ -5,24 +5,48 @@ import (
...
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/s3"
 	"github.com/aws/aws-sdk-go/service/s3/s3iface"
 	"github.com/stellar/go/support/errors"
 )
 
+type s3HttpProxy interface {
```
sreuland, line 21: In order to enable test access, i refactored the s3 `Send` to be composed as app level interface, so tests could override it and simulate err's, the native s3 sdk was struct and couldn't be mocked.

exp/lighthorizon/index/builder.go
```diff
@@ -294,16 +294,16 @@ func (builder *IndexBuilder) Watch(ctx context.Context) error {
...
 		// minute before we give up entirely.
 		timedCtx, cancel := context.WithTimeout(ctx, 60*time.Second)
 		defer cancel()
 
-		sleepTime := time.Second
-
+		sleepTime := (3 * time.Second)
 	outer:
 		for {
+			time.Sleep(sleepTime)
```
sreuland, line 306: added an up-front wait before first attempt at next ledger tx meta, noticed it was always getting not-found on first attempt of new iteration, watching logs with this in place, it is pretty consistent at showing not-found on first attempt and the second attempt usually resolves with getting the ledger's tx meta.

Shaptic (approved):
Seems like a good adjustment :+1:

PR #4587, opened 2022-09-14
services/horizon: Lock ingestion while applying DB migrations

### LINKED ISSUE #4588: Debug deadlock while applying DB migrations

2.20.0

### What did you do?

Apply `58_add_index_by_id_optimization.sql` migration file while ingestion is running on another ingesting instance.

No error.

### What did you see instead?

Deadlock error:
```
pq: deadlock detected handling 58_add_index_by_id_optimization.sql
```

---

There are two things that are important to remember: 1. the migrations are run in a db transaction and 2. because we have multiple ingesting instance, ingestion is running along db migrations (also in a DB transaction). This is the chain of events I think that leads to deadlock:
1. first `CREATE INDEX` query in the migration file acquires `ShareLock` on `history_operation_claimable_balances` (this lock is going to be released at the end of transaction).
1. one of ingesting instances `INSERT` a row into `history_operation_liquidity_pools` and by doing this acquires `RowExclusiveLock` (it conflicts with `ShareLock` and is also released at the end of transaction).
1. now, `CREATE INDEX` on `history_operation_liquidity_pools` is executed, again `ShareLock` acquire attempt but on `history_operation_liquidity_pools` table and needs to wait until ingestion commits because there’s already a conflicting lock on this table.
1. ingesting instance attempts to `INSERT` a new row into `history_operation_claimable_balances` but waits because in step 1 a `ShareLock` was acquired.
1. deadlock because both txs wait for the other tx lock to be released.

I created a simple local repro.

session 1:
```
deadlock_repro=# create table t1 (f1 int);
CREATE TABLE
deadlock_repro=# create table t2 (f1 int);
CREATE TABLE
deadlock_repro=# begin;
BEGIN
deadlock_repro=*# create index i1 on t1(f1);
CREATE INDEX
deadlock_repro=*# create index i2 on t2(f1);
CREATE INDEX
deadlock_repro=*# 
```
session 2:
```
deadlock_repro=# begin;
BEGIN
deadlock_repro=*# insert into t2 (f1) values (0);
INSERT 0 1
deadlock_repro=*# insert into t1 (f1) values (0);
ERROR:  deadlock detected
LINE 1: insert into t1 (f1) values (0);
                    ^
DETAIL:  Process 80691 waits for RowExclusiveLock on relation 1638638 of database 1638622; blocked by process 80642.
Process 80642 waits for ShareLock on relation 1638641 of database 1638622; blocked by process 80691.
HINT:  See server log for query details.
```
(the queries within txs are run one after another in two sessions: S1, S2, S1, S2)

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Acquire ingestion lock before applying DB migrations to stop ingestion.

### Why

When applying `58_add_index_by_id_optimization.sql` migration file we experienced a deadlock. It can happen when ingestion is running because ingestion by inserting or updating certain rows can acquire `RowExclusiveLock` which conflicts with other locks (like `ShareLock` when creating a new index).

Close #4588.

### Known limitations

We lock ingestion only for `up` migrations. `down` migrations can remove key-value table which can lead to deadlocks.


services/horizon/internal/db2/schema/main.go
```diff
@@ -46,6 +48,28 @@ var Migrations migrate.MigrationSource = &migrate.AssetMigrationSource{
 // upward back to the current version at the start of the process. If count is
 // 0, a count of 1 will be assumed.
 func Migrate(db *sql.DB, dir MigrateDir, count int) (int, error) {
+	txConn, err := db.Conn(context.Background())
+	if err != nil {
+		return 0, err
+	}
+
+	tx, err := txConn.BeginTx(context.Background(), nil)
```
tsachiherman, line 56: given that you've already used `BeginTx` instead of the "classic" `Begin`, why don't you take advantage of that and use the `TxOptions` ?

services/horizon/internal/db2/schema/main.go
```diff
@@ -46,6 +48,28 @@ var Migrations migrate.MigrationSource = &migrate.AssetMigrationSource{
 // upward back to the current version at the start of the process. If count is
 // 0, a count of 1 will be assumed.
 func Migrate(db *sql.DB, dir MigrateDir, count int) (int, error) {
+	txConn, err := db.Conn(context.Background())
```
tsachiherman, line 51: don't forget to add `defer txConn.Close()`

services/horizon/internal/db2/schema/main.go
```diff
@@ -46,6 +48,28 @@ var Migrations migrate.MigrationSource = &migrate.AssetMigrationSource{
...
+
+	tx, err := txConn.BeginTx(context.Background(), nil)
+	if err != nil {
+		return 0, err
+	}
+
+	// Lock ingestion
+	_, err = tx.Query("select value from key_value_store where key = 'exp_ingest_last_ledger' for update")
+	if err != nil {
+		if !strings.Contains(err.Error(), `relation "key_value_store" does not exist`) {
```
tsachiherman, line 64: It's a bit weird to test a specific error like that.. but regardless of that, please add error handling for the case where an error has occurred, and it's *not* the absence of the table. i.e. connection failure, etc.

tsachiherman (commented):
I think that the implementation is almost there - I've added few comments.

services/horizon/internal/db2/schema/main.go
bartekn: `Conn` only [exposes](https://pkg.go.dev/database/sql#pkg-index) `BeginTx` method.

tsachiherman (approved):

PR #4596, opened 2022-09-19
clients/horizonclient: fix theoretical bug in currentServerTime

tsachiherman:
## Description

The code in currentServerTime was not testing the existence of the entry in `ServerTimeMap` correctly. As a result, it would have missed the case where an entry could be missing from the map.

That being said, this is a benign bug. It would never triggered unless the clock is set around epoch time ( which is not really realistic ). This fix is just to make sure that our code is correct. There should be no expectation of any behavioral change.
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

The logic of `currentServerTime` is being update to properly test for absence of hosts entry in the map.

### Why

Potential issue was found during code review.

### Known limitations

N/A

### DIFF
--- a/clients/horizonclient/internal_test.go
+++ b/clients/horizonclient/internal_test.go
@@ -7,28 +7,40 @@ import (
 )
 
 func TestCurrentServerTime(t *testing.T) {
-	currentTime := currentServerTime("non-existing-host-name", 60)
+	t.Run("non-existing-host-name", func(t *testing.T) {
+		currentTime := currentServerTime("non-existing-host-name", 60)
+		defer func() {
+			serverTimeMapMutex.Lock()
+			ServerTimeMap["TestCurrentServerTime-server-behind"] = ServerTimeRecord{ServerTime: 27, LocalTimeRecorded: 23}
+			serverTimeMapMutex.Unlock()
+		}()
 
-	require.Zerof(t, currentTime, "server time for non-existing time is expected to be zero, but was %d instead", currentTime)
+		require.Zerof(t, currentTime, "server time for non-existing time is expected to be zero, but was %d instead", currentTime)
+	})
 
-	serverTimeMapMutex.Lock()
-	ServerTimeMap["TestCurrentServerTime-server-behind"] = ServerTimeRecord{ServerTime: 27, LocalTimeRecorded: 23}
-	serverTimeMapMutex.Unlock()
+	t.Run("server-behind", func(t *testing.T) {
+		currentTime := currentServerTime("TestCurrentServerTime-server-behind", 500)
+		defer func() {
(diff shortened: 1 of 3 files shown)


clients/horizonclient/main_test.go
```diff
@@ -1455,7 +1455,9 @@ func TestFetchTimebounds(t *testing.T) {
 	// When no saved server time, return local time
 	st, err := client.FetchTimebounds(100)
 	if assert.NoError(t, err) {
+		serverTimeMapMutex.Lock()
```
tsachiherman, line 1458: it's important to use locks here so that we could run the unit tests in parallel.

bartekn (approved):
Good catch!

ire-and-curses (approved):
Great!

clients/horizonclient/internal_test.go
```diff
@@ -0,0 +1,34 @@
...
+	serverTimeMapMutex.Lock()
+	delete(ServerTimeMap, "TestCurrentServerTime-server-behind")
+	ServerTimeMap["TestCurrentServerTime-server"] = ServerTimeRecord{ServerTime: 27, LocalTimeRecorded: 23}
+	serverTimeMapMutex.Unlock()
+
+	currentTime = currentServerTime("TestCurrentServerTime-server", 37)
+
+	require.Equalf(t, currentTime, int64(41), "currentServerTime should have returned %d, but returned %d instead", 41, currentTime)
+
+	serverTimeMapMutex.Lock()
```
sreuland, line 31: this looks like general tear-down for test, should it be defer'd  so it always runs, minor, but if the require on ln 29 fails, it would exit before here.

tsachiherman: done. do you mind taking another look ?

clients/horizonclient/internal_test.go
```diff
@@ -0,0 +1,46 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestCurrentServerTime(t *testing.T) {
+	t.Run("non-existing-host-name", func(t *testing.T) {
+		currentTime := currentServerTime("non-existing-host-name", 60)
+		defer func() {
+			serverTimeMapMutex.Lock()
```
sreuland, line 13: this test doesn't need a cleanup as it didn't alter state, but this appears to be the setup for next test, minor suggestion, up to you, maybe could just do this inline at top of the next test so it keeps expected input/outputs together in same test?

sreuland (approved):

clients/horizonclient/internal_test.go
tsachiherman: done.

Issue #4514, opened 2022-08-05
Horizon Lite: Understand request latency characteristics


jcx120:
Can the Lite design support webscale response times? (3-5s)

1. Measure current latency of /account/* mvp endpoints
2. Understand history age vs variation of data
 - Does latency vary with history? provide a distribution, understand the optimal sweet spot
 - Does latency vary with data variance (account, ledger mix etc)? 
3. Request profiling
 - Understand current hotspots (indices lookup > txmeta download > request processing etc) and where would potential optimizations be targeted

Shaptic:
https://docs.google.com/document/d/1hlUbEqD5Jek5YTar9kQ7JvMDLly2wVTr_FdssdPba8c/edit?usp=sharing

PR #4601, opened 2022-09-20
exp/lighthorizon/index: Add ability to disable bits in index.

Shaptic:
### What
 * Adds `BitmapIndex.SetInactive(index uint32)`, which _disables_ (i.e. flips off) the bit representing `index`.
 * Adds a `index purge <args>` tool that takes advantage of the above

### Why
If the index and ledger source are ever out of sync, it's impossible to reconcile the index. The alternative to this is actually having all ledgers available, but this isn't always feasible. This lets you purge out older ledgers you might not have available.

This is (unfortunately) related to #4514.

### Known limitations
The test suite should be comprehensive, but a second set of eyes on edge cases would be great.


exp/lighthorizon/index/types/bitmap.go
```diff
@@ -113,6 +129,83 @@ func (i *BitmapIndex) setActive(index uint32) error {
...
+			i.firstBit = 0
+			i.lastBit = 0
+			i.bitmap = []byte{}
+		} else if err != nil {
+			return err
+		} else {
+			i.firstBit = nextBit
+		}
+
+		for i.bitmap[0] == 0 {
```
tsachiherman, line 157: If I'm reading this right, I suspect that we can avoid this for loop and explicitly calculate the
Byte offset. 
I.e.
I.bitmap = i.bitmap[Nextbit/8-firstbit/8:]

exp/lighthorizon/index/types/bitmap.go
```diff
@@ -113,6 +129,83 @@ func (i *BitmapIndex) setActive(index uint32) error {
...
+				}
+
+				offset, _ := maxBitAfter(i.bitmap[j], 0)
+				byteOffset := uint32(len(i.bitmap)-1) * 8
+				i.lastBit = i.rangeFirstBit() + byteOffset + offset
+				i.bitmap = i.bitmap[:j+1]
+				break
+			}
+		} else {
+			// Otherwise, do we need to adjust the range?
```
tsachiherman, line 177: I think that this section belongs to the upper if statement.
( Do we have a unit test that deletes a bit in the middle ? )

tsachiherman (commented):
Adding a fuzzer test here could help us to detect missing test cases.

exp/lighthorizon/index/types/bitmap.go
Shaptic: We only need to check this if the last byte is non-zero, but in fact your comment made me realize we actually only need to check it in the exact case where the last bit got unset, so I added a986ef8b to simplify this.

exp/lighthorizon/index/types/bitmap.go
Shaptic: You're right, great catch! Done in a986ef8b.

tsachiherman (commented):
I don't believe that the code works correctly. I wrote a fuzzer that is constantly fail - 
```golang
// TestFuzzerSetInactive attempt to fuzz random bits into two
// bitmap sets, one by addition, and one by subtraction - then,
// it compares the outcome.
func TestFuzzerSetInactive(t *testing.T) {
	permLen := uint32(32)
	setBitsCount := 16
	for n := 0; n < 10000; n++ {
		randBits := rand.Perm(int(permLen))
		setBits := randBits[:setBitsCount]
		clearBits := randBits[setBitsCount:]
		setBitmap := &BitmapIndex{}
		for _, bit := range setBits {
			setBitmap.setActive(uint32(bit))
		}
		clearBitmap := &BitmapIndex{}
		for i := uint32(0); i < permLen; i++ {
			clearBitmap.setActive(i)
		}
		for _, bit := range clearBits {
			setBitmap.setInactive(uint32(bit))
		}
		require.Equal(t, setBitmap, clearBitmap)
	}
}
```

Shaptic:
@tsachiherman in your test:

1. after the first loop, `setBitmap` has half of its bits flipped
2. after the second loop, `clearBitmap` has _all_ of its bits flipped
3. after the third lop, `setBitmap` has the same state as in (1), because `setInactive` on an already-inactive bit is a no-op, and we know that `clearBits` does not intersect `setBits` because it's a permutation

Thus, how could `setBitmap` ever match `clearBitmap`? I think what you meant is to do `clearBitmap.setInactive` in the third loop. I made this change and it revealed a couple of subtle bugs (how `lastBit`, when/how compression happens), **and** it actually revealed an optimization opportunity in `setActive` to avoid expanding an extra empty byte sometimes :tada: 

All of these are fixed in 3381caf2. Nice catch, thanks! Fuzzing was a great idea :+1:

tsachiherman (approved):
looks great!

PR #4603, opened 2022-09-21
all: add golangci linter as a GitHub action

tsachiherman:
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This PR adds the golangci linter as a GitHub action.

The command line version would be more capable than this version, as the integration with GitHub isn't that great.
( adding comments require elevated privileges, which might be a security issue ).

### Why

It would allow us to get *some* linting feedback for our proposed changes.

### Known limitations

Using this without code comments won't make it as useful as it could be.. but it's better than not having this at all. We should re-iterate on this one once in a month or so - after we've gathered some feedback from engineers.

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -19,10 +19,8 @@ jobs:
         with:
           fetch-depth: 0 # required for new-from-rev option in .golangci.yml
       - uses: actions/setup-go@v3
-    steps:
-      - run: >
-        mkdir -p /home/runner/work/go/src/github.com/stellar
-        mv /home/runner/work/go/go /home/runner/work/go/src/github.com/stellar
+      - run: mkdir -p /home/runner/work/go/src/github.com/stellar
+      - run: mv /home/runner/work/go/go /home/runner/work/go/src/github.com/stellar
       - name: Run linters & report
         uses: golangci/golangci-lint-action@537aa1903e5d359d0b27dbc19ddd22c5087f3fbc
         with:
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -1,4 +1,4 @@
-name: golangci-lint
+name: Linters
 on:
   push:
     branches:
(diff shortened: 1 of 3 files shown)


.github/workflows/reviewdog.yml
```diff
@@ -0,0 +1,29 @@
...
+    name: runner / golangci-lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0 # required for new-from-rev option in .golangci.yml
+      - name: Run linters & report
+        env:
+          REVIEWDOG_GITHUB_API_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        uses: reviewdog/action-golangci-lint@v2
```
leighmcculloch, line 24: It's safer to pin the action to a specific commit per https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions.
```suggestion
        uses: reviewdog/action-golangci-lint@70498f2d1f75a55ee9a4d719e74e21ed68aebea3
```

leighmcculloch (commented):
💯 It would be great to switch to using golangci-lint in this repo.

Fyi we started using golangci-lint on some Go projects that do not live in this repo.
- https://github.com/stellar/starlight
- https://github.com/stellar/starbridge
- https://github.com/stellar/common-platform-backend

.github/workflows/reviewdog.yml
leighmcculloch, line 24: I wonder if we need reviewdog. The `golangci/golangci-lint` action that we're using below posts comments into pull request checks too.

https://github.com/stellar/starlight/blob/919096552e44eec5ec3fd6278922396926239874/.github/workflows/sdk.yml#L20-L25

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,33 @@
...
+jobs:
+  golangci:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b # version v3.0.2
+        with:
+          fetch-depth: 0 # required for new-from-rev option in .golangci.yml
+      - name: Setup GO
+        uses: actions/setup-go@268d8c0ca0432bb2cf416faae41297df9d262d7f # version v3.3.0
```
leighmcculloch, line 22: No need to pin GitHub owned actions, we only need to pin third party actions.

If GitHub's own action repos change it is no different to their runner-image repo changing which is the image all our code runs on, so no need to pin. Using `v3` for both should suffice.

leighmcculloch (commented):
One ask, otherwise LGTM.

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,33 @@
...
+      - master
+  pull_request:
+
+permissions:
+  contents: read
+  # Optional: allow read access to pull request. Use with `only-new-issues` option.
+  pull-requests: read
+
+jobs:
+  golangci:
```
leighmcculloch, line 14: One more thing, if you want this to be a required check, add a `complete` job like this:
https://github.com/stellar/go/blob/35023acd3efbd7e52fadba34cfd1ff22ce4d677b/.github/workflows/go.yml#L9-L15

leighmcculloch (approved):

.github/workflows/golangci-lint.yml
tsachiherman: I was hoping to phase it slowly, as I don't want to break anything. I want to see that people are getting feedback from this first, and take some feedback before making this mandatory.

leighmcculloch: Yup no worries, good idea. Just mentioning that this is how we hook into required checks, we don't add required checks for the actual jobs themselves if we don't have to.

PR #4606, opened 2022-09-23
ingest: Add support for tx meta v3

### LINKED ISSUE #4610: horizon: support TxMetaV3
Update ingestion and the transaction endpoint to provide TransactionMetaV3 (which provides the contract events). Otherwise Horizon can't provide event information to soroban-rpc.

Refer to #4600 for parent issue and context of overall horizon support ites.

bartekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Adds support for `LedgerCloseMeta.V3`.

### Why

New meta version is emitted in protocol 20.
Closes #4610 

### Known limitations

If new operations added in protocol 20 are added to the ledger, ingestion will halt.


tamirms (approved):

paulbellamy (approved):

ingest/ledger_transaction.go
```diff
@@ -76,6 +75,27 @@ func (t *LedgerTransaction) GetChanges() ([]Change, error) {
 
 		txChangesAfter := GetChangesFromLedgerEntryChanges(v2Meta.TxChangesAfter)
 		changes = append(changes, txChangesAfter...)
+	case 3:
```
sreuland, line 78: does this code get executed by integration tests with core on ver 20? minor suggestion, if you think worthwhile to add simple unit test in ledger_transaction_test.go to demonstrate coverage on this V3 block?

ingest/ledger_transaction.go
```diff
@@ -117,6 +137,14 @@ func (t *LedgerTransaction) GetOperationChanges(operationIndex uint32) ([]Change
 
 		v2Meta := t.UnsafeMeta.MustV2()
 		changes = operationChanges(v2Meta.Operations, operationIndex)
+	case 3:
```
tsachiherman, line 140: unless you're expecting that this function would need completely new logic for each of the meta versions, I'd suggest you'd reimplement this method so that it would avoid code duplication.

ingest/ledger_transaction.go
tsachiherman: same as my other comment - consider writing this method is a way that would avoid code duplication.

xdr/ledger_close_meta.go
```diff
@@ -1,21 +1,148 @@
...
+		panic(fmt.Sprintf("Unsupported LedgerCloseMeta.V: %d", l.V))
+	}
+}
+
+func (l LedgerCloseMeta) TransactionEnvelopes() []TransactionEnvelope {
+	switch l.V {
+	case 0:
+		return l.MustV0().TxSet.Txs
+	case 1, 2:
+		var envelopes []TransactionEnvelope
```
tsachiherman, line 56: maybe preallocate this array to avoid reallocations ?

tsachiherman (commented):
I think that as our produce grow, `LedgerTransaction` here becomes more clogged with `UnsafeMeta` variations.
I would suggest the following - 
keep `LedgerTransaction` structure as is. Add `LedgerTransactionInterface` as an interface which would implement the various operations we have on the `LedgerTransactions`.
Add 3 more structs : `LedgerTransactionV1, LedgerTransactionV2, LedgerTransactionV3`.
have these structs implement the `LedgerTransactionInterface`. each one of these would only need to support the relevant changes.
Ideally, this would allow us to eliminate most of the versioning-switching.
what do you think ?

sreuland (approved):

bartekn:
> I think that as our produce grow, `LedgerTransaction` here becomes more clogged with `UnsafeMeta` variations.
> I would suggest the following - 
> keep `LedgerTransaction` structure as is. Add `LedgerTransactionInterface` as an interface which would implement the various operations we have on the `LedgerTransactions`.
> Add 3 more structs : `LedgerTransactionV1, LedgerTransactionV2, LedgerTransactionV3`.
> have these structs implement the `LedgerTransactionInterface`. each one of these would only need to support the relevant changes.
> Ideally, this would allow us to eliminate most of the versioning-switching.
> what do you think ?

@tsachiherman  The point of `LedgerTransaction` is to hide all raw xdr structs from outside developers - it is meant to be as developer friendly as possible. I'm afraid that if we start returning three versions (and more in the future) through public `ingest` API it will cause unnecessary confusion.

We could consider creating private structs (`ledgerTransactionV1`, ...)  and return `LedgerTransactionInterface` in public API. However, given that it's possible to decrease code duplication in many cases in version-switches (I implemented it in recent commits - great suggestion!) I think we should probably stay with the current code.

Let me know what you think.

xdr/ledger_close_meta.go
```diff
@@ -1,21 +1,148 @@
...
+func (l LedgerCloseMeta) LedgerHeaderHistoryEntry() LedgerHeaderHistoryEntry {
+	switch l.V {
+	case 0:
+		return l.MustV0().LedgerHeader
+	case 1:
+		return l.MustV1().LedgerHeader
+	case 2:
+		return l.MustV2().LedgerHeader
+	default:
+		panic(fmt.Sprintf("Unsupported LedgerCloseMeta.V: %d", l.V))
```
tsachiherman, line 14: shouldn't the object passed to panic be an error object ?

tsachiherman:
> > I think that as our produce grow, `LedgerTransaction` here becomes more clogged with `UnsafeMeta` variations.
> > I would suggest the following -
> > keep `LedgerTransaction` structure as is. Add `LedgerTransactionInterface` as an interface which would implement the various operations we have on the `LedgerTransactions`.
> > Add 3 more structs : `LedgerTransactionV1, LedgerTransactionV2, LedgerTransactionV3`.
> > have these structs implement the `LedgerTransactionInterface`. each one of these would only need to support the relevant changes.
> > Ideally, this would allow us to eliminate most of the versioning-switching.
> > what do you think ?
> 
> @tsachiherman The point of `LedgerTransaction` is to hide all raw xdr structs from outside developers - it is meant to be as developer friendly as possible. I'm afraid that if we start returning three versions (and more in the future) through public `ingest` API it will cause unnecessary confusion.
> 
> We could consider creating private structs (`ledgerTransactionV1`, ...) and return `LedgerTransactionInterface` in public API. However, given that it's possible to decrease code duplication in many cases in version-switches (I implemented it in recent commits - great suggestion!) I think we should probably stay with the current code.
> 
> Let me know what you think.

I think that hiding it internally is perfectly fine. We re-evaluate this refactoring at a later stage. hopefully, that would be before we have version 4.

bartekn:
Merged with tests failing as agreed in @stellar/horizon-committers meeting. Going to revisit failing tests once a new Stellar-Core build is pushed to package repo.

PR #4607, opened 2022-09-23
ingest: add error code checks on few unit tests

tsachiherman:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This PR add some validation for functions returned error codes that were ignored in the unit test.

### Why

This PR provide some trivial fixes for issues identified by using the linters.

### Known limitations

Changes are intuitive & trivial.

### DIFF
--- a/ingest/checkpoint_change_reader_test.go
+++ b/ingest/checkpoint_change_reader_test.go
@@ -757,7 +757,7 @@ func entryAccount(t xdr.BucketEntryType, id string, balance uint32) xdr.BucketEn
 			Type: xdr.BucketEntryTypeDeadentry,
 			DeadEntry: &xdr.LedgerKey{
 				Type:    xdr.LedgerEntryTypeAccount,
-				Account: &xdr.LedgerKeyAccount{xdr.MustAddress(id)},
+				Account: &xdr.LedgerKeyAccount{AccountId: xdr.MustAddress(id)},
 			},
 		}
 	default:
--- a/ingest/ledger_transaction_test.go
+++ b/ingest/ledger_transaction_test.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/stellar/go/xdr"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestChangeAccountChangedExceptSignersInvalidType(t *testing.T) {
@@ -13,7 +14,9 @@ func TestChangeAccountChangedExceptSignersInvalidType(t *testing.T) {
 	}
 
 	assert.Panics(t, func() {
-		change.AccountChangedExceptSigners()
+		_, err := change.AccountChangedExceptSigners()
+		// the following is here only to avoid false-positive warning by the linter.
+		require.NoError(t, err)
 	})
 }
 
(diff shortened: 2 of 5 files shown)


ingest/ledgerbackend/remote_captive_core_test.go
```diff
@@ -23,11 +24,14 @@ func TestGetLedgerSucceeds(t *testing.T) {
...
 	called := 0
+	var encodeFailed int64
 	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		called++
-		json.NewEncoder(w).Encode(LedgerResponse{
+		if nil != json.NewEncoder(w).Encode(LedgerResponse{
 			Ledger: Base64Ledger(expectedLedger),
-		})
+		}) {
+			atomic.AddInt64(&encodeFailed, 1)
```
tsachiherman, line 33: this needs to be done via atomics, since the function is being called via a goroutine, and we cannot call any of the testing.T methods from a different goroutine.

ingest/ledger_transaction_test.go
```diff
@@ -13,7 +14,9 @@ func TestChangeAccountChangedExceptSignersInvalidType(t *testing.T) {
 	}
 
 	assert.Panics(t, func() {
-		change.AccountChangedExceptSigners()
+		_, err := change.AccountChangedExceptSigners()
+		// the following is here only to avoid false-positive warning by the linter.
+		require.NoError(t, err)
```
paulbellamy, line 19: this invalidates the `assert.Panics` wrapper, as the require here might panic.

ingest/checkpoint_change_reader_test.go
```diff
@@ -757,7 +757,7 @@ func entryAccount(t xdr.BucketEntryType, id string, balance uint32) xdr.BucketEn
 			Type: xdr.BucketEntryTypeDeadentry,
 			DeadEntry: &xdr.LedgerKey{
 				Type:    xdr.LedgerEntryTypeAccount,
-				Account: &xdr.LedgerKeyAccount{xdr.MustAddress(id)},
+				Account: &xdr.LedgerKeyAccount{AccountId: xdr.MustAddress(id)},
```
paulbellamy, line 760: I think this is good, but one risk is that, if you do:
`MyStruct{foo}` in go, and MyStruct has 2 fields, the code will fail to compile, while if you do named field, it will silently ignore the other fields. This introduces risk if refactoring later and adding fields to the struct. Overall, probably nicer, though.

paulbellamy (approved):
One note to fix, but lgtm, other than that

ingest/checkpoint_change_reader_test.go
tsachiherman: It's not really for looking nicer - it's so that we can ensure that the value get assigned to the correct variable in case we change the variables order in the struct definition. As a sidenote, it also eliminates a linter error..

ingest/ledgerbackend/remote_captive_core_test.go
sreuland: if encoding did fail in the mocked handler, ln 44 would still fail correct? If so, would seem adequate for unit test purpose, i.e. verify inputs/outputs of target code. this seems to introduce bit more complexity by verifying mock state also, which does help debug diagnostics, wanted to mention for consideration as another option, just log out the encoding err from the mock handler?

tsachiherman: not necessarily. The encoded bytes could still be correct while the Encode method return an error.
( shouldn't happen, clearly. but that's what we're testing here )

tamirms (approved):

Issue #4595, opened 2022-09-19
services/soroban-rpc: getTransactionStatus method


tsachiherman:
Depends on https://github.com/stellar/stellar-core/issues/3484, which is merged now.

This API should follow the soroban-rpc API Docs on Confluence, so that devs can swap out API server they are using `soroban-rpc`/`cli serve`/`horizon` seamlessly.

## getTransactionStatus

Clients will poll this to tell when the transaction has been completed.

TODO: add a websocket version for pubsub
TODO: Figure out result/error and how that fits into the jsonrpc response object

Parameters:

- `<hash>` - transaction hash to query

Returns:

- `<object>`
  - `id: <hash>` - hash of the transaction as a hex-encoded string
  - `status: <status>` - the current status of the transaction by hash. (`pending`, `success`, `error`)
  - `results: <object[]>` - (optional) Will be present on completed successful transactions.
    - `xdr: <xdr.ScVal>` - xdr-encoded return value of the contract call
    - TODO: Anything else we should include? Gas spent?
    - TODO: How does this work for classic transactions?
  - `error: <object>` - (optional) Will be present on failed transactions.
    - `code: <string>` - Short unique string representing the type of error
    - `message: <string>` - Human friendly summary of the error
    - `data: <object>` - (optional) More data related to the error if available

PR #4609, opened 2022-09-26
  exp/services/soroban-rpc: Add getTransactionStatus and sendTransaction json rpc methods

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add json rpc methods for sending a transaction and querying a transaction status. This is implemented by proxying to the horizon txsub endpoint and the horizon transaction detail endpoint.

### Why

https://github.com/stellar/go/issues/4595
https://github.com/stellar/go/issues/4595

### Known limitations

* As a shortcut to save time ahead of meridian, we proxy transaction submissions to Horizon instead of submitting directly to captive core
* Pending transactions and transaction submission validation errors are stored in an memory data structure 
* I have not yet tested creating soroban smart contracts or invoking host functions because horizon does not support that functionality yet
* No websocket functionality implemented, so the client will need to poll the getTransactionStatus method

### DIFF
--- a/exp/services/soroban-rpc/internal/methods/transaction.go
+++ b/exp/services/soroban-rpc/internal/methods/transaction.go
@@ -3,23 +3,25 @@ package methods
 import (
 	"context"
 	"encoding/hex"
+	"fmt"
 	"net/http"
 	"sync"
 	"time"
 
 	"github.com/creachadair/jrpc2"
-	"github.com/creachadair/jrpc2/code"
 	"github.com/creachadair/jrpc2/handler"
 
 	"github.com/stellar/go/clients/horizonclient"
 	"github.com/stellar/go/network"
 	"github.com/stellar/go/protocols/horizon"
+	"github.com/stellar/go/support/render/problem"
 	"github.com/stellar/go/xdr"
 )
 
 const (
-	TransactionComplete = "complete"
-	TransactionPending  = "pending"
+	TransactionSuccess = "success"
+	TransactionPending = "pending"
+	TransactionError   = "error"
 )
 
 type SendTransactionRequest struct {
@@ -34,16 +36,19 @@ type TransactionStatusResponse struct {
 	ID     string               `json:"id"`
 	Status string               `json:"status"`
 	Result *horizon.Transaction `json:"result"`
+	Error  *problem.P           `json:"error"`
 }
 
 type SendTransactionResponse struct {
-	ID string `json:"id"`
+	ID     string     `json:"id"`
+	Status string     `json:"status"`
+	Error  *problem.P `json:"error"`
 }
 
 type transactionResult struct {
 	timestamp time.Time
 	pending   bool
-	err       error
+	err       *problem.P
 }
 
 type horizonRequest struct {
@@ -52,57 +57,77 @@ type horizonRequest struct {
 }
 
 type TransactionProxy struct {
-	lock            sync.RWMutex
-	results         map[string]transactionResult
-	client          *horizonclient.Client
-	passphrase      string
-	queue           chan horizonRequest
-	workers         int
-	expiryFrequency time.Duration
-	ttl             time.Duration
+	lock       sync.RWMutex
+	results    map[string]transactionResult
+	client     *horizonclient.Client
+	passphrase string
+	queue      chan horizonRequest
+	workers    int
+	ttl        time.Duration
+	ctx        context.Context
+	cancel     context.CancelFunc
+	wg         sync.WaitGroup
 }
 
 func NewTransactionProxy(
 	client *horizonclient.Client,
 	workers, queueSize int,
 	networkPassphrase string,
-	expiryFrequency, ttl time.Duration) *TransactionProxy {
+	ttl time.Duration,
+) *TransactionProxy {
+	if workers > queueSize {
+		queueSize = workers
+	}
+	ctx, cancel := context.WithCancel(context.Background())
 	return &TransactionProxy{
-		results:         map[string]transactionResult{},
-		client:          client,
-		passphrase:      networkPassphrase,
-		queue:           make(chan horizonRequest, queueSize),
-		workers:         workers,
-		expiryFrequency: expiryFrequency,
-		ttl:             ttl,
+		results:    map[string]transactionResult{},
+		client:     client,
+		passphrase: networkPassphrase,
+		queue:      make(chan horizonRequest, queueSize),
+		workers:    workers,
+		ttl:        ttl,
+		ctx:        ctx,
+		cancel:     cancel,
 	}
 }
 
-func (p *TransactionProxy) Start(ctx context.Context) {
+func (p *TransactionProxy) Start() {
 	for i := 0; i < p.workers; i++ {
-		go p.startWorker(ctx)
+		p.wg.Add(1)
+		go p.startWorker(p.ctx)
 	}
-
-	go p.startExpiryTicker(ctx)
 }
 
-func (p *TransactionProxy) SendTransaction(ctx context.Context, request SendTransactionRequest) (SendTransactionResponse, error) {
+func (p *TransactionProxy) Close() {
+	p.cancel()
+	p.wg.Wait()
+}
+
+func (p *TransactionProxy) SendTransaction(ctx context.Context, request SendTransactionRequest) SendTransactionResponse {
+	defer p.deleteExpiredEntries(time.Now())
+
 	var envelope xdr.TransactionEnvelope
 	err := xdr.SafeUnmarshalBase64(request.Transaction, &envelope)
 	if err != nil {
-		return SendTransactionResponse{}, (&jrpc2.Error{
-			Code:    code.InvalidParams,
-			Message: "Invalid transaction XDR",
-		}).WithData(err)
+		return SendTransactionResponse{
+			Status: TransactionError,
+			Error: problem.MakeInvalidFieldProblem(
+				"transaction",
+				fmt.Errorf("cannot unmarshall transaction: %v", err),
+			),
+		}
 	}
 
 	var hash [32]byte
 	hash, err = network.HashTransactionInEnvelope(envelope, p.passphrase)
 	if err != nil {
-		return SendTransactionResponse{}, (&jrpc2.Error{
-			Code:    code.InvalidParams,
-			Message: "Transaction could not be hashed",
-		}).WithData(err)
+		return SendTransactionResponse{
+			Status: TransactionError,
+			Error: problem.MakeInvalidFieldProblem(
+				"transaction",
+				fmt.Errorf("cannot hash transaction: %v", err),
+			),
+		}
 	}
 	txHash := hex.EncodeToString(hash[:])
 
@@ -114,21 +139,26 @@ func (p *TransactionProxy) SendTransaction(ctx context.Context, request SendTran
 	// response
 	if result.pending || (ok && result.err == nil) {
 		return SendTransactionResponse{
-			ID: txHash,
-		}, nil
+			ID:     txHash,
+			Status: TransactionPending,
+		}
 	}
 
 	p.results[txHash] = transactionResult{pending: true}
 	select {
 	case p.queue <- horizonRequest{txHash: txHash, transactionXDR: request.Transaction}:
 		return SendTransactionResponse{
-			ID: txHash,
-		}, nil
+			ID:     txHash,
+			Status: TransactionPending,
+		}
 	default:
 		delete(p.results, txHash)
-		return SendTransactionResponse{}, &jrpc2.Error{
-			Code:    code.InternalError,
-			Message: "Request queue is full",
+		problemErr := problem.ServerError
+		problemErr.Detail = "Transaction queue is full"
+		return SendTransactionResponse{
+			ID:     txHash,
+			Status: TransactionError,
+			Error:  &problemErr,
 		}
 	}
 }
@@ -146,6 +176,7 @@ func (p *TransactionProxy) deletePendingEntry(txHash string) {
 }
 
 func (p *TransactionProxy) startWorker(ctx context.Context) {
+	defer p.wg.Done()
 	for {
 		select {
 		case <-ctx.Done():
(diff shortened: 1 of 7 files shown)


exp/services/soroban-rpc/internal/methods/transaction.go
```diff
@@ -0,0 +1,247 @@
...
+	expiryFrequency time.Duration
+	ttl             time.Duration
+}
+
+func NewTransactionProxy(
+	client *horizonclient.Client,
+	workers, queueSize int,
+	networkPassphrase string,
+	expiryFrequency, ttl time.Duration) *TransactionProxy {
+	return &TransactionProxy{
```
tsachiherman, line 70: you might want to adjust the queueSize in case workers > queueSize.

exp/services/soroban-rpc/internal/methods/transaction.go
```diff
@@ -0,0 +1,247 @@
...
+		ttl:             ttl,
+	}
+}
+
+func (p *TransactionProxy) Start(ctx context.Context) {
+	for i := 0; i < p.workers; i++ {
+		go p.startWorker(ctx)
+	}
+
+	go p.startExpiryTicker(ctx)
```
tsachiherman, line 86: I think that expiry ticker here is a bit of an overkill ( but functionally correct). you could easily prune the expired entries during SendTransaction. 
i.e. creating a go routine that is going to be idle most of the time isn't great.

tsachiherman (changes requested):
Overall looks good. Beside the notes added, I have one more request:
can you add a Stop()/Shutdown() that would cancel the context and block until all the child go routines are done ?
i.e. use sync/workgroup + child context so you could shut it down.
The goal here is to allow deterministic cleanup.

exp/services/soroban-rpc/internal/methods/transaction.go
```diff
@@ -0,0 +1,247 @@
...
+		}
+	} else {
+		return TransactionStatusResponse{
+			ID:     request.Hash,
+			Status: TransactionComplete,
+			Result: &tx,
+		}, nil
+	}
+
+	p.lock.RLock()
```
tsachiherman, line 194: could you add a comment here -
// herr.Problem.Status == http.StatusNotFound
it was very tricky for me to figure this out by reading it.

exp/services/soroban-rpc/main.go
```diff
@@ -73,20 +101,33 @@ func main() {
...
+			ctx, cancel := context.WithCancel(context.Background())
 			supporthttp.Run(supporthttp.Config{
 				ListenAddr: fmt.Sprintf(":%d", port),
 				Handler:    handler,
 				OnStarting: func() {
 					logger.Infof("Starting Soroban JSON RPC server on %v", port)
+					transactionProxy.Start(ctx)
 				},
 				OnStopping: func() {
+					cancel()
```
tsachiherman, line 130: I think that you would want to call `cancel` after `handler.Close()`, since you don't want any request to go from the handler to the proxy *after* it's being canceled.

exp/services/soroban-rpc/main.go
```diff
@@ -73,20 +101,33 @@ func main() {
...
 			})
 			if err != nil {
 				logger.Fatalf("could not create handler: %v", err)
 			}
+			ctx, cancel := context.WithCancel(context.Background())
 			supporthttp.Run(supporthttp.Config{
 				ListenAddr: fmt.Sprintf(":%d", port),
 				Handler:    handler,
 				OnStarting: func() {
 					logger.Infof("Starting Soroban JSON RPC server on %v", port)
```
tsachiherman, line 126: the port here should be pull from the actual TCP connection rather than the port argument.
Separately, could we consider changing the port into endpoint, for security reasons, and default it to "localhost:8000" ? that way, the end user can select the interface he would want to bind.

exp/services/soroban-rpc/internal/methods/transaction.go
```diff
@@ -0,0 +1,247 @@
...
+}
+
+type GetTransactionStatusRequest struct {
+	Hash string `json:"hash"`
+}
+
+type TransactionStatusResponse struct {
+	ID     string               `json:"id"`
+	Status string               `json:"status"`
+	Result *horizon.Transaction `json:"result"`
```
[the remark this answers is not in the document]
sreuland: > Also, error handling here?

+1, if not a big stretch to populate `error` as optionally mentioned in the api docs, seems like could help diagnostics for clients.

exp/services/soroban-rpc/internal/methods/transaction.go
```diff
@@ -0,0 +1,247 @@
...
+}
+
+type SendTransactionResponse struct {
+	ID string `json:"id"`
```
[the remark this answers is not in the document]
sreuland: > it should be an error TransactionStatusResponse

updated [send transaction api docs](https://stellarorg.atlassian.net/wiki/spaces/PJC/pages/4402151440/soroban-rpc+API+Docs#sendTransaction-(Meridian-or-use-Horizon)) to capture the behavior mentioned here, i.e. for `status` and `error` on the tx responses.

exp/services/soroban-rpc/internal/methods/transaction.go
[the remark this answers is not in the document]
tamirms: ok I will try to follow that format but I will keep Result to have type `*horizon.Transaction` for now. Once we're able to submit soroban transactions through horizon I will update the `TransactionStatusResponse` to include the the smart contract return value. We will also need https://github.com/stellar/stellar-core/pull/3543 to be merged  to include the smart contract return value

exp/services/soroban-rpc/internal/methods/transaction.go
```diff
@@ -0,0 +1,247 @@
...
+				return TransactionStatusResponse{}, (&jrpc2.Error{
+					Code:    code.InvalidRequest,
+					Message: herr.Problem.Title,
+				}).WithData(herr.Problem.Extras)
+			}
+		} else {
+			return TransactionStatusResponse{}, err
+		}
+	} else {
+		return TransactionStatusResponse{
```
sreuland, line 187: could p.deletePendingEntry(request.Hash) be done at this point as optimization, since tx status is known now, to keep the pending queue size down, rather than letting it wait longer for the expired loop to purge it.

exp/services/soroban-rpc/internal/methods/transaction.go
```diff
@@ -0,0 +1,247 @@
...
+	hash, err = network.HashTransactionInEnvelope(envelope, p.passphrase)
+	if err != nil {
+		return SendTransactionResponse{}, (&jrpc2.Error{
+			Code:    code.InvalidParams,
+			Message: "Transaction could not be hashed",
+		}).WithData(err)
+	}
+	txHash := hex.EncodeToString(hash[:])
+
+	p.lock.Lock()
```
sreuland, line 109: in general, and not suggesting for now, unless you saw value in it and not a stretch, has a context-aware lock been considered for usage, like `go-lock` with `tryLockWithContext(ctx)` just curious with advanced threading, potential for getting locked up and not honoring the callers ctx.

tsachiherman: @sreuland where do we have a use case where we use `tryLockWithContext` ? I'm familiar with the concept, but always found it too of a headache to implement.

exp/services/soroban-rpc/internal/methods/transaction_test.go
```diff
@@ -0,0 +1,54 @@
...
+import (
+	"github.com/stretchr/testify/assert"
+	"testing"
+	"time"
+)
+
+func TestDeleteExpiredTransaction(t *testing.T) {
+	ttl := time.Minute
+	proxy := NewTransactionProxy(
+		nil,
```
sreuland, line 12: great test coverage btw, from this unit test layer, noticed may have been able to mock out `HTTP` in `horizonclient.Client`, and possibly exercise much of the new code paths, rather than system integration tests later, just thinking out loud, not suggesting change.

exp/services/soroban-rpc/internal/methods/transaction.go
sreuland: I was thinking on ln 109, if there were a mis-matchted lock/unlock at some point, then ensuing caller requests would get stuck there on the `Lock()`, also if the caller defined a ctx with timeout shorter than the http client timeout(is configurable, I think is default of 10s) and passes that in here, could be another potential way that caller's context is ignored if the http response takes longer.

exp/services/soroban-rpc/internal/methods/transaction.go
```diff
@@ -0,0 +1,281 @@
...
+	ID     string               `json:"id"`
+	Status string               `json:"status"`
+	Result *horizon.Transaction `json:"result"`
+	Error  *problem.P           `json:"error"`
+}
+
+type SendTransactionResponse struct {
+	ID     string     `json:"id"`
+	Status string     `json:"status"`
+	Error  *problem.P `json:"error"`
```
tsachiherman, line 45: could you add a code comment : 
// optional error in case status == error

exp/services/soroban-rpc/internal/methods/transaction.go
```diff
@@ -0,0 +1,281 @@
...
+		workers:    workers,
+		ttl:        ttl,
+		ctx:        ctx,
+		cancel:     cancel,
+	}
+}
+
+func (p *TransactionProxy) Start() {
+	for i := 0; i < p.workers; i++ {
+		p.wg.Add(1)
```
tsachiherman, line 96: nit: you could have called
```golang
p.wg.Add(p.workers)
```
outside of the loop.

exp/services/soroban-rpc/internal/methods/transaction.go
```diff
@@ -0,0 +1,281 @@
...
+		go p.startWorker(p.ctx)
+	}
+}
+
+func (p *TransactionProxy) Close() {
+	p.cancel()
```
tsachiherman, line 102: add code comments - 
```golang
// signal the worker go routines to abort
p.cancel()
// wait until the worker go routines are done
p.wg.Wait()
```

exp/services/soroban-rpc/internal/methods/transaction.go
```diff
@@ -0,0 +1,281 @@
...
+}
+
+func (p *TransactionProxy) startWorker(ctx context.Context) {
+	defer p.wg.Done()
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case request := <-p.queue:
+			_, err := p.client.SubmitTransactionXDR(request.transactionXDR)
```
tsachiherman, line 185: I'm surprised to see there is no context being passed to SubmitTransactionXDR; I would have expected to see that there is some sort of cancellation option.

exp/services/soroban-rpc/internal/methods/transaction.go
```diff
@@ -0,0 +1,281 @@
...
+				if herr, ok := err.(*horizonclient.Error); ok {
+					result.err = &herr.Problem
+				} else {
+					problemErr := problem.ServerError
+					problemErr.Detail = fmt.Sprintf("transaction submission failed: %v", err)
+					result.err = &problemErr
+				}
+				p.setTxResult(request.txHash, result)
+			} else {
+				p.deletePendingEntry(request.txHash)
```
tsachiherman, line 197: if we did send the transaction successfully, don't we want to mark it as "success" ?

exp/services/soroban-rpc/main.go
```diff
@@ -2,8 +2,10 @@ package main
 
 import (
 	"fmt"
+	"github.com/stellar/go/network"
```
tsachiherman, line 5: nit: move the the bottom import group.

exp/services/soroban-rpc/internal/methods/transaction.go
```diff
@@ -0,0 +1,281 @@
...
+	}
+}
+
+func (p *TransactionProxy) Close() {
+	p.cancel()
+	p.wg.Wait()
+}
+
+func (p *TransactionProxy) SendTransaction(ctx context.Context, request SendTransactionRequest) SendTransactionResponse {
+	defer p.deleteExpiredEntries(time.Now())
```
tsachiherman, line 107: I think it would be better to move this after taking the lock so that we won't take the lock twice within the function call.

exp/services/soroban-rpc/internal/methods/transaction.go
tsachiherman: I generally agree with your observation here; as you've noted, in this case the lock is taken only for a short amount of time, and never for a blocking operation ( therefore, I can't see any concrete potential issue here )

exp/services/soroban-rpc/internal/methods/transaction.go
tamirms: Yeah, I agree but the horizon client doesn’t support context passing. Otherwise I would have used it.

exp/services/soroban-rpc/internal/methods/transaction.go
tamirms: Could you clarify what you are proposing? The delete expired entries function currently acquires a lock. Are you suggesting to remove the locking entirely and have the caller of the function responsible for the thread safety?

tsachiherman: sure. Move the defer call from line 107 to line 136, and remove acquiring the lock from within `deleteExpiredEntries`.

sreuland: was curious why overall cache expiration purge is triggered here during request, could the purge be de-coupled from here and instead refactored as a timer loop started in `TransactionProxy.Start()` and stopped in `TransactionProxy.Close()` ?

sreuland (approved):
great approach and implementation to get this realized!

exp/services/soroban-rpc/internal/methods/transaction.go
tamirms: we don't want to store the successful response in the `results` map because horizon already will have the response available in its database.

That is why in `GetTransactionStatus()` the first thing we do is call `p.client.TransactionDetail(request.Hash)` which will query horizon for the response.

tsachiherman (approved):

PR #4615, opened 2022-09-27
exp/lighthorizon: Add parameters to preload ledger cache.

Shaptic:
### What
This adds two new command line parameters:
 - `--ledger-cache-preload=<count>`, which starts a goroutine in the background to fill the cache with `count` ledgers
 - `--ledger-cache-preload-start=<ledger>`, which specifies the ledger to start building the cache at. By default, it will just count backwards from the latest ledger stored at the ledger source.

### Why
#4526

### Known limitations
n/a


exp/lighthorizon/tools/index.go
```diff
@@ -124,7 +124,7 @@ view s3:///indices GAXLQGKIUAIIUHAX4GJO3J7HFGLBCNF6ZCZSTLJE7EKO5IUHGLQLMXZO
 view file:///tmp/indices --limit=0 GAXLQGKIUAIIUHAX4GJO3J7HFGLBCNF6ZCZSTLJE7EKO5IUHGLQLMXZO
 view gcs://indices --limit=10 GAXLQGKIUAIIUHAX4GJO3J7HFGLBCNF6ZCZSTLJE7EKO5IUHGLQLMXZO,GBUUWQDVEEXBJCUF5UL24YGXKJIP5EMM7KFWIAR33KQRJR34GN6HEDPV,GBYETUYNBK2ZO5MSYBJKSLDEA2ZHIXLCFL3MMWU6RHFVAUBKEWQORYKS`,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			if len(args) > 2 {
+			if len(args) < 1 || len(args) > 2 {
```
tsachiherman, line 127: I'm all for using the strongest predicate you can - but writing `Len(args) < 1` is a bit misleading, when you really wanted to test `len(args) == 0`. I'd advise you'd use either `len(args) == 0` or `len(args) <= 0` so that the zero value would be there explicitly.

exp/lighthorizon/main.go
```diff
@@ -96,6 +97,27 @@ break down accounts by active ledgers.`,
...
+			cachePreloadCount, _ := cmd.Flags().GetUint32("ledger-cache-preload")
+			cachePreloadStart, _ := cmd.Flags().GetUint32("ledger-cache-preload-start")
+			if cacheDir != "" && cachePreloadCount > 0 {
+				startLedger := latestLedger - cachePreloadCount
+				if cachePreloadStart > 0 {
+					startLedger = cachePreloadStart
+				}
+
+				go func() {
+					tools.BuildCache(sourceUrl, cacheDir,
```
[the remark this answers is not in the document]
Shaptic: This is a good point, but I'd like to make a counterargument: If the cache build fails, what should we do? It's not a critical error, so probably nothing. It already does logging, so there's a trace of what went wrong, but it doesn't seem to me that there's a meaningful notion of "handling" the error.

You're right with regards to process aborting, though. If the process gets aborted, the goroutine goes down with it, but we are running the risk of the cache being corrupted. Admittedly, if an abort occurs (i.e. an _unexpected_ shutdown), you'd probably need manual intervention (hence the existence of `cache purge <ledger>`), but support for a _graceful_ shutdown should definitely be added here.

Unfortunately, though, that's a bigger change, because we'd need to add signal handling in general, and that's absent from the main service anyway. (In other words, if you ctrl+c the web server while it's doing _normal_ caching (i.e. not the cache preload in question here), you'd still risk cache corruption.)

How do you feel about capturing that in a larger issue, e.g. "exp/lighthorizon: Add graceful shutdowns."?

exp/lighthorizon/tools/index.go
Shaptic: That's an interesting perspective. Why does 0 need to be explicit? If I needed exactly 2 or 3 arguments, I'd put `< 2 || > 3`, right? So it's the same rationale here, since I need exactly 1 or 2, not necessarily "more than zero." I guess I could say `if len(args) != 1 && len(args) != 2`, but again, I'm not sure why the zero needs to be explicit.

exp/lighthorizon/main.go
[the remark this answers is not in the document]
tsachiherman: I'm perfectly fine with that.

exp/lighthorizon/tools/index.go
tsachiherman: I think that my rationale is that `len(args)` is always a natural number, and writing `len(args) < 1` picks the entire range that is less than one - implying a single entry. Clearly, there is no logical flaw here.
If you were to explain to me, verbally, what this function is doing - would you say "number of arguments is less than 1" or "no arguments" ?

Shaptic: I'd say neither, heh. I'd say "at least one argument, but no more than two".

exp/lighthorizon/main.go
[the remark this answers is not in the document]
Shaptic: Done! #4619.

exp/lighthorizon/main.go
```diff
@@ -96,6 +97,27 @@ break down accounts by active ledgers.`,
...
+			latestLedger, err := ingester.GetLatestLedgerSequence(context.Background())
+			if err != nil {
+				log.Fatalf("Failed to retrieve latest ledger from %s: %v", sourceUrl, err)
+				return
+			}
+			log.Infof("The latest ledger stored at %s is %d.", sourceUrl, latestLedger)
+
+			cachePreloadCount, _ := cmd.Flags().GetUint32("ledger-cache-preload")
+			cachePreloadStart, _ := cmd.Flags().GetUint32("ledger-cache-preload-start")
+			if cacheDir != "" && cachePreloadCount > 0 {
```
sreuland, line 109: is there a way to mark some flags as dependent on others? i.e. `cacheDir` 
 is now required when setting the new `cachePreloadCount` and `cachePreloadStart`, was curious if there's a way to express that in cmd config or maybe just guard check it manually upfront here at same time for end user feedback, otherwise if user sets `cachePreloadCount`  and not `cacheDir` , then this will silently do no caching which isn't user intent?

```
if cacheDir == "" && cachePreloadCount > 0 {
   log.Fatal("cacheDir required ")
   return
}
```

Shaptic: Good idea! Added in 6c29639.

Shaptic: Do you think it should be `Fatal` or `Warn`?

sreuland: I'd treat as a blocking validation error for immediate feedback to user to fix, so `Fatal`

exp/lighthorizon/main.go
```diff
@@ -96,6 +97,36 @@ break down accounts by active ledgers.`,
...
+				return
+			}
+			log.Infof("The latest ledger stored at %s is %d.", sourceUrl, latestLedger)
+
+			cachePreloadCount, _ := cmd.Flags().GetUint32("ledger-cache-preload")
+			cachePreloadStart, _ := cmd.Flags().GetUint32("ledger-cache-preload-start")
+			if cachePreloadCount > 0 {
+				if cacheDir == "" {
+					log.Fatalf("--ledger-cache-preload=%d specified but no "+
+						"--ledger-cache directory provided, ignoring...",
```
sreuland, line 112: i think just need to remove the `ignoring` part of message correct? b/c it's exiting the process early.

exp/lighthorizon/build/k8s/lighthorizon_web.yml
```diff
@@ -7,11 +7,13 @@ metadata:
...
   name: lighthorizon-pubnet-web-env
 data:
-  TXMETA_SOURCE: "s3://horizon-ledgermeta-prodnet-test"
-  INDEXES_SOURCE: "s3://horizon-index-prodnet-test"
+  TXMETA_SOURCE: "s3://horizon-indices-pubnet"
+  INDEXES_SOURCE: "s3://horizon-ledgermeta-pubnet"
   NETWORK_PASSPHRASE: "Public Global Stellar Network ; September 2015"
   MAX_PARALLEL_DOWNLOADS: 16
   CACHE_PATH: "/ledgercache"
+  CACHE_PRELOAD_START_LEDGER: 0
```
sreuland, line 15: could remove preload start env property, as cmd defaults its absence to same value?

exp/lighthorizon/services/main.go
```diff
@@ -133,6 +133,11 @@ func searchAccountTransactions(ctx context.Context,
 
 		start := time.Now()
 		ledger, innerErr := config.Ingester.GetLedger(ctx, nextLedger)
+
```
sreuland, line 136: yes, have been shuffling this around, created a ticket to get visibility on it for prioritization:

https://github.com/stellar/go/issues/4657

exp/lighthorizon/main.go
```diff
@@ -96,6 +97,36 @@ break down accounts by active ledgers.`,
...
+						cachePreloadCount)
+					return
+				} else {
+					startLedger := latestLedger - cachePreloadCount
```
sreuland, line 116: should this be sanity checked for at least > 0 and if negative set to 0 i.e. beginning and log a WARN?

sreuland (approved):

Issue #4593, opened 2022-09-19
services/soroban-rpc: simulateTransaction


tsachiherman:
Depends on https://github.com/stellar/stellar-core/issues/3484, which is merged now.

This API should follow the soroban-rpc API Docs on Confluence, so that devs can swap out API server they are using `soroban-rpc`/`cli serve`/`horizon` seamlessly.

## simulateTransaction

Submit a trial contract invocation to get back return values, expected ledger footprint, and expected costs.

TODO: Should/can we offer to auto-calculate the footprint as part of sendTransaction to save round-trips? Solana does this, but it might conflict with our signature scheme?

Parameters:

- `<xdr.Transaction>` - The transaction to be simulated

Returns:

- `<object>`
  - `cost: <object>` - Information about the fees expected, instructions used, etc.
    - `cpuInsns: <string>` - Stringified-number of the total cpu instructions consumed by this transaction
    - `memBytes: <string>` - Stringified-number of the total memory bytes allocated by this transaction
  - `footprint: <object>`
    - `readOnly: <xdr.LedgerKey[]>` - Array of ledger keys expected to be read by this transaction
    - `readWrite: <xdr.LedgerKey[]>` - Array of ledger keys expected to be written by this transaction.
  - `results: <object[]>`
    - `xdr: <xdr.ScVal>` - (optional) Only present on success. xdr-encoded return value of the contract call
  - `latestLedger: <number>` - The current latest ledger observed by the node when this response was generated.

Possible Errors:

- TODO: Figure out failure modes.
- Call panicked/rejected/trapped (whatever we’re calling it)

2opremio:
@tamirms can we close this?

PR #4621, opened 2022-09-30
exp/services/soroban-rpc: Add simulate transaction endpoint

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add simulate transaction endpoint to the soroban JSON RPC service. This is implemented by proxying the request to stellar core.

### Why

https://github.com/stellar/go/issues/4593

### Known limitations

The stellar core response doesn't include a `latestLedger` field in the response.

https://github.com/stellar/stellar-core/pull/3500/files#diff-12e712ade191c619e26f360d118226148677a891b5d23f2474ebe64c427351e9R219-R225


clients/stellarcore/client.go
```diff
@@ -52,6 +53,34 @@ func (c *Client) Upgrade(ctx context.Context, version int) error {
...
 
+// Preflight submits a preflight request to the stellar core instance. The response will contain the footprint
+// of the invoke host function operation among other information.
+func (c *Client) Preflight(ctx context.Context, invokeHostFunctionOp xdr.InvokeHostFunctionOp) (proto.PreflightResponse, error) {
+	b64, err := xdr.MarshalBase64(invokeHostFunctionOp)
+	if err != nil {
+		return proto.PreflightResponse{}, errors.Wrap(err, "failed to marshal invoke host function op")
+	}
+	q := url.Values{}
+	q.Set("blob", b64)
```
2opremio, line 64: Not directly related to your PR but … how come Core is using a query parameter? They official have a limit of 2KB (if I recall properly)

tamirms: yeah, I am also surprised by this. I would have expected the http request to be a POST but the transaction submission endpoint is also a GET request which passes the xdr as a query parameter (see https://github.com/stellar/stellar-core/blob/master/docs/software/commands.md , search for tx?blob=Base64 ).

regarding the 2k limit, that is enforced by browsers. IE had a limit of 2k but the actual protocol doesn't really define a limit: https://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers

exp/services/soroban-rpc/internal/methods/simulate_transaction.go
```diff
@@ -0,0 +1,55 @@
...
+type SimulateTransactionRequest struct {
+	InvokeHostFunctionOp string `json:"invoke_host_function_op"`
+}
+
+type SimulateTransactionCost struct {
+	CPUInstructions uint64 `json:"cpu_insns,string"`
+	MemoryBytes     uint64 `json:"mem_bytes,string"`
+}
+
+type SimulateTransactionResponse struct {
```
paulbellamy, line 21: These should probably match the api docs https://stellarorg.atlassian.net/wiki/spaces/PJC/pages/4402151440/soroban-rpc+API+Docs#simulateTransaction-(Meridian)

exp/services/soroban-rpc/internal/methods/simulate_transaction.go
```diff
@@ -0,0 +1,55 @@
...
+	Cost      SimulateTransactionCost `json:"cost"`
+}
+
+// NewSimulateTransactionHandler returns a json rpc handler to execute preflight requests to stellar core
+func NewSimulateTransactionHandler(coreClient *stellarcore.Client) jrpc2.Handler {
+	return handler.New(func(ctx context.Context, request SimulateTransactionRequest) SimulateTransactionResponse {
+		var xdrOp xdr.InvokeHostFunctionOp
+		if err := xdr.SafeUnmarshalBase64(request.InvokeHostFunctionOp, &xdrOp); err != nil {
+			return SimulateTransactionResponse{
+				Error: "Could unmarshal invoke host function op: " + err.Error(),
```
tsachiherman, line 34: could you wrap the inner error somehow so that we can still us errors.Is and error.As to find the underlying error ? 
maybe `fmt.Errorf` with `%w` ?

exp/services/soroban-rpc/internal/methods/simulate_transaction.go
```diff
@@ -0,0 +1,55 @@
...
+			}
+		}
+
+		coreResponse, err := coreClient.Preflight(ctx, xdrOp)
+		if err != nil {
+			return SimulateTransactionResponse{
+				Error: "Could not submit request to core: " + err.Error(),
```
tsachiherman, line 41: same as above; it's particularly important since the passed-in context could expired during the `Preflight` execution, making the `Preflight` return the `ctx.Err`. In that case, we need to return the explicit `ctx.Err` upstream to let the caller know that it's the context expiration that led to this failure.

exp/services/soroban-rpc/internal/test/simulate_transaction_test.go
```diff
@@ -0,0 +1,152 @@
...
+	}).BuildXDR()
+	assert.NoError(t, err)
+	return op
+}
+
+func TestSimulateTransactionSucceeds(t *testing.T) {
+	test := NewTest(t)
+
+	ch := jhttp.NewChannel(test.server.URL, nil)
+	cli := jrpc2.NewClient(ch, nil)
```
tsachiherman, line 89: this one confused me for a little bit. I used to call "cli" as in "command line interface"; it took me a while that it's the first three letters of "client" as well.

exp/services/soroban-rpc/internal/test/simulate_transaction_test.go
```diff
@@ -0,0 +1,152 @@
...
+
+	invokeHostOpB64, err := xdr.MarshalBase64(createInvokeHostOperation(t).Body.MustInvokeHostFunctionOp())
+	require.NoError(t, err)
+
+	request := methods.SimulateTransactionRequest{InvokeHostFunctionOp: invokeHostOpB64}
+	var result methods.SimulateTransactionResponse
+	err = cli.CallResult(context.Background(), "simulateTransaction", request, &result)
+	assert.NoError(t, err)
```
tsachiherman, line 97: Please add another test for CallResult passing in an expired context. The output of the call should be the ctx.Err value. ( a nice-to-have is to make a context expired during the http call, but it's a bit more tricky to implement ).

exp/services/soroban-rpc/main.go
```diff
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"github.com/stellar/go/clients/stellarcore"
```
tsachiherman, line 5: move to the group down.

tsachiherman (commented):
Few small requests; looks great otherwise.

exp/services/soroban-rpc/internal/methods/simulate_transaction.go
bartekn, line 41: Not sure if `soroban-rpc` will ever be publicly exposed but if so, `err.Error()` can contain some sensitive data, like Stellar-Core IP address. We could instead print a generic message and log the error message.

tsachiherman: I like this approach. Sensitive information should remain local, and return code should be client side actionable.

exp/services/soroban-rpc/internal/methods/simulate_transaction.go
paulbellamy: If you're going to also change soroban-cli, js-soroban-sdk, and soroban-example-dapp, be my guest.

paulbellamy: Oh, good point. I guess cli doesn't call it.

If it's only taking a single operation, maybe it should be `simulateOperation`, not `simulateTransaction` 🤔 But with auth, it does actually use information from the transaction, because the result can depend on the source account of the txn...

tsachiherman: I think that I'm leaning toward `simulateTransaction` for few reasons; please let me know what you think:
1. The end user is familiar with the construct of a "Transaction"; introducing another sub-object is less preferable.
2. Unlike etherium, we support multiple operations - which could be multiple smart contact invocations. The gas price we want to calculate is the *sum* of all of them.
3. Smart contracts within a single transaction can be reactive; i.e. the effect of the first one would make functional changes to the second one. Therefore, we must evaluate the entire transaction as a whole.
4. The number of round trips to evaluate a transaction would ideally be 1, rather then the number of operations.
@paulbellamy @tamirms

exp/services/soroban-rpc/internal/methods/simulate_transaction.go
```diff
@@ -0,0 +1,67 @@
...
+	proto "github.com/stellar/go/protocols/stellarcore"
+	"github.com/stellar/go/support/log"
+	"github.com/stellar/go/xdr"
+
+	"github.com/creachadair/jrpc2"
+	"github.com/creachadair/jrpc2/handler"
+)
+
+type SimulateTransactionRequest struct {
+	InvokeHostFunctionOp string `json:"invoke_host_function_op"`
```
2opremio, line 16: According to the RPC Server docs, this `simulateTransaction` takes a full transaction envelope.

I think that passing the function makes a lot more sense, but ... I think we should sync it with the `js-soroban-sdk` implementation.

CC @paulbellamy

exp/services/soroban-rpc/internal/methods/simulate_transaction.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+type SimulateTransactionCost struct {
+	CPUInstructions uint64 `json:"cpu_insns,string"`
+	MemoryBytes     uint64 `json:"mem_bytes,string"`
```
2opremio, line 21:
```suggestion
	CPUInstructions uint64 `json:"cpuInsns,string"`
	MemoryBytes     uint64 `json:"memBytes,string"`
```

tsachiherman: do we have any standard to follow picking json field names ? ( I'm not picky )
just wondering which of the `low_caps` / `pascalNotation` we should be using.

2opremio: The RPC Server docs (read spec) had these fields in  `camelCase`, and I think they use `camelCase` in general. See https://stellarorg.atlassian.net/wiki/spaces/PJC/pages/4402151440/soroban-rpc+API+Docs#simulateTransaction-(Meridian)

exp/services/soroban-rpc/internal/methods/simulate_transaction.go
2opremio: We decided to go for a `TransactionEnvelope`

.github/workflows/horizon.yml
```diff
@@ -35,7 +35,7 @@ jobs:
       HORIZON_INTEGRATION_TESTS_ENABLED: true
       HORIZON_INTEGRATION_TESTS_CORE_MAX_SUPPORTED_PROTOCOL: ${{ matrix.protocol-version }}
       PROTOCOL_20_CORE_DEBIAN_PKG_VERSION: 19.4.1-1085.fc63f3a6a.focal~soroban
```
sreuland, line 37: I think the deb pkg needs to be bumped also - 19.4.1-1093.9a33c7392.focal~soroban

sreuland (approved):

Issue #4679, opened 2022-10-31
exp/soroban-rpc: relocate soroban-rpc to soroban-tools repository


sreuland:
epic - https://github.com/stellar/go/issues/4672

### What would you like to see?
soroban rpc code is moved into it's own repo:

- [x] Rename soroban-cli repo to soroban-tools https://github.com/stellar/terraform/pull/1002
- [x] Copy Soroban RPC to soroban-CLI repo ( will be deleted from go repository at a later stage ) https://github.com/stellar/soroban-cli/pull/251
- [x] Add go repository as a dependency to soroban-rpc. https://github.com/stellar/soroban-cli/pull/251
  - [x] use go.mod to point to go repository
- [ ] Update Jenkins [soroban-package-builder](https://buildmeister-v3.stellar-ops.com/job/Horizon/job/soroban-package-builder/) to reference new soroban rpc code location
- [ ] Verify quickstart  [soroban-dev](https://github.com/stellar/quickstart/blob/master/Dockerfile.soroban-dev) image still works with latest build of Jenkins pkg
- [ ] Remove soroban-rpc from go repo

### What alternatives are there?
leave in mono-repo, rely on new tags to reference versions.

paulbellamy:
What are the issues that this is attempting to solve?

tomerweller:
> What are the issues that this is attempting to solve?

I'd like to reiterate this @paulbellamy's question. @tsachiherman ?

tsachiherman:
> > What are the issues that this is attempting to solve?
> 
> I'd like to reiterate @paulbellamy's question. @tsachiherman ?

Moving the soroban-rpc to the same place as the soroban-cli addresses several issues:
- the components are now gurenteed to be synchronized.
- the components can be tested to work together. 
- the soroban-rpc is going to use the soroban-env-host to run simulateTransaction. That's a rust package that would need to be wrapped. Ideally, it would live where we already have other rust components.
- less error prone : the soroban tools are all in one repository.

PR #4692, opened 2022-11-15
exp/services/soroban-rpc: Use Rust to obtain preflights

2opremio:
Instead of using Core (through its now deprecated `/preflight` HTTP endpoint) to obtain preflights, do it locally through the [`soroban-env-host`](https://github.com/stellar/rs-soroban-env/tree/main/soroban-env-host) library.

This is done through a small C-ABI layer communicating Go (using CGO) with Rust's FFI. The Rust code is compiled to a static library (`libpreflight.a`). The compilation is done through `go generate`

Unfortunately it doesn't still work since we need a way to supply `soroban-env-host` with:

* The raw Ledger entries it requests.
* Ledger information.

(all of the above marked with `TODO`s in the code)

~Additionally we need to decide whether we want to try to link the resulting binary statically. See https://github.com/2opremio/soroban-go-rust-preflight-poc/issues/3 for more details.~ 

I decided to link dynamically against glibc in Linux for now, since the `musl` targets (which allow linking truly statically) don't seem to allow unwinding Rust panics (See https://github.com/stellar/go/actions/runs/3484488730/jobs/5829099175 , in which you can see the rust call causing an ABORT signal)


sreuland:
do you think this should target the soroban-rpc source that has been relocated to soroban-tools:

https://github.com/stellar/soroban-tools/tree/main/cmd/soroban-rpc

the exp/services/soroban-rpc path will be removed, possibly during this sprint, need to confirm with @tsachiherman on timing though :
https://github.com/stellar/go/issues/4679

2opremio:
I can definitely move the PR there, thanks

2opremio:
@sreuland I can't move the PR just yet, the code is not in sync. See https://github.com/stellar/soroban-tools/pull/251#issuecomment-1324125295

2opremio:
https://github.com/stellar/go/pull/4698 (if successful) should be able to fulfill the remaining requirements of this PR.

2opremio:
I've migrated this PR to https://github.com/stellar/soroban-tools/pull/269 . 

This PR is not going to be updated further but I haven't closed it yet because some of the changes aren't migratable:

* Preflight method removal in the core client
* Test adjustments

I will apply those once we remove RPC-server from this repo

CC @tsachiherman

tsachiherman:
we won't be using this one anymore, right ?

2opremio:
No, but I left it open because there is still work to do as mentioned above. That's why I left it open.

tsachiherman:
> No, but I left it open because there is still work to do as mentioned above. That's why I left it open.

Yes, I missed these. Thanks!

2opremio:
Yes!

PR #4697, opened 2022-11-23
verify-range: revert on-disk mode, use S3 URLs for history archives

bartekn:
Revert on-disk mode in Captive-Core config and use S3 URLs for history archives. The first change ensures that verify-range container runs fast (currently it takes 2 hours in AWS to ingest state alone for a single range), the second makes the download of bucket files faster in AWS Batch.


sreuland:
may want to keep an eye on the verify-range ec2's, they should be pretty large compute profile already, just mentioning as will need to have the extra RAM capacity now that core will be using in-mem.

sreuland (approved):
thanks!

PR #4694, opened 2022-11-17
exp/services/soroban-rpc: getEvents method (Horizon-backed)

paulbellamy:
<details>
  <summary>PR Checklist</summary>
  
### PR Structure

* [ ] This PR has reasonably narrow scope (if not, break it down into smaller PRs).
* [ ] This PR avoids mixing refactoring changes with feature changes (split into two PRs
  otherwise).
* [ ] This PR's title starts with name of package that is most changed in the PR, ex.
  `services/friendbot`, or `all` or `doc` if the changes are broad or impact many
  packages.

### Thoroughness

* [ ] This PR adds tests for the most critical parts of the new functionality or fixes.
* [ ] I've updated any docs ([developer docs](https://developers.stellar.org/api/), `.md`
  files, etc... affected by this change). Take a look in the `docs` folder for a given service,
  like [this one](https://github.com/stellar/go/tree/master/services/horizon/internal/docs).

### Release planning

* [ ] I've updated the relevant CHANGELOG ([here](services/horizon/CHANGELOG.md) for Horizon) if
  needed with deprecations, added features, breaking changes, and DB schema changes.
* [ ] I've decided if this PR requires a new major/minor version according to
  [semver](https://semver.org/), or if it's mainly a patch change. The PR is targeted at the next
  release branch if it's not a patch change.
</details>

Fixes #4554 

### What

Implement the `getEvents` method in soroban-rpc API. For now, backed by fetching transactions from Horizon.

### Why

To get this in the hands of devs for field-testing without needing to wait for the real data backend to be complete in https://github.com/stellar/go/issues/4681

### Todo

- [x] Test it is usable, and not horrendously slow
  - [x] try it against futurenet for reference
- [x] Add tests
- [x] Resolve TODOs in the code
- [x] Merge and use https://github.com/stellar/stellar-xdr/pull/59
- [x] Handle horizon retries, rate-limiting etc. Use an exponential backoff

PR #4698, opened 2022-11-24
exp/services/soroban-rpc: Add LedgerEntry in-memory storage

2opremio:
Use a history archive checkpoint + captivecore txmeta to maintain an in-memory storage for the ledger entries needed for:

1. Obtaining preflight information (see #4692 ) for the `simulateTransaction` method
2. Implementing the `getContractData` method
3. Implemeting `getAccount` method.

Using this storage (together with #4692) we should be able to completely remove the existing Soroban RPC dependencies on 

* Core's `getledgerentry` endpoint
* Horizon (used for the `getAccount` method)

https://github.com/stellar/go/pull/4694 could also piggyback onto the txmeta stream of captive core to remove its Horizon dependency.

The PR has only been very mildly tested on futurenet.

TODO:
 * ~Test on pubnet to check feasibility:~
   * ~whether the memory consumed is overkill~
   * ~whether the initialization time is overkill~
* Fix tests
* Address code TODOs
* Make the code production ready (more robust/readable)

### DIFF
--- a/exp/services/soroban-rpc/internal/ledgerentry_storage.go
+++ b/exp/services/soroban-rpc/internal/ledgerentry_storage.go
@@ -1,0 +1,202 @@
+package internal
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"sync"
+	"time"
+
+	"github.com/stellar/go/historyarchive"
+	"github.com/stellar/go/ingest"
+	backends "github.com/stellar/go/ingest/ledgerbackend"
+	"github.com/stellar/go/xdr"
+)
+
+type LedgerEntryStorage interface {
+	GetLedgerEntry(key xdr.LedgerKey) (xdr.LedgerEntry, bool, uint32, error)
+	io.Closer
+}
+
+func NewLedgerEntryStorage(
+	networkPassPhrase string,
+	archive historyarchive.ArchiveInterface,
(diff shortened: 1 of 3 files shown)


2opremio:
I have tested it against pubnet on my mac and the memory consumption is pretty high (around 20GB for soroban-rpc):


Screenshot 2022-11-29 at 02 19 41

Note that this is without `ContractData` ledger entries, which don't yet exist in pubnet.

Processing the first checkpoint took ~3 minutes.

As expected, starting up (also in-memory) captive core took ~7 minutes to initialize (catchup to the checkpoint, sync and start emitting txmeta)

I am going to try to use _compressed_  ledger keys to see if the memory consumption goes down, but I don't think we will be able to get away with having everything in memory due to the memory consumption.

2opremio:
Using compressed ledger keys _only_ reduced the memory usage by 3GB.


Screenshot 2022-11-29 at 15 32 03

2opremio:
@tsachiherman quite the opposite, I am going to migrate this to the tools repo and use it as the basis of the Soroban RPC storage work (which needs to rely on captive core)

tsachiherman:
> @tsachiherman quite the opposite, I am going to migrate this to the tools repo and use it as the basis of the Soroban RPC storage work (which needs to rely on captive core)

yes, that was what I meant. We won't be needing this *here*, but may consider porting that.

PR #4703, opened 2022-11-29
services/horizon allow horizon docker image to be built using testing packages

jacekn:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This PR will allow us to use testing packages for producing stellar-horizon docker images.
stellar-core will still be installed from the stable repo.

### Why

This will let us test things better in some circumstances. Currently we are unable to build even temporary images using non-stable horizon releases.

### Known limitations

None


bartekn (approved):
LGTM but if we move adding a testing repo above `stellar-core` install command we could use core testing packages in the future as well.

jacekn:
> LGTM but if we move adding a testing repo above `stellar-core` install command we could use core testing packages in the future as well.

I wanted to avoid changing the original behavior. Before the change we always installed latest stable core. Adding the testing repo before installing core would give us latest stable OR testing core.
If it's desired we could change the build process to specify core version as parameter in the same way we do with horizon

PR #4708, opened 2022-12-07
Horizon 2.23.1 - CHANGELOG

bartekn:


tsachiherman (approved):

tsachiherman (commented):
I'm wondering : is that means that we won't be testing out horizon on Ubuntu 18 ?
I'm clearly don't mind that we make our releases on Ubuntu 20, but I don't want to break existing in installations that wishes to upgrade their horizon installation and find that it won't work.

bartekn:
> I'm wondering : is that means that we won't be testing out horizon on Ubuntu 18 ?

I think it's the other way around, before this PR we used `ubuntu-latest` which was `ubuntu-20.04` then. However, it seems that GH bumped `ubuntu-latest` to `ubuntu-22.04` quite recently. So there is no difference in our tests and for users.

PR #4724, opened 2022-12-20
#4721: updated soroban/p20 docker image and debian pkg to latest 19.6.1

### LINKED ISSUE #4721: release: update core docker image
update core docker image for go monorepo to compile against, tests in there, docker image.

sreuland:
updated GH CI config to use latest soroban enable core pkg release - `19.6.1-1158.c0ad35aa1.focal~soroban` 

Closes #4721


sreuland:
looking into integration test failure on `TestInvokeHostFunctionInvokeStatelessContractFn ` ...

tsachiherman (approved):
Looks great. We need to find a better solution for these wasms, though.

2opremio (approved):

sreuland:
> Looks great. We need to find a better solution for these wasms, though.

the [revised rpc design](https://github.com/stellar/go/issues/4681) may allow this version of horizon to be deprecated overall rather than try to optimize there, but, yes, good point.

Issue #4710, opened 2022-12-14
Enable BucketListDB by default in captive-core


SirTyson:
### What problem does your feature solve?

Recently, core added the `EXPERIMENTAL_BUCKETLIST_DB` flag. When this flag is set, `stellar-core` uses the BucketList for key value lookup for most queries instead of a SQL database. This reduces disk usage by over 50% since state does not need to be duplicated between a SQL database and the BucketList. BucketListDB also has up to 4x lookup performance compared to SQL. Finally, startup is significantly faster. Because we query buckets directly, we can skip most `BucketApply` logic since we no longer need to apply buckets to a SQL database. 

While this feature is still hidden behind an experimental flag, we think it's ready to be turned on by default when `captive-core` is using on disk mode. BucketListDB is not currently compatible in in-memory mode. 

### What would you like to see?

We would like to add `EXPERIMENTAL_BUCKETLIST_DB=true` to the default `captive-core` configuration when Horizon is run with the `--captive-core-use-db` flag. In addition to enabling BucketListDB, we will also need to tune some low-level optimization parameters for `captive-core`. There are two configuration parameters we will need to tune to the Horizon SKU, `EXPERIMENTAL_BUCKETLIST_DB_INDEX_PAGE_SIZE_EXPONENT` and `EXPERIMENTAL_BUCKETLIST_DB_INDEX_CUTOFF`. `EXPERIMENTAL_BUCKETLIST_DB_INDEX_PAGE_SIZE_EXPONENT` should be set to the machine storage device's physical page size. Currently, the default page size is 16KB, which is the optimal value for AWS NVMe storage. If Horizon nodes typically use a different storage class like EBS, this value will have to be changed. Unfortunately, during testing we found that page sizes reported by the OS are incorrect on AWS. The best way to figure out this value is just by trial and error with our microbenchmarking tool Tracy.
`EXPERIMENTAL_BUCKETLIST_DB_INDEX_CUTOFF` controls how much memory BucketListDB uses. A higher `EXPERIMENTAL_BUCKETLIST_DB_INDEX_CUTOFF` means more RAM usage but faster lookup performance. 

I'm currently working on getting a Horizon SKU test machine running where I can fine tune these values. In the mean time, I'll put out a PR so anyone else can also run a captive-core test, but it should not be merged until we find the optimal configuration settings.

PR #4733, opened 2023-01-05
services/horizon: Enable BucketListDB by default in captive core

SirTyson:
Addresses #4710.

This PR exposes BucketListDB related configuration parameters in the captive-core toml file and enables BucketListDB by default in captive-core when using `--captive-core-use-db`. When enabled, BucketListDB reduces captive-core disk usage by around 50% and significantly reduces startup and catchup time.

This update exposes the captive core flags `EXPERIMENTAL_BUCKETLIST_DB`, `EXPERIMENTAL_BUCKETLIST_DB_INDEX_PAGE_SIZE_EXPONENT`, and `EXPERIMENTAL_BUCKETLIST_DB_INDEX_CUTOFF`. These flags control the memory usage of BucketListDB. See the [stellar-core docs](https://github.com/stellar/stellar-core/blob/master/src/bucket/readme.md) for more information on configuring these flags.

By default, captive-core will enable BucketListDB with a page size of 4 KB. This is different than the validator's default page size of 16 KB. While SDF's validators have NVME storage, the Horizon SKU uses EBS, requiring a different page size. If a Horizon operator is using storage medium other than AWS EBS, the page size parameter should be tuned accordingly.

With the default BucketListDB configuration, captive-core will use an additional ~2 GB of memory. However, this is partially offset by the memory savings from reducing the size of the SQL database. Should this memory overhead be too high, memory usage can be configured using the exposed configuration parameters.

 This update will not automatically trigger a state rebuild. However, if `EXPERIMENTAL_BUCKETLIST_DB` is set to false in the captive-core toml, a state rebuild will be triggered.

### DIFF
--- a/ingest/ledgerbackend/toml.go
+++ b/ingest/ledgerbackend/toml.go
@@ -4,7 +4,9 @@ import (
 	"bytes"
 	"fmt"
 	"io/ioutil"
+	"os/exec"
 	"regexp"
+	"strconv"
 	"strings"
 
 	"github.com/stellar/go/support/errors"
@@ -83,6 +85,9 @@ type captiveCoreTomlValues struct {
 	Validators                           []Validator          `toml:"VALIDATORS,omitempty"`
 	HistoryEntries                       map[string]History   `toml:"-"`
 	QuorumSetEntries                     map[string]QuorumSet `toml:"-"`
+	UseBucketListDB                      bool                 `toml:"EXPERIMENTAL_BUCKETLIST_DB,omitempty"`
+	BucketListDBPageSizeExp              *uint                `toml:"EXPERIMENTAL_BUCKETLIST_DB_INDEX_PAGE_SIZE_EXPONENT,omitempty"`
+	BucketListDBCutoff                   *uint                `toml:"EXPERIMENTAL_BUCKETLIST_DB_INDEX_CUTOFF,omitempty"`
 }
 
 // QuorumSetIsConfigured returns true if there is a quorum set defined in the configuration.
@@ -317,6 +322,8 @@ type CaptiveCoreTomlParams struct {
 	Strict bool
 	// If true, specifies that captive core should be invoked with on-disk rather than in-memory option for ledger state
 	UseDB bool
+	// the path to the core binary, used to introspect core at runtie, determine some toml capabilities
+	CoreBinaryPath string
 }
 
 // NewCaptiveCoreTomlFromFile constructs a new CaptiveCoreToml instance by merging configuration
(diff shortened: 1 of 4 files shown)


services/horizon/CHANGELOG.md
```diff
@@ -6,6 +6,12 @@ file. This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+## 2.23.2
```
sreuland, line 9: can remove this line with version, it might be different depending on release time, i.e. 2.24.0 etc.

services/horizon/CHANGELOG.md
```diff
@@ -6,6 +6,12 @@ file. This project adheres to [Semantic Versioning](http://semver.org/).
...
+
+### Changes
+
+- Add support for BucketListDB params in captive core cfg/.toml file and enable BucketListDB by default when `--captive-core-use-db` set.
```
sreuland, line 13: thanks for adding the changelog entry, do you mind including link to the PR as well:

```suggestion
- Add support for BucketListDB params in captive core cfg/.toml file and enable BucketListDB by default when `--captive-core-use-db` set. ([4733](https://github.com/stellar/go/pull/4733))
```

ingest/ledgerbackend/toml.go
```diff
@@ -83,6 +83,9 @@ type captiveCoreTomlValues struct {
 	Validators                           []Validator          `toml:"VALIDATORS,omitempty"`
 	HistoryEntries                       map[string]History   `toml:"-"`
 	QuorumSetEntries                     map[string]QuorumSet `toml:"-"`
+	UseBucketListDB                      bool                 `toml:"EXPERIMENTAL_BUCKETLIST_DB,omitempty"`
+	BucketListDBPageSizeExp              *uint                `toml:"EXPERIMENTAL_BUCKETLIST_DB_INDEX_PAGE_SIZE_EXPONENT,omitempty"`
```
sreuland, line 87: what is reason on using pointer, other settings are mostly by-value, not disagreeing, just wanted to know the reason, thanks!

sreuland:
@SirTyson , thanks for incorporating theses changes, was checking out that one CI test failure in `ubuntu-20.04, 1.18.6, 9.6.5, captive-core-remote-storage, 19`, which seems to be failing on `TestClaimableBalanceCreationOperationsAndEffects`, maybe merge latest from master back in here, and let's see if that clears up?

services/horizon/CHANGELOG.md
mollykarcher: And this sounds like a warning that I think warrants being added to the CHANGELOG
```
This update will not automatically trigger a state rebuild. However, if EXPERIMENTAL_BUCKETLIST_DB is set to false in the captive-core toml, a state rebuild will be triggered.
```

mollykarcher:
Very excited to see the difference this ends up making, thanks for testing this out! Would love it if you could share the results (+process) of the benchmarking/calibration you're doing once that's complete.

ingest/ledgerbackend/toml.go
SirTyson: I want to have `omitempty` behavior, but 0 is a valid configuration that should not be omitted. If I use a regular `uint`, 0 is treated as an empty entry and not included. If I use a pointer, 0 is a valid entry, but empty entries are omitted, which is the proper behavior.

SirTyson:
> Very excited to see the difference this ends up making, thanks for testing this out! Would love it if you could share the results (+process) of the benchmarking/calibration you're doing once that's complete.

What would be the best format for something like this? It's a bit of an involved process where you have to build a special version of core and use some some low-level bench marking tools, but I could put a process doc together with some results if it's helpful.

ingest/ledgerbackend/toml.go
```diff
@@ -415,6 +418,15 @@ func (c *CaptiveCoreToml) setDefaults(params CaptiveCoreTomlParams) {
 		c.Database = "sqlite3://stellar.db"
 	}
 
+	if def := c.tree.Has("EXPERIMENTAL_BUCKETLIST_DB"); !def && params.UseDB {
```
sreuland, line 421: @SirTyson , I think we need to restrict when `UseBucketListDB` is enabled on the toml config to an external horizon command line flag also, so we get a new flag like `CaptiveCoreTomlParams.UseBucketListDB` available at this point  `https://github.com/stellar/go/pull/4733/files#diff-c7702ef819e52917ed3868588dff5273774c61a6aa8d36c1644d7ba251fb7756R305` since different versions of core will either support or not support this new toml config flag, looks like older core versions complain when it's present?

ingest/ledgerbackend/toml.go
```diff
@@ -409,12 +416,50 @@ func (c *CaptiveCoreToml) CatchupToml() (*CaptiveCoreToml, error) {
 	return offline, nil
 }
 
-func (c *CaptiveCoreToml) setDefaults(params CaptiveCoreTomlParams) {
+func (c *CaptiveCoreToml) checkCoreVersion(coreBinaryPath string) bool {
```
sreuland, line 419: I think this approach of dynamically checking core's runtime version to determine whether bucketlistdb is applied is good, but would like to get additional review from others on team if agree: @tamirms , @paulbellamy , @2opremio , @bartekn , any concerns on this?

sreuland (approved):
looks good, approved, but before merging, let's get a review from at least one other horizon team member to confirm this approach with core runtime version checking is acceptable approach:

https://github.com/stellar/go/pull/4733#discussion_r1069955700

thanks!

SirTyson:
Looks like all the comments have been addressed and the regex stuff has been reviewed, should be good to merge?

sreuland:
> Looks like all the comments have been addressed and the regex stuff has been reviewed, should be good to merge?

@SirTyson , yes, I resolved those older comments, hopefully the repo allows you to proceed on merge button, if not lmk can click it, thanks again, great work!

PR #4745, opened 2023-01-24
Enable BucketListDB by default in captive core

sreuland:
This is a port of a recent change that went in on mainline for enabling bucketlistdb for captive core on-disk mode:
https://github.com/stellar/go/pull/4733

it could potentially help performance, so wanted to see if we can port this into the soroban branch so that soroban-rpc can reference it?


sreuland:
> @sreuland usually we merge master ( eg [soroban-xdr-next...master](https://github.com/stellar/go/compare/soroban-xdr-next...master) ) into the soroban branch rather than cherry-picking individual commits.

yes, that's definitely ideal, I wasn't sure on the plan with this branch, i.e. if it will be merged to master for the protocol change. I'll close this pr and spin up a new one for merging latest master back to here, thanks!

PR #4746, opened 2023-01-24
txnbuild: add muxed account & memo support to SEP-10 utility functions

### LINKED ISSUE #3937: SEP-10: muxed account and memo support
## SEP-10 Muxed Account & Memo Support

SEP-10 now supports authenticating users of shared, pooled, or omnibus Stellar accounts.

These users are either represented using a muxed account (an `M...` address) or a Stellar account (`G...`) accompanied by an ID memo.

#### `buildChallengeTx()` or the equivalent function

- Allow muxed accounts as client account parameter values
- Add a optional `memo` parameter that accepts a 64-bit integer representation and attach it to the challenge transaction returned
  - If your language does not support optional parameters, this may be a breaking change
- Disallow muxed client accounts and memos to be passed in the same call; they are mutually exclusive parameters

#### `readChallengeTx()` or the equivalent function

- Allow muxed accounts as source accounts of the first operation within challenge transactions
- Add a 64-bit integer representation `memo` to the return value if present in the challenge transaction
  - Depending on the data type of the return value, this may be a breaking change
- Disallow muxed client accounts and memos to be present in the challenge transaction; they are mutually exclusive

### Implementations

The JS and Python SDKs have implemented this functionality, please use them as references:

- (stellar/js-stellar-sdk#709)[https://github.com/stellar/js-stellar-sdk/pull/709] included in release [v8.3.0](https://github.com/stellar/js-stellar-sdk/releases/tag/v8.3.0)
- (stellarCN/py-stellar-base#521)[https://github.com/StellarCN/py-stellar-base/pull/521] included in release [v4.2.0](https://github.com/StellarCN/py-stellar-base/releases/tag/4.2.0)

JakeUrban:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

resolves #3937

Adds support for muxed accounts to be specified as client account IDs and memos to be added to challenge transactions. This is a breaking change.

### Why

SEP-10 was modified to support users of shared Stellar accounts. The two approaches to identifying users of a shared account, in the context of SEP-10, is by assigning each user a muxed account address or an integer memo. 

### Outstanding Questions

The Python SDK moved away from returning mutliple values from the `BuildChallengeTx()` and `ReadChallengeTx()` equivalent functions and instead introduced a `ChallengeTransaction` container object for all the relevant values. We could use this same approach in the Go SDK instead of adding an additional value to the existing set.

### DIFF
--- a/exp/services/webauth/internal/serve/challenge.go
+++ b/exp/services/webauth/internal/serve/challenge.go
@@ -34,7 +34,9 @@ func (h challengeHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	queryValues := r.URL.Query()
 
 	account := queryValues.Get("account")
-	if !strkey.IsValidEd25519PublicKey(account) {
+	isStellarAccount := strkey.IsValidEd25519PublicKey(account)
+	isMuxedAccount := strkey.IsValidMuxedAccountEd25519PublicKey(account)
+	if !isStellarAccount && !isMuxedAccount {
 		badRequest.Render(w)
 		return
 	}
@@ -60,6 +62,10 @@ func (h challengeHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	var memo txnbuild.MemoID
 	if queryValues.Get("memo") != "" {
+		if isMuxedAccount {
+			badRequest.Render(w)
+			return
+		}
 		memoInt, err := strconv.ParseUint(queryValues.Get("memo"), 10, 64)
 		if err != nil {
 			badRequest.Render(w)
--- a/exp/services/webauth/internal/serve/challenge.go
+++ b/exp/services/webauth/internal/serve/challenge.go
@@ -60,7 +60,7 @@ func (h challengeHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		homeDomain = h.HomeDomains[0]
 	}
 
-	var memo txnbuild.MemoID
+	var memo txnbuild.Memo
 	if queryValues.Get("memo") != "" {
 		if isMuxedAccount {
 			badRequest.Render(w)
--- a/exp/services/webauth/internal/serve/challenge.go
+++ b/exp/services/webauth/internal/serve/challenge.go
@@ -63,10 +63,6 @@ func (h challengeHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	var memo *txnbuild.MemoID
 	memoParam := queryValues.Get("memo")
 	if memoParam != "" {
-		if isMuxedAccount {
-			badRequest.Render(w)
-			return
-		}
 		memoInt, err := strconv.ParseUint(memoParam, 10, 64)
 		if err != nil {
 			badRequest.Render(w)
@@ -87,7 +83,7 @@ func (h challengeHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	)
 	if err != nil {
 		h.Logger.Ctx(ctx).WithStack(err).Error(err)
-		serverError.Render(w)
+		badRequest.Render(w)
 		return
 	}
 
--- a/exp/services/webauth/internal/serve/challenge.go
+++ b/exp/services/webauth/internal/serve/challenge.go
@@ -2,6 +2,7 @@ package serve
 
 import (
 	"net/http"
+	"strconv"
 	"strings"
 	"time"
 
@@ -57,6 +58,16 @@ func (h challengeHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		homeDomain = h.HomeDomains[0]
 	}
 
+	var memo txnbuild.MemoID
+	if queryValues.Get("memo") != "" {
+		memoInt, err := strconv.ParseUint(queryValues.Get("memo"), 10, 64)
+		if err != nil {
+			badRequest.Render(w)
+			return
+		}
+		memo = txnbuild.MemoID(memoInt)
+	}
+
 	tx, err := txnbuild.BuildChallengeTx(
 		h.SigningKey.Seed(),
 		account,
@@ -64,6 +75,7 @@ func (h challengeHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		homeDomain,
 		h.NetworkPassphrase,
 		h.ChallengeExpiresIn,
+		memo,
 	)
 	if err != nil {
 		h.Logger.Ctx(ctx).WithStack(err).Error(err)
--- a/exp/services/webauth/internal/serve/challenge.go
+++ b/exp/services/webauth/internal/serve/challenge.go
@@ -60,7 +60,7 @@ func (h challengeHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		homeDomain = h.HomeDomains[0]
 	}
 
-	var memo txnbuild.MemoID
+	var memo *txnbuild.MemoID
 	memoParam := queryValues.Get("memo")
 	if memoParam != "" {
 		if isMuxedAccount {
@@ -72,7 +72,8 @@ func (h challengeHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 			badRequest.Render(w)
 			return
 		}
-		memo = txnbuild.MemoID(memoInt)
+		memoId := txnbuild.MemoID(memoInt)
+		memo = &memoId
 	}
 
 	tx, err := txnbuild.BuildChallengeTx(
@@ -82,7 +83,7 @@ func (h challengeHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		homeDomain,
 		h.NetworkPassphrase,
 		h.ChallengeExpiresIn,
-		&memo,
+		memo,
 	)
 	if err != nil {
 		h.Logger.Ctx(ctx).WithStack(err).Error(err)
--- a/exp/services/webauth/internal/serve/challenge.go
+++ b/exp/services/webauth/internal/serve/challenge.go
@@ -60,13 +60,14 @@ func (h challengeHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		homeDomain = h.HomeDomains[0]
 	}
 
-	var memo txnbuild.Memo
-	if queryValues.Get("memo") != "" {
+	var memo txnbuild.MemoID
+	memoParam := queryValues.Get("memo")
+	if memoParam != "" {
 		if isMuxedAccount {
 			badRequest.Render(w)
 			return
 		}
-		memoInt, err := strconv.ParseUint(queryValues.Get("memo"), 10, 64)
+		memoInt, err := strconv.ParseUint(memoParam, 10, 64)
 		if err != nil {
 			badRequest.Render(w)
 			return
@@ -81,7 +82,7 @@ func (h challengeHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		homeDomain,
 		h.NetworkPassphrase,
 		h.ChallengeExpiresIn,
-		memo,
+		&memo,
 	)
 	if err != nil {
 		h.Logger.Ctx(ctx).WithStack(err).Error(err)
--- a/exp/services/webauth/internal/serve/token.go
+++ b/exp/services/webauth/internal/serve/token.go
@@ -2,6 +2,7 @@ package serve
 
 import (
 	"net/http"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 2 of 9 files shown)


exp/services/webauth/internal/serve/challenge.go
```diff
@@ -57,13 +60,28 @@ func (h challengeHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		homeDomain = h.HomeDomains[0]
 	}
 
+	var memo txnbuild.Memo
+	if queryValues.Get("memo") != "" {
```
tsachiherman, line 64: two comments:
1. make the "memo" into a const and use that const.
2. call `queryValues.Get("memo")` only once and re-use the output instead of calling it twice.

exp/services/webauth/internal/serve/token.go
```diff
@@ -140,10 +149,21 @@ func (h tokenHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	var sub string
+	if muxedAccount == (xdr.MuxedAccount{}) {
+		sub = clientAccountID
+		if memo != nil {
```
tsachiherman, line 155: given that memo is of datatype `txnbuild.Memo`, it would never be a `nil`; you might want to compare it to `txnbuild.Memo{}`

exp/services/webauth/internal/serve/token.go
```diff
@@ -140,10 +149,21 @@ func (h tokenHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+			xdrMemo, _ := memo.ToXDR()
```
tsachiherman, line 156: don't skip error testing. if you can't make the conversion, issue a server-side error.

txnbuild/transaction_challenge_example_test.go
```diff
@@ -75,7 +75,7 @@ func ExampleVerifyChallengeTxThreshold() {
 
 	// Server verifies signed challenge transaction
 	{
-		_, txClientAccountID, _, err := txnbuild.ReadChallengeTx(challengeTx, serverAccount.Address(), network.TestNetworkPassphrase, "webauthdomain.stellar.org", []string{"test"})
+		_, txClientAccountID, _, _, err := txnbuild.ReadChallengeTx(challengeTx, serverAccount.Address(), network.TestNetworkPassphrase, "webauthdomain.stellar.org", []string{"test"})
```
tsachiherman, line 78: add validation to the returned memo field.

tsachiherman (commented):
Overall, I think that it looks good, but I do have few comments that would improve the code quality and readability.

exp/services/webauth/internal/serve/token.go
```diff
@@ -140,10 +149,21 @@ func (h tokenHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	var sub string
+	if muxedAccount == (xdr.MuxedAccount{}) {
```
leighmcculloch, line 153: Is the muxed account ever going to be completely empty? It will at least contain the account ID I think.

JakeUrban:
Thanks @tsachiherman,

Independently from your comments, what do you think about the approach described in the "Outstanding Questions" section of the description? Should we stick we adding another return value or use a `ChallengeTransaction` struct?

txnbuild/transaction.go
```diff
@@ -998,8 +998,9 @@ func NewFeeBumpTransaction(params FeeBumpTransactionParams) (*FeeBumpTransaction
 
 // BuildChallengeTx is a factory method that creates a valid SEP 10 challenge, for use in web authentication.
 // "timebound" is the time duration the transaction should be valid for, and must be greater than 1s (300s is recommended).
+// Muxed accounts or ID memos can be provided to identity a user of a shared Stellar account.
 // More details on SEP 10: https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0010.md
-func BuildChallengeTx(serverSignerSecret, clientAccountID, webAuthDomain, homeDomain, network string, timebound time.Duration) (*Transaction, error) {
+func BuildChallengeTx(serverSignerSecret, clientAccountID, webAuthDomain, homeDomain, network string, timebound time.Duration, memo Memo) (*Transaction, error) {
```
leighmcculloch, line 1003: I think the `memo` field could probably be of type `MemoID`. We can always change it to `Memo` in the future if we end up supporting other types of memos. It would help make the API more compile time safe rather than rely on runtime checks.

leighmcculloch (commented):
Thanks for expanding support for memos and muxeds in the SDK and also updating the webauth server too, this has been lacking for a while.

I left some comments inline.

leighmcculloch:
> The Python SDK moved away from returning mutliple values from the `BuildChallengeTx()` and `ReadChallengeTx()` equivalent functions and instead introduced a `ChallengeTransaction` container object for all the relevant values. We could use this same approach in the Go SDK instead of adding an additional value to the existing set.

I think a refactor like that should take place as an independent change either ahead of or after this change. I'd avoid making the change in the same pull request because it's really easy for the complexity of the diff to expand to the point where we miss subtle changes in the logic of these sensitive functions.

That aside, there are pros and cons.

I generally prefer a struct and I think it is more ergonomic. A side-effect of returning a struct is that each time we add a new field existing software will continue to work. This side-effect is usually a good thing because for most software consumers of values only need to be aware of a partial set of data contained in the value.

However, in the case of authentication it is usually important for the consumer of a value to be aware of the entire value. So for this code, while it is less ergonomic, I think the many return values actually play a beneficial role.

txnbuild/transaction.go
JakeUrban: I wanted to take this approach, but since `MemoID` is defined as `type MemoID uint64`, how would I specify _not_ to attach a memo to the challenge? I can't pass `nil` if the parameter type is `MemoID`.

txnbuild/transaction.go
```diff
@@ -1105,57 +1115,61 @@ func generateRandomNonce(n int) ([]byte, error) {
 // one of the following functions to completely verify the transaction:
 // - VerifyChallengeTxThreshold
 // - VerifyChallengeTxSigners
-func ReadChallengeTx(challengeTx, serverAccountID, network, webAuthDomain string, homeDomains []string) (tx *Transaction, clientAccountID string, matchedHomeDomain string, err error) {
+//
+// The returned clientAccountID may be a Stellar account or Muxed account address. If
+// the address is muxed, or if the memo returned is non-nil, the challenge transaction
+// is being used to authenticate a user of a shared Stellar account.
+func ReadChallengeTx(challengeTx, serverAccountID, network, webAuthDomain string, homeDomains []string) (tx *Transaction, clientAccountID string, matchedHomeDomain string, memo Memo, err error) {
```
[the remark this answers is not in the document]
JakeUrban: Same question about how to specify that no memo was used.

JakeUrban:
Ok, thanks Leigh. We'll keep the use of multiple return values here.

exp/services/webauth/internal/serve/token.go
JakeUrban: `txnbuild.Memo` is defined as an interface, so it can be nil right? 

```go
package main

import "fmt"

type A interface {}

func main() {
    test(nil)
}

func test(a A) {
    fmt.Println(a)
}
```

This outputs `<nil>` to stdout.

txnbuild/transaction.go
leighmcculloch: I'd make the type `*MemoID`, so that you can pass `nil` or the value.

exp/services/webauth/internal/serve/challenge.go
```diff
@@ -57,13 +60,30 @@ func (h challengeHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+			badRequest.Render(w)
+			return
+		}
+		memoInt, err := strconv.ParseUint(memoParam, 10, 64)
+		if err != nil {
+			badRequest.Render(w)
+			return
+		}
+		memoId := txnbuild.MemoID(memoInt)
+		memo = &memoId
```
JakeUrban, line 76: I feel like I'm missing something here. I tried to do the following originally,

```go
memo = &txnbuild.MemoID(memoInt)
```

But the compiler wouldn't let me.

txnbuild/transaction.go
```diff
@@ -1021,7 +1022,16 @@ func BuildChallengeTx(serverSignerSecret, clientAccountID, webAuthDomain, homeDo
...
-		return nil, errors.Wrapf(err, "%s is not a valid account id", clientAccountID)
+		if _, err = xdr.AddressToMuxedAccount(clientAccountID); err != nil {
+			return nil, errors.Wrapf(err, "%s is not a valid account id or muxed account", clientAccountID)
+		} else if memo != nil {
+			return nil, errors.New("memos are not valid for challenge transactions with a muxed client account")
+		}
+	} else if memo != nil {
+		var xdrMemo xdr.Memo
+		if xdrMemo, err = memo.ToXDR(); err != nil || xdrMemo.Type != xdr.MemoTypeMemoId {
+			return nil, errors.New("memo must be of type MemoID")
```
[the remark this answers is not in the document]
JakeUrban: Using the `*MemoID` type makes the memo.toXDR()` calls unnecessary so I think this is resolved.

exp/services/webauth/internal/serve/token.go
```diff
@@ -140,10 +149,21 @@ func (h tokenHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		sub = clientAccountID
+		if memo != nil {
+			xdrMemo, _ := memo.ToXDR()
+			sub += ":" + strconv.FormatUint(uint64(xdrMemo.MustId()), 10)
```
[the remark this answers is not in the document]
JakeUrban: I think you're right but the request response wouldn't be the same as the schema defined for `serverError` so I used `getId()` instead.

exp/services/webauth/internal/serve/token.go
JakeUrban: It will be completely empty. `AddressToMuxedAccount()` initializes an empty `MuxedAccount{}` then calls `setAddress()` on it, which only assigns the address if it isn't muxed.

exp/services/webauth/internal/serve/challenge.go
leighmcculloch: Unfortunately you can't take the address of a literal unless it is a struct. This is super annoying feature of Go. Hopefully this will change in the future. There's a proposal to change it here: https://github.com/golang/go/issues/45624.

exp/services/webauth/internal/serve/token.go
leighmcculloch: According to the code below the value returned won't be empty for a `G...` address:
https://github.com/stellar/go/blob/e4288dbc66f5a6f225dbb03d1b5e74ed37bc85c7/xdr/muxed_account.go#L65-L75

leighmcculloch (commented):
One comment outstanding above:
- https://github.com/stellar/go/pull/4746/files#r1087283430

I think that there should be some tests added to the webauth service. There's a new parameter that's accepted, and new logic being added there.

For example:

Screenshot 2023-01-25 at 3 33 04 PM

Screenshot 2023-01-25 at 3 34 09 PM

exp/services/webauth/internal/serve/token.go
JakeUrban: Ah you're right. Let me fix that.

leighmcculloch:
I noticed another couple spots without coverage that seem important:

Screenshot 2023-01-26 at 11 13 47 AM


Screenshot 2023-01-26 at 11 14 06 AM

I think this is an area that was previously untested, but given we're adding some more complexity to whether an account is valid or not, I think it's probably worth adding tests for both cases.

JakeUrban:
The 2nd check in first screenshot, where it errors if a memo and muxed account are provided, is tested. But I'll add tests for the 2 other checks.

leighmcculloch (approved):
👏🏻

exp/services/webauth/internal/serve/token.go
```diff
@@ -89,7 +89,11 @@ func (h tokenHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	l.Info("Start verifying challenge transaction.")
 
 	muxedAccount, err := xdr.AddressToMuxedAccount(clientAccountID)
-	if err == nil {
+	if err != nil {
+		serverError.Render(w)
```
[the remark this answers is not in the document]
JakeUrban: I'm conflicted on this. At first I agreed, then I figured that transactions wouldn't the serializable unless their operations source accounts were valid. But looking at the Go SDK's `NewTransaction` function and `ManageData.Validate()`, it doesn't seem like operation source accounts are confirmed to the valid keys. So I guess we should use bad request.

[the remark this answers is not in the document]
leighmcculloch: 👍🏻

PR #4748, opened 2023-01-25
Port latest master to soroban-xdr-next, including bucketlistdb changes

sreuland:
wanted to port latest master to the soroban branch as it includes new change for bucketlistdb toml config for captive core on-disk mode, so soroban-rpc server could use captive core cfg's that reference these new toml params.


services/horizon/internal/integration/parameters_test.go
```diff
@@ -2,15 +2,16 @@
...
-	"github.com/stellar/go/services/horizon/internal/simplepath"
 	"io/ioutil"
 	"os"
 	"os/exec"
 	"path"
 	"strings"
 	"testing"
 
+	"github.com/stellar/go/services/horizon/internal/paths"
+	"github.com/stellar/go/services/horizon/internal/simplepath"
```
tsachiherman, line 13: can you check out the above comment?

```
"github.com/stellar/go/services/horizon/internal/paths" imported but not used (typecheck)
```

sreuland: strange, the 'paths' package is referenced in this file version - https://github.com/stellar/go/pull/4748/files/a586f53e3cb8f129833eff431519e0b619efdc21#diff-6f737140ca2744b223417ca5d5f6f4a20e9812494960796e774932e97b85da5fR187

not sure why golang-lint is logging this out, but it's not causing the gha lint action to fail, that passed regardless. I loaded this branch in VSCode, and not seeing it highlight it. This is port of master, no new dev, so, this would likely be triggering same lint on that branch.

sreuland: Spoke with @tsachiherman , he identified the root cause as a gh go compiler version issue, I spun up new ticket to track that fix - https://github.com/stellar/go/issues/4749 , will proceed with merging on this pr as it's not related.

PR #4752, opened 2023-01-26
services/submitter: Prototype of Transaction Submission Service

bartekn:
### What

### Why

### Known limitations

[TODO or N/A]

### DIFF
--- a/services/submitter/channel.go
+++ b/services/submitter/channel.go
@@ -1,0 +1,51 @@
+package main
+
+import (
+	"sync"
+
+	"github.com/stellar/go/clients/horizonclient"
+	"github.com/stellar/go/keypair"
+)
+
+// Channel contains current state of the channel account and provides methods to reload and read sequence number
+type Channel struct {
+	Seed string
+
+	mutex          sync.Mutex
+	accountID      string
+	sequenceNumber int64
+}
+
+// ReloadState loads the current state of the channel account using given horizon client
+func (ch *Channel) ReloadState(client horizonclient.ClientInterface) (accountID string, sequenceNumber uint64, err error) {
+	ch.mutex.Lock()
+	defer ch.mutex.Unlock()
+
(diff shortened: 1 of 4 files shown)


JakeUrban:
Thanks for getting this started, I'm super excited about the project.

### Error Handling

I think the next step to take here is to implement automatic resubmission on recoverable errors. Recoverable errors are:

- 504 status code
- 429 status code
- tx_too_late transaction result code
- tx_bad_seq transaction result code
- tx_insufficient_fee transaction result code

Errors that are not recoverable:

- 500 status code
- tx_insufficient_balance result code
- all payment operation result codes (except success)

All other errors indicate a bug in the package.

### Interface Design

I also think we need to simplify the interface exposed to the developer using this package. From their perspective, there should be three transaction statuses:

- pending
- success
- error

There should be no distinction between a transaction that has not been picked up from the store, one that has been retrieved from the store but not submitted, and one that has been submitted but has not been included in a ledger.

JakeUrban:
### What

Refactors @bartekn's initial implementation of the transaction submitter to better manage concurrent transaction submission.

- Each channel account is given a go routine that pull from a buffered queue of transactions
- Every second, the submitter checks if the queue is full, and if its not, fills the queue with more pending transactions
- For every transaction placed in the queue, a go routine picks up the transaction and uses it's channel account to submit it

This PR also adds the following:

- Signs the transactions with the root & channel account
- Uses fee bump transactions so channel accounts don't need to hold XLM

### Why

The existing implementation assigns transactions to channels that could still be working on submitting previously assigned transactions. This would cause the sequence number errors and channel accounts would constantly have to reload their state, reconstruct transactions, and resubmit them.

Using a buffered channel allows for:

1. the submitter to only query the DB when the channel is not full
2. The channel accounts to never be idle -- unless there are no pending transactions in the database

PR #4753, opened 2023-01-27
allow optional stellar_core_version to be passed to horizon docker build

mwtzzz:
### What
allow optional stellar_core_version to be passed to horizon docker build

### Why
need it for the SRE hackathon project and could be useful for other things

### Testing
will test with branch

### Issue addressed by this PR
https://github.com/stellar/ops/issues/2076

### DIFF
--- a/services/horizon/docker/Dockerfile
+++ b/services/horizon/docker/Dockerfile
@@ -6,7 +6,7 @@ ARG DEBIAN_FRONTEND=noninteractive
 RUN apt-get update && apt-get install -y wget apt-transport-https gnupg2 && \
     wget -qO /etc/apt/trusted.gpg.d/SDF.asc https://apt.stellar.org/SDF.asc && \
     echo "deb https://apt.stellar.org focal stable" | tee -a /etc/apt/sources.list.d/SDF.list && \
-    apt-get update && apt-cache madison stellar-core && apt-get install -y stellar-core && \
+    apt-get update && apt-cache madison stellar-core && apt-get install -y stellar-core=${STELLAR_CORE_VERSION} && \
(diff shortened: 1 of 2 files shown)


services/horizon/docker/Dockerfile
```diff
@@ -1,12 +1,13 @@
...
 
 ARG VERSION
+ARG STELLAR_CORE_VERSION
 ARG DEBIAN_FRONTEND=noninteractive
 
 RUN apt-get update && apt-get install -y wget apt-transport-https gnupg2 && \
     wget -qO /etc/apt/trusted.gpg.d/SDF.asc https://apt.stellar.org/SDF.asc && \
     echo "deb https://apt.stellar.org focal stable" | tee -a /etc/apt/sources.list.d/SDF.list && \
-    apt-get update && apt-cache madison stellar-core && apt-get install -y stellar-core && \
+    apt-get update && apt-cache madison stellar-core && eval "apt-get install -y stellar-core${STELLAR_CORE_VERSION+=$STELLAR_CORE_VERSION}" && \
```
jacekn, line 10: At this point we only have stable repo. We should support testing or even unstable packages too so that we can build soroban images

jacekn (approved):

services/horizon/docker/Dockerfile
mwtzzz: so this needs to get passed in as a parameter from Jenkins.

mwtzzz: I'm thinking a checkbox in Jenkins : "yes" include testing and unstable repos, and pass that into Dockerfile

jacekn: You could enable unstable/testing when core version is explicitly provided, otherwise default to the existing behavior which is to use latest stable package

PR #4755, opened 2023-01-31
Bump core to latest stable 19.8.0

marta-lokhova:


.github/workflows/horizon.yml
```diff
@@ -110,7 +110,7 @@ jobs:
     name: Test (and push) verify-range image
     runs-on: ubuntu-20.04
     env:
-      STELLAR_CORE_VERSION: 19.6.0-1138.b3a6bc281.focal
+      STELLAR_CORE_VERSION: 19.7.0-1175.7249363c6.focal
```
sreuland, line 113: I noticed references to other core versions above also, https://github.com/stellar/go/pull/4755/files#diff-734d2e7c3b283e8b975c1e9e58292168bc7040248b3931d7610c8eee19591c1cR37, which are using 19.5, but I would think all core references here should be consistent, @Shaptic @2opremio @tamirms , do you know why these have diverged a bit, is it ok to just get them all up to this current 19.7 at this point?

sreuland:
looked like one integration test failed, likely transient, I kicked off re-run, if that passes s/b all good!

sreuland (approved):

Shaptic (approved):
confirmed that the version matches debian repos :+1:

mbsdf (approved):

Issue #4785, opened 2023-03-03
HTTP Setup: Default WriteTimeout


C0x41lch0x41:
### What problem does your feature solve?
There is no default `WriteTimeout` in the setup of http server. If the server instance is used using a configuration without `WriteTimeout` it will be set to 0 which means `no timeout`. That may be an issue if a request takes too much time to process and can potentially lead to DoS the server.
We need to be careful how this value is set though. It should be more than ReadTimeout.

Those two articles provide useful information in my opinion:
* [So you want to expose Go on the Internet](https://blog.cloudflare.com/exposing-go-on-the-internet/)
* [The complete guide to Go net/http timeouts](https://blog.cloudflare.com/the-complete-guide-to-golang-net-http-timeouts/)

### What would you like to see?
A default value for `WriteTimeout` in [HTTP setup function](https://github.com/stellar/go/blob/fb9e9435abed40725cfeae9603bfb9addefff074/support/http/main.go#L91)

### What alternatives are there?
We already have a default `ReadTimeout` and [Go >= 1.13](https://go.dev/doc/go1.13#net) sets a [pretty aggressive TCP timeout](https://cs.opensource.google/go/go/+/master:src/net/dial.go;l=17), so this may be sufficient if our services does not let the client control parameters that influence the processing time of a request. But it is good practice in case there is a edge case.

stellarsaur:
We discussed this, and came to the conclusion we need to take a more careful look at our existing timeouts to make sure we use the right default WriteTimeout. Ideally, this should be longer than all our existing timeouts, but not unlimited like it currently is.

Moved back to our backlog for now.

leighmcculloch:
Fyi this was something we investigated previously. There are some findings [here](https://github.com/stellar/go/security/advisories/GHSA-5v3w-33m8-hhq5#advisory-comment-1382).

PR #4796, opened 2023-03-07
support/http - Added default WriteTimeout of 30s to HTTP server

stellarsaur:
See [this issue](https://github.com/stellar/go/issues/4785) for full context.

30 seconds seems like a reasonable default to account for large payloads and network congestion.

### DIFF
--- a/support/http/main_test.go
+++ b/support/http/main_test.go
@@ -25,7 +25,7 @@ func TestRun_setupDefault(t *testing.T) {
 
 	assert.Equal(t, defaultShutdownGracePeriod, srv.Timeout)
 	assert.Equal(t, defaultReadTimeout, srv.ReadTimeout)
-	assert.Equal(t, time.Duration(0), srv.WriteTimeout)
+	assert.Equal(t, defaultWriteTimeout, srv.WriteTimeout)
 	assert.Equal(t, time.Duration(0), srv.IdleTimeout)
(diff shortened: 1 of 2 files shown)


Shaptic (commented):
Thanks for the PR! I think it'd be worthwhile to discuss this in the team meeting tomorrow because we have a lot of existing timeouts and this one needs to be bigger than all of them. For example, we kill DB requests after 10s and web requests after 30s, I think? So it's possible that this timeout needs to be higher, e.g. 45s. And others might have more thoughts. Can you add an [agenda item](https://docs.google.com/document/d/1Y5_nl9yXYfXkRR0vA7BPgnWJm2l99-KhZ_NLZVZTXtM/)?

support/http/main_test.go
```diff
@@ -25,7 +25,7 @@ func TestRun_setupDefault(t *testing.T) {
 
 	assert.Equal(t, defaultShutdownGracePeriod, srv.Timeout)
 	assert.Equal(t, defaultReadTimeout, srv.ReadTimeout)
-	assert.Equal(t, time.Duration(0), srv.WriteTimeout)
+	assert.Equal(t, defaultWriteTimeout, srv.WriteTimeout)
```
tsachiherman, line 28: This isn't wrong, but it also doesn't verify that for a non-zero config value, the same value is being set.

Issue #4725, opened 2022-12-22
horizon: ingest Account Credited / Debited effects for smart contract asset transfers


tamirms:
The upcoming soroban release will implement single asset balance assets. This issue is a subtask of https://github.com/stellar/go/issues/4722 .

We will need to emit "Account Credited" / "Account Debited" effects whenever a Stellar asset is transferred or clawed back from a Stellar account via a smart contract invocation. This will require modifications to the [effects ingestion processor](https://github.com/stellar/go/blob/soroban-xdr-next/services/horizon/internal/ingest/processors/effects_processor.go#L226). The changes should be pretty straight forward because we can iterate through the tx meta produced by the invoke host function operation and see if there are any ledger entry changes for trustlines.

prerequisite: #4747

Shaptic:
Some remarks:

 * Is it possible to infer that a contract data ledger entry belongs to a SAC from the entry itself? e.g. how it's possible to detect whether a particular Ethereum contract represents (or at least mimics) an ERC-20 token? Or do you run into an issue of "it looks like a SAC and acts like a SAC but it isn't one"?
 * If we decide to store entries, could/should we do it in a way that's distinct from the rest of our database, e.g. storing it in sqlite? This would let us heavily imply that we are storing this data simply for the purposes of persistence, rather than trying to make it a "full-fledged" part of Horizon (e.g. part of the overall DB schema, plumbed to endpoints, complex queries, scalability, etc.).
 * Should Horizon talk to Soroban RPC for this data? :sweat_smile:

Shaptic:
Thoughtful answers, thanks! I agree with you on all counts, just needed some rationale. 

Horizon/RPC ingestion sync can be mitigated by sharing a Captive Core, but I think you're right that the dependency might get hairy. My intent was to try to keep the SC and Classic worlds as decoupled as possible, but that doesn't seem possible :confused: 

On the first point, is there a way to generate meta that **must** be unique to the SAC? For example, one could have a method that returns a signature signed by the issuing account. Such a blob couldn't be returned by a non-SAC SC, since it'd need the issuer's secret. Though I guess in that case, the issuer itself could mimic the SAC and that method, but can an issuer deploy its own token twice (once via SAC and once via custom)?? 

Anyway, that's just a thought - maybe there's a clever construction there that Core can pull off and we can leverage to glean information.

tomerweller:
I think we should embrace the events approach. and I agree there's no escape from saving some minimal contract data information. 

[...]

PR #4798, opened 2023-03-08
ingest: Adds basic event parsing for standard Stellar Asset Contract events

Shaptic:
### What
Adds an abstraction layer for standardized contract events that come from the Stellar Asset Contract.

### Why
Event parsing is necessary to generate effects and record payments as part of #4722 (and #4725).

### Known limitations
This only covers the `transfer` events generated by calling `xfer` and `xfer_from`. It could also use more test cases.

Ideally, @sreuland and I would build the parsing/testing on top of this PR for the events we need as they pertain to our issues.

### DIFF
--- a/ingest/event.go
+++ b/ingest/event.go
@@ -139,9 +139,9 @@ func NewStellarAssetContractEvent(event *Event, networkPassphrase string) (*Stel
 }
 
 // parseTransferEvent tries to parse the given topics and value as a SAC
-// "transfer" event. It assumes that the `topics` EXCLUDES the function name,
-// i.e. that its been validated previously. It will return a best-effort parsing
-// even in error cases.
+// "transfer" event. It assumes that the `topics` array has already validated
+// both the function name AND the asset <--> contract ID relationship. It will
+// return a best-effort parsing even in error cases.
 func (event *StellarAssetContractEvent) parseTransferEvent(topics xdr.ScVec, value xdr.ScVal) error {
 	//
 	// The transfer event format is:
@@ -157,7 +157,6 @@ func (event *StellarAssetContractEvent) parseTransferEvent(topics xdr.ScVec, val
 		return ErrNotTransferEvent
 	}
 
-	fmt.Println("xfer here")
 	from, to := topics[1], topics[2]
 	if from.Type != xdr.ScValTypeScvObject || to.Type != xdr.ScValTypeScvObject {
 		return ErrNotTransferEvent
--- a/ingest/event.go
+++ b/ingest/event.go
@@ -104,16 +104,20 @@ func NewStellarAssetContractEvent(event *Event, networkPassphrase string) (*Stel
 
 	asset, err := txnbuild.ParseAssetString(string(assetBytes))
 	if err != nil {
-		return evt, ErrNotStellarAssetContract
+		return evt, errors.Wrap(ErrNotStellarAssetContract, err.Error())
 	}
 
-	xdrAsset, err := xdr.NewCreditAsset(asset.GetCode(), asset.GetIssuer())
+	switch asset.IsNative() {
+	case true:
+		evt.Asset = xdr.MustNewNativeAsset()
+	case false:
+		evt.Asset, err = xdr.NewCreditAsset(asset.GetCode(), asset.GetIssuer())
+	}
 	if err != nil {
-		return evt, ErrNotStellarAssetContract
+		return evt, errors.Wrap(ErrNotStellarAssetContract, err.Error())
 	}
-	evt.Asset = xdrAsset
 
-	expectedId, err := xdrAsset.ContractID(networkPassphrase)
+	expectedId, err := evt.Asset.ContractID(networkPassphrase)
 	if err != nil {
 		return evt, errors.Wrap(ErrNotStellarAssetContract, err.Error())
 	}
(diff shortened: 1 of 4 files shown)


ingest/event.go
```diff
@@ -0,0 +1,185 @@
...
+	// For all parsing errors, we just continue, since it's not a real
+	// error, just an event non-complaint with SAC events.
+	rawAsset := topics[len(topics)-1]
+	assetContainer, ok := rawAsset.GetObj()
+	if !ok || assetContainer == nil {
+		return evt, ErrNotStellarAssetContract
+	}
+
+	assetBytes, ok := assetContainer.GetBin()
+	fmt.Println("here", len(assetBytes), assetBytes)
```
sreuland, line 100: good debug logging, just a little more descriptive, `fmt.Printf('SAC asset event received for %v', string(assetBytes))` ?

Shaptic: lol I left this in while I was debugging tests :facepalm: maybe we shouldn't print stuff in this package? Horizon itself should probably do that, though

ingest/event.go
```diff
@@ -0,0 +1,185 @@
...
+		xdr.ScSymbol("burn"):     EventTypeBurn,
+	}
+
+	// TODO: Better parsing errors
+	ErrNotStellarAssetContract = errors.New("event was not from a Stellar Asset Contract")
+	ErrEventUnsupported        = errors.New("this type of Stellar Asset Contract event is unsupported")
+	ErrNotTransferEvent        = errors.New("event is an invalid 'transfer' event")
+)
+
+type StellarAssetContractEvent struct {
```
sreuland, line 45: This event is meant to model a sub-group of similar functional events from SAC related to balance changes mostly correct? Maybe can indicate that with name like `StellarAssetContractBalanceChangedEvent` to infer that meaning a bit more? 

Do you think this eventually scales to point of being 1-1, a separate event struct specific to each SAC method, i.e. `StellarAssetContract<sac_method>Event`, each supporting common interface:

type StellarAssetContractEvent interface {
   getType() int
}

`NewStellarAssetContractEvent()` becomes more polymorphic, returning instance of event specific struct that matched contract event, caller uses `getType()` to cast to real struct type.

ingest/event.go
```diff
@@ -0,0 +1,218 @@
...
+	// SAC event. At this point, we can parse the event and treat it as
+	// truth, mapping it to effects where appropriate.
+	if expectedId != *event.ContractId { // nil check was earlier
+		return evt, ErrNotStellarAssetContract
+	}
+
+	switch evt.Type {
+	case EventTypeTransfer:
+		return evt, evt.parseTransferEvent(topics, value)
+	case EventTypeMint:
```
sreuland, line 131: these seem to return an un-usable event, as only `Type` has been populated, maybe just throw unsupported also to avoid callers trying to do something, or do you think we can squeeze mint/clawback into the current `StellarAssetContractEvent`?

 I was wondering if mint is qualified for including in payments, I see payments includes `CreateAccount` operation, which is somewhat similar to mint?

ingest/event.go
Shaptic: > This event is meant to model a sub-group of similar functional events from SAC related to balance changes mostly correct? 

I thought, rather, that this would be done for all SAC events, but we'd start with balance-related ones since those have the immediate need. I do like the idea of making it polymorphic (kinda like `txnbuild.Operation` is, see [here](https://github.com/stellar/go/blob/ece49d110dd2c68d3ace79bf1fc4bb061a280e9d/txnbuild/operation.go#L9-L15)), but maybe it's too soon? For now, it's easier to use the `struct` and just know which fields are set for which types.

sreuland (approved):
great work to capture contract events in a re-usable format, I'm sure will iterate more on it, but it provides insight on event model and will be re-used!

ingest/event.go
Shaptic: As for payments, I think it should count, because in the Classic world, a "mint" of your custom asset only **_really_** happens after you make a `PaymentOp` to someone. Same thing with burns.

ingest/event.go
sreuland: yep, we don't want to make it too fancy yet, as it may change, but thinking of how callers will use the event struct, they're gonna end up hardcoding what fields in the uber-struct(`StellarAssetContractEvent `) are valid for each type of event(xfer, mint, clawback, etc), whereas if we bite the bullet now and capture the fields-per-contract-event in separate structs, then callers can handle different events in more deterministic type-safe way

Shaptic: yeah I 100% agree, was just being lazy tbh - added in latest push (a7ebfd6184380f22eb310a58f664f4815cb49fee)! is that what you had in mind?

sreuland: @Shaptic , nice, I see you put up https://github.com/stellar/go/pull/4802 which captures the different event models succinctly, thanks!

ingest/event.go
```diff
@@ -0,0 +1,244 @@
...
+		return ""
+	}
+
+	var result string
+	var err error
+
+	switch address.Type {
+	case xdr.ScAddressTypeScAddressTypeAccount:
+		pubkey := address.MustAccountId().Ed25519
+		fmt.Println("pubkey:", address.MustAccountId())
```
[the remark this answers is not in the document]
Shaptic: good catches, thanks! should be gone in the other PR

Issue #4691, opened 2022-11-11
support/render/problem: include detail when printing problem as an error


howardtw:
### What problem does your feature solve?

Including error detail provides more information for the developer when it comes to debugging. Right now we only see the following log line for all 400 errors:
```
"request failed due to error" error="problem: bad_request"
```
which is not very helpful.

### What would you like to see?

Include `Problem.Detail` in the [Error()](https://github.com/stellar/go/blob/7a67fed427d7ebbe8e8a58c3d32d517412edcf06/support/render/problem/problem.go#L63-L65) function.

### What alternatives are there?

Add logging for the Problem.Detail [here](https://github.com/stellar/go/blob/7a67fed427d7ebbe8e8a58c3d32d517412edcf06/support/render/problem/problem.go#L172-L175)

leighmcculloch:
I've had similar experiences when encountering this error. I almost always want to know more, and it's inconvenient needing to write some additional code to go and get that information.

PR #4801, opened 2023-03-09
Include full problem details when printing problem as an error

stellarsaur:
See [this issue](https://github.com/stellar/go/issues/4691) for full context.

This PR aims to make error messages more useful by including more detail, instead of just the type of error.

### DIFF
--- a/support/render/problem/problem.go
+++ b/support/render/problem/problem.go
@@ -61,7 +61,7 @@ type P struct {
 }
 
 func (p P) Error() string {
-	return fmt.Sprintf("problem: %s", p.Type)
+	return fmt.Sprintf("problem: %s. full details: %s", p.Type, p.Detail)
 }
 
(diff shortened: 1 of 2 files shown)


support/render/problem/problem.go
```diff
@@ -61,7 +61,7 @@ type P struct {
 }
 
 func (p P) Error() string {
-	return fmt.Sprintf("problem: %s", p.Type)
+	return fmt.Sprintf("problem: %s. full details: %s", p.Type, p.Detail)
```
Shaptic, line 64: any chance we relied on the previous format programmatically, somewhere?

Shaptic (approved):
yeet it, but you should cc @howardtw to see if this resolves what they need

support/render/problem/problem.go
howardtw: That's what I was thinking as well. I wondered if the Horizon team would know better 😃

PR #4802, opened 2023-03-09
ingest: Move SAC event parsing to its own package, add more events.

Shaptic:
### What
Building on #4798, this moves standardized SAC event parsing into its own package, `support/contractevents`.

This also adds parsing for the other "balance change-like" events:
 - `mint`
 - `burn`
 - `clawback`

### Why
It just makes more sense as event parsing is unrelated to ingestion.

### Known limitations
We still need parsing for the rest of the events.

The event structures are modeled after the [token contract](https://github.com/stellar/rs-soroban-env/blob/main/soroban-env-host/src/native_contract/token/event.rs). However, it'd be best for testing if we had either (a) events come from Core itself or (b) raw XDR taken from Core generating these events.


support/contractevents/event_test.go
```diff
@@ -0,0 +1,296 @@
...
+	require.Equal(t, EventTypeBurn, sacEvent.GetType())
+
+	burnEvent := sacEvent.(*BurnEvent)
+	require.Equal(t, randomAccount, burnEvent.From)
+	require.EqualValues(t, 10000, burnEvent.Amount.Lo)
+	require.EqualValues(t, 0, burnEvent.Amount.Hi)
+}
+
+func TestFuzzingSACEventParser(t *testing.T) {
+	gen := randxdr.NewGenerator()
```
sreuland, line 167: wow, I've not paid attention to fuzzing much in tests, is this typical to run many randomized blobs to verify against non-compliant xdr, is just one non-compliant xdr equivalent to same outcome?

support/contractevents/utils.go
```diff
@@ -0,0 +1,90 @@
+package contractevents
```
sreuland, line 1: would `main.go` be considered more idiomatic file name rather than `utils.go` when it comes to these common package declarations?

support/contractevents/utils.go
```diff
@@ -0,0 +1,90 @@
...
+	"github.com/stellar/go/strkey"
+	"github.com/stellar/go/support/errors"
+	"github.com/stellar/go/xdr"
+)
+
+var ErrNotBalanceChangeEvent = errors.New("event doesn't represent a balance change")
+
+// MustScAddressToString converts the low-level `xdr.ScAddress` union into the
+// appropriate strkey (contract C... or account ID G...).
+func MustScAddressToString(address *xdr.ScAddress) string {
```
sreuland, line 15: for design consideration, what if these methods here in utils were instead defined on `sacEvent` struct, usable from concrete event structs like mint,clawback,transfer,etc through their existing embedded usage of `sacEvent`.

sreuland (approved):
great refactor for extended event types.

support/contractevents/event_test.go
Shaptic: my goal here was to make sure that the parser itself doesn't blow up under stress, regardless of the event form. fuzzing like this helps test more paths than a single non-complaint example, I think. for example I hit a nil pointer deref under a subtle case w/ a native asset. a SAC is pretty unlikely to automatically generate using this method, but testing the various "unhappy paths" is still useful, I think, since there are many exit points in the parser

support/contractevents/utils.go
Shaptic: I actually think this helper specifically would be better in the `xdr` package as I imagine this isn't the last time that someone will want the `strkey` representation of an `ScAddress`. sticking it on as a private method of `sacEvent` lowers its utility further. for the others, that might make sense, but then it hurts re-use in the non-SAC cases if people want to use the lib (maybe they should be public instead of private, then)

support/contractevents/utils.go
Shaptic: I feel like `main.go` is misleading to me because then I expect it to be a "runnable" package when this is only a library.

support/contractevents/utils.go
sreuland: yes, sounds reasonable, early still, so that refactor can happen as/if needed on later passes, thanks!

support/contractevents/event_test.go
sreuland: ok, thanks for further clarification.

PR #4806, opened 2023-03-10
services/horizon: Add `account_credited`/`debited` effects for SAC events.

Shaptic:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Adds ability to generate effects for `invokeHostFunction` operations based on Stellar Asset Contract events.

### Why
#4725 as part of #4722.

### Known limitations
Effects are not generated if the involved party is a contract address (in the form `C...`). This is by design by nature of the effect name (i.e. _account_`_debited`), though we may introduce new, contract-related effects in the future.

### DIFF
--- a/amount/main.go
+++ b/amount/main.go
@@ -117,6 +117,11 @@ func String(v xdr.Int64) string {
 	return StringFromInt64(int64(v))
 }
 
+// TODO: Make this actually work.
+func String128(v xdr.Int128Parts) string {
+	return StringFromInt64(int64(v.Lo))
+}
+
 // StringFromInt64 returns an "amount string" from the provided raw int64 value `v`.
 func StringFromInt64(v int64) string {
 	r := big.NewRat(v, 1)
--- a/amount/main.go
+++ b/amount/main.go
@@ -10,6 +10,8 @@
 package amount
 
 import (
+	"fmt"
+	"math"
 	"math/big"
 	"regexp"
 	"strconv"
@@ -117,8 +119,16 @@ func String(v xdr.Int64) string {
 	return StringFromInt64(int64(v))
 }
 
-// TODO: Make this actually work.
+// String128 converts a *Classic-compatible* 128-bit integer into a string. This
+// means that it's 128 bits in name only, and this will panic if its underlying
+// value exceeds an int64. A panic is okay here because it indicates a violation
+// of invariants that are key to the Stellar Asset Contract.
 func String128(v xdr.Int128Parts) string {
+	if v.Hi != 0 || v.Lo > math.MaxInt64 {
+		panic(fmt.Errorf("expected int64, got hi=%d, lo=%d",
+			v.Hi, v.Lo))
+	}
+
 	return StringFromInt64(int64(v.Lo))
 }
 
--- a/amount/main.go
+++ b/amount/main.go
@@ -119,6 +119,9 @@ func String(v xdr.Int64) string {
 
 // String128 converts a signed 128-bit integer into a string, boldly assuming
 // 7-decimal precision.
+//
+// TODO: This should be adapted to variable precision when appopriate, but 7
+// decimals is the correct default for Stellar Classic amounts.
 func String128(v xdr.Int128Parts) string {
 	// the upper half of the i128 always indicates its sign regardless of its
 	// value, just like a native signed type
--- a/amount/main.go
+++ b/amount/main.go
@@ -123,9 +123,9 @@ func String128(v xdr.Int128Parts) string {
 	// the upper half of the i128 always indicates its sign regardless of its
 	// value, just like a native signed type
 	val := big.NewInt(int64(v.Hi))
-	val.Lsh(val, 64).Add(val, big.NewInt(0).SetUint64(uint64(v.Lo)))
+	val.Lsh(val, 64).Add(val, new(big.Int).SetUint64(uint64(v.Lo)))
 
-	rat := big.NewRat(0, 1).SetInt(val)
+	rat := new(big.Rat).SetInt(val)
 	rat.Quo(rat, bigOne)
 	return rat.FloatString(7)
 }
--- a/amount/main.go
+++ b/amount/main.go
@@ -10,8 +10,6 @@
 package amount
 
 import (
-	"fmt"
-	"math"
 	"math/big"
 	"regexp"
 	"strconv"
@@ -119,17 +117,17 @@ func String(v xdr.Int64) string {
 	return StringFromInt64(int64(v))
 }
 
-// String128 converts a *Classic-compatible* 128-bit integer into a string. This
-// means that it's 128 bits in name only, and this will panic if its underlying
-// value exceeds an int64. A panic is okay here because it indicates a violation
-// of invariants that are key to the Stellar Asset Contract.
+// String128 converts a signed 128-bit integer into a string, boldly assuming
+// 7-decimal precision.
 func String128(v xdr.Int128Parts) string {
-	if v.Hi != 0 || v.Lo > math.MaxInt64 {
-		panic(fmt.Errorf("expected int64, got hi=%d, lo=%d",
-			v.Hi, v.Lo))
-	}
+	// the upper half of the i128 always indicates its sign regardless of its
+	// value, just like a native signed type
+	val := big.NewInt(int64(v.Hi))
+	val.Lsh(val, 64).Add(val, big.NewInt(0).SetUint64(uint64(v.Lo)))
 
-	return StringFromInt64(int64(v.Lo))
+	rat := big.NewRat(0, 1).SetInt(val)
+	rat.Quo(rat, bigOne)
+	return rat.FloatString(7)
 }
 
 // StringFromInt64 returns an "amount string" from the provided raw int64 value `v`.
--- a/services/horizon/internal/ingest/processors/effects_processor.go
+++ b/services/horizon/internal/ingest/processors/effects_processor.go
@@ -1438,14 +1438,13 @@ func (e *effectsWrapper) addInvokeHostFunctionEffects(events []contractevents.Ev
 		case contractevents.EventTypeTransfer:
 			xferEvent := evt.(*contractevents.TransferEvent)
 			details["amount"] = amount.String128(xferEvent.Amount)
-			if strkey.IsValidContract(xferEvent.From) {
+			if strkey.IsValidEd25519PublicKey(xferEvent.From) ||
+				strkey.IsValidEd25519PublicKey(xferEvent.To) {
 				e.addUnmuxed(
 					xdr.MustAddressPtr(xferEvent.From),
 					history.EffectAccountDebited,
 					details,
 				)
-			}
-			if strkey.IsValidContract(xferEvent.To) {
 				e.addUnmuxed(
 					xdr.MustAddressPtr(xferEvent.To),
 					history.EffectAccountCredited,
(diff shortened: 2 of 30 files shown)


services/horizon/internal/ingest/processors/effects_processor.go
```diff
@@ -1386,3 +1408,78 @@ func (e *effectsWrapper) addLiquidityPoolWithdrawEffect() error {
 	e.addMuxed(e.operation.SourceAccount(), history.EffectLiquidityPoolWithdrew, details)
 	return nil
 }
+
+// addInvokeHostFunctionEffects iterates through the events and generates
+// account_credited and account_debited effects when it sees events related to
+// the Stellar Asset Contract corresponding to those effects.
+func (e *effectsWrapper) addInvokeHostFunctionEffects(events []contractevents.Event) error {
```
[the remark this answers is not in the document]
Shaptic: Thanks for raising these subtleties!

The code should handle accounts and contracts transparently by nature of `contractevents` using `strkey` appropriately to fill out the e.g. `From` fields, and because these are SAC events, they can't have more than 64 bits of an asset because of the limitation from Stellar Classic (if I understand it correctly). (The new `amount.String128` helper should handle this sanely, too, but we'd need more work across the board to understand non-standard events anyway.)

However, you make a compelling point that the name of the effect itself implies that it should only have `G...` (or `M...`, but that doesn't apply for contracts) values in the fields. I guess, in my mind, contracts are a form of accounts, but you're right that it may not be intuitive. If we drop events with `C...` values, though, there are two issues I see: 

 - **many** events will have at least one field be a contract (by nature of the event coming from a contract), e.g. anyone letting a contract hold a balance on their behalf will cause a `transfer-to-contract` event, and this is the first thing many will do
 - excluding said events gives an incomplete picture of the credit/debit effects pertaining to an asset

There are a few ways to resolve this, I think:

 1. introduce `contract_debited` and `contract_credited` effects: This gives a fuller picture and can handle, for example, intra-contract transfers. The downside is that downstream users now need to 
 2. let `account_debited` and `account_credited` include contracts: aka do nothing different in this PR. This has the issue of being confusing in name (as you mentioned) if there are purely contract-to-contract events captured here.
 3. introduce generic `debited` and `credited` effects and drop the `account_*` versions: this is a breaking change and would probably be a bunch of work, but it would give everyone a unified view into what's happening to an asset

(This isn't an exhaustive list by any means!)

Thoughts? I can also cross-post to Slack so we can get more design input.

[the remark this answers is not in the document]
Shaptic: > through the integration tests I wrote for asset stats I discovered that contracts can hold more than 64 bits worth of an asset. only the accounts are restricted in having at most 64 bits.

Oh woah, okay, yeah that makes sense, e.g. if two accounts hold the max amount and both deposit into a contract then it should work. Noted, great catch :clap: 

Good point regarding contracts. I do think that the debit/credit of an asset into/out of a contract lies directly on the bridge between Classic and Soroban, but contract-to-contract can be out of scope. We can cross that bridge if we get to it. It's unfortunate that we still have to store/propogate/parse the whole event only to find it as being irrelevant at the final stage, but :man_shrugging: 

I'll amend this PR to include `account_debited` when the funds come from an account (regardless of destination, incl. contracts) and include `account_credited` when the funds come to an account (again, regardless of source).

services/horizon/internal/ingest/processors/effects_processor.go
```diff
@@ -1386,3 +1408,78 @@ func (e *effectsWrapper) addLiquidityPoolWithdrawEffect() error {
...
+				history.EffectAccountCredited,
+				details,
+			)
+
+		case contractevents.EventTypeBurn:
+			burnEvent := evt.(*contractevents.BurnEvent)
+			details["amount"] = amount.String128(burnEvent.Amount)
+			e.addUnmuxed(
+				xdr.MustAddressPtr(burnEvent.From),
+				history.EffectAccountCredited,
```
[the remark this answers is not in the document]
Shaptic: bizarre - guess that just shows not to implicitly trust the first result on Google :facepalm: I flipped these all around in the latest push(es)

services/horizon/internal/ingest/processors/effects_processor.go
```diff
@@ -1386,3 +1408,92 @@ func (e *effectsWrapper) addLiquidityPoolWithdrawEffect() error {
...
+					details,
+				)
+			}
+
+		// Clawback events result in a debit to the `from` address, but acts
+		// like a burn to the recipient, so these are functionally equivalent
+		case contractevents.EventTypeClawback:
+			cbEvent := evt.(*contractevents.ClawbackEvent)
+			details["amount"] = amount.String128(cbEvent.Amount)
+			if cbEvent.From[0] != 'C' {
```
sreuland, line 1473: minor for consideration, encapsulate as `contractevents.IsContractAddress(string) bool`, will likely be (re)using this in many places.

services/horizon/internal/ingest/processors/effects_processor.go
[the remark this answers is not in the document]
sreuland: good discovery, I think this aspect of dropping sac events which are strictly contract-2-contract(have no classic account reference) as they are not applicable to current horizon data model, applies when deriving payments as well, will take same approach on [the upcoming pr for that](https://github.com/stellar/go/pull/4807)

amount/main.go
```diff
@@ -117,6 +117,19 @@ func String(v xdr.Int64) string {
 	return StringFromInt64(int64(v))
 }
 
+// String128 converts a signed 128-bit integer into a string, boldly assuming
+// 7-decimal precision.
```
sreuland, line 121: where did you find this spec on the sac event payload amounts having the decimal scale factored in? wanted to refer to it for context, thanks!

services/horizon/internal/ingest/processors/effects_processor.go
Shaptic: used `strkey`'s validation helpers instead!

services/horizon/internal/integration/sac_test.go
```diff
@@ -299,7 +314,7 @@ func TestContractTransferBetweenContracts(t *testing.T) {
 	)
```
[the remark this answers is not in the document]
Shaptic: thanks for your comments! I think I've added assertions in all of the scenarios where appropriate, now :pray:

amount/main.go
Shaptic: I didn't :joy: this should probably be a TODO, in the sense that it doesn't handle variable-precision, but that only applies to non-SAC amounts. In the SAC case, it's always 7 decimals because that's the Stellar standard: https://developers.stellar.org/docs/fundamentals-and-concepts/stellar-data-structures/assets#amount-precision.

services/horizon/internal/ingest/processors/effects_processor.go
```diff
@@ -1386,3 +1409,96 @@ func (e *effectsWrapper) addLiquidityPoolWithdrawEffect() error {
...
+
+		switch evt.GetType() {
+		// Transfer events generate an `account_debited` effect for the `from`
+		// (sender) and an `account_credited` effect for the `to` (recipient).
+		case contractevents.EventTypeTransfer:
+			xferEvent := evt.(*contractevents.TransferEvent)
+			details["amount"] = amount.String128(xferEvent.Amount)
+			if strkey.IsValidEd25519PublicKey(xferEvent.From) ||
+				strkey.IsValidEd25519PublicKey(xferEvent.To) {
+				e.add(
```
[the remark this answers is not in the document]
Shaptic: I had that initially, but my thought later was that a transfer event doesn't make sense without full context.

With the other events, there's only one party involved (e.g. a mint implicitly comes from the issuer), whereas a transfer doesn't make sense if you can only see the `to` just because there was a `from` to the contract. In other words, if I'm watching debits of an asset from my account, I'm going to be confused/upset that I can't see where the debit went, just because it went to a contract.

Maybe I'm overthinking this? I'm happy to revert the behavior if the above thought is misguided.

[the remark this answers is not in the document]
sreuland: tracking [same design question for payments representations pr](https://github.com/stellar/go/pull/4807#discussion_r1140762278), in terms of these half-n-half sac balance changed events with one participant as a classic account and other participant is a contract.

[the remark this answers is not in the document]
Shaptic: > Otherwise, it will probably be a breaking change for clients because they don't expect contract strings for the account string.

I forgot about this; great point, I'm convinced! Separated out in 5a1d104.

I think we can add contract support separately to the server side (at least as it pertains to this issue, since it's such a small amount of work) once we have it on the client side. I can make a separate issue to address & scope this.

services/horizon/internal/ingest/processors/effects_processor.go
```diff
@@ -173,8 +181,9 @@ func (operation *transactionOperationWrapper) effects() ([]effect, error) {
 	}
 
 	wrapper := &effectsWrapper{
-		effects:   []effect{},
-		operation: operation,
+		effects:    []effect{},
+		operation:  operation,
+		passphrase: operation.network,
```
sreuland, line 186: can `passphrase` be removed, as it's reference existing data already available from `operation`?

services/horizon/internal/ingest/processors/effects_processor.go
```diff
@@ -1386,3 +1409,97 @@ func (e *effectsWrapper) addLiquidityPoolWithdrawEffect() error {
...
+				e.add(
+					burnEvent.From,
+					null.String{},
+					history.EffectAccountDebited,
+					details,
+				)
+			}
+
+		// Other events are irrelevant to debit/credit effects
+		default:
```
sreuland, line 1499: is there an idiomatic subtlety for the no-op default clause, it seems self-describing w/o, it's worth extra code?

services/horizon/internal/ingest/processors/effects_processor_test.go
```diff
@@ -3453,5 +3463,200 @@ func TestLiquidityPoolEffects(t *testing.T) {
...
+		return keypair.MustRandom().Address()
+	}
+
+	admin := randAddr()
+	asset := xdr.MustNewCreditAsset("TESTER", admin)
+	// nativeAsset := xdr.MustNewNativeAsset()
+	from, to := randAddr(), randAddr()
+
+	amount := big.NewInt(12345)
+	// bigAmount := big.NewInt(math.MaxInt64)
```
sreuland, line 3479: super minor, is this dead code, can it be removed?

services/horizon/internal/integration/sac_test.go
```diff
@@ -700,6 +757,46 @@ func assertAssetStats(itest *integration.Test, expected assetStats) {
...
+}
+
+// getTxEffects returns a transaction's effects, limited to 2 because it's to be
+// used for checking SAC effects.
+func getTxEffects(itest *integration.Test, txHash string, asset xdr.Asset) []effects.Effect {
+	t := itest.CurrentTest()
+	effects, err := itest.Client().Effects(horizonclient.EffectRequest{
+		ForTransaction: txHash,
+		Order:          horizonclient.OrderDesc,
+		Limit:          2, // no tx should have more than 2 effects
```
sreuland, line 788: would it be worth allowing larger limit and\ asserting size=2 on the results, in case a regression was lurking?

sreuland (approved):
nice work!

services/horizon/internal/integration/sac_test.go
Shaptic: Good call!

services/horizon/internal/ingest/processors/effects_processor.go
Shaptic: I guess just showing that it's intentional rather than forgetting some other event types :shrug:

services/horizon/internal/ingest/processors/effects_processor.go
Shaptic: good call!

services/horizon/internal/ingest/processors/effects_processor_test.go
Shaptic: better yet: added to new tests!

PR #4807, opened 2023-03-14
#4728: include contract asset balance changes in payments listing end…

### LINKED ISSUE #4728: horizon: add single asset balance payment support
master issue: #4722 

Payments

@bartekn raised a good point: horizon has a payments endpoint which returns all operations that result in a payment to / from an account. Since stellar assets can be transferred via smart contract invocations it would make sense to include those operations in the payments endpoint.

prerequisite: #4747

sreuland:
### What

draft on using the new `contractevents` package and applying asset balance changes from contract events in txmeta into `/payments` api endpoint.

### Why

the `/payments` api endpoint needs to include asset movements initiated by contract, which aren't emitted as classic payment operations.

Closes #4728

### DIFF
--- a/protocols/horizon/operations/main.go
+++ b/protocols/horizon/operations/main.go
@@ -352,9 +352,10 @@ type LiquidityPoolWithdraw struct {
 // just contains a base64 encoded string of it's xdr serialization.
 type InvokeHostFunction struct {
 	Base
-	Parameters []HostFunctionParameter `json:"parameters"`
-	Function   string                  `json:"function"`
-	Footprint  string                  `json:"footprint"`
+	Parameters          []HostFunctionParameter      `json:"parameters"`
+	Function            string                       `json:"function"`
+	Footprint           string                       `json:"footprint"`
+	AssetBalanceChanges []AssetContractBalanceChange `json:"asset_balance_changes"`
 }
 
 // InvokeHostFunction parameter model, intentionally simplified, Value
@@ -362,6 +363,17 @@ type InvokeHostFunction struct {
 type HostFunctionParameter struct {
 	Value string `json:"value"`
 	Type  string `json:"type"`
+}
+
+type AssetContractBalanceChange struct {
+	base.Asset
+	From        string `json:"from"`
+	FromMuxed   string `json:"from_muxed,omitempty"`
+	FromMuxedID uint64 `json:"from_muxed_id,omitempty,string"`
+	To          string `json:"to"`
+	ToMuxed     string `json:"to_muxed,omitempty"`
+	ToMuxedID   uint64 `json:"to_muxed_id,omitempty,string"`
+	Amount      string `json:"amount"`
 }
 
 // Operation interface contains methods implemented by the operation types
--- a/services/horizon/internal/db2/history/operation.go
+++ b/services/horizon/internal/db2/history/operation.go
@@ -236,15 +236,18 @@ func (q *OperationsQ) ForTransaction(ctx context.Context, hash string) *Operatio
 
 // OnlyPayments filters the query being built to only include operations that
 // are in the "payment" class of operations:  CreateAccountOps, Payments, and
-// PathPayments.
+// PathPayments and contract asset balance changes expressed in 'asset_balance_changed' flag.
 func (q *OperationsQ) OnlyPayments() *OperationsQ {
-	q.sql = q.sql.Where(sq.Eq{"hop.type": []xdr.OperationType{
-		xdr.OperationTypeCreateAccount,
-		xdr.OperationTypePayment,
-		xdr.OperationTypePathPaymentStrictReceive,
-		xdr.OperationTypePathPaymentStrictSend,
-		xdr.OperationTypeAccountMerge,
-	}})
+	q.sql = q.sql.Where(sq.Or{
+		sq.Eq{"hop.type": []xdr.OperationType{
+			xdr.OperationTypeCreateAccount,
+			xdr.OperationTypePayment,
+			xdr.OperationTypePathPaymentStrictReceive,
+			xdr.OperationTypePathPaymentStrictSend,
+			xdr.OperationTypeAccountMerge,
+		}},
+		sq.Eq{"hop.asset_balance_changed": 1}})
+
 	return q
 }
 
@@ -390,6 +393,7 @@ var selectOperation = sq.Select(
 		"hop.details, " +
 		"hop.source_account, " +
 		"hop.source_account_muxed, " +
+		"hop.asset_balance_changed, " +
 		"ht.transaction_hash, " +
 		"ht.tx_result, " +
 		"COALESCE(ht.successful, true) as transaction_successful").
--- a/services/horizon/internal/db2/schema/migrations/64_add_payment_flag_history_ops.sql
+++ b/services/horizon/internal/db2/schema/migrations/64_add_payment_flag_history_ops.sql
@@ -1,0 +1,7 @@
+-- +migrate Up
+
+ALTER TABLE history_operations ADD asset_balance_changed smallint DEFAULT 0;
+
+-- +migrate Down
+
+ALTER TABLE history_operations DROP COLUMN asset_balance_changed;
--- a/services/horizon/internal/ingest/processors/operations_processor.go
+++ b/services/horizon/internal/ingest/processors/operations_processor.go
@@ -11,6 +11,7 @@ import (
 	"github.com/stellar/go/ingest"
 	"github.com/stellar/go/protocols/horizon/base"
 	"github.com/stellar/go/services/horizon/internal/db2/history"
+	"github.com/stellar/go/support/contractevents"
 	"github.com/stellar/go/support/errors"
 	"github.com/stellar/go/toid"
 	"github.com/stellar/go/xdr"
@@ -20,15 +21,17 @@ import (
 type OperationProcessor struct {
 	operationsQ history.QOperations
 
-	sequence uint32
-	batch    history.OperationBatchInsertBuilder
+	sequence          uint32
+	batch             history.OperationBatchInsertBuilder
+	networkPassphrase string
 }
 
-func NewOperationProcessor(operationsQ history.QOperations, sequence uint32) *OperationProcessor {
+func NewOperationProcessor(operationsQ history.QOperations, sequence uint32, networkPassphrase string) *OperationProcessor {
 	return &OperationProcessor{
-		operationsQ: operationsQ,
-		sequence:    sequence,
-		batch:       operationsQ.NewOperationBatchInsertBuilder(maxBatchSize),
+		operationsQ:       operationsQ,
+		sequence:          sequence,
+		batch:             operationsQ.NewOperationBatchInsertBuilder(maxBatchSize),
+		networkPassphrase: networkPassphrase,
 	}
 }
 
@@ -36,10 +39,11 @@ func NewOperationProcessor(operationsQ history.QOperations, sequence uint32) *Op
 func (p *OperationProcessor) ProcessTransaction(ctx context.Context, transaction ingest.LedgerTransaction) error {
 	for i, op := range transaction.Envelope.Operations() {
 		operation := transactionOperationWrapper{
-			index:          uint32(i),
-			transaction:    transaction,
-			operation:      op,
-			ledgerSequence: p.sequence,
+			index:             uint32(i),
+			transaction:       transaction,
+			operation:         op,
+			ledgerSequence:    p.sequence,
+			networkPassphrase: p.networkPassphrase,
 		}
 		details, err := operation.Details()
 		if err != nil {
@@ -65,6 +69,7 @@ func (p *OperationProcessor) ProcessTransaction(ctx context.Context, transaction
 			detailsJSON,
 			acID.Address(),
 			sourceAccountMuxed,
+			operation.AssetBalanceChanged(),
 		); err != nil {
 			return errors.Wrap(err, "Error batch inserting operation rows")
 		}
@@ -79,10 +84,11 @@ func (p *OperationProcessor) Commit(ctx context.Context) error {
 
 // transactionOperationWrapper represents the data for a single operation within a transaction
 type transactionOperationWrapper struct {
-	index          uint32
-	transaction    ingest.LedgerTransaction
-	operation      xdr.Operation
-	ledgerSequence uint32
+	index             uint32
+	transaction       ingest.LedgerTransaction
+	operation         xdr.Operation
+	ledgerSequence    uint32
+	networkPassphrase string
 }
 
 // ID returns the ID for the operation.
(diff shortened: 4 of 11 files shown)


services/horizon/internal/ingest/processors/operations_processor.go
```diff
@@ -655,6 +705,64 @@ func (operation *transactionOperationWrapper) Details() (map[string]interface{},
...
+	return balanceChanges, nil
+}
+
+// fromAccount - strkey format of contract or address
+// toAccount - strkey format of contract or address, or nillable
+// amountChanged - absolute value that asset balance changed
+// asset - the fully qualified issuer:code for asset that had balance change
+//
+// return - a balance changed record expressed as map of key/value's
+func createSACBalanceChangeEntry(fromAccount string, toAccount string, amountChanged xdr.Uint64, asset xdr.Asset) map[string]interface{} {
```
sreuland, line 754: need to get clarification on correct interpretation of the i128 bits in amount , this first pass of assuming just the lowest 64 bits as uint64 is incorrect.

services/horizon/internal/ingest/processors/operations_processor.go
```diff
@@ -655,6 +705,64 @@ func (operation *transactionOperationWrapper) Details() (map[string]interface{},
...
 
+// Searches an operation for SAC events that are of a type which represent asset balances changed between contracts and/or classic.
+// SAC events have a one-to-one association to SAC contract fn invocations, i.e. invoke the 'mint' function, will trigger one Mint Event to be emitted capturing the fn args.
+// SAC events that involve asset balance changes follow some standard data formats:
+//     amount expressed as uint64 only, the event type in this case provides the context of whether an amount was credit/debit to a balance.
+//     the 'from' and 'to' attributes represent an account or a contract
+func (operation *transactionOperationWrapper) parseAssetBalanceChangesFromContractEvents() ([]map[string]interface{}, error) {
+	balanceChanges := []map[string]interface{}{}
+
+	txMeta, ok := operation.transaction.UnsafeMeta.GetV3()
```
Shaptic, line 716: I think you can leverage `operation.transaction.GetOperationEvents(...)` here

services/horizon/internal/ingest/processors/operations_processor.go
```diff
@@ -655,6 +705,64 @@ func (operation *transactionOperationWrapper) Details() (map[string]interface{},
...
+//
+// return - a balance changed record expressed as map of key/value's
+func createSACBalanceChangeEntry(fromAccount string, toAccount string, amountChanged xdr.Uint64, asset xdr.Asset) map[string]interface{} {
+	balanceChange := map[string]interface{}{}
+
+	balanceChange["from"] = fromAccount
+	if toAccount != "" {
+		balanceChange["to"] = toAccount
+	}
+	balanceChange["amount"] = string(amountChanged)
```
Shaptic, line 761: bit of a nit: any reason not to initialize it directly with the "known" fields? e.g.
```suggestion
	balanceChange := map[string]interface{}{
		"amount": string(amountChanged),
		"from": fromAccount,
	}
	if toAccount != "" {
		balanceChange["to"] = toAccount
	}
```

Shaptic: I find it easier to visualize the structure this way rather than adding entries one by one in my head

Shaptic (commented):
I know it's a draft, but just adding some comments as you go!

services/horizon/internal/db2/schema/migrations/64_add_payment_flag_history_ops.sql
```diff
@@ -0,0 +1,7 @@
+-- +migrate Up
+
+ALTER TABLE history_operations ADD asset_balance_changed smallint DEFAULT 0;
```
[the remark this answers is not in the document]
sreuland: yes, we discussed that further in team meet, will go back to `is_payment`.

services/horizon/internal/db2/history/operation.go
```diff
@@ -236,15 +236,18 @@ func (q *OperationsQ) ForTransaction(ctx context.Context, hash string) *Operatio
...
-	}})
+	q.sql = q.sql.Where(sq.Or{
+		sq.Eq{"hop.type": []xdr.OperationType{
+			xdr.OperationTypeCreateAccount,
+			xdr.OperationTypePayment,
+			xdr.OperationTypePathPaymentStrictReceive,
+			xdr.OperationTypePathPaymentStrictSend,
+			xdr.OperationTypeAccountMerge,
+		}},
+		sq.Eq{"hop.asset_balance_changed": 1}})
```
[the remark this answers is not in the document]
sreuland: yes, ideally don't want to attempt to update that flag as part of migration, due to number of rows potentially in hops table.

services/horizon/internal/ingest/processors/operations_processor.go
sreuland: pro tip, thanks!

services/horizon/internal/ingest/processors/operations_processor.go
sreuland: sure, may be a little more fluent.

protocols/horizon/operations/main.go
```diff
@@ -352,9 +352,10 @@ type LiquidityPoolWithdraw struct {
 // just contains a base64 encoded string of it's xdr serialization.
 type InvokeHostFunction struct {
 	Base
-	Parameters []HostFunctionParameter `json:"parameters"`
-	Function   string                  `json:"function"`
-	Footprint  string                  `json:"footprint"`
+	Parameters          []HostFunctionParameter      `json:"parameters"`
+	Function            string                       `json:"function"`
+	Footprint           string                       `json:"footprint"`
+	AssetBalanceChanges []AssetContractBalanceChange `json:"asset_balance_changes"`
```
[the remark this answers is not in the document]
Shaptic: My general concern is that this sets a precedent we may not want to set: i.e. that Horizon will give you nice JSON interpretations of the side-effects of host function invocations as part of its responses. I can imagine us constantly getting feature requests for some other side-effects of contract calls. But because contract calls are so opaque, we should have a "hands off" policy that moves the complexity to the client, who knows best how to interpret the results.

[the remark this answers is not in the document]
sreuland: ah, ok, thanks for the insight, gtk, no new data models that reference contract aspects, but where contract model can be reflected through an existing horizon model like payments, then it may be worthwhile, i'll remove `AssetBalanceChanges`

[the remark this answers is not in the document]
Shaptic: Right, yeah, that was my thought - but to be clear this is just my opinion; we can/should open up the discussion to the team, maybe there's downstream value to adding this that we aren't aware of?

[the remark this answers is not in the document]
sreuland: I was looking at the resource model of other operations that qualify for 'is payment'  .e.g. Payment, CreateAccount, AccountMerge, etc. Those api resource models include attributes representing some generalized notion of payment aka `to account`,`from account`,`amount`,  but with naming terms more specific to the context of the op like `funder` as the 'from account' for the CreateAccountOp. 

I think can apply that same pattern of payment oriented info in the resource model for InvokeHostFunctionOp, with `asset_balance_changed` having an array of `{to,from,amount}`. This may retain some consistency/convenience for downstream api usage, with a precedence of payment details being present in the op resource. I'll ask for review on this in chat, thanks!

services/horizon/internal/ingest/processors/operations_processor.go
```diff
@@ -655,6 +706,77 @@ func (operation *transactionOperationWrapper) Details() (map[string]interface{},
...
+}
+
+// fromAccount - strkey format of contract or address
+// toAccount - strkey format of contract or address, or nillable
+// amountChanged - absolute value that asset balance changed
+// asset - the fully qualified issuer:code for asset that had balance change
+//
+// return - a balance changed record expressed as map of key/value's
+func createSACBalanceChangeEntry(fromAccount string, toAccount string, amountChanged xdr.Int128Parts, asset xdr.Asset) map[string]interface{} {
+	balanceChange := map[string]interface{}{}
```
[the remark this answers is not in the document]
sreuland: yes. that would be nice context to have, added 'type' to the json model in details - https://github.com/stellar/go/pull/4807/commits/1af0c9aaf79dc9213e416729cbbf468d50e2d0da

protocols/horizon/operations/main.go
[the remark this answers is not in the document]
Shaptic: That's true, but the big difference is that those operations will **always** have those fields, whereas the InvokeHostFunctionOp will **only** have them for payment-like, Stellar Asset Contract specific invocations. So the array will only sometimes (in fact, probably rarely once the ecosystem takes off) be filled.

And what if, later, we start parsing other things out of the events? Will this schema continue to grow and grow with every Horizon feature request? There will be `[]AssetBalanceChanges`, `[]Participants`, `[]AllowanceChanges`, etc. etc. That's what I meant about setting a precedent.

Happy to discuss more in Slack, will keep an eye out for the thread!

services/horizon/internal/test/scenarios/base-horizon.sql
```diff
@@ -495,7 +495,8 @@ CREATE TABLE history_operations (
     application_order integer NOT NULL,
     type integer NOT NULL,
     details jsonb,
-    source_account character varying(64) DEFAULT ''::character varying NOT NULL
+    source_account character varying(64) DEFAULT ''::character varying NOT NULL,
+    is_payment smallint default 0
```
sreuland, line 499: ouch, dragging these scenarios along, don't have a choice as they fail against new db schema for history_ops table.

services/horizon/internal/ingest/processors/operations_processor.go
sreuland: uses a temp function for conversion now, have TODO - https://github.com/stellar/go/pull/4807/files#diff-5f37fbbec67583bbaa94cbe645d8a4aa046505cee2c37be660818ef340645ab1R797 to refactor to use common function `amount.String128()`

services/horizon/internal/db2/schema/migrations/64_add_payment_flag_history_ops.sql
```diff
@@ -0,0 +1,9 @@
+-- +migrate Up
+
+ALTER TABLE history_operations ADD is_payment smallint DEFAULT 0;
```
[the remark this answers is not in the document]
sreuland: yes, good points, this can be a huge table in pubnet deployments and this suggestion helps with migration performance and overall semantics, thanks! - https://github.com/stellar/go/pull/4807/commits/f9ea51becabaddb9f7f66e75171c8a933cee3551

services/horizon/internal/ingest/processors/operations_processor.go
```diff
@@ -655,6 +706,103 @@ func (operation *transactionOperationWrapper) Details() (map[string]interface{},
...
+// Searches an operation for SAC events that are of a type which represent
+//    Asset balances changed with at least one participant of the change referencing a classic account.
+//
+// SAC events have a one-to-one association to SAC contract fn invocations,
+//    i.e. invoke the 'mint' function, will trigger one Mint Event to be emitted capturing the fn args.
+//
+// SAC events that involve asset balance changes follow some standard data formats:
+//     The 'amount' in the event is expressed as Int128Parts, however it's expected that it's value will not be signed.
+//     It represents a absolute delta, the event type can provide the context of whether an amount
+//     was considered incremental or decreental, i.e. credit or debit to a balance.
```
Shaptic, line 718: nit: typo
```suggestion
//     was considered incremental or decremental, i.e. credit or debit to a balance.
```

support/contractevents/utils.go
```diff
@@ -38,6 +38,10 @@ func MustScAddressToString(address *xdr.ScAddress) string {
 	return result
 }
 
+func IsContractAddress(account string) bool {
+	return account[0] == 'C'
+}
+
```
Shaptic, line 44: I like the thought and I noted your comment on my PR, but this makes way more sense in the `strkey` package, since you can get full, proper validation. See this commit for a reference implementation: https://github.com/stellar/go/pull/4806/commits/5b36a76fc989459af2cf090183516c1d21bdb39f.

But what I realized later is that what I really wanted was `strkey.IsValidEd25519PublicKey` to check for an account as opposed to checking for "not a contract" (see this commit: https://github.com/stellar/go/pull/4806/commits/61b281db4ca5edf5ae4d7d5cdc316c91d8cd9534 for that).

I think you want the same idea here.

sreuland: another pro tip, great pivot, thanks!

services/horizon/internal/ingest/processors/operations_processor.go
```diff
@@ -655,6 +708,105 @@ func (operation *transactionOperationWrapper) Details() (map[string]interface{},
...
+
+	if toAccount != "" {
+		if strkey.IsValidEd25519PublicKey(toAccount) {
+			// TODO, place the 'C...' contract address here, if/when horizon has facility for
+			// contract addresses
+			balanceChange["to"] = "contract"
+		} else {
+			balanceChange["to"] = toAccount
+		}
+	}
```
Shaptic, line 794: oop, mishaps here after the last commit: you had `isContract` before, so you want `!isPublicKey` now, or maybe you want to add a validator for it to `strkey`.

If you're going to put `"contract"` into the entry, though, you can just put `fromAccount` and `toAccount` into them directly? But the outcome of this is probably tightly related to the outcome of the discussion @tamirms and I are having [here](https://github.com/stellar/go/pull/4806#discussion_r1139836753) with the same question regarding including contract IDs in effects.

sreuland: yep, this is sans tests still, working on that, thanks for heads up.

Yes, on the contract addresses, I'm trying to include the 'half-in' sac events, with one participant of the asset balance change triggered by contract referring to a classic account, but then not sure what to put for the opposing participant, or, maybe we just omit the `from` or `to` if it was for contract? @tamirms , did you have thoughts on how this payment representation would be best for downstream clients to use?

protocols/horizon/operations/main.go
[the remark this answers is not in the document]
tsachiherman: I'd be happy to hear @mollykarcher's thought around that as well. My intuition is that in this particular case, I don't see a great value in returning the parsed value assuming the client SDK is capable of figuring these by other means. I do not believe that there is a right/wrong decision here. Given that we don't have any concrete consumers that require that ( afaik ), I would opt into skipping on adding these features. We would always be able to add these in the future, but taking these away would be an issue.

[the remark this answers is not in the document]
Shaptic: We aligned offline: it makes sense to include this in the response because it's already parsed out as part of supporting `/payments`.

services/horizon/internal/db2/history/operation.go
```diff
@@ -235,16 +235,20 @@ func (q *OperationsQ) ForTransaction(ctx context.Context, hash string) *Operatio
...
-	}})
+	q.sql = q.sql.Where(sq.Or{
+		sq.Eq{"hop.type": []xdr.OperationType{
+			xdr.OperationTypeCreateAccount,
+			xdr.OperationTypePayment,
+			xdr.OperationTypePathPaymentStrictReceive,
+			xdr.OperationTypePathPaymentStrictSend,
+			xdr.OperationTypeAccountMerge,
+		}},
+		sq.Eq{"hop.is_payment": true}})
```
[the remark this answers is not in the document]
sreuland: ok, on staging, there won't be many rows with is_payment being non-null yet to begin with, just forward ingestion history after the version is installed, great suggestion, I need to confirm though on the core ingestion config, b/c this version requires the soroban enabled core and staging has config for the non-soroban core.

PR #4813, opened 2023-03-16
474/soroban diagnostic events

paulbellamy:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add support for captive-core to emit Soroban diagnostic events

### Why

Part of https://github.com/stellar/soroban-tools/issues/474

- For the captive-core config, soroban-rpc needs to be able to enable this on captive-core.
- ~For horizon, I just added the flag for completeness. Not super committed to it, as we don't do anything with the events yet anyway (afaik?)~

### Known limitations

[TODO or N/A]

### DIFF
--- a/services/horizon/internal/config.go
+++ b/services/horizon/internal/config.go
@@ -19,17 +19,18 @@ type Config struct {
 	Port               uint
 	AdminPort          uint
 
-	EnableCaptiveCoreIngestion  bool
-	EnableIngestionFiltering    bool
-	UsingDefaultPubnetConfig    bool
-	CaptiveCoreBinaryPath       string
-	RemoteCaptiveCoreURL        string
-	CaptiveCoreConfigPath       string
-	CaptiveCoreTomlParams       ledgerbackend.CaptiveCoreTomlParams
-	CaptiveCoreToml             *ledgerbackend.CaptiveCoreToml
-	CaptiveCoreStoragePath      string
-	CaptiveCoreReuseStoragePath bool
-	CaptiveCoreConfigUseDB      bool
+	EnableCaptiveCoreIngestion                     bool
+	EnableIngestionFiltering                       bool
+	UsingDefaultPubnetConfig                       bool
+	CaptiveCoreBinaryPath                          string
+	RemoteCaptiveCoreURL                           string
+	CaptiveCoreConfigPath                          string
+	CaptiveCoreTomlParams                          ledgerbackend.CaptiveCoreTomlParams
+	CaptiveCoreToml                                *ledgerbackend.CaptiveCoreToml
+	CaptiveCoreStoragePath                         string
+	CaptiveCoreReuseStoragePath                    bool
+	CaptiveCoreConfigUseDB                         bool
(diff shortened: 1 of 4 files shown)


services/horizon/internal/flags.go
```diff
@@ -40,6 +40,8 @@ const (
 	CaptiveCoreConfigPathName = "captive-core-config-path"
 	// captive-core-use-db is the command line flag for enabling captive core runtime to use an external db url connection rather than RAM for ledger states
 	CaptiveCoreConfigUseDB = "captive-core-use-db"
+	// captive-core-enable-soroban-diagnostic-events is the command line flag for enabling soroban diagnostic events in captive core
+	CaptiveCoreConfigEnableSorobanDiagnosticEvents = "captive-core-enable-soroban-diagnostic-events"
```
tsachiherman, line 44: This is not a bug.. but I can't think of any reason we would need this. Am I missing something ?

tsachiherman (commented):
I could be wrong, but I don't believe that horizon would need to support this flag.
This flag would only be used in the soroban-rpc.

services/horizon/internal/config.go
```diff
@@ -19,17 +19,18 @@ type Config struct {
...
+	UsingDefaultPubnetConfig                       bool
+	CaptiveCoreBinaryPath                          string
+	RemoteCaptiveCoreURL                           string
+	CaptiveCoreConfigPath                          string
+	CaptiveCoreTomlParams                          ledgerbackend.CaptiveCoreTomlParams
+	CaptiveCoreToml                                *ledgerbackend.CaptiveCoreToml
+	CaptiveCoreStoragePath                         string
+	CaptiveCoreReuseStoragePath                    bool
+	CaptiveCoreConfigUseDB                         bool
+	CaptiveCoreConfigEnableSorobanDiagnosticEvents bool
```
Shaptic, line 33: Wasn't the point of introducing the captive core toml file (#3442) to reduce the need for adding incredibly specific flags that configure captive core's behavior? We can probably just get away with leaving it there

services/horizon/internal/flags.go
paulbellamy: Horizon shouldn't need it no. Just included it for completeness. Happy to remove if people feel that's better.

services/horizon/internal/config.go
paulbellamy: You mean to horizon? or generally? Like are you suggesting we should just always enable diagnostic events with captive core?

Shaptic: I'm saying that Horizon should parse whatever Captive Core gives it, and people should use the toml file to control whether or not CC includes diagnostic events.

services/horizon/internal/flags.go
paulbellamy: Good point. Removed.

services/horizon/internal/config.go
paulbellamy: Agreed. Removed.

tsachiherman (approved):

Shaptic (approved):
:shipit:

Issue #4823, opened 2023-03-28
services/horizon: remove remote captive core


mollykarcher:
### What problem does your feature solve?

[Remote captive core](https://developers.stellar.org/docs/run-api-server/remote-core) is currently publicized in our docs, but we do not maintain it or test it when we do releases. It has not kept up with feature development on main line (ex. on-disk mode), and we do not dogfood it ourselves so are unaware of any potential bugs or operational issues (ex. recent [issue](https://stellarfoundation.slack.com/archives/C01BC71B1RV/p1679939687179609?thread_ts=1679939668.963439&cid=C01BC71B1RV) with metrics endpoint).

### What would you like to see?

- Remove the configuration options / env variables for it
- Remove all functionality and codepaths associated with it
- Remove any reference to it or mention of it in public-facing documentation

### What alternatives are there?

1. Start maintaining + testing it so that we can confidently document and recommend it. I don't see this option being worthwhile as I find it unlikely anyone is actually using this option, and we don't have a solid use case for it internally
2. Don't remove it from code, just remove it from public-facing documentation. This could still achieve the goal of dissuading people from using it, but it would still be available if they were currently using it
3. Remove it from code, but leave mention of it (and it's deprecation) in public-facing documentation. This could be useful for anyone that was currently using it and give them a clear path as to how to migrate away from it
4. Don't remove it from code or documentation, but change the existing documentation disclaimer about it being experimental to make it clear that it is not maintained at all and to use at your own risk. _This is probably the quickest and safest option that achieves the end goal_

Shaptic:
I support (2), since it's still a useful piece of code (even if it needs some maintenance), because it already lives in the `exp/` (as in `exp`erimental) folder which should imply that its usage is risky.

PR #4826, opened 2023-03-29
Remove the Remote Captive Core implementation

Shaptic:
Per #4823 and the Platform Team meeting discussion, we're removing all references to Remote Captive Core. The corresponding docs PR is stellar/stellar-docs#111.


mollykarcher (approved):
thanks for hopping on this so fast!

tsachiherman (approved):

PR #4834, opened 2023-04-04
Set xdr version on every download of new xdr

leighmcculloch:
### What
Set xdr version on every download of new xdr

### Why
The xdr_commit_generated.txt file is intended to store the revision of the xdr that has been downloaded and generated. The current Makefile target isn't triggered when running `xdr-update`. A side-effect of that is that in the last Soroban v0.0.7 release the value in the txt file is inaccurate.

For the sake of simplicity, instead of having it be a standalone target that we need to structure to be triggered at the right time, we can simply update it whenever the xdr files are updated. Since any time the xdr files are updated is a time we'd expect the revision to change.


Shaptic (approved):
speedy :clap:

paulbellamy (approved):

PR #4841, opened 2023-04-13
support/log: Add support for JSON log formatting

2opremio:
Needed for https://github.com/stellar/soroban-tools/issues/497 (since soroban-tools uses the `support/log`)


paulbellamy (approved):

tsachiherman (commented):
Implementation is good, but I'd be happy to see a unit test that prove these changes works as expected.

PR #4852, opened 2023-04-20
all: enforce simplified Golang code formatting

2opremio:
### What

Simplify all golang code with `gofmt -s` and enforce that in CI.

### Why

The resulting code is simpler and cleaner

### Known limitations

N/A


tsachiherman (approved):
I was hoping you'd apply this to the xdr-next as well... Or merge it with your other pr.

2opremio:
I will merge it into soroban-xdr-next afterwards.

Issue #4856, opened 2023-04-26
services/horizon: change default for `--captive-core-use-db` to true


mollykarcher:
### What problem does your feature solve?

Currently, there's a lot of inconsistency in what configuration we recommend for horizon, which leads to confusion both internally and externally. At present, the default value for `CAPTIVE_CORE_USE_DB` is `false`, however our recommendations to external partners (and what we've chosen to do in our deployment) is `true`. The size of the ledger grows unbounded over time, so with an in-memory configuration, the RAM required of horizon/captive-core will eventually become unreasonably large and/or prohibitively expensive. 

### What would you like to see?

- [x] Update the default of `CAPTIVE_CORE_USE_DB` to `true`
- [x] Update quickstart to use new default (this may happen automatically, not sure if quickstart is explicitly setting this flag)
- [ ] Update external documentation to reflect new default

Updating the default is a breaking change, as any customers not explicitly setting this flag will experience a behavior flip. However, it's unlikely that anyone using in-memory is doing it explicitly/intentionally and in a worse-case scenario where it causes an issue for them, the revert is quite simple (ie. explicitly set the flag to `false`)

### What alternatives are there?

- Another option would be to remove in-memory option altogether. We've chosen this route for [soroban-rpc](https://github.com/stellar/soroban-tools/issues/615); being a new product offering, we don't want to muddy the waters with this option when our recommendation is to always use on-disk mode. We could go the same route here (and maybe we'll want to eventually), but it's probably best to at least give users the option to explicitly now set/choose this option for now, since it's a quick and easy back-stop if they do notice or complain when the default changes.
- We could consider deprecating the in-memory option (for removal in a later release)

Shaptic:
We also want this to be the new default in the stub configuration provided when you install the `.deb` package. I think it's [here](https://github.com/stellar/packages/blob/master/stellar-captive-core/debian/stellar-captive-core.default).

PR #4877, opened 2023-05-19
services/horizon: Update default for  --captive-core-use-db to true

urvisavla:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Modify the default value of the `--captive-core-use-db` command line flag to true. This modification ensures that CaptiveCore utilizes an external database instead of relying on memory (RAM). The configuration of the external database can be customized within the CaptiveCore configuration file with the default setting being `sqlite3://stellar.db.`

If the `ingest` option is not enabled, this particular flag will be disregarded. Previously, when it was necessary to explicitly set this flag, an error message would be displayed if it was set without configuring other ingestion parameters. However, now, if the ingest option is disabled, this flag will simply be ignored.

To explicitly set it to false via the command line, use the following syntax: `--captive-core-use-db=false`. Alternatively, you can set the environment variable `CAPTIVE_CORE_USE_DB=false` in order to revert to the in-memory option.

### Why
See [4856](https://github.com/stellar/go/issues/4856) for details

### Known limitations
Due to the increasing size of the ledger, supporting the in-memory option is no longer practical and should be deprecated.

### DIFF
--- a/services/horizon/CHANGELOG.md
+++ b/services/horizon/CHANGELOG.md
@@ -5,6 +5,11 @@ file. This project adheres to [Semantic Versioning](http://semver.org/).
 
 
 ## Unreleased
+### Changes
+### Breaking Changes
+- Modify the default value of `--captive-core-use-db` to true ([4856](https://github.com/stellar/go/issues/4856))
+  - This updates the default behavior of captive core to start in on-disk mode.
+  - To continue using the previous in-memory mode, explicitly set the `--captive-core-use-db` flag to false
 
 ## 2.24.1
 
--- a/services/horizon/internal/flags.go
+++ b/services/horizon/internal/flags.go
@@ -177,7 +177,7 @@ func Flags() (*Config, support.ConfigOptions) {
 		&support.ConfigOption{
 			Name:        CaptiveCoreConfigUseDB,
 			OptType:     types.Bool,
-			FlagDefault: false,
+			FlagDefault: true,
 			Required:    false,
 			Usage: `when enabled, Horizon ingestion will instruct the captive
 			              core invocation to use an external db url for ledger states rather than in memory(RAM).\n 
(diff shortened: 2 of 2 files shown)


Shaptic (approved):
Seems good, but this definitely needs a breaking changelog entry.

urvisavla:
> Seems good, but this definitely needs a breaking changelog entry.

Do we need a major version update since it's a breaking change? Asking so that I can update the changelog with appropriate release version.

cc @mollykarcher

urvisavla:
> > Seems good, but this definitely needs a breaking changelog entry.
> 
> Do we need a major version update since it's a breaking change? Asking so that I can update the changelog with appropriate release version.
> 
> cc @mollykarcher

Updated changelog entry.

services/horizon/CHANGELOG.md
```diff
@@ -5,6 +5,11 @@ file. This project adheres to [Semantic Versioning](http://semver.org/).
 
 
 ## Unreleased
+### Changes
+### Breaking Changes
+- Modify the default value of `--captive-core-use-db` to true ([4856](https://github.com/stellar/go/issues/4856))
+  - This updates the default behavior of captive core to start in on-disk mode.
+  - To continue using the previous in-memory mode, explicitly set the `--captive-core-use-db` flag to false
```
Shaptic, line 12: nitpick: spacing on the first bullet. also we don't need an empty `Changes` heading

```suggestion
### Breaking Changes
  - Modify the default value of `--captive-core-use-db` to true ([4856](https://github.com/stellar/go/issues/4856))
  - This updates the default behavior of captive core to start in on-disk mode.
  - To continue using the previous in-memory mode, explicitly set the `--captive-core-use-db` flag to false
```

Shaptic (approved):
@urvisavla nah, we don't follow semver, so just a minor bump is appropriate, but we'll only set that at release time

Issue #4883, opened 2023-05-27
Re-enable SAC integration tests


sreuland:
### What problem does your feature solve?
the SAC integration tests have been disabled, causing lack of coverage related to contract parsing aspects.

### What would you like to see?
identify how to generate the necessary simulation footprint and soroban tx data to include in the InvokeHostFunctionOp for tests to use, since the preflight endpoint was deprecated and the tests were previously using that to obtain.

### Epic
https://github.com/stellar/soroban-tools/issues/657

PR #4887, opened 2023-05-31
services/horizon: Reenable InvokeHostFunction integration tests

2opremio:
Addresses part of #4883 

We still need to reenable the tests for `sac_test.go` (which is going to be harder since we need to figure out the footprint of each transaction).


sreuland (approved):
looks good!

Issue #4902, opened 2023-06-06
services/horizon: race condition found in integration tests


2opremio:
See https://github.com/stellar/go/actions/runs/5187706213/jobs/9350401104?pr=4892

This time it was found in the `soroban-xdr-next` branch, but I am pretty sure it will also surface in `master`:

```
023-06-06T11:16:56.5035479Z ==================
2023-06-06T11:16:56.5035918Z WARNING: DATA RACE
2023-06-06T11:16:56.5036230Z Write at 0x00c00000243c by goroutine 91835:
2023-06-06T11:16:56.5036772Z   github.com/stellar/go/ingest/ledgerbackend.(*CaptiveStellarCore).handleMetaPipeResult()
2023-06-06T11:16:56.5037512Z       /home/runner/work/go/go/ingest/ledgerbackend/captive_core_backend.go:650 +0x78d
2023-06-06T11:16:56.5038064Z   github.com/stellar/go/ingest/ledgerbackend.(*CaptiveStellarCore).GetLedger()
2023-06-06T11:16:56.5039031Z       /home/runner/work/go/go/ingest/ledgerbackend/captive_core_backend.go:605 +0xcf7
2023-06-06T11:16:56.5039651Z   github.com/stellar/go/ingest/ledgerbackend.metricsLedgerBackend.GetLedger()
2023-06-06T11:16:56.5040569Z       /home/runner/work/go/go/ingest/ledgerbackend/metrics.go:37 +0xba
2023-06-06T11:16:56.5041246Z   github.com/stellar/go/ingest/ledgerbackend.(*metricsLedgerBackend).GetLedger()
2023-06-06T11:16:56.5041703Z       <autogenerated>:1 +0x95
2023-06-06T11:16:56.5042159Z   github.com/stellar/go/services/horizon/internal/ingest.resumeState.run()
2023-06-06T11:16:56.5042813Z       /home/runner/work/go/go/services/horizon/internal/ingest/fsm.go:396 +0x4f1
2023-06-06T11:16:56.5043335Z   github.com/stellar/go/services/horizon/internal/ingest.(*resumeState).run()
2023-06-06T11:16:56.5043763Z       <autogenerated>:1 +0x5c
2023-06-06T11:16:56.5044233Z   github.com/stellar/go/services/horizon/internal/ingest.(*system).runStateMachine()
2023-06-06T11:16:56.5044894Z       /home/runner/work/go/go/services/horizon/internal/ingest/main.go:567 +0x398
[...]

PR #4903, opened 2023-06-09
4902 Add mutex for concurrent access in GetLatestLedgerSequence

urvisavla:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add a mutex in `CaptiveStellarCore` to ensure concurrent read access to `GetLatestLedgerSequence` method. The mutex ensures that no write operations occur concurrently with read operations. This is required because of read access of `c.nextLedger` when posting `captive_stellar_core_latest_ledger` metric while being concurrently written in `handleMetaPipeResult`.

This issue does not occur in master because we're not posting `captive_stellar_core_latest_ledger` metric in master. However, `GetLatestLedgerSequence` does get called from outside in `maybeReapLookupTables` in master but it's in the same thread as the ingestion so the mutex is not necessary.

### Why
Fixes data race reported in #4902 

### Known limitations
Locking can have an impact on the performance but we're not expecting any noticeable impact in this case.

### DIFF
--- a/ingest/ledgerbackend/captive_core_backend.go
+++ b/ingest/ledgerbackend/captive_core_backend.go
@@ -97,6 +97,12 @@ type CaptiveStellarCore struct {
 	// cachedMeta keeps that ledger data of the last fetched ledger. Updated in GetLedger().
 	cachedMeta *xdr.LedgerCloseMeta
 
+	// ledgerSequenceLock mutex is used to protect the member variables used in the
+	// read-only GetLatestLedgerSequence method from concurrent write operations.
+	// This is required when GetLatestLedgerSequence is called from other goroutine
+	// such as writing Prometheus metric captive_stellar_core_latest_ledger.
+	ledgerSequenceLock sync.RWMutex
+
 	prepared           *Range  // non-nil if any range is prepared
 	closed             bool    // False until the core is closed
 	nextLedger         uint32  // next ledger expected, error w/ restart if not seen
@@ -307,9 +313,12 @@ func (c *CaptiveStellarCore) openOfflineReplaySubprocess(from, to uint32) error 
 	// The next ledger should be the first ledger of the checkpoint containing
 	// the requested ledger
 	ran := BoundedRange(from, to)
+	c.ledgerSequenceLock.Lock()
 	c.prepared = &ran
 	c.nextLedger = c.roundDownToFirstReplayAfterCheckpointStart(from)
 	c.lastLedger = &to
+	c.ledgerSequenceLock.Unlock()
+
 	c.previousLedgerHash = nil
 
 	return nil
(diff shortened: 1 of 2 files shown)


urvisavla:
> looks good but can you fix the lint errors in ingest/ledgerbackend/captive_core_backend_test.go ?

Fixed linter errors.

ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -307,9 +313,12 @@ func (c *CaptiveStellarCore) openOfflineReplaySubprocess(from, to uint32) error
 	// The next ledger should be the first ledger of the checkpoint containing
 	// the requested ledger
 	ran := BoundedRange(from, to)
+	c.ledgerSequenceLock.Lock()
 	c.prepared = &ran
 	c.nextLedger = c.roundDownToFirstReplayAfterCheckpointStart(from)
 	c.lastLedger = &to
+	c.ledgerSequenceLock.Unlock()
```
sreuland, line 320: I think should use go's `defer`  to run the Unlock, to follow the idiomatic go approach to achieve 'finally' semantics and ensure lock is not left inconsistent/deadlocked.

ingest/ledgerbackend/captive_core_backend_test.go
```diff
@@ -645,6 +646,71 @@ func TestGetLatestLedgerSequence(t *testing.T) {
...
+			case <-ctx.Done():
+				return
+			default:
+				_, _ = captiveBackend.GetLatestLedgerSequence(ctx)
+			}
+		}
+	}(ctx)
+
+	for i := fromSeq; i < toSeq; i++ {
+		_, err = captiveBackend.GetLedger(ctx, i)
```
sreuland, line 705: this isn't deterministic in how it replicates the race correct, i.e. there's not a way to ensure the threads access lock at same time, but with this test, go `-race` shouldn't detect anything at execution time? If you comment out the new mutex, and then run the test with `-race`, does it detect?

urvisavla: Yes, you're correct that it's not possible to replicate two threads accessing simultaneously. But, -race is able to detect unsynchronized reads and writes. If I remove the lock from the code and run the test with the -race flag enabled, the test fails.

ingest/ledgerbackend/captive_core_backend.go
urvisavla: updated.

ingest/ledgerbackend/captive_core_backend_test.go
sreuland: nice, then it's effective, thanks!

sreuland (approved):
looks good, thanks for responding to suggestion.

PR #4926, opened 2023-06-21
all: Fix improper use of errors.Wrap

bartekn:
`errors.Wrap` method returns nil if the first argument passed is also nil. If `errors.Wrap` is copied from a condition like `if err != nil` to another one which  also returns `errors.Wrap` but does not overwrite `err` before the returned value will always be `nil`.


services/horizon/internal/db2/history/claimable_balances.go
```diff
@@ -47,7 +47,7 @@ func (cbq ClaimableBalancesQuery) Cursor() (int64, string, error) {
 		}
 		r = parts[1]
 		if l < 0 {
-			return l, r, errors.Wrap(err, "Invalid cursor - first value should be higher than 0")
+			return l, r, errors.New("Invalid cursor - first value should be higher than 0")
```
Shaptic, line 50: just want to piggy-back on this PR to fix a Golang convention: error messages should start with lowercase letters (link: [golang/go](https://github.com/golang/go/wiki/CodeReviewComments#error-strings)).

```suggestion
			return l, r, errors.New("invalid cursor - first value should be higher than 0")
```

Shaptic (approved):
:wave: hi bartek! do you miss us? :smile: thank you for this PR! it's a subtle thing to catch - did you use a tool for this?

bartekn:
> it's a subtle thing to catch - did you use a tool for this?

Yes, I've been playing with CodeQL recently and this was one of the first queries I worked on 😉 . I've been looking for something like this since: https://github.com/dominikh/go-tools/issues/529.

PR #4941, opened 2023-06-30
LedgerChangeReader: Support State Expiration & Eviction

paulbellamy:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add LedgerCloseMetaV2 support to LedgerChangeReader.

### Why

So it emits correct changes for state expiration. Particularly:
- Entry extensions
- Entry evictions

This is to support: https://github.com/stellar/soroban-tools/pull/695

### Known limitations

[TODO or N/A]


ingest/ledger_change_reader.go
```diff
@@ -122,6 +124,14 @@ func (r *LedgerChangeReader) Read() (Change, error) {
 			r.pending = append(r.pending, metaChanges...)
 		}
 		return r.Read()
+	case evictionChangesState:
+		// Get contract ledgerEntry evictions
+		changes := GetChangesFromLedgerEntryEvictions(
+			r.ledgerCloseMeta.EvictedLedgerKeys(),
```
tsachiherman, line 130: could you add an error case to `EvictedLedgerKeys` instead of ignoring the error coming back from `SetContractCode`/`SetContractData` ?

ingest/ledger_change_reader_test.go
```diff
@@ -77,29 +78,55 @@ func buildChange(account string, balance int64) xdr.LedgerEntryChange {
...
 
-type balanceEntry struct {
-	address string
-	balance int64
+type changePredicate func(*testing.T, int, Change)
+
+func isBalance(address string, balance int64) changePredicate {
+	return func(t *testing.T, idx int, change Change) {
+		msg := fmt.Sprintf("change %d", idx)
+		assert.NotNil(t, change.Post, msg)
```
tsachiherman, line 86: change
```golang
assert.NotNil(t, change.Post, msg)
```
to
```golang
require.NotNil(t, change.Post, msg)
```
since assert is going to continue with the execution, and the subsequent line would generate a segmentation fault.

ingest/ledger_change_reader_test.go
```diff
@@ -108,11 +135,15 @@ func assertChangesEqual(
...
 		}
-
-		changes = append(changes, parseChange(change))
+		changes = append(changes, change)
 	}
 
-	assert.Equal(t, expected, changes)
+	assert.Len(t, changes, len(expectations), "unexpected number of changes")
+
+	// Check each change is what we expect
```
tsachiherman, line 143: are the changes are always guaranteed to be in order ?

ingest/ledger_change_reader_test.go
```diff
@@ -318,6 +349,337 @@ func TestLedgerChangeReaderOrder(t *testing.T) {
...
+		},
+	}
+	secondTxHash, err := network.HashTransactionInEnvelope(secondTx, network.TestNetworkPassphrase)
+	assert.NoError(t, err)
+
+	baseFee := xdr.Int64(100)
+	tempKey := xdr.ScSymbol("TEMPKEY")
+	persistentKey := xdr.ScSymbol("TEMPVAL")
+	persistentVal := xdr.ScSymbol("PERSVAL")
+	contractIdBytes, err := hex.DecodeString("df06d62447fd25da07c0135eed7557e5a5497ee7d15b7fe345bd47e191d8f577")
```
tsachiherman, line 391: ST1003: var contractIdBytes should be contractIDBytes (stylecheck)

ingest/ledger_change_reader_test.go
```diff
@@ -318,6 +349,337 @@ func TestLedgerChangeReaderOrder(t *testing.T) {
...
+	assert.NoError(t, err)
+	var contractId xdr.Hash
```
tsachiherman, line 393: ST1003: var contractId should be contractID (stylecheck)

xdr/ledger_close_meta.go
```diff
@@ -99,6 +93,9 @@ func (l LedgerCloseMeta) TransactionResultPair(i int) TransactionResultPair {
 	case 1:
 		return l.MustV1().TxProcessing[i].Result
 	case 2:
+		if l.MustV2().TxProcessing[i].TxApplyProcessing.V != 3 {
```
tsachiherman, line 96: shouldn't this be applied only for the `LedgerCloseMeta.TxApplyProcessing` ?

tsachiherman (changes requested):
Few small requests.

ingest/ledger_change_reader_test.go
Shaptic: we should `s/assert\./require.` in the whole repo, honestly - it rarely makes sense to continue a test after a failed assertion

ingest/change.go
```diff
@@ -68,6 +68,39 @@ func GetChangesFromLedgerEntryChanges(ledgerEntryChanges xdr.LedgerEntryChanges)
...
+// GetChangesFromLedgerEntryEvictions transforms evicted LedgerKeys to []Change.
+// The generated changes always remove the entries.
+func GetChangesFromLedgerEntryEvictions(keys []xdr.LedgerKey) []Change {
+	changes := []Change{}
+
+	for _, key := range keys {
+		state := xdr.LedgerEntry{}
+		switch key.Type {
+		case xdr.LedgerEntryTypeContractData:
+			state.Data.SetContractData(&xdr.ContractDataEntry{
```
Shaptic, line 80: `golangci` says check your error and I think that's valid: this could help us catch places to update XDR in the future?

ingest/change.go
```diff
@@ -68,6 +68,39 @@ func GetChangesFromLedgerEntryChanges(ledgerEntryChanges xdr.LedgerEntryChanges)
...
+				Durability: key.ContractData.Durability,
+			})
+		case xdr.LedgerEntryTypeContractCode:
+			state.Data.SetContractCode(&xdr.ContractCodeEntry{
+				Hash: key.ContractCode.Hash,
+			})
+		default:
+			// Currently only contractData and contractCode are evicted by core, so
+			// we only need to handle those two.
+			panic("Invalid LedgerEntry eviction type")
```
Shaptic, line 92:
```suggestion
			panic(fmt.Errorf("Invalid LedgerEntry eviction type: %v", key.Type))
```

friendlier :) (might need an `import fmt`)

ingest/ledger_change_reader.go
```diff
@@ -122,6 +124,14 @@ func (r *LedgerChangeReader) Read() (Change, error) {
...
+		)
+		r.pending = append(r.pending, changes...)
+		r.state++
```
Shaptic, line 133: TIHI - should we just set the enum appropriately? :laughing:

ingest/ledger_change_reader_test.go
```diff
@@ -77,29 +78,55 @@ func buildChange(account string, balance int64) xdr.LedgerEntryChange {
...
+	return func(t *testing.T, idx int, change Change) {
+		msg := fmt.Sprintf("change %d", idx)
+		assert.NotNil(t, change.Post, msg)
+		assert.NotNil(t, change.Pre, msg)
+		assert.Equal(t, xdr.LedgerEntryTypeContractData.String(), change.Post.Data.Type.String(), msg)
+		assert.Equal(t, contract, change.Post.Data.ContractData.Contract, msg)
+		assert.Equal(t, key, change.Post.Data.ContractData.Key, msg)
+		newExpiry := change.Post.Data.ContractData.ExpirationLedgerSeq
+		oldExpiry := change.Pre.Data.ContractData.ExpirationLedgerSeq
+		assert.Equal(t, extension, uint32(newExpiry-oldExpiry), msg)
```
Shaptic, line 103: protip: `assert.EqualValues` lets you ignore type differences :sunglasses:

ingest/ledger_change_reader_test.go
```diff
@@ -108,11 +135,15 @@ func assertChangesEqual(
...
-		changes = append(changes, parseChange(change))
+		changes = append(changes, change)
 	}
 
-	assert.Equal(t, expected, changes)
+	assert.Len(t, changes, len(expectations), "unexpected number of changes")
+
+	// Check each change is what we expect
+	for i, change := range changes {
+		expectations[i](t, i+1, change)
```
Shaptic, line 145: clever!

xdr/ledger_close_meta.go
```diff
@@ -146,3 +143,23 @@ func (l LedgerCloseMeta) UpgradesProcessing() []UpgradeEntryMeta {
...
 	}
 }
+
+// EvictedLedgerKeys returns the LedgerKeys for both the
+// EvictedTemporaryLedgerKeys and and the EvictedPersistentLedgerEntries in a
+// ledger.
+func (l LedgerCloseMeta) EvictedLedgerKeys() []LedgerKey {
+	switch l.V {
+	case 0, 1:
+		return nil
```
Shaptic, line 153: safe to do this over an empty slice?

Shaptic (commented):
drive-by review :racing_car:

xdr/ledger_close_meta.go
tsachiherman: Hmm. The answer is "it depends".
If it's just range iterations and appends, than it would work just fine.

However, if we were to have pointer receiver functions implemented on that slice, then it might be problematic.

ingest/ledger_change_reader_test.go
paulbellamy: Yes. There isn't really a reason for them to be in any specific order, but the order is consistent due to the internal ledgerChangeReader implementation

ingest/ledger_change_reader.go
paulbellamy: Probably yes, but this code generally needs a cleanup/refactor tbh.

xdr/ledger_close_meta.go
paulbellamy: Pointer methods on nil arrays work fine, afaik: https://go.dev/play/p/TOly3j6kQqy

Where it gets tricky is `encoding/json` treats them differently. So if clients have specific expectations about json structures it might be wrong.

tsachiherman (approved):

PR #4949, opened 2023-07-10
services/horizon: Add optional configuration parameter called NETWORK

### LINKED ISSUE #4913: services/horizon: Add a new optional config parameter called "NETWORK"
As part of our effort to streamline the Horizon configuration process and provide a more user friendly experience when setting up Horizon, we want to add a new optional config parameter called `NETWORK`. The new parameter will allow specifying the desired network for Horizon with two available values: "pubnet" or "testnet." Horizon will automatically adjust configuration settings and generate the core toml/cfg based on the network chosen. 

Introduce a new optional configuration parameter called `NETWORK` to specify the desired Stellar network in Horizon. Available values are `pubnet` and `testnet`. This parameter can be set as an environment variable similar to other environment variables in Horizon. When the `NETWORK` parameter is set, Horizon automatically adjusts the remaining configuration settings and generates the corresponding core toml/cfg settings.

- If the `NETWORK` parameter is specified, the following fields are automatically populated based on the chosen network:
  - `HISTORY_ARCHIVE_URLS`
  - `NETWORK_PASSPHRASE`
  - `CAPTIVE_CORE_CONFIG_PATH`: The default configuration file for `pubnet` and `testnet` is embedded in the Horizon binary and used to generate the default configuration.

- By default, `CAPTIVE_CORE_STORAGE_PATH` is set to the current directory where Horizon is running.

- By default, `STELLAR_CORE_URL` is set to `http://localhost:<CAPTIVE_CORE_HTTP_PORT>`.

- If no `NETWORK` parameter is specified, the following fields become mandatory and must be explicitly set:
  - `HISTORY_ARCHIVE_URLS`
  - `NETWORK_PASSPHRASE`

-  if the NETWORK parameter is supplied and HISTORY_ARCHIVE_URLS or NETWORK_PASSPHRASE is supplied, an error message will be displayed.

urvisavla:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

The PR introduces a new optional configuration parameter called `NETWORK` in Horizon. This parameter allows users to specify the desired Stellar network, `pubnet` or `testnet`. When the `NETWORK` parameter is set, Horizon automatically adjusts the remaining configuration settings and generates the corresponding captive core config file.

- If the `NETWORK` parameter is specified, the following fields are automatically populated based on the chosen network:
  - `HISTORY_ARCHIVE_URLS`
  - `NETWORK_PASSPHRASE`
  - `CAPTIVE_CORE_CONFIG_PATH`: The default configuration file for `pubnet` and `testnet` is embedded in the Horizon binary and used to generate the default configuration.

- By default, `CAPTIVE_CORE_STORAGE_PATH` is set to the current directory where Horizon is running.

- By default, `STELLAR_CORE_URL` is set to `http://localhost:<CAPTIVE_CORE_HTTP_PORT>`

- _`CAPTIVE_CORE_STORAGE_PATH` and `STELLAR_CORE_URL` fields are configured by default if not explicitly specified, regardless of whether the `NETWORK` parameter is specified or not._

- If no `NETWORK` parameter is specified, the following fields become mandatory and an error message is displayed if not supplied:
  - `HISTORY_ARCHIVE_URLS`
  - `NETWORK_PASSPHRASE`

-  if the `NETWORK` parameter is specified and `HISTORY_ARCHIVE_URLS` or `NETWORK_PASSPHRASE` is supplied, an error message is displayed.

### Why

Simplified Configuration: By specifying the `NETWORK` parameter, users no longer need to separately set the `HISTORY_ARCHIVE_URLS`, `NETWORK_PASSPHRASE`, and `CAPTIVE_CORE_CONFIG_PATH`. These fields are automatically populated based on the chosen network, streamlining the configuration process.

closes #4913 

### Known limitations
[N/A]


services/horizon/internal/flags_test.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			e := createCaptiveCoreDefaultConfig(&tt.config)
+			if tt.errStr == "" {
+				assert.NoError(t, e)
+				assert.Equal(t, tt.networkPassphrase, tt.config.NetworkPassphrase)
+				assert.Equal(t, tt.historyArchiveURLs, tt.config.HistoryArchiveURLs)
+			} else {
+				assert.Equal(t, tt.errStr, e.Error())
```
Shaptic, line 79: could prob do a `require.Error()` first before accessing `e.Error()`

Shaptic (commented):
looks really good! much easier to follow the code, too :clap: just left some minor comments below

services/horizon/internal/flags.go
```diff
@@ -589,6 +616,162 @@ type ApplyOptions struct {
...
+func loadDefaultCaptiveCoreToml(config *Config, defaultConfigData []byte) error {
+
+	config.CaptiveCoreTomlParams.CoreBinaryPath = config.CaptiveCoreBinaryPath
+	config.CaptiveCoreTomlParams.HistoryArchiveURLs = config.HistoryArchiveURLs
+	config.CaptiveCoreTomlParams.NetworkPassphrase = config.NetworkPassphrase
+
+	var err error
+	config.CaptiveCoreToml, err = ledgerbackend.NewCaptiveCoreTomlFromData(defaultConfigData, config.CaptiveCoreTomlParams)
+	if err != nil {
+		return fmt.Errorf("invalid captive core toml: %v", err)
```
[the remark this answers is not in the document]
urvisavla: good catch. `errors.Wrap` looks cleaner to me. Fixed.

services/horizon/internal/flags_test.go
urvisavla: updated.

services/horizon/internal/config.go
```diff
@@ -114,4 +114,6 @@ type Config struct {
 	BehindAWSLoadBalancer bool
 	// RoundingSlippageFilter excludes trades from /trade_aggregations with rounding slippage >x bps
 	RoundingSlippageFilter int
+	// Stellar network: 'testnet' or 'pubnet'
```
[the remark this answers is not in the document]
urvisavla: Good point. What do you think @sreuland?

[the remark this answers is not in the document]
Shaptic: we also use it in the integration tests so we could leverage it there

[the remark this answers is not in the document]
sreuland: sounds good, that's also similar to how quickstart offers the same 3 networks.

services/horizon/internal/flags.go
```diff
@@ -42,6 +43,12 @@ const (
 	CaptiveCoreConfigUseDB = "captive-core-use-db"
 
 	captiveCoreMigrationHint = "If you are migrating from Horizon 1.x.y, start with the Migration Guide here: https://developers.stellar.org/docs/run-api-server/migrating/"
+
+	StellarPubnet              = "pubnet"
```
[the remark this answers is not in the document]
urvisavla: fixed.

services/horizon/internal/flags.go
```diff
@@ -589,6 +616,162 @@ type ApplyOptions struct {
...
+var (
+	//go:embed configs/captive-core-pubnet.cfg
+	PubnetDefaultConfig []byte
+
+	//go:embed configs/captive-core-testnet.cfg
+	TestnetDefaultConfig []byte
+
+	pubnetConf = networkConfig{
+		defaultConfig:      PubnetDefaultConfig,
+		historyArchiveURLs: []string{"https://history.stellar.org/prd/core-live/core_live_001/"},
```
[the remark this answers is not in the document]
urvisavla: Added multiple urls.

services/horizon/docker/docker-compose.standalone.yml
```diff
@@ -34,13 +34,8 @@ services:
 
   horizon:
     environment:
-      - HISTORY_ARCHIVE_URLS=http://host.docker.internal:1570
-      - NETWORK_PASSPHRASE=Standalone Network ; February 2017
-      - CAPTIVE_CORE_CONFIG_APPEND_PATH=/captive-core-standalone.cfg
+      - NETWORK=standalone
```
sreuland, line 37: nice, making the integration tests use the new `NETWORK` parameter, do you think we might need to also retain the integration tests based on individual network configurations also to assert those flags are not regressing in behavior?

services/horizon/internal/configs/captive-core-standalone.cfg
```diff
@@ -1,12 +1,17 @@
...
+
 [[VALIDATORS]]
 NAME="local_core"
 HOME_DOMAIN="core.local"
 # From "SACJC372QBSSKJYTV5A7LWT4NXWHTQO6GHG4QDAVC2XDPX6CNNXFZ4JK"
 PUBLIC_KEY="GD5KD2KEZJIGTC63IGW6UMUSMVUVG5IHG64HUTFWCHVZH2N2IBOQN7PS"
-ADDRESS="host.docker.internal"
-QUALITY="MEDIUM"
\ No newline at end of file
+ADDRESS="localhost:11625"
```
sreuland, line 15: hmm, this might be interesting to see how it affects things at run times, not sure if integration tests use this cfg? but in general, `localhost` only refers to internal container network loopback interface, if the core instance it's trying to connect to is located in another docker container or any other process on host machine, then it would need to use `host.docker.internal` to reference that network path to host. Also, based on team discussion earlier, do you think this ADDRESS could potentially be dynamically provided as another env parameter when NETWORK=STANDALONE? that way people can specify a remote core instance that they are running on standalone, and it may not be at this default of their host machine, ie. localhost:11625

urvisavla: Good catch!  I agree with you that it's likely not being used because I recently updated the `STELLAR_CORE_URL` in the [docker-compose-standalone.yml ](https://github.com/stellar/go/pull/4956) to run Horizon standalone in Docker, but I totally missed the default config. So as you correctly pointed out, it seems that this won't work for Docker unless we parameterize the ADDRESS and HISTORY_ARCHIVE_URLS as well to make it compatible with running Horizon from Docker.

The only valid use case for adding the default for `standalone` configuration is if someone wants to start the Horizon binary directly (without Docker) in standalone mode. And in that scenario, we expect it to work with the defaults only if the core and history archives are running locally. This behavior differs from the Quickstart where the core runs in the same container. So having the `standalone` `NETWORK` option in Horizon might confuse users who are used to using the Quickstart with standalone because of the additional steps.

I'm wondering whether it's alright to delay adding the standalone option until we have given it more thought. For now we can proceed with the default configuration for `pubnet `and `testnet`. @sreuland @Shaptic @mollykarcher What do you think?

services/horizon/docker/docker-compose.standalone.yml
urvisavla: That's a great point, and I think it makes sense to revert all ( testnet, pubnet, and standalone) docker-compose changes for now to make sure individual configurations work and that we have no regression.  Also it seems like the tests are passing with the new `NETWORK` parameter, so I think it's safe to assume that it's working as expected(?).

services/horizon/internal/configs/captive-core-standalone.cfg
Shaptic: Great points all around. I'll just echo what I said in today's meeting heh: let's not let `standalone` hold back this PR as it's already a big win!

sreuland: yes, agreed, thanks for vetting that further!

urvisavla: 🙏  Reverted standalone for now.

services/horizon/internal/flags.go
```diff
@@ -589,6 +616,162 @@ type ApplyOptions struct {
...
+	var defaultNetworkConfig networkConfig
+	switch config.Network {
+	case StellarPubnet:
+		defaultNetworkConfig = pubnetConf
+	case StellarTestnet:
+		defaultNetworkConfig = testnetConf
+	default:
+		return fmt.Errorf("no default configuration found for network %s", config.Network)
+	}
+	config.NetworkPassphrase = defaultNetworkConfig.networkPassphrase
```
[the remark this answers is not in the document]
urvisavla: Since our intention is to discourage users from specifying individual network settings and promote the use of default configurations, I think it is OK to throw an error in this situation. wdyt?

services/horizon/internal/flags.go
```diff
@@ -589,6 +624,166 @@ type ApplyOptions struct {
...
+var (
+	//go:embed configs/captive-core-pubnet.cfg
+	PubnetDefaultConfig []byte
+
+	//go:embed configs/captive-core-testnet.cfg
+	TestnetDefaultConfig []byte
+
+	pubnetConf = networkConfig{
+		defaultConfig: PubnetDefaultConfig,
+		historyArchiveURLs: []string{"https://history.stellar.org/prd/core-live/core_live_001/",
```
sreuland, line 642: for consideration, rather than hardcoding the archive urls, network passwords in-line, what if create a property file to hold these non core.cfg network settings such as `configs/pubnet-params.toml` with:
```
HISTORY_ARCHIVE_URLS=..
```

and might be able to get Network passphrase by opening the existing `captive-core-testnet|pubnet.cfg` files for that key?

services/horizon/internal/flags.go
```diff
@@ -589,6 +624,166 @@ type ApplyOptions struct {
...
+	config.CaptiveCoreToml, err = ledgerbackend.NewCaptiveCoreTomlFromFile(config.CaptiveCoreConfigPath, config.CaptiveCoreTomlParams)
+	if err != nil {
+		return errors.Wrap(err, "invalid captive core toml file")
+	}
+	return nil
+}
+
+// createCaptiveCoreDefaultConfig generates the default Captive Core configuration.
+// validates the configuration settings, sets default values, and loads the Captive Core TOML file.
+func createCaptiveCoreDefaultConfig(config *Config) error {
```
sreuland, line 699: this is named so close to the other method, maybe a rename to differentiate them more, `createCaptiveCoreConfigBasedOnNetwork` and change `createCaptiveCoreConfig` to `createCaptiveCoreConfigBasedOnFlags`

services/horizon/docker/docker-compose.standalone.yml
sreuland: i looked into the integration tests a bit, I think the way this is setup at present, no integration tests will run with the new `network` flag. but it would be good to establish a few integration tests that exercise this from external command line, fortunately, there is an existing integration test, [parameters_test.go](https://github.com/urvisavla/go/blob/4913/network-config-param/services/horizon/internal/integration/parameters_test.go), which does just that, running horizon from cmd line with different flags to verify, so I think would be worthwhile to add a test method in there that validates horizon startup behavior for the different combos of `network`,`network-passphrase`,`history-archive-urls`. for reference, check out how `NewTest()` sets up the horizon flags right before it starts horizon - https://github.com/urvisavla/go/blob/4913/network-config-param/services/horizon/internal/test/integration/integration.go#L341-L344, as that is how to change those cmd flags from test initiation.

sreuland (approved):
looks good, I left a question on establishing some new integration test coverage for `network` and some minor refactor for network constants and method names for consideration, thanks!

services/horizon/internal/flags.go
urvisavla: updated.

services/horizon/internal/flags.go
urvisavla: updated.

services/horizon/docker/docker-compose.standalone.yml
urvisavla: Currently, integration tests, both the code and the docker-compose-integration.yaml file are designed to test Horizon and Core in standalone mode. I spent some time modifying the integration tests to also support `pubnet` and `testnet`, but some more plumbing and testing in integration test framework will be required. I can create a separate issue for this.

urvisavla: Created  https://github.com/stellar/go/issues/4967 for adding integration tests.

urvisavla:
> looks good, I left a question on establishing some new integration test coverage for `network` and some minor refactor for network constants and method names for consideration, thanks!

I've updated network constants and method names as suggested and created  an issue for new integration tests https://github.com/stellar/go/issues/4967

Dezzj (commented):
Veiwed

services/horizon/internal/flags.go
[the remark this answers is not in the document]
hethm999: good

network/main.go
```diff
@@ -19,6 +19,18 @@ const (
 	TestNetworkPassphrase = "Test SDF Network ; September 2015"
 )
 
+var (
+	// PublicNetworkhistoryArchiveURLs is a list of history archive URLs for stellar 'pubnet'
+	PublicNetworkhistoryArchiveURLs = []string{"https://history.stellar.org/prd/core-live/core_live_001/",
+		"https://history.stellar.org/prd/core-live/core_live_002/",
+		"https://history.stellar.org/prd/core-live/core_live_003/"}
```
jacekn, line 26: I know this is very late in the process but this adds hard dependency on the SDF history archives.
To help with decentralization I think we should avoid doing this and instead spread traffic across multiple archives.
Another side effect of this is that this will put more stress on SDF history archives.

sreuland: this is the backstop/default to land on, it doesn't force/stop at this, as things progress towards decentralization and there are other history archive url options, users can still provide the `HISTORY_ARCHIVE_URLS` setting to override and reference other archive servers, as it's complementary with NETWORK, does this usage help with that concern?

jacekn: My concern is that the default value for pubnet is, IMO, not what we should recommend or encourage. For sure users who dig a bit deeper and understand decentralization effort can override defaults that SDF provides. But in reality many won't and I don't think we can rely on users knowing this setting should be changed.

Generally speaking default values should be good for most use cases and should follow best practices. If we want to encourage network decentralization then best practice would be to avoid using single entity for history archives. To follow that best practice we could make the setting use wider selection or archives or, if you expect users to override the setting, make the setting mandatory so that it's a conscious choice.

sreuland: @jacekn , I've created ticket #5142 to track decentralization requirements on archiver server url config further, thanks for calling out!

PR #4955, opened 2023-07-11
Add functionality to hide ingestion enabled flag from --help output

### LINKED ISSUE #4906: ingest/filtering: enable filtering by default
un-necessary configuration flag to enable ingestion filtering rather than default

ingestion filtering enabled by default, don't need to configure the runtime flag, remove `--exp-enable-ingestion-filtering` flag from the help output, but keep the flag definition. Log a warning message saying the flag is deprecated if it's present. find any references to the flag in the repo and remove them or document as deprecated.

There will be no filter rules created by default which results in effectively same end result, nothing filtered. Need to confirm this is the outcome when filtering enabled with no rules.

remove reference of the flag in the[ public ingestion docs ](https://developers.stellar.org/docs/run-api-server/ingestion-filtering)

set the `--exp-enable-ingestion-filtering` in horizon runtime config.

aditya1702:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Closes #4906

- Add a new `Hidden` config option which makes it easy to hide flags from our --help output.
- Hide the flag: `--exp-enable-ingestion-filtering` since it is now enabled by default.
- Give a DEPRECATION warning if a user still specifies the flag.
- Integration tests to test the change.

### Why

Ingestion filtering is now always enabled and does not require user specified flag.

### Known limitations

[TODO or N/A]


support/config/config_option.go
```diff
@@ -28,6 +33,13 @@ func (cos ConfigOptions) Init(cmd *cobra.Command) error {
 			return err
 		}
 	}
+
+	// Hide the ingestion filtering flag from the --help output.
+	if cmd.Use == HorizonCmdUsageString {
```
urvisavla, line 38: config package doesn't seem like the right place to handle specific horizon flags. Instead, it should be managed within the horizon package itself. Consider adding a new function like `Hide` or `HideFlag` in the config package. Use this function from within the horizon package (possibly in the flags.go ?).

services/horizon/internal/flags.go
```diff
@@ -608,6 +618,9 @@ func ApplyFlags(config *Config, flags support.ConfigOptions, options ApplyOption
 		config.Ingest = true
 	}
 
+	// Ingestion filtering enabled by default
+	config.EnableIngestionFiltering = true
```
urvisavla, line 622: Is this necessary? Considering the flag default is already updated to true on line 200 .

services/horizon/internal/integration/ingestion_filtering_test.go
```diff
@@ -165,3 +182,67 @@ func TestFilteringAssetWhiteList(t *testing.T) {
...
+	accountFilter, err := itest.AdminClient().GetIngestionAccountFilter()
+	tt.NoError(err)
+
+	tt.ElementsMatch(expectedAccountFilter.Whitelist, accountFilter.Whitelist)
+	tt.Equal(expectedAccountFilter.Enabled, accountFilter.Enabled)
+
+	// Ensure the latest filter configs are reloaded by the ingestion state machine processor
+	time.Sleep(time.Duration(filters.GetFilterConfigCheckIntervalSeconds()) * time.Second)
+
+	// Make sure that when using a non-whitelisted account, the transaction is not stored
```
urvisavla, line 227: If ingestion filtering is disabled (line 192), the account should not be filtered out, right?

services/horizon/internal/flags.go
aditya1702: I only added this since we are still keeping the flag. And if someone passes the "--exp-enable-ingestion-filtering false", this should take care of it. What do you think?

services/horizon/internal/integration/ingestion_filtering_test.go
aditya1702: @urvisavla But the idea is that ingestion-filtering will always be allowed. So even if someone passes the flag as false, it should do the filtering.

If someone wants ingestion filtering to stop, one should remove the filters themselves and not set the flag to false (since it is deprecated and going to be removed eventually).

@sreuland What do you think about this? Is my understanding correct?

support/config/config_option.go
aditya1702: Yeah, I agree with that.

services/horizon/internal/integration/ingestion_filtering_test.go
urvisavla: Ok, so as long as there's at least one enabled rule, the filtering will kick in. If that's the case, it makes sense. Thanks for clarifying!

services/horizon/internal/flags.go
```diff
@@ -194,12 +195,21 @@ func Flags() (*Config, support.ConfigOptions) {
...
-			Name:        "exp-enable-ingestion-filtering",
+			Name:        EnableIngestionFilteringFlag,
 			OptType:     types.Bool,
-			FlagDefault: false,
+			FlagDefault: true,
 			Required:    false,
-			Usage:       "causes Horizon to enable the experimental Ingestion Filtering and the ingestion admin HTTP endpoint at /ingestion/filter",
 			ConfigKey:   &config.EnableIngestionFiltering,
+			CustomSetValue: func(opt *support.ConfigOption) error {
+				if val := viper.GetString(opt.Name); val != "" {
```
urvisavla, line 204: Should this be `viper.GetBool(opt.Name);` since it's of type Bool? Also, I'm not sure if we can determine whether the flag was explicitly set here because by default it's set to true.

urvisavla (commented):
Overall it looks good to me. Would be good to get another set of 👀 to review it as well. Also, there are a few linter warnings, it'll be great if you could address those as well. Thanks!

urvisavla (approved):
Please add a changelog entry to reflect this change. Also, we need to update the public documentation for [ingestion filtering ](https://developers.stellar.org/docs/run-api-server/ingestion-filtering#configuration) with these modifications.

services/horizon/cmd/root.go
```diff
@@ -48,6 +52,12 @@ func init() {
 	if err != nil {
 		stdLog.Fatal(err.Error())
 	}
+
+	// Hide the ingestion filtering flag from --help output.
+	err = flags.HideFlag(RootCmd, EnableIngestionFilteringFlag)
```
sreuland, line 57: why not do hiding internally in `horizon.Flags()`, and keep this type of lower level knowledge out of this upper layer, and skips needing to redeclare `EnableIngestionFilteringFlag`.

services/horizon/internal/flags.go
```diff
@@ -194,12 +195,23 @@ func Flags() (*Config, support.ConfigOptions) {
...
+					stdLog.Printf(
+						"DEPRECATED - %s causes Horizon to enable the experimental Ingestion Filtering and the "+
+							"ingestion admin HTTP endpoint at /ingestion/filter. Ingestion filtering is now enabled by "+
+							"default and this flag will soon be removed. If you do not want any filtering, you need to "+
```
sreuland, line 208: `No ingestion filter rules are defined by default, which equates to no filtering of historical data. If you have never added filter rules to this deployment, then nothing further needed. If you have defined ingestion filter rules prior but disabled filtering overall by setting this flag disabled with --exp-enable-ingestion-filtering=false, then you should now delete the filter rules using the Horizon Admin API to achieve the same no-filtering result. Remove usage of this flag in all cases.`

services/horizon/internal/flags.go
```diff
@@ -194,12 +195,23 @@ func Flags() (*Config, support.ConfigOptions) {
...
+							"remove any filters you might have set, instead of passing the flag.",
+						EnableIngestionFilteringFlag,
+					)
+				}
+				return nil
```
sreuland, line 213: should you return an error here, so horizon does not start up, and the user will then be informed to the need to remove the flag usage?

services/horizon/internal/flags.go
```diff
@@ -608,6 +620,9 @@ func ApplyFlags(config *Config, flags support.ConfigOptions, options ApplyOption
 		config.Ingest = true
 	}
 
+	// Ingestion filtering enabled by default
+	config.EnableIngestionFiltering = true
```
sreuland, line 624: I think you can remove this line, it should be set to true all the time now after flags.SetValues() is run due to the default value and the CustomSetValue handler defined above.

services/horizon/internal/integration/ingestion_filtering_test.go
```diff
@@ -165,3 +182,67 @@ func TestFilteringAssetWhiteList(t *testing.T) {
 	_, err = itest.Client().TransactionDetail(txResp.Hash)
 	tt.NoError(err)
 }
+
+func TestFilteringAccountWhiteListWithFalseIngestionFilteringFlag(t *testing.T) {
```
sreuland, line 186: I think can remove this test which is validating exp-enable-ingestion-filtering usage and move it into /services/horizon/internal/integration/parameters_test.go, where you can check the cmd line output from the stdout/stderr for matching expectation

sreuland:
@aditya1702 , you can add `Closes #4906` in the pr description, and then GH will automatically link the issue to this PR and also migrate the status of the issue to closed when the pr merges.

services/horizon/cmd/root.go
aditya1702: I did consider this approach but the Init method needs to be called before making any flag hidden. Let me try to work around this thing

urvisavla: Consider a more generic approach to handle this, as we might want to use it for other flags as well. In my opinion, an approach as the following would be cleaner.

1. Modify the `ConfigOption` struct in config_option.go to include a flag `Hide`:
```
@@ -69,6 +69,7 @@ type ConfigOption struct {
        CustomSetValue func(*ConfigOption) error // Optional function for custom validation/transformation
        ConfigKey      interface{}               // Pointer to the final key in the linked Config struct
        flag           *pflag.Flag               // The persistent flag that the config option is attached to
+      Hide         bool                // Field to indicate whether the flag is hidden
 }

```

2. Modify `ConfigOption::Init()` in config_option.go to hide the flag it based on the value of `Hide`:
```
@@ -79,7 +80,11 @@ func (co *ConfigOption) Init(cmd *cobra.Command) error {
                co.EnvVar = strutils.KebabToConstantCase(co.Name)
        }
        // Initialise and bind the persistent flags
-       return co.setFlag(cmd)
+       err := co.setFlag(cmd)
+       if co.Hide {
+               co.flag.Hidden = true
+       }
+       return err
 }
```

3. Update the `Flags` function in flags.go to set `Hide : true` for the flag you want to hide:
```
@@ -198,6 +198,7 @@ func Flags() (*Config, support.ConfigOptions) {
                        OptType:     types.Bool,
                        FlagDefault: false,
                        Required:    false,
+                       Hide:      true,
                        Usage:       "causes Horizon to enable the experimental Ingestion Filtering and the ingestion admin HTTP endpoint at /ingestion/filter",
                        ConfigKey:   &config.EnableIngestionFiltering,
                },

```

services/horizon/internal/flags.go
```diff
@@ -125,13 +126,15 @@ func Flags() (*Config, support.ConfigOptions) {
 			OptType:   types.String,
 			Required:  true,
 			Usage:     "horizon postgres database to connect with",
+			IsHidden:  false,
```
urvisavla, line 129: It's essential that the flag's default value in `ConfigOptions` is `false` without the need to explicitly set it for each flag like this. Otherwise, it will break in all the other places where the `ConfigOptions` flags are used because by default the all the flags will be hidden.

support/config/config_option.go
```diff
@@ -69,6 +69,7 @@ type ConfigOption struct {
 	CustomSetValue func(*ConfigOption) error // Optional function for custom validation/transformation
 	ConfigKey      interface{}               // Pointer to the final key in the linked Config struct
 	flag           *pflag.Flag               // The persistent flag that the config option is attached to
+	IsHidden       bool                      // A flag which indicates whether to hide the flag from --help output
```
urvisavla, line 72: nit: rename it to `Hide` or `Hidden`?

services/horizon/internal/flags.go
aditya1702: @urvisavla Just checked that by default the value will be False. I will remove the individual options from each flag except the ingestion-filtering one

urvisavla: Great! Thanks for verifying that!

services/horizon/internal/integration/ingestion_filtering_test.go
sreuland: wanted to add more context, the scope of this test is to exercise the horizon flags parsing routine rather than ingestion filtering, and to verify that horizon runtime flags will have `exp-enable-ingestion-filtering=true` regardless of the incoming cmd line and environment params which `parameters_test.go` could be a better unit test for this aspect.

services/horizon/internal/integration/parameters_test.go
```diff
@@ -191,6 +191,21 @@ func TestMaxPathFindingRequests(t *testing.T) {
 	})
 }
 
+//func TestHelpOutputForNoIngestionFilteringFlag(t *testing.T) {
```
sreuland, line 194: @aditya1702 , do you think capturing the sysout from flags code parsing can be caught here in test by spinning up a usage of the cobra commend wrapper, checkout the `StartHorizon()` helper method which creates similar for integration tests that start horizon as a reference example:
https://github.com/aditya1702/go/blob/remove-ingestion-filtering-flag/services/horizon/internal/test/integration/integration.go#L302

and then reroute the sysout with `cmd.SetOutput(writer)` first using a writer to a buffer you create in the test, and then after cmd.Init() or cmd.Execute() the sysout would be generated and in present in the buffer? 

I think in this usage, the test could also be able to check the result from `NewAppFromFlags()`,  App.Config.EnableIngestionFiltering, to assert it's true regardless of cmd/env having it set otherwise, since the flag parsing code overrides that to true in all cases correct?

aditya1702: @sreuland Yes, I am actually trying to figure that out right now. Testing command line arguments. I will make a commit when done with the changes 👍

support/config/config_option.go
```diff
@@ -82,6 +83,13 @@ func (co *ConfigOption) Init(cmd *cobra.Command) error {
 	return co.setFlag(cmd)
 }
 
+// SetHidden Hides the flag from --help output
+func (co *ConfigOption) SetHidden(cmd *cobra.Command) {
```
sreuland, line 87: does meaning of 'hidden' correlate closely with 'deprecated' , if it aligns, would suggest renaming this to SetDeprecated to drive that point more.

services/horizon/cmd/root.go
sreuland: for consideration, use the term  'Deprecated' rather than 'Hide' to point out the purpose of this a bit more, where hiding is a resulting effect of this?

services/horizon/internal/integration/ingestion_filtering_test.go
```diff
@@ -28,33 +27,64 @@ func TestFilteringAccountWhiteList(t *testing.T) {
...
 			Destination: nonWhitelistedAccount.GetAccountID(),
 			Amount:      "10",
 			Asset:       defaultAllowedAsset,
 		},
 	)
-
 	txResp, err := itest.Client().TransactionDetail(txResp.Hash)
 	tt.NoError(err)
 
+	// Make sure that when using a whitelisted account, that transaction is also allowed.
```
sreuland, line 47: I don't think this second test `whitelistedAccount` is necessary, it not getting any more coverage, b/c at this point no rules have been defined for `whitelistedAccount` so it's same as `nonWhitelistedAccount`

services/horizon/internal/integration/parameters_test.go
```diff
@@ -212,6 +216,54 @@ func TestDisablePathFinding(t *testing.T) {
...
+
+func TestHelpOutputForNoIngestionFilteringFlag(t *testing.T) {
+	config, flags := horizon.Flags()
+
+	horizonCmd := &cobra.Command{
+		Use:           "horizon",
+		Short:         "client-facing api server for the Stellar network",
+		SilenceErrors: true,
+		SilenceUsage:  true,
+		Long:          "Client-facing API server for the Stellar network. It acts as the interface between Stellar Core and applications that want to access the Stellar network. It allows you to submit transactions to the network, check the status of accounts, subscribe to event streams and more.",
```
sreuland, line 242: for the test usage, can probably skip defining a Long description, less code.

services/horizon/internal/integration/parameters_test.go
```diff
@@ -212,6 +216,54 @@ func TestDisablePathFinding(t *testing.T) {
...
+	}
+
+	var writer io.Writer = &bytes.Buffer{}
+	horizonCmd.SetOutput(writer)
+
+	horizonCmd.SetArgs([]string{"-h"})
+	if err := flags.Init(horizonCmd); err != nil {
+		fmt.Println(err)
+	}
+	if err := horizonCmd.Execute(); err != nil {
```
sreuland, line 259: can you try to add a second test case here for `horizon --exp-enable-ingestion-filtering false`, and verify the output includes `DEPRECATED - No ingestion ...` from the changes in flags.go?

sreuland (approved):
looks good, left a few minor comments on test coverage, thanks for evaluating the feedback.

services/horizon/internal/integration/ingestion_filtering_test.go
aditya1702: @sreuland Actually I think we should create a new file for command_line_args since we will also be adding tests for the other flags

PR #4963, opened 2023-07-14
horizon: fix EffectRestoreFootprint enum

tsachiherman:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

The EffectRestoreFootprint had the same value as EffectBumpFootprintExpiration, which would generate uncertainties when processing effects.

### Why

Trivial

### Known limitations

n/a

### DIFF
--- a/protocols/horizon/effects/main.go
+++ b/protocols/horizon/effects/main.go
@@ -327,6 +327,14 @@ type ContractDebited struct {
 	Amount   string `json:"amount"`
 }
 
+type BumpFootprintExpiration struct {
+	Base
+}
+
+type RestoreFootprint struct {
+	Base
+}
+
 type AccountThresholdsUpdated struct {
 	Base
 	LowThreshold  int32 `json:"low_threshold"`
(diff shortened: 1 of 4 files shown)


chowbao (approved):

sreuland (approved):
not suggesting here/now, noticed no integration tests against these new ops yet, probably some tech debt we want to address at good point.

tsachiherman:
> not suggesting here/now, noticed no integration tests against these new ops yet, probably some tech debt we want to address at good point.

Do we have integration tests against any of the other effects ?

tsachiherman:
> not suggesting here/now, noticed no integration tests against these new ops yet, probably some tech debt we want to address at good point.

Do we have integration tests against any of the other effects ?

sreuland:
> > not suggesting here/now, noticed no integration tests against these new ops yet, probably some tech debt we want to address at good point.
> 
> Do we have integration tests against any of the other effects ?

looks like some protocol/effects are indirectly verified through internal horizon integration tests which use the go sdk for api request/results, such as this one that refers to cb effects - https://github.com/stellar/go/blob/soroban-xdr-next/services/horizon/internal/integration/sponsorship_test.go#L721-L725, but is more incidental usage, for consideration in future, may be worthwhile to include some unit test coverage here in same protocols package to verify these.

PR #4968, opened 2023-07-18
all: Bump Soroban XDR definitions

2opremio:
integration tests won't be passing until we get a new Core release with the XDR changes.


services/horizon/internal/ingest/processors/operations_processor_test.go
```diff
@@ -100,7 +100,11 @@ func (s *OperationsProcessorTestSuiteLedger) TestOperationTypeInvokeHostFunction
 	sourceAddress := "GAUJETIZVEP2NRYLUESJ3LS66NVCEGMON4UDCBCSBEVPIID773P2W6AY"
 	source := xdr.MustMuxedAddress(sourceAddress)
 
-	contractParamVal1 := xdr.ScSymbol("func1")
+	contractParamVal0 := xdr.ScSymbol("func1")
```
tsachiherman, line 103: I think that swapping the variables names `contractParamVal0` and `contractParamVal1` would make it cleaner, since it would retain the same order used when creating the []xdr.ScVal

tsachiherman (commented):
looks good.

services/horizon/internal/ingest/processors/operations_processor_test.go
2opremio: Yep, you are right. This I changed the order on the details and should had changed the indexes too.

It should be changed as per your suggestion.

tsachiherman: thank you!

2opremio:
I cannot fix the integration tests until we get a new proper soroban-rpc image

2opremio:
Unfortunately, the integration tests are still failing.

Although soroban-rpc has been updated at https://github.com/stellar/soroban-tools/pull/798 and the corresponding soroban-rpc docker image has been added to the PR, `resource_limit_exceeded` errors still seem to be occuring.

This makes me think that the hack at https://github.com/stellar/soroban-tools/pull/798  to compensate for the lack rent fee estimations.

2opremio:
@tsachiherman this is ready to go. Question is whether we want to wait for other P10 changes before merging.

tsachiherman (approved):

PR #4974, opened 2023-07-24
Services/horizon: Integration tests for NETWORK parameter

### LINKED ISSUE #4967: services/horizon: Add integration tests for testing default configuration using NETWORK parameter
The main purpose of this is to make sure we have thorough testing to verify that the "NETWORK" parameter is working correctly.

The current integration tests cover testing Horizon and Core in `standalone` mode. The goal is to enhance the integration tests by adding support for `pubnet` and `testnet` networks so that we can add integration tests for the "NETWORK" configuration parameter. 

- Add integration tests targeting "NETWORK" configuration parameter. See this [PR](https://github.com/stellar/go/pull/4949) for the implementation details of the parameter

urvisavla:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Introduce integration tests to verify that Horizon starts successfully with the default configuration when the NETWORK parameter is set to either "pubnet" or "testnet". The tests cover `Network` parameter using both command-line argument and  environment variable. Added support to skip starting core containers for running these tests.

Updated the `StartHorizon()` function to return the error instead of just exiting. Updated the existing test that was affected by it.

Fixed an issue in an existing test where it failed to clean up the environment variable properly.

### Why
closes #4967 

### Known limitations

Typically during integration tests, we initiate Horizon in standalone mode and simultaneously start the stellar-core container in standalone mode as well. We wait for Horizon to begin ingesting to verify the test's success. However, in the case of "pubnet" or "testnet," waiting for Horizon to start ingesting is not practical. As a result, we run only validate that the Horizon successfully starts the captive-core subprocess using the default configuration.

### DIFF
--- a/services/horizon/internal/integration/parameters_test.go
+++ b/services/horizon/internal/integration/parameters_test.go
@@ -138,148 +138,129 @@ func TestInvalidNetworkParameters(t *testing.T) {
 		t.Skip()
 	}
 
-	errMsg := "history-archive-urls parameter not allowed with the network parameter"
-	t.Run("history archive urls validation", func(t *testing.T) {
-		localParams := integration.MergeMaps(networkParamArgs, map[string]string{
-			horizon.NetworkFlagName:                    horizon.StellarPubnet,
-			horizon.EnableCaptiveCoreIngestionFlagName: "true",
-			horizon.HistoryArchiveURLsFlagName:         "HISTORY_ARCHIVE_URLS",
-		})
-		testConfig := integration.GetTestConfig()
-		testConfig.SkipCoreContainerCreation = true
-		testConfig.HorizonIngestParameters = localParams
-		test := integration.NewTest(t, *testConfig)
-		err := test.StartHorizon()
-		assert.Equal(t, err.Error(), integration.HorizonInitErrStr+
-			": "+CaptiveCoreConfigErrMsg+errMsg)
-		// Adding sleep as a workaround for the race condition in the ingestion system.
-		// https://github.com/stellar/go/issues/5005
-		time.Sleep(2 * time.Second)
-		test.StopHorizon()
-		test.Shutdown()
-	})
+	var captiveCoreConfigErrMsg = integration.HorizonInitErrStr + ": error generating captive " +
+		"core configuration: invalid config: %s parameter not allowed with the %s parameter"
+	testCases := []struct {
+		name         string
+		errMsg       string
+		networkValue string
+		param        string
+	}{
+		{
+			name: "history archive urls validation",
+			errMsg: fmt.Sprintf(captiveCoreConfigErrMsg, horizon.HistoryArchiveURLsFlagName,
+				horizon.NetworkFlagName),
+			networkValue: horizon.StellarPubnet,
+			param:        horizon.HistoryArchiveURLsFlagName,
+		},
+		{
+			name: "network-passphrase validation",
+			errMsg: fmt.Sprintf(captiveCoreConfigErrMsg, horizon.NetworkPassphraseFlagName,
+				horizon.NetworkFlagName),
+			networkValue: horizon.StellarTestnet,
+			param:        horizon.NetworkPassphraseFlagName,
+		},
+	}
 
-	errMsg = "network-passphrase parameter not allowed with the network parameter"
-	t.Run("network-passphrase validation", func(t *testing.T) {
-		localParams := integration.MergeMaps(networkParamArgs, map[string]string{
-			horizon.NetworkFlagName:                    horizon.StellarTestnet,
-			horizon.EnableCaptiveCoreIngestionFlagName: "true",
-			horizon.NetworkPassphraseFlagName:          "NETWORK_PASSPHRASE",
+	for _, testCase := range testCases {
+		t.Run(testCase.name, func(t *testing.T) {
+			localParams := integration.MergeMaps(networkParamArgs, map[string]string{
+				horizon.NetworkFlagName:                    testCase.networkValue,
+				horizon.EnableCaptiveCoreIngestionFlagName: "true",
+				testCase.param:                             "NETWORK_PARAM",
+			})
+			testConfig := integration.GetTestConfig()
+			testConfig.SkipCoreContainerCreation = true
+			testConfig.HorizonIngestParameters = localParams
+			test := integration.NewTest(t, *testConfig)
+			err := test.StartHorizon()
+			// Adding sleep as a workaround for the race condition in the ingestion system.
+			// https://github.com/stellar/go/issues/5005
+			time.Sleep(2 * time.Second)
+			assert.Equal(t, testCase.errMsg, err.Error())
+			test.Shutdown()
 		})
-		testConfig := integration.GetTestConfig()
-		testConfig.SkipCoreContainerCreation = true
-		testConfig.HorizonIngestParameters = localParams
-		test := integration.NewTest(t, *testConfig)
-		err := test.StartHorizon()
-		assert.Equal(t, err.Error(), integration.HorizonInitErrStr+
-			": "+CaptiveCoreConfigErrMsg+errMsg)
-		// Adding sleep as a workaround for the race condition in the ingestion system.
-		// https://github.com/stellar/go/issues/5005
-		time.Sleep(2 * time.Second)
-		test.StopHorizon()
-		test.Shutdown()
-	})
+	}
 }
 
 // TestNetworkParameter Ensure that Horizon successfully starts the captive-core
 // subprocess using the default configuration when --network [testnet|pubnet]
 // commandline parameter.
 //
-// Typically during integration tests, we initiate Horizon in standalone mode and simultaneously start the
-// stellar-core container in standalone mode as well. We wait for Horizon to begin ingesting to verify the test's
-// success. However, in the case of "pubnet" or "testnet," waiting for Horizon to start ingesting is not practical.
-// So we don't start stellar-core containers.
+// In integration tests, we start Horizon and stellar-core containers in standalone mode
+// simultaneously. We usually wait for Horizon to begin ingesting to verify the test's
+// success. However, for "pubnet" or "testnet," we can not wait for Horizon to catch up,
+// so we skip starting stellar-core containers.
 func TestNetworkParameter(t *testing.T) {
 	if !integration.RunWithCaptiveCore {
 		t.Skip()
 	}
 
-	t.Run("network parameter pubnet", func(t *testing.T) {
-		localParams := integration.MergeMaps(networkParamArgs, map[string]string{
-			horizon.NetworkFlagName: horizon.StellarPubnet,
-		})
-		testConfig := integration.GetTestConfig()
-		testConfig.SkipCoreContainerCreation = true
-		testConfig.HorizonIngestParameters = localParams
-		test := integration.NewTest(t, *testConfig)
-		err := test.StartHorizon()
-		// Adding sleep as a workaround for the race condition in the ingestion system.
-		// https://github.com/stellar/go/issues/5005
-		time.Sleep(2 * time.Second)
-		assert.NoError(t, err)
-		test.StopHorizon()
-		test.Shutdown()
-	})
+	testCases := []string{
+		horizon.StellarPubnet,
+		horizon.StellarTestnet,
+	}
 
-	t.Run("network parameter testnet", func(t *testing.T) {
-		localParams := integration.MergeMaps(networkParamArgs, map[string]string{
-			horizon.NetworkFlagName: horizon.StellarTestnet,
+	for _, networkValue := range testCases {
+		t.Run(fmt.Sprintf("NETWORK parameter %s", networkValue), func(t *testing.T) {
+			localParams := integration.MergeMaps(networkParamArgs, map[string]string{
+				horizon.NetworkFlagName: networkValue,
+			})
+			testConfig := integration.GetTestConfig()
+			testConfig.SkipCoreContainerCreation = true
+			testConfig.HorizonIngestParameters = localParams
+			test := integration.NewTest(t, *testConfig)
+			err := test.StartHorizon()
+			// Adding sleep as a workaround for the race condition in the ingestion system.
+			// https://github.com/stellar/go/issues/5005
+			time.Sleep(2 * time.Second)
+			assert.NoError(t, err)
+			test.Shutdown()
 		})
-		testConfig := integration.GetTestConfig()
-		testConfig.SkipCoreContainerCreation = true
-		testConfig.HorizonIngestParameters = localParams
-		test := integration.NewTest(t, *testConfig)
-		err := test.StartHorizon()
-		// Adding sleep as a workaround for the race condition in the ingestion system.
-		// https://github.com/stellar/go/issues/5005
-		time.Sleep(2 * time.Second)
-		assert.NoError(t, err)
-		test.StopHorizon()
-		test.Shutdown()
-	})
+	}
 }
 
 // TestNetworkEnvironmentVariable Ensure that Horizon successfully starts the captive-core
 // subprocess using the default configuration when the NETWORK environment variable is set
 // to either pubnet or testnet.
 //
-// Typically during integration tests, we initiate Horizon in standalone mode and simultaneously start the
-// stellar-core container in standalone mode as well. We wait for Horizon to begin ingesting to verify the test's
-// success. However, in the case of "pubnet" or "testnet," waiting for Horizon to start ingesting is not practical.
-// So we don't start stellar-core containers.
+// In integration tests, we start Horizon and stellar-core containers in standalone mode
+// simultaneously. We usually wait for Horizon to begin ingesting to verify the test's
+// success. However, for "pubnet" or "testnet," we can not wait for Horizon to catch up,
+// so we skip starting stellar-core containers.
 func TestNetworkEnvironmentVariable(t *testing.T) {
 	if !integration.RunWithCaptiveCore {
 		t.Skip()
 	}
+	testCases := []string{
+		horizon.StellarPubnet,
+		horizon.StellarTestnet,
+	}
 
-	value, isSet := os.LookupEnv("NETWORK")
-	defer func() {
-		if isSet {
-			_ = os.Setenv("NETWORK", value)
-		} else {
-			_ = os.Unsetenv("NETWORK")
-		}
-	}()
+	for _, networkValue := range testCases {
+		t.Run(fmt.Sprintf("NETWORK environment variable %s", networkValue), func(t *testing.T) {
+			value, isSet := os.LookupEnv("NETWORK")
+			defer func() {
+				if isSet {
+					_ = os.Setenv("NETWORK", value)
+				} else {
+					_ = os.Unsetenv("NETWORK")
+				}
+			}()
 
-	t.Run("NETWORK environment variable pubnet", func(t *testing.T) {
-		testConfig := integration.GetTestConfig()
-		testConfig.SkipCoreContainerCreation = true
-		testConfig.HorizonIngestParameters = networkParamArgs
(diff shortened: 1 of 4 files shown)


services/horizon/internal/integration/parameters_test.go
```diff
@@ -25,6 +26,16 @@ var defaultCaptiveCoreParameters = map[string]string{
 	horizon.StellarCoreDBURLFlagName:  "",
 }
 
+var networkParamArgs = map[string]string{
+	horizon.EnableCaptiveCoreIngestionFlagName: "true",
```
sreuland, line 30: NETWORK usage means no need to set ENABLE_CAPTIVE_CORE_INGESTION and it's defaulted to true correct? seems like its presence as a parameter should trigger a startup error when NETWORK is used as it's in conflict, should it be removed from here?

services/horizon/internal/integration/parameters_test.go
```diff
@@ -41,6 +52,20 @@ const (
 		QUALITY="MEDIUM"`
 )
 
+func NewParameterTestSkipContainerCreation(t *testing.T, params map[string]string) *integration.Test {
+	return NewParameterTestWithEnvSkipContainerCreation(t, params, map[string]string{})
+}
+
+func NewParameterTestWithEnvSkipContainerCreation(t *testing.T, params, envvars map[string]string) *integration.Test {
```
sreuland, line 59: instead of extending the existing pattern with two more `NewParameterTest..`methods and the duplication that comes with it for adding one new config param, is it worth refactoring, eliminate the `NewParameterTest..` methods, each test can still use a common base config struct to avoid most of dupe, but having 4 methods for initiating the test seems like more to maintain, i.e. simplify to something like:

```
func GetConfig() *integration.Config {
     return &integration.Config {
		ProtocolVersion:         17,
		SkipHorizonStart:        true,
		SkipContainerCreation:   false,
		HorizonIngestParameters: map[string]string{},
		HorizonEnvironment:      map[string]string{},
     }
}

...

func TestSomeParams() {
     config := GetConfig()
     config.HorizonIngestParameters =  ...
     config.HorizonEnvironment = ...
     config.SkipContainerCreation = true
     test := integration.NewTest(t, config)
     err := test.StartHorizon()
     assert.NoError(t, err)
}


```

services/horizon/internal/integration/parameters_test.go
```diff
@@ -120,6 +148,96 @@ func (suite *FatalTestCase) TestEnvironmentPreserved() {
...
+		})
+		test := NewParameterTestSkipContainerCreation(t, localParams)
+		err := test.StartHorizon()
+		time.Sleep(10 * time.Second)
+		assert.NoError(t, err)
+		test.StopHorizon()
+		test.Shutdown()
+	})
+
+	t.Run("network parameter testnet", func(t *testing.T) {
```
sreuland, line 180: is it worth adding another test that submits invalid combo of NETWORK and one of the direct network configs like CAPTIVE_CORE_CONFIG_PATH and assert horizon cli invocation doesn't result in success, and the cmd output has the expected error message about parameter conflict?  or is that error outcome covered in different test already?

services/horizon/internal/integration/parameters_test.go
```diff
@@ -41,6 +52,20 @@ const (
...
 
+func NewParameterTestSkipContainerCreation(t *testing.T, params map[string]string) *integration.Test {
+	return NewParameterTestWithEnvSkipContainerCreation(t, params, map[string]string{})
+}
+
+func NewParameterTestWithEnvSkipContainerCreation(t *testing.T, params, envvars map[string]string) *integration.Test {
+	config := integration.Config{
+		ProtocolVersion:         17,
+		SkipHorizonStart:        true,
+		SkipContainerCreation:   true,
```
sreuland, line 63: interesting find, so `SkipContainerCreation` already existed, but was not used anywhere, and now it can be! One thing I noticed as reading the code is that name doesn't convey much context, for consideration, renaming to `SkipCoreStart` or something similar.

services/horizon/internal/integration/parameters_test.go
```diff
@@ -120,6 +148,96 @@ func (suite *FatalTestCase) TestEnvironmentPreserved() {
...
+		t.Skip()
+	}
+
+	t.Run("network parameter pubnet", func(t *testing.T) {
+		localParams := integration.MergeMaps(networkParamArgs, map[string]string{
+			horizon.NetworkFlagName: horizon.StellarPubnet,
+		})
+		test := NewParameterTestSkipContainerCreation(t, localParams)
+		err := test.StartHorizon()
+		time.Sleep(10 * time.Second)
```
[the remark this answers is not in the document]
sreuland: yes, would be unfortunate if the test assertions rely on connectivity to external systems that are not under control of test like actual testnet or pubnet. 

Instead, may need to take a different approach for test verification, such as check `StartHorizon()` doesn't result in cmd line errors/warnings for the NETWORK usage, and then inspect the local expected outcomes from the Horizon App instance such as horizon's ingestion config has the correct network params defined and the generated core config file has the correct content for testnet/pubnet.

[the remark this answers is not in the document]
urvisavla: >why is the sleep necessary?

The sleep was added to ensure that Horizon fully starts before sending a shutdown signal. Without the sleep, the shutdown is initiated before ingestion co-routine (ingest.StartStateMachine) is started, resulting in a race condition error where [s.wg.Add()](https://github.com/stellar/go/blob/master/services/horizon/internal/ingest/main.go#L641) is called after [s.wg.Wait()](https://github.com/stellar/go/blob/master/services/horizon/internal/ingest/main.go#L847).  After giving it some more thought, I think it might be more appropriate to handle this scenario within the app/ingestion rather than adding sleep in the test??



>yes, would be unfortunate if the test assertions rely on connectivity to external systems that are not under control of test like actual testnet or pubnet.
>
>Instead, may need to take a different approach for test verification, such as check StartHorizon() doesn't result in cmd line errors/warnings for the NETWORK usage, and then inspect the local expected outcomes from the Horizon App instance such as horizon's ingestion config has the correct network params defined and the generated core config file has the correct content for testnet/pubnet.

The tests are independent of external connectivity. They can run without any network connectivity and will fail if any of the conditions mentioned in the [comments](https://github.com/stellar/go/pull/4974/files#diff-6f737140ca2744b223417ca5d5f6f4a20e9812494960796e774932e97b85da5fR196-R200) do not match.
```
/The test will fail for scenarios with an invalid static configuration, such as:
// - The default passphrase in Horizon is different from the one specified in the captive-core config.
// - History archive URLs are not configured.
// - The captive-core TOML config file is invalid.
// The test will also fail if an invalid parameter is specified."

```

services/horizon/internal/integration/parameters_test.go
urvisavla: You're correct. By default, ENABLE_CAPTIVE_CORE_INGESTION is already set to true, so there's no need to explicitly set it to true. However,  it is explicitly set to `false` in [StartHorizon()](https://github.com/stellar/go/blob/master/services/horizon/internal/test/integration/integration.go#L331),  so to override that, I still need to provide an empty value for ENABLE_CAPTIVE_CORE_INGESTION.

services/horizon/internal/integration/parameters_test.go
urvisavla: I completely agree with you. It was bothering me as well. I was considering adding varargs but your approach is even better. I love it! I've updated the code based on your suggestion.

services/horizon/internal/integration/parameters_test.go
urvisavla: In these tests, the StartHorizon() function will return an error if an invalid combinations of parameters, whether provided through command-line or environment variables. For better coverage, we have the option of either adding more integration tests to handle different parameter combinations and verify that StartHorizon() fails to start the Horizon and returns an error, or introduce unit tests to test each combination of configuration parameters (if adding more integration tests feels excessive). If we choose the unit test approach, we can add unit tests for the flags.go `ApplyFlags` function. This function is responsible for validating both commandline and environment variables. Though when we added the `NETWORK` parameter, we've done some refactoring (and added [unit tests](https://github.com/stellar/go/blob/master/services/horizon/internal/flags_test.go))  in flags.go, further refactoring  will be needed to remove database related functionality so that we can unit test `ApplyFlags`.

Please let me know your thoughts on the best way to proceed here. Thanks!

aditya1702: @urvisavla @sreuland I think this should be done as an entire separate issue/task of its own - adding tests for command line flags and other parameters. Currently, the tests are a little disjoint with some in flags_test.go, parameters_test.go and some in command_line_args_test.go (which I added in my previous PR). Maybe we can combine these into a proper test suite for testing the ApplyFlags function.

services/horizon/internal/integration/parameters_test.go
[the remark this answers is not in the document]
urvisavla: >Without the sleep, the shutdown is initiated before ingestion co-routine (ingest.StartStateMachine) is started, resulting in a race condition error where [s.wg.Add()](https://github.com/stellar/go/blob/master/services/horizon/internal/ingest/main.go#L641) is called after [s.wg.Wait()](https://github.com/stellar/go/blob/master/services/horizon/internal/ingest/main.go#L847).

I think that my previous statement was not accurate and the actual reason for the race condition is not the order in which Add() and Wait() are called, but rather the fact that they are being concurrently accessed.

We are currently seeing this problem in our master branch in the ingestion filtering test cases. I have created an [issue](https://github.com/stellar/go/issues/4987) with my analysis for the root cause. We can discuss the best solution on the issue, if we decide to address it.

 @aditya1702, In the meantime, to fix the CI on master, we can either revert the test that is currently failing in the master branch or add sleep in it??

services/horizon/internal/integration/parameters_test.go
sreuland: got it, thanks. if ENABLE_CAPTIVE_CORE_INGESTION is not set to empty by the test, then the `ENABLE_CAPTIVE_CORE_INGESTION=false` would have triggered a conflict on tests that submit NETWORK flag correct? which is the expectation when both ENABLE_CAPTIVE_CORE_INGESTION=false and NETWORK=pubnet|testnet are present.

services/horizon/internal/integration/parameters_test.go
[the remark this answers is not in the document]
sreuland: > The tests are independent of external connectivity. They can run without any network connectivity and will fail if any of the conditions mentioned in the comments do not match.

ok, I see that now, thanks. the `SkipContainerCreation = true` prevents the docker-compose for core and it also disables captive core settings. It might be worthwhile to rename that to `SkipCoreContainerCreation`, as it's a bit ambiguous on what it's doing.

services/horizon/internal/integration/parameters_test.go
sreuland: it looks like there is decent separation of unit tests(flags_test.go) which check the expected output at that layer, and then integration tests which overlap across unit tests but check outputs at a higher system level like the Horizon App instance or err's which parameters_test.go looks positioned in. @aditya1702 , do you think the consolidation could boil down to refactoring command_line_args_test.go to here?

services/horizon/internal/integration/parameters_test.go
urvisavla: done.

services/horizon/internal/integration/parameters_test.go
urvisavla: Changed it to `SkipCoreContainerCreation`

services/horizon/internal/integration/parameters_test.go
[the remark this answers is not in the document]
urvisavla: Renamed `SkipContainerCreation` to `SkipCoreContainerCreation`.


>and it also disables captive core settings

Not sure if I understand that.

services/horizon/internal/integration/parameters_test.go
urvisavla: Yes, your expectation is correct. But the check was missing, so instead it was displaying "flag --stellar-core-url  can not be empty" error message. I've now added the required check to show a more helpful error message in this [PR](https://github.com/stellar/go/pull/4988).

services/horizon/internal/integration/parameters_test.go
aditya1702: @sreuland @urvisavla Yeah I think I was confused where to add my tests and created a new file. I have consolidated it to parameters_test.go in #4994

services/horizon/internal/integration/parameters_test.go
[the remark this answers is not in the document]
urvisavla: > > and it also disables captive core settings
> 
> Not sure if I understand that.

Never mind, I got it and yes it also disables captive core settings.

services/horizon/internal/integration/parameters_test.go
urvisavla: Adding 2 sleep in the tests to work around the race condition. Created this issue #5005  to address it in the code.

services/horizon/internal/integration/parameters_test.go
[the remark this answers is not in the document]
urvisavla: Adding 2 sleep in the tests to work around the race condition. Created this issue #5005  to address it in the code.

services/horizon/internal/integration/parameters_test.go
urvisavla: I've included two additional integration tests [c36e824](https://github.com/stellar/go/pull/4974/commits/c36e82490d039e9cae086bdf9dea4a3b0e5b7164) to validate  `--network-passphrase` and  `--history-archive-urls` parameters in conjunction with the `--network` parameter..  Different combinations of these parameters are covered in [unit test ](https://github.com/stellar/go/blob/master/services/horizon/internal/flags_test.go) so I've not repeated them.

services/horizon/internal/integration/parameters_test.go
```diff
@@ -127,6 +130,139 @@ func (suite *FatalTestCase) TestEnvironmentPreserved() {
 	assert.Equal(t, "original value", envValue)
 }
 
+// TestInvalidNetworkParameters Ensure that Horizon returns an error when
+// using NETWORK environment variables, history archive urls or network passphrase
+// parameters are also set.
+func TestInvalidNetworkParameters(t *testing.T) {
+	if !integration.RunWithCaptiveCore {
```
sreuland, line 137: I think can remove this skip on run with captive core, this test can run regardless, because `testConfig.SkipCoreContainerCreation = true` pre-empts the test from getting to the point of checking or running captive core in any case?

services/horizon/internal/integration/parameters_test.go
```diff
@@ -127,6 +130,139 @@ func (suite *FatalTestCase) TestEnvironmentPreserved() {
...
+			param:        horizon.NetworkPassphraseFlagName,
+		},
+	}
+
+	for _, testCase := range testCases {
+		t.Run(testCase.name, func(t *testing.T) {
+			localParams := integration.MergeMaps(networkParamArgs, map[string]string{
+				horizon.NetworkFlagName:                    testCase.networkValue,
+				horizon.EnableCaptiveCoreIngestionFlagName: "true",
+				testCase.param:                             "NETWORK_PARAM",
```
sreuland, line 170: at first I was confused about what `NETWORK_PARAM` meant since it has no relevance to do the param key used, now see it's just any string value to trigger presence rather than nil.  it might be easier to convey the test intent by adding a `testcase.paramValue` also , that way can show the typical/expected values for the params, so, it's more obvious as to what the test is validating will get rejected.

services/horizon/internal/integration/parameters_test.go
```diff
@@ -127,6 +130,139 @@ func (suite *FatalTestCase) TestEnvironmentPreserved() {
...
+// TestNetworkParameter Ensure that Horizon successfully starts the captive-core
+// subprocess using the default configuration when --network [testnet|pubnet]
+// commandline parameter.
+//
+// In integration tests, we start Horizon and stellar-core containers in standalone mode
+// simultaneously. We usually wait for Horizon to begin ingesting to verify the test's
+// success. However, for "pubnet" or "testnet," we can not wait for Horizon to catch up,
+// so we skip starting stellar-core containers.
+func TestNetworkParameter(t *testing.T) {
+	if !integration.RunWithCaptiveCore {
```
sreuland, line 195: similar question here, this check may not be necessary since `SkipCoreContainerCreation=true` is used

services/horizon/internal/integration/parameters_test.go
urvisavla: Yes, but it will not prevent the test from running and the test will fail if it can't find the local stellar-core binary. I believe we run integration tests w/ and w/o captive-core so the check here is required for when running w/o captive-core.

services/horizon/internal/integration/parameters_test.go
```diff
@@ -127,6 +130,139 @@ func (suite *FatalTestCase) TestEnvironmentPreserved() {
...
+			})
+			testConfig := integration.GetTestConfig()
+			testConfig.SkipCoreContainerCreation = true
+			testConfig.HorizonIngestParameters = localParams
+			test := integration.NewTest(t, *testConfig)
+			err := test.StartHorizon()
+			// Adding sleep as a workaround for the race condition in the ingestion system.
+			// https://github.com/stellar/go/issues/5005
+			time.Sleep(2 * time.Second)
+			assert.NoError(t, err)
```
sreuland, line 217: this could be a chance to assert that the NETWORK param triggered correct creation of toml struct in memory:
`assert(test.GetHorizonIngestConfig().CaptiveCoreToml.NetworkPassphrase, testcase.Passphrase)`

and it should also have set archive urls correctly for network:
`assert(test.GetHorizonIngestConfig().HistoryArchiveURLs, testcase.ArchiveURLs)`

services/horizon/internal/integration/parameters_test.go
```diff
@@ -127,6 +130,139 @@ func (suite *FatalTestCase) TestEnvironmentPreserved() {
...
+			test.Shutdown()
+		})
+	}
+}
+
+// TestNetworkEnvironmentVariable Ensure that Horizon successfully starts the captive-core
+// subprocess using the default configuration when the NETWORK environment variable is set
+// to either pubnet or testnet.
+//
+// In integration tests, we start Horizon and stellar-core containers in standalone mode
```
sreuland, line 227: can probably skip restating this aspect as it was mentioned on other test, or can move this documentation off the test method and onto the  `SkipCoreContainerCreation` definition.

sreuland (approved):
the new test coverage looks good, i just left a few minor suggestions here to consider before merge.

services/horizon/internal/integration/parameters_test.go
urvisavla: same as above

services/horizon/internal/integration/parameters_test.go
urvisavla: The code validates against the parameter name and not the value and the expected error message conveys which invalid parameter was passed not the value. Changed it to use parameter name for the value and added a comment to make it clear.

services/horizon/internal/integration/parameters_test.go
urvisavla: I left the comment here instead of SkipCoreContainerCreation def because there might use cases for skipping container creation other than what we're using it for here i.e. testing `pubnet` or `testnet`.  

I know the comment appears twice but if it's alright i'd like to also keep the comment here. Mainly because we don't want to lose the context when we be remove `TestNetworkParameter ` the test as part of config param deprecation.

services/horizon/internal/integration/parameters_test.go
urvisavla: done.

PR #4978, opened 2023-07-26
services/horizon add horizon binary building Makefile

jacekn:
### What

This PR adds Makefile entries to allow us to build horizon in an isolated, docker based, environment.

### Why

This will make build process more transparent and it will simplify build systems where horizon is built.

### DIFF
--- a/services/horizon/Makefile
+++ b/services/horizon/Makefile
@@ -1,0 +1,16 @@
+SUDO := $(shell docker version >/dev/null 2>&1 || echo "sudo")
+
+ifndef VERSION_STRING
+        $(error VERSION_STRING environment variable must be set. For example "2.26.0-d2d01d39759f2f315f4af59e4b95700a4def44eb")
+endif
+
+binary-build:
+	$(SUDO) docker run --rm $(DOCKER_OPTS) -v $(shell pwd)/../../:/go/src/github.com/stellar/go \
+		--pull always \
+		--env CGO_ENABLED=0 \
+		--env GOFLAGS="-ldflags=-X=github.com/stellar/go/support/app.version=$(VERSION_STRING)" \
+		golang:1.20-bullseye \
+		/bin/bash -c '\
+			git config --global --add safe.directory /go/src/github.com/stellar/go && \
+			cd /go/src/github.com/stellar/go && \
+			go build -o horizon -trimpath -v ./services/horizon'
--- a/services/horizon/Makefile
+++ b/services/horizon/Makefile
@@ -4,8 +4,10 @@ ifndef VERSION_STRING
         $(error VERSION_STRING environment variable must be set. For example "2.26.0-d2d01d39759f2f315f4af59e4b95700a4def44eb")
 endif
 
+DOCKER_PLATFORM := $(shell docker system info --format '{{.OSType}}/{{.Architecture}}')
+
 binary-build:
-	$(SUDO) docker run --rm $(DOCKER_OPTS) -v $(shell pwd)/../../:/go/src/github.com/stellar/go \
+	$(SUDO) docker run --platform $(DOCKER_PLATFORM) --rm $(DOCKER_OPTS) -v $(shell pwd)/../../:/go/src/github.com/stellar/go \
 		--pull always \
 		--env CGO_ENABLED=0 \
 		--env GOFLAGS="-ldflags=-X=github.com/stellar/go/support/app.version=$(VERSION_STRING)" \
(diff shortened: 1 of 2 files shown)


mwtzzz (approved):
LGTM

services/horizon/Makefile
```diff
@@ -0,0 +1,16 @@
...
+binary-build:
+	$(SUDO) docker run --rm $(DOCKER_OPTS) -v $(shell pwd)/../../:/go/src/github.com/stellar/go \
+		--pull always \
+		--env CGO_ENABLED=0 \
+		--env GOFLAGS="-ldflags=-X=github.com/stellar/go/support/app.version=$(VERSION_STRING)" \
+		golang:1.20-bullseye \
+		/bin/bash -c '\
+			git config --global --add safe.directory /go/src/github.com/stellar/go && \
+			cd /go/src/github.com/stellar/go && \
+			go build -o horizon -trimpath -v ./services/horizon'
```
sreuland, line 16: how will this work when the host is on a different o/s than linux variant, like mac, since the compiled horizon bin will be linked with linux libs related to `1.20-bullseye`, the resulting bin is placed on the host path through the vol mount, but the bin is only runnable on same linux o/s and arch type correct? maybe a pre-check that host is on linux, or dynamically set `golang:1.20-bullseye` when host is linux and no tag qualifier if non-linux?

jacekn: Good point, I'll make some improvements

jacekn: @sreuland I pushed commit with some improvements. Could you check if that makes sense?

sreuland: I'll try the pr locally on mac to verify, but I think there may be an edge case where the built go binary is linked with linux .so libs when running in the docker container, and then that bin is placed on the mac host, at which point the mac won't be able to execute the bin because the link loader won't find same libs.

tsachiherman: unless you use cgo, there is no risk for that to happen, since it would always compile all the libraries statically.
(keep in mind that we might use a third-party package that use cgo..).
for mac compiling I think that it's doable to have an emulated mac docker and compile it there.. and it would also be ok to run it as long as the host is a mac computer.

sreuland: it worked on mac m1:
```
$ VERSION_STRING=xyz  make -B horizon
```

afterwards inspected the binary file, and it does show the exec format as ELF which is linux, not Mach-O format which is supposed to be for Mac, but it runs on the mac, so, I guess it's not significant at least for mac hosts:
```
$ file horizon
horizon: ELF 64-bit LSB executable, ARM aarch64, version 1 (SYSV), statically linked, Go BuildID=pyvzo4OeJB9uylH-L_KY/9axYX1yBlHzpwa4s5iTT/RW9wBMRFUmH_bodTZw0M/FZxIlZ_Ab0fwNO5GtM99, with debug_info, not stripped
```

I was able to run `horizon version` and `horizon db init`

just for comparison when I run go build directly on mac, I see the Mach-O format:

```
$ go build -o horizon -trimpath -v ./services/horizon
$ file horizon
horizon: Mach-O 64-bit executable arm64
```

sreuland (approved):

services/horizon/Makefile
```diff
@@ -0,0 +1,18 @@
...
+binary-build:
+	$(SUDO) docker run --platform $(DOCKER_PLATFORM) --rm $(DOCKER_OPTS) -v $(shell pwd)/../../:/go/src/github.com/stellar/go \
+		--pull always \
+		--env CGO_ENABLED=0 \
+		--env GOFLAGS="-ldflags=-X=github.com/stellar/go/support/app.version=$(VERSION_STRING)" \
+		golang:1.20-bullseye \
+		/bin/bash -c '\
+			git config --global --add safe.directory /go/src/github.com/stellar/go && \
+			cd /go/src/github.com/stellar/go && \
+			go build -o horizon -trimpath -v ./services/horizon'
```
urvisavla, line 18: Is there a specific reason for keeping the binary name as "horizon" instead of "stellar-horizon"? The package manager appears to install it as "stellar-horizon," and in our [public documentation](https://developers.stellar.org/docs/run-api-server/installing#completing-and-testing-your-installation), we advise users to copy the binary to "stellar-horizon" for consistency with the package manager.

services/horizon/Makefile
jacekn: Thanks for testing Shawn!

services/horizon/Makefile
jacekn: Thank you, this is a very good point. I updated binary name to `stellar-horizon`

Issue #4915, opened 2023-06-16
services/horizon: Deprecation and Removal of --db-url Command Line Parameter


urvisavla:
### What problem does your feature solve?

In our docs, we recommend using environment variables instead of command line flags for Horizon as they provide better security for sensitive information.

### What would you like to see?

- [ ]  Display warning messages regarding the deprecation of the `--db-url` command line argument by year-end and  transition to using environment variable.
- [ ]  Identify any additional command-line parameters that require deprecation.
- [ ] Phase out the `--db-url` command line parameter by year-end.

urvisavla:
As part of this we can also:
1. Assess whether there are any command-line flags that are not inherently "global" and could potentially vary based on the specific subcommand.  In such case, we may need to allow users to override these specific environment variables using command-line options. We can start with identifying any flags falling within this category. 
2. Review current validations for each command and their respective subcommands tree and make adjustments/improvements where necessary.

urvisavla:
I have a [PR](https://github.com/stellar/go/pull/5044) that introduces a deprecation warning for the `--db-url` command-line  parameter. This PR also includes the necessary changes to support a config flag exclusively via environment variables. However, I'm not sure if this is the best way to go. Here are a few reasons why:
1. If we make one config parameter exclusively environment variables, the inconsistency could confuse users.
2. Users can run multiple Horizon instances, and these instances might need to connect to different databases. 
For example, they might have an API-only instance and an ingestion instance on the same machine. In such cases, the API instance might need to use a read-only database (ro db instance), while the ingestion instance needs the `db-url` parameter to point to a read-write database (rw db instance). To manage this, users would have to create extra scripts to pass different environment variables. Keeping the option to set `db-url` via the command line might be simpler for users.

I agree that command-line parameters may not be the most secure approach, and we may need to rethink how we handle Horizon's configuration as a whole. However, recommending the use of environment variables for configuring Horizon in our public documentation guides users toward a more secure and proper setup.

I'd love to hear your thoughts on this, @mollykarcher, @sreuland, @aditya1702.

urvisavla:
As discussed during the Partner Experience team meeting, we have decided not to proceed with this and instead favor https://github.com/stellar/go/issues/5045. Hence, closing this issue.

Issue #4855, opened 2023-04-26
services/horizon: fully remove non-captive core horizon support


mollykarcher:
### What problem does your feature solve?

Currently, running horizon in non-captive-core (or legacy) mode is still supported, just [deprecated](https://developers.stellar.org/docs/run-api-server/migrating). We also still use 2 legacy instances for ingestion and transaction submission in our own public horizon deployment. We've maintained this setup due to concern over startup time of the captive-core instances when they are restarted or state is rebuilt.

However, those concerns may have been alleviated now that core has merged a [change](https://github.com/stellar/stellar-core/pull/3655) to significantly speed-up that start-up time.

### What would you like to see?

- [x] Observe startup time on the core release where this change goes out (v19.10) to validate it's quick/acceptable. We may also want to sanity-check this restart startup time a few times on staging (or similar)
- [x] Work with ops to remove the legacy ingesters from our deployment. This requires doing in staging first, and consider having more than 2 instances of captive core ingesters, need to confirm stability and responsiveness on restarts in staging, identify the typical wait for catch up ties after restarts, so, no surprises when lands in prod. Ops ticket referencing this work:
    - [x] https://github.com/stellar/ops/issues/2557
- [x] Note that these are currently also serving transaction submission requests, so we may want to first move txsub to the captive-core instances and observe that behavior/performance for some time before fully killing these off. Consider keeping transaction submission isolated from ingestion, this would keep all user behavior/API requests from ever bottlenecking the ingestion process. So instead of completely dropping the legacy ingestors, we redeploy them as captive-core instances and disable ingestion on them, so they're only serving transaction submission. But we should consider the pros/cons here. Ops ticket referencing this work:
    - [x] https://github.com/stellar/ops/issues/2558
- [x] Remove any code and integration tests supporting the db backend or legacy setup:
  * `--stellar-core-db-url`
- [x] Update external developer docs 
- [x] deprecate usage of `ENABLE_CAPTIVE_CORE_INGESTION` flag, the distinction is no longer relevant, `INGEST` flag presumes captive core ingestion.

sydneynotthecity:
@Shaptic says that captive core nodes in prod on an upgrade sometimes fall out of sync (~20 min+) with current state. We need to figure out why they fall behind because Horizon does fall back on these ingestion nodes when that occurs.

mollykarcher:
> captive core nodes in prod on an upgrade sometimes fall out of sync (~20 min+) with current state
[...]

PR #4979, opened 2023-07-26
Add DISABLE-TX-SUB configuration parameter

### LINKED ISSUE #4914: services/horizon: Add "DISABLE_TX_SUB" configuration parameter
To streamline configuration and enhance flexibility in Horizon, we want to introduce the `TX_SUB_DISABLED` configuration parameter. This parameter enables or disables transaction submission functionality in Horizon.

Add `DISABLE_TX_SUB` configuration parameter for Transaction Submission in Horizon. This parameter can be set as an environment variable similar to other parameters in Horizon. 

Key features:
1. Enable/disable Transaction Submission: By default `DISABLE_TX_SUB` should be set to false. Horizon should accept or reject POST requests with a 405 HTTP error code, method not supported fro the `/tx` endpoint based on this setting.

2. Enhance startup validation: When `DISABLE_TX_SUB` is false, Horizon should perform a startup check to make sure data ingestion is happening by checking it's DB and If no ingestion is detected, Horizon should exit with an appropriate error.

3. If either "INGESTION=true" or "DISABLE_TX_SUB=false" is set in the configuration, Horizon should validate incoming config parameters include network config via either the new parameter "NETWORK" or specify the lower level parameters, if not, a startup error should be emitted stating network config is needed.  Otherwise, when `DISABLE_TX_SUB=true` and `INGEST=false`,  all network oriented config params are no longer needed, such as `NETWORK` and `STELLAR_CORE_URL`because no core connectivity is required, if the are present, horizon should emit error saying the flags are in conflict.

Confirm that Horizon can use captive core's local `HTTP_PORT` to submit transactions or does it need to send to a watcher/validator node? We need to mention this expectation accurately in new [config docs](https://github.com/stellar/stellar-docs/issues/141).

aditya1702:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Closes #4914 

- Introduce a `DISABLE-TX-SUB` parameter which disables transaction submission functionality for Horizon.
- Add integration and unit tests for the change.

### Why

This change helps separate the transaction submission functionality of Horizon. If users want to just use Horizon as an API layer, they can switch off this functionality.

### Known limitations

None


services/horizon/internal/integration/parameters_test.go
```diff
@@ -461,9 +461,9 @@ func TestHelpOutputForNoIngestionFilteringFlag(t *testing.T) {
 	if err := horizonCmd.Execute(); err != nil {
 		fmt.Println(err)
 	}
-
 	output := writer.(*bytes.Buffer).String()
 	assert.NotContains(t, output, "--exp-enable-ingestion-filtering")
+	assert.NotContains(t, output, "--disable-tx-sub")
```
Shaptic, line 466: I'm a bit confused and I haven't seen this test before: why is this parameter expected to be missing from the output?

services/horizon/internal/actions/submit_transaction_test.go
```diff
@@ -153,3 +157,52 @@ func TestClientDisconnectSubmission(t *testing.T) {
...
+		Submitter:         mockSubmitter,
+		NetworkPassphrase: network.PublicNetworkPassphrase,
+		DisableTxSub:      true,
+		CoreStateGetter:   mock,
+	}
+
+	form := url.Values{}
+	form.Set("tx", TransactionEnvelopeXDR)
+
+	var p = &problem.P{
```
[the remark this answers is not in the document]
aditya1702: @Shaptic I am not reusing it anywhere else but the tx-disabled test.

services/horizon/internal/flags.go
```diff
@@ -820,8 +838,6 @@ func ApplyFlags(config *Config, flags support.ConfigOptions, options ApplyOption
 		config.Ingest = true
 	}
 
-	config.EnableIngestionFiltering = true
```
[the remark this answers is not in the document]
aditya1702: @Shaptic Yes this is on by default. I just moved it CustomSetValue function in the flag definition.

services/horizon/internal/integration/parameters_test.go
aditya1702: @Shaptic Ah my bad, I meant to do assert.Contains instead.

services/horizon/internal/flags_test.go
```diff
@@ -131,3 +132,34 @@ func Test_createCaptiveCoreConfig(t *testing.T) {
...
+		{
+			name: "disable-tx-sub set to false and no network and ingestion params set",
+			config: Config{
+				NetworkPassphrase:  "",
+				HistoryArchiveURLs: []string{},
+				DisableTxSub:       false,
+			},
+			flags:   flags,
+			options: ApplyOptions{RequireCaptiveCoreConfig: true, AlwaysIngest: false},
+			errStr: "invalid config: --disable-tx-sub set to false but ingestion has been disabled." +
```
urvisavla, line 155: I can't seem to find where this error is generated in the code.

services/horizon/internal/integration/parameters_test.go
aditya1702: @Shaptic So my reasoning is that we are recommending that users specify an environment variable rather than a flag. Hence the help output will not contain this flag. Is this a good way or should we allow it to be shown in the help output?

services/horizon/internal/flags.go
```diff
@@ -798,6 +808,14 @@ func setCaptiveCoreConfiguration(config *Config) error {
 		config.StellarCoreURL = fmt.Sprintf("http://localhost:%d", config.CaptiveCoreToml.HTTPPort)
 	}
 
+	if !config.DisableTxSub {
+		if config.Network == "" && (config.NetworkPassphrase == "" || len(config.HistoryArchiveURLs) == 0) {
+			return fmt.Errorf("invalid config: --%s set to false but no --%s flag or "+
```
urvisavla, line 813: In what scenario does this code path get executed?  I think if this condition were met, it would have bailed before reaching this point?

services/horizon/internal/flags_test.go
aditya1702: Ah I removed the config error but did not remove this test.

services/horizon/internal/flags.go
aditya1702: I think you are correct. The network passphrase and history archive URLs will always be checked so this code will never be hit.

services/horizon/internal/httpx/router.go
```diff
@@ -319,6 +320,7 @@ func (r *Router) addRoutes(config *RouterConfig, rateLimiter *throttled.HTTPRate
 	r.Method(http.MethodPost, "/transactions", ObjectActionHandler{actions.SubmitTransactionHandler{
 		Submitter:         config.TxSubmitter,
 		NetworkPassphrase: config.NetworkPassphrase,
+		DisableTxSub:      config.DisableTxSub,
```
urvisavla, line 323: Do we need to add test to validate the expected http response on this endpoint when the tx_sub is disabled?

services/horizon/internal/flags.go
urvisavla: According to the [docs](https://github.com/stellar/stellar-docs/blob/partner_experience/docs/run-platform-server/configuring.mdx#L63-L69), it's possible to have tx sub enabled on a non-ingesting instance. Are we handling that scenario?

services/horizon/internal/actions/submit_transaction.go
```diff
@@ -155,6 +156,19 @@ func (handler SubmitTransactionHandler) GetResource(w HeaderWriter, r *http.Requ
 		return nil, hProblem.StaleHistory
 	}
 
+	if handler.DisableTxSub {
+		return nil, &problem.P{
+			Type:   "transaction_submission_disabled",
+			Title:  "Transaction Submission Disabled",
+			Status: http.StatusMethodNotAllowed,
+			Detail: "Transaction submission has been disabled for Horizon. " +
+				"To enable it again, set env variable DISABLE_TX_SUB to false.",
```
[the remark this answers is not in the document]
Shaptic: Since the default is false (as in, txsub is enabled), recommending you remove it leads to a simpler config. For me, at least, it's some unnecessary mental load to decipher what `DISABLE_TX_SUB=false` means due to the double-negative.

services/horizon/internal/integration/parameters_test.go
Shaptic: Ah, I see. Is that something we're doing with new flags going forward? I didn't know that was our recommendation.

services/horizon/internal/flags.go
Shaptic: This seems odd: you need ingestion to get tx status after submission. Otherwise, you don't know what to return.

urvisavla: > You can run none or multiple instances to serve transaction submission requests. If you run an instance with transaction submission enabled, your deployment is required to perform the ingestion role on the same database referenced here. Horizon transaction submission depends on **live** ingestion to confirm tx submission status.

> If ingestion is planned to be done on a separate instance, add `INGEST=false` on this instance, otherwise don't include the parameter, Horizon will default to `INGEST=true` if absent.


>  When enabled, requires **live** ingestion process to be running on the same database because Horizon depends on new ledgers from network to confirm a transaction submission status, Horizon will report startup error if it detects no **live** ingestion. Requires `NETWORK` or `STELLAR_CORE_URL` be defined for access to network.

hmm, as per the docs, as long at there is another instance live ingesting to the same db, it should work. It further says that the Horizon will detect if there is no live ingestion on the db and report error 🤔

services/horizon/internal/integration/parameters_test.go
urvisavla: IMO, it's better to keep things consistent by adding it to the help. As far as I know, the help serves as documentation for the most parameters (apart from checking the code).  So while we'll be making an exception for[ --db-url ](https://github.com/stellar/go/issues/4915) due to security reasons, it should be ok to accept this one as command-line parameter for now.  thoughts?

services/horizon/internal/flags.go
Shaptic: oh okay perfect, that seems fine then

services/horizon/internal/integration/parameters_test.go
aditya1702: @urvisavla @Shaptic Yes, I agree with keeping it consistent. But we should discuss changing the help output in future so that it doesnt encourage using the configurations as flags.

services/horizon/internal/flags.go
aditya1702: @urvisavla Do I need to add any check for this scenario? Since, if you pass INGEST=false, it requires you to specify --stellar-core-db-url and gives an error if not specified.

services/horizon/internal/httpx/router.go
aditya1702: I have already added one to test the response - https://github.com/stellar/go/pull/4979/files#diff-35959631e97ae0aed640ab44ddfec192b275ae8ea02bd8c999f87a6814d019cbR161

urvisavla: Cool! Sorry I missed that.

services/horizon/internal/flags.go
sreuland: @aditya1702 , `--stellar-core-db-url` is slated to be removed in https://github.com/stellar/go/issues/4855 as part of deprecating non-captive core ingestion config and that would have removed it entirely, but this is a bit strange that there is validation requiring --stellar-core-db-url when ingest=false, as that means it doesn't need to perform ingestion e.g. by connecting to --stellar-core-db-url or using captive core, I wonder if the validation is getting triggered due to just absence of any captive core settings, in any case, it appears will need to adjust that validation related to `--stellar-core-db-url` and `INGEST=false` to allow the expected outcomes when DISABLE-TX-SUB is present or not for good housekeeping?

services/horizon/internal/actions/submit_transaction.go
```diff
@@ -155,6 +156,19 @@ func (handler SubmitTransactionHandler) GetResource(w HeaderWriter, r *http.Requ
 		return nil, hProblem.StaleHistory
 	}
 
+	if handler.DisableTxSub {
```
sreuland, line 159: should this flag should be checked first at top of handler method to short circuit? because these other validations are moot if DisableTxSub is true, and some of these may even trigger fail before getting here like `handler.GetCoreState()` since --stellar-core-url can be undefined when DisableTxSub is true.

services/horizon/internal/actions/submit_transaction_test.go
```diff
@@ -20,6 +20,10 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+const (
```
sreuland, line 23: if you follow the [suggestion above to check disabled first in the handler](https://github.com/stellar/go/pull/4979/files#r1300735228), then can remove this variable with valid tx xdr, as can set the `tx` parameter to any value on a test for disabled tx flag, since it will not be parsed anyways.

services/horizon/internal/integration/parameters_test.go
sreuland: @aditya1702 , good point, the help output for sub-commands and flags will need to cover that as best practice, and to avoid it scope creeping into here, how about we track that as a separate concern on - https://github.com/stellar/go/issues/5022, that way in this pr, can follow the recommendations here in the interim on flag consistency and just initiate DisableTxSub flag with `Hidden: false`?

services/horizon/internal/actions/submit_transaction.go
aditya1702: Yes, nice idea. I will move it to the top.

services/horizon/internal/integration/parameters_test.go
```diff
@@ -397,6 +397,55 @@ func TestIngestionFilteringAlwaysDefaultingToTrue(t *testing.T) {
 	})
 }
 
+func TestDisableTxSub(t *testing.T) {
+	t.Run("DISABLE_TX_SUB=false, INGEST=false, STELLAR_CORE_URL=''", func(t *testing.T) {
```
sreuland, line 401: minor, on the test cases, use a statement for the title, to express the expected outcome for clarity, such as 'It requires core url when tx sub not disabled'

services/horizon/internal/integration/txsub_test.go
sreuland, line 1: I think it might be worth bringing this file back but add two new tests `itSubmitsWhenTxSubNotDisabled` and 'itReturnsErrorWhenTxSubisDisabled', since the endpoint now has new error behavior when flag is disabled. 

copy from the same tests done in parameters_test.go,  after starting horizon then initiate a tx sub to confirm the txsub flag values match the outcomes from the http endpoint, an example tx sub frag to use:

```
txResp, err := itest.SubmitOperations(itest.MasterAccount(), itest.Master(),
		&txnbuild.Payment{
				Destination: master.Address(),
				Amount:      "10",
				Asset:       txnbuild.NativeAsset{},
			},
	)

```

services/horizon/internal/integration/parameters_test.go
```diff
@@ -397,6 +397,55 @@ func TestIngestionFilteringAlwaysDefaultingToTrue(t *testing.T) {
 	})
 }
 
+func TestDisableTxSub(t *testing.T) {
+	t.Run("require stellar-core-url when both DISABLE_TX_SUB=false and INGEST=false", func(t *testing.T) {
+		localParams := integration.MergeMaps(networkParamArgs, map[string]string{
+			horizon.NetworkFlagName:          "testnet",
+			horizon.IngestFlagName:           "false",
+			horizon.DisableTxSubFlagName:     "false",
+			horizon.StellarCoreDBURLFlagName: "",
```
sreuland, line 406: core-db-url isn't relevant to coverage over this test area correct? I would remove if if so, to avoid any confusion, and keep test small as possible.

sreuland (approved):
tests look good! make sure CI tests pass before merge though, noticed one failing, not sure on the reason.

PR #4980, opened 2023-07-27
Ignore liquidity pool overflow errors for slippage

Shaptic:


tsachiherman (approved):
looks reasonable, although I can't really validate that this is the correct action, as I'm not familiar with the details of the CAP.

PR #4985, opened 2023-07-27
ingest/ledgerbackend: Add parameter to enforce diagnostic events in captive core backend

2opremio:
The configuration parameter (`EnforceSorobanDiagnosticEvents`) is optional and makes sure `ENABLE_SOROBAN_DIAGNOSTIC_EVENTS` is set to true for captive core unless the user explicitly sets `ENABLE_SOROBAN_DIAGNOSTIC_EVENTS=false` in the appendix file.

This fixes https://github.com/stellar/soroban-tools/issues/806 (when `EnforceSorobanDiagnosticEvents` is set to true)


ingest/ledgerbackend/toml.go
```diff
@@ -326,6 +326,8 @@ type CaptiveCoreTomlParams struct {
 	UseDB bool
 	// the path to the core binary, used to introspect core at runtie, determine some toml capabilities
 	CoreBinaryPath string
+	// Enforce EnableSorobanDiagnosticEvents when not disabled explicitly
+	EnforceSorobanDiagnosticEvents bool
```
sreuland, line 330: would setting EnableSorobanDiagnosticEvents=true in `setDefaults()` achieve the same end result of being enabled in all cases except when explicit set to false w/o additional override flag?

2opremio: Yes, but we don't want to enforce it in all cases, only in soroban-rpc. I don't think we should enforce it in Horizon.

sreuland (approved):

PR #4995, opened 2023-07-28
captive-core: improve core http client handling

tsachiherman:
### What

This PR fixes few minor http-level conditions that were missed when implementing the core http client package.

### Why

This PR addresses the following:
1. Missing status code testing. A success status code is in the range of [200-299]. The existing code for GetLedgerEntry, SetCursor and SubmitTransaction  wasn't validating it.
2. Using `NewRequestWithContext` instead of `NewRequest` + `WithContext` is preferable, as it's more implicit and aligns with the underlying implementation.
3. Missing drainage of response body. In order for the http package to reuse a connection, the response body is required to be fully read. Otherwise, the session would not be reused, and another http session would be established. I've updated all the handlers here to correctly drain the response body before closing it.
4. Removal of the `Preflight` method, which is no longer supported by core.

### Known limitations

* The `ManualClose` is doing some questionable json parsing.. That logic is likely need to be fixed ( outside of the scope of this PR ).


clients/stellarcore/client.go
```diff
@@ -104,42 +86,64 @@ func (c *Client) GetLedgerEntry(ctx context.Context, ledgerKey xdr.LedgerKey) (p
...
 
-	hresp, err := c.http().Do(req)
+	var hresp *http.Response
+	hresp, err = c.http().Do(req)
 	if err != nil {
 		err = errors.Wrap(err, "http request errored")
 		return
 	}
-	defer hresp.Body.Close()
+	defer func() {
```
Shaptic, line 129: seems repeated a few times: should it be a helper?

clients/stellarcore/client.go
```diff
@@ -31,59 +30,42 @@ type Client struct {
...
+		err2 = hresp.Body.Close()
+		if err == nil && err2 != nil {
+			err = errors.Wrap(err2, "unable to close response body")
+		}
+	}()
 
 	if !(hresp.StatusCode >= 200 && hresp.StatusCode < 300) {
-		return errors.New("http request failed with non-200 status code")
+		err = errors.New("http request failed with non-200 status code")
+		return
```
Shaptic, line 64: confused about this change - extra line for no reason?

clients/stellarcore/client.go
```diff
@@ -31,59 +30,42 @@ type Client struct {
...
+	hresp, err = c.http().Do(req)
 	if err != nil {
 		return errors.Wrap(err, "http request errored")
 	}
-	defer hresp.Body.Close()
+	defer func() {
+		// drain the remaining of the Body
+		_, err2 := io.Copy(ioutil.Discard, hresp.Body)
+		if err == nil && err2 != nil {
+			err = errors.Wrap(err2, "unable to read excess data from response")
```
Shaptic, line 54: is this an error we care about? seems "ignorable" from a caller perspective

Shaptic (commented):
Does this mean a "breaking change" in the sense that the RPC will no longer return a JSON response body at all in error cases? Imo we should return "best effort" responses even in error cases, e.g. if I `getLedgerEntries` and it fails, if you can give me **some** of them, that's better than a 400.

clients/stellarcore/client.go
tsachiherman: from functional perspective - yes, that's correct. we can ignore that. But - if we are capable of detecting this issue - we should be able to log/report it somewhere. It does imply that there was a problem, and we shouldn't hide problems..

clients/stellarcore/client.go
tsachiherman: I've just tried to align with `set err first, then return. It's more clean in the context of the defer method which examines the content of `err`.

clients/stellarcore/client.go
tsachiherman: I'm all for unifying the logic, but wanted to make sure we're functional-correct first.

tsachiherman:
> Does this mean a "breaking change" in the sense that the RPC will no longer return a JSON response body at all in error cases? Imo we should return "best effort" responses even in error cases, e.g. if I `getLedgerEntries` and it fails, if you can give me **some** of them, that's better than a 400.

how is this change anyway relates to what the RPC would return ? what am I missing here ?

Shaptic:
@tsachiherman because it looks like instead of reading the body into the response indiscriminately:

```go
	responseBytes, err := io.ReadAll(hresp.Body)
```

we are now yeeting the bytes into oblivion:

```
if !(hresp.StatusCode >= 200 && hresp.StatusCode < 300) {
	// drain the remaining of the Body
	_, err = io.Copy(ioutil.Discard, hresp.Body)
```

in error cases. So this would change the response body in error cases. My question is: were we returning partial responses previously? Or was json unmarshalling failing anyway, so clients never saw a valid json body?

tsachiherman:
> @tsachiherman because it looks like instead of reading the body into the response indiscriminately:
> 
> ```go
> 	responseBytes, err := io.ReadAll(hresp.Body)
> ```
> 
> we are now yeeting the bytes into oblivion:
> 
> ```
> if !(hresp.StatusCode >= 200 && hresp.StatusCode < 300) {
> 	// drain the remaining of the Body
> 	_, err = io.Copy(ioutil.Discard, hresp.Body)
> ```
> 
> in error cases. So this would change the response body in error cases. My question is: were we returning partial responses previously? Or was json unmarshalling failing anyway, so clients never saw a valid json body?

let's divide and conquer, shall we ? you're referring specifically about the `GetLedgerEntry` method. In the case of this method, the core code is here:
https://github.com/stellar/stellar-core/blob/9215e8d7a79f7be7345b1c9d7554e800d9f09f06/src/main/CommandHandler.cpp#L708
As you can see, it would return a valid json in the case of a key found, and a 400 in the case of failure.

We don't want to hit the "failed-to-parse-json" path here. instead, we want to want to detect the error code ( 2xx vs 4xx ) and use that to determine the success/fail. If we don't need the data, I'd rather just toss it away - unless you have any concrete use for it ( we didn't before )

tsachiherman:
> Does this mean a "breaking change" in the sense that the RPC will no longer return a JSON response body at all in error cases? Imo we should return "best effort" responses even in error cases, e.g. if I `getLedgerEntries` and it fails, if you can give me **some** of them, that's better than a 400.

This is not the RPC's `getLedgerEntries`.. it's the core's version. We don't use it at this time anywhere.

Shaptic:
> This is not the RPC's getLedgerEntries..

completely missed that :facepalm: thanks for clarifying!

ingest/ledgerbackend/toml.go
```diff
@@ -516,6 +517,22 @@ func (c *CaptiveCoreToml) setDefaults(params CaptiveCoreTomlParams) {
 	}
 }
 
+// checkTestingWithProtocol19 is a temporary workaround that would let us keep conducting the
+// horizon testings against an older core version without brakage. Note that at this point, we
+// don't know if the core would support the new flags - so we're defering this to an env variable.
+// post release, we should just remove this test.
+func checkTestingAboveProtocol19() bool {
+	str := os.Getenv("HORIZON_INTEGRATION_TESTS_CORE_MAX_SUPPORTED_PROTOCOL")
```
2opremio, line 525: This shouldn't be in production code

clients/stellarcore/client.go
```diff
@@ -30,60 +29,61 @@ type Client struct {
...
 	}
 
-	hresp, err := c.http().Do(req)
+	var hresp *http.Response
+	hresp, err = c.http().Do(req)
 	if err != nil {
 		return errors.Wrap(err, "http request errored")
 	}
-	defer hresp.Body.Close()
+	defer func() {
```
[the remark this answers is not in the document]
tsachiherman: historical reasons.. I've removed the closure.

PR #4999, opened 2023-07-28
Correctly handle LP overflow slippage

Shaptic:
### What
Set maximum slippage to exclude overflow spikes from plots.

### Why
Corrects the hotfix behavior from #4981.


Shaptic:
wrong target branch

PR #5000, opened 2023-07-28
Properly set maximum slippage in case of LP overflows

Shaptic:
### What
Set maximum slippage to exclude overflow spikes from plots.

### Why
Corrects the hotfix behavior from #4981. Original issue documented here: https://status.stellar.org/incidents/gsltshfy60rf


services/horizon/internal/ingest/processors/trades_processor.go
```diff
@@ -260,7 +260,10 @@ func (p *TradeProcessor) roundingSlippage(
 			true,
 		)
 		if !ok {
-			return null.Int{}, errors.New("Liquidity pool overflows from this exchange")
+			// Temporary workaround for https://github.com/stellar/go/issues/4203
+			// Given strict sends that would overflow here, set maximum slippage
+			// so they get excluded.
+			roundingSlippageBips = xdr.Int64(math.MaxInt64)
```
urvisavla, line 266: Should we be logging it to gather data on its frequency of occurrence?

Shaptic: Eh, it happened once in two years :joy:

PR #5002, opened 2023-07-31
core config: update to support multiple versions

tsachiherman:
### What

Previous PR broke multi-version compatibility. This PR brings it back.

### Why

We need to have horizon work seamlessly with both v19 as well as v20, for upgrade purposes.

### Known limitations

n/a


2opremio:
@tsachiherman integration tests are failing

tsachiherman:
> aside from the lint errors, this PR looks good. but we should look into why the integration tests are failing

not sure why you see failing integration tests.. it looks to me like all the tests are passing.

tsachiherman:
The verify-range is still failing, but we'll fix it later on. I want to "unblock" the branch first.

PR #5026, opened 2023-08-23
xdr: Add function to obtain the type of a binary-compressed key 

2opremio:
I need it for soroban-tools

### DIFF
--- a/xdr/main.go
+++ b/xdr/main.go
@@ -232,8 +232,15 @@ func (e *EncodingBuffer) LedgerKeyUnsafeMarshalBinaryCompress(key LedgerKey) ([]
 }
 
 // GetBinaryCompressedLedgerKeyType gets the key type from the result of LedgerKeyUnsafeMarshalBinaryCompress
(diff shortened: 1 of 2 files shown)


xdr/main.go
```diff
@@ -230,6 +231,11 @@ func (e *EncodingBuffer) LedgerKeyUnsafeMarshalBinaryCompress(key LedgerKey) ([]
 	return e.xdrEncoderBuf.Bytes(), nil
 }
 
+// GetBinaryCompressedLedgerKeyType gets the key type from the result of LedgerKeyUnsafeMarshalBinaryCompress
+func GetBinaryCompressedLedgerKeyType(compressedKey []byte) LedgerEntryType {
+	return LedgerEntryType(compressedKey[0])
```
tsachiherman, line 236: in general, seems to be unsafe since len(compressedKey) could be zero. But if you're using it only in a test, I think that it might be ok. ( although it would be better to move that to the `main_test.go` file.

tsachiherman (approved):

PR #5030, opened 2023-08-29
Remove Horizon Submission Queue

aditya1702:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

### Why

### Known limitations

[TODO or N/A]


services/horizon/internal/txsub/system.go
```diff
@@ -134,75 +132,40 @@ func (sys *System) Submit(
...
-			// If error is txBAD_SEQ, check for the result again
-			tx, err = txResultByHash(ctx, db, hash)
-			if err != nil {
-				// finally, return the bad_seq error if no result was found on 2nd attempt
-				sys.finish(ctx, hash, resultCh, Result{Err: sr.Err})
-				return
-			}
-			// If we found the result, use it as the result
-			sys.finish(ctx, hash, resultCh, Result{Transaction: tx})
+		if err = sys.waitUntilAccountSequence(ctx, db, sourceAddress, uint64(envelope.SeqNum())); err != nil {
```
Shaptic, line 150: I know this is a draft, but I want to throw a thought in here: I don't think we should do this anymore, either. For context, this means Horizon allows out-of-order txsub and will wait for the gap to be filled. For example:

* account is at seq N
* account txsubs a tx `T` with seqnum `N + 3`
* horizon will hold `T` for a while until it sees `N+1` and `N+2`
* it submits all of them once the gap is filled

This has caused confusion and problems in the past because the holding behavior is not documented, and, more importantly, **this optimization no longer works**, i.e. you'd need to submit 1 tx/ledger anyway. That means there's no purpose in holding gaps since you can't shotgun them at once, anyway.

Since we're here and simplifying the txsub behavioral contract, we might as well continue to rip out stuff that is no longer useful.

aditya1702: @Shaptic Yes, I removed it now. I thought that this function blocks the transaction endpoint till it is successfully submitted to core. However, that is done by the open submissions list.

PR #5032, opened 2023-08-30
services/horizon: Fixing Claimable Balances Query Limit Issue

### LINKED ISSUE #4907: services/horizon:`limit` flag on `claimable_balances` behaving oddly
Noticed something strange with the limit param while filtering claimable balances. Given https://horizon-testnet.stellar.org/claimable_balances/?asset=RPCIEGE00001%3AGC3DQ2W7BJUF7OFKJTJH3EPGUVR3FRN2A2263PUK6DD2HN4FWW2K3OVE&claimant=GDAYVCINVNUZ57EOCN4FK2VVWGQ3L3NW37L6UJLZCK3C7S7CNSS5EHHG&limit=200 you can see 2 items but when loading https://horizon-testnet.stellar.org/claimable_balances/?asset=RPCIEGE00001%3AGC3DQ2W7BJUF7OFKJTJH3EPGUVR3FRN2A2263PUK6DD2HN4FWW2K3OVE&claimant=GDAYVCINVNUZ57EOCN4FK2VVWGQ3L3NW37L6UJLZCK3C7S7CNSS5EHHG&limit=1 it shows zero

(as of the date of this posting)

urvisavla:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This PR addresses the issue identified in #4907. 

Issue: 
When querying claimable balances with a combination of asset and claimant and setting a limit of 200, the response displayed the items correctly. However, when reducing the limit to 1 in the query, the response showed zero results.

Analysis:
The query to filter claimable balances by claimants is as follows:
```
SELECT cb.id, cb.claimants, cb.asset, cb.amount, cb.sponsor, cb.last_modified_ledger, cb.flags 
FROM claimable_balances cb 
WHERE cb.asset = '?' AND cb.id IN 
   (SELECT id FROM claimable_balance_claimants
   WHERE destination = 'GDB3GJCDWLAMVV7ZL6Q7BH3PGVMFWHA5KSRIY2NML3RXX35ESL3MBOXU' 
   ORDER BY last_modified_ledger ASC, id ASC LIMIT 10) 
ORDER BY last_modified_ledger ASC, id ASC LIMIT 10;
```

In this query, there are no asset or sponsor filters in the subquery because there is no asset or sponsor information in `caimable_balance_claimants` table. Consequently, when the same limit as the outer query is applied to the subquery and an asset or sponsor filter is provided, the balance ids returned by the subquery may not necessarily match the asset type hence returning incorrect (incomplete) result. The same observation was made for the sponsor filter.

Solution:
The solution is to use a different query that uses a JOIN with `claimable_balance_claimants` when asset or sponsor filters are specified:
```
SELECT cb.id, cb.claimants, cb.asset, cb.amount, cb.sponsor, cb.last_modified_ledger, cb.flags 
FROM claimable_balances cb 
JOIN claimable_balance_claimants ON claimable_balance_claimants.id = cb.id 
WHERE claimable_balance_claimants.destination = 'GDB3GJCDWLAMVV7ZL6Q7BH3PGVMFWHA5KSRIY2NML3RXX35ESL3MBOXU' 
AND cb.asset = 'AAAAAUFRVUEAAAAAW5QuU6wzyP0KgMx8GxqF19g4qcQZd6rRizrwV/jjPfA=' 
AND cb.sponsor = 'GDFCYDQOVJ2OEWPLEGIRQVAM3VTOQ6JDNLJTDZP5S5OGTEHM5CIWMYBH' 
ORDER BY cb.last_modified_ledger ASC, cb.id ASC 
LIMIT 10;
```

During testing, it was found that the JOIN query performs poorly when no filters other than the claimant are supplied. Therefore, in such case, we will continue to use the original subquery to retrieve claimable balances by claimants.

Test coverage:
Implemented a unit test to replicate the scenario described in the bug.
 
### Why

Fixes #4907

### Known limitations
N/A

### DIFF
--- a/services/horizon/internal/db2/history/claimable_balances.go
+++ b/services/horizon/internal/db2/history/claimable_balances.go
@@ -291,7 +291,7 @@ func (q *Q) GetClaimableBalances(ctx context.Context, query ClaimableBalancesQue
 
 	var results []ClaimableBalance
 	if err := q.Select(ctx, &results, sql); err != nil {
-		return nil, errors.Wrap(err, fmt.Sprintf("could not run select query"))
+		return nil, errors.Wrap(err, "could not run select query")
 	}
 
 	return results, nil
--- a/services/horizon/internal/db2/history/claimable_balances.go
+++ b/services/horizon/internal/db2/history/claimable_balances.go
@@ -88,10 +88,9 @@ func applyClaimableBalancesQueriesCursor(sql sq.SelectBuilder, lCursor int64, rC
 // ClaimableBalanceClaimant is a row of data from the `claimable_balances_claimants` table.
 // This table exists to allow faster querying for claimable balances for a specific claimant.
 type ClaimableBalanceClaimant struct {
-	BalanceID          string    `db:"id"`
-	Destination        string    `db:"destination"`
-	LastModifiedLedger uint32    `db:"last_modified_ledger"`
-	Asset              xdr.Asset `db:"asset"`
+	BalanceID          string `db:"id"`
+	Destination        string `db:"destination"`
+	LastModifiedLedger uint32 `db:"last_modified_ledger"`
 }
 
 // ClaimableBalance is a row of data from the `claimable_balances` table.
@@ -248,23 +247,30 @@ func (q *Q) GetClaimableBalances(ctx context.Context, query ClaimableBalancesQue
 		return nil, errors.Wrap(err, "could not apply query to page")
 	}
 
-	if query.Asset != nil {
-		// when search by asset, profiling has shown best performance to have the LIMIT on inner query
-		sql = sql.Where("cb.asset = ?", query.Asset)
-	}
+	if query.Asset != nil || query.Sponsor != nil {
 
-	if query.Sponsor != nil {
-		sql = sql.Where("cb.sponsor = ?", query.Sponsor.Address())
-	}
+		// JOIN with claimable_balance_claimants table to query by claimants
+		if query.Claimant != nil {
+			sql = sql.Join("claimable_balance_claimants on claimable_balance_claimants.id = cb.id")
+			sql = sql.Where("claimable_balance_claimants.destination = ?", query.Claimant.Address())
+		}
 
-	if query.Claimant != nil {
+		// Apply filters for asset and sponsor
+		if query.Asset != nil {
+			sql = sql.Where("cb.asset = ?", query.Asset)
+		}
+		if query.Sponsor != nil {
+			sql = sql.Where("cb.sponsor = ?", query.Sponsor.Address())
+		}
+
+	} else if query.Claimant != nil {
+		// If only the claimant is provided without additional filters, a JOIN with claimable_balance_claimants
+		// does not perform efficiently. Instead, use a subquery (with LIMIT) to retrieve claimable balances based on
+		// the claimant's address.
+
 		var selectClaimableBalanceClaimants = sq.Select("id").From("claimable_balance_claimants").
 			Where("destination = ?", query.Claimant.Address())
 
-		// https://github.com/stellar/go/issues/4907
-		if query.Asset != nil {
-			selectClaimableBalanceClaimants = selectClaimableBalanceClaimants.Where("asset = ?", query.Asset)
-		}
 		selectClaimableBalanceClaimants.Limit(query.PageQuery.Limit)
 
 		subSql, err := applyClaimableBalancesQueriesCursor(selectClaimableBalanceClaimants, l, r, query.PageQuery.Order)
@@ -285,7 +291,7 @@ func (q *Q) GetClaimableBalances(ctx context.Context, query ClaimableBalancesQue
 
 	var results []ClaimableBalance
 	if err := q.Select(ctx, &results, sql); err != nil {
-		return nil, errors.Wrap(err, "could not run select query")
+		return nil, errors.Wrap(err, fmt.Sprintf("could not run select query"))
 	}
 
 	return results, nil
--- a/services/horizon/internal/db2/history/claimable_balances.go
+++ b/services/horizon/internal/db2/history/claimable_balances.go
@@ -88,9 +88,10 @@ func applyClaimableBalancesQueriesCursor(sql sq.SelectBuilder, lCursor int64, rC
 // ClaimableBalanceClaimant is a row of data from the `claimable_balances_claimants` table.
 // This table exists to allow faster querying for claimable balances for a specific claimant.
 type ClaimableBalanceClaimant struct {
-	BalanceID          string `db:"id"`
-	Destination        string `db:"destination"`
-	LastModifiedLedger uint32 `db:"last_modified_ledger"`
+	BalanceID          string    `db:"id"`
+	Destination        string    `db:"destination"`
+	LastModifiedLedger uint32    `db:"last_modified_ledger"`
+	Asset              xdr.Asset `db:"asset"`
 }
 
 // ClaimableBalance is a row of data from the `claimable_balances` table.
@@ -261,15 +262,10 @@ func (q *Q) GetClaimableBalances(ctx context.Context, query ClaimableBalancesQue
 			Where("destination = ?", query.Claimant.Address())
 
 		// https://github.com/stellar/go/issues/4907
-		// If a filter by asset is provided, do not add LIMIT to the subquery. This is because
-		// claimable_balance_claimants table cannot be filtered by asset, so when the LIMIT specified is small,
-		// the balance IDs returned by this subquery might be of different asset than the one specified
-		// in the filter. This could lead to an empty result list.
-		if query.Asset == nil {
-			// Given that each destination can be a claimant for each balance maximum once
-			// we can LIMIT the subquery.
-			selectClaimableBalanceClaimants.Limit(query.PageQuery.Limit)
+		if query.Asset != nil {
+			selectClaimableBalanceClaimants = selectClaimableBalanceClaimants.Where("asset = ?", query.Asset)
 		}
+		selectClaimableBalanceClaimants.Limit(query.PageQuery.Limit)
 
 		subSql, err := applyClaimableBalancesQueriesCursor(selectClaimableBalanceClaimants, l, r, query.PageQuery.Order)
 		if err != nil {
--- a/services/horizon/internal/db2/history/claimable_balances.go
+++ b/services/horizon/internal/db2/history/claimable_balances.go
@@ -258,10 +258,7 @@ func (q *Q) GetClaimableBalances(ctx context.Context, query ClaimableBalancesQue
 
 	if query.Claimant != nil {
 		var selectClaimableBalanceClaimants = sq.Select("id").From("claimable_balance_claimants").
-			Where("destination = ?", query.Claimant.Address()).
-			// Given that each destination can be a claimant for each balance maximum once
-			// we can LIMIT the subquery.
-			Limit(query.PageQuery.Limit)
+			Where("destination = ?", query.Claimant.Address())
 		subSql, err := applyClaimableBalancesQueriesCursor(selectClaimableBalanceClaimants, l, r, query.PageQuery.Order)
 		if err != nil {
 			return nil, errors.Wrap(err, "could not apply subquery to page")
--- a/services/horizon/internal/db2/history/claimable_balances.go
+++ b/services/horizon/internal/db2/history/claimable_balances.go
@@ -259,6 +259,18 @@ func (q *Q) GetClaimableBalances(ctx context.Context, query ClaimableBalancesQue
 	if query.Claimant != nil {
 		var selectClaimableBalanceClaimants = sq.Select("id").From("claimable_balance_claimants").
 			Where("destination = ?", query.Claimant.Address())
+
+		// https://github.com/stellar/go/issues/4907
+		// If a filter by asset is provided, do not add LIMIT to the subquery. This is because
+		// claimable_balance_claimants table cannot be filtered by asset, so when the LIMIT specified is small,
+		// the balance IDs returned by this subquery might be of different asset than the one specified
+		// in the filter. This could lead to an empty result list.
+		if query.Asset == nil {
+			// Given that each destination can be a claimant for each balance maximum once
+			// we can LIMIT the subquery.
+			selectClaimableBalanceClaimants.Limit(query.PageQuery.Limit)
+		}
+
 		subSql, err := applyClaimableBalancesQueriesCursor(selectClaimableBalanceClaimants, l, r, query.PageQuery.Order)
 		if err != nil {
 			return nil, errors.Wrap(err, "could not apply subquery to page")
--- a/services/horizon/internal/db2/history/claimable_balances.go
+++ b/services/horizon/internal/db2/history/claimable_balances.go
@@ -269,9 +269,7 @@ func (q *Q) GetClaimableBalances(ctx context.Context, query ClaimableBalancesQue
 		// the claimant's address.
 
 		var selectClaimableBalanceClaimants = sq.Select("id").From("claimable_balance_claimants").
-			Where("destination = ?", query.Claimant.Address())
-
-		selectClaimableBalanceClaimants.Limit(query.PageQuery.Limit)
+			Where("destination = ?", query.Claimant.Address()).Limit(query.PageQuery.Limit)
 
 		subSql, err := applyClaimableBalancesQueriesCursor(selectClaimableBalanceClaimants, l, r, query.PageQuery.Order)
 		if err != nil {
--- a/services/horizon/internal/db2/history/claimable_balances_test.go
+++ b/services/horizon/internal/db2/history/claimable_balances_test.go
@@ -240,7 +240,6 @@ func insertClaimants(q *Q, tt *test.T, cBalance ClaimableBalance) error {
 			BalanceID:          cBalance.BalanceID,
 			Destination:        claimant.Destination,
 			LastModifiedLedger: cBalance.LastModifiedLedger,
-			Asset:              cBalance.Asset,
 		}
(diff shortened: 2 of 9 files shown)


services/horizon/internal/db2/history/claimable_balances.go
```diff
@@ -258,10 +258,7 @@ func (q *Q) GetClaimableBalances(ctx context.Context, query ClaimableBalancesQue
 
 	if query.Claimant != nil {
 		var selectClaimableBalanceClaimants = sq.Select("id").From("claimable_balance_claimants").
-			Where("destination = ?", query.Claimant.Address()).
-			// Given that each destination can be a claimant for each balance maximum once
-			// we can LIMIT the subquery.
-			Limit(query.PageQuery.Limit)
+			Where("destination = ?", query.Claimant.Address())
```
Shaptic, line 261: When this is RFR, can you `git blame` this to find the PR and tag the relevant author/participants? I remember we did a TON of stuff here trying to optimize the query so it'd be great to get the full perspective.

services/horizon/internal/db2/history/claimable_balances.go
```diff
@@ -258,10 +258,19 @@ func (q *Q) GetClaimableBalances(ctx context.Context, query ClaimableBalancesQue
...
+		// https://github.com/stellar/go/issues/4907
+		// If a filter by asset is provided, do not add LIMIT to the subquery. This is because
+		// claimable_balance_claimants table cannot be filtered by asset, so when the LIMIT specified is small,
+		// the balance IDs returned by this subquery might be of different asset than the one specified
+		// in the filter. This could lead to an empty result list.
+		if query.Asset == nil {
 			// Given that each destination can be a claimant for each balance maximum once
 			// we can LIMIT the subquery.
-			Limit(query.PageQuery.Limit)
+			selectClaimableBalanceClaimants.Limit(query.PageQuery.Limit)
```
Shaptic, line 271: This solves the bug directly, of course, but there are definitely performance implications. Obviously, our behavior is incorrect before this PR, but I'm wondering if there's some balance we need to strike. 

* **Would it be better to add SOME limit to the query?** There are a **lot** of claimants in the DB, iirc. What are the chances that we can do an arbitrary e.g. `Limit(500)` in the `else` case and still fill out the result list? 
* **Maybe we should paginate the claimants, instead?** In other words, set an arbitrary limit here, then keep querying until we fill out the overall `limit=<N>` param? Maybe that would be slower, as we'd need to do a sort in order to paginate.
* **Maybe we just modify the table to allow filtering by asset?** But this would involve a very computationally-intensive reingestion, afaict, since we'd need to parse out the claimants.

sydneynotthecity: +1 to adding an arbitrary limit. Or at least performance testing the complete removal of the limit. The original query did not include a limit, which was what caused the original crappy query performance (some queries would take 200+ seconds to return) [This gist](https://gist.github.com/Shaptic/a91d0babc99aa022f3ca86461f845e30) gave an example of one of the long running queries, might be worthwhile to test with these specific parms.

urvisavla:
I executed the query against both the staging pubnet database, experimenting with different limit values. The parameters I selected were:

1. The most frequently occurring claimants from the `claimable_balance_claimants` table, `GDB3GJCDWLAMVV7ZL6Q7BH3PGVMFWHA5KSRIY2NML3RXX35ESL3MBOXU` which 1368944 occurrences.

2. The asset Id with the highest number of occurrences in the `claimable_balances` table, `AAAAAUFRVUEAAAAAW5QuU6wzyP0KgMx8GxqF19g4qcQZd6rRizrwV/jjPfA=` with 1397329 occurrences.

Here are the various scenarios I tested:

1. Limit 10 for both the main and subquery (this is the current default when no limit is specified):
    Execution Time: 2.354 ms

2. No limit on the subquery with a main query limit of 10 (proposed fix):
   Execution Time: 2572.291 ms (It's clear that some kind of limit is needed)

3. Arbitrarily set a limit of 1000 on the inner query with an outer query limit of 10:
   - Execution Time: 26.086 ms (This limit works for 99.9% of claimants in the current dataset)

4. No limit on either the main or subquery:
   - Execution Time: 5840.723 ms

I think the correct approach to address this issue would involve adding asset information to the claimants' table. However, as others have noted, this will require schema migration and data reingestion for the claimants table. Alternatively, following the suggestion of others, implementing an arbitrary limit should work well in most of cases. While it doesn't fix the bug, it significantly reduces the chances of running into it. 
I think we need to make a decision on how critical it is for the query to always work correctly versus mostly working correctly. And if we want to temporarily raise the limit to an arbitrary number as stopgap solution.

Shaptic:
Awesome analysis @urvisavla! This gives a really good picture.

> Arbitrarily set a limit of 1000 on the inner query with an outer query limit of 10:
> Execution Time: 26.086 ms (This limit works for 99.9% of claimants in the current dataset)
> 
> While it doesn't fix the bug, it significantly reduces the chances of running into it.

This seems the most reasonable to me. Yes it's not perfect, but it's probably more than good enough to immediately help people affected by this. Is the 0.1% a random chance, or is there a particular type of query that would continue to trigger this bug? And how certain is that %? Or do you just colloquially mean "really unlikely it'll fail"?

urvisavla:
> > Arbitrarily set a limit of 1000 on the inner query with an outer query limit of 10:
> > Execution Time: 26.086 ms (This limit works for 99.9% of claimants in the current dataset)
> > While it doesn't fix the bug, it significantly reduces the chances of running into it.
> 
> This seems the most reasonable to me. Yes it's not perfect, but it's probably more than good enough to immediately help people affected by this. Is the 0.1% a random chance, or is there a particular type of query that would continue to trigger this bug? And how certain is that %? Or do you just colloquially mean "really unlikely it'll fail"?


Currently, there are 615539 unique claimants, out of which 90 have more than 1000 entries in the claimants table, accounting for approximately 0.14% of the total claimants. When executing the query with a specific asset filter for these claimants when the limit is set to 1000, there is a possibility that the query may return incomplete result if all (1000) balance Ids returned by the subquery are for a different asset than the one the user has requested.

services/horizon/internal/db2/schema/migrations/65_add_asset_in_claimable_balance_claimants.sql
```diff
@@ -0,0 +1,9 @@
+-- +migrate Up
+
+ALTER TABLE claimable_balance_claimants ADD asset text;
+CREATE INDEX "claimants_by_asset_destination_last_modified_ledger_id" ON claimable_balance_claimants USING btree (asset, destination, last_modified_ledger, id);
```
Shaptic, line 4: can you see how big this index would be on staging? I imagine it'd be ginormous... can we get away with any less fields in the btree vs. performance? e.g. does `last_modified_ledger` really matter?

services/horizon/internal/db2/history/claimable_balances.go
```diff
@@ -88,9 +88,10 @@ func applyClaimableBalancesQueriesCursor(sql sq.SelectBuilder, lCursor int64, rC
 // ClaimableBalanceClaimant is a row of data from the `claimable_balances_claimants` table.
 // This table exists to allow faster querying for claimable balances for a specific claimant.
 type ClaimableBalanceClaimant struct {
-	BalanceID          string `db:"id"`
-	Destination        string `db:"destination"`
-	LastModifiedLedger uint32 `db:"last_modified_ledger"`
+	BalanceID          string    `db:"id"`
+	Destination        string    `db:"destination"`
+	LastModifiedLedger uint32    `db:"last_modified_ledger"`
+	Asset              xdr.Asset `db:"asset"`
```
Shaptic, line 94: I'm very surprised you can just use the raw xdr object here... does it automatically serialize it to a string in the [canonical SEP-11 format](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0011.md#alphanum4-alphanum12)? (i.e. `native` or `code:issuer`)

services/horizon/internal/db2/schema/migrations/65_add_asset_in_claimable_balance_claimants.sql
tamirms, line 4: how does the query perform when there is no asset filter?

also, I noticed that in migration 62_claimable_balance_claimants.sql there is already an existing index:

```sql
CREATE INDEX "claimable_balance_claimants_by_destination_last_modified_ledger_id" ON claimable_balance_claimants USING btree (destination, last_modified_ledger, id);
```

do we need both of these indexes?

services/horizon/internal/db2/history/claimable_balances.go
```diff
@@ -258,10 +259,14 @@ func (q *Q) GetClaimableBalances(ctx context.Context, query ClaimableBalancesQue
 
 	if query.Claimant != nil {
```
tamirms, line 260: I noticed there is also a `sponsor` filter on the query that is included on the outer query but not a part of the claimable_balance_claimants table. Is there bug similar to https://github.com/stellar/go/issues/4907 which can be triggered when the `sponsor` filter is present?

services/horizon/internal/db2/schema/migrations/65_add_asset_in_claimable_balance_claimants.sql
urvisavla: There are currently two indices on the table:
- `claimable_balance_claimants_pkey`: 3.1GB
- `claimable_balance_claimants_by_destination_last_modified_ledger`: 3.6GB

The new index `claimants_by_asset_destination_last_modified_ledger_id` adds another 5GB, making the total of 12GB. This makes it one of the largest "state" tables, second only to `trust_line` which is 15GB.

I experimented with the combinations of different indices for the two queries and found that:
Using all three indices provides the best performance with the query taking ~1 ms.
The second best option is to use no additional index, with the existing two index it takes  < 10 ms. 
All other combinations I tried took > 500 ms.  

So not adding the new index seems to be the most efficient option both in terms of space and time.

Query 1
```
   SELECT cb.id, cb.claimants, cb.asset, cb.amount, cb.sponsor, cb.last_modified_ledger, cb.flags
   FROM claimable_balances cb
   WHERE cb.asset = 'AAAAAUFRVUEAAAAAW5QuU6wzyP0KgMx8GxqF19g4qcQZd6rRizrwV/jjPfA='
     AND cb.id IN (
       SELECT id
       FROM claimable_balance_claimants
       WHERE destination = 'GDB3GJCDWLAMVV7ZL6Q7BH3PGVMFWHA5KSRIY2NML3RXX35ESL3MBOXU'
       ORDER BY last_modified_ledger ASC, id ASC
       LIMIT 10
     )
   ORDER BY last_modified_ledger ASC, id ASC
   LIMIT 10;
   ```

Query 2
   ```
   SELECT cb.id, cb.claimants, cb.asset, cb.amount, cb.sponsor, cb.last_modified_ledger, cb.flags
   FROM claimable_balances cb
   WHERE cb.asset = 'AAAAAUFRVUEAAAAAW5QuU6wzyP0KgMx8GxqF19g4qcQZd6rRizrwV/jjPfA='
     AND cb.id IN (
       SELECT id
       FROM claimable_balance_claimants
       WHERE asset = 'AAAAAUFRVUEAAAAAW5QuU6wzyP0KgMx8GxqF19g4qcQZd6rRizrwV/jjPfA='
         AND destination = 'GDB3GJCDWLAMVV7ZL6Q7BH3PGVMFWHA5KSRIY2NML3RXX35ESL3MBOXU'
       ORDER BY last_modified_ledger ASC, id ASC
       LIMIT 10
     )
   ORDER BY last_modified_ledger ASC, id ASC
   LIMIT 10;
   ```

Comment (urvisavla) [reply services/horizon/internal/db2/schema/migrations/65_add_asset_in_claimable_balance_claimants.sql:4]: No we do not. see the comment above https://github.com/stellar/go/pull/5032#discussion_r1326952718

File: services/horizon/internal/db2/history/claimable_balances.go (hunk 4 above)
Comment (tamirms) [reply services/horizon/internal/db2/history/claimable_balances.go:94]: the xdr.Asset struct implements the sql.Scanner and sql.Valuer interfaces:

https://github.com/stellar/go/blob/master/xdr/db.go#L44-L52

that is how the assets are getting serialized into strings in SQL queries

File: services/horizon/internal/db2/history/claimable_balances.go (hunk 5 above)
Comment (urvisavla) [reply services/horizon/internal/db2/history/claimable_balances.go:260]: Good catch. Yes, the same bug will occur when `sponsor` filter is used. And If we were to also include `sponsor` in the `claimable_balance_claimants` table then we don't need to query the `claimable_balances` table (so no outer query) when filtered by `claimant`. Although this additional denormalization fixes the issue may impact the write performance? Do you have any other suggestions or alternatives?

Comment (tamirms) [reply services/horizon/internal/db2/history/claimable_balances.go:260]: > Although this additional denormalization fixes the issue may impact the write performance? 

I think it's worth trying out and measuring the impact of the performance.

Not every claimable balance is sponsored. If you count the number of claimable balance rows that have a non-empty sponsor and it turns out that only a minority of claimable balances are sponsored, we could make the sponsor column on claimable_balance_claimants nullable. I think that would reduce disk space and perhaps reduce the impact of write performance.

Comment (urvisavla) [reply services/horizon/internal/db2/history/claimable_balances.go:260]: It appears that all entries in the `claimable_balances` have the `sponsor` column populated.

Comment (tamirms) [reply services/horizon/internal/db2/history/claimable_balances.go:260]: I just thought of another idea you could try. Instead of using a subquery to filter by destination you could instead join claimable_balances with claimable_balance_claimants and add filters in the where clause:

```sql
SELECT cb.id, cb.claimants, cb.asset, cb.amount, cb.sponsor, cb.last_modified_ledger, cb.flags
FROM claimable_balances cb
JOIN claimable_balance_claimants ON claimable_balance_claimants.id = cb.id
     WHERE claimable_balance_claimants.destination = 'GDB3GJCDWLAMVV7ZL6Q7BH3PGVMFWHA5KSRIY2NML3RXX35ESL3MBOXU'
     AND cb.asset = 'AAAAAUFRVUEAAAAAW5QuU6wzyP0KgMx8GxqF19g4qcQZd6rRizrwV/jjPfA='
ORDER BY cb.last_modified_ledger asc, cb.id asc
LIMIT 10
```

This approach wouldn't require us to add extra columns to the claimable_balance_claimants table. But this query needs to be tested in staging. You might need to tinker with adding extra claimable_balances indexes to make the claimable balances query fast for all combinations of filters (sponsor, asset, destination).

urvisavla: That's a great idea! When I first tried using an `JOIN` for the `claimable_balance_claimants` table to fix the original issue, it was very slow, taking more than 10 seconds, so I abandoned the idea. But when I ran the query you provided with asset and/or sponsor, it worked much faster,  less than 50 ms. 

So, I suggest we go with the `JOIN` when asset or sponsor filters are used. But if no filters other than the claimants are there, we can stick to the subquery with a limit. It still gives us the right results in that case and is much quicker.

As you said, by using this approach, we can avoid adding extra columns that would increase the table size. Plus, we'll maintain good performance.

services/horizon/internal/db2/history/claimable_balances.go
```diff
@@ -247,21 +247,32 @@ func (q *Q) GetClaimableBalances(ctx context.Context, query ClaimableBalancesQue
...
 
-	if query.Claimant != nil {
 		var selectClaimableBalanceClaimants = sq.Select("id").From("claimable_balance_claimants").
-			Where("destination = ?", query.Claimant.Address()).
-			// Given that each destination can be a claimant for each balance maximum once
-			// we can LIMIT the subquery.
-			Limit(query.PageQuery.Limit)
+			Where("destination = ?", query.Claimant.Address())
+
+		selectClaimableBalanceClaimants.Limit(query.PageQuery.Limit)
```
tamirms, line 274: should `selectClaimableBalanceClaimants.Limit(query.PageQuery.Limit)` be assigned to `selectClaimableBalanceClaimants`? I don't think `Limit(query.PageQuery.Limit)` will not modify `selectClaimableBalanceClaimants`

services/horizon/internal/db2/history/claimable_balances.go
```diff
@@ -280,7 +291,7 @@ func (q *Q) GetClaimableBalances(ctx context.Context, query ClaimableBalancesQue
 
 	var results []ClaimableBalance
 	if err := q.Select(ctx, &results, sql); err != nil {
-		return nil, errors.Wrap(err, "could not run select query")
+		return nil, errors.Wrap(err, fmt.Sprintf("could not run select query"))
```
tamirms, line 294: S1039: unnecessary use of fmt.Sprintf (gosimple)

services/horizon/internal/db2/history/claimable_balances_test.go
```diff
@@ -233,6 +233,177 @@ func TestFindClaimableBalancesByDestination(t *testing.T) {
...
+		}
+		err := claimantsInsertBuilder.Add(tt.Ctx, claimant)
+		if err != nil {
+			return err
+		}
+	}
+	return claimantsInsertBuilder.Exec(tt.Ctx)
+}
+
+func validateClaimableBalanceQuery(t *test.T, q *Q, query ClaimableBalancesQuery, expectedLen int, expectedClaimants []string, expectedAsset string, expectedSponsor string) {
```
tamirms, line 252: this function will not validate `expectedClaimants` properly if the length is greater than 2

services/horizon/internal/db2/history/claimable_balances.go
urvisavla: fixed in 40d3587

services/horizon/internal/db2/history/claimable_balances_test.go
urvisavla: fixed in https://github.com/stellar/go/commit/40d35871cd0a9e7164528ccbab6fe68c1ecf8c77

Issue #5031, opened 2023-08-30
Eviction of Soroban ledger entries can lead to Horizon having invalid asset stats


tamirms:
Horizon ingests two types of contract data ledger entries from the Stellar Asset Contract:

- [AssetInfo](https://github.com/stellar/rs-soroban-env/blob/main/soroban-env-host/src/native_contract/token/contract.rs#L141) which is used by horizon to map contract ids to Stellar Assets
- [BalanceValue](https://github.com/stellar/rs-soroban-env/blob/main/soroban-env-host/src/native_contract/token/balance.rs#L61) which is used by horizon to keep track of how many contracts hold a given Stellar Asset and the total balance of that asset held by all contracts

Horizon will only ingest [BalanceValue](https://github.com/stellar/rs-soroban-env/blob/main/soroban-env-host/src/native_contract/token/balance.rs#L61) ledger entries from contracts which have an [AssetInfo](https://github.com/stellar/rs-soroban-env/blob/main/soroban-env-host/src/native_contract/token/contract.rs#L141) ledger entry. But, it is possible for a Stellar asset contract to have an evicted [AssetInfo](https://github.com/stellar/rs-soroban-env/blob/main/soroban-env-host/src/native_contract/token/contract.rs#L141) ledger entry while also having live [BalanceValue](https://github.com/stellar/rs-soroban-env/blob/main/soroban-env-host/src/native_contract/token/balance.rs#L61) ledger entries.

In that scenario, a Horizon instance starting from an empty DB will ignore all live [BalanceValue](https://github.com/stellar/rs-soroban-env/blob/main/soroban-env-host/src/native_contract/token/balance.rs#L61) ledger entries from that contract because the [AssetInfo](https://github.com/stellar/rs-soroban-env/blob/main/soroban-env-host/src/native_contract/token/contract.rs#L141) ledger entry is evicted and therefore not present in the ledger. If the  [AssetInfo](https://github.com/stellar/rs-soroban-env/blob/main/soroban-env-host/src/native_contract/token/contract.rs#L141) ledger entry was restored, then Horizon would ingest it, but it would not ingest the pre-existing live [BalanceValue](https://github.com/stellar/rs-soroban-env/blob/main/soroban-env-host/src/native_contract/token/balance.rs#L61) ledger entries. So, in that case the Horizon DB would not have an accurate count of contract holders and state verification would fail.

Here's a concrete example:

[...]

PR #5033, opened 2023-08-31
services/horizon/internal/ingest/processors: Accommodate eviction of soroban ledger entries in asset stats endpoint

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

https://github.com/stellar/go/issues/5031 describes a scenario where the asset stats endpoint will return incorrect data for contract holders when the ledger entries which store asset balances outlives the asset contract instance storage.

To fix this issue we will ingest all [`BalanceValue`](https://github.com/stellar/rs-soroban-env/blob/main/soroban-env-host/src/native_contract/token/balance.rs#L61) ledger entries and compute the number of contract holders and the total amount held by all contracts. These stats will be stored in a new table:

```sql
CREATE TABLE contract_asset_stats (
     -- contract id of the asset contract
     contract_id BYTEA PRIMARY KEY ,
     -- stat contains the number of contracts which hold this asset and
     -- the total amount held by all the holders
     stat       JSONB NOT NULL
);
```

We will still ingest [`AssetInfo`](https://github.com/stellar/rs-soroban-env/blob/main/soroban-env-host/src/native_contract/token/contract.rs#L141) ledger entries so we can map stellar classic assets to their corresponding soroban contracts.

In the asset stats endpoint we will fetch Stellar classic asset stats and, if we have a mapping from the Stellar classic asset to its soroban contract, we will join the `contract_asset_stats` table to fetch stats on contract holders. If there is no matching row in `contract_asset_stats` we will assume there are no contract holders.

With this fix, if an asset contract is evicted but still has live [`BalanceValue`](https://github.com/stellar/rs-soroban-env/blob/main/soroban-env-host/src/native_contract/token/balance.rs#L61) ledger entries, the asset stats endpoint will omit all the contract holders. However, once the asset contract is restored, the asset stats endpoint will correctly include all the contract holders.

### Why

See https://github.com/stellar/go/issues/5031

### Known limitations

It is possible for Horizon to ingest [`BalanceValue`](https://github.com/stellar/rs-soroban-env/blob/main/soroban-env-host/src/native_contract/token/balance.rs#L61) ledger entries which do not belong to real Stellar asset contracts. The stats for such fake asset contracts will never be included in the asset stats endpoint response, but, they will still occupy space in the  `contract_asset_stats` table. There will be 1 row in `contract_asset_stats` for each "fake" asset contract.


Shaptic (commented):
Mostly nitpicks with a couple of deeper questions.

I know I missed the discussion of this issue earlier in the week, but I'm wondering if there's another approach. I added a comment on the issue about it; it's probably better to discuss there: https://github.com/stellar/go/issues/5031#issuecomment-1701721962.

services/horizon/internal/ingest/processors/asset_stats_set.go
```diff
@@ -478,7 +479,8 @@ func (s AssetStatSet) ingestAssetContractBalance(change ingest.Change) {
 	// there was no balance
 	stats, ok := s.contractAssetStats[*pContractID]
 	if !ok {
-		stats = contractAssetStatValue{
+		stats = assetContractStatValue{
+			contractID: *pContractID,
```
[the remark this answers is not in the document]
tamirms: > does this deref create a copy in stats.contractID?
yes, because the contract id has a type of [32]byte. If it were a normal byte slice, dereferencing it would not make a copy

tamirms:
I have updated this PR to fix https://github.com/stellar/go/issues/5071 ( Horizon does not account for expired contract balances in asset stats endpoint). Now we store all asset balances held by contracts in the `contract_asset_balances` table:

```sql
CREATE TABLE contract_asset_balances (
     key_hash BYTEA PRIMARY KEY,
     asset_contract_id BYTEA NOT NULL,
     amount       numeric(39,0) NOT NULL, -- 39 digits is sufficient for a 128 bit integer
     expiration_ledger integer NOT NULL
);

CREATE INDEX "contract_asset_balances_by_expiration" ON contract_asset_balances USING btree (expiration_ledger);
```
When ingesting contract asset balances count how many balances are active vs archived (expired). We make sure to update these statistics when balances expire or are restored.

I recommend reviewing the PR in the following order for a better understanding of the code:

- First look at the schema changes in the migrations to understand the data model
- Then, review the db functions in services/horizon/internal/db2/history/asset_stats.go
- Finally, review the asset stats processor changes

tamirms:
Unfortunately I was not able to add any integration tests which cover the case where an asset balance is expired / restored because the stellar asset contract configures the lifetime of asset balances to span a month's worth of ledgers:

https://github.com/stellar/rs-soroban-env/blob/main/soroban-env-host/src/builtin_contracts/stellar_asset_contract/balance.rs#L42-L47

even with accelerated ledger close times, it would take too long for an asset balance to expire. So I have only tested the functionality through unit tests.

2opremio:
Maybe we should ask core to provide a way to test this more easily?

services/horizon/internal/db2/history/asset_stats.go
```diff
@@ -53,6 +63,181 @@ func (q *Q) InsertAssetStats(ctx context.Context, assetStats []ExpAssetStat, bat
...
+		return nil
+	}
+	for len(keys) > 0 {
+		var args []interface{}
+		var values []string
+
+		for i := 0; len(keys) > 0 && (i+1)*2 < maxUpdateBatchSize; i++ {
+			args = append(args, keys[0][:], amounts[0])
+			values = append(values, "(cast(? as bytea), cast(? as numeric))")
+			keys = keys[1:]
```
urvisavla, line 137: Wondering why remove a key from the slice vs incrementing the index as the first will incur copy overhead? This question is more in the context of go programming.

services/horizon/internal/db2/history/asset_stats.go
```diff
@@ -53,6 +63,181 @@ func (q *Q) InsertAssetStats(ctx context.Context, assetStats []ExpAssetStat, bat
...
+// (if they exist in the db).
+func (q *Q) UpdateContractAssetBalanceAmounts(ctx context.Context, keys []xdr.Hash, amounts []string) error {
+	if len(keys) == 0 {
+		return nil
+	}
+	for len(keys) > 0 {
+		var args []interface{}
+		var values []string
+
+		for i := 0; len(keys) > 0 && (i+1)*2 < maxUpdateBatchSize; i++ {
```
urvisavla, line 134: Why is this limited half of maxUpdateBatchSize?

services/horizon/internal/db2/history/asset_stats.go
```diff
@@ -33,11 +34,20 @@ func assetStatToPrimaryKeyMap(assetStat ExpAssetStat) map[string]interface{} {
 	}
 }
 
+// ContractStatRow represents a row in the contract_asset_stats table
+type ContractStatRow struct {
```
sreuland, line 38: minor, for consideration on naming consistency, `ContractAssetStatRow`, and ` func (q *Q) InsertContractAssetStats` on line 67, to align with `contract_asset_stats`

services/horizon/internal/ingest/processors/asset_stats_processor.go
```diff
@@ -13,116 +14,304 @@ import (
...
-	default:
-		return nil
+	// only ingest contract data entries which could be relevant to
+	// asset stats
+	if change.Type == xdr.LedgerEntryTypeContractData {
+		ledgerEntry := change.Post
+		if ledgerEntry == nil {
+			ledgerEntry = change.Pre
+		}
+		asset := AssetFromContractData(*ledgerEntry, p.networkPassphrase)
```
sreuland, line 62: how about `assetContractBalanceFound` instead of `ok`

services/horizon/internal/ingest/processors/asset_stats_processor.go
```diff
@@ -13,116 +14,304 @@ import (
...
+		_, _, ok := ContractBalanceFromContractData(*ledgerEntry, p.networkPassphrase)
```
sreuland, line 63: how about `assetContractBalanceFound` instead of `ok`

services/horizon/internal/db2/history/asset_stats.go
tamirms: `keys = keys[1:]` does not actually create a copy. A slice in Go is essentially just a pointer into a specific location in an array. So `keys = keys[1:]` will just advance the pointer in an O(1) operation.

This might be useful in explaining how slices workin Go:

https://stackoverflow.com/questions/39993688/are-slices-passed-by-value

services/horizon/internal/db2/history/asset_stats.go
tamirms: postgres has allows a maximum of 65,534 placeholder arguments to a sql statement. so my intention was to limit how big `args` can grow so that it falls below the 65,534 limit. I'll try to rewrite this code to make it less confusing.

services/horizon/internal/ingest/processors/asset_stats_processor.go
```diff
@@ -13,116 +14,304 @@ import (
...
+
+	return p.updateDB(ctx, assetStatSet, contractAssetStatSet)
+}
+
+func (p *AssetStatsProcessor) updateDB(
+	ctx context.Context,
+	assetStatSet AssetStatSet,
+	contractAssetStatSet *ContractAssetStatSet,
+) error {
+	if p.ingestFromHistoryArchive {
```
sreuland, line 156: can a brief comment be added to state why special path when ledger is from archive, it appears like it is specific to creation of asset stats only.

services/horizon/internal/ingest/processors/contract_asset_stats.go
```diff
@@ -0,0 +1,395 @@
...
+		contractID := *pContractID
+		if change.Post == nil {
+			s.contractToAsset[contractID] = nil
+			return nil
+		}
+		// The contract id for a stellar asset should never change and
+		// therefore we return a fatal ingestion error if we encounter
+		// a stellar asset changing contract ids.
+		postAsset := AssetFromContractData(*change.Post, s.networkPassphrase)
+		if postAsset == nil || !(*postAsset).Equals(*asset) {
```
sreuland, line 122: should potential forged entries be cordoned, i.e. `postAsset == nil`, any approach to avoid ingestion halt? if this ever happened, what would need to happen for horizon forward ingestion to be able to proceed?

tamirms: I updated the comment "The contract id for any soroban contract should never change..." So it is not possible to even have a forged entry which changes contract id

services/horizon/internal/ingest/processors/asset_stats_set_test.go
```diff
@@ -44,10 +46,15 @@ func assertAssetStatsAreEqual(t *testing.T, all []history.ExpAssetStat, expected
...
 }
 
-func assertAllFromSnapshotEquals(t *testing.T, set AssetStatSet, expected []history.ExpAssetStat) {
-	all, err := set.AllFromSnapshot()
-	assert.NoError(t, err)
-	assertAssetStatsAreEqual(t, all, expected)
+func assertAssetContractStatsAreEqual(t *testing.T, expected, got []history.ContractStatRow) {
+	assert.Len(t, got, len(expected))
+	sort.Slice(expected, func(i, j int) bool {
+		return hex.EncodeToString(expected[i].ContractID) < hex.EncodeToString(expected[j].ContractID)
```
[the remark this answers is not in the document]
tamirms: storing the contract id instead of a strkey is marginally better in terms of efficiency:

https://stackoverflow.com/questions/48208050/encoding-byte-data-and-storing-as-text-vs-storing-in-bytea-in-postgresql

but I don't think the difference is significant.

However, I don't see the benefit of changing the representation to strkey at this point. If, for example, storing the strkey representation would simplify the ingestion code I would be on board with making the change. But, I don't think the DB representation of contract ids improves the code one way or another.

services/horizon/internal/db2/history/main.go
```diff
@@ -364,6 +364,44 @@ type Asset struct {
 	Issuer string `db:"asset_issuer"`
 }
 
+type ContractStat struct {
+	Balance string `json:"balance"`
+	Holders int32  `json:"holders"`
```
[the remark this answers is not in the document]
tamirms: XLM is excluded from asset stats and I think we're a long way off from having the number of holders overflow a 32 bit integer for an issued asset.

I'm hesitant to make this change because the asset stats endpoint already represents account holders with an int32 field and it would be a breaking change to change the type to an int64 / json string.

services/horizon/internal/db2/history/asset_stats.go
```diff
@@ -158,8 +226,10 @@ func parseAssetStatsCursor(cursor string) (string, string, error) {
 }
 
 // GetAssetStats returns a page of exp_asset_stats rows.
-func (q *Q) GetAssetStats(ctx context.Context, assetCode, assetIssuer string, page db2.PageQuery) ([]ExpAssetStat, error) {
-	sql := selectAssetStats
+func (q *Q) GetAssetStats(ctx context.Context, assetCode, assetIssuer string, page db2.PageQuery) ([]AssetAndContractStat, error) {
+	sql := sq.Select("exp_asset_stats.*, contract_asset_stats.stat as contracts").
+		From("exp_asset_stats").
```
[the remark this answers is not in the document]
tamirms: yeah, I think it would be better done in another PR. There are a lot of changes in this one and I would like to limit the scope to make the change set less risky

sreuland (approved):
lgtm, thanks for addressing feedback!

urvisavla (approved):
LGTM!  However, the VerifyRange integration test is failing. Not sure if that needs to be addressed.

tamirms:
> However, the VerifyRange integration test is failing. Not sure if that needs to be addressed.

This issue was fixed by @sreuland in the master branch.

PR #5034, opened 2023-09-01
all: add support for new expiration ledger entry

2opremio:
As usual, the integration tests won't work until we update soroban-rpc (see https://github.com/stellar/soroban-tools/pull/913 )


tsachiherman (approved):
changes in this PR looks good.

PR #5039, opened 2023-09-05
services/horizon: Remove Horizon Submission Queue

### LINKED ISSUE #4996: services/horizon: remove internal transaction submission queue

Horizon manages an internal [submission queue](https://github.com/stellar/go/blob/master/services/horizon/internal/txsub/sequence/manager.go). This intent of this is to protect/isolate clients from potentially receiving error messages on transaction submission when multiple transactions are submitted very quickly in succession. If transactions (from the same source account) are submitted with sequence numbers out-of-order, then this submission queue will hold those transactions in-memory so that it can re-order them in line with their sequence numbers, allowing all of them to succeed when they are actually submitted to core. This behavior has limited benefit:
- It is unclear which end-users or clients that this use case supports
- The queue is in-memory, so the benefit it provides is not realized on load-balanced or horizontally scaled instances (ex. horizon.stellar.rg)
- This behavior is confusing to end-users, and obfuscates underlying network behavior from the end user (ie. you must submit sequential sequence numbers)
- Partners have been surprised by this behavior, causing outages (ie. Circle's bulk transaction submission incident)
- Submitting out-of-order increases the probability of txsub requests 504-ing, in which case clients need to poll Horizon to validate their transaction was submitted

Recently, core has [released a change](https://www.stellar.org/developers-blog/proposed-changes-to-transaction-submission) that prevents the submission of multiple transactions in a ledger for a given account. With this change, the Horizon submission queue is even less likely to provide any benefit to end-users, and if anything cause more confusion. Transactions would be queued up in-memory, only to have only one of them be able to be submitted when they were finally submitted to core.

We should remove the transaction submission queue entirely, and simply pass-through transaction submission requests directly to core. This has the potential to result in more `tx_bad_seq` errors for clients that were relying on this behavior. However, it drastically simplifies the interface, makes the underlying network behavior clearer, and reduces user confusion that could have been exacerbated by the recent core update.

Not an alternative, but a related issue that further simplifies the developer experience of transaction submission would be [allowing submission to be async](https://github.com/stellar/go/issues/4082). With the submission queue fully removed, transaction submission should always be able to respond immediately with the knowledge that the transaction was at least accepted/validated by core. Note that there is still the possibility that it could not be included in a ledger, so clients would still need to poll Horizon to be sure.

aditya1702:
Closes #4996 

<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Horizon manages an internal [submission queue](https://github.com/stellar/go/blob/master/services/horizon/internal/txsub/sequence/manager.go). This intent of this is to protect/isolate clients from potentially receiving error messages on transaction submission when multiple transactions are submitted very quickly in succession. If transactions (from the same source account) are submitted with sequence numbers out-of-order, then this submission queue will hold those transactions in-memory so that it can re-order them in line with their sequence numbers, allowing all of them to succeed when they are actually submitted to core

This PR: 
- removes this queue and routes transactions directly to core for submission.
- removes the tests related to the queue.

### Why

Recently, core has [released a change](https://www.stellar.org/developers-blog/proposed-changes-to-transaction-submission) that prevents the submission of multiple transactions in a ledger for a given account. With this change, the Horizon submission queue is even less likely to provide any benefit to end-users, and if anything cause more confusion. Transactions would be queued up in-memory, only to have only one of them be able to be submitted when they were finally submitted to core.

### Known limitations

N/A


services/horizon/internal/txsub/system.go
```diff
@@ -134,118 +132,36 @@ func (sys *System) Submit(
...
-		if err == sequence.ErrBadSequence {
-			// convert the internal only ErrBadSequence into the FailedTransactionError
-			err = ErrBadSequence
-		}
+	sr := sys.submitOnce(ctx, rawTx)
+	sys.updateTransactionTypeMetrics(envelope)
 
+	if sr.Err != nil {
+		// any error other than "txBAD_SEQ" is a failure
+		isBad, err := sr.IsBadSeq()
```
Shaptic, line 140: Does this still need to be checked? I think we should be able to pass `txBAD_SEQ` errors as well, since there's no reason to try to hold onto them anymore. 

In general a short comment explaining the _why_ behind these lines of code would be great (e.g. `// add transaction to open list` is obvious from the code, but why are we doing that?)!

services/horizon/internal/txsub/system_test.go
```diff
@@ -148,73 +146,6 @@ func (suite *SystemTestSuite) TestSubmit_Basic() {
 	assert.False(suite.T(), suite.submitter.WasSubmittedTo)
 }
 
-func (suite *SystemTestSuite) TestTimeoutDuringSequnceLoop() {
```
Shaptic, line 151: Removing these tests is nice as it indicates these behaviors/features should not be supported anymore, but should we also add explicit "to-fail" test coverage? e.g. that out of order submissions aren't supported. Maybe this is already covered outside of this diff; I'm not super in-the-know about the existing test coverage.

Shaptic (commented):
Loving that there's so much code going away :partying_face: Left a couple of notes where I'm not sure about behavior. 

Are there DB tables that can be dropped as a result of this PR?

services/horizon/internal/txsub/system.go
```diff
@@ -122,7 +120,7 @@ func (sys *System) Submit(
 		return
 	}
 
-	tx, sequenceNumber, err := checkTxAlreadyExists(ctx, db, hash, sourceAddress)
+	tx, _, err := checkTxAlreadyExists(ctx, db, hash, sourceAddress)
```
sreuland, line 123: it looks like `sequenceNumber` can be removed from `checkTxAlreadyExists` response now, nothing else uses it?

services/horizon/internal/txsub/system_test.go
sreuland: yes, can these unit tests be kept but change their expected outcomes(and title) to align with results from new functionality, i.e. core responds with errors?  b/c this is what the real world can still throw at it? to @Shaptic's point also, is there an integration test(using a captive core instance) that submits transactions of same invalidity and validate same outcomes but triggered from real core error response rather than mock?

services/horizon/internal/txsub/system.go
```diff
@@ -134,118 +132,36 @@ func (sys *System) Submit(
...
-// waitUntilAccountSequence blocks until either the context times out or the sequence number of the
-// given source account is greater than or equal to `seq`
-func (sys *System) waitUntilAccountSequence(ctx context.Context, db HorizonDB, sourceAddress string, seq uint64) error {
-	timer := time.NewTimer(sys.accountSeqPollInterval)
-	defer timer.Stop()
-
-	for {
-		sequenceNumbers, err := db.GetSequenceNumbers(ctx, []string{sourceAddress})
+		// If error is txBAD_SEQ, check for the result again
+		tx, err = txResultByHash(ctx, db, hash)
```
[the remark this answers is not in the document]
aditya1702: @tamirms I removed this w.r.t. George's comment here - https://github.com/stellar/go/pull/5030#discussion_r1309373894

But it looks like the scenario you described above has to be handled and this function does that. What do you think @Shaptic?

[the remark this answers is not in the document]
Shaptic: I'm certain that @tamirms is right because I was looking at the code from an optimization standpoint rather than a reliability standpoint in regards to ingestion lag & timing inconsistencies :detective:

services/horizon/internal/txsub/system_test.go
aditya1702: @Shaptic @sreuland I am adding back these tests since I have added the waitUntilAccountSequence function back as per Tamir's comment below.

services/horizon/internal/txsub/system.go
```diff
@@ -39,7 +38,6 @@ type System struct {
 	DB                func(context.Context) HorizonDB
```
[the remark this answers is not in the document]
aditya1702: No need to remove it since we are adding back waitUntilAccountSequence

services/horizon/internal/txsub/system.go
aditya1702: @Shaptic As Tamir described in the scenario below, we will get bad sequence errors if ingestion is lagging behind.

I will add some comments there to make it easy to understand.

aditya1702:
> Loving that there's so much code going away 🥳 Left a couple of notes where I'm not sure about behavior.
> 
> Are there DB tables that can be dropped as a result of this PR?

@Shaptic No db tables will be dropped for this change.

services/horizon/internal/txsub/system.go
```diff
@@ -134,75 +132,40 @@ func (sys *System) Submit(
...
+			sys.finish(ctx, hash, resultCh, Result{Err: sr.Err})
+			return
+		}
+		// If we found the result, use it as the result
+		sys.finish(ctx, hash, resultCh, Result{Transaction: tx})
+		return
 	}
 
+	// add transaction to open list
+	sys.Pending.Add(ctx, hash, resultCh)
```
tamirms, line 168: Error return value of `sys.Pending.Add` is not checked (errcheck)

services/horizon/internal/txsub/results.go
```diff
@@ -70,14 +70,14 @@ func checkTxAlreadyExists(ctx context.Context, db HorizonDB, hash, sourceAddress
...
 		}
 
-		num, ok := sequenceNumbers[sourceAddress]
+		_, ok := sequenceNumbers[sourceAddress]
 		if !ok {
-			return tx, 0, ErrNoAccount
+			return tx, ErrNoAccount
 		}
-		return tx, num, ErrNoResults
+		return tx, ErrNoResults
```
tamirms, line 80: it appears that all of this is dead code. we don't need to be calling `db.GetSequenceNumbers()` if we don't do anything with the results

aditya1702: @tamirms Ah my bad, I did not realise it was just fetching the sequence number. I have removed the code.

services/horizon/internal/txsub/system.go
```diff
@@ -134,75 +132,47 @@ func (sys *System) Submit(
...
-		sys.SubmissionQueue.NotifyLastAccountSequences(map[string]uint64{
-			sourceAddress: uint64(envelope.SeqNum()),
-		})
-	case <-ctx.Done():
-		sys.finish(ctx, hash, resultCh, Result{Err: sys.deriveTxSubError(ctx)})
+		// If error is txBAD_SEQ, check for the result again
+		tx, err = txResultByHash(ctx, db, hash)
+		if err != nil {
+			// finally, return the bad_seq error if no result was found on 2nd attempt
+			sys.finish(ctx, hash, resultCh, Result{Err: sr.Err})
```
sreuland, line 162: since `sys.finish` has to be called in all paths of this `Submit` method, and it's repeated alot, can that be accomplished as a single `defer` statement instead?

tamirms:
I think [ErrNoAccount](https://github.com/stellar/go/blob/43e3c17b9614924fe9d6e19cda37c5360edb4354/services/horizon/internal/txsub/errors.go#L22) can be removed since it seems to not be referenced anywhere

services/horizon/internal/txsub/results.go
```diff
@@ -55,29 +55,15 @@ func txResultFromHistory(tx history.Transaction) (history.Transaction, error) {
 // queries execute on different ledgers. In this case, txsub can mistakenly respond with a bad_seq error
 // because the first query occurs when the tx is not yet ingested and the second query occurs when the tx
 // is ingested.
-func checkTxAlreadyExists(ctx context.Context, db HorizonDB, hash, sourceAddress string) (history.Transaction, uint64, error) {
+func checkTxAlreadyExists(ctx context.Context, db HorizonDB, hash string) (history.Transaction, error) {
```
tamirms, line 58: I think you can remove this function entirely. it is no longer necessary to establish a repeatable read transaction because we are not fetching account sequence numbers anymore. Any code is calling `checkTxAlreadyExists()` can be replaced with a call to `txResultByHash()`

tamirms:
I think we can simplify [OpenSubmissionList](https://github.com/stellar/go/blob/43e3c17b9614924fe9d6e19cda37c5360edb4354/services/horizon/internal/txsub/open_submission_list.go) by removing all the error return values from the function signatures. I noticed that none of the functions actually returned any errors except for [`Add()`](https://github.com/stellar/go/blob/43e3c17b9614924fe9d6e19cda37c5360edb4354/services/horizon/internal/txsub/open_submission_list.go#L45) which tries to validate the transaction hash. I don't think it's necessary to have this validation because we have already validated the transaction hash before invoking txsub

services/horizon/internal/txsub/system.go
aditya1702: Yeah that is good idea.

services/horizon/internal/txsub/system.go
```diff
@@ -99,6 +99,11 @@ func (sys *System) Submit(
 	sys.Init()
 	resultCh := make(chan Result, 1)
 	resultReadCh = resultCh
+	var result Result
+
+	defer func() {
```
tamirms, line 104: I don't think deferring will work because there is a codepath where `sys.finish()` should not be called. Notice how `sys.finish()` is not called at the end of the function:

```golang
	// Add transaction to open list of pending txns: the transaction has been successfully submitted to core
	// but that does not mean it is included in the ledger. The txn status remains pending
	// until we see an ingestion in the db.
	sys.Pending.Add(hash, resultCh)
	return

```

aditya1702: @tamirms I had actually added a change for that which I did not commit yet. Can we add this check:

`defer func() {
	if result.Err != nil || result.Transaction.TransactionHash != "" {
		sys.finish(ctx, hash, resultCh, result)
	}
}()`

aditya1702: Lets just avoid defer then. cc @sreuland

services/horizon/internal/txsub/system.go
```diff
@@ -39,7 +38,6 @@ type System struct {
 	DB                func(context.Context) HorizonDB
 	Pending           OpenSubmissionList
 	Submitter         Submitter
-	SubmissionQueue   *sequence.Manager
 	SubmissionTimeout time.Duration
 	Log               *log.Entry
```
tamirms, line 43: you can delete the `BufferedSubmissionsGauge` field in the metrics below because it's no longer being used

aditya1702: Its weird, Goland used to give updates on which params are not used. Might have to enable it somewhere. Thanks @tamirms

tamirms (approved):
Looks good!

PR #5051, opened 2023-09-13
services/horizon: Add deprecation warning for using command-line flags

### LINKED ISSUE #5045: services/horizon: deprecate command-line parameter usage in favor of env vars

We recommend environment variable usage in our docs. We want to deprecate this behavior in the service so that developers/users that are currently using it are nudged into moving over.

A deprecation warning is logged when a user runs horizon with cli args. For now, it's still possible for them to send them. This ticket should encapsulate the work only to emit the warning, and we'll discuss timeline for full removal/deprecation and create a follow-up ticket to track the work to remove it completely.

this is consistent to standard conventions generally acknowledged in [12 Factor App](https://12factor.net/config) for micro-services 

Don't remove it.

aditya1702:
Closes #5045 

<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

We recommend environment variable usage in our docs. We want to deprecate command-line flag usage and remove it completely in the future. This deprecation warning will make users move towards using env vars.

### Why

Currently horizon has a ton of flags and this makes using them all very cumbersome. We want to move towards using environment variables.

### Known limitations

[TODO or N/A]


urvisavla (commented):
While we're currently only adding a deprecation warning, I'm wondering if we should transition our code that uses command line parameters, both in test cases and potentially other code, to instead use environment variables before publishing the warning?

services/horizon/cmd/root.go
```diff
@@ -16,7 +16,11 @@ var (
 		Short:         "client-facing api server for the Stellar network",
 		SilenceErrors: true,
 		SilenceUsage:  true,
-		Long:          "Client-facing API server for the Stellar network. It acts as the interface between Stellar Core and applications that want to access the Stellar network. It allows you to submit transactions to the network, check the status of accounts, subscribe to event streams and more.",
+		Long: "Client-facing API server for the Stellar network. It acts as the interface between Stellar Core " +
+			"and applications that want to access the Stellar network. It allows you to submit transactions to the " +
+			"network, check the status of accounts, subscribe to event streams and more.\n" +
+			"DEPRECATED - the use of command-line flags has been deprecated in favor of environment variables. Please" +
+			"consult our Configuring section in the developer documentation on how to use them - https://developers.stellar.org/docs/run-api-server/configuring",
```
sreuland, line 23: this highlights a separate concern, will need the partner experience docs to support backwards-compat urls for this link, since the url path name changed to `docs/run-platform-server/`

sreuland: the mention of `DEPRECATED` in help output may not have enough visibility since it's only seen when using the `-h` for help output.

Is it worth a step back to consider a different approach, rather than mention in static help output, a more dynamic cli usage for more timely feedback, such as a new function like `printDeprecatable(flags)` and then invoke that first from `RunE` command function callbacks. It would print to stdout/err a message that mentions all deprecatable flags that were detected as cli flags, such as:
```
Flags a,b,c, are deprecated on command line, please migrate the usage from cli to environment variables instead per https://developers.stellar.org/docs/run-api-server/configuring
```

`printDeprecatable` becomes a single place to define list of all deprecatable flags.

sreuland: @aditya1702 , I was referring to dynamically detecting flags presence on cli and tailoring the warning message displayed on cli to just the flags that were present in addition to the current approach in pr which is to include a statement of deprecated cli usage in the static global/general help output. That static help output alone doesn't interactively or dynamically detect when flags are actually on the cli and inform users of ongoing deprecated usages. Maybe the #5045's acceptance criteria is a bit ambiguous on expected outcome for use cases, but, I think it can be enhanced to provide more detailed outcome expectation:

```
1. Mention in horizon global help output that command line usage of flags is deprecated and environment variable usage is preferred instead.

2. Dynamically detect if any flags are present on horizon cli, such as:

stellar-horizon serve --a --b --c

and print a warning in cli output that states the flags presented should not be passed on cli.

The dynamically generated warning message:
Flags a,b,c, are deprecated on command line, please migrate the usage from cli to environment variables instead per https://developers.stellar.org/docs/run-api-server/configuring.
```

urvisavla (commented):
Can we avoid duplicating [environment.go](https://github.com/stellar/go/blob/master/services/horizon/internal/test/integration/environment.go)?

services/horizon/internal/flags_test.go
```diff
@@ -180,3 +183,55 @@ func Test_createCaptiveCoreConfig(t *testing.T) {
...
+	assert.Equal(t, config.ApplyMigrations, true)
+	assert.Equal(t, config.EnableCaptiveCoreIngestion, false)
+	assert.Equal(t, config.CheckpointFrequency, uint32(8))
+	assert.Equal(t, config.MaxDBConnections, 50)
+	assert.Equal(t, config.AdminPort, uint(6060))
+	assert.Equal(t, config.Port, uint(8001))
+	assert.Equal(t, config.CaptiveCoreBinaryPath, os.Getenv("HORIZON_INTEGRATION_TESTS_CAPTIVE_CORE_BIN"))
+	assert.Equal(t, config.CaptiveCoreConfigPath, "../docker/captive-core-classic-integration-tests.cfg")
+	assert.Equal(t, config.CaptiveCoreConfigUseDB, true)
+	envManager.Restore()
```
urvisavla, line 236: It might be a good idea to call envManager.Restore() early in a defer block.

support/config/config_option.go
```diff
@@ -58,6 +58,16 @@ func (cos ConfigOptions) SetValues() error {
 	return nil
 }
 
+func (cos ConfigOptions) GetAllFlagsPassedByUser() []string {
```
urvisavla, line 61: nit: rename it to explicitly mention that the function returns command line flags  e.g: `GetCommandLineFlags` or `GetFlagsFromCommandLine`. Also, add function docstring.

support/config/config_option.go
```diff
@@ -58,6 +58,16 @@ func (cos ConfigOptions) SetValues() error {
...
+	var flagsPassedByUser []string
+	for _, co := range cos {
+		if co.flag.Changed {
+			flagsPassedByUser = append(flagsPassedByUser, "--"+co.flag.Name)
```
urvisavla, line 65: nit: I don't know if this will be used anywhere else, but if it is, would it make sense to return the flag names here and add "--" when displaying them, like in `ApplyFlags` in this case?

services/horizon/internal/integration/parameters_test.go
```diff
@@ -450,40 +444,94 @@ func TestDisableTxSub(t *testing.T) {
...
+
+		outputBytes, _ := io.ReadAll(r)
+		wg.Wait() // Wait for the goroutine to finish before proceeding
+		_ = r.Close()
+		os.Stderr = originalStderr
+
+		assert.Contains(t, string(outputBytes), "DEPRECATED - No ingestion filter rules are defined by default, which equates to "+
+			"no filtering of historical data. If you have never added filter rules to this deployment, then nothing further needed. "+
+			"If you have defined ingestion filter rules prior but disabled filtering overall by setting this flag disabled with "+
+			"--exp-enable-ingestion-filtering=false, then you should now delete the filter rules using the Horizon Admin API to achieve "+
```
urvisavla, line 480: Should this message be updated to environment variable name instead of command line?

services/horizon/internal/integration/parameters_test.go
```diff
@@ -450,40 +444,94 @@ func TestDisableTxSub(t *testing.T) {
...
+	})
+	t.Run("deprecated output for command-line flags", func(t *testing.T) {
+		originalStderr := os.Stderr
+		r, w, _ := os.Pipe()
+		os.Stderr = w
+		stdLog.SetOutput(os.Stderr)
+
+		config, flags := horizon.Flags()
+
+		horizonCmd := &cobra.Command{
```
urvisavla, line 491: Why didn't we use the `StartHorizon` utility function in this case like we did [here](https://github.com/stellar/go/pull/5051/files#diff-6f737140ca2744b223417ca5d5f6f4a20e9812494960796e774932e97b85da5fR454-R459) ? Is it because `StartHorizon` would start both the web and ingest instances which would take more time?

aditya1702: @urvisavla Yes thats the reason. Testing the deprecated command doesnt require spinning up Horizon and the docker containers along with the web and ingest args.

services/horizon/internal/integration/parameters_test.go
aditya1702: Yeah good catch. I will do it to make it consistent with using env variables

urvisavla (approved):
LGTM!

PR #5054, opened 2023-09-14
Re-point Soroban RPC to the final Docker image

Shaptic:
self-explanatory: this is the "final" (_knocks on wood_) Soroban RPC Docker image for Preview 11.


ire-and-curses (approved):
Nice!

Issue #3942, opened 2021-09-20
Unstable order of claimable balance effects during liquidity pool revocation


2opremio:
The Liquidity Pool revocation integration test is flaky due to the claimable balance effects not being ordered consistently in different executions.

See https://app.circleci.com/pipelines/github/stellar/go/9376/workflows/560678ef-ba51-40af-8d78-cec2dee83151 (e.g. https://app.circleci.com/pipelines/github/stellar/go/9376/workflows/560678ef-ba51-40af-8d78-cec2dee83151/jobs/53849 ).

This is (most probably) due to core's metadata not being ordered consistently.

We could consider this "feature" but I think core should always output the same order.

jonjove:
There is no meaningful ordering on the meta within a single operation. We could pick an order, but sorting it into that order would make core slower. There is no reason to assume that every consumer expects the meta to have a specific order, so I don't really see a justification for making core slower.

PR #5070, opened 2023-10-03
ingest: Extract ledger entry changes from Tx Meta in a deterministic order

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

In https://github.com/stellar/stellar-core/issues/3952 we discovered that the tx meta emitted by captive core can appear in a random order. This is problematic for Horizon because we would like ingestion to always be deterministic and reproducible. For some releases we test our ingestion system by ingesting history twice, once with the release candidate and the second time with the previous stable version and then we assert that content produced by both runs are the same. If tx meta can appear in a random order this can result in horizon effects being produced in a different order.

This commit is an attempt to make the ordering of Tx Meta consistent in Horizon ingestion.
### Known limitations

https://github.com/stellar/stellar-core/pull/3954 is an attempt to fix this issue in core. Once that change is included in a core stable release we can revert this PR.

### DIFF
--- a/ingest/change.go
+++ b/ingest/change.go
@@ -72,7 +72,8 @@ func GetChangesFromLedgerEntryChanges(ledgerEntryChanges xdr.LedgerEntryChanges)
 		}
 	}
 
-	return sortChanges(changes)
+	sortChanges(changes)
+	return changes
 }
 
 type sortableChanges struct {
@@ -118,9 +119,8 @@ func (s sortableChanges) Swap(i, j int) {
 // multiple changes with the same ledger key in a LedgerEntryChanges group,
 // but if that is the case, we fall back to the original ordering of the changes
 // by using a stable sorting algorithm.
-func sortChanges(changes []Change) []Change {
+func sortChanges(changes []Change) {
 	sort.Stable(newSortableChanges(changes))
-	return changes
 }
 
 // LedgerEntryChangeType returns type in terms of LedgerEntryChangeType.
--- a/ingest/change.go
+++ b/ingest/change.go
@@ -2,6 +2,7 @@ package ingest
 
 import (
 	"bytes"
+	"sort"
 
 	"github.com/stellar/go/support/errors"
 	"github.com/stellar/go/xdr"
@@ -18,6 +19,13 @@ type Change struct {
 	Type xdr.LedgerEntryType
 	Pre  *xdr.LedgerEntry
 	Post *xdr.LedgerEntry
+}
+
+func (c *Change) ledgerKey() (xdr.LedgerKey, error) {
+	if c.Pre != nil {
+		return c.Pre.LedgerKey()
+	}
+	return c.Post.LedgerKey()
 }
 
 // GetChangesFromLedgerEntryChanges transforms LedgerEntryChanges to []Change.
@@ -64,6 +72,54 @@ func GetChangesFromLedgerEntryChanges(ledgerEntryChanges xdr.LedgerEntryChanges)
 		}
 	}
 
+	return sortChanges(changes)
+}
+
+type sortableChanges struct {
+	changes    []Change
+	ledgerKeys [][]byte
+}
+
+func newSortableChanges(changes []Change) sortableChanges {
+	ledgerKeys := make([][]byte, len(changes))
+	for i, c := range changes {
+		lk, err := c.ledgerKey()
+		if err != nil {
+			panic(err)
+		}
+		lkBytes, err := lk.MarshalBinary()
+		if err != nil {
+			panic(err)
+		}
+		ledgerKeys[i] = lkBytes
+	}
+	return sortableChanges{
+		changes:    changes,
+		ledgerKeys: ledgerKeys,
+	}
+}
+
+func (s sortableChanges) Len() int {
+	return len(s.changes)
+}
+
+func (s sortableChanges) Less(i, j int) bool {
+	return bytes.Compare(s.ledgerKeys[i], s.ledgerKeys[j]) < 0
+}
+
+func (s sortableChanges) Swap(i, j int) {
+	s.changes[i], s.changes[j] = s.changes[j], s.changes[i]
+	s.ledgerKeys[i], s.ledgerKeys[j] = s.ledgerKeys[j], s.ledgerKeys[i]
+}
+
+// sortChanges is applied on a list of changes to ensure that LedgerEntryChanges
+// from Tx Meta are ingested in a deterministic order.
+// The changes are sorted by ledger key. It is unexpected for there to be
+// multiple changes with the same ledger key in a LedgerEntryChanges group,
+// but if that is the case, we fall back to the original ordering of the changes
+// by using a stable sorting algorithm.
+func sortChanges(changes []Change) []Change {
+	sort.Stable(newSortableChanges(changes))
 	return changes
 }
 
--- a/ingest/ledger_change_reader.go
+++ b/ingest/ledger_change_reader.go
@@ -139,6 +139,7 @@ func (r *LedgerChangeReader) Read() (Change, error) {
 				Post: nil,
 			}
 		}
+		changes = sortChanges(changes)
 		r.pending = append(r.pending, changes...)
 		r.state++
 		return r.Read()
--- a/ingest/ledger_change_reader.go
+++ b/ingest/ledger_change_reader.go
@@ -139,7 +139,7 @@ func (r *LedgerChangeReader) Read() (Change, error) {
 				Post: nil,
 			}
 		}
-		changes = sortChanges(changes)
+		sortChanges(changes)
 		r.pending = append(r.pending, changes...)
 		r.state++
 		return r.Read()
--- a/ingest/change_test.go
+++ b/ingest/change_test.go
@@ -1,0 +1,215 @@
+package ingest
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/stellar/go/xdr"
+)
+
+func assertChangesAreEqual(t *testing.T, a, b Change) {
+	assert.Equal(t, a.Type, b.Type)
+	if a.Pre == nil {
+		assert.Nil(t, b.Pre)
+	} else {
+		aBytes, err := a.Pre.MarshalBinary()
+		assert.NoError(t, err)
+		bBytes, err := b.Pre.MarshalBinary()
+		assert.NoError(t, err)
+		assert.Equal(t, aBytes, bBytes)
+	}
+	if a.Post == nil {
+		assert.Nil(t, b.Post)
+	} else {
+		aBytes, err := a.Post.MarshalBinary()
+		assert.NoError(t, err)
+		bBytes, err := b.Post.MarshalBinary()
+		assert.NoError(t, err)
+		assert.Equal(t, aBytes, bBytes)
+	}
+}
+
+func TestSortChanges(t *testing.T) {
+	for _, testCase := range []struct {
+		input    []Change
+		expected []Change
+	}{
+		{[]Change{}, []Change{}},
+		{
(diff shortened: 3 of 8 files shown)


MonsieurNicolas:
This change is surprising to me: the key order in meta has never been deterministic (since 2015), so what changed in Horizon and/or clients?

If this is it to help with testing, it seems odd to take a dependency on ordering in production?

Also, meta already stored basically everywhere (Horizon DB, core-DB in "legacy mode") is already non-canonical so this change may not be enough.

MonsieurNicolas:
I mean: if it makes the logic simpler, you can try to normalize meta in some way across all data sources, but I am not sure it's something you really want.

tamirms:
@MonsieurNicolas 


> This change is surprising to me: the key order in meta has never been deterministic (since 2015), so what changed in Horizon and/or clients?
If this is it to help with testing, it seems odd to take a dependency on ordering in production?

Having horizon ingestion be deterministic and reproducible makes it a lot easier for us catch bugs in release candidates. The most straightforward way to guarantee that horizon ingestion is deterministic is to ensure that the order of tx meta is deterministic. What would be the alternative approach?

> Also, meta already stored basically everywhere (Horizon DB, core-DB in "legacy mode") is already non-canonical so this change may not be enough.

This change will re-sort tx meta before Horizon consumes it in ingestion. The sorting will be applied regardless if the source of the tx meta is captive core or the core-db in legacy mode.

MonsieurNicolas:
> Having horizon ingestion be deterministic and reproducible makes it a lot easier for us catch bugs in release candidates. The most straightforward way to guarantee that horizon ingestion is deterministic is to ensure that the order of tx meta is deterministic. What would be the alternative approach?

Maybe it depends at which layer you test, which is why I was asking about what changed in Horizon: I thought you would just compare ingested state with expected state (so set comparisons) and not really worry about the format of the diff.

tamirms:
@MonsieurNicolas Horizon stores some ledger entries in its database and those are periodically compared against the ledger entries from history archive snapshots. That comparison is a set comparison like you described (where ordering doesn't matter). 

But we also compare horizon effects between the release candidate and the last stable release. The ordering of meta does affect the ordering of horizon effects in some cases.

Here is an example comparing the effects response for a specific [operation](https://horizon.stellar.org/operations/207289144773263361) in [horizon staging](https://horizon-stg-pubnet.stellar.org/operations/207289144773263361/effects) and [horizon production](https://horizon.stellar.org/operations/207289144773263361/effects)

In staging the effects response looks like:

```
[
      {
        "id": "0207289144773263361-0000000001",
        "paging_token": "207289144773263361-1",
        "account": "GDPCYRFMTUIGB37GJ76LNJP7NXUENBACS5QTBLMICFIOYAJNGMGHBJIU",
        "type": "trustline_deauthorized",
        "type_i": 24,
        "created_at": "2023-09-24T22:46:03Z",
        "trustor": "GDWAC25VFDIIX442O34FEKFPEALTRDUN5LU3ANNIDWO25JJMBAI2JXC4",
        "asset_type": "credit_alphanum12",
        "asset_code": "NICKEL"
      },
     ..., // other effects omitted for brevity  
      {
        "id": "0207289144773263361-0000000008",
        "paging_token": "207289144773263361-8",
        "account": "GDPCYRFMTUIGB37GJ76LNJP7NXUENBACS5QTBLMICFIOYAJNGMGHBJIU",
        "type": "claimable_balance_sponsorship_created",
        "type_i": 69,
        "created_at": "2023-09-24T22:46:03Z",
        "balance_id": "00000000362c689c6734eca6de952a0cbb4eaf28545885307e66710747aa03935f446199",
        "sponsor": "GDWAC25VFDIIX442O34FEKFPEALTRDUN5LU3ANNIDWO25JJMBAI2JXC4"
      },
      {
        "id": "0207289144773263361-0000000009",
        "paging_token": "207289144773263361-9",
        "account": "GDPCYRFMTUIGB37GJ76LNJP7NXUENBACS5QTBLMICFIOYAJNGMGHBJIU",
        "type": "claimable_balance_sponsorship_created",
        "type_i": 69,
        "created_at": "2023-09-24T22:46:03Z",
        "balance_id": "00000000ba96b670dfbd77fe9cf207336439806022dd77dfd1fcdda0f3a8c48559e11dee",
        "sponsor": "GDWAC25VFDIIX442O34FEKFPEALTRDUN5LU3ANNIDWO25JJMBAI2JXC4"
      },
      {
        "id": "0207289144773263361-0000000010",
        "paging_token": "207289144773263361-10",
        "account": "GDPCYRFMTUIGB37GJ76LNJP7NXUENBACS5QTBLMICFIOYAJNGMGHBJIU",
        "type": "liquidity_pool_removed",
        "type_i": 94,
        "created_at": "2023-09-24T22:46:03Z",
        "liquidity_pool_id": "3e1337116f38a6ff6b4161cb4455ea4f3f2777562e52dbeffebaa0582cc7e561"
      }
    ]
```

In production the effects response looks like:

```
[
      {
        "id": "0207289144773263361-0000000001",
        "paging_token": "207289144773263361-1",
        "account": "GDPCYRFMTUIGB37GJ76LNJP7NXUENBACS5QTBLMICFIOYAJNGMGHBJIU",
        "type": "trustline_deauthorized",
        "type_i": 24,
        "created_at": "2023-09-24T22:46:03Z",
        "trustor": "GDWAC25VFDIIX442O34FEKFPEALTRDUN5LU3ANNIDWO25JJMBAI2JXC4",
        "asset_type": "credit_alphanum12",
        "asset_code": "NICKEL"
      },
      {
        "id": "0207289144773263361-0000000008",
        "paging_token": "207289144773263361-8",
        "account": "GDPCYRFMTUIGB37GJ76LNJP7NXUENBACS5QTBLMICFIOYAJNGMGHBJIU",
        "type": "claimable_balance_sponsorship_created",
        "type_i": 69,
        "created_at": "2023-09-24T22:46:03Z",
        "balance_id": "00000000ba96b670dfbd77fe9cf207336439806022dd77dfd1fcdda0f3a8c48559e11dee",
        "sponsor": "GDWAC25VFDIIX442O34FEKFPEALTRDUN5LU3ANNIDWO25JJMBAI2JXC4"
      },
     ..., // other effects omitted for brevity  
      {
        "id": "0207289144773263361-0000000009",
        "paging_token": "207289144773263361-9",
        "account": "GDPCYRFMTUIGB37GJ76LNJP7NXUENBACS5QTBLMICFIOYAJNGMGHBJIU",
        "type": "claimable_balance_sponsorship_created",
        "type_i": 69,
        "created_at": "2023-09-24T22:46:03Z",
        "balance_id": "00000000362c689c6734eca6de952a0cbb4eaf28545885307e66710747aa03935f446199",
        "sponsor": "GDWAC25VFDIIX442O34FEKFPEALTRDUN5LU3ANNIDWO25JJMBAI2JXC4"
      },
      {
        "id": "0207289144773263361-0000000010",
        "paging_token": "207289144773263361-10",
        "account": "GDPCYRFMTUIGB37GJ76LNJP7NXUENBACS5QTBLMICFIOYAJNGMGHBJIU",
        "type": "liquidity_pool_removed",
        "type_i": 94,
        "created_at": "2023-09-24T22:46:03Z",
        "liquidity_pool_id": "3e1337116f38a6ff6b4161cb4455ea4f3f2777562e52dbeffebaa0582cc7e561"
      }
    ]
```

The responses are almost identical. The only difference is in the `claimable_balance_sponsorship_created` effects. In [horizon staging](https://horizon-stg-pubnet.stellar.org/operations/207289144773263361/effects)  `00000000362c689c6734eca6de952a0cbb4eaf28545885307e66710747aa03935f446199` is first and is followed by `00000000ba96b670dfbd77fe9cf207336439806022dd77dfd1fcdda0f3a8c48559e11dee`. In [horizon production](https://horizon.stellar.org/operations/207289144773263361/effects) the order is reversed: `00000000ba96b670dfbd77fe9cf207336439806022dd77dfd1fcdda0f3a8c48559e11dee` is first and is followed by `00000000362c689c6734eca6de952a0cbb4eaf28545885307e66710747aa03935f446199`.


After doing some digging, I found that this issue of random tx meta ordering affecting the effects response has occurred before in https://github.com/stellar/go/issues/3942 . If you look at the conversation in the issue, @2opremio argues that core should provide a deterministic ordering and he was met with objections from Jon Jove who was concerned about the impact of core's performance.

The way we fixed this issue is by adding sorting logic locally in the ingestion processor: 

https://github.com/stellar/go/pull/3943/commits/883ccb318bbdad968c5089e8699ac3c33406f631

We could fix this new issue in the same way by adding sorting logic locally in this part of the code:

https://github.com/stellar/go/blob/facabfc2f4c4a8d01ab5ad7da14955247ad7dba0/services/horizon/internal/ingest/processors/effects_processor.go#L262-L267

However, I believe a more robust solution would be sorting tx meta at the source either in core (the ideal solution) or from the ingestion library which reads tx meta from core (which is what this PR does).

ingest/change.go
```diff
@@ -64,6 +72,54 @@ func GetChangesFromLedgerEntryChanges(ledgerEntryChanges xdr.LedgerEntryChanges)
...
+}
+
+// sortChanges is applied on a list of changes to ensure that LedgerEntryChanges
+// from Tx Meta are ingested in a deterministic order.
+// The changes are sorted by ledger key. It is unexpected for there to be
+// multiple changes with the same ledger key in a LedgerEntryChanges group,
+// but if that is the case, we fall back to the original ordering of the changes
+// by using a stable sorting algorithm.
+func sortChanges(changes []Change) []Change {
+	sort.Stable(newSortableChanges(changes))
```
sreuland, line 122: since `sortableChanges` can mutate the underlying array in `changes` passed from caller as part of Swap method, should a copy of `changes` be made to avoid side effect?

ingest/ledger_change_reader.go
```diff
@@ -139,6 +139,7 @@ func (r *LedgerChangeReader) Read() (Change, error) {
 				Post: nil,
 			}
 		}
+		changes = sortChanges(changes)
```
sreuland, line 142: this now results in all changes being (re)sorted, which is great for the goal stated, do you think it may have potential to trigger false warnings from [verify-range ](https://github.com/stellar/go/blob/master/services/horizon/docker/verify-range/start#L155-L163) as that compares history table dumps, at least in the interim while the production dumps are from a horizon release prior to the one having this change, maybe it's just an aspect worth mentioning in the changelog notes?

sreuland (approved):

ingest/change.go
sreuland: ok, that works also, thanks!

ingest/ledger_change_reader.go
tamirms: > do you think it may have potential to trigger false warnings from verify-range ?

yes, that's true but we actually already have false positives from verify-range due to random ordering of tx meta from core. In fact, I discovered this issue by debugging verify-range failures on the soroban release candidate.

The aim of this PR is to fix this issue for all releases going forward. If we want to compare a release candidate with an older version then we would need to cherry-pick this commit which sorts the ingestion changes deterministically on top of the older version. Then we could run verify-range on the release candidate and compare it against the older version + cherry-picked commit.

Good call about adding this to the changelog notes. I can do that

MonsieurNicolas:
Thanks @tamirms for the background. I didn't realize this was needed for effects that don't have good global "IDs".
Now that I understand what this is, I think this change looks fine as Horizon may emit effects from basically anything, right?

Soroban-RPC / other ingesters may not need this though (but maybe we have a lot of spare CPU there)?

As far as doing the sort at the core layer: as we'd be looking at some pretty bad migration (to eventually remove this from Horizon), it does not seem to be worth doing at this time (we may revisit this if we do SPVs/parallel execution).

PR #5077, opened 2023-10-06
cmd: limit global flags displayed on cli help output

### LINKED ISSUE #5022: services/horizon/flags: refined cli help content
current help output printed to console from `stellar-horizon -h` and sub-commands  like`stellar-horizon <sub-command> -h` prints all Global Flags in any case, many of which are not relevant within specific sub-commands.

the Global Flags section can be noisy, making difficult to understand what parameters are relevant for each specific sub-command like `serve`, `db`, `ingest`, `record-metrics`

reduced cli help output, less copy, but what is printed is more concise to context.

* `stellar-horizon -h` doesn't print Global Flags, just the sub-commands available and new 'best practices' blurb to recommend using environment variables overall.

* `stellar-horizon <sub-command> ... -h`, doesn't print Global Flags, just the flags relevant for the `<sub-command>`'s given, i.e. the help output for `stellar-horizon db init` vs `stellar-horizon db reingest`

sreuland:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
* Limit which global flags will display on `-h` cli output to when it's only relevant to the command's general roles that it may be performing 'ingestion', 'db update tool' or 'api server'.
* Don't show global flags on help output when the most specific sub-command still has child sub-commands, limit help output to just list of available sub-commands.

### Why

The global flags are a fixed set of parameters and get displayed as a single large block on the console help output for any help request regardless of the level of `horizon` command. Since it's a fixed list, many of the global parameters tend to not be relevant to the context of command being requested by help and therefore can be somewhat mis-leading.

Closes #5022 

### Notes for review:

Start here, the solution overrides the cobra HelpFunc:
https://github.com/stellar/go/pull/5077/files#diff-7ad09d8c6e6a91967287e26e0343f0262ed02aa9fab85a0741cf665eeab514efR52

Some screen shots of help output from this PR on top, and current horizon on bottom.  
  
<br>
`stellar-horizon -h`. 

Screen Shot 2023-10-05 at 8 40 06 PM

<br>
<br>
`stellar-horizon ingest -h`. 

Screen Shot 2023-10-05 at 8 41 50 PM
<br>
<br>
Screen shot of help output from this PR on left, and current horizon on right:  
`stellar-horizon db reingest range -h`. 

Screen Shot 2023-10-05 at 8 50 39 PM


aditya1702 (approved):
LGTM. Just need to resolve the merge conflicts.

urvisavla (commented):
The changes look great!  The only thing I noticed is that `horizon db reingest` displays all global flags, whereas `horizon db reingest -h` does not. I believe `horizon db reingest` cannot be used on its own without specifying a range so I expected the output of `horizon db reingest` to same as `horizon db reingest -h`

sreuland:
> The changes look great! The only thing I noticed is that `horizon db reingest` displays all global flags, whereas `horizon db reingest -h` does not. I believe `horizon db reingest` cannot be used on its own without specifying a range so I expected the output of `horizon db reingest` to same as `horizon db reingest -h`

good catch, yeah, I think cobra is routing through a 'usage' output function there, so the hooked 'help' didn't get a chance to run, I'll look into it further to see if can capture that path, thanks!

sreuland:
> The changes look great! The only thing I noticed is that `horizon db reingest` displays all global flags, whereas `horizon db reingest -h` does not. I believe `horizon db reingest` cannot be used on its own without specifying a range so I expected the output of `horizon db reingest` to same as `horizon db reingest -h`

@urvisavla , fortunately, cobra allowed overriding the usage output in same way as help output, so i was able to apply the same global flags behavior in this case of usage - https://github.com/stellar/go/pull/5077/commits/deca1ccebc4e03a2165f67215a120e3c6b396328

urvisavla (approved):

Issue #5061, opened 2023-09-21
Remove unused indexes from horizon schema


tamirms:
While investigating https://github.com/stellar/go/issues/5059 , I discovered that the `history_operations_on_type` was not used by Horizon despite being present in the db schema for several years. I used the following query from https://pgdash.io/blog/finding-unused-indexes-in-postgresql.html to find all the indexes in the horizon schema and the number of times each index has been scanned:

```
select schemaname || '.' || indexrelname as index, idx_scan from pg_stat_user_indexes;
```

The output from that query indicates that there are other indexes which are also unused:

```
index                                  |   idx_scan    
------------------------------------------------------------------------+---------------
 public.accounts_data_pkey                                              |             0
 public.accounts_pkey                                                   |    1031025647
 public.accounts_signers_pkey                                           |        108852
 public.claimable_balance_claimants_pkey                                |             0
 public.claimable_balances_pkey                                         |     854140668
 public.exp_asset_stats_pkey                                            |      25320968
 public.gorp_migrations_pkey                                            |             0
 public.history_assets_asset_code_asset_type_asset_issuer_key           |     374988189
 public.history_assets_pkey                                             |   51865712229
 public.history_trades_60000_pkey                                       |       9023515
 public.key_value_store_pkey                                            |    2560632613
 public.liquidity_pools_pkey                                            |      12866984
 public.offers_pkey                                                     |     487719082
 public.trust_lines_pkey                                                |             0
 public.accounts_by_sponsor                                             |          1336
 public.accounts_data_account_id_name                                   |     549970728
 public.accounts_data_by_sponsor                                        |          1605
 public.accounts_home_domain                                            |             0
 public.accounts_inflation_destination                                  |             0
 public.accounts_signers_by_sponsor                                     |          1330
 public.asset_by_issuer                                                 |             0
 public.best_offer                                                      |    1122713103
 public.by_hash                                                         |     799480837
 public.by_inner_hash                                                   |     799480721
 public.by_ledger                                                       |    1095649867
 public.claimabable_balances_by_claimants                               |             0
 public.claimable_balance_claimants_by_destination_last_modified_ledger |      32094806
 public.claimable_balances_by_asset_last_modified_ledger_id             |       3789899
 public.claimable_balances_by_last_modified_ledger_and_id               |       9862204
 public.claimable_balances_by_sponsor_last_modified_ledger_id           |       1022056
 public.exp_asset_stats_by_issuer                                       |      95818931
[...]

PR #5081, opened 2023-10-12
5061/remove unused indexes

urvisavla:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Added db migration to remove unused indices from the Horizon database. 

### Why

see #5061 for details

### Known limitations
- The migration took about 40 mins on staging db which has full history. 

 - This migration needs to occur without a transaction. Otherwise, the migration process will maintain an exclusive lock on the tables even after dropping the index, leading to deadlock when concurrent requests that require a shared lock are being served. An alternative approach is to have a separate migration file for each index.


sreuland (approved):
nice work!

tamirms (approved):

Issue #4909, opened 2023-06-15
Refactor ingestion data-flow


tamirms:
In https://docs.google.com/document/d/1YETNALx5EzqZDNSVWzTfaK5Ogw84PsBlrt64nOr-Njg/edit?usp=sharing we were able to speed up ingestion by refactoring the ingestion data-flow. Currently the Horizon processors for the history tables are coupled to a single ledger. We need to break this coupling and allow a Horizon processor instance to be used on batches of ledgers.

Also, we need to refactor how the DB is used in a Horizon processor. Currently, we periodically insert batches of rows to the DB and flush all remaining rows to the DB at the end of the ingestion round. We should instead accumulate all the rows in memory and only use the DB at the very end of the ingestion round to flush all the rows using [COPY statements](https://github.com/stellar/go/issues/4908).

Note that these changes only need to be applied on the Horizon processors for the history tables. We do not need to modify the Horizon processors for the state tables.

In the [spike branch](https://github.com/stellar/go/compare/master...tamirms:go:ingest-perf-pq) the interface for the Horizon history processors was changed to:

```go
type horizonTransactionProcessor interface {
	ProcessTransaction(xdr.LedgerCloseMeta, ingest.LedgerTransaction) error
	Commit(ctx context.Context, session db.SessionInterface) error
}
```

Note how `ProcessTransaction()` now takes `xdr.LedgerCloseMeta` parameter which allows the processor to be used on transactions spanning multiple ledgers. Also, `ProcessTransaction()` no longer has a context parameter because we don't expect to have any DB operations in `ProcessTransaction()`. Instead, `ProcessTransaction()` will only accumulate rows for the history tables in-memory. The `Commit()` function is the only part of the processor which should have access to the db and it will be used to flush the in-memory rows to the DB.

Some of the history tables rely on lookup tables to obtain integer ids for accounts, assets, claimable balances, and liquidity pools. In this case the data-flow will be slightly more complex. In the [spike branch](https://github.com/stellar/go/compare/master...tamirms:go:ingest-perf-pq), the code for keeping track of all the id lookups was encapsulated into a "loader" component. Whenever the processor encountered an account string in `ProcessTransaction()`, the processor would register the account string in the loader component. The resulting data-flow looked like:

```go
	accountLoader := history.NewAccountLoader()
	cbLoader := history.NewClaimableBalanceLoader()
	lpLoader := history.NewLiquidityPoolLoader()
	assetLoader := history.NewAssetLoader()
[...]

PR #5083, opened 2023-10-16
integrating new loaders and builders into processors

sreuland:
wiring up the new processor Flush method and loaders with the fast batch builders into the existing processor flow.
fixing all tests.

This is part of #4909

### DIFF
--- a/services/horizon/docker/verify-range/dependencies
+++ b/services/horizon/docker/verify-range/dependencies
@@ -21,6 +21,3 @@ cd stellar-go
 git config --add remote.origin.fetch "+refs/pull/*/head:refs/remotes/origin/pull/*"
 git fetch --force --quiet origin
 /usr/local/go/bin/go build -v ./services/horizon/.
-
-
-
--- a/services/horizon/docker/verify-range/dependencies
+++ b/services/horizon/docker/verify-range/dependencies
@@ -11,8 +11,8 @@ echo "deb https://apt.stellar.org $(lsb_release -cs) stable" | sudo tee -a /etc/
 apt-get update
 apt-get install -y stellar-core=${STELLAR_CORE_VERSION}
 
-wget -q https://dl.google.com/go/go1.18.linux-amd64.tar.gz
-tar -C /usr/local -xzf go1.18.linux-amd64.tar.gz
+wget -q https://dl.google.com/go/go1.20.linux-amd64.tar.gz
+tar -C /usr/local -xzf go1.20.linux-amd64.tar.gz
 
 git clone https://github.com/stellar/go.git stellar-go
 cd stellar-go
@@ -20,4 +20,7 @@ cd stellar-go
 # Below ensures we also fetch PR refs
 git config --add remote.origin.fetch "+refs/pull/*/head:refs/remotes/origin/pull/*"
 git fetch --force --quiet origin
-/usr/local/go/bin/go build -v ./services/horizon
+/usr/local/go/bin/go build -v ./services/horizon/.
+
+
+
--- a/services/horizon/internal/db2/history/account_loader.go
+++ b/services/horizon/internal/db2/history/account_loader.go
@@ -28,7 +28,7 @@ const loaderLookupBatchSize = 50000
 
 // Value implements the database/sql/driver Valuer interface.
 func (a FutureAccountID) Value() (driver.Value, error) {
-	return a.loader.GetNow(a.address), nil
+	return a.loader.GetNow(a.address)
 }
 
 // AccountLoader will map account addresses to their history
@@ -71,11 +71,15 @@ func (a *AccountLoader) GetFuture(address string) FutureAccountID {
 // GetNow should only be called on values which were registered by
 // GetFuture() calls. Also, Exec() must be called before any GetNow
 // call can succeed.
-func (a *AccountLoader) GetNow(address string) int64 {
-	if id, ok := a.ids[address]; !ok {
-		panic(fmt.Errorf("address %v not present", address))
+func (a *AccountLoader) GetNow(address string) (int64, error) {
+	if !a.sealed {
+		return 0, fmt.Errorf(`invalid account loader state,  
+		Exec was not called yet to properly seal and resolve %v id`, address)
+	}
+	if internalID, ok := a.ids[address]; !ok {
+		return 0, fmt.Errorf(`account loader address %q was not found`, address)
 	} else {
-		return id
+		return internalID, nil
 	}
 }
 
@@ -206,4 +210,8 @@ func NewAccountLoaderStub() AccountLoaderStub {
 // address is mapped to the provided history account id
 func (a AccountLoaderStub) Insert(address string, id int64) {
 	a.Loader.ids[address] = id
+}
+
+func (a AccountLoaderStub) Sealed() {
+	a.Loader.sealed = true
 }
--- a/services/horizon/internal/db2/history/account_loader.go
+++ b/services/horizon/internal/db2/history/account_loader.go
@@ -209,9 +209,6 @@ func NewAccountLoaderStub() AccountLoaderStub {
 // Insert updates the wrapped AccountLoader so that the given account
 // address is mapped to the provided history account id
 func (a AccountLoaderStub) Insert(address string, id int64) {
-	a.Loader.ids[address] = id
-}
-
-func (a AccountLoaderStub) Sealed() {
 	a.Loader.sealed = true
+	a.Loader.ids[address] = id
 }
--- a/services/horizon/internal/db2/history/asset_loader.go
+++ b/services/horizon/internal/db2/history/asset_loader.go
@@ -23,6 +23,9 @@ type AssetKey struct {
 }
 
 func (key AssetKey) String() string {
+	if key.Type == xdr.AssetTypeToString[xdr.AssetTypeAssetTypeNative] {
+		return key.Type
+	}
 	return key.Type + "/" + key.Code + "/" + key.Issuer
 }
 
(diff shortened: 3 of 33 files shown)


services/horizon/internal/db2/history/asset_loader.go
```diff
@@ -22,6 +22,10 @@ type AssetKey struct {
 	Issuer string
 }
 
+func (key AssetKey) String() string {
+	return key.Type + "/" + key.Code + "/" + key.Issuer
```
tamirms, line 26: the String() implementation should match https://github.com/stellar/go/blob/master/xdr/asset.go#L246

so if the asset is native `String()` should return "native" instead of "native//"

sreuland: @tamirms , I adjusted to that recommendation. 

The PR is now passing all tests. How about I merge this PR to complete the functionality. And then will do a follow-on PR to port latest master back to ingestion-next and then can follow that up with on more PR to merge ingestion-next to master?

services/horizon/docker/verify-range/dependencies
```diff
@@ -11,13 +11,16 @@ echo "deb https://apt.stellar.org $(lsb_release -cs) stable" | sudo tee -a /etc/
...
 
 git clone https://github.com/stellar/go.git stellar-go
 cd stellar-go
 # By default "git fetch" only fetches refs/<branchname>
 # Below ensures we also fetch PR refs
 git config --add remote.origin.fetch "+refs/pull/*/head:refs/remotes/origin/pull/*"
 git fetch --force --quiet origin
-/usr/local/go/bin/go build -v ./services/horizon
+/usr/local/go/bin/go build -v ./services/horizon/.
+
```
tamirms, line 24: nit: trailing newlines

services/horizon/internal/db2/history/fee_bump_scenario.go
```diff
@@ -269,6 +270,8 @@ func FeeBumpScenario(tt *test.T, q *Q, successful bool) FeeBumpFixture {
 	details, err := json.Marshal(map[string]string{
 		"bump_to": "98",
 	})
+
+	fmt.Print(string(details))
```
tamirms, line 274: I assume this was for debugging. we can remove this now

services/horizon/internal/ingest/processor_runner.go
```diff
@@ -311,46 +323,51 @@ func (s *ProcessorRunner) RunTransactionProcessorsOnLedger(ledger xdr.LedgerClos
 		transactionReader      *ingest.LedgerTransactionReader
 	)
 
+	if err = s.checkIfProtocolVersionSupported(ledger.ProtocolVersion()); err != nil {
+		err = errors.Wrap(err, "Error while checking for supported protocol version")
+		return
+	}
+
+	// ensure capture of the ledger to history regardless of whether it has transactions.
+	ledgersProcessor := processors.NewLedgerProcessor(s.historyQ.NewLedgerBatchInsertBuilder(), CurrentVersion)
```
tamirms, line 332: why is `ledgersProcessor` created in this function as opposed to `buildTransactionProcessor()` where all the other transaction processors are created?

services/horizon/internal/ingest/processors/ledgers_processor.go
```diff
@@ -10,7 +10,7 @@ import (
 	"github.com/stellar/go/xdr"
 )
 
-type ledgerInfo struct {
+type LedgerInfo struct {
```
tamirms, line 13: why is this exported now?

support/db/internal_test.go
```diff
@@ -7,6 +7,7 @@ const testSchema = `
 CREATE TABLE  IF NOT EXISTS people (
     name character varying NOT NULL,
     hunger_level integer NOT NULL,
+	json_value jsonb,
```
tamirms, line 10: nit: indentation is unaligned

services/horizon/internal/db2/history/account_loader.go
```diff
@@ -207,3 +211,7 @@ func NewAccountLoaderStub() AccountLoaderStub {
 func (a AccountLoaderStub) Insert(address string, id int64) {
 	a.Loader.ids[address] = id
 }
+
+func (a AccountLoaderStub) Sealed() {
```
tamirms, line 215: I don't see the purpose of this function. I only see it used in services/horizon/internal/ingest/processors/trades_processor_test.go but I don't think it's necessary

services/horizon/internal/db2/history/asset_loader.go
tamirms: @sreuland that sounds good! I reviewed the PR and left some comments but overall the PR looks great!

services/horizon/internal/ingest/processor_runner.go
sreuland: line 333 is the reason, it invokes the ledger processor via new `LedgersProcessor.ProcessLedger()` to populate it's internal state with the current seqnum up front, regardless of number of tx's in the ledger, even if there are 0, since the processor flow on `StreamLedgerTransactions` only invokes `ProcessTransaction` for all processors in the group if the ledger had at least on tx, it allowed an edge case in ingestion, where empty ledgers would not add a row to history_ledgers table.

services/horizon/internal/ingest/processors/ledgers_processor.go
sreuland: oops, revereted that, thx.

services/horizon/internal/db2/history/fee_bump_scenario.go
sreuland: yes, removed.

services/horizon/internal/ingest/processor_runner.go
tamirms: oh I see, that's a really good catch! Is this a pre-existing bug that's also present in the master branch?

services/horizon/internal/db2/history/account_loader.go
sreuland: this was done to enable unit tests in trades_procesor_test, which have already established a path of instrumenting a trades processor instance under test with mocks for everything and then asserting the processor output on `ProcessTransaction()` and `Flush()`, to accomplish running both of those methods in sequence with mocks, the test has to manually transition the loader state to sealed, as the test is not running the processor through process_runner which does that.

tamirms: ok, what do you think about making the `Insert()` function set `sealed` to true? Then we can avoid having a `Sealed()` function that is used in trades_procesor_test.

services/horizon/internal/ingest/processor_runner.go
sreuland: no, it's a new aspect related to re-org of processors flow with the runner, on master, the LedgerProcessor instance is created with the xdr.Ledger, and runner always executes LedgerProcessor.Commit() which inserts the ledger to ledger_history, whereas, in this new design, LedgerProcessor instance is not created with a Ledger, instead LedgerProcessor.ProcessTransaction() was adjusted to add a ledger to the internal map which is used in Flush(), but that was where it skipped if no Tx's, so, I adjusted slightly, adding LedgerProcessor.ProcessLedger() which runner invokes always per ledger, before ProcessTransaction's.

services/horizon/internal/db2/history/account_loader.go
sreuland: yes, that sounds good, done - https://github.com/stellar/go/pull/5083/commits/625ce1f5f983a9cebb22e7f3bd9d9810c6a3a4fe

services/horizon/docker/verify-range/dependencies
```diff
@@ -11,13 +11,13 @@ echo "deb https://apt.stellar.org $(lsb_release -cs) stable" | sudo tee -a /etc/
...
+tar -C /usr/local -xzf go1.20.linux-amd64.tar.gz
 
 git clone https://github.com/stellar/go.git stellar-go
 cd stellar-go
 # By default "git fetch" only fetches refs/<branchname>
 # Below ensures we also fetch PR refs
 git config --add remote.origin.fetch "+refs/pull/*/head:refs/remotes/origin/pull/*"
 git fetch --force --quiet origin
-/usr/local/go/bin/go build -v ./services/horizon
+/usr/local/go/bin/go build -v ./services/horizon/.
```
tamirms, line 23: looks like there was a trailing `/.` added to this line. Is that intentional?

services/horizon/internal/ingest/processor_runner.go
```diff
@@ -135,24 +134,37 @@ func buildChangeProcessor(
 func (s *ProcessorRunner) buildTransactionProcessor(
 	ledgerTransactionStats *processors.StatsLedgerTransactionProcessor,
 	tradeProcessor *processors.TradeProcessor,
-	ledger xdr.LedgerHeaderHistoryEntry,
+	ledgersProcessor *processors.LedgersProcessor,
+	ledger xdr.LedgerCloseMeta,
```
tamirms, line 138: `(*ProcessorRunner).buildTransactionProcessor` - `ledger` is unused (unparam)

tamirms (approved):
🎉

services/horizon/docker/verify-range/dependencies
sreuland: yes, the result in either form is the same horizon binary being built.

services/horizon/internal/ingest/processor_runner.go
sreuland: good eye, interesting, VS Code didn't warn on that nor did staticcheck, fixed it. thanks.

PR #5088, opened 2023-10-23
Fix Go dependencies

2opremio:
https://github.com/stellar/go/pull/5084 broke the Go dependencies, because:

1. `go mod tidy` wasn't invoked before merging leading to a misaligned Go version in go.mod (we should probably enforce this in CI and make sure there are no diffs)
2. It ugraded github.com/creachadair/jrpc2 to 1.1.1 which requires Go 1.21 (breaking the two-Go-release compatibility guarantee we make)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/stellar/go
 
-go 1.19
+go 1.20
 
 require (
 	firebase.google.com/go v3.12.0+incompatible
@@ -11,7 +11,7 @@ require (
 	github.com/adjust/goautoneg v0.0.0-20150426214442-d788f35a0315
 	github.com/asaskevich/govalidator v0.0.0-20230301143203-a9d515a09cc2
 	github.com/aws/aws-sdk-go v1.45.26
-	github.com/creachadair/jrpc2 v1.1.1
+	github.com/creachadair/jrpc2 v1.1.0
 	github.com/elazarl/go-bindata-assetfs v1.0.1
 	github.com/getsentry/raven-go v0.2.0
 	github.com/go-chi/chi v4.1.2+incompatible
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/stellar/go
 
-go 1.19
+go 1.20
```
2opremio, line 3: it used to be 1.21 when running `go mod tidy`

2opremio: I actually changed CI in this PR. Why not.

tsachiherman:
I'm surprised you're saying that `go mod tidy` wasn't called. I did call that, and deployed 1.19 before doing that.

Also, if it broke the build, how come all the CI tests passed ?!

2opremio:
It wasn't, since otherwise it would had written 1.21 to go.mod

2opremio:
> Also, if it broke the build, how come all the CI tests passed ?!

Because go updates go.mod when building

stellarsaur (approved):

tsachiherman:
> > Also, if it broke the build, how come all the CI tests passed ?!
> 
> I never wrote it broke the build, it broke the two-Golang release policy. I guess CI works because Go updates go.mod when building with 1.21 (or does a best effort when using an earlier version)

that's a bit concerning. should we add a test to verify that the go.mod/go.sum doesn't get updated during the build ? i.e. so that it's only developer-driven ?

2opremio:
> that's a bit concerning. should we add a test to verify that the go.mod/go.sum doesn't get updated during the build ? i.e. so that it's only developer-driven ?

Yes, I suggested that in the PR description:

> we should probably enforce this in CI and make sure there are no diffs

leighmcculloch:
> I guess CI works because Go updates go.mod when building with 1.21

This is surprising. Go since version 1.16 has not updated the go.mod automatically.

In the release notes for Go 1.16 it says:

> Build commands like go build and go test no longer modify go.mod and go.sum by default. Instead, they report an error if a module requirement or checksum needs to be added or updated

Ref: https://go.dev/doc/go1.16

This was strengthened in Go 1.18 with additional commands being changed to not modify go.mod:

> **Automatic go.mod and go.sum updates**
>
> The go mod graph, go mod vendor, go mod verify, and go mod why subcommands no longer automatically update the go.mod and go.sum files. (Those files can be updated explicitly using go get, go mod tidy, or go mod download.)

Ref: https://go.dev/doc/go1.18

I think something else is probably happening.

PR #5092, opened 2023-10-26
services/horizon: Explicitly supply objectives in Prometheus SummaryOpts to fix metrics dashboards

urvisavla:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This PR explicitly supplies objectives in `Prometheus.SummaryOpts` because the new version of the Prometheus go client no longer provides the default value.

### Why

In previous versions of the Prometheus go client, default objectives were automatically provided in `prometheus.SummaryOpts`. However, this was [deprecated](
https://github.com/prometheus/client_golang/pull/262/files
) in [version 1.0.0.](https://github.com/prometheus/client_golang/blob/main/CHANGELOG.md#100--2019-06-15). As a result, when we [updated](https://github.com/stellar/go/commit/d50c63e68acca8e2c59e2a8c53d167665d18e7e9#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6L35-L36) the dependencies in our repo, certain dashboards, such as the "Top 20 slowest routes" and "Occurrences of DB query taking longer than 5 seconds", stopped showing any data. This is because Horizon was unable to post "summary" metrics type without objectives.

This PR addresses this issue by explicitly supplying objectives for all summary metrics.

### Known limitations
N/A

### DIFF
--- a/ingest/ledgerbackend/metrics.go
+++ b/ingest/ledgerbackend/metrics.go
@@ -17,7 +17,8 @@ func WithMetrics(base LedgerBackend, registry *prometheus.Registry, namespace st
 	summary := prometheus.NewSummary(
 		prometheus.SummaryOpts{
 			Namespace: namespace, Subsystem: "ingest", Name: "ledger_fetch_duration_seconds",
-			Help: "duration of fetching ledgers from ledger backend, sliding window = 10m",
+			Help:       "duration of fetching ledgers from ledger backend, sliding window = 10m",
+			Objectives: map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.99: 0.001},
 		},
 	)
 	registry.MustRegister(summary)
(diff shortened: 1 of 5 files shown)


tamirms (approved):
looks good! can you also implement this fix for soroban-rpc in https://github.com/stellar/soroban-tools ?

tsachiherman (commented):
The PR looks great; one thing that I'd like to suggest is to define a global map that would include these default objectives, and use that everywhere.

urvisavla:
> The PR looks great; one thing that I'd like to suggest is to define a global map that would include these default objectives, and use that everywhere.

It's being used in many different packages, making it difficult to place it in a centralized location

PR #5096, opened 2023-10-31
Merge latest master to ingestion-next

sreuland:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

merged latest master to ingestion-next, required some changes to fast batch builder for converting byte[] columnar values to string to avoid postgres driver issue.

Related to #4909 

### Why

fix the pending merge conflicts that awaited on master, as it has been a while since last updated


services/horizon/internal/ingest/processors/effects_processor.go
```diff
@@ -1426,3 +1460,115 @@ func (e *effectsWrapper) addLiquidityPoolWithdrawEffect() error {
...
+
+	source := e.operation.SourceAccount()
+	for _, event := range events {
+		evt, err := contractevents.NewStellarAssetContractEvent(&event, e.operation.network)
+		if err != nil {
+			continue // irrelevant or unsupported event
+		}
+
+		details := make(map[string]interface{}, 4)
+		addAssetDetails(details, evt.GetAsset(), "")
```
tamirms, line 1480: Error return value is not checked (errcheck)

services/horizon/internal/ingest/processors/effects_processor.go
```diff
@@ -1426,3 +1460,115 @@ func (e *effectsWrapper) addLiquidityPoolWithdrawEffect() error {
...
+		case contractevents.EventTypeTransfer:
+			transferEvent := evt.(*contractevents.TransferEvent)
+			details["amount"] = amount.String128(transferEvent.Amount)
+			toDetails := map[string]interface{}{}
+			for key, val := range details {
+				toDetails[key] = val
+			}
+
+			if strkey.IsValidEd25519PublicKey(transferEvent.From) {
+				e.add(
```
tamirms, line 1499: Error return value is not checked (errcheck)

services/horizon/internal/ingest/processors/effects_processor.go
```diff
@@ -1426,3 +1460,115 @@ func (e *effectsWrapper) addLiquidityPoolWithdrawEffect() error {
...
+					history.EffectAccountDebited,
+					details,
+				)
+			} else {
+				details["contract"] = transferEvent.From
+				e.addMuxed(source, history.EffectContractDebited, details)
+			}
+
+			if strkey.IsValidEd25519PublicKey(transferEvent.To) {
+				e.add(
```
tamirms, line 1511: Error return value is not checked (errcheck)

services/horizon/internal/ingest/processors/effects_processor.go
```diff
@@ -1426,3 +1460,115 @@ func (e *effectsWrapper) addLiquidityPoolWithdrawEffect() error {
...
+					transferEvent.To,
+					null.String{},
+					history.EffectAccountCredited,
+					toDetails,
+				)
+			} else {
+				toDetails["contract"] = transferEvent.To
+				e.addMuxed(source, history.EffectContractCredited, toDetails)
```
tamirms, line 1519: Error return value is not checked (errcheck)

support/db/fast_batch_insert_builder.go
```diff
@@ -58,6 +58,10 @@ func (b *FastBatchInsertBuilder) Row(row map[string]interface{}) error {
 		if !ok {
 			return errors.Errorf(`column "%s" does not exist`, column)
 		}
+
+		if b, ok := val.([]byte); ok {
+			val = string(b)
```
tamirms, line 63: I think we only want to do this for json/jsonb columns. Could you check that this code will still correctly insert rows when the schema includes [bytea](https://www.postgresql.org/docs/current/datatype-binary.html) columns?

tamirms:
The "Horizon / Test (and push) verify-range image (pull_request)" CI check is failing

services/horizon/internal/ingest/processors/effects_processor.go
sreuland: I checked blame, this change comes from master several months ago, i'm fine with cleaning here, btw, what utility are you using to catch these? staticcheck.sh not tripping on it, and VSCode IDE not showing me a warn either.

support/db/fast_batch_insert_builder.go
sreuland: it passes on all unit/integration tests(locally, will see why not here), the only table in schema with a column of this type is exp_asset_stats, which is not a table updated by fast batch builder as of yet, which is limited to history tables for effects, cb's, lp's, ops, tx's. 

I took this approach to avoid scope creep on getting to work with latest master. Initially, this can work for now, but i think should create a ticket to convert to pgx driver usage instead for longer term, I tried that locally, but it leads to wider code change than I think we want to introduce, namely the SessionInterface needs to expose a sql.Conn reference that pgx can wrap over, or just convert all sql.Conn usages to pass 'pgx' instead of 'postgres' for driver dialect, but that seems best isolated in it's own follow-on refactor/change.

services/horizon/internal/ingest/processors/effects_processor.go
tamirms: yeah, the error return values were not checked in master. But, the ingestion-next branch explicitly fixed these warnings in earlier PRs. we should fix them here and then it will be fixed when merging back to master.

> what utility are you using to catch these?

These warnings were generated by [golangci-lint](https://github.com/stellar/go/actions/runs/6711501505/job/18239005949?pr=5096)

support/db/fast_batch_insert_builder.go
tamirms: I agree that using pgx would require a lot of work. But I think this issue can be fixed in the short term without much effort. We can convert byte slices to strings before passing them into `Row()` or `RowStruct()`.

In your PR you do the conversion automatically inside the `Row()` and `RowStruct()` functions. But the problem with that approach is that it might not work for [bytea](https://www.postgresql.org/docs/current/datatype-binary.html) columns. So I think it would be better to convert the byte slice strings in the code which calls `Row()` / `RowStrct()`, e.g. https://github.com/stellar/go/blob/59cc1f0d3d69362a7523de30f088c8663887de8b/services/horizon/internal/db2/history/effect_batch_insert_builder.go#L54

sreuland: https://github.com/stellar/go/pull/5096/commits/769a287371478837d1d65235f911e41fd5da3a25

tamirms (approved):
looks good! once  "Horizon / Test (and push) verify-range image (pull_request)" is fixed I think we can merge

PR #5100, opened 2023-11-01
Remove code related to legacy ingestion support

aditya1702:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Removes configuration flags (and related tests) supporting legacy ingestion:
- stellar-core-db-url
- enable-captive-core-ingestion

Note: The flag definitions are still present and a deprecation warning is added to the command-line.

### Why

As part of  #4855, we want to remove support for legacy ingestion from Horizon. 
- Ops team has successfully shut down all legacy ingestors on public Horizon - https://github.com/stellar/ops/issues/2557
- As part of partner experience revamp, we have removed all references of above config variables from docs - https://github.com/stellar/stellar-docs/pull/249

### Known limitations

N/A


services/horizon/internal/integration/parameters_test.go
```diff
@@ -55,6 +53,8 @@ const (
 		PUBLIC_KEY="GD5KD2KEZJIGTC63IGW6UMUSMVUVG5IHG64HUTFWCHVZH2N2IBOQN7PS"
 		ADDRESS="localhost"
 		QUALITY="MEDIUM"`
+
+	StellarCoreUrl = "http://localhost:11626"
```
stellarsaur, line 57: Nit: StellarCoreUrl -> StellarCoreURL

stellarsaur (approved):

services/horizon/internal/flags.go
```diff
@@ -916,25 +891,19 @@ func ApplyFlags(config *Config, flags support.ConfigOptions, options ApplyOption
...
-			if err != nil {
-				return errors.Wrap(err, "error generating captive core configuration")
-			}
+		err := setCaptiveCoreConfiguration(config, options)
+		if err != nil {
+			return errors.Wrap(err, "error generating captive core configuration")
 		}
 	} else {
-		if config.EnableCaptiveCoreIngestion && (config.CaptiveCoreBinaryPath != "" || config.CaptiveCoreConfigPath != "") {
+		if config.CaptiveCoreBinaryPath != "" || config.CaptiveCoreConfigPath != "" {
```
urvisavla, line 899: I think we can remove this 'else' block, as ingest is set to true by default so it doesn't provide much value.

services/horizon/cmd/ingest.go
```diff
@@ -315,26 +289,13 @@ var ingestInitGenesisStateCmd = &cobra.Command{
...
-				return fmt.Errorf("flag --%s cannot be empty", horizon.StellarCoreDBURLFlagName)
-			}
-
-			coreSession, dbErr := db.Open("postgres", globalConfig.StellarCoreDatabaseURL)
-			if dbErr != nil {
-				return fmt.Errorf("cannot open Core DB: %v", dbErr)
-			}
-			ingestConfig.CoreSession = coreSession
-		}
+		ingestConfig.CaptiveCoreBinaryPath = globalConfig.CaptiveCoreBinaryPath
```
urvisavla, line 297: Should this be moved to when creating the object above (line 288)?

services/horizon/cmd/ingest.go
```diff
@@ -315,26 +289,13 @@ var ingestInitGenesisStateCmd = &cobra.Command{
...
+		ingestConfig.CaptiveCoreConfigUseDB = globalConfig.CaptiveCoreConfigUseDB
```
urvisavla, line 298: same as above

services/horizon/docker/verify-range/start
```diff
@@ -72,9 +72,7 @@ dump_horizon_db() {
 export NETWORK_PASSPHRASE="Public Global Stellar Network ; September 2015"
 export HISTORY_ARCHIVE_URLS="https://s3-eu-west-1.amazonaws.com/history.stellar.org/prd/core-live/core_live_001"
 export DATABASE_URL="postgres://postgres:postgres@localhost:5432/horizon?sslmode=disable"
-export CAPTIVE_CORE_CONFIG_APPEND_PATH="/captive-core-pubnet.cfg" 
-export STELLAR_CORE_BINARY_PATH="/usr/bin/stellar-core"
```
urvisavla, line 76: Any reason for removing this env variable here?

services/horizon/internal/integration/parameters_test.go
```diff
@@ -99,34 +99,34 @@ func TestEnvironmentPreserved(t *testing.T) {
 	// running an integration test.
 
 	// Note that we ALSO need to make sure we don't modify parent env state.
-	value, isSet := os.LookupEnv("CAPTIVE_CORE_CONFIG_PATH")
+	value, isSet := os.LookupEnv("STELLAR_CORE_URL")
```
urvisavla, line 102: Any reason for changing the env variable used in this test?

services/horizon/docker/verify-range/start
aditya1702: Ah removed this by mistake. Confused it with STELLAR_CORE_DB_URL

services/horizon/internal/flags.go
```diff
@@ -250,15 +246,6 @@ func Flags() (*Config, support.ConfigOptions) {
 			ConfigKey:      &config.CaptiveCoreConfigUseDB,
 			UsedInCommands: IngestionCommands,
 		},
-		&support.ConfigOption{
```
urvisavla, line 253: Because removing the flag here could break for users who explicitly pass the flag, I'm wondering if we deprecate the flag instead of removing it outright and add a warning that the flag is deprecated and that captive-core ingestion is enabled by default. We can remove the flag in the future.

sreuland: yes, agreed, need the gentler deprecation approach for a phase-out, I think the same appriacg has already been done on other legacy ingestion related flags like `stellar-core-db-url`

services/horizon/internal/flags.go
aditya1702: @urvisavla While INGEST will be true by default, aren't we also recommending a clear separation between keep ingesting instances and transaction submission instances separate using DISABLE_TX_SUB = True, INGEST = false.

services/horizon/internal/integration/parameters_test.go
aditya1702: INGEST is set to false in the tests, so specifying CAPTIVE_CORE_CONFIG_PATH  triggers the above else condition and fails the test.

services/horizon/internal/flags.go
aditya1702: Yeah agree with the above points.

services/horizon/internal/flags.go
urvisavla: I think this check is to warn users if they have forgotten to set the ingest flag. But, since the flag is now set by default and users must explicitly set it to False, I'm not sure the error message is helpful. May be the error message should be something like "you have disabled ingestion but are still passing captive-core config parameters, which are not required....". wdyt?

services/horizon/internal/integration/parameters_test.go
urvisavla: 👍

services/horizon/internal/flags.go
aditya1702: @urvisavla Yeah I think thats a better error message.

services/horizon/internal/flags.go
```diff
@@ -251,12 +250,21 @@ func Flags() (*Config, support.ConfigOptions) {
...
-			ConfigKey:      &config.EnableCaptiveCoreIngestion,
+			Name:        EnableCaptiveCoreIngestionFlagName,
+			OptType:     types.String,
+			FlagDefault: "",
+			Required:    false,
+			Hidden:      true,
+			CustomSetValue: func(opt *support.ConfigOption) error {
+				if val := viper.GetString(opt.Name); val != "" {
+					stdLog.Printf(
+						"DEPRECATED - The usage of the flag --enable-captive-core-ingestion has been deprecated. " +
```
sreuland, line 261: can an assertion be added to parameters_test to verify this message is output to console when the parameter is present, is the expectation to emit this message when it's present on command line or environment variable or either?

services/horizon/internal/integration/parameters_test.go
```diff
@@ -531,6 +527,45 @@ func TestDeprecatedOutputs(t *testing.T) {
...
+				return
+			}
+		}()
+
+		outputBytes, _ := io.ReadAll(r)
+		wg.Wait() // Wait for the goroutine to finish before proceeding
+		_ = r.Close()
+		os.Stderr = originalStderr
+
+		assert.Contains(t, string(outputBytes), "DEPRECATED - The usage of the flag --stellar-core-db-url has been deprecated. "+
```
sreuland, line 562: this is great coverage for command line usage, how about when the same params are present as environment variables, is the expected outcome the same behavior of printing these messages to console or different? Can the test demonstrate that outcome with env variables here also?

aditya1702: @sreuland Yes the outcome would be the same. I added a similar test for testing the env variables

sreuland (approved):
nice work!

Issue #5086, opened 2023-10-19
Use COPY to speed up ClaimableBalancesChangeProcessor


tamirms:
In https://github.com/stellar/go/issues/4908 we observed that inserting rows is much faster using the postgres `COPY` command. We can use this property to improve the performance of the [ClaimableBalancesChangeProcessor](https://github.com/stellar/go/blob/d50c63e68acca8e2c59e2a8c53d167665d18e7e9/services/horizon/internal/ingest/processors/claimable_balances_change_processor.go#L12) . The [ClaimableBalancesChangeProcessor](https://github.com/stellar/go/blob/d50c63e68acca8e2c59e2a8c53d167665d18e7e9/services/horizon/internal/ingest/processors/claimable_balances_change_processor.go#L12) upserts rows into the `claimable_balances` and `claimable_balance_claimants` tables.

However, because claimable balances are either created or removed it turns out that we never have to modify existing  rows of `claimable_balances` and `claimable_balance_claimants`. The create claimable balance operation creates a claimable balance and the claim operation removes the claimable balance from the ledger. There is no way to mutate a claimable balance once created.

This means we can update the [ClaimableBalancesChangeProcessor](https://github.com/stellar/go/blob/d50c63e68acca8e2c59e2a8c53d167665d18e7e9/services/horizon/internal/ingest/processors/claimable_balances_change_processor.go#L12) to use COPY to insert into the the `claimable_balances` and `claimable_balance_claimants` tables. We expect that to be faster than the current code which bulk inserts rows into the `claimable_balances` and `claimable_balance_claimants` tables.

tamirms:
We can also use COPY to speed up inserts in the [SignersProcessor](https://github.com/stellar/go/blob/d50c63e68acca8e2c59e2a8c53d167665d18e7e9/services/horizon/internal/ingest/processors/signers_processor.go#L18)

urvisavla:
When I switched the [claimable balance processor](https://github.com/stellar/go/pull/5085) to use [FastBatchInsertBuilder](https://github.com/stellar/go/blob/9ce82de12cbade47de862ad5e9d0684029558564/support/db/fast_batch_insert_builder.go),' which uses ‘COPY' instead of 'INSERT,' it worked in my local environment but it failed on the CI pipeline with the following error in json parsing `pq: invalid input syntax for type json`.

[...]

Issue #5099, opened 2023-11-01
Ingest batches of ledgers in-memory before flushing to DB 


tamirms:
In https://github.com/stellar/go/issues/4909 we have updated all the transaction processors to use the FastBatchInsertBuilder to insert rows into the history tables with the postgres COPY command. We also refactored the transaction processor interface to allow a single processor to accumulate records across multiple ledgers:

```go
type horizonTransactionProcessor interface {
	ProcessTransaction(xdr.LedgerCloseMeta, ingest.LedgerTransaction) error
	Flush(ctx context.Context, session db.SessionInterface) error
}
```

Now we can build on this work to improve the performance of reingestion. https://github.com/stellar/go/issues/4909 improved the performance of ingesting a single ledger. But we can extract more performance gains by ingesting multiple ledgers within a single processor lifetime.

In the [spike branch](https://github.com/stellar/go/compare/master...tamirms:go:ingest-perf-pq) the dataflow for ingesting batches of ledgers within a single processor lifetime is:

```go
	accountLoader := history.NewAccountLoader()
	cbLoader := history.NewClaimableBalanceLoader()
	lpLoader := history.NewLiquidityPoolLoader()
	assetLoader := history.NewAssetLoader()
	processors := buildTransactionProcessors(
		s.historyQ,
		accountLoader,
		cbLoader,
		lpLoader,
		assetLoader,
	)

       // apply all the ledgers in the batch on the processors
       for _, ledger := range ledgers {
		if err = s.runner.ApplyProcessorsOnLedger(processors, ledgerCloseMeta); err != nil {
			return err
		}
       }

       // use the loaders to lookup all the accounts, assets, claimable balances, and liquidity pools registered
       // by the processors
       err = func() error {
		if err := s.historyQ.Begin(); err != nil {
			return errors.Wrap(err, "Error starting a transaction")
		}
		defer s.historyQ.Rollback()

		if err := accountLoader.Exec(s.ctx, s.historyQ); err != nil {
			return err
		}
		if err := cbLoader.Exec(s.ctx, s.historyQ); err != nil {
			return err
		}
		if err := lpLoader.Exec(s.ctx, s.historyQ); err != nil {
			return err
		}
		if err := assetLoader.Exec(s.ctx, s.historyQ); err != nil {
			return err
		}
		if err := s.historyQ.Commit(); err != nil {
			return errors.Wrap(err, commitErrMsg)
		}
		return nil
	}()

[...]

Issue #5098, opened 2023-11-01
Use COPY to speed up all ChangeProcessors


tamirms:
In https://github.com/stellar/go/issues/5086 we improved performance of the ClaimableBalancesChangeProcessor by using COPY to bulk insert claimable balance rows. We should use COPY to improve the performance of all the remaining change processors.

Most of the change processors combine new ledger entries and updates of existing ledger entries into a single list which is then upserted into the DB. We should change this workflow so that we handle new ledger entries separately from updating existing ledger entries because we can bulk insert new ledger entries efficiently with COPY.

The claimable balance change processor has already been taken care of in https://github.com/stellar/go/issues/5086 and the asset stats processor is still in flux because of https://github.com/stellar/go/pull/5033 . So we should implement the COPY optimization on the following remaining change processors:  

- [x] [SignersProcessor](https://github.com/stellar/go/blob/d50c63e68acca8e2c59e2a8c53d167665d18e7e9/services/horizon/internal/ingest/processors/signers_processor.go#L14)
- [x] [AccountDataProcessor](https://github.com/stellar/go/blob/d50c63e68acca8e2c59e2a8c53d167665d18e7e9/services/horizon/internal/ingest/processors/account_data_processor.go#L12)
- [x] [AccountsProcessor](https://github.com/stellar/go/blob/d50c63e68acca8e2c59e2a8c53d167665d18e7e9/services/horizon/internal/ingest/processors/accounts_processor.go#L13)
- [x] [LiquidityPoolsChangeProcessor](https://github.com/stellar/go/blob/d50c63e68acca8e2c59e2a8c53d167665d18e7e9/services/horizon/internal/ingest/processors/liquidity_pools_change_processor.go)
- [x] [OffersProcessor](https://github.com/stellar/go/blob/d50c63e68acca8e2c59e2a8c53d167665d18e7e9/services/horizon/internal/ingest/processors/offers_processor.go#L16)
- [x] [TrustLinesProcessor](https://github.com/stellar/go/blob/d50c63e68acca8e2c59e2a8c53d167665d18e7e9/services/horizon/internal/ingest/processors/trust_lines_processor.go#L12)

I anticipate that once COPY is used in all the change processors, ingesting from the history archives will be significantly faster, which means that state rebuilds will much faster than before.

hethm999:
ok

PR #5104, opened 2023-11-02
services/horizon: Use COPY to speed up ClaimableBalanceChangeProcessor

urvisavla:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Use [FastBatchInsertBuilder](https://github.com/stellar/go/blob/9ce82de12cbade47de862ad5e9d0684029558564/support/db/fast_batch_insert_builder.go) for writing to claimable_balances and claimable_balance_claimants tables. FastBatchInsertBuilder uses ‘COPY' to insert into the db which is faster than using 'INSERT'.

### Why
https://github.com/stellar/go/issues/5086

### Known limitations
N/A

### DIFF
--- a/services/horizon/internal/db2/history/claimable_balance_batch_insert_builder.go
+++ b/services/horizon/internal/db2/history/claimable_balance_batch_insert_builder.go
@@ -1,0 +1,47 @@
+package history
+
+import (
+	"context"
+
+	"github.com/stellar/go/support/db"
+	"github.com/stellar/go/xdr"
+)
+
+// ClaimableBalanceBatchInsertBuilder is used to insert claimable balance into the
+// claimable_balances table
+type ClaimableBalanceBatchInsertBuilder interface {
+	Add(claimableBalance ClaimableBalance) error
+	Exec(ctx context.Context, session db.SessionInterface) error
+	Reset() error
+}
+
+// ClaimableBalanceBatchInsertBuilder is a simple wrapper around db.FastBatchInsertBuilder
+type claimableBalanceBatchInsertBuilder struct {
+	encodingBuffer *xdr.EncodingBuffer
+	builder        db.FastBatchInsertBuilder
+	table          string
+}
+
+// NewClaimableBalanceBatchInsertBuilder constructs a new ClaimableBalanceBatchInsertBuilder instance
+func (q *Q) NewClaimableBalanceBatchInsertBuilder() ClaimableBalanceBatchInsertBuilder {
+	return &claimableBalanceBatchInsertBuilder{
+		encodingBuffer: xdr.NewEncodingBuffer(),
+		builder:        db.FastBatchInsertBuilder{},
+		table:          "claimable_balances",
+	}
+}
+
+// Add adds a new claimable balance to the batch
+func (i *claimableBalanceBatchInsertBuilder) Add(claimableBalance ClaimableBalance) error {
+	return i.builder.RowStruct(claimableBalance)
+}
+
+// Exec inserts claimable balance rows to the database
+func (i *claimableBalanceBatchInsertBuilder) Exec(ctx context.Context, session db.SessionInterface) error {
+	return i.builder.Exec(ctx, session, i.table)
+}
+
+func (i *claimableBalanceBatchInsertBuilder) Reset() error{
+	i.builder.Reset()
+	return nil
+}
--- a/services/horizon/internal/db2/history/claimable_balance_batch_insert_builder.go
+++ b/services/horizon/internal/db2/history/claimable_balance_batch_insert_builder.go
@@ -41,7 +41,7 @@ func (i *claimableBalanceBatchInsertBuilder) Exec(ctx context.Context, session d
 	return i.builder.Exec(ctx, session, i.table)
 }
 
-func (i *claimableBalanceBatchInsertBuilder) Reset() error{
+func (i *claimableBalanceBatchInsertBuilder) Reset() error {
 	i.builder.Reset()
 	return nil
 }
--- a/services/horizon/internal/db2/history/claimable_balance_batch_insert_builder.go
+++ b/services/horizon/internal/db2/history/claimable_balance_batch_insert_builder.go
@@ -36,7 +36,7 @@ func (i *claimableBalanceBatchInsertBuilder) Add(claimableBalance ClaimableBalan
 	return i.builder.RowStruct(claimableBalance)
 }
 
-// Exec inserts claimable balance rows to the database
+// Exec writes the batch of claimable balances to the database.
 func (i *claimableBalanceBatchInsertBuilder) Exec(ctx context.Context, session db.SessionInterface) error {
 	return i.builder.Exec(ctx, session, i.table)
 }
--- a/services/horizon/internal/db2/history/claimable_balance_batch_insert_builder.go
+++ b/services/horizon/internal/db2/history/claimable_balance_batch_insert_builder.go
@@ -4,30 +4,28 @@ import (
 	"context"
 
 	"github.com/stellar/go/support/db"
-	"github.com/stellar/go/xdr"
 )
 
 // ClaimableBalanceBatchInsertBuilder is used to insert claimable balance into the
 // claimable_balances table
 type ClaimableBalanceBatchInsertBuilder interface {
 	Add(claimableBalance ClaimableBalance) error
-	Exec(ctx context.Context, session db.SessionInterface) error
-	Reset()
+	Exec(ctx context.Context) error
 }
 
 // ClaimableBalanceBatchInsertBuilder is a simple wrapper around db.FastBatchInsertBuilder
 type claimableBalanceBatchInsertBuilder struct {
-	encodingBuffer *xdr.EncodingBuffer
-	builder        db.FastBatchInsertBuilder
-	table          string
+	session db.SessionInterface
+	builder db.FastBatchInsertBuilder
+	table   string
 }
 
 // NewClaimableBalanceBatchInsertBuilder constructs a new ClaimableBalanceBatchInsertBuilder instance
-func (q *Q) NewClaimableBalanceBatchInsertBuilder() ClaimableBalanceBatchInsertBuilder {
+func (q *Q) NewClaimableBalanceBatchInsertBuilder(session db.SessionInterface) ClaimableBalanceBatchInsertBuilder {
 	return &claimableBalanceBatchInsertBuilder{
-		encodingBuffer: xdr.NewEncodingBuffer(),
-		builder:        db.FastBatchInsertBuilder{},
-		table:          "claimable_balances",
+		session: session,
+		builder: db.FastBatchInsertBuilder{},
+		table:   "claimable_balances",
 	}
 }
 
@@ -37,11 +35,6 @@ func (i *claimableBalanceBatchInsertBuilder) Add(claimableBalance ClaimableBalan
 }
 
 // Exec writes the batch of claimable balances to the database.
-func (i *claimableBalanceBatchInsertBuilder) Exec(ctx context.Context, session db.SessionInterface) error {
-	return i.builder.Exec(ctx, session, i.table)
-}
-
-// Reset clears out the current batch of claimable balances
-func (i *claimableBalanceBatchInsertBuilder) Reset() {
-	i.builder.Reset()
+func (i *claimableBalanceBatchInsertBuilder) Exec(ctx context.Context) error {
+	return i.builder.Exec(ctx, i.session, i.table)
 }
--- a/services/horizon/internal/db2/history/claimable_balance_batch_insert_builder.go
+++ b/services/horizon/internal/db2/history/claimable_balance_batch_insert_builder.go
@@ -12,7 +12,7 @@ import (
 type ClaimableBalanceBatchInsertBuilder interface {
 	Add(claimableBalance ClaimableBalance) error
 	Exec(ctx context.Context, session db.SessionInterface) error
-	Reset() error
+	Reset()
 }
 
 // ClaimableBalanceBatchInsertBuilder is a simple wrapper around db.FastBatchInsertBuilder
@@ -41,7 +41,7 @@ func (i *claimableBalanceBatchInsertBuilder) Exec(ctx context.Context, session d
 	return i.builder.Exec(ctx, session, i.table)
 }
 
-func (i *claimableBalanceBatchInsertBuilder) Reset() error {
+// Reset clears out the current batch of claimable balances
+func (i *claimableBalanceBatchInsertBuilder) Reset() {
 	i.builder.Reset()
-	return nil
 }
--- a/services/horizon/internal/db2/history/claimable_balance_batch_insert_builder.go
+++ b/services/horizon/internal/db2/history/claimable_balance_batch_insert_builder.go
@@ -21,9 +21,9 @@ type claimableBalanceBatchInsertBuilder struct {
 }
 
 // NewClaimableBalanceBatchInsertBuilder constructs a new ClaimableBalanceBatchInsertBuilder instance
-func (q *Q) NewClaimableBalanceBatchInsertBuilder(session db.SessionInterface) ClaimableBalanceBatchInsertBuilder {
+func (q *Q) NewClaimableBalanceBatchInsertBuilder() ClaimableBalanceBatchInsertBuilder {
 	return &claimableBalanceBatchInsertBuilder{
-		session: session,
+		session: q,
 		builder: db.FastBatchInsertBuilder{},
 		table:   "claimable_balances",
 	}
--- a/services/horizon/internal/db2/history/claimable_balance_claimant_batch_insert_builder.go
+++ b/services/horizon/internal/db2/history/claimable_balance_claimant_batch_insert_builder.go
@@ -10,33 +10,38 @@ import (
 // ClaimableBalanceClaimantBatchInsertBuilder is used to insert transactions into the
 // history_transactions table
 type ClaimableBalanceClaimantBatchInsertBuilder interface {
-	Add(ctx context.Context, claimableBalanceClaimant ClaimableBalanceClaimant) error
-	Exec(ctx context.Context) error
+	Add(claimableBalanceClaimant ClaimableBalanceClaimant) error
+	Exec(ctx context.Context, session db.SessionInterface) error
+	Reset() error
 }
 
-// ClaimableBalanceClaimantBatchInsertBuilder is a simple wrapper around db.BatchInsertBuilder
+// ClaimableBalanceClaimantBatchInsertBuilder is a simple wrapper around db.FastBatchInsertBuilder
 type claimableBalanceClaimantBatchInsertBuilder struct {
 	encodingBuffer *xdr.EncodingBuffer
-	builder        db.BatchInsertBuilder
+	builder        db.FastBatchInsertBuilder
+	table          string
 }
 
 // NewClaimableBalanceClaimantBatchInsertBuilder constructs a new ClaimableBalanceClaimantBatchInsertBuilder instance
-func (q *Q) NewClaimableBalanceClaimantBatchInsertBuilder(maxBatchSize int) ClaimableBalanceClaimantBatchInsertBuilder {
+func (q *Q) NewClaimableBalanceClaimantBatchInsertBuilder() ClaimableBalanceClaimantBatchInsertBuilder {
 	return &claimableBalanceClaimantBatchInsertBuilder{
 		encodingBuffer: xdr.NewEncodingBuffer(),
-		builder: db.BatchInsertBuilder{
-			Table:        q.GetTable("claimable_balance_claimants"),
-			MaxBatchSize: maxBatchSize,
-			Suffix:       "ON CONFLICT (id, destination) DO UPDATE SET last_modified_ledger=EXCLUDED.last_modified_ledger",
-		},
+		builder:        db.FastBatchInsertBuilder{},
+		table:          "claimable_balance_claimants",
 	}
 }
 
-// Add adds a new transaction to the batch
-func (i *claimableBalanceClaimantBatchInsertBuilder) Add(ctx context.Context, claimableBalanceClaimant ClaimableBalanceClaimant) error {
-	return i.builder.RowStruct(ctx, claimableBalanceClaimant)
+// Add adds a new claimant for a claimable Balance to the batch
+func (i *claimableBalanceClaimantBatchInsertBuilder) Add(claimableBalanceClaimant ClaimableBalanceClaimant) error {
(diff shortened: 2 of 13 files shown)


urvisavla:
Since we'll be updating [all](https://github.com/stellar/go/issues/5098) change processors to use `FastBatchInsertBuilder`, I changed the change processor interface to accept the database session required for `FastBatchInsertBuilder` in both the `Commit` and `ProcessChange` (internally calls `Commit`) methods. This meant that I had to pass nil to the database session for the [StatsChangeProcessor](https://github.com/stellar/go/pull/5104/files#diff-bee14211e4439838b73cf0d9091990c8d02e3fe140e01348954666cc57ba97ebR81). Even though StatsChangeProcessor implements the change processor interface, it is different from other change processors in that It updates in-mem counters instead of storing it in the database. 

This solution works, but I am not sure about the design. Here are some other ideas I had:
1. Create a separate interface for the StatsChangeProcessor because it is different from other change processors and handle it separately in the processor runner [RunAllProcessorsOnLedger](https://github.com/stellar/go/blob/4670509a3e872359fbe11f6ee0383ecf883e8ace/services/horizon/internal/ingest/processor_runner.go#L394)  and [RunHistoryArchiveIngestion](https://github.com/stellar/go/blame/4670509a3e872359fbe11f6ee0383ecf883e8ace/services/horizon/internal/ingest/processor_runner.go#L233) 
2. Create a separate interface for the StatsChangeProcessor and pass an instance of StatsChangeProcessor to each of the change processors, so they can update the stats.
3. Pass the database session to the constructors of the change processors that need it instead of modifying `ProcessChange` and `Commit`. This would mean that each processor would be bound to a session throughout its lifetime, which might be okay in our case.

services/horizon/internal/ingest/processors/claimable_balances_change_processor.go
```diff
@@ -53,59 +58,32 @@ func (p *ClaimableBalancesChangeProcessor) ProcessChange(ctx context.Context, ch
...
 	var (
-		cbsToUpsert   []history.ClaimableBalance
+		cbsToInsert   []history.ClaimableBalance
 		cbIDsToDelete []string
 	)
 	changes := p.cache.GetChanges()
 	for _, change := range changes {
-		switch {
-		case change.Pre == nil && change.Post != nil:
+		if change.Post != nil {
```
tamirms, line 66: can you return an error if there is ledger entry change update?

services/horizon/internal/ingest/verify_test.go
```diff
@@ -324,8 +331,13 @@ func TestStateVerifier(t *testing.T) {
 	test.ResetHorizonDB(t, tt.HorizonDB)
 	q := &history.Q{&db.Session{DB: tt.HorizonDB}}
 
+	tt.Assert.NoError(q.SessionInterface.BeginTx(tt.Ctx, &sql.TxOptions{}))
+	defer func() {
```
tamirms, line 335: I don't think the rollback is necessary because you are committing the transaction in the test. If the test fails before the commit, it doesn't matter because the entire test db is disposable.

services/horizon/internal/ingest/processors/claimable_balances_change_processor.go
```diff
@@ -132,6 +110,44 @@ func (p *ClaimableBalancesChangeProcessor) Commit(ctx context.Context) error {
 	return nil
 }
 
+func (p *ClaimableBalancesChangeProcessor) insertClaimableBalancesAndClaimants(ctx context.Context,
+	claimableBalances []history.ClaimableBalance) error {
+	defer p.claimantsInsertBuilder.Reset()
```
tamirms, line 115: there is no need to reset the cb builders because does not retry with the same processor instance. The processors are created at the start of every ingestion iteration and after the ingestion iteration they are discarded. In fact, I think we should remove the `Reset()` functions from   `ClaimableBalanceBatchInsertBuilder` and `ClaimableBalanceClaimantBatchInsertBuilder`

services/horizon/internal/db2/history/claimable_balance_batch_insert_builder.go
```diff
@@ -0,0 +1,47 @@
...
+// claimable_balances table
+type ClaimableBalanceBatchInsertBuilder interface {
+	Add(claimableBalance ClaimableBalance) error
+	Exec(ctx context.Context, session db.SessionInterface) error
+	Reset()
+}
+
+// ClaimableBalanceBatchInsertBuilder is a simple wrapper around db.FastBatchInsertBuilder
+type claimableBalanceBatchInsertBuilder struct {
+	encodingBuffer *xdr.EncodingBuffer
```
tamirms, line 20: I think this field is never used

services/horizon/internal/db2/history/claimable_balance_claimant_batch_insert_builder.go
```diff
@@ -7,36 +7,41 @@ import (
...
-	Exec(ctx context.Context) error
+	Add(claimableBalanceClaimant ClaimableBalanceClaimant) error
+	Exec(ctx context.Context, session db.SessionInterface) error
+	Reset()
 }
 
-// ClaimableBalanceClaimantBatchInsertBuilder is a simple wrapper around db.BatchInsertBuilder
+// ClaimableBalanceClaimantBatchInsertBuilder is a simple wrapper around db.FastBatchInsertBuilder
 type claimableBalanceClaimantBatchInsertBuilder struct {
 	encodingBuffer *xdr.EncodingBuffer
```
tamirms, line 20: I think this field is never used

services/horizon/internal/ingest/processors/claimable_balances_change_processor.go
tamirms: > In fact, I think we should remove the Reset() functions from ClaimableBalanceBatchInsertBuilder and ClaimableBalanceClaimantBatchInsertBuilder

nevermind, it appears we still need the Reset() functions in the case that `p.cache.Size() > maxBatchSize` in `ProcessChange()`

services/horizon/internal/ingest/processors/claimable_balances_change_processor.go
```diff
@@ -53,59 +58,32 @@ func (p *ClaimableBalancesChangeProcessor) ProcessChange(ctx context.Context, ch
...
-		case change.Pre == nil && change.Post != nil:
+		if change.Post != nil {
 			// Created
 			row, err := p.ledgerEntryToRow(change.Post)
 			if err != nil {
 				return err
 			}
-			cbsToUpsert = append(cbsToUpsert, row)
-		case change.Pre != nil && change.Post == nil:
+			cbsToInsert = append(cbsToInsert, row)
```
tamirms, line 72: we don't need to keep a list of cbs to insert because we can add the cbs to the batch insert builder directly right here

services/horizon/internal/db2/history/claimable_balance_batch_insert_builder.go
```diff
@@ -0,0 +1,47 @@
...
+	builder        db.FastBatchInsertBuilder
+	table          string
```
tamirms, line 22: I think we should add a 	`session` field of type `db.SessionInterface` so we can remove the need to pass it into `Exec()`:

```diff
diff --git a/services/horizon/internal/db2/history/claimable_balance_batch_insert_builder.go b/services/horizon/internal/db2/history/claimable_balance_batch_insert_builder.go
index 93a6abd9d..bba562a3f 100644
--- a/services/horizon/internal/db2/history/claimable_balance_batch_insert_builder.go
+++ b/services/horizon/internal/db2/history/claimable_balance_batch_insert_builder.go
@@ -11,7 +11,7 @@ import (
 // claimable_balances table
 type ClaimableBalanceBatchInsertBuilder interface {
 	Add(claimableBalance ClaimableBalance) error
-	Exec(ctx context.Context, session db.SessionInterface) error
+	Exec(ctx context.Context) error
 	Reset()
 }
 
@@ -20,6 +20,7 @@ type claimableBalanceBatchInsertBuilder struct {
 	encodingBuffer *xdr.EncodingBuffer
 	builder        db.FastBatchInsertBuilder
 	table          string
+	session        db.SessionInterface
 }
 
 // NewClaimableBalanceBatchInsertBuilder constructs a new ClaimableBalanceBatchInsertBuilder instance
@@ -28,6 +29,7 @@ func (q *Q) NewClaimableBalanceBatchInsertBuilder() ClaimableBalanceBatchInsertB
 		encodingBuffer: xdr.NewEncodingBuffer(),
 		builder:        db.FastBatchInsertBuilder{},
 		table:          "claimable_balances",
+		session:        q,
 	}
 }
 
@@ -37,8 +39,8 @@ func (i *claimableBalanceBatchInsertBuilder) Add(claimableBalance ClaimableBalan
 }
 
 // Exec writes the batch of claimable balances to the database.
-func (i *claimableBalanceBatchInsertBuilder) Exec(ctx context.Context, session db.SessionInterface) error {
-	return i.builder.Exec(ctx, session, i.table)
+func (i *claimableBalanceBatchInsertBuilder) Exec(ctx context.Context) error {
+	return i.builder.Exec(ctx, i.session, i.table)
 }
 
```

I know that the batch insert builders for the history tables take `session` as an argument to `Exec()`. But, the difference between the processors for the history tables and the change processors is that the history tables require the DB on the `Flush()` function but not on the `ProcessTransaction()` function, which allows us to minimize the lifetime of the DB transaction. However, the change processors need the DB connection in both the `ProcessChange()` and `Commit()` functions.

sreuland: @tamirms , was looking into `ProcessChange()` on a few change processors to see how they were using db during `Exec()`, and doesn't appear like they are actually querying db states, it looked like mostly xdr methods and then it calls out to ChangeCompactor which just works with in-mem objects, maybe I'm missing the db call or it's just in a specific changeprocessor, in which case disregard, otherwise, it it turns out there is no db access from `ProcessChange()` implementations, we could get by with same pattern for `Exec(ctx, sessionInterface)` on change processors, just for consistency of a single pattern for processor to fastbatchbuilder.

tamirms: @sreuland you're right but there's a limit to how many objects we accumulate in-memory. In `ProcessChange()` you will find code like below:

```go
	if p.cache.Size() > maxBatchSize {
		err = p.Commit(ctx)
		if err != nil {
			return errors.Wrap(err, "error in Commit")
		}
		p.reset()
	}
```

In the scenario where we accumulate more than `maxBatchSize` ledger entry changes we will invoke the DB in `ProcessChange()`.

services/horizon/internal/ingest/processors/claimable_balances_change_processor.go
sreuland: in that edge case mentioned in ProcessChange(), what if we just change that conditional block, add a deferred call to processor `reset()`, since it currently short-circuited if `Commit()` had an err, missing that last call to processor reset(), which would ensure the same builder reset is accomplished by creating and assigning new builder instances to the current processor instance, which would seem to allow for removing the need for builders to have the `Reset()` as mentioned, could be ideal as that behavior is more of a processor state concern rather than something a builder should know how to do.

sreuland (approved):
nice work, left comments for consideration on two ongoing threads related to builder interface design choices for `Exec()` and `Reset()`.

services/horizon/internal/ingest/processors/claimable_balances_change_processor.go
tamirms: thanks @sreuland , I just re-read the code and it seems that `reset()` does not call the builder's `Reset()` function and instead it recreates the builder instance:

```go
func (p *ClaimableBalancesChangeProcessor) reset() {
	p.cache = ingest.NewChangeCompactor()
	p.claimantsInsertBuilder = p.qClaimableBalances.NewClaimableBalanceClaimantBatchInsertBuilder()
	p.claimableBalanceInsertBuilder = p.qClaimableBalances.NewClaimableBalanceBatchInsertBuilder()
}
```

So, I think we can proceed with removing the `Reset()` function on the builder.

services/horizon/internal/db2/history/claimable_balance_batch_insert_builder.go
sreuland: ah, we haven't followed a size limit in fast batch builder yet, the `Row` methods just increases internal slice, so the wrapping batch builder `Add` used by processors in ProcessTransaction()/ProcessChange() don't need reference to a db session, but it seems likely that fastbatchbuilder.Row methods may need to trigger this same size-exceeded-flush-now pattern, probably by throwing a `SizeLimitExceededErr`,  from `Add(ctx)` which wrapping builders would catch and handle by running `Exec(ctx, sessionInterface)` inline, preventing it from bubbling up any further. Not recommending to do in this PR, but maybe that is needed as part of enabling new range of ledgers for a processor in https://github.com/stellar/go/issues/5099.

since reference to a sessionInterface could ultimately be needed during any interaction(add or exec) between  fastbatch wrapped builder and processor, maybe the interface for wrapped builders, regardless of change or transaction oriented, can follow a single shape of:
```
Add(ctx Context, ...entity fields)
Exec (ctx Context)
```

where no session is passed, rather, all fastbatch wrapped batch builders, would take a sessionInterface in their `New` methods and have that available to use as needed for internal calls to `fastbatchbuilder.Exec(ctx, sessionInterface)` to flush memory over to db.

tamirms: > but it seems likely that fastbatchbuilder.Row methods may need to trigger this same size-exceeded-flush-now pattern

we can't apply the same pattern to the transaction processors because the loaders need to be executed first before we can start flushing the fast batch builder contents to the DB. For https://github.com/stellar/go/issues/5099 , we can limit the memory usage of the builders implicitly by restricting the number of ledgers applied on a given processor.

sreuland: yes, ok, general take-away on design aspect, don't apply max sizing concerns in fast batch or wrapped builders, rather the callers(tx or change processors) of wrapped builders enforce notion of max memory sizing at a higher level by checking their own cached models(change processors) and for upcoming multi-ledger tx processors(#5099) checking the number of ledgers processed. thanks for walking through the subtleties between change and tx processor flows, didn't want to derail pr conclusion here, wrapped builders for change processors follow the interface @urvisavla has proposed with the slight tweak for [recommended session reference](https://github.com/stellar/go/pull/5104#discussion_r1383096995)

services/horizon/internal/ingest/verify_test.go
urvisavla: removed rollback.

services/horizon/internal/ingest/verify_test.go
```diff
@@ -271,8 +271,13 @@ func TestStateVerifierLockBusy(t *testing.T) {
...
```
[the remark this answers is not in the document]
urvisavla: removed rollback.

services/horizon/internal/db2/history/claimable_balance_batch_insert_builder.go
urvisavla: removed encodingBuffer

services/horizon/internal/db2/history/claimable_balance_claimant_batch_insert_builder.go
urvisavla: removed encodingBuffer

services/horizon/internal/ingest/processors/claimable_balances_change_processor.go
urvisavla: updated.

services/horizon/internal/ingest/processors/claimable_balances_change_processor.go
urvisavla: Removed Reset() from both the batch builders.

services/horizon/internal/ingest/processors/claimable_balances_change_processor.go
urvisavla: done

services/horizon/internal/db2/history/claimable_balance_batch_insert_builder.go
```diff
@@ -0,0 +1,40 @@
...
+type claimableBalanceBatchInsertBuilder struct {
+	session db.SessionInterface
+	builder db.FastBatchInsertBuilder
+	table   string
+}
+
+// NewClaimableBalanceBatchInsertBuilder constructs a new ClaimableBalanceBatchInsertBuilder instance
+func (q *Q) NewClaimableBalanceBatchInsertBuilder(session db.SessionInterface) ClaimableBalanceBatchInsertBuilder {
+	return &claimableBalanceBatchInsertBuilder{
+		session: session,
```
tamirms, line 26: actually you don't need to have a `session db.SessionInterface` argument to this function because `q` is itself a session.

```suggestion
func (q *Q) NewClaimableBalanceBatchInsertBuilder() ClaimableBalanceBatchInsertBuilder {
	return &claimableBalanceBatchInsertBuilder{
		session: q,
```

services/horizon/internal/db2/history/claimable_balance_claimant_batch_insert_builder.go
```diff
@@ -4,39 +4,37 @@ import (
...
-	encodingBuffer *xdr.EncodingBuffer
-	builder        db.BatchInsertBuilder
+	session db.SessionInterface
+	builder db.FastBatchInsertBuilder
+	table   string
 }
 
 // NewClaimableBalanceClaimantBatchInsertBuilder constructs a new ClaimableBalanceClaimantBatchInsertBuilder instance
-func (q *Q) NewClaimableBalanceClaimantBatchInsertBuilder(maxBatchSize int) ClaimableBalanceClaimantBatchInsertBuilder {
+func (q *Q) NewClaimableBalanceClaimantBatchInsertBuilder(session db.SessionInterface) ClaimableBalanceClaimantBatchInsertBuilder {
```
tamirms, line 24: same comment applies from the claimable balance batch insert builder:

you don't need to have a session db.SessionInterface argument to this function because q is itself a session.

services/horizon/internal/ingest/verify_test.go
```diff
@@ -271,8 +271,10 @@ func TestStateVerifierLockBusy(t *testing.T) {
 	test.ResetHorizonDB(t, tt.HorizonDB)
 	q := &history.Q{&db.Session{DB: tt.HorizonDB}}
 
+	tt.Assert.NoError(q.SessionInterface.BeginTx(tt.Ctx, &sql.TxOptions{}))
```
tamirms, line 274: you don't need to reference `SessionInterface` when accessing `q`. You can invoke all the `SessionInterface` functions directly on `q`. For example:

https://github.com/stellar/go/blob/4670509a3e872359fbe11f6ee0383ecf883e8ace/services/horizon/internal/ingest/fsm.go#L145

tamirms: this comment also applies to the other tests where `q.SessionInterface` is referenced:

services/horizon/internal/actions/claimable_balance_test.go
services/horizon/internal/db2/history/claimable_balances_test.go

services/horizon/internal/db2/history/claimable_balance_batch_insert_builder.go
urvisavla: updated

services/horizon/internal/db2/history/claimable_balance_claimant_batch_insert_builder.go
urvisavla: updated

services/horizon/internal/ingest/verify_test.go
urvisavla: updated

services/horizon/internal/ingest/processors/claimable_balances_change_processor.go
```diff
@@ -2,32 +2,38 @@ package processors
...
+	claimableBalanceInsertBuilder history.ClaimableBalanceBatchInsertBuilder
+	session                       db.SessionInterface
 }
 
-func NewClaimableBalancesChangeProcessor(Q history.QClaimableBalances) *ClaimableBalancesChangeProcessor {
+func NewClaimableBalancesChangeProcessor(Q history.QClaimableBalances, session db.SessionInterface) *ClaimableBalancesChangeProcessor {
 	p := &ClaimableBalancesChangeProcessor{
 		encodingBuffer:     xdr.NewEncodingBuffer(),
 		qClaimableBalances: Q,
+		session:            session,
```
tamirms, line 27: the `session` field is no longer necessary because we create the insert builders from `qClaimableBalances`

urvisavla: updated

services/horizon/internal/ingest/processor_runner.go
```diff
@@ -109,6 +109,7 @@ func (s *ProcessorRunner) DisableMemoryStatsLogging() {
 
 func buildChangeProcessor(
 	historyQ history.IngestionQ,
+	session db.SessionInterface,
```
tamirms, line 112: `buildChangeProcessor` - `session` is unused (unparam)

services/horizon/internal/ingest/processors/claimable_balances_change_processor_test.go
```diff
@@ -20,19 +21,29 @@ func TestClaimableBalancesChangeProcessorTestSuiteState(t *testing.T) {
...
-	ctx                    context.Context
-	processor              *ClaimableBalancesChangeProcessor
-	mockQ                  *history.MockQClaimableBalances
-	mockBatchInsertBuilder *history.MockClaimableBalanceClaimantBatchInsertBuilder
+	ctx                                    context.Context
+	processor                              *ClaimableBalancesChangeProcessor
+	mockQ                                  *history.MockQClaimableBalances
+	mockClaimantsBatchInsertBuilder        *history.MockClaimableBalanceClaimantBatchInsertBuilder
+	mockClaimableBalanceBatchInsertBuilder *history.MockClaimableBalanceBatchInsertBuilder
+	session                                db.SessionInterface
```
tamirms, line 29: field `session` is unused (unused)

services/horizon/internal/ingest/processors/claimable_balances_change_processor_test.go
```diff
@@ -109,25 +120,35 @@ func TestClaimableBalancesChangeProcessorTestSuiteLedger(t *testing.T) {
...
```
tamirms, line 128: field `session` is unused (unused)

tamirms (approved):
there are a few lint warnings about unused fields / variables. But once those are fixed, we should be good to merge!

tamirms (approved):

PR #5116, opened 2023-11-15
xdr: Sanity-check allocations when decoding

2opremio:
Depends on https://github.com/stellar/xdrgen/pull/182 and https://github.com/stellar/go-xdr/pull/19


xdr/main.go
```diff
@@ -56,10 +56,17 @@ func safeUnmarshalString(decoder func(reader io.Reader) io.Reader, data string,
 // decoding the xdr into the provided destination. Also ensures that the reader
 // is fully consumed.
 func SafeUnmarshalBase64(data string, dest interface{}) error {
+	return SafeUnmarshalBase64WithMaxAllocSize(data, 0, dest)
+}
+
+// SafeUnmarshalBase64WithMaxAllocSize works just like SafeUnmarshalBase64, except a maximum
+// allocation size is provided.
+func SafeUnmarshalBase64WithMaxAllocSize(data string, maxAllocSize int, dest interface{}) error {
```
Shaptic, line 64: we should rename it as `EvenSaferUnmarshalBase64` lol

Shaptic: More seriously, `BoundedUnmarshalBase64` might be a better name

Shaptic: Ah sorry I think this belongs in the xdrgen PR (https://github.com/stellar/xdrgen/pull/182), instead

services/horizon/internal/actions/submit_transaction.go
```diff
@@ -36,12 +36,9 @@ type envelopeInfo struct {
...
 }
 
-func extractEnvelopeInfo(raw string, passphrase string) (envelopeInfo, error) {
+func (handler SubmitTransactionHandler) extractEnvelopeInfo(raw string, passphrase string) (envelopeInfo, error) {
 	result := envelopeInfo{raw: raw}
-	if err := gxdr.ValidateTransactionEnvelope(raw, gxdr.DefaultMaxDepth); err != nil {
-		return result, err
-	}
-	err := xdr.SafeUnmarshalBase64(raw, &result.parsed)
+	err := xdr.SafeUnmarshalBase64WithMaxAllocSize(raw, int(handler.MaxHTTPRequestSize), &result.parsed)
```
sreuland, line 41: that worked out nice!

sreuland: I was going to suggest suggest a new test in submit_transaction_test.go to assert error when request xdr payload exceeds `MaxHTTPRequestSize` , but not sure if it could even get here or middleware would reject that request first before forwarding to this handler.

sreuland (approved):
lgtm!

services/horizon/internal/actions/submit_transaction.go
2opremio: The middleware would reject it before it gets here.

services/horizon/internal/actions/submit_transaction.go
```diff
@@ -23,9 +22,10 @@ type NetworkSubmitter interface {
 }
 
 type SubmitTransactionHandler struct {
-	Submitter         NetworkSubmitter
-	NetworkPassphrase string
-	DisableTxSub      bool
+	Submitter          NetworkSubmitter
+	NetworkPassphrase  string
+	DisableTxSub       bool
+	MaxHTTPRequestSize uint
```
tamirms, line 28: I don't think `MaxHTTPRequestSize` is used anywhere

2opremio:
True, I will remove it

services/horizon/internal/actions/submit_transaction.go
2opremio: True, I will remove it

xdr/xdr_generated.go
```diff
@@ -25,10 +25,14 @@ import (
...
 	"fmt"
 	"io"
+	"unsafe"
 
 	"github.com/stellar/go-xdr/xdr3"
 )
 
+// Needed since unsafe is not used in all cases
+var _ = unsafe.Sizeof(0)
+
```
2opremio, line 35: This isn't needed anymore, I should remove it.

xdr/xdr_generated.go
```diff
@@ -368,8 +380,11 @@ func (s *ScpNomination) DecodeFrom(d *xdr.Decoder, maxDepth uint) (int, error) {
 	}
 	s.Votes = nil
 	if l > 0 {
+		if il, ok := d.InputLen(); ok && uint(il) < uint(l) {
+			return n, fmt.Errorf("decoding Value: length (%d) exceeds remaining input length (%d)", l, il)
+		}
 		s.Votes = make([]Value, l)
-		for i := uint32(0); i < l; i++ {
+		for i := uint32(0); uint(i) < uint(l); i++ {
```
2opremio, line 387: This isn't needed either

PR #5117, opened 2023-11-15
services/horizon/ingest: historyRange and reingestHistoryRange states send batches of ledgers to tx processors

### LINKED ISSUE #5099: Ingest batches of ledgers in-memory before flushing to DB

sreuland:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Changed the `historyRange` and `reingestHistoryRange` state machines to send batches of ledgers to tx processors for each db tx begin/commit round trip, rather then just one ledger.

For context, a pre-req, the tx processors have already been enabled with fast batch builders for holding ledger updates in memory longer and then flushing  per #4909, this changeset builds upon that by now increasing to a range of ledgers sent to the tx processors before flushing.

The batch size of ledgers is configurable per `Config.MaxLedgerPerFlush`

### Why

Ingestion performance, reducing the number of db round trips for tx commits to lower overall time to finish ingestion range.
Closes #5099 

### Known limitations

this batching of ledgers for db round trips was only applicable to the two reingest states `historyRange` and `reingestHistoryRange`


services/horizon/cmd/db.go
```diff
@@ -275,6 +276,14 @@ func ingestRangeCmdOpts() support.ConfigOptions {
 			FlagDefault: uint32(100000),
 			Usage:       "[optional] parallel workers will run jobs processing ledger batches of the supplied size",
 		},
+		{
+			Name:        "ledgers-per-flush",
+			ConfigKey:   &maxLedgersPerFlush,
+			OptType:     types.Uint32,
+			Required:    false,
+			FlagDefault: uint32(10),
```
sreuland, line 284: i wondered about externalizing this config for ledger size, as it's just more config noise for user to be aware of, maybe the default here is enough and users would never need to change, but in that case, should we expose this as a flag or just hardcode a sensible value internally?

services/horizon/internal/ingest/fsm_reingest_history_range_state.go
```diff
@@ -0,0 +1,187 @@
...
+		if err := s.historyQ.Begin(s.ctx); err != nil {
+			return stop(), errors.Wrap(err, "Error starting a transaction")
+		}
+		defer s.historyQ.Rollback()
+
+		if e := h.ingestRange(s, h.fromLedger, h.toLedger); e != nil {
+			return stop(), e
+		}
+
+		if e := s.historyQ.Commit(); e != nil {
```
sreuland, line 170: this is the big gain here, as it only does a commit once at end after all ledgers flushed from processor memory to db tx, before it was in a loop of tx begin/commit per each ledger in the range

services/horizon/internal/ingest/fsm_history_range_state.go
```diff
@@ -0,0 +1,107 @@
...
+	for cur := h.fromLedger; cur <= h.toLedger; cur++ {
+		var ledgerCloseMeta xdr.LedgerCloseMeta
+
+		log.WithField("sequence", cur).Info("Waiting for ledger to be available in the backend...")
+		startTime := time.Now()
+
+		ledgerCloseMeta, err = s.ledgerBackend.GetLedger(s.ctx, cur)
+		if err != nil {
+			// Commit finished work in case of ledger backend error.
+			commitErr := s.historyQ.Commit()
```
urvisavla, line 74: Do we need to process the accumulated ledgers with by calling `RunTransactionProcessorsOnLedgers` before committing?

services/horizon/internal/ingest/fsm_reingest_history_range_state.go
```diff
@@ -0,0 +1,187 @@
...
+	ledgers := []xdr.LedgerCloseMeta{}
+	for cur := fromLedger; cur <= toLedger; cur++ {
+		var ledgerCloseMeta xdr.LedgerCloseMeta
+
+		log.WithField("sequence", cur).Info("Waiting for ledger to be available in the backend...")
+		startTime := time.Now()
+		ledgerCloseMeta, err = s.ledgerBackend.GetLedger(s.ctx, cur)
+
+		if err != nil {
+			return errors.Wrap(err, "error getting ledger")
```
urvisavla, line 61: Same as above. Should we process the ledgers accumulated so far in case of an error rather than discarding the batch?

services/horizon/internal/ingest/processor_runner.go
```diff
@@ -317,49 +309,158 @@ func (s *ProcessorRunner) runChangeProcessorOnLedger(
...
+	if err != nil {
+		return errors.Wrap(err, "Error streaming changes from ledger")
+	}
+
+	transactionStats := groupProcessors.transactionStatsProcessor.GetResults()
+	transactionStats.TransactionsFiltered = groupFilterers.droppedTransactions
+
+	tradeStats := groupProcessors.tradeProcessor.GetStats()
+
+	curHeap, sysHeap := getMemStats()
```
tamirms, line 348: I don't think we were getting these mem stats before. I don't know if we should unconditionally get and log these stats after ingesting every ledger. This might degrade performance.

services/horizon/internal/ingest/fsm_history_range_state.go
```diff
@@ -0,0 +1,107 @@
...
+
+	// We should be ingesting the ledger which occurs after
+	// lastHistoryLedger. Otherwise, some other horizon node has
+	// already completed the ingest history range operation and
+	// we should go back to the init state
+	if lastHistoryLedger != h.fromLedger-1 {
+		return start(), nil
+	}
+
+	ledgers := []xdr.LedgerCloseMeta{}
```
tamirms, line 64:
```suggestion
	ledgers := make([]xdr.LedgerCloseMeta, 0, s.maxLedgerPerFlush)
```

you can reset the slice after flushing with this code:

```go
ledgers = ledgers[0:]
```

services/horizon/cmd/db.go
tamirms: yeah, I don't think we need to expose this flag. If `parallel-job-size` is configured, then we should set `maxLedgersPerFlush` to the same value. Otherwise, I think we can default to 100 which is what I found to be effective during my testing in https://docs.google.com/document/d/1YETNALx5EzqZDNSVWzTfaK5Ogw84PsBlrt64nOr-Njg/edit#heading=h.ay7b6e6q2akx

services/horizon/internal/ingest/fsm_reingest_history_range_state.go
sreuland: good suggestion, since this currently would rollback all prior processor batch flushes upon an err, I've changed this to commit after an err, which results in any prior batch flushes getting committed at least to retain that progress, not reaching further to try triggering flush of processors on in-progress batch also, will let that drop off, as it would only be 100 ledgers at most.  https://github.com/stellar/go/pull/5117/commits/2386a2083ca2e7b84e486e125b37958222d58d4b

services/horizon/internal/ingest/fsm_history_range_state.go
sreuland: the extra call to RunTransactionProcessorsOnLedgers as part of error handling could potentially process up to 100 more ledgers as max batch size, but not sure if extra code/test is justified for the gain, this already commits up to the last flushed batch of the requested range which is consistent with the same error handling outcome in reingestHistoryRange.

services/horizon/internal/ingest/processor_runner.go
sreuland: ok, yes, I threw that in, removed now.

services/horizon/internal/ingest/fsm_history_range_state.go
sreuland: thanks for the suggestions on better slice handling, worked that in - https://github.com/stellar/go/pull/5117/commits/c2aea6ab64612b13b934c446ed0b92fc7f5afa30

services/horizon/cmd/db.go
sreuland: ok, done, thanks - https://github.com/stellar/go/pull/5117/commits/c2aea6ab64612b13b934c446ed0b92fc7f5afa30

sreuland:
> @sreuland this PR looks good! The 2 bugs mentioned at the end of #5099 are not addressed in this PR. Is the plan to spin out separate issues for those 2 bugs?

@tamirms, yes, have created https://github.com/stellar/go/issues/5128 and https://github.com/stellar/go/issues/5127

tamirms (approved):

PR #5135, opened 2023-12-03
services/horizon: Remove the optimization that uses COPY for liquidity pool insertion

### LINKED ISSUE #5134: services/horizon: Duplicate key error in LiquidityPoolsChangeProcessor
Latest from master branch.

### What did you do?
Ran Horizon in ingest mode.

Horizon to ingest without errors.

### What did you see instead?

Horizon ingestion halted with the following error during ingestion of ledger `49247417`:

ERRO[2023-11-30T15:59:26.921Z] Error in ingestion state machine current_state="resume(latestSuccessfullyProcessedLedger=49247416)" error="Error running processors on ledger: Error committing changes from processor: error in *processors.LiquidityPoolsChangeProcessor.Commit: error executing LiquidityPoolsBatchInsertBuilder: pq: duplicate key value violates unique constraint "liquidity_pools_pkey"" next_state="resume(latestSuccessfullyProcessedLedger=49247416)" pid=1647120 service=ingest

This error indicates that a duplicate key value was found while inserting data into the liquidity_pools table.

urvisavla:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Remove the optimization that uses COPY for liquidity pool insertion and revert to the previous method.

### Why
Fixes #5134 

To improve liquidity pool data insert performance, the COPY command was introduced to perform bulk insertions (https://github.com/stellar/go/pull/5110). This optimization assumed only newly created liquidity pools would be inserted using the COPY method, preventing conflicts with existing records. However, when the final trust-line for a liquidity pool is removed, the corresponding liquidity pool record is [marked for later deletion](https://github.com/stellar/go/pull/3898). And if a new trust-line is added before the actual deletion, inserting the updated liquidity pool data using COPY fails due to the duplicate record.

The solution is to revert to the original upsert method for liquidity pool insertions which handles both new insertions and updates to existing records.

### Known limitations
N/A


tamirms (approved):

services/horizon/internal/ingest/processors/liquidity_pools_change_processor.go
```diff
@@ -45,24 +43,21 @@ func (p *LiquidityPoolsChangeProcessor) ProcessChange(ctx context.Context, chang
 		if err != nil {
 			return errors.Wrap(err, "error in Commit")
 		}
+		p.reset()
```
sreuland, line 46: should internal compactor state be reset regardless of error? it looked like prior path deferred to that but maybe it's not relevant given the removal of builder.

sreuland (approved):

services/horizon/internal/ingest/processors/liquidity_pools_change_processor.go
urvisavla: It's a valid question but I'm not sure about resetting the compactor if the commit fails. I rolled back the change without considering the compactor's impact.  @tamirms what's your opinion, should we reset the compactor regardless of the commit outcome?

tamirms: When `ProcessChange()` returns an error the error gets bubbled up and eventually the processor is discarded.

If it were possible to reuse a processor instance across multiple ingestion rounds it would make sense to reset the processor. But in this case, it doesn't matter if we reset the processor upon returning an error because the processor will never used again.

PR #5140, opened 2023-12-12
services/horizon/internal/db2/history: Use FastBatchInsertBuilder for asset stats

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Use FastBatchInsertBuilder for asset stats to speed up db operations where we need to insert a lot of asset stat rows.

### DIFF
--- a/services/horizon/internal/db2/history/asset_stats_test.go
+++ b/services/horizon/internal/db2/history/asset_stats_test.go
@@ -21,6 +21,8 @@ func TestAssetStatContracts(t *testing.T) {
 	test.ResetHorizonDB(t, tt.HorizonDB)
 	q := &Q{tt.HorizonSession()}
 
+	tt.Assert.NoError(q.Begin(context.Background()))
+
 	assetStats := []ExpAssetStat{
 		{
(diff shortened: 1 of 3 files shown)


services/horizon/internal/db2/history/asset_stats_test.go
```diff
@@ -183,13 +188,18 @@ func TestAssetContractStats(t *testing.T) {
 		tt.Assert.NoError(err)
 		tt.Assert.Equal(result, row)
 	}
+
+	tt.Assert.NoError(q.Rollback())
```
sreuland, line 192: the same db tx session is shared via `q` between fast builder and test usage so commit isn't needed in test scope, should these tests be doing commit/rollback at all since not material to the outcomes?

sreuland (approved):
cool, nice add!

services/horizon/internal/db2/history/asset_stats_test.go
tamirms: I think the rollback is not necessary but I prefer to not leave the transaction open

PR #5147, opened 2023-12-21
services/horizon/docker/verify-range: Fix comparison of claimable balance ids in verify-range script

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This commit contains two fixes for the verify-range script:

1. The verify range script should compare hex claimable balance ids instead of internal history claimable balance integer ids.
2. The verify range script does not check the history_operation_liquidity_pools and history_transaction_liquidity_pools tables. 

The reason for (1) is because the generation of integer history ids can vary between implementations. For example, in horizon 2.28.0 we generate the history ids for all claimable balances in all the transactions of the ledger. Whereas in previous horizon releases we generate the history ids for claimable balances when ingesting each transaction. The differences in this behavior can lead to different history ids being assigned to the same set of claimable balances.

### Known limitations

[N/A]


sreuland (approved):
awesome, thanks for fix!

PR #5150, opened 2024-01-05
ingest/ledgerbackend: Fix captive core toml history entries

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Remove trailing slash from history archive urls when generating the captive core toml file.

### Why

If you ran the `horizon db reingest range` command with a `NETWORK` configured to pubnet, the command would fail because the history archive urls we use all have trailing slashes:

https://github.com/stellar/go/blob/master/network/main.go#L26-L28

This would result in core making curl requests to invalid urls.

### Known limitations

[N/A]

### DIFF
--- a/ingest/ledgerbackend/toml.go
+++ b/ingest/ledgerbackend/toml.go
@@ -558,7 +558,7 @@ func (c *CaptiveCoreToml) setDefaults(params CaptiveCoreTomlParams) {
 		for i, val := range params.HistoryArchiveURLs {
 			name := fmt.Sprintf("HISTORY.h%d", i)
 			c.HistoryEntries[c.tablePlaceholders.newPlaceholder(name)] = History{
-				Get: fmt.Sprintf("curl -sf %s/{0} -o {1}", val),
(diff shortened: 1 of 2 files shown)


mollykarcher (approved):

2opremio (approved):

urvisavla:
Considering whether removing the trailing slash [here](https://github.com/stellar/go/blob/master/network/main.go#L26-L28) is a better option as it seems to be used in [this](https://github.com/stellar/go/pull/5150/files#diff-c7702ef819e52917ed3868588dff5273774c61a6aa8d36c1644d7ba251fb7756L561)?

tamirms:
@urvisavla history archive urls can be configured via command line. So, even if we strip the trailing slash in https://github.com/stellar/go/blob/master/network/main.go#L26-L28 , it is still possible for horizon to fail a reingest command because the history archive urls provided by the user contain a trailing slash

PR #5158, opened 2024-01-10
services/horizon/ingest: removed legacy core cursor update against during ledger ingestion

### LINKED ISSUE #5153: No need to update cursor when ingesting from captive core
As @sreuland reported, we always update the cursor via the stellar core client in the resume state:

https://github.com/stellar/go/blob/master/services/horizon/internal/ingest/fsm.go#L526

We should only be doing this if we're running legacy ingestion via the stellar core postgres DB. When running captive core, there is no need to update the cursor.

sreuland:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Removed the core cursor update that is attempted by ingest fsm on each ledger, as it was only relevant to legacy core db ingestion which is no longer enabled, only captive core ingestion.

Removed the configuration flags `--stellar-core-url-db`, `--cursor-name` `--skip-cursor-update` , as they are related to legacy core db ingestion which is no longer usable. 

if one of the aforementioned flags is present, horizon will exit and report the error on console
```
$ DATABASE_URL=dburl STELLAR_CORE_DATABASE_URL=xxx ./horizon
flag --stellar-core-db-url and environment variable STELLAR_CORE_DATABASE_URL have been removed and no longer valid, must use captive core configuration for ingestion
```

Closes #5153 

### Why

discovered the un-necessary core cursor update taking time during performance benchmarking.


services/horizon/internal/ingest/main.go
```diff
@@ -273,13 +272,6 @@ func NewSystem(config Config) (System, error) {
 			cancel()
 			return nil, errors.Wrap(err, "error creating captive core backend")
 		}
-	} else {
```
sreuland, line 276: this was unreachable code, `config.RemoteCaptiveCoreEnabled()` is the inverse of `config.LocalCaptiveCoreEnabled()`, horizon assumes only captive core.

services/horizon/internal/ingest/main.go
```diff
@@ -252,7 +250,8 @@ func NewSystem(config Config) (System, error) {
 			cancel()
 			return nil, errors.Wrap(err, "error creating captive core backend")
 		}
-	} else if config.LocalCaptiveCoreEnabled() {
+	} else {
```
urvisavla, line 253: I think we can remove lines 247 to 252 as we've deprecated remote captive core as well.

urvisavla (approved):
lgtm!

services/horizon/internal/ingest/main.go
sreuland: sure, I swept out the remnants of that - https://github.com/stellar/go/pull/5158/commits/a781a11ea006e418491043bb5482fed073ebcb34

urvisavla: Didn't realize it was still left in multiple places. Thanks for cleaning it up!

PR #5171, opened 2024-01-18
historyarchive: Cache bucket files from history archives on disk.

Shaptic:
### What
Add an on-disk cache to the history archive interface for streaming buckets and a metric for measuring cache hits, hooking both into the ingestion state machine.

### Why
This avoids unnecessary retries and repeated downloads from the HAs.

Doing a simple `Wrap` on the archive's `backend` is not viable, because we need the `stellar-history.json` file and its variants to **never** be cached, which cannot be controlled if letting the [`storage.OnDiskCache`](https://github.com/stellar/go/blob/master/support/storage/ondisk_cache.go) implementation from `master` implicitly cache all requests.

This replaces #5165.

### Known limitations
The cache size is an estimated guess, see https://github.com/stellar/go/pull/5171#discussion_r1458118423.

### DIFF
--- a/historyarchive/archive_cache.go
+++ b/historyarchive/archive_cache.go
@@ -1,7 +1,6 @@
 package historyarchive
 
 import (
-	"fmt"
 	"io"
 	"os"
 	"path"
@@ -95,12 +94,16 @@ func (abc *ArchiveBucketCache) GetFile(
 	if err != nil {
 		// Uh-oh, the cache and the disk are not in sync somehow? Let's evict
 		// this value and try again (recurse) w/ the remote version.
-		L.WithError(err).Warn("Opening cached ledger failed")
+		L.WithError(err).Warn("Opening cached file failed")
 		abc.lru.Remove(localPath)
 		return abc.GetFile(filepath, upstream)
 	}
 
 	return local, nil
+}
+
+func (abc *ArchiveBucketCache) Exists(filepath string) bool {
+	return abc.lru.Contains(path.Join(abc.path, filepath))
 }
 
 // Close purges the cache, then forwards the call to the wrapped backend.
@@ -158,7 +161,8 @@ func NameLockfile(file string) string {
 
 type trc struct {
 	io.Reader
-	close func() error
+	close  func() error
+	closed bool
 }
 
 func (t trc) Close() error {
@@ -166,23 +170,28 @@ func (t trc) Close() error {
 }
 
 func teeReadCloser(r io.ReadCloser, w io.WriteCloser, onClose func() error) io.ReadCloser {
-	fmt.Printf("Making teeReadCloser onto %v and %v\n", r, w)
-	return trc{
+	closer := trc{
 		Reader: io.TeeReader(r, w),
-		close: func() error {
-			// Always run all closers, but return the first error
-			err1 := r.Close()
-			err2 := w.Close()
-			err3 := onClose()
-
-			fmt.Println("Errors were:", err1, err2, err3)
-
-			if err1 != nil {
-				return err1
-			} else if err2 != nil {
-				return err2
-			}
-			return err3
-		},
+		closed: false,
 	}
+	closer.close = func() error {
+		if closer.closed {
+			return nil
+		}
+
+		// Always run all closers, but return the first error
+		err1 := r.Close()
+		err2 := w.Close()
+		err3 := onClose()
+
+		closer.closed = true
+		if err1 != nil {
+			return err1
+		} else if err2 != nil {
+			return err2
+		}
+		return err3
+	}
+
+	return closer
 }
--- a/historyarchive/archive_cache.go
+++ b/historyarchive/archive_cache.go
@@ -112,13 +112,9 @@ func (abc *ArchiveBucketCache) Exists(filepath string) bool {
 	return abc.lru.Contains(path.Join(abc.path, filepath))
 }
 
-// Close purges the cache, then forwards the call to the wrapped backend.
+// Close purges the cache and cleans up the filesystem.
 func (abc *ArchiveBucketCache) Close() error {
-	// We only purge the cache, leaving the filesystem untouched:
-	// https://github.com/stellar/go/pull/4457#discussion_r929352643
 	abc.lru.Purge()
-
-	// Only bubble up the disk purging error if there is no other error.
 	return os.RemoveAll(abc.path)
 }
 
@@ -134,7 +130,7 @@ func (abc *ArchiveBucketCache) onEviction(key, value interface{}) {
 	if err := os.Remove(path); err != nil { // best effort removal
 		abc.log.WithError(err).
 			WithField("key", path).
-			Warn("removal failed after cache eviction")
+			Warn("Removal failed after cache eviction")
 	}
 }
 
@@ -171,6 +167,10 @@ type trc struct {
 }
 
 func (t trc) Close() error {
+	if t.closed {
+		return nil
+	}
+
 	return t.close()
 }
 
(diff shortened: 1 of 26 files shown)


historyarchive/archive_pool.go
```diff
@@ -31,14 +31,7 @@ func NewArchivePool(archiveURLs []string, config ConnectOptions) (ArchivePool, e
...
 	for _, url := range archiveURLs {
-		archive, err := Connect(
-			url,
-			ConnectOptions{
-				NetworkPassphrase:   config.NetworkPassphrase,
-				CheckpointFrequency: config.CheckpointFrequency,
-				Context:             config.Context,
-			},
-		)
+		archive, err := Connect(url, config)
```
sreuland, line 34: this was also resolved in a recent merge, might a conflict here now.

services/horizon/internal/ingest/main.go
```diff
@@ -222,6 +223,11 @@ func NewSystem(config Config) (System, error) {
 			NetworkPassphrase:   config.NetworkPassphrase,
 			CheckpointFrequency: config.CheckpointFrequency,
 			UserAgent:           fmt.Sprintf("horizon/%s golang/%s", apkg.Version(), runtime.Version()),
+			CacheConfig: historyarchive.CacheOptions{
```
sreuland, line 226: should we also add the cache to the captive core runner - https://github.com/stellar/go/blob/release-horizon-v2.28.0/ingest/ledgerbackend/captive_core_backend.go#L175, i don't think the archives get used alot by it, just a few times up front before launching core..

historyarchive/archive_cache.go
```diff
@@ -0,0 +1,188 @@
...
+		// Since it's not on-disk, pull it from the remote backend, shove it
+		// into the cache, and write it to disk.
+		remote, err := upstream.GetFile(filepath)
+		if err != nil {
+			return remote, err
+		}
+
+		local, err := abc.createLocal(filepath)
+		if err != nil {
+			// If there's some local FS error, we can still continue with the
```
sreuland, line 80: wondering about thread safety in GetFile, we know there will be live ingestion calling archives and a separate thread running state verify against archives at times concurrently, this seems thread-safe in here based on err handling/best effort, do see any concerns in that area?

historyarchive/archive_test.go
```diff
@@ -607,7 +614,9 @@ func TestGetLedgers(t *testing.T) {
 		[]xdrEntry{results[0], results[1], results[2]},
 	)
 
-	ledgers, err := archive.GetLedgers(1000, 1002)
+	_, err = archive.GetLedgers(1000, 1002)
+	assert.NoError(t, err)
+	ledgers, err := archive.GetLedgers(1000, 1002) // invoke cache
```
sreuland, line 619: after you port latest 2.28.0 branch, it has new Stats added to Archive/ArchivePool, can use archive.GetStats().[GetDownloads()](https://github.com/stellar/go/blob/release-horizon-v2.28.0/historyarchive/archive.go#L96) here in test to confirm the cache was used on 2nd attempt and not archive based on the counter.

sreuland: hmm, actually, we might want to change [ArchiveStats](https://github.com/stellar/go/blob/release-horizon-v2.28.0/historyarchive/archive.go#L71-L75), and add counters for cache activity on the archive?

historyarchive/archive_pool.go
Shaptic: thanks for the heads up! I merged upstream and fixed it up

services/horizon/internal/ingest/main.go
Shaptic: It looks like the runner only uses it in `runFromParams` when calling `GetLedgers`, which means the cache wouldn't make a difference anyway because it'll be empty. better to leave it as-is, I think!

historyarchive/archive_cache.go
Shaptic: Great question: I used the same strategy as we used in `OnDiskCache` from Zenith which was tested implicitly at length by the parallel download feature (https://github.com/stellar/go/pull/4548), we had discussions here during the original design: https://github.com/stellar/go/pull/4575.

In this port, I tried to very aggressively bail out if there were any concerns about the filesystem not being in sync, including cache evictions at the `Archive` level and retries with the backend. And I think contention should be very low because live ingestion does not really call archives in Horizon; rather, Captive Core does.

historyarchive/archive_test.go
```diff
@@ -641,4 +648,19 @@ func TestGetLedgers(t *testing.T) {
...
 		assertXdrEquals(t, results[i], ledger.TransactionResult)
 	}
+
+	// Repeat the same check but ensure the cache was used
+	ledgers, err = archive.GetLedgers(1000, 1002) // all cached
+	assert.NoError(t, err)
+	assert.Len(t, ledgers, 3)
+
+	// requests and downloads should not change because of the cache
+	assert.Equal(t, uint32(7), archive.GetStats()[0].GetRequests())
```
sreuland, line 658: do you think it's worthwhile to add cache hit counters now to [ArchiveStats](https://github.com/stellar/go/blob/release-horizon-v2.28.0/historyarchive/archive.go#L71-L75) to gain metrics on cache efficacy?

historyarchive/archive_test.go
Shaptic: great idea! added in 5c1c597c

historyarchive/archive_test.go
Shaptic: another great idea! done in 8728a85

services/horizon/internal/ingest/main.go
```diff
@@ -226,6 +227,11 @@ func NewSystem(config Config) (System, error) {
...
+				Cache:    true,
+				Path:     path.Join(config.CaptiveCoreStoragePath, "bucket-cache"),
+				MaxFiles: 100,
```
sreuland, line 233: is constant of 100 good middle ground to start with, seems like a good starter, not asking to push out any further to a config param etc, will have to watch it. any napkin math on disk space usage expecatations given this cache size, not sure on size of a recent archive xdr.gz. if it's more than a couple GB, we should probably mention in the CHANGELOG this impact on local disk.

sreuland (approved):
lgtm, great work!

services/horizon/internal/ingest/main.go
Shaptic: I asked @SirTyson and he ball-parked the following:

> Current BL size is like 11.5 GB
> I think it's about 18 files

It's not a lot, but you're definitely right about noting the disk requirement changes! We could change the 100 to 25 or something more conservative, too. Done in efea5084.

historyarchive/stats.go
```diff
@@ -0,0 +1,57 @@
...
+	fileUploads   atomic.Uint32
+	cacheHits     atomic.Uint32
+	backendName   string
+}
+
+type ArchiveStats interface {
+	GetRequests() uint32
+	GetDownloads() uint32
+	GetUploads() uint32
+	GetCacheHits() uint32
```
sreuland, line 18: nice, let's add cache hits to the [archive metrics emission in fsm](https://github.com/Shaptic/go/blob/cache-HA-buckets/services/horizon/internal/ingest/fsm.go#L576) also:

```
s.Metrics().HistoryArchiveStatsCounter.
	With(prometheus.Labels{
		"source": historyServerStat.GetBackendName(),
		"type":   "cache_hits"}).
	Add(float64(historyServerStat.GetCacheHits()))	
```

Shaptic: Oh awesome, thank you! Didn't realize there were so many spots to tweak to add the metric, heh. Fixed up in f6e9e24 :+1:

sreuland:
@Shaptic , i started prep on the rel branch, updated notes, you can revert the CHANGELOG here, is included on rel branch now, thanks!

Shaptic:
Bah, bad timing! I was fixing the conflict at the same time lol. Anyway I think we can merge this now :+1:

Issue #5174, opened 2024-01-19
services/horizon: add config setting for toggle soroban tx-meta presence


sreuland:
### What problem does your feature solve?
ability to control whether serialized tx meta for a soroban transaction is included in ingested transaction history

### What would you like to see?
a new horizon config parameter `ENABLE_SOROBAN_TXMETA_HISTORY`, defaults to `FALSE`

when it's TRUE, and an incoming transaction is for soroban, the row for it in `history_transactions` table should have serialized base64 for an empty `xdr.TransactionMetaV3{}`  in`tx_meta`, and for the API resource model emitted for `Transaction` for this row needs to show the same value for 
`result_meta_xdr`

also, confirm tx_sub doesn't leave anything around but still works for soroban tx's after this.

### What alternatives are there?
this should target [2.28.0](https://github.com/stellar/go/tree/release-horizon-v2.28.0) release first, not master.

sreuland:
closing, this was done in #5175 instead

PR #5176, opened 2024-01-20
services/horizon: Add DISABLE_SOROBAN_INGEST flag to skip soroban ingestion processing

### LINKED ISSUE #5175: services/horizon: Add the ability to toggle ingestion of Soroban elements
Some `Change` and `Transaction` processors in the ledger ingestion pipeline are parsing Soroban transaction metadata and are extracting data. These include, **at minimum**:

 - <del>the `AssetStatsProcessor` which extensively tracks changes to wrapped Stellar assets (see `ingestAssetContractMetadata`)</del> - we discovered fundamental conflict here due to asset balances are based on trustlines/cb/lp entries which can also be changed by contracts and state verifier needs the contract tx-meta to re-assemble the full asset state value to compare against archive ledger entries of which trustlines/cb/lp include contract changes.
 - the `EffectProcessor` which produces `(account|contract)_(credited|debited)` effects (see `ingestEffects`)
 - history_transactions table should have serialized base64 for an empty xdr.TransactionMetaV3{} intx_meta, and for the API resource model emitted for Transaction for this row needs to show the same value for
result_meta_xdr
- 'asset_balance_changes` in `details` of API Operation resource model, will be empty when Op is of type  `InvokeHostFunctionOp`, i.e. it wont parse contract events from tx meta to derive this in operations processor.

(We should take care to ensure the list above is complete, it may be non-exhaustive.)

These may incur significant increases in the compute time needed to finish ingesting a ledger when the Soroban transaction volume increases, and we need a way to disable this at runtime as a safety valve.

A new Horizon configuration setting, `DISABLE_SOROBAN_INGEST_PROCESSORS`, which defaults to `FALSE`. When it's `TRUE`, it will trigger specific routines in the aforementioned processors to skip Soroban-related parsing/extraction.

**Note:** This should target the [v2.28.0](https://github.com/stellar/go/tree/release-horizon-v2.28.0) release first, not master.

Wait for Horizon's ingestion time to blow up as Soroban resource limits increase.

urvisavla:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Added `DISABLE_SOROBAN_INGEST_PROCESSORS` flag to optionally skip Soroban related ingestion processing. It's set to `false` by default. #5175 provides specs on specific processes and storage points that are elided.

### Why

Fixes #5175 

### Known limitations
Manually tested. So would be good to add unit tests to validate the flag and correct processing skipping.

### DIFF
--- a/services/horizon/internal/flags.go
+++ b/services/horizon/internal/flags.go
@@ -57,6 +57,8 @@ const (
 	EnableIngestionFilteringFlagName = "exp-enable-ingestion-filtering"
 	// DisableTxSubFlagName is the command line flag for disabling transaction submission feature of Horizon
 	DisableTxSubFlagName = "disable-tx-sub"
+	// SkipSorobanIngestionFlagName is the command line flag for disabling Soroban related ingestion processing
+	SkipSorobanIngestionFlagName = "disable-soroban-ingest-processors"
 
 	// StellarPubnet is a constant representing the Stellar public network
 	StellarPubnet = "pubnet"
@@ -728,6 +730,15 @@ func Flags() (*Config, support.ConfigOptions) {
 				" It automatically configures network settings, including %s, %s, and %s.",
 				StellarPubnet, StellarTestnet, NetworkPassphraseFlagName,
 				HistoryArchiveURLsFlagName, CaptiveCoreConfigPathName),
+			UsedInCommands: IngestionCommands,
+		},
+		&support.ConfigOption{
+			Name:           SkipSorobanIngestionFlagName,
+			ConfigKey:      &config.SkipSorobanIngestion,
+			OptType:        types.Bool,
+			FlagDefault:    false,
+			Required:       false,
+			Usage:          "excludes Soroban data during ingestion processing",
 			UsedInCommands: IngestionCommands,
 		},
 	}
--- a/services/horizon/internal/flags.go
+++ b/services/horizon/internal/flags.go
@@ -58,7 +58,7 @@ const (
 	// DisableTxSubFlagName is the command line flag for disabling transaction submission feature of Horizon
 	DisableTxSubFlagName = "disable-tx-sub"
 	// SkipSorobanIngestionFlagName is the command line flag for disabling Soroban related ingestion processing
-	SkipSorobanIngestionFlagName = "disable-soroban-ingest-processors"
+	SkipSorobanIngestionFlagName = "disable-soroban-ingest"
 
 	// StellarPubnet is a constant representing the Stellar public network
 	StellarPubnet = "pubnet"
--- a/services/horizon/internal/ingest/processor_runner.go
+++ b/services/horizon/internal/ingest/processor_runner.go
@@ -140,22 +140,12 @@ func (s *ProcessorRunner) buildTransactionProcessor(ledgersProcessor *processors
 	lazyLoaders := []horizonLazyLoader{accountLoader, assetLoader, lpLoader, cbLoader}
 	statsLedgerTransactionProcessor := processors.NewStatsLedgerTransactionProcessor()
 
-	var skipOperationType []xdr.OperationType
-
-	if s.config.SkipSorobanIngestion {
-		skipOperationType = []xdr.OperationType{
-			xdr.OperationTypeInvokeHostFunction,
-			xdr.OperationTypeExtendFootprintTtl,
-			xdr.OperationTypeRestoreFootprint,
-		}
-	}
-
 	tradeProcessor := processors.NewTradeProcessor(accountLoader,
 		lpLoader, assetLoader, s.historyQ.NewTradeBatchInsertBuilder())
 
 	processors := []horizonTransactionProcessor{
 		statsLedgerTransactionProcessor,
-		processors.NewEffectProcessor(accountLoader, s.historyQ.NewEffectBatchInsertBuilder(), s.config.NetworkPassphrase, skipOperationType),
+		processors.NewEffectProcessor(accountLoader, s.historyQ.NewEffectBatchInsertBuilder(), s.config.NetworkPassphrase, s.config.SkipSorobanIngestion),
 		ledgersProcessor,
 		processors.NewOperationProcessor(s.historyQ.NewOperationBatchInsertBuilder(), s.config.NetworkPassphrase),
 		tradeProcessor,
--- a/services/horizon/internal/ingest/processor_runner.go
+++ b/services/horizon/internal/ingest/processor_runner.go
@@ -111,9 +111,21 @@ func buildChangeProcessor(
 	source ingestionSource,
 	ledgerSequence uint32,
 	networkPassphrase string,
+	skipSorobanIngestion bool,
 ) *groupChangeProcessors {
 	statsChangeProcessor := &statsChangeProcessor{
 		StatsChangeProcessor: changeStats,
+	}
+
+	var skipEntryType []xdr.LedgerEntryType
+
+	if skipSorobanIngestion {
+		skipEntryType = []xdr.LedgerEntryType{
+			xdr.LedgerEntryTypeContractData,
+			xdr.LedgerEntryTypeContractCode,
+			xdr.LedgerEntryTypeTtl,
+			xdr.LedgerEntryTypeConfigSetting,
+		}
 	}
 
 	useLedgerCache := source == ledgerSource
@@ -127,7 +139,7 @@ func buildChangeProcessor(
 		processors.NewTrustLinesProcessor(historyQ),
 		processors.NewClaimableBalancesChangeProcessor(historyQ),
 		processors.NewLiquidityPoolsChangeProcessor(historyQ, ledgerSequence),
-	})
+	}, skipEntryType)
 }
 
 func (s *ProcessorRunner) buildTransactionProcessor(ledgersProcessor *processors.LedgersProcessor) *groupTransactionProcessors {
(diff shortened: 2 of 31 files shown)


services/horizon/internal/ingest/processor_runner.go
```diff
@@ -111,11 +111,23 @@ func buildChangeProcessor(
...
 	}
 
+	var skipEntryType []xdr.LedgerEntryType
+
+	if skipSorobanIngestion {
+		skipEntryType = []xdr.LedgerEntryType{
+			xdr.LedgerEntryTypeContractData,
+			xdr.LedgerEntryTypeContractCode,
+			xdr.LedgerEntryTypeTtl,
+			xdr.LedgerEntryTypeConfigSetting,
```
urvisavla, line 127: Is this Soroban related ledger entry type?

services/horizon/internal/ingest/processor_runner.go
```diff
@@ -111,11 +111,23 @@ func buildChangeProcessor(
 	source ingestionSource,
 	ledgerSequence uint32,
 	networkPassphrase string,
+	skipSorobanIngestion bool,
 ) *groupChangeProcessors {
 	statsChangeProcessor := &statsChangeProcessor{
```
urvisavla, line 116: Should we also skip it in transaction stats [StatsLedgerTransactionProcessor](https://github.com/stellar/go/blob/0292a74fffd433ab411f616403bca7e2619580a1/services/horizon/internal/ingest/processors/stats_ledger_transaction_processor.go#L131) ?

services/horizon/internal/ingest/processor_runner.go
2opremio: Yes, it contains network configuration settings for Soroban. And AFAICR only for Soroban

services/horizon/internal/ingest/processor_runner.go
sreuland: @urvisavla , I'm wondering if an alternate idea may be feasible for consideration - elide the soroban related meta once here at top/beginning of ledger ingestion from tx's in the incoming `LedgerCloseMeta` in [RunAllProcessorsOnLedger()](https://github.com/stellar/go/pull/5176/files#diff-ab79ee5d60a3baacf3e3323b77eaf66fe51c50e4380e3bbd32d5beaddcc50156L482), then all these downstream consumers don't need to change or have awareness of soroban types to skip , rather they just won't receive any meta that triggers their soroban processing routines.

eliding the meta from the incoming ledgerclosemeta would follow the same proposal as here in poc pr to prevent soroban tx meta from being stored in history table, which blanks out the tx xdr's meta if it was soroban:
https://github.com/stellar/go/pull/5172/files#diff-db88e1e0df8102b1614a43f058ea6bfd87d7c21a8cc907b1cdb12585ca70dc4fR179-R186

a somewhat pseudo example of doing the same here from processor_runner:
```
func (s *ProcessorRunner) RunAllProcessorsOnLedger(ledger xdr.LedgerCloseMeta) (
	stats ledgerStats,
	err error,
) {
        for tx_index := 0; tx_index < ledger.CountTransactions(); i++ {
             tx_meta := ledger.TxApplyProcessing(tx_index)
	    if v3_meta, ok := tx_meta.GetV3(); ok && v3_meta.SorobanMeta != nil {
                    // it's soroban, elide the tx meta, force empty meta into the xdr instead
		    tx_meta.V3 = &xdr.TransactionMetaV3{
			    Ext:             xdr.ExtensionPoint{},
			    TxChangesBefore: xdr.LedgerEntryChanges{},
			    Operations:      []xdr.OperationMeta{},
			    TxChangesAfter:  xdr.LedgerEntryChanges{},
			    SorobanMeta:     nil,
		    }
	    }
       }

       //  `ledger` now has no soroban tx meta, and gets passed to existing change and tx processor pipelines, which should silently skip most of their processing, since there's no meta to trigger anything.
      ... 

}


    
```

Shaptic: Hmm, I think this is a great idea. As long as it's done in a way that doesn't overlap with the other, meta eliding PR (#5174), it would be far less fragile than trying to identify the specific areas where we need to elide processing!

services/horizon/internal/ingest/processor_runner.go
```diff
@@ -142,11 +154,22 @@ func (s *ProcessorRunner) buildTransactionProcessor(ledgersProcessor *processors
...
 		lpLoader, assetLoader, s.historyQ.NewTradeBatchInsertBuilder())
 
+	var skipOperationType []xdr.OperationType
+
+	if s.config.SkipSorobanIngestion {
+		skipOperationType = []xdr.OperationType{
+			xdr.OperationTypeInvokeHostFunction,
+			xdr.OperationTypeExtendFootprintTtl,
+			xdr.OperationTypeRestoreFootprint,
+			xdr.OperationTypeExtendFootprintTtl,
```
Shaptic, line 164: repeated

services/horizon/internal/ingest/processor_runner.go
urvisavla: This is a great idea, I didn't realize this was an option. I tested the code snippet but it doesn't seem to be updating the XDR itself. Any thoughts on what might be the issue?

services/horizon/internal/flags.go
```diff
@@ -57,6 +57,8 @@ const (
 	EnableIngestionFilteringFlagName = "exp-enable-ingestion-filtering"
 	// DisableTxSubFlagName is the command line flag for disabling transaction submission feature of Horizon
 	DisableTxSubFlagName = "disable-tx-sub"
+	// SkipSorobanIngestionFlagName is the command line flag for disabling Soroban related ingestion processing
+	SkipSorobanIngestionFlagName = "disable-soroban-ingest-processors"
```
Shaptic, line 61: thoughts on simplifying the flag name? e.g. just `--disable-soroban-ingestion`

services/horizon/internal/ingest/processors/effects_processor.go
```diff
@@ -50,14 +53,29 @@ func (p *EffectProcessor) ProcessTransaction(
 		return nil
 	}
 
-	for opi, op := range transaction.Envelope.Operations() {
+	elidedTransaction := transaction
+
+	if p.skipSoroban &&
+		elidedTransaction.UnsafeMeta.V == 3 &&
+		elidedTransaction.UnsafeMeta.MustV3().SorobanMeta != nil {
```
Shaptic, line 60: nit: just accessing the struct will be slightly faster since you did the `V == 3` check already
```suggestion
		elidedTransaction.UnsafeMeta.V3.SorobanMeta != nil {
```

services/horizon/internal/ingest/processors/operations_processor.go
```diff
@@ -688,11 +697,18 @@ func (operation *transactionOperationWrapper) Details() (map[string]interface{},
 			}
 			details["parameters"] = params
 
-			if balanceChanges, err := operation.parseAssetBalanceChangesFromContractEvents(); err != nil {
-				return nil, err
+			var balanceChanges []map[string]interface{}
+			var parseErr error
+			if !operation.skipSorobanDetails {
```
Shaptic, line 702: nit suggestion: double-negatives can be annoying to read - can we invert the conditionals? i.e.
```suggestion
			if operation.skipSorobanDetails { // elide
```

services/horizon/internal/integration/invokehostfunction_test.go
```diff
@@ -24,13 +26,42 @@ const increment_contract = "soroban_increment_contract.wasm"
...
+func CaseContractInvokeHostFunctionInstallContract(t *testing.T) {
 	if integration.GetCoreMaxSupportedProtocol() < 20 {
 		t.Skip("This test run does not support less than Protocol 20")
 	}
 
 	itest := integration.NewTest(t, integration.Config{
-		ProtocolVersion:  20,
+		ProtocolVersion: 20,
+		HorizonEnvironment: map[string]string{
+			"DISABLE_SOROBAN_INGEST_PROCESSORS": fmt.Sprint(DisabledSoroban)},
```
Shaptic, line 64: does sprint work here w/o the formatting specifier? I thought you needed
```suggestion
			"DISABLE_SOROBAN_INGEST_PROCESSORS": strconv.FormatBool(DisabledSoroban)},
```

Shaptic (approved):
Straightforward solution and nice test coverage! :clap: :shipit: 
Coupla small nitpicks below :arrow_down: but it's good to merge as-is imo!

services/horizon/internal/integration/invokehostfunction_test.go
sreuland: yes, it outputs stringified true/false

services/horizon/internal/ingest/processors/operations_processor.go
sreuland: yep, done - https://github.com/stellar/go/pull/5176/commits/4765d5242e69008bc977377cbb7be80d017217c8

services/horizon/internal/ingest/processors/effects_processor.go
sreuland: ok, i like the encapsulated getter, but there's so much usage of the struct way, changed it.

services/horizon/internal/flags.go
sreuland: yep, good suggestion, changed it - https://github.com/stellar/go/pull/5176/commits/4765d5242e69008bc977377cbb7be80d017217c8

services/horizon/internal/integration/invokehostfunction_test.go
Shaptic: neat, thanks !

PR #5178, opened 2024-01-23
historyarchive: Make the library target the same log as Horizon

Shaptic:
### What
When using a cache for history archives, use the same log target to match Horizon.

### Why
Logs from this package do not appear in Horizon's logs.

### Known limitations
This means library users need to use the same logging setup as Horizon, but this is an attempt at a temporary fix, so this is an acceptable quirk for now.


sreuland (approved):
great find!

PR #5183, opened 2024-01-29
Merge 2.28.0 back to mainline

sreuland:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

merging 2.28.0 back to mainline, it has a few material changes that were done solely on release branch to start such as #5171 and #5175. And there was a significant conflict in histoaryarchive package between the branches as master had taken a refactoring from lighthorizin and 2.28. had a somewhat partial/different refactoring in the same package. i updated all to be consistent to the historyarchive package from master. 

### Why

part of post release steps


historyarchive/archive_pool.go
```diff
@@ -51,8 +51,18 @@ func NewArchivePool(archiveURLs []string, opts ArchiveOptions) (ArchivePool, err
 	return validArchives, nil
 }
 
+func (pa ArchivePool) GetStats() []ArchiveStats {
+	stats := []ArchiveStats{}
```
Shaptic, line 55: we can pre-allocate the known capacity here, yeah?
```suggestion
	stats := make([]ArchiveStats, 0, len(pa))
```

historyarchive/archive_pool.go
```diff
@@ -51,8 +51,18 @@ func NewArchivePool(archiveURLs []string, opts ArchiveOptions) (ArchivePool, err
...
+	for _, archive := range pa {
+		if len(archive.GetStats()) == 1 {
```
Shaptic, line 57: bit confused about the check here?

ingest/ledgerbackend/captive_core_backend_test.go
```diff
@@ -138,22 +140,33 @@ func TestCaptiveNew(t *testing.T) {
 	require.NoError(t, err)
 	defer os.RemoveAll(storagePath)
 
+	var userAgent string
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		userAgent = r.Header["User-Agent"][0]
```
Shaptic, line 145: just checking: why `[0]` here?

Shaptic (approved):
looks like a clean merge to me - thanks for resolving all of the conflicts!! :clap:

historyarchive/archive_pool.go
sreuland: i think was left over from initial impl, it works out now to just rollup whatever's in the array, good catch - https://github.com/stellar/go/pull/5183/commits/98d54d1623e2bd3f1370061b40fb04aa53dcc03c

historyarchive/archive_pool.go
sreuland: I think potential capacity on stats isn't [one-to-one irt archive](https://github.com/stellar/go/pull/5183/commits/98d54d1623e2bd3f1370061b40fb04aa53dcc03c), there could be more than one stats instance, not at present, it doesn't seem like big gain vs. code either way?

ingest/ledgerbackend/captive_core_backend_test.go
sreuland: the expectation is that only a single value for user-agent header is sent by the archive client when config'd correctly, the test should break if anything different emitted, i.e. multi or no value.

Shaptic: got it, thanks for clarity!

PR #5188, opened 2024-01-31
services/horizon + horizonclient: Add new async transaction submission endpoint

### LINKED ISSUE #4082: services/horizon: support async transaction submission

Horizon accepts transactions for submission to the network. It does this by:
1. Accepting the transaction.
2. Adding the transaction to a queue of transaction for the account.
3. Submitting the transaction to stellar-core when it is that transactions turn.
4. Waiting for the transaction result to be ingested.
5. Finally responding to the client with the transaction result.

It can take Horizon 5-30 seconds to perform the above steps.

A client's process/thread is held and blocked for the 5-30 seconds, consuming resources. During this time the client has no visibility into if the Horizon server has received the transaction or what is happening, and so the client must wait until it gets a response to know if any progress is being made.

Applications do not always want to keep a http connection open for ~5-30 seconds to find out the final state of the submitted transaction. In some cases applications would prefer to submit and lookup the transaction later to see if it was successful, failed, or just not included. In cases of scale and depending on the technologies the client is using, holding a http connection open for ~5-30 seconds may be prohibitive. Applications of scale usually avoid holding up processes for long periods of time. Applications that wish to submit multiple transactions must start multiple threads and handle error cases across each which is challenging to code and get right. 

Additionally, all applications need to monitor the state of the transaction in failure or timeout cases anyway, since 

@tomerweller and I found that when writing example code that submitted many transactions that Horizon's synchronous submission API made it difficult to adequately use resources on the client.

I also found that when writing code for github.com/stellar/starlight that Horizon's submission logic was inconvenient since I had to write concurrency code to be able to do other things while a transaction was sent to Horizon, or block the application while waiting unnecessarily.

In both cases it would have been much more convenient if immediate validation failure errors were available immediately and transaction results from being included in a ledger were available later.

I'd like an option for applications to submit transactions to Horizon asynchronously.

In other words, Horizon won't wait for the transaction result to be seen on the ledger before providing a response to the client.

An immediate failure from stellar-core's validation that runs during submission would still be returned immediately, but if stellar-core indicates that the transaction is `"PENDING"`, horizon would return immediately in that case so that the calling application can move on and check back on transaction result later.

I think the existing transaction queue to be utilized, so if multiple transactions are submitted horizon still orders their submission to stellar-core in order of their sequence number, as it does today. Preserving this behavior for async transaction submission means that async and sync submission would have many of the same properties and behaviors, all that would be different is how the caller experiences a response.

Keep things the way they are and require that clients block their own processes.

### Experimenting

I experimented with an incomplete implementation of this idea, although the transaction queue functionality was not preserved: https://github.com/stellar/go/pull/3564. I came to the conclusion during this experiment that preserving the transaction queue functionality is pretty important, otherwise the semantics between async and sync submission would be subtly different which would be difficult to understand as a user.

aditya1702:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Closes #4082 

Adds the following:
- New endpoint `transactions-async` that allows clients to submit transactions and get a response immediately. It does not block unlike the current endpoint.
- Unittests for the endpoint.
- Integration tests for the endpoint.
- Corresponding code in `horizonclient` SDK.

### Why

- The current endpoint blocks until Horizon ingests the transaction in the DB. This implementation removes the wait and returns an immediate response.
- It brings the transaction submission process in sync with Soroban-RPC's implementation.

More on the requirements about this endpoint here: https://docs.google.com/document/d/1uiEnZJ-Yv1-8fBe8A8FDeR_HbpsyQLnY7KnlsTKl4vk/edit

### Known limitations

Once implementation is finalised from the team, I will update the documentation wherever necessary in the stellar-docs.


services/horizon/internal/httpx/router.go
```diff
@@ -331,6 +334,17 @@ func (r *Router) addRoutes(config *RouterConfig, rateLimiter *throttled.HTTPRate
 		CoreStateGetter:   config.CoreGetter,
 	}})
 
+	// Async Transaction submission API
+	r.Method(http.MethodPost, "/v2/transactions", ObjectActionHandler{actions.AsyncSubmitTransactionHandler{
```
mollykarcher, line 338: I'd prefer not to use this naming scheme for a couple reasons:
1. URI/path versioning like this is typically implemented as the version for the _entire_ API, so it's unexpected/strange to have a single endpoint with a `v2` in it, while all other endpoints are absent a version in their path
2. `v2` implies that people should be moving away from the original `transactions` endpoint and towards this one, but that's not really the case here. We're offering a new version we think might be useful, but there's no concerted need to push people towards it or deprecate the older one

I would just recommend something like `/transactions-async`/`/async-transactions`. I wouldn't even be opposed to keeping it in the same endpoint but gating the functionality with a query param (ex. `/transactions?async=true`)

mollykarcher: It might be easier to flesh out all of this request/response format stuff in the design [doc](https://docs.google.com/document/d/1uiEnZJ-Yv1-8fBe8A8FDeR_HbpsyQLnY7KnlsTKl4vk/edit?disco=AAABE0uhZuc)

.gitignore
```diff
@@ -5,6 +5,7 @@
 /services/horizon/captive-core
 /services/horizon/horizon
 /services/horizon/stellar-horizon
+/bucket-cache
```
tamirms, line 8: seems unrelated to the rest of the changes in this PR

aditya1702: @tamirms If we dont have this in gitignore, then all the bucket cache files show up in the changed files. However, yes, this change is not related to the PR. Do you want me to include this separately?

tamirms: > Do you want me to include this separately?

no worries, you can leave it in. but for next time it is generally preferable to separate unrelated changes in different PRs

protocols/horizon/main.go
```diff
@@ -567,6 +567,27 @@ type InnerTransaction struct {
...
+	ErrorResultXDR string `json:"errorResultXdr,omitempty"`
+	// DiagnosticEventsXDR is present only if Status is equal to proto.TXStatusError.
+	// DiagnosticEventsXDR is a base64-encoded slice of xdr.DiagnosticEvent
+	DiagnosticEventsXDR string `json:"diagnosticEventsXdr,omitempty"`
+	// TxStatus represents the status of the transaction submission returned by stellar-core.
+	// It can be one of: proto.TXStatusPending, proto.TXStatusDuplicate,
+	// proto.TXStatusTryAgainLater, or proto.TXStatusError.
+	TxStatus string `json:"tx_status"`
+	// HttpStatus represents the corresponding http status code.
+	HttpStatus int `json:"status"`
```
tamirms, line 585: the http status is already part of the response so there is no need to include it in the response body

protocols/horizon/main.go
```diff
@@ -567,6 +567,27 @@ type InnerTransaction struct {
...
+// AsyncTransactionSubmissionResponse represents the response returned by Horizon
+// when using the transaction-async endpoint.
+type AsyncTransactionSubmissionResponse struct {
+	// ErrorResultXDR is present only if Status is equal to proto.TXStatusError.
+	// ErrorResultXDR is a TransactionResult xdr string which contains details on why
+	// the transaction could not be accepted by stellar-core.
+	ErrorResultXDR string `json:"errorResultXdr,omitempty"`
+	// DiagnosticEventsXDR is present only if Status is equal to proto.TXStatusError.
+	// DiagnosticEventsXDR is a base64-encoded slice of xdr.DiagnosticEvent
+	DiagnosticEventsXDR string `json:"diagnosticEventsXdr,omitempty"`
```
tamirms, line 579: I'm not sure if we want to include diagnostic events in the horizon response. This seems like it would be more relevant for soroban-rpc than horizon

cc @2opremio @mollykarcher

services/horizon/internal/actions/submit_transaction_async.go
```diff
@@ -0,0 +1,164 @@
...
+type AsyncSubmitTransactionHandler struct {
+	NetworkPassphrase string
+	DisableTxSub      bool
+	ClientWithMetrics stellarcore.ClientWithMetricsInterface
+	CoreStateGetter
+}
+
+// AsyncTransactionSubmissionResponse represents the response returned by Horizon
+// when using the transaction-submission-async endpoint.
+type AsyncTransactionSubmissionResponse struct {
```
tamirms, line 27: this seems identical to the struct defined in protocols/horizon/main.go . Do we need to duplicate it here?

services/horizon/internal/actions/submit_transaction_async.go
```diff
@@ -0,0 +1,164 @@
...
+	hProblem "github.com/stellar/go/services/horizon/internal/render/problem"
+	"github.com/stellar/go/support/render/problem"
+	"net/http"
+)
+
+const (
+	HTTPStatusCodeForPending       = http.StatusCreated
+	HTTPStatusCodeForDuplicate     = http.StatusConflict
+	HTTPStatusCodeForTryAgainLater = http.StatusServiceUnavailable
+	HTTPStatusCodeForError         = http.StatusBadRequest
```
tamirms, line 15: I think it would be cleaner to avoid having these constants and just refer to the http statuses directly in the handler.

services/horizon/internal/actions/submit_transaction.go
```diff
@@ -29,53 +25,6 @@ type SubmitTransactionHandler struct {
 	CoreStateGetter
 }
 
-type envelopeInfo struct {
```
tamirms, line 32: I think it's fine to leave these functions in this file instead of moving them to utils.go because that would minimize the diff in the code review. These functions are only relevant to transaction submission rather than general utility functions that can be used by other endpoints

clients/stellarcore/metrics_client.go
```diff
@@ -0,0 +1,130 @@
...
+
+type ClientWithMetricsInterface interface {
+	SubmitTransaction(ctx context.Context, rawTx string, envelope xdr.TransactionEnvelope) (resp *proto.TXResponse, err error)
+	UpdateTxSubMetrics(duration float64, envelope xdr.TransactionEnvelope, response *proto.TXResponse, err error)
+}
+
+type ClientWithMetrics struct {
+	CoreClient ClientInterface
+
+	AsyncTxSubMetrics struct {
```
tamirms, line 19: it seems these metrics are duplicated between the existing txsub implementation and the new async txsub implementation. Could you instead define and update this metrics only in here and then have both the legacy txsub and async txsub use this client with separate prometheuse subsystems?

clients/stellarcore/metrics_client.go
```diff
@@ -0,0 +1,130 @@
...
+	"context"
+	"github.com/prometheus/client_golang/prometheus"
+	proto "github.com/stellar/go/protocols/stellarcore"
+	"github.com/stellar/go/xdr"
+	"time"
+)
+
+type ClientWithMetricsInterface interface {
+	SubmitTransaction(ctx context.Context, rawTx string, envelope xdr.TransactionEnvelope) (resp *proto.TXResponse, err error)
+	UpdateTxSubMetrics(duration float64, envelope xdr.TransactionEnvelope, response *proto.TXResponse, err error)
```
tamirms, line 13: this seems like a private helper function and therefore should not be exposed publicly in the interface

clients/stellarcore/client.go
```diff
@@ -29,6 +29,18 @@ type Client struct {
...
+type ClientInterface interface {
+	Upgrade(ctx context.Context, version int) (err error)
+	GetLedgerEntry(ctx context.Context, ledgerKey xdr.LedgerKey) (proto.GetLedgerEntryResponse, error)
+	Info(ctx context.Context) (resp *proto.InfoResponse, err error)
+	SetCursor(ctx context.Context, id string, cursor int32) (err error)
+	SubmitTransaction(ctx context.Context, envelope string) (resp *proto.TXResponse, err error)
+	WaitForNetworkSync(ctx context.Context) error
+	ManualClose(ctx context.Context) (err error)
+	http() HTTP
+	simpleGet(ctx context.Context, newPath string, query url.Values) (*http.Request, error)
```
tamirms, line 41: private functions should not be included in an interface

clients/stellarcore/client.go
```diff
@@ -29,6 +29,18 @@ type Client struct {
 	URL string
 }
 
+type ClientInterface interface {
```
tamirms, line 32: https://go.dev/wiki/CodeReviewComments#interfaces

services/horizon/internal/actions/submit_transaction_async.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+
+	coreState := handler.GetCoreState()
+	if !coreState.Synced {
+		return nil, hProblem.StaleHistory
+	}
+
+	resp, err := handler.ClientWithMetrics.SubmitTransaction(r.Context(), info.raw, info.parsed)
+	if err != nil {
+		return nil, &problem.P{
```
tamirms, line 93: I don't think this error case is because the request is bad. This seems more like an internal error

services/horizon/internal/actions/submit_transaction_async.go
```diff
@@ -0,0 +1,164 @@
...
+				"https://developers.stellar.org/api/errors/http-status-codes/horizon-specific/transaction-submission-async/transaction_submission_failed",
+			Extras: map[string]interface{}{
+				"envelope_xdr": raw,
+				"error":        err,
+			},
+		}
+	}
+
+	if resp.IsException() {
+		return nil, &problem.P{
```
tamirms, line 109: what's the reasoning for this case to be classified as a bad request instead of an internal error?

services/horizon/internal/actions/submit_transaction_async.go
tamirms: actually a better idea would be to create a map:

```
var coreStatusToHTTPStatus = map[string]int{
   proto.TXStatusPending: http.StatusCreated,
}
```

.gitignore
aditya1702: @tamirms Will keep in mind for future PRs

services/horizon/internal/actions/submit_transaction_async.go
aditya1702: I added the struct in protocols/horizon/main later on. Yeah this can be reused here and no need for duplication.

services/horizon/internal/actions/submit_transaction_async.go
aditya1702: Same as the above one

protocols/horizon/main.go
aditya1702: @tamirms I did not understand. Where is the http response included in the response? Which response are you talking about here?

clients/horizonclient/client.go
```diff
@@ -521,6 +521,78 @@ func (c *Client) SubmitTransactionWithOptions(transaction *txnbuild.Transaction,
...
+//
+// If you want to skip this check, use SubmitTransactionWithOptions.
+func (c *Client) AsyncSubmitFeeBumpTransaction(transaction *txnbuild.FeeBumpTransaction) (txResp hProtocol.AsyncTransactionSubmissionResponse, err error) {
+	return c.AsyncSubmitFeeBumpTransactionWithOptions(transaction, SubmitTxOpts{})
+}
+
+// AsyncSubmitFeeBumpTransactionWithOptions submits an async fee bump transaction to the network, allowing
+// you to pass SubmitTxOpts. err can be either an error object or a horizon.Error object.
+func (c *Client) AsyncSubmitFeeBumpTransactionWithOptions(transaction *txnbuild.FeeBumpTransaction, opts SubmitTxOpts) (txResp hProtocol.AsyncTransactionSubmissionResponse, err error) {
+	// only check if memo is required if skip is false and the inner transaction
```
sreuland, line 546: for consideration, some code de-dupe, instead of repeating this same validate/serialize block in both async/sync methods, pull up lines 546-559 into `validateFeeBumpTx(transaction *txnbuild.FeeBumpTransaction, opts SubmitTxOpts) (txeBase64 string, err error)` and use from both async/sync methods

clients/stellarcore/metrics_client.go
```diff
@@ -0,0 +1,129 @@
...
+	proto "github.com/stellar/go/protocols/stellarcore"
+	"github.com/stellar/go/xdr"
+	"time"
+)
+
+type ClientWithMetricsInterface interface {
+	SubmitTransaction(ctx context.Context, rawTx string, envelope xdr.TransactionEnvelope) (resp *proto.TXResponse, err error)
+}
+
+type ClientWithMetrics struct {
```
sreuland, line 15: for encapsulation, could make this private as` type clientWithMetrics struct`, rename the interface to ClientWithMetrics and then add factory method `func NewClientWithMetrics() ClientWithMetrics`

protocols/horizon/main.go
sreuland: I think is referring to the HTTP 'Status' header, which gets set in these layers by go action-handler/router which handler can optionally return an `problem.P` error with Status set to a non 200 http status code, which ultimately gets placed as http `Status` header on response back to client. an example here - https://github.com/stellar/go/pull/5188/files#diff-d76b3705da30a9483cd8e70a9b0c62c016ddd85eabf5becf668f0e317c16f307R165

services/horizon/internal/integration/txsub_async_test.go
```diff
@@ -0,0 +1,118 @@
...
+
+import (
+	"github.com/stellar/go/protocols/horizon"
+	"github.com/stellar/go/services/horizon/internal/test/integration"
+	"github.com/stellar/go/txnbuild"
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func TestAsyncTxSub_SuccessfulSubmission(t *testing.T) {
```
sreuland, line 11: for consideration, adding a test to confirm `GET /transactions-async` is success response, returns oapi doc.

clients/stellarcore/metrics_client.go
aditya1702: @sreuland Made the change for ClientWithMetrics, however, the ClientInterface will require a bigger change so defering it for now.

services/horizon/internal/httpx/router.go
```diff
@@ -331,6 +333,17 @@ func (r *Router) addRoutes(config *RouterConfig, rateLimiter *throttled.HTTPRate
 		CoreStateGetter:   config.CoreGetter,
 	}})
 
+	// Async Transaction submission API
+	r.Method(http.MethodPost, "/transactions-async", ObjectActionHandler{actions.AsyncSubmitTransactionHandler{
```
JakeUrban, line 337: I think we need to continue using underscores to match the other multi-word endpoints

aditya1702: Yeah makes sense.

protocols/horizon/main.go
aditya1702: @tamirms We do include the diagnostic events in normal transaction submission response when the core resp is an error - https://github.com/stellar/go/blob/master/services/horizon/internal/actions/submit_transaction.go#L116

protocols/horizon/main.go
tamirms: @aditya1702 here is a description of responses in the HTTP protocol https://en.wikipedia.org/wiki/HTTP#HTTP/1.1_response_messages

Note how there is a a status line containing the [response status code](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes). So, I don't understand why you want to duplicate that information and put it in the response body

services/horizon/internal/actions/submit_transaction_async.go
```diff
@@ -0,0 +1,135 @@
...
+			Hash:                info.hash,
+		}, nil
+	case proto.TXStatusPending, proto.TXStatusDuplicate, proto.TXStatusTryAgainLater:
+		return horizon.AsyncTransactionSubmissionResponse{
+			TxStatus:   resp.Status,
+			HttpStatus: coreStatusToHTTPStatus[resp.Status],
+			Hash:       info.hash,
+		}, nil
+	default:
+		return nil, &problem.P{
```
tamirms, line 120: if we never expect to enter this case the status should be an internal error and we should log the unexpected status

protocols/horizon/main.go
aditya1702: Talked with Fons and we can remove the diagnostic events

support/render/hal/handler.go
```diff
@@ -1,16 +0,0 @@
-package hal
```
tamirms, line 1: why is this file and support/render/hal/io.go deleted in this pr?

clients/stellarcore/metrics_client.go
```diff
@@ -0,0 +1,129 @@
...
+		Name:       "submission_duration_seconds",
+		Help:       "submission durations to Stellar-Core, sliding window = 10m",
+		Objectives: map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.99: 0.001},
+	}, []string{"status"})
+	submissionsCounter := prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace: "horizon",
+		Subsystem: prometheusSubsystem,
+		Name:      "submissions_count",
+		Help:      "number of submissions, sliding window = 10m",
+	}, []string{"status"})
```
tamirms, line 85: this metric is not necessary because the submission duration summary already includes a count, see https://prometheus.io/docs/concepts/metric_types/#summary

clients/stellarcore/metrics_client.go
```diff
@@ -0,0 +1,129 @@
...
+		Subsystem: prometheusSubsystem,
+		Name:      "v1_count",
+		Help:      "number of v1 transaction envelopes submitted, sliding window = 10m",
+	}, []string{"status"})
+	feeBumpTransactionsCounter := prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace: "horizon",
+		Subsystem: prometheusSubsystem,
+		Name:      "feebump_count",
+		Help:      "number of fee bump transaction envelopes submitted, sliding window = 10m",
+	}, []string{"status"})
```
tamirms, line 103: I think you can get rid of these metrics and just add an envelope_type label to submissionDuration

clients/horizonclient/internal.go
```diff
@@ -27,7 +27,10 @@ func decodeResponse(resp *http.Response, object interface{}, horizonUrl string,
 	}
 	setCurrentServerTime(u.Hostname(), resp.Header["Date"], clock)
 
-	if !(resp.StatusCode >= 200 && resp.StatusCode < 300) {
+	// While this part of code assumes that any error < 200 or error >= 300 is a Horizon problem, it is not
+	// true for the response from /transactions_async endpoint which does give these codes for certain responses
+	// from core.
+	if !(resp.StatusCode >= 200 && resp.StatusCode < 300) && (resp.Request == nil || resp.Request.URL == nil || resp.Request.URL.Path != "/transactions_async") {
```
tamirms, line 33: are there any tests which exercise this new behavior?

clients/stellarcore/metrics_client.go
```diff
@@ -0,0 +1,129 @@
+package stellarcore
+
+import (
+	"context"
```
tamirms, line 4: we try to group imports in a specific order, see https://github.com/stellar/go/pull/5004#discussion_r1289005979

services/horizon/internal/actions/submit_transaction_async.go
```diff
@@ -0,0 +1,147 @@
...
+	proto "github.com/stellar/go/protocols/stellarcore"
+	hProblem "github.com/stellar/go/services/horizon/internal/render/problem"
+	"github.com/stellar/go/support/errors"
+	"github.com/stellar/go/support/log"
+	"github.com/stellar/go/support/render/problem"
+	"net/http"
+)
+
+var (
+	logger = log.New().WithField("service", "async-txsub")
```
tamirms, line 15: there is a logger attached to the request context. checkout the `Ctx()` function in https://github.com/stellar/go/blob/master/support/log/main.go#L58 and how it's used. we should use that function to obtain the logger attached to the request

services/horizon/internal/actions/submit_transaction_async.go
```diff
@@ -0,0 +1,147 @@
...
+	DisableTxSub      bool
+	ClientWithMetrics stellarcore.ClientWithMetrics
+	CoreStateGetter
+}
+
+func (handler AsyncSubmitTransactionHandler) GetResource(_ HeaderWriter, r *http.Request) (interface{}, error) {
+	logger.SetLevel(log.DebugLevel)
+
+	if err := validateBodyType(r); err != nil {
+		logger.WithError(err).Error("Could not validate request body type")
```
tamirms, line 29: we don't need to log errors where the user provides a malformed transaction. we should only log cases which are unexpected

services/horizon/internal/actions/submit_transaction_async.go
```diff
@@ -0,0 +1,147 @@
...
+				"convenience.",
+			Extras: map[string]interface{}{
+				"envelope_xdr": raw,
+			},
+		}
+	}
+
+	coreState := handler.GetCoreState()
+	if !coreState.Synced {
+		logger.WithField("envelope_xdr", raw).Error("Stellar-core is not synced")
```
tamirms, line 73: no need to log this

services/horizon/internal/actions/submit_transaction_async.go
```diff
@@ -0,0 +1,147 @@
...
+				"To enable it again, remove env variable DISABLE_TX_SUB.",
+			Extras: map[string]interface{}{
+				"envelope_xdr": raw,
+			},
+		}
+	}
+
+	info, err := extractEnvelopeInfo(raw, handler.NetworkPassphrase)
+	if err != nil {
+		logger.WithField("envelope_xdr", raw).Error("Could not parse transaction envelope")
```
tamirms, line 55: no need to log this

services/horizon/internal/actions/submit_transaction_async.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	raw, err := getString(r, "tx")
+	if err != nil {
+		logger.WithError(err).Error("Could not read transaction string from request URL")
+		return nil, err
+	}
+
+	if handler.DisableTxSub {
+		logger.WithField("envelope_xdr", raw).Error("Could not submit transaction: transaction submission is disabled")
```
tamirms, line 40: no need to log this

services/horizon/internal/actions/submit_transaction_async.go
```diff
@@ -0,0 +1,147 @@
...
+	logger.SetLevel(log.DebugLevel)
+
+	if err := validateBodyType(r); err != nil {
+		logger.WithError(err).Error("Could not validate request body type")
+		return nil, err
+	}
+
+	raw, err := getString(r, "tx")
+	if err != nil {
+		logger.WithError(err).Error("Could not read transaction string from request URL")
```
tamirms, line 35: no need to log this

clients/horizonclient/internal.go
tamirms: nevermind, I think this is exercised in services/horizon/internal/integration/txsub_async_test.go

clients/stellarcore/metrics_client.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	proto "github.com/stellar/go/protocols/stellarcore"
+	"github.com/stellar/go/xdr"
+	"time"
+)
+
+type ClientWithMetrics interface {
```
tamirms, line 11: https://go.dev/wiki/CodeReviewComments#interfaces

support/render/httpjson/io.go
```diff
@@ -27,7 +36,12 @@ func renderToString(data interface{}, pretty bool) ([]byte, error) {
 // Render write data to w, after marshaling to json. The response header is
 // set based on cType.
 func Render(w http.ResponseWriter, data interface{}, cType contentType) {
-	RenderStatus(w, http.StatusOK, data, cType)
+	statusCode := http.StatusOK
+	if asyncTxSubResponse, ok := data.(horizon.AsyncTransactionSubmissionResponse); ok {
+		statusCode = coreStatusToHTTPStatus[asyncTxSubResponse.TxStatus]
```
tamirms, line 41: What if there's another endpoint that requires custom http status codes in their response? we'd have to add another if statement to this function:

```go
	if asyncTxSubResponse, ok := data.(horizon.AsyncTransactionSubmissionResponse); ok {
		statusCode = coreStatusToHTTPStatus[asyncTxSubResponse.TxStatus]
	} else if otherResponse, ok := data.(horizon.MyEndpointResponse); ok {
		...
	}
```

it would be better to define an interface with a `HTTPStatus() int` function. That would provide a general way for any response to define its own custom http status behavior rather than implementing that logic here.

services/horizon/internal/httpx/router.go
```diff
@@ -372,6 +385,15 @@ func (r *Router) addRoutes(config *RouterConfig, rateLimiter *throttled.HTTPRate
 		w.Header().Set("Content-Type", "application/openapi+yaml")
 		w.Write(p)
 	})
+	r.Internal.Get("/transactions_async", func(w http.ResponseWriter, r *http.Request) {
+		p, err := staticFiles.ReadFile("static/txsub_async_oapi.yaml")
+		if err != nil {
+			w.WriteHeader(http.StatusNotFound)
```
tamirms, line 391: this should probably be an internal error since we don't expect `staticFiles.ReadFile("static/txsub_async_oapi.yaml")` to fail

support/render/httpjson/io.go
aditya1702: Yep makes sense

clients/stellarcore/metrics_client.go
aditya1702: @tamirms I see it is written in the link that we should not create interfaces for mocking, but we have done it in Horizon code at a lot of places. How should I then generate mocks for unittesting? Do you have any suggestions?

services/horizon/internal/actions/submit_transaction_async.go
```diff
@@ -0,0 +1,152 @@
...
+	case proto.TXStatusError, proto.TXStatusPending, proto.TXStatusDuplicate, proto.TXStatusTryAgainLater:
+		response := horizon.AsyncTransactionSubmissionResponse{
+			TxStatus: resp.Status,
+			Hash:     info.hash,
+		}
+
+		if resp.Status == proto.TXStatusError {
+			logger.WithFields(log.F{
+				"envelope_xdr": raw,
+				"error_xdr":    resp.Error,
```
Shaptic, line 127: Is this the same in the sync endpoint? I thought it just used the result_xdr field the same way. Also made me realize that https://developers.stellar.org/network/horizon/resources/submit-a-transaction is lacking error cases.

services/horizon/internal/actions/submit_transaction_async_test.go
```diff
@@ -0,0 +1,214 @@
...
+	TxHash = "3389e9f0f1a65f19736cacf544c2e825313e8447f569233bb8db39aa607c8889"
+)
+
+func createRequest() *http.Request {
+	form := url.Values{}
+	form.Set("tx", TxXDR)
+
+	request, _ := http.NewRequest(
+		"POST",
+		"http://localhost:8000/v2/transactions_async",
```
Shaptic, line 32: I thought we weren’t doing versioning

services/horizon/internal/actions/submit_transaction_async_test.go
```diff
@@ -0,0 +1,214 @@
...
+		}
+
+		request := createRequest()
+		w := httptest.NewRecorder()
+
+		resp, err := handler.GetResource(w, request)
+		assert.NoError(t, err)
+		assert.Equal(t, resp, testCase.expectedResponse)
+	}
+}
```
Shaptic, line 214: Nice coverage

services/horizon/internal/integration/txsub_async_test.go
```diff
@@ -0,0 +1,125 @@
...
+			TimeBounds:   txnbuild.NewInfiniteTimeout(),
+			LedgerBounds: &txnbuild.LedgerBounds{MinLedger: 0, MaxLedger: 100},
+		},
+	}
+
+	txResp, err := itest.AsyncSubmitTransaction(master, txParams)
+	assert.NoError(t, err)
+	assert.Equal(t, txResp, horizon.AsyncTransactionSubmissionResponse{
+		TxStatus: "PENDING",
+		Hash:     "6cbb7f714bd08cea7c30cab7818a35c510cbbfc0a6aa06172a1e94146ecf0165",
```
Shaptic, line 39: How did you predict this hash? 😲

services/horizon/internal/integration/txsub_async_test.go
```diff
@@ -0,0 +1,125 @@
...
+		Hash:           "6cbb7f714bd08cea7c30cab7818a35c510cbbfc0a6aa06172a1e94146ecf0165",
+	})
+
+	txResp, err = itest.AsyncSubmitTransaction(master, txParams)
+	assert.NoError(t, err)
+	assert.Equal(t, txResp, horizon.AsyncTransactionSubmissionResponse{
+		ErrorResultXDR: "",
+		TxStatus:       "TRY_AGAIN_LATER",
+		Hash:           "d5eb72a4c1832b89965850fff0bd9bba4b6ca102e7c89099dcaba5e7d7d2e049",
+	})
```
Shaptic, line 110: It’d be good to have an integration case that confirms that the transaction can eventually be retrieved in the PENDING, as well, ie test the GET /transactions/:id path that we expect people to take, as well as one that is PENDING then fails later.

Shaptic (commented):
Solid test coverage, just a few small comments below ⬇️

services/horizon/internal/actions/submit_transaction_async.go
```diff
@@ -0,0 +1,152 @@
...
+		return nil, err
+	}
+
+	if handler.DisableTxSub {
+		return nil, &problem.P{
+			Type:   "transaction_submission_disabled",
+			Title:  "Transaction Submission Disabled",
+			Status: http.StatusMethodNotAllowed,
+			Detail: "Transaction submission has been disabled for Horizon. " +
+				"To enable it again, remove env variable DISABLE_TX_SUB.",
```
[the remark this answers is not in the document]
Shaptic: Like I’d rather say something like “reach out to the provider” rather than give specific horizon flags

services/horizon/internal/integration/txsub_async_test.go
JakeUrban: Correct me if I'm wrong, but transactions in `PENDING` should return 404 from `GET /transactions/:id` since they haven't been included in a ledger yet right?

services/horizon/internal/httpx/static/txsub_async_oapi.yaml
```diff
@@ -0,0 +1,168 @@
+openapi: 3.0.0
```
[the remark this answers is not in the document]
aditya1702: Thank you!

services/horizon/internal/actions/submit_transaction_async.go
```diff
@@ -0,0 +1,152 @@
...
+			Extras: map[string]interface{}{
+				"envelope_xdr": raw,
+			},
+		}
+	}
+
+	info, err := extractEnvelopeInfo(raw, handler.NetworkPassphrase)
+	if err != nil {
+		return nil, &problem.P{
```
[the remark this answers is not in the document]
aditya1702: Yeah nice suggestion 👍

services/horizon/internal/actions/submit_transaction_async_test.go
aditya1702: Left it here by mistake.

services/horizon/internal/actions/submit_transaction_async.go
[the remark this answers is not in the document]
aditya1702: @Shaptic This flag has already been introduced last year. Its mainly for separating ingesting and web/txsub instances for horizon. Was that what you were asking?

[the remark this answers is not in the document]
Shaptic: No I understand the purpose of the flag, I'm saying it seems unnecessary to include the specific CLI flag that the operator needs to change in the actual error response which will be seen by a regular user.

clients/stellarcore/metrics_client.go
tamirms: interfaces should be defined where they are used. the transaction submission endpoint already defines an interface for the core client:

https://github.com/stellar/go/blob/9f50dddb3997ab9cbac812f0612d086652f21691/services/horizon/internal/actions/submit_transaction.go#L21-L23

so I don't see the need to define the `ClientWithMetrics` interface

support/render/hal/handler.go
tamirms: the circular import only exists because of https://github.com/stellar/go/pull/5188#discussion_r1544149260 . Once that is addressed you won't have the circular import

clients/stellarcore/metrics_client.go
tamirms: I just realized that there is a separate handler for the async endpoint:

https://github.com/stellar/go/blob/9f50dddb3997ab9cbac812f0612d086652f21691/services/horizon/internal/actions/submit_transaction_async.go

So you it is more appropriate to define an interface there with the minimal set of functions required for your use case

support/render/httpjson/io.go
tamirms: [this](https://github.com/stellar/go/pull/5188/commits/e9318930ab8244bf180745fc72eadf9555736a4f) isn't what I had in mind.

In services/horizon/internal/httpx/handler.go , you can define an interface with a `HTTPStatus() int` function. If the response matches that interface, invoke `HTTPStatus()` to determine the http status, otherwise default to 200.

services/horizon/internal/actions/submit_transaction_async.go
```diff
@@ -93,7 +88,6 @@ func (handler AsyncSubmitTransactionHandler) GetResource(_ HeaderWriter, r *http
 	}
 
 	if resp.IsException() {
-		logger.WithField("envelope_xdr", raw).WithError(errors.Errorf(resp.Exception)).Error("Transaction submission exception from stellar-core")
```
tamirms: we should still be logging this case as exceptions from core are unexpected

services/horizon/internal/actions/submit_transaction_async.go
```diff
@@ -0,0 +1,152 @@
...
+
+	switch resp.Status {
+	case proto.TXStatusError, proto.TXStatusPending, proto.TXStatusDuplicate, proto.TXStatusTryAgainLater:
+		response := horizon.AsyncTransactionSubmissionResponse{
+			TxStatus: resp.Status,
+			Hash:     info.hash,
+		}
+
+		if resp.Status == proto.TXStatusError {
+			logger.WithFields(log.F{
```
tamirms, line 125: we don't need to log this case because core can return an error response if the transaction is invalid

services/horizon/internal/actions/submit_transaction_async.go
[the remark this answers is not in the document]
aditya1702: I am going to add a TODO for this and do it later. Want to get this merged now 😅

services/horizon/internal/actions/submit_transaction_async.go
aditya1702: Oops forgot to remove this one. Thanks

services/horizon/internal/actions/submit_transaction_async.go
aditya1702: @Shaptic Yeah I will update the docs also after this gets merged.

clients/stellarcore/metrics_client.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/stellar/go/xdr"
+)
+
+type ClientWithMetrics struct {
+	CoreClient Client
+
+	TxSubMetrics struct {
+		// SubmissionDuration exposes timing metrics about the rate and latency of
+		// submissions to stellar-core
+		SubmissionDuration *prometheus.SummaryVec
```
tamirms, line 19: why is this part of `TxSubMetrics` instead of being included in `ClientWithMetrics` directly?

also, `SubmissionDuration` does not need to be public since this field should only be accessed internally. Similarly, I don't see why `CoreClient` is a public field

clients/stellarcore/metrics_client.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+}
+
+func (c ClientWithMetrics) SubmitTransaction(ctx context.Context, rawTx string, envelope xdr.TransactionEnvelope) (*proto.TXResponse, error) {
```
tamirms, line 23: this function should take either a `rawTx string` parameter or a `envelope xdr.TransactionEnvelope` parameter but not both because it is expected that `rawTx` must be equivalent to `envelope`. Therefore it is possible to derive one from the other. By allowing the function to take both parameters you allow the possibility for the function to be used incorrectly because someone can call the function with `rawTx` and `envelope` values which don't match

clients/stellarcore/mocks.go
```diff
@@ -0,0 +1,22 @@
...
+	mock.Mock
+}
+
+// SubmitTransaction mocks the SubmitTransaction method
+func (m *MockClientWithMetrics) SubmitTransaction(ctx context.Context, rawTx string, envelope xdr.TransactionEnvelope) (*proto.TXResponse, error) {
+	args := m.Called(ctx, rawTx, envelope)
+	return args.Get(0).(*proto.TXResponse), args.Error(1)
+}
+
+func (m *MockClientWithMetrics) UpdateTxSubMetrics(duration float64, envelope xdr.TransactionEnvelope, response *proto.TXResponse, err error) {
```
tamirms, line 20: this function is no longer necessary

clients/stellarcore/mocks.go
```diff
@@ -0,0 +1,22 @@
+package stellarcore
+
+import (
+	"context"
+	proto "github.com/stellar/go/protocols/stellarcore"
+	"github.com/stellar/go/xdr"
+	"github.com/stretchr/testify/mock"
+)
+
+type MockClientWithMetrics struct {
```
tamirms, line 10: given that this struct is only used in services/horizon/internal/actions/submit_transaction_async_test.go . It would be better to move this code to that package

services/horizon/internal/actions/submit_transaction_async.go
```diff
@@ -0,0 +1,139 @@
...
+	raw, err := getString(r, "tx")
+	if err != nil {
+		return nil, err
+	}
+
+	if handler.DisableTxSub {
+		return nil, &problem.P{
+			Type:   "transaction_submission_disabled",
+			Title:  "Transaction Submission Disabled",
+			Status: http.StatusMethodNotAllowed,
```
tamirms, line 44: method not allowed refers to the HTTP method (E.g. POST, GET, PUT, etc). Using it here doesn't make sense because we expect the request to be a POST. I suggest using a 403 status code. This comment also applies to the disable tx sub check in services/horizon/internal/actions/submit_transaction.go

services/horizon/internal/test/integration/integration.go
```diff
@@ -911,6 +911,11 @@ func (i *Test) MetricsURL() string {
 	return fmt.Sprintf("http://localhost:%d/metrics", i.AdminPort())
 }
 
+// AsyncTxSubOpenAPISpecURL returns the URL for getting the openAPI spec yaml for async-txsub endpoint.
+func (i *Test) AsyncTxSubOpenAPISpecURL() string {
+	return fmt.Sprintf("http://localhost:%d/transactions_async", i.AdminPort())
+}
+
```
tamirms, line 918: the functions added to this file are never used so no need to include them in the PR

clients/stellarcore/metrics_client.go
aditya1702: @tamirms Ah I added this because earlier there were a lot of other metrics too. So I wanted to abstract them within a struct. But not it does not make sense. I will remove it

services/horizon/internal/test/integration/integration.go
aditya1702: @tamirms It is used here - https://github.com/stellar/go/pull/5188/files#diff-3ffac184a24ab183dd63a08516cde2bb086d31ad0a24295e4d506eceb7ce4197R136.

tamirms (approved):
🎉

overcat:
Hi @aditya1702, according to the design intention, a user submitted a transaction through this interface, and it was successfully executed. If the user submits the same transaction again through this interface (which is completely identical to the previous one), what kind of exception will the user receive?

clients/horizonclient/internal.go
overcat, line 33: I think there is a problem with the implementation here. For AsyncSubmitTransactionXDR, if the user passes in an invalid XDR, the data returned by Horizon is as follows:

```
{
  "type": "https://stellar.org/horizon-errors/transaction_malformed",
  "title": "Transaction Malformed",
  "status": 400,
  "detail": "Horizon could not decode the transaction envelope in this request. A transaction should be an XDR TransactionEnvelope struct encoded using base64.  The envelope read from this request is echoed in the `extras.envelope_xdr` field of this response for your convenience.",
  "extras": {
    "envelope_xdr": "AAAAAgAAAAAoXi7Kz8/PAl9GZl5AD09H080g961ptb1MMfj6evUsJwAAAGQACuJqAAAAEQAAAAEAAAAAAAAAAAAAAABmoxT9AAAAAQAAAA9IZWxsbywgU3RlbGxhciEAAAAAAQAAAAAAAAABAAAAAPSbkKYju7E2GtKbpAnQ0twJCAGBwIcxea690WueeVSjAAAAAAAAAADQsJmHAAAAAAAAAAF69SwnAAAAQHgHTIiwgcIvdm0//yTH9o9rsGlMCng7YjKrOszNaJ0qwvPeLs8/bNdTRdg51gs3/RQAhT4lDeJR4Bs5wgZjeMgM=",
    "error": {}
  }
}
```
However, due to this change, the err returned by AsyncSubmitTransactionXDR will be nil, which does not meet expectations.

protocols/horizon/main.go
```diff
@@ -567,6 +576,26 @@ type InnerTransaction struct {
 	MaxFee     int64    `json:"max_fee,string"`
 }
 
+// AsyncTransactionSubmissionResponse represents the response returned by Horizon
+// when using the transaction-async endpoint.
+type AsyncTransactionSubmissionResponse struct {
+	// ErrorResultXDR is present only if Status is equal to proto.TXStatusError.
+	// ErrorResultXDR is a TransactionResult xdr string which contains details on why
+	// the transaction could not be accepted by stellar-core.
+	ErrorResultXDR string `json:"errorResultXdr,omitempty"`
```
overcat, line 585: Perhaps using `error_result_xdr` would be more consistent.

aditya1702: @overcat The `error_result_xdr` is returned from core when the transaction submission has issues. In the invalid XDR case, it never reaches core but fails as a validation check from Horizon.

JakeUrban: @aditya1702 I think he's pointing out the naming convention `errorResultXdr` should probably be `error_result_xdr`.

aditya1702: Ah my bad, misunderstood your comment. @overcat Yes that is a good catch and I will include this change for the next release. However, I do want to note that this will be a breaking change to the API

overcat: Hi @aditya1702, 

Yes, this will be a breaking update. Some measures need to be considered before making the changes in order to ensure a smooth transition downstream.

Did you check my [other comment](https://github.com/stellar/go/pull/5188#discussion_r1692439324)? That's a bug. I think we need to fix it. Below is the code to reproduce it.

```go
package main

import client "github.com/stellar/go/clients/horizonclient"

func main() {
	client := client.Client{
		HorizonURL: "https://horizon-testnet.stellar.org",
	}
	_, err := client.AsyncSubmitTransactionXDR("invalidXdrAAAAAgAAAAoXi7Kz8/PAl9GZl5AD09H080g961tb1MMfj6evUsJwAAAGQACuJqAAAAEgAAAAEAAAAAAAAAAAAAAABmo4DLAAAAAQAAAA9IZWxsbywgU3RlbGxhciEAAAAAAQAAAAAAAAABAAAAAPSbkKYju7E2GtKbpAnQ0twJCAGBwIcxea690WueeVSjAAAAAAAAAADQsJmHAAAAAAAAAAF69SwnAAAAQFAaIAq+4YAPnY7BEukwtmtiXVvHvjPi+6kJUpl/ljoYSmnZcMIUSFHH7LGct0ftWl7PnUKGE4JGDFk38tg8YAs=")
	if err != nil {
		println("error is not empty")
		panic(err)
	}
	println("We shouldn't have arrived here.")
}

```

PR #5191, opened 2024-02-02
Fix for transaction submission timeout

### LINKED ISSUE #5192: services/horizon/txsub: submitting transaction with future sequence may result in 60s timeout

2.28.0

### What did you do?

submit a tx with a sequence number that is gapped farther out than current source account seq+1.

```
curl --location 'https://<horizon_2.28.0_on_testnet>/transactions' \
--header 'Content-Type: application/x-www-form-urlencoded' \
--data-urlencode 'tx=AAAAAgAAAACI3DQX1QWOxLRQPgwS6hoKib4gD+mJIkI9QzQBT6aw7gAAJxAAAYJ5AAABZAAAAAAAAAABAAAAHEJXUnhramQ5V3VkdXVrZXdrakN1dVVjQlk0YWgAAAABAAAAAQAAAACI3DQX1QWOxLRQPgwS6hoKib4gD+mJIkI9QzQBT6aw7gAAAAEAAAAAO5kROA7+mIugqJAOsc/kTzZvfb6Ua+0HckD39iTfFcUAAAAAAAAAAAAAA/IAAAAAAAAAAU+msO4AAABANwm//iyoc4FjpAp4awoQLQmakCKF4HNiNZp02fm+cHZtJ1QOKR6ZjZGPUipUj/AkyhYDtqOBMNfXvR49YExTCg=='

```

tx_bad_seq error returned

### What did you see instead?
60 second timeout from server

aditya1702:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Adds a check in the transaction submission workflow, where it returns the bad sequence error immediately if Horizon and Core sync up with each other.

### Why

During a particular edge-case, if core reports bad-seq error, and Horizon and Core are synced up with each other, the txsub endpoint will still timeout instead of returning the error immediately.

closes #5192 

This is a bug due to how Horizon will keep waiting for the account sequence to catch up to the submitted tx seq.

### Known limitations

NA

### DIFF
--- a/services/horizon/internal/ledger/main.go
+++ b/services/horizon/internal/ledger/main.go
@@ -1,3 +1,8 @@
+// Package ledger provides useful utilities concerning ledgers within stellar,
+// specifically as a central location to store a cached snapshot of the state of
+// both horizon's and stellar-core's views of the ledger.  This package is
+// intended to be at the lowest levels of horizon's dependency tree, please keep
+// it free of dependencies to other horizon packages.
 package ledger
 
 import (
--- a/services/horizon/internal/ledger/main.go
+++ b/services/horizon/internal/ledger/main.go
@@ -1,15 +1,9 @@
-// Package ledger provides useful utilities concerning ledgers within stellar,
-// specifically as a central location to store a cached snapshot of the state of
-// both horizon's and stellar-core's views of the ledger.  This package is
-// intended to be at the lowest levels of horizon's dependency tree, please keep
-// it free of dependencies to other horizon packages.
 package ledger
 
 import (
+	"github.com/prometheus/client_golang/prometheus"
 	"sync"
 	"time"
-
-	"github.com/prometheus/client_golang/prometheus"
 )
 
 // Status represents a snapshot of both horizon's and stellar-core's view of the
@@ -31,7 +25,7 @@ type HorizonStatus struct {
 }
 
 // State is an in-memory data structure which holds a snapshot of both
-// horizon's and stellar-core's view of the the network
+// horizon's and stellar-core's view of the network
 type State struct {
 	sync.RWMutex
 	current Status
@@ -42,6 +36,14 @@ type State struct {
 		HistoryElderLedgerCounter         prometheus.CounterFunc
 		CoreLatestLedgerCounter           prometheus.CounterFunc
 	}
+}
+
+type StateInterface interface {
+	CurrentStatus() Status
+	SetStatus(next Status)
+	SetCoreStatus(next CoreStatus)
+	SetHorizonStatus(next HorizonStatus)
+	RegisterMetrics(registry *prometheus.Registry)
 }
 
 // CurrentStatus returns the cached snapshot of ledger state
--- a/services/horizon/internal/txsub/system.go
+++ b/services/horizon/internal/txsub/system.go
@@ -41,7 +41,7 @@ type System struct {
 	Submitter         Submitter
 	SubmissionTimeout time.Duration
 	Log               *log.Entry
-	LedgerState       *ledger.State
+	LedgerState       ledger.StateInterface
 
 	Metrics struct {
 		// SubmissionDuration exposes timing metrics about the rate and latency of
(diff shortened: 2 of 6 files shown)


services/horizon/internal/txsub/system.go
```diff
@@ -142,6 +144,14 @@ func (sys *System) Submit(
 			return
 		}
 
+		// Check if Horizon and Core have synced up: If yes, then no need to wait for account sequence
+		// and send txBAD_SEQ right away.
+		currentStatus := sys.LedgerState.CurrentStatus()
+		if currentStatus.CoreLatest == currentStatus.HistoryLatest {
+			sys.finish(ctx, hash, resultCh, Result{Err: sr.Err})
+			return
+		}
```
mollykarcher, line 153: Defer to @Shaptic or @sreuland on if this is the best way to compare core/ingest states. But we need to move this logic _inside_ `waitUntilAccountSequence`. Will it as it is here, imagine this sequence of events:
- state: account is at sequence number 5 and horizon ingestion is lagging behind core
- a user submits a tx with sequence number 7
- we execute this check on line 150, and it continues on because horizon is lagging behind
- we enter `waitUntilAccountSequence`
- horizon ingestion catches up
- `waitUntilAccountSequence` continues to loop waiting for the account sequence number to go to 6, which will never happen

What we need is for `waitUntilAccountSequence` to exit/terminate when ingestion is caught up.

aditya1702: @mollykarcher Yeah that makes sense. Did not think of this edge-case

sreuland: yea, good point, repeating this same [status check](https://github.com/stellar/go/pull/5191/files#diff-5846d16bb162a4a6ca1e880ef01f59b654f19302c0ef9e90f3d7fab959ceb5b1R149-R153) from the polling iteration in `waitUntilAccountSequence` instead of here gives it a bit more resiliency, same timeliness of up front check as realized here on 1st iteration of polling, and additional opportunity to still exit early if the same condition happens on ensuing poll iterations any time up to the context time out.

Shaptic: Yep, agreed with this.

And I _think_ this seems like a good way to check state (from the code, those statuses seem like they're updated on each app tick), but you could `<=` to be safer.

services/horizon/internal/txsub/system_test.go
```diff
@@ -281,6 +315,64 @@ func (suite *SystemTestSuite) TestSubmit_BadSeqNotFound() {
...
+		suite.successXDR,
+		suite.successTx.Transaction.TransactionHash,
+	)
+
+	assert.NotNil(suite.T(), r.Err)
+	assert.True(suite.T(), suite.submitter.WasSubmittedTo)
+}
+
+// If error is bad_seq and horizon and core are in sync, then return error
+func (suite *SystemTestSuite) TestSubmit_BadSeqNotFoundWhenInSync() {
```
sreuland, line 345: should this test case read `TestSubmit_BadSeqReturnedWhenInSync()` , maybe that can be demonstrated further with additional assert on `r.Err = BadSeq` after [line 387](https://github.com/stellar/go/pull/5191/files#diff-28500168bb477ec1663bf8c8c364b8f87c1886bdafd087f557c04a41af3bd398R387)?

sreuland (approved):
lgtm, nice work on pulling up the StateInterface to enable a unit test with mocking!

services/horizon/internal/txsub/system_test.go
aditya1702: Hmm, yeah that would work. Will change that.

Shaptic (approved):

services/horizon/internal/ledger/main.go
```diff
@@ -1,15 +1,9 @@
-// Package ledger provides useful utilities concerning ledgers within stellar,
-// specifically as a central location to store a cached snapshot of the state of
-// both horizon's and stellar-core's views of the ledger.  This package is
-// intended to be at the lowest levels of horizon's dependency tree, please keep
-// it free of dependencies to other horizon packages.
```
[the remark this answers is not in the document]
aditya1702: Oops did it by mistake. Good catch 😅

PR #5197, opened 2024-02-05
services/horizon: Add cache toggle and use libary for on-disk caching

Shaptic:
### What
* Introduce a library to do on-disk caching ([`fscache`](https://github.com/djherbis/fscache)) to avoid the issue we were seeing with the custom implementation.
* Changes the cache to be size-based rather than count-based (we ensure it will always be <= 10GiB this way).
 * A new `--history-archive-caching=(true|false)` flag (default: true) controls whether or not there will be an on-disk cache used for history archive fetches. Most people will not need to set this to `false` unless they're encountering issues. This is to prevent folks from being blocked if issues are encountered with this feature in the future.
 * Add a metric for showing how much bandwidth is saved by caching.

### Why
Battle-tested reliability.

Closes https://github.com/stellar/quickstart/issues/557.


historyarchive/archive.go
```diff
@@ -396,18 +398,45 @@ func (a *Archive) GetXdrStream(pth string) (*XdrStream, error) {
 
 func (a *Archive) cachedGet(pth string) (io.ReadCloser, error) {
 	if a.cache != nil {
```
sreuland, line 400: for consideration, guard check up front to remove the simple path:

```
if a.cache == nil {
   a.stats.incrementDownloads()
   return a.backend.GetFile(pth)
}
```

historyarchive/archive.go
```diff
@@ -396,18 +398,45 @@ func (a *Archive) GetXdrStream(pth string) (*XdrStream, error) {
...
+				return teeReadCloser(
+					upstreamReader,
+					wrtr,
+					func() error { return nil }, // no add'l work to do
+				), nil
+			}
+
+			// To track metrics, we make a best-effort to see how much bandwidth
+			// we've saved by caching.
+			size, err := a.backend.Size(pth)
```
sreuland, line 432: this will result in blocking http HEAD request to server on every file request even when local cache used, gives up some caching gains, can a `sync.Map` be used as a cached `pth->size` lookup instead?

historyarchive/xdrstream.go
```diff
@@ -107,7 +107,7 @@ func (x *XdrStream) ExpectedHash() ([sha256.Size]byte, bool) {
 func (x *XdrStream) Close() error {
 	if x.validateHash {
 		// Read all remaining data from rdr
-		_, err := io.Copy(ioutil.Discard, x.rdr)
+		_, err := io.Copy(io.Discard, x.rdr)
```
sreuland, line 110: oh, interesting, was this and [ln 144](https://github.com/stellar/go/pull/5197/files#diff-0ea9abf4b94ff3c2d78357c9156e6f754f1d062fb95c342d72732ae750bc4edfR144) related to root cause of corruption seen in [quickstart/557](https://github.com/stellar/quickstart/issues/557)?

historyarchive/archive.go
```diff
@@ -396,18 +398,45 @@ func (a *Archive) GetXdrStream(pth string) (*XdrStream, error) {
 
 func (a *Archive) cachedGet(pth string) (io.ReadCloser, error) {
 	if a.cache != nil {
-		rdr, foundInCache, err := a.cache.GetFile(pth, a.backend)
-		if !foundInCache {
-			a.stats.incrementDownloads()
+		rdr, wrtr, err := a.cache.Get(pth)
+		if err != nil {
+			a.cache.Remove(pth) // just in case
+			log.WithError(err).Warnf("On-disk cache retrieval failed")
```
tamirms, line 404: it is worth including the error from `a.cache.Remove(pth)` in case it's non-nil in the log message

historyarchive/archive.go
```diff
@@ -396,18 +398,45 @@ func (a *Archive) GetXdrStream(pth string) (*XdrStream, error) {
...
+			// you're expected to write your upstream as well as a reader that
+			// will read directly from it.
+			if wrtr != nil {
+				rdr.Close() // we make our own
+				a.stats.incrementDownloads()
+
+				upstreamReader, err := a.backend.GetFile(pth)
+				if err != nil {
+					wrtr.Close()
+					a.cache.Remove(pth)
```
tamirms, line 416: I think it's worth logging errors from a.cache.Remove(pth) and from the Close() invocations

Shaptic: yep, added lots of logging!

historyarchive/archive.go
Shaptic: I checked the lib's code and this shouldn't be needed anyway

historyarchive/xdrstream.go
Shaptic: nah this was just a drive-by fix of deprecated usage - they're functionally equivalent

historyarchive/archive.go
Shaptic: good call!

historyarchive/archive.go
```diff
@@ -476,13 +505,25 @@ func Connect(u string, opts ConnectOptions) (*Archive, error) {
...
 
-	if opts.CacheConfig.Cache {
-		cache, innerErr := MakeArchiveBucketCache(opts.CacheConfig)
-		if innerErr != nil {
-			return &arch, innerErr
+	if opts.CachePath != "" {
+		haunter := fscache.NewLRUHaunterStrategy(
+			fscache.NewLRUHaunter(0, 10<<30 /* ~10GiB */, time.Minute),
+		)
+		fs, err := fscache.NewFs(opts.CachePath, 0755 /* drwxr-xr-x */)
```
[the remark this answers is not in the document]
Shaptic: it is now :smile:

historyarchive/archive.go
Shaptic: yep, nice idea

historyarchive/archive.go
```diff
@@ -395,27 +405,65 @@ func (a *Archive) GetXdrStream(pth string) (*XdrStream, error) {
...
-	if a.cache != nil {
-		rdr, foundInCache, err := a.cache.GetFile(pth, a.backend)
-		if !foundInCache {
-			a.stats.incrementDownloads()
-		} else {
-			a.stats.incrementCacheHits()
-		}
-		if err == nil {
-			return rdr, nil
+	cache := a.cache.cache
```
tamirms, line 408: `a.cache.cache` is a bit confusing. I think it would be better to get rid of the `cacheState` struct and replace it with:

```golang
type archiveBucketCache struct {
    fscache.Cache
    path  string
    sizes sync.Map
}
```
Note that `fscache.Cache` is embedded in the `archiveBucketCache` struct.
Then you can include a `cache   *archiveBucketCache` field in the `Archive` struct`.

Shaptic: ahh nice, very clean :+1: done in e92029cb

historyarchive/archive.go
```diff
@@ -395,23 +406,62 @@ func (a *Archive) GetXdrStream(pth string) (*XdrStream, error) {
...
+	// If a NEW key is being retrieved, it returns a writer to which
+	// you're expected to write your upstream as well as a reader that
+	// will read directly from it.
+	if wrtr != nil {
+		log.WithField("path", pth).Info("Caching file...")
+		a.stats.incrementDownloads()
+		upstreamReader, err := a.backend.GetFile(pth)
+		if err != nil {
+			L.WithError(err).Warn("Download failed, purging from cache")
+			wrtr.Close()
```
tamirms, line 432: does `rdr` also need to be closed?

Shaptic: it will be closed by the caller once reading fails, right?

historyarchive/archive.go
```diff
@@ -395,23 +406,62 @@ func (a *Archive) GetXdrStream(pth string) (*XdrStream, error) {
...
+			if err != nil {
+				L.WithError(err).Warn("Failed to download and cache file")
+				if removalErr := a.cache.Remove(pth); removalErr != nil {
+					L.WithError(removalErr).Warn("Removing cached file failed")
+				}
+			} else {
+				L.Infof("Cached %dKiB file", written/1024)
+
+				// Track how much bandwidth we've saved from caching.
+				a.cache.sizes.Store(pth, written)
```
tamirms, line 453: do we also want a metric for how much bandwidth was spent downloading?

historyarchive/archive.go
tamirms: `rdr` is never returned in this code path. On line 434 we return `nil` instead of `rdr` because this is an error case

Shaptic: oh duh :facepalm: so we can close `rdr` or we can return it to bubble up a bad reader. thoughts?

historyarchive/archive.go
Shaptic: we could, sure, but we can see that through cloudflare in a way heh. wouldn't be a big deal to add, though

historyarchive/archive.go
tamirms: I think we can close `rdr` and still return nil because we're in the error case. There's no need to return both a reader and an error

Shaptic: yep true. fixed up!

historyarchive/xdrstream.go
sreuland: ok, how do we know this fixed https://github.com/stellar/quickstart/issues/557, build quickstart locally with this branch for horizon and run with `--pubnet` ?

historyarchive/archive.go
tamirms: > we can see that through cloudflare in a way

good point. Given that we already have that data in cloudflare we can defer collecting bandwidth metrics on dwonloads until later

historyarchive/archive.go
```diff
@@ -395,23 +406,66 @@ func (a *Archive) GetXdrStream(pth string) (*XdrStream, error) {
...
+	// If a NEW key is being retrieved, it returns a writer to which
+	// you're expected to write your upstream as well as a reader that
+	// will read directly from it.
+	if wrtr != nil {
+		log.WithField("path", pth).Info("Caching file...")
+		a.stats.incrementDownloads()
+		upstreamReader, err := a.backend.GetFile(pth)
+		if err != nil {
+			L.WithError(err).Warn("Download failed, purging from cache")
+			L.Debugf("Writer: %v, reader: %v; cache removal: %v",
```
tamirms, line 432: can we combine these two log statements into one?

historyarchive/xdrstream.go
Shaptic: yep, that's how I've been doing it this whole time locally and I can confirm that it no longer encounters the read error when processing ledgers

tamirms (approved):

sreuland (approved):
nice work!

historyarchive/archive.go
```diff
@@ -395,23 +406,71 @@ func (a *Archive) GetXdrStream(pth string) (*XdrStream, error) {
...
+	// will read directly from it.
+	if wrtr != nil {
+		log.WithField("path", pth).Info("Caching file...")
+		a.stats.incrementDownloads()
+		upstreamReader, err := a.backend.GetFile(pth)
+		if err != nil {
+			L.WithError(err).WithFields(log.Fields{
+				"write-close": wrtr.Close(),
+				"read-close":  rdr.Close(),
+				"cache-rm":    a.cache.Remove(pth),
```
tamirms, line 436: I think the evaluation order of the entries in `log.Fields` is not guaranteed according to https://go101.org/article/evaluation-orders.html

so it may be possible that the `"cache-rm":    a.cache.Remove(pth)` entry is evaluated before `"write-close": wrtr.Close()` and `"read-close":  rdr.Close()` and that could result in a deadlock

Shaptic: I was thinking about that theoretical possibility but didn't look up the details. Great call out! Fixed in ee02995c.

PR #5205, opened 2024-02-13
[Draft] Merge horizon-v2.28.2 back into master

Shaptic:
### What
Self-explanatory.

### Why
Self-explanatory.

This is a very precarious merge (lots of conflicts were resolved) so I'd appreciate multiple eyes on it for sanity checking.

PR #5206, opened 2024-02-13
Merge upstream `master` into `horizon-v2.28.2` to facilitate the back-merge

Shaptic:
### What
I took the following steps:

```bash
git switch master
git pull upstream master   # sync to stellar/go
git checkout horizon-v2.28.2
git switch -C v2.28.2-backmerge
git merge -S master
# <resolve all conflicts>
git commit -S
git push origin v2.28.2-backmerge
```

The rationale is self-explanatory: we need the patch release changes back in master.

### Why
Self-explanatory. Replaces https://github.com/stellar/go/pull/5205.

This was a **very** precarious merge (lots of conflicts were resolved) so I'd appreciate multiple eyes on it for sanity checking.


historyarchive/archive.go
```diff
@@ -119,8 +131,9 @@ func (arch *Archive) GetCheckpointManager() CheckpointManager {
 func (a *Archive) GetPathHAS(path string) (HistoryArchiveState, error) {
 	var has HistoryArchiveState
 	rdr, err := a.backend.GetFile(path)
-	// this is a query on the HA server state, not a data/bucket file download
-	a.stats.incrementRequests()
+	a.stats.incrementDownloads()
+	// 	// this is a query on the HA server state, not a data/bucket file download
+	// 	a.stats.incrementRequests()
```
Shaptic, line 136: this is a diff I'm a little worried about because I'm not sure which is correct. @sreuland you're on the `git blame` for this so can you help give some insight?

services/horizon/internal/db2/history/claimable_balances.go
```diff
@@ -67,17 +67,17 @@ func applyClaimableBalancesQueriesCursor(sql sq.SelectBuilder, lCursor int64, rC
...
+		sql = sql.OrderBy("cb.last_modified_ledger asc, cb.id asc")
 	case db2.OrderDescending:
 		if hasPagedLimit {
 			sql = sql.
-				Where(sq.Expr("(last_modified_ledger, id) < (?, ?)", lCursor, rCursor))
+				Where(sq.Expr("(cb.last_modified_ledger, cb.id) < (?, ?)", lCursor, rCursor))
 		}
 
-		sql = sql.OrderBy("last_modified_ledger desc, id desc")
+		sql = sql.OrderBy("cb.last_modified_ledger desc, cb.id desc")
```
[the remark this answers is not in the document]
urvisavla: yes, looks good.

historyarchive/archive.go
sreuland: ugh, this looks like a bad merge conflict resolution, that's not [what 2.28.2 looks like in this area](https://github.com/stellar/go/blob/release-horizon-v2.28.2/historyarchive/archive.go#L146).

hmm, when i ported [2.28.0](https://github.com/stellar/go/pull/5183) and [2.28.1](https://github.com/stellar/go/pull/5195) branches back to master, I used interactive rebase instead of merge-master-to-rel-then-merge-rel-back-to-master and found that resulted in less conflicts and easier to understand the remaining ones which did arise from archive pool stuff of light horizon going into master but not not on 2.28 yet

it may be worth re-trying this merge of 2.28.2 to master with a rebase instead just to see - try create new branch off release branch, from that branch `git rebase -i master` choose just the [new commits on the 2.28.2 branch](https://github.com/stellar/go/compare/release-horizon-v2.28.1...release-horizon-v2.28.2), and see how conflicts look from there, should be a lot less, and should only see the commits relevant to the 2.28.2 show up in git compare to master, whereas this pr is showing a lot of commits

Shaptic: It's not a bad resolution because I did it on purpose: you can see that [on master](https://github.com/stellar/go/blob/master/historyarchive/archive.go#L122-L123) it has the commented-out lines! I left them both in specifically because I wanted to ensure that the behavior preserved here (i.e. incrementing downloads) is actually the correct one! So can you confirm that? :smile: 

The merge actually went well otherwise! The conflicts were almost exclusively related to my code in #5197 so I had a good handle on how to resolve everything. 

I think the reason why there are so many commits showing here is exactly because you did an interactive rebase, so it still thinks that those commits from .0 and .1 are different than the ones you merged into master (since history got modified). I'll try it via the rebase approach to see if it's cleaner but I also think it's fine as-is since the conflicts got worked out.

Shaptic: Actually while doing the `git rebase -i master` off of `horizon-v2.28.2` I actually get the exact same conflicts lol so there's really no benefit besides a mildly cleaner commit log (at the expense of me having to carefully resolve the conflicts twice :laughing:).

PR #5208, opened 2024-02-14
/services/horizon/ingest: add SKIP_TXMETA

### LINKED ISSUE #5189: Elide tx-meta from transaction data model in db and api.
A new Horizon configuration setting, DISABLE_TXMETA, which defaults to FALSE. When it's TRUE, horizon will no longer store serialized tx-meta xdr in the history database or show it on the transaction API model.

history_transactions table should have serialized base64 for an empty xdr.TransactionMetaV3{} in `tx_meta`, and for the API resource model emitted for Transaction for this row needs to show the same empty value representation for
`result_meta_xdr`

revert the [recently added DISABLE_SOROBAN_INGEST](https://github.com/stellar/go/pull/5176), as it removed soroban related parsing from tx-meta for effects and asset balances, however, in hindsight those were not necessary and do not incur a performance overhead.

sreuland:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Added new `SKIP_TXMETA`. Defaults to `FALSE`. When `TRUE` the transaction model for classic or soroban tx's won't have tx meta populated anymore, on database, history_transactions.tx_meta column will have serialized xdr that equates to empty for any protocol version, such as for protocol 20,`xdr.TransactionMeta.V3`, `Operations`, `TxChangesAfter`, `TxChangesBefore` will be empty arrays and `SorobanMeta` will be nil and API responses with transaction model will have same empty/nil.

Removed `DISABLE_SOROBAN_INGEST` configuration parameter and it's cohort by reverting https://github.com/stellar/go/pull/5176, it did similar tx-meta removal but was limited to just soroban specific,  `SKIP_TXMETA` parameter instead and removed parsing effects/ops aspects derived from tx-meta, which wasn't needed.

### Why

to prevent extended storage space needs on db.
Closes #5189

### DIFF
--- a/services/horizon/internal/ingest/processors/transactions_processor.go
+++ b/services/horizon/internal/ingest/processors/transactions_processor.go
@@ -11,40 +11,23 @@ import (
 )
 
 type TransactionProcessor struct {
-	batch       history.TransactionBatchInsertBuilder
-	skipSoroban bool
+	batch history.TransactionBatchInsertBuilder
 }
 
-func NewTransactionFilteredTmpProcessor(batch history.TransactionBatchInsertBuilder, skipSoroban bool) *TransactionProcessor {
+func NewTransactionFilteredTmpProcessor(batch history.TransactionBatchInsertBuilder) *TransactionProcessor {
 	return &TransactionProcessor{
-		batch:       batch,
-		skipSoroban: skipSoroban,
+		batch: batch,
 	}
 }
 
-func NewTransactionProcessor(batch history.TransactionBatchInsertBuilder, skipSoroban bool) *TransactionProcessor {
+func NewTransactionProcessor(batch history.TransactionBatchInsertBuilder) *TransactionProcessor {
 	return &TransactionProcessor{
-		batch:       batch,
-		skipSoroban: skipSoroban,
+		batch: batch,
 	}
 }
 
 func (p *TransactionProcessor) ProcessTransaction(lcm xdr.LedgerCloseMeta, transaction ingest.LedgerTransaction) error {
-	elidedTransaction := transaction
-
-	if p.skipSoroban &&
-		elidedTransaction.UnsafeMeta.V == 3 &&
-		elidedTransaction.UnsafeMeta.MustV3().SorobanMeta != nil {
-		elidedTransaction.UnsafeMeta.V3 = &xdr.TransactionMetaV3{
-			Ext:             xdr.ExtensionPoint{},
-			TxChangesBefore: xdr.LedgerEntryChanges{},
-			Operations:      []xdr.OperationMeta{},
-			TxChangesAfter:  xdr.LedgerEntryChanges{},
-			SorobanMeta:     nil,
-		}
-	}
-
-	if err := p.batch.Add(elidedTransaction, lcm.LedgerSequence()); err != nil {
+	if err := p.batch.Add(transaction, lcm.LedgerSequence()); err != nil {
 		return errors.Wrap(err, "Error batch inserting transaction rows")
 	}
 
--- a/services/horizon/internal/ingest/processors/transactions_processor.go
+++ b/services/horizon/internal/ingest/processors/transactions_processor.go
@@ -11,23 +11,52 @@ import (
 )
 
 type TransactionProcessor struct {
-	batch history.TransactionBatchInsertBuilder
+	batch      history.TransactionBatchInsertBuilder
+	skipTxmeta bool
 }
 
-func NewTransactionFilteredTmpProcessor(batch history.TransactionBatchInsertBuilder) *TransactionProcessor {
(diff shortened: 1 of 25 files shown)


services/horizon/internal/ingest/processors/transactions_processor.go
```diff
@@ -11,23 +11,52 @@ import (
...
+				TxChangesAfter:  xdr.LedgerEntryChanges{},
+				SorobanMeta:     nil,
+			}
+		case 2:
+			elidedTransaction.UnsafeMeta.V2 = &xdr.TransactionMetaV2{
+				TxChangesBefore: xdr.LedgerEntryChanges{},
+				Operations:      []xdr.OperationMeta{},
+				TxChangesAfter:  xdr.LedgerEntryChanges{},
+			}
+		case 1:
```
tamirms, line 51: I think we should a default case which returns an error

services/horizon/internal/integration/transaction_test.go
```diff
@@ -0,0 +1,133 @@
...
+	clientTx := itest.MustSubmitOperations(&masterAccount, itest.Master(), op)
+
+	var txMetaResult xdr.TransactionMeta
+	err = xdr.SafeUnmarshalBase64(clientTx.ResultMetaXdr, &txMetaResult)
+	require.NoError(t, err)
+
+	assert.Greater(t, len(txMetaResult.MustV2().Operations), 0)
+	assert.Greater(t, len(txMetaResult.MustV2().TxChangesBefore), 0)
+	// TODO, generate TxChangesAfter also
+	//assert.Greater(t, len(txMetaResult.MustV2().TxChangesAfter), 0)
```
tamirms, line 42: TODO

services/horizon/internal/integration/transaction_test.go
```diff
@@ -0,0 +1,133 @@
...
+	op := &txnbuild.Payment{
+		SourceAccount: itest.Master().Address(),
+		Destination:   itest.Master().Address(),
+		Asset:         txnbuild.NativeAsset{},
+		Amount:        "10",
+	}
+
+	clientTx := itest.MustSubmitOperations(&masterAccount, itest.Master(), op)
+
+	var txMetaResult xdr.TransactionMeta
```
tamirms, line 35: it doesn't look like we check anything in `txMetaResult`. Should we be asserting something about the value / shape of `txMetaResult`? this comment also applies to the other tests in this file

services/horizon/internal/integration/transaction_test.go
```diff
@@ -0,0 +1,133 @@
...
+	err = xdr.SafeUnmarshalBase64(clientTx.ResultMetaXdr, &txMetaResult)
+	require.NoError(t, err)
+
+	assert.Greater(t, len(txMetaResult.MustV2().Operations), 0)
+	assert.Greater(t, len(txMetaResult.MustV2().TxChangesBefore), 0)
+	// TODO, generate TxChangesAfter also
+	//assert.Greater(t, len(txMetaResult.MustV2().TxChangesAfter), 0)
+}
+func TestP19MetaDisabledTransaction(t *testing.T) {
```
tamirms, line 44: nit: there should be a new line separateing these two functions

services/horizon/internal/integration/transaction_test.go
sreuland: do you know of any operations that would trigger TxChangesAfter, i tried several and FeeBumps, but only generated TxChangesBefore.

services/horizon/internal/integration/transaction_test.go
sreuland: values in txMetaResult are asserted [a few lines lower ](https://github.com/stellar/go/pull/5208/files#diff-f3057c0125a179b20f65f27fd71674ae9fa76f9947afe42ac03a703c86d06f67R39-R40) , is that what you were looking for?

services/horizon/internal/ingest/processors/transactions_processor.go
sreuland: yes, I considered that iniitally. but didn't follow through, good callout, done - https://github.com/stellar/go/pull/5208/commits/6264a8c5897e2a302736473ddd2817e98a670d8a

services/horizon/internal/integration/transaction_test.go
tamirms: oops, sorry I misread the code. please ignore this comment

services/horizon/internal/integration/transaction_test.go
tamirms: I think soroban refunds can appear in TxChangesAfter:

https://github.com/stellar/stellar-protocol/blob/8571ed1e9f9a5f26e92d2469316d7997d61ec093/core/cap-0046-07.md?plain=1#L242

but if this case is difficult to produce, I think you can just delete the commented out code

tamirms (approved):

PR #5214, opened 2024-02-19
clients/horizonclient: allow sending user-defined headers on requests

RedCuckoo:
### What

Add Headers field to the Client struct in order to allow passing custom HTTP headers with requests made by the SDK.

### Why

This change enables users of SDK to include custom headers, such as authorization headers, when making HTTP requests to Stellar API. Previously, there was no built-in mechanism to add custom headers, which limited the flexibility of the SDK in certain scenarios. By introducing the Headers field, users can now easily include any additional headers required for their specific use cases.

### Known limitations

N/A


mollykarcher (commented):
Thanks for the contribution!

RedCuckoo:
@mollykarcher Thank you for the comments, I have applied requested changes and wrote unit tests for this specific case.

Shaptic (approved):

PR #5216, opened 2024-02-21
services/horizon/internal/ingest: Fix transaction processor metrics

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

When horizon v2.27.0 was deployed to production the ingestion duration of the transaction processor appeared to decrease dramatically:

image

Unfortunately, this is not due to any performance improvements but rather a bug in our reporting of metrics to prometheus.

In the 2.27.0 release ingestion filtering was [enabled by default](https://github.com/stellar/go/commit/f7065378eb39c33cfab69d28095d0f7ae3f68c9d).  When filtering is enabled, Horizon will run an [additional transaction processor](https://github.com/stellar/go/blob/horizon-v2.27.0/services/horizon/internal/ingest/processor_runner.go#L169-L172) on the `history_transactions_filtered_tmp` table. The [transaction processor](https://github.com/stellar/go/blob/horizon-v2.27.0/services/horizon/internal/ingest/processors/transactions_processor.go#L17-L23) for the `history_transactions_filtered_tmp` table has the same type is the regular transaction processor. The type of a transaction is used as the [label](https://github.com/stellar/go/blob/horizon-v2.27.0/services/horizon/internal/ingest/group_processors.go#L71) for the prometheus metric capturing ingestion run duration. So, the metric for the `history_transactions_filtered_tmp` processor was getting reported instead of the normal transactions processor because they both had the [same label](https://github.com/stellar/go/blob/horizon-v2.27.0/services/horizon/internal/ingest/processor_runner.go#L364-L370). 

This PR aims to fix this bug and to prevent bugs like this from occurring again by ensuring that all ingestion processor metrics have unique label names.
 
### Known limitations

[N/A]


services/horizon/internal/ingest/processor_runner.go
```diff
@@ -413,13 +436,52 @@ func (s *ProcessorRunner) RunTransactionProcessorsOnLedger(ledger xdr.LedgerClos
 	return
 }
 
+func checkUniqueTransactionProcessorNames(
```
sreuland, line 439: this has un-necessary side effects, it mutates the input Set 's', and appears that none of callers refer to the set after calling this function, so, is probably not necessary?

taking a step back, wondering if there may be a different design pattern to consider to weave in the same processor/filterer name uniqueness but encapsulate it in `Name()` and avoid it bubbling up to the processor/filter, with a singleton factory function where `Name()` calls the factory to build the name and throws err if not-unique, the singleton maintains an in-mem set for lookup of names to enforce uniquness.

the name factory
```
var registeredNames = set.Set[string]{},
func ProcessorNameRegister(name string) (string, error) {
     if registeredNames.Contains(name) {
	return fmt.Errorf("%s is duplicated", name)
    }
    registeredNames.Add(name)
    // decorate the name further ..
    return name
}

```

filter and processor use the factory
```
func (xyz *ProcessorOrFilter) Name() (string, error) {
	return ProcessorNameRegister("groupTransactionFilterers")
}
```

and when 'group' processors iterate their nested processors to capture run duration measurements, they use `Name()`
```
...
groupProcessor.AddRunDuration(nestedProcessor.Name()), startTime)
```

tamirms: I don't think this approach would work because it assumes that `Name()` will be invoked only once (on repeat invocations you will get trigger the error path in `ProcessorNameRegister()`). However, `Name()` can be called multiple times (for example we sume the duration of `ProcessTransaction()` and `Flush()`).

sreuland: yes, good point, that was half-baked, for further consideration, could use memoization in each processor/filter`Name()` to allow thread-safe re-entrancy, like:

```
var xyzRegisteredName atomic.Pointer[string]

func (xyz *ProcessorOrFilter) Name() (string, error) {
        var err error
	var myName string
        if myName = *xyzRegisteredName.Load(); myName == "" {
		if myName, err = ProcessorNameRegister("xyzProcessor"); err != nil {
			return "", err
		}
		xyzRegisteredName.Store(&myName)
	}
	return myName, nil
}
```

tamirms: I tried pursuing this approach but it resulted in a lot of code duplication as this function and the `xyzRegisteredName` global variable needs to be present in every ingestion processor, filterer, and loader. So I prefer keeping the code as is

sreuland: agreed, and the approach you took with `nameRegistry` scoped per ledger run looks good, approved. 

... there is another approach which could encapsulate once in the register function by caller passing its identity, captured in this example as `requestingSource` :
```
var registryMutex sync.RWMutex
var registeredNames = map[string]string{}

func ProcessorNameRegister(requestingSource string, name string) (string, error) {
	registryMutex.Lock()
	defer registryMutex.Unlock()
        if registeredSource, found :=registeredNames[name]; found {
		if registeredSource != requestingSource {
	        return "", fmt.Errorf("%s has already been registered by another source %v", name, registeredSource)
		}
		return name, nil	
        }

       registeredNames[name] = requestingSource
       return name, nil
}
```

and then `Name()` can go back to a thin wrapper:
```
func (xyz *ProcessorOrFilter) Name() (string, error) {
	return ProcessorNameRegister(reflect.TypeOf(xyz).Elem().Name(), "xyzProcessor")
}
```

defer to your choice, thanks for taking time to consider the suggestions!

PR #5223, opened 2024-02-26
support/db: Delay canceling queries from client side when there's a statement / transaction timeout configured in postgres

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add a `--cancel-db-query-timeout` flag which configures a timeout on all db queries invoked by a horizon HTTP request. This timeout is a postgres client side timeout which means that it is triggered by horizon establishing a new connection to postgres in order to cancel the ongoing query.

### Why

Previously, we configured both postgres client side timeouts and server side timeouts (in the form of the [statement_timeout](https://www.postgresql.org/docs/current/runtime-config-client.html) postgres setting). However, the client side timeout is redundant when the postgres server will enforce its own statement timeout. In fact, the client side timeout can result in additional load which will exacerbate situations when the postgres DB cannot keep up with the ongoing query traffic.

In our horizon deployments, we plan to configure the client side db timeout to be longer than the server side db timeout. We expect the server side timeout to terminate a majority of the long running queries and the client side timeout will be a fallback in the unexpected scenario where the db query has surpassed the server side timeout.
 
### Known limitations

[N/A]

### DIFF
--- a/services/horizon/internal/flags.go
+++ b/services/horizon/internal/flags.go
@@ -453,7 +453,7 @@ func Flags() (*Config, support.ConfigOptions) {
 				if duration < 0 {
 					return fmt.Errorf("%s cannot be negative", co.Name)
 				}
-				*(co.ConfigKey.(*time.Duration)) = time.Duration(viper.GetInt(co.Name)) * time.Second
+				*(co.ConfigKey.(*time.Duration)) = time.Duration(duration) * time.Second
 				return nil
 			},
 			Usage: "defines the timeout for when horizon will cancel all postgres queries connected to an HTTP request. The timeout is measured in seconds since the start of the HTTP request. Note, this timeout does not apply to POST /transactions. " +
--- a/services/horizon/internal/flags.go
+++ b/services/horizon/internal/flags.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"os/exec"
 	"strings"
+	"time"
 
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/viper"
@@ -69,6 +70,7 @@ const (
 	StellarTestnet = "testnet"
 
 	defaultMaxHTTPRequestSize = uint(200 * 1024)
+	clientQueryTimeoutNotSet  = -1
 )
 
 var (
@@ -438,15 +440,27 @@ func Flags() (*Config, support.ConfigOptions) {
 			UsedInCommands: ApiServerCommands,
 		},
 		&support.ConfigOption{
-			Name:           "cancel-db-query-timeout",
-			ConfigKey:      &config.CancelDBQueryTimeout,
-			OptType:        types.Int,
-			FlagDefault:    0,
-			CustomSetValue: support.SetDuration,
+			Name:        "client-query-timeout",
+			ConfigKey:   &config.ClientQueryTimeout,
+			OptType:     types.Int,
+			FlagDefault: clientQueryTimeoutNotSet,
+			CustomSetValue: func(co *support.ConfigOption) error {
+				if !support.IsExplicitlySet(co) {
+					*(co.ConfigKey.(*time.Duration)) = time.Duration(co.FlagDefault.(int))
+					return nil
+				}
+				duration := viper.GetInt(co.Name)
+				if duration < 0 {
+					return fmt.Errorf("%s cannot be negative", co.Name)
+				}
+				*(co.ConfigKey.(*time.Duration)) = time.Duration(viper.GetInt(co.Name)) * time.Second
+				return nil
+			},
 			Usage: "defines the timeout for when horizon will cancel all postgres queries connected to an HTTP request. The timeout is measured in seconds since the start of the HTTP request. Note, this timeout does not apply to POST /transactions. " +
-				"The difference between cancel-db-query-timeout and connection-timeout is that connection-timeout applies a postgres statement timeout whereas cancel-db-query-timeout will send an additional request to postgres to cancel the ongoing query. " +
-				"Generally, cancel-db-query-timeout should be configured to be higher than connection-timeout to allow the postgres statement timeout to kill long running queries without having to send the additional cancel request to postgres. " +
-				"By default, cancel-db-query-timeout will be set to twice the connection-timeout.",
+				"The difference between client-query-timeout and connection-timeout is that connection-timeout applies a postgres statement timeout whereas client-query-timeout will send an additional request to postgres to cancel the ongoing query. " +
+				"Generally, client-query-timeout should be configured to be higher than connection-timeout to allow the postgres statement timeout to kill long running queries without having to send the additional cancel request to postgres. " +
+				"By default, client-query-timeout will be set to twice the connection-timeout. Setting client-query-timeout to 0 will disable the timeout which means that Horizon will never kill long running queries using the cancel request, however, " +
+				"long running queries can still be killed through the postgres statement timeout which is configured via the connection-timeout flag.",
 			UsedInCommands: ApiServerCommands,
 		},
 		&support.ConfigOption{
@@ -995,9 +1009,9 @@ func ApplyFlags(config *Config, flags support.ConfigOptions, options ApplyOption
 			" If Horizon is behind both, use --behind-cloudflare only")
 	}
 
-	if config.CancelDBQueryTimeout == 0 {
+	if config.ClientQueryTimeout == clientQueryTimeoutNotSet {
 		// the default value for cancel-db-query-timeout is twice the connection-timeout
-		config.CancelDBQueryTimeout = config.ConnectionTimeout * 2
+		config.ClientQueryTimeout = config.ConnectionTimeout * 2
 	}
 
 	return nil
--- a/services/horizon/internal/flags.go
+++ b/services/horizon/internal/flags.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"os/exec"
 	"strings"
+	"time"
 
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/viper"
@@ -435,6 +436,17 @@ func Flags() (*Config, support.ConfigOptions) {
 			FlagDefault:    55,
 			CustomSetValue: support.SetDuration,
 			Usage:          "defines the timeout of connection after which 504 response will be sent or stream will be closed, if Horizon is behind a load balancer with idle connection timeout, this should be set to a few seconds less that idle timeout, does not apply to POST /transactions",
+			UsedInCommands: ApiServerCommands,
+		},
+		&support.ConfigOption{
+			Name:           "cancel-db-query-timeout",
+			ConfigKey:      &config.CancelDBQueryTimeout,
+			OptType:        types.Int,
+			CustomSetValue: support.SetDuration,
+			Usage: "defines the timeout for when horizon will cancel all postgres queries connected to an HTTP request. The timeout is measured in seconds since the start of the HTTP request. Note, this timeout does not apply to POST /transactions. " +
+				"The difference between cancel-db-query-timeout and connection-timeout is that connection-timeout applies a postgres statement timeout whereas cancel-db-query-timeout will send an additional request to postgres to cancel the ongoing query. " +
+				"Generally, cancel-db-query-timeout should be configured to be higher than connection-timeout to allow the postgres statement timeout to kill long running queries without having to send the additional cancel request to postgres. " +
+				"By default, cancel-db-query-timeout will be set to 2 seconds more than connection-timeout.",
 			UsedInCommands: ApiServerCommands,
 		},
 		&support.ConfigOption{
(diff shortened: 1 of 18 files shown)


services/horizon/internal/init.go
```diff
@@ -45,38 +45,29 @@ func mustInitHorizonDB(app *App) {
 			log.Fatalf("max open connections to horizon db must be greater than %d", ingest.MaxDBConnections)
 		}
 	}
+	serverSidePGTimeoutConfigs := []db.ClientConfig{
+		db.StatementTimeout(app.config.ConnectionTimeout),
+		db.IdleTransactionTimeout(app.config.ConnectionTimeout),
+	}
```
tamirms, line 51: the ingestion system uses a separate connection pool https://github.com/stellar/go/blob/6036a0779c6cd6299ffddb7c70088d755a379474/services/horizon/internal/init.go#L87-L89

so we can apply these timeout settings to `app.historyQ` and `app.primaryHistoryQ` because we know that those db connections will never be used in ingestion and will only be used when serving HTTP requests

services/horizon/internal/flags.go
```diff
@@ -437,6 +437,18 @@ func Flags() (*Config, support.ConfigOptions) {
...
+		&support.ConfigOption{
+			Name:           "cancel-db-query-timeout",
+			ConfigKey:      &config.CancelDBQueryTimeout,
+			OptType:        types.Int,
+			FlagDefault:    0,
+			CustomSetValue: support.SetDuration,
+			Usage: "defines the timeout for when horizon will cancel all postgres queries connected to an HTTP request. The timeout is measured in seconds since the start of the HTTP request. Note, this timeout does not apply to POST /transactions. " +
+				"The difference between cancel-db-query-timeout and connection-timeout is that connection-timeout applies a postgres statement timeout whereas cancel-db-query-timeout will send an additional request to postgres to cancel the ongoing query. " +
+				"Generally, cancel-db-query-timeout should be configured to be higher than connection-timeout to allow the postgres statement timeout to kill long running queries without having to send the additional cancel request to postgres. " +
+				"By default, cancel-db-query-timeout will be set to twice the connection-timeout.",
```
sreuland, line 449:
```suggestion
				"When set to 0, which is the default, it will trigger cancel-db-query-timeout to be set to twice the connection-timeout.",
```

sreuland: minor, should the flag allow for disabling by the user, I see later in middleware, `0` [equates to skipping application of the flag](https://github.com/stellar/go/pull/5223/files#diff-b09eb588ef67c1bb0180fa0ed67e2ab66968d5fc317a330740a1f9fed5e9a962R351), but it would never be exercised as of now.

support/db/session.go
```diff
@@ -12,28 +12,60 @@ import (
...
+}
+
 // Begin binds this session to a new transaction.
 func (s *Session) Begin(ctx context.Context) error {
 	if s.tx != nil {
 		return errors.New("already in transaction")
 	}
+	ctx, cancel, err := s.context(ctx)
+	if err != nil {
+		cancel()
```
sreuland, line 51: the cancel invoke could be removed, it can only be noop when Session.context() returns an err

support/db/session.go
```diff
@@ -12,28 +12,60 @@ import (
...
+		return err
+	}
 
 	tx, err := s.DB.BeginTxx(ctx, nil)
 	if err != nil {
 		if knownErr := s.handleError(err, ctx); knownErr != nil {
+			cancel()
```
sreuland, line 58: could reduce cancel() invokes in this err block to once before knownErr check

services/horizon/internal/flags.go
```diff
@@ -437,6 +437,18 @@ func Flags() (*Config, support.ConfigOptions) {
 			Usage:          "defines the timeout of connection after which 504 response will be sent or stream will be closed, if Horizon is behind a load balancer with idle connection timeout, this should be set to a few seconds less that idle timeout, does not apply to POST /transactions",
 			UsedInCommands: ApiServerCommands,
 		},
+		&support.ConfigOption{
+			Name:           "cancel-db-query-timeout",
```
Shaptic, line 441: just driving by - can we simplify the name to `query-timeout`?

support/db/session.go
```diff
@@ -135,7 +177,13 @@ func (s *Session) Get(ctx context.Context, dest interface{}, query sq.Sqlizer) e
 // GetRaw runs `query` with `args`, setting the first result found on
 // `dest`, if any.
 func (s *Session) GetRaw(ctx context.Context, dest interface{}, query string, args ...interface{}) error {
-	query, err := s.ReplacePlaceholders(query)
+	ctx, cancel, err := s.context(ctx)
```
sreuland, line 180: just to confirm on these session 'Raw' methods, is the expectation that a Session.Begin() could have been called prior to this, i.e. `s.txCancel` may be set or no tx is possible also? It seems like should use `s.txCancel` here if present rather than wrap the context again?

support/db/session_test.go
```diff
@@ -140,6 +140,38 @@ func TestStatementTimeout(t *testing.T) {
...
+	deadline := time.Now().Add(time.Hour)
+	requestCtx := context.WithValue(context.Background(), &DeadlineCtxKey, deadline)
+	resultCtx, _, err = sess.context(requestCtx)
+	assert.NoError(t, err)
+	d, ok := resultCtx.Deadline()
+	assert.True(t, ok)
+	assert.Equal(t, deadline, d)
+
+	requestCtx, cancel := context.WithDeadline(requestCtx, time.Now().Add(time.Minute*30))
+	resultCtx, _, err = sess.context(requestCtx)
```
sreuland, line 164: it might be worthwhile to capture the cancel func returned from Session.context(), and assert that invoking it does not effect the original source context, `requestCtx` in this case.

support/db/session_test.go
```diff
@@ -140,6 +140,38 @@ func TestStatementTimeout(t *testing.T) {
...
+	assert.NoError(t, err)
+	d, ok = resultCtx.Deadline()
+	assert.True(t, ok)
+	assert.Equal(t, deadline, d)
+
+	cancel()
```
sreuland, line 170: this looks like a good opportunity to assert the new expectation that when an 'upstream' context does get cancelled, it does not propagate to the 'deadline' context, i.e. `assert.NoError(resultCtx.Err())`

sreuland (approved):
nice integration, left some optional comments for consideration.

services/horizon/internal/flags.go
Shaptic: sure, good point! maybe `client-query-timeout` then? I guess in my mind I read the flag as "cancel query" and think "what the heck is a cancel query, and why does it need a timeout?"

(also to be clear I don't want to delay this PR at all)

services/horizon/internal/flags.go
tamirms: I didn't want to allow the flag to be disabled because I think it could result in confusion. If we skip the application of this flag, the connection timeout will be set to both the postgres server side statement timeouts and the query cancelations on the horizon side. I think this would be unintuitive because, as a user, my first thought is that disabling the query cancelation timeout would result in no query cancelation whatsoever.

If we wanted to allow the option of having no query cancelation at all we could equate 0 to that behavior. -1 could be a default which means cancel-db-query-timeout will be set to twice the connection-timeout value. The user still has the option of setting cancel-db-query-timeout and connection-timeout to the same value which will result in more or less the status quo behavior of horizon prior to this PR.

support/db/session.go
tamirms: I think that might change the behavior of how the error counter is incremented in `handleError()`

support/db/session.go
tamirms: > is the expectation that a Session.Begin() could have been called prior to this, i.e. s.txCancel may be set or no tx is possible also?

yes, both cases are possible.

services/horizon/internal/flags_test.go
```diff
@@ -207,22 +208,69 @@ func Test_createCaptiveCoreConfig(t *testing.T) {
...
-		"NETWORK_PASSPHRASE":       "Standalone Network ; February 2017",
-		"APPLY_MIGRATIONS":         "true",
-		"CHECKPOINT_FREQUENCY":     "8",
-		"MAX_DB_CONNECTIONS":       "50",
-		"ADMIN_PORT":               "6060",
-		"PORT":                     "8001",
-		"CAPTIVE_CORE_BINARY_PATH": os.Getenv("HORIZON_INTEGRATION_TESTS_CAPTIVE_CORE_BIN"),
-		"CAPTIVE_CORE_CONFIG_PATH": "../docker/captive-core-classic-integration-tests.cfg",
-		"CAPTIVE_CORE_USE_DB":      "true",
+func TestClientQueryTimeoutFlag(t *testing.T) {
```
sreuland, line 211: thorough test coverage!

services/horizon/internal/flags.go
```diff
@@ -437,6 +439,30 @@ func Flags() (*Config, support.ConfigOptions) {
...
+			CustomSetValue: func(co *support.ConfigOption) error {
+				if !support.IsExplicitlySet(co) {
+					*(co.ConfigKey.(*time.Duration)) = time.Duration(co.FlagDefault.(int))
+					return nil
+				}
+				duration := viper.GetInt(co.Name)
+				if duration < 0 {
+					return fmt.Errorf("%s cannot be negative", co.Name)
+				}
+				*(co.ConfigKey.(*time.Duration)) = time.Duration(viper.GetInt(co.Name)) * time.Second
```
sreuland, line 456: minor, could reference `duration` instead of making viper call again

support/db/session.go
```diff
@@ -12,28 +12,66 @@ import (
...
+// original context.
+// Note that the override will not be applied if requestCtx has already been terminated.
+// The timeout can be disabled by setting the DeadlineCtxKey value to 0.
+func (s *Session) context(requestCtx context.Context) (context.Context, context.CancelFunc, error) {
+	var ctx context.Context
+	var cancel context.CancelFunc
+
+	deadline, ok := requestCtx.Value(&DeadlineCtxKey).(time.Time)
+	if !ok {
+		ctx, cancel = context.WithCancel(requestCtx)
```
sreuland, line 34: are there any existing paths in code that depend on this condition of absence of `DeadlineCtxKey`? if not, this is dead code, just for consideration could be more strict and return `(requestCtx, nil, errors.New("invalid context config, no DeadlineCtxKey"))`.

support/db/session.go
sreuland: yes, thanks!

support/db/session.go
sreuland: ok, and in the case where a tx has been started and `s.txCancel` is populated with that ctx, we still want to ignore that here and use caller passed 'ctx'

support/db/session.go
tamirms: the ingestion db session will not have `DeadlineCtxKey` . only the db sessions used in the request path will set a deadline override

support/db/session.go
```diff
@@ -12,28 +12,67 @@ import (
...
+// If the override exists, we return a new context with the desired deadline. Otherwise, we return the
+// original context.
+// Note that the override will not be applied if requestCtx has already been terminated.
+// The timeout can be disabled by setting the DeadlineCtxKey value to 0.
+func (s *Session) context(requestCtx context.Context) (context.Context, context.CancelFunc, error) {
+	var ctx context.Context
+	var cancel context.CancelFunc
+
+	deadline, ok := requestCtx.Value(&DeadlineCtxKey).(time.Time)
+	if !ok {
```
sreuland, line 35: this condition is only used by tests such as those in session_test.go, which don't config `DeadlineCtxKey`, it's not a valid production code path correct? if so, a comment line stating the purpose would be good to highlight.

sreuland (approved):
lgtm, thanks for incorporating feedback.

support/db/session.go
```diff
@@ -12,28 +12,67 @@ import (
...
+		return requestCtx, noop, nil
+	}
+
+	// if requestCtx is already terminated don't proceed with the db statement
+	if requestCtx.Err() != nil {
+		return requestCtx, nil, requestCtx.Err()
+	}
+
+	if deadline.IsZero() {
```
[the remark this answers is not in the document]
tamirms: if the http client disconnects, the context attached to the http request in horizon would be canceled. If that happens before the query is made to the DB, horizon will not attempt to do the query and will just return with an error that eventually gets translated to ClientDisconnected in the response. If the request context is canceled during the query, horizon will still wait until the query finishes. If that was the last query in the horizon endpoint, horizon will respond with a 200 and populate the response body with the json payload. If there was another query required, the session would detect the request context was cancelled and would return an error that eventually gets translated to ClientDisconnected in the response.

Even without this change, it is still possible to get a 200 response when the http client disconnects. That case will occur if the client disconnects after the last db query finishes but before horizon written any http header / response

PR #5224, opened 2024-02-27
historyarchive: Add round-robin, error-resilience, and back-off to the `ArchivePool`

### LINKED ISSUE #5167: historyarchive: Harden the `ArchivePool` to be robust against errors.
The current implementation of `ArchivePool` uses a random archive for each individual call, bubbling up any errors to the caller. It does an initial filtering on creation, but it does not handle any errors that occur during usage.

The `ArchivePool` should intelligently deal with problematic archives:

 * evaluate whether rand or round-robin makes more sense, in general
 * if there is an error when making a call on a certain archive, it should retry with another
   - this should have an intelligent back-off mechanism: i.e. retry immediately if it's a new archive, but back-off if it's the same or a problematic archive
 * the pool should track problematic archives, i.e. ones that give errors with high frequency/likelihood
   - should it remove problematic archives?
   - keep in mind how many remain in the pool after removal
 * consider edge cases the whole time, e.g. just one archive in the pool

This can be handled at the caller level, but it makes sense to do this in the pool itself, if possible.

Prior work: #3402

Shaptic:
### What
This adds two main components to the `ArchivePool` implementation that will significantly add resilience to the interface:

* Previously, each method call would just forward the method to a randomly-chosen archive in the pool. Now, it will (persistently) **round-robin** through each archive in the pool with a **constant back-off** mechanism (250ms/archive) of up to 3 retries (so 4 total queries per attempt in the worst case).

### Why
Closes #5167.

### Known limitations
For feedback purposes, I haven't actually implemented this for every `ArchiveInterface` method as I'm seeking feedback prior to writing a bunch of repetitive code I may have to wipe out later.


historyarchive/archive_pool.go
```diff
@@ -6,30 +6,46 @@ package historyarchive
...
-func NewArchivePool(archiveURLs []string, opts ArchiveOptions) (ArchivePool, error) {
+func NewArchivePool(archiveURLs []string, opts ArchiveOptions) (ArchiveInterface, error) {
 	if len(archiveURLs) <= 0 {
 		return nil, errors.New("No history archives provided")
 	}
 
+	ap := ArchivePool{
+		pool:   make([]ArchiveInterface, 0, len(archiveURLs)),
+		errors: make(map[ArchiveInterface]*errStats, len(archiveURLs)),
+		curr:   rand.Intn(len(archiveURLs)), // don't necessarily start at zero
```
tamirms, line 44: not all `archiveURLs` are guaranteed to be included in `pool` because you skip a url if it's invalid. `curr` should be set after `pool` is populated

tamirms: or you could make `cur` start at 0 and you can shuffle the `pool` list before returning from `NewArchivePool()`

historyarchive/archive_pool.go
```diff
@@ -62,80 +79,234 @@ func (pa ArchivePool) GetStats() []ArchiveStats {
...
+	return bsize, pa.runOnEach(func(ai ArchiveInterface) error {
+		var err error
+		bsize, err = ai.BucketSize(bucket)
+		return err
+	})
 }
 
-func (pa ArchivePool) BucketSize(bucket Hash) (int64, error) {
-	return pa.GetAnyArchive().BucketSize(bucket)
+func (pa *ArchivePool) CategoryCheckpointExists(cat string, chk uint32) (bool, error) {
```
tamirms, line 221: why do the remaining methods not use `runOnEach()`?

tamirms: nevermind, I just re-read the pr description:

> For that reason, I haven't actually implemented this for every ArchiveInterface method as I'm seeking feedback prior to writing a bunch of repetitive code I may have to wipe out later.

historyarchive/archive_pool.go
```diff
@@ -62,80 +79,234 @@ func (pa ArchivePool) GetStats() []ArchiveStats {
...
+			// > A negative or zero duration causes Sleep to return immediately.
+			//
+			// https://pkg.go.dev/time#Sleep
+			time.Sleep(bestBackoff)
+			ai = bestPool
+		}
+
+		// Reaching this point means either no back-off or `ai` was backed-off
+		// from, so we can actually do execution now.
+		if err := runner(ai); err != nil {
```
tamirms, line 157: if we get a context cancel or context deadline exceeded error we don't want to retry because subsequent attempts are guaranteed to fail

tamirms:
The checkpoint change reader retries history archive operations here:

https://github.com/stellar/go/blob/master/ingest/checkpoint_change_reader.go#L132-L142

We should get rid of the retry logic in the above code because there's no use in having retries implemented in multiple layers.

historyarchive/archive_pool.go
Shaptic: hmm, that may be true but I don't think that's the responsibility of the pool itself. once they're all exhausted, the same error will bubble up to a higher level, anyway.

Shaptic:
> We should get rid of the retry logic in the above code because there's no use in having retries implemented in multiple layers.

While I agree in practice, I don't think we should project behavior from one layer to another. For example, a user of the ingest package may choose to only use a single archive and yet still want retries.

Shaptic:
> I think we should start with a simpler implementation (e.g. retry at most n times in a round-robin fashion). 

Yeah, I generally agree, @tamirms. And I don't think we have any metrics that we're trying to fight against, per se: it's more that we wanted the `ArchivePool` to retry automatically so that we would get resilience "for free" in every place that uses the pool. And I thought that if you're going to add retries, you might as well add back-offs, since we've seen issues with a lack of back-off before in other places.

However, I think keeping it simple is a totally fine call.

tamirms:
> a user of the ingest package may choose to only use a single archive and yet still want retries.

this is a good point. would it make sense to have another archive wrapper which solely encapsulates the retry / backoff logic? then we could enable both the pooling and retry properties by composing the `ArchivePool` with the `RetryArchive` wrapper.

> And I thought that if you're going to add retries, you might as well add back-offs, since we've seen issues with a lack of back-off before in other places.

I think it's ok to include back-offs but I'm not sure that we should have something sophisticated which tries to be clever about adjusting the backoff based on the estimated reliability of each archive in the pool.

Shaptic:
> would it make sense to have another archive wrapper which solely encapsulates the retry / backoff logic

It's a good idea, but I think no: retrying across the pool is better than retrying on each individual archive (in the sense that it is likelier to result in success), so we need the pool-level retries first. If we did add a `RetryArchive`, we'd still need this PR.

> adjusting the backoff based on the estimated reliability of each archive in the pool

The back-off is linear with the number of consecutive errors; there's no other adjustment. Here, we just prefer archives with a lower back-off.

------

@tamirms I dropped the back-off related code in 75e50a1e96e2921090d1b5a65834cd0b433309dd and just kept a simple round-robin retry approach so that we could evaluate it. Do you think we should keep it simple and just start there?

historyarchive/archive_pool.go
Shaptic: good catch! fixed in 2585dfe7b8a2c113af59bd536dfbda5356b5af49

tamirms:
> retrying across the pool is better than retrying on each individual archive (in the sense that it is likelier to result in success), so we need the pool-level retries first. If we did add a RetryArchive, we'd still need this PR.

if you compose `RetryArchive` with `PoolArchive` you would achieve the desired retry behavior.

Assume `PoolArchive` cycles through different urls for each invocation in a round robin. You can wrap the `PoolArchive` in a `RetryArchive` and, when the `RetryArchive` invokes the underlying `PoolArchive` again during retries, the `PoolArchive's` round robin behavior will ensure that a different history archive url is selected during the retried request.

tamirms:
> The back-off is linear with the number of consecutive errors; there's no other adjustment. Here, we were (i.e. before the latest push) just preferring archives with a lower back-off.

I think having a constant back-off should be simple and effective

historyarchive/archive_pool.go
tamirms: > hmm, that may be true but I don't think that's the responsibility of the pool itself. once they're all exhausted, the same error will bubble up to a higher level, anyway.

If there is a backoff in between each request then that will delay termination by the backoff time multiplied by the number of retries we allow.

We use https://github.com/cenkalti/backoff in soroban-rpc to do retries and that library does have an option to support context cancellation and they also allow you to classify some errors as permanent so that the retry logic is immediately aborted. So, I do think this feature is relevant to the retry logic.

Shaptic:
@tamirms I've simplified the design somewhat in line with how we discussed it, but not to the same level of depth. Rather than designing a `RetriableArchive` (which I think would be better scoped in a separate issue relative to #5167), I just used the [`backoff.Retry`](https://pkg.go.dev/github.com/cenkalti/backoff/v4#Retry) you referenced to back-off the loop of 

This isn't as... equitable? as the previous solution (which distributed backoffs better to each individual archive) since it backs off on the entire round-robin for a single call, but I think it's still a substantial improvement relative to what we have before and so it achieves the goals in #5167.

It also classifies context errors as permanent like you advised. PTA(nother)L before I do it for the remaining methods!

historyarchive/archive_pool.go
```diff
@@ -62,80 +79,135 @@ func (pa ArchivePool) GetStats() []ArchiveStats {
...
+	return pa.pool[pa.curr]
+}
+
+// runRoundRobin is a helper method that will run a particular action on every
+// archive in the pool until it succeeds or the pool is exhausted (whichever
+// comes first), repeating with a constant 500ms backoff.
+func (pa *ArchivePool) runRoundRobin(runner func(ai ArchiveInterface) error) error {
+	return backoff.Retry(func() error {
+		var lastErr error
+		for range pa.pool {
```
tamirms, line 99: if I understand correctly, in the worst case, we will try every single archive 3 times each, which I think might be too much.

If you remove the for loop entirely from this closure, in the worst case, we will try up to 3 different archives with back off of half a second between each attempt. That seems more reasonable to me.

Shaptic: It's a good point! I would argue there's no reason to back off before retrying a _different_ archive, but you're right that this strikes a much better balance between back-off and load distribution than back-offs on the whole cycle. Fixed in efbbda6f.

Shaptic: (I also lowered the back-off time.)

historyarchive/archive_pool.go
```diff
@@ -62,80 +79,133 @@ func (pa ArchivePool) GetStats() []ArchiveStats {
...
+			return nil
+		}
+
+		if stats := ai.GetStats(); len(stats) > 0 {
+			log.WithField("error", err).Warnf(
+				"Encountered an error with archive '%s'",
+				stats[0].GetBackendName())
+		}
+
+		if errors.Is(err, context.Canceled) ||
```
tamirms, line 110: nit: we usually try to avoid logging context canceled errors (e.g. https://github.com/stellar/go/blob/master/services/horizon/internal/ingest/main.go#L675-L677 ) , so I think it would be better to do this check before emitting the "Encountered an error with archive" log

tamirms:
@Shaptic this approach looks good to me!

historyarchive/archive_pool.go
Shaptic: great catch!

historyarchive/archive_pool.go
```diff
@@ -62,80 +79,189 @@ func (pa ArchivePool) GetStats() []ArchiveStats {
...
+		c, err = ai.ListBucket(dp)
+		return <-err
+		// TODO: Does the above block in the absence of errors?? Otherwise, can do this:
+		//
+		// if errCount := drainErrors(err); errCount > 0 {
+		// 	return fmt.Errorf("encountered %d errors with ListBucket on %+v", errCount, dp)
+		// }
+		// return nil
+	})
+	return c, err
```
Shaptic, line 229: @tamirms I added it for the rest of the methods, but I wasn't sure what the right resolution method was here. We can't assume that the `chan error` is peekable, so I'm not certain how we should bubble errors up. Do you have a preference between the two approaches here?

MonsieurNicolas:
Out of curiosity, why is random (assuming uniform distribution) worst than round robin? The issue linked seems to imply that either would work as long as failure information is taken into account

historyarchive/archive_pool.go
tamirms: I would recommend leaving the functions which return `(chan string, chan error)` as is. We can't really redo retries on these functions in an asynchronous manner. As you have noted, in the case where there are no errors, we will essentially make  `ListBucket()` block until the entire operation is complete

Shaptic:
@MonsieurNicolas yeah you're right that there's nothing inherently better about either, but because we want retries then it's better to have certainty that we won't reuse the same archive again (which is of course doable with random selection, too, but is cleaner with round robin).

historyarchive/archive_pool.go
Shaptic: makes sense, done!

tamirms (approved):
🎉

MonsieurNicolas:
> @MonsieurNicolas yeah you're right that there's nothing inherently better about either, but because we want retries then it's better to have certainty that we won't reuse the same archive again (which is of course doable with random selection, too, but is cleaner with round robin).

makes sense. we just need to make sure to shuffle the list on startup so that you don't introduce a bias towards the first archive across deployments/restarts

Shaptic:
@MonsieurNicolas yep, great call-out! that's sort of being done with [this line](https://github.com/stellar/go/pull/5224/files#diff-02a66587953ebf40035ed81414f2035cfb11f05e1aed683d31b0f6073d99a8ecR67):

```golang
ap.curr = rand.Intn(len(ap.pool)) // don't necessarily start at zero
```

which effectively achieves the same thing: a random starting point for the round robin.

Shaptic:
Hmm, thinking about that more: a random starting point is definitely not the same level of randomness as a shuffle (you could argue that you are still biasing against _groups_ of archives since most people will pass them in order) when it comes to _traversal_, but I think since the goal is to alleviate concentrating load on a particular group during _restarts_, the same end goal is achieved either way.

PR #5225, opened 2024-02-28
services/horizon: Add new metrics counters for db connection close events

### LINKED ISSUE #5217: services/horizon: emit new metrics counters for timeouts
no metrics keys capture when horizon app server triggers a db session disconnect/abort due to CONNECTION_TIMEOUT or http client disconnects which trigger a similar db session disconnect/abort.

new counter metrics key:
`horizon-db-session-forced-disconnects_total`
with two labels:
`server` - counter of occurrences of db session disconnects that were triggered due to CONNECTION_TIMEOUT
`client` - counter of occurrences of db session disconnects that were triggered due to early closure of client http connection.

sreuland:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Added  new db layer metrics:
`client_closed_session_total` 
`server_timeout_closed_session_total`
`statement_timeout_closed_session_total`

### Why

obtain insights on what types of events underly db sessions timeouts
Closes #5217


tamirms:
Instead of having 3 separate prometheus counters for these specific errors, I think it would be better to have one prometheus counter for all possible [postgres errors](https://www.postgresql.org/docs/12/errcodes-appendix.html) returned by the driver. This could be implemented by having a label in the prometheus counter which represents the postgres error code.

Then, you can wrap the session functions with some code which checks if the returned error is a postgres server error ( https://stackoverflow.com/questions/37560534/does-the-error-returned-by-db-exec-have-a-code ) and , in that case, you can increment the metric with the appropriate error code label.

sreuland:
> I think it would be better to have one prometheus counter for all possible [postgres errors](https://www.postgresql.org/docs/12/errcodes-appendix.html) returned by the driver. This could be implemented by having a label in the prometheus counter which represents the postgres error code.

I re-worked it per suggestion for single metric with labels, the new [metrics gathering routine](https://github.com/stellar/go/pull/5225/files#diff-d1193552ca1c1b01294508b0d7319c0d56b9b7797e9b3034cbc80b805b67ae58R372-R420) attempts to get pg server error code if libpq provides it.

support/db/metrics.go
```diff
@@ -356,6 +369,55 @@ func getQueryType(ctx context.Context, query squirrel.Sqlizer) QueryType {
...
+	if dbErr == nil || s.NoRows(dbErr) {
+		return
+	}
+
+	abendOrigin := "libpq"
+	abendType := "error"
+	abendCondition := ""
+	var abendDbErrorCode pq.ErrorCode
+
+	if err, ok := dbErr.(*pq.Error); ok {
```
tamirms, line 382: we should use https://pkg.go.dev/errors#As because that will handle cases where the db error is hidden under several layers of wrapping

support/db/session.go
```diff
@@ -232,28 +233,48 @@ func (s *Session) NoRows(err error) bool {
...
 	}
 
+	for _, handler := range s.errorHandlers {
+		handler(dbErr, ctx)
+	}
+
+	var abendDbErrorCode pq.ErrorCode
+
+	// libpq will only return pg server error if context did not cancel/timeout
+	if err, ok := dbErr.(*pq.Error); ok {
```
tamirms, line 259: as I mentioned in the other comment, we should use https://pkg.go.dev/errors#As

support/db/metrics.go
```diff
@@ -58,6 +60,7 @@ type SessionWithMetrics struct {
 	maxLifetimeClosedCounter prometheus.CounterFunc
 	roundTripProbe           *roundTripProbe
 	roundTripTimeSummary     prometheus.Summary
+	abendCounter             *prometheus.CounterVec
```
tamirms, line 63: I think errorCounter would be a more intuitive name as I think many engineers might not be familiar with the term "abend"

support/db/metrics.go
```diff
@@ -400,14 +461,21 @@ func (s *SessionWithMetrics) Select(ctx context.Context, dest interface{}, query
...
 }
 
+func (s *SessionWithMetrics) Query(ctx context.Context, query squirrel.Sqlizer) (*sqlx.Rows, error) {
+	return s.SessionInterface.Query(ctx, query)
+}
+
+func (s *SessionWithMetrics) QueryRaw(ctx context.Context, query string, args ...interface{}) (*sqlx.Rows, error) {
+	return s.Query(ctx, squirrel.Expr(query, args...))
+}
+
```
tamirms, line 478: these two functions are not necessary because they delegate to `SessionInterface` but that is already the default behavior when you have embedding in Go structs:

https://gobyexample.com/struct-embedding

support/db/metrics.go
```diff
@@ -374,8 +436,7 @@ func (s *SessionWithMetrics) Get(ctx context.Context, dest interface{}, query sq
 		}).Inc()
 	}()
 
-	err = s.SessionInterface.Get(ctx, dest, query)
-	return err
```
tamirms, line 378: I think this structure might be intentional because `err` is referenced the timer closure and the defer closure

support/db/session.go
```diff
@@ -232,28 +233,48 @@ func (s *Session) NoRows(err error) bool {
...
-		// if libpq waits too long to obtain conn from pool, can get ctx timeout before server trip
-		return ErrTimeout
-	case strings.Contains(err.Error(), "pq: canceling statement due to user request"):
+		// when horizon's context times out(it's set to app connection-timeout), it triggers pg to emit "pq: canceling statement due to user request"
+		// so, in order of precedence, check the ctx deadline first to classify this as a timeout, not a cancel
 		return ErrTimeout
-	case strings.Contains(err.Error(), "pq: canceling statement due to conflict with recovery"):
+	case abendDbErrorCode.Name() == "query_canceled":
+		// https://www.postgresql.org/docs/12/errcodes-appendix.html, query_canceled
+		return ErrStatementTimeout
```
tamirms, line 272: we previously returned `ErrTimeout` for this case ("pq: canceling statement due to user request"). Any reason for changing that behavior?

support/db/session.go
```diff
@@ -232,28 +233,48 @@ func (s *Session) NoRows(err error) bool {
...
+	case strings.Contains(dbErr.Error(), "pq: canceling statement due to conflict with recovery"):
 		return ErrConflictWithRecovery
-	case strings.Contains(err.Error(), "driver: bad connection"):
+	case strings.Contains(dbErr.Error(), "driver: bad connection"):
 		return ErrBadConnection
-	case strings.Contains(err.Error(), "pq: canceling statement due to statement timeout"):
```
tamirms, line 254: why is the statement timeout case deleted?

support/db/metrics.go
```diff
@@ -356,6 +369,55 @@ func getQueryType(ctx context.Context, query squirrel.Sqlizer) QueryType {
...
+	abendType := "error"
+	abendCondition := ""
+	var abendDbErrorCode pq.ErrorCode
+
+	if err, ok := dbErr.(*pq.Error); ok {
+		// libpq only provides a pg.Error if a server trip was made, otherwise it may not be present
+		// the error could just be context or libpq err
+		abendDbErrorCode = err.Code
+		abendOrigin = "db"
+		abendCondition = abendDbErrorCode.Name()
```
tamirms, line 387: unfortunately the lib/pq library has not updated its error code mapping for several years:

https://github.com/lib/pq/blob/master/error.go#L77

It seems like it only supports errors from postgres 9.3 and we recommend running postgres 12 for horizon. So I think it would be better to include the error code in the prometheus metric. The error code is not as user friendly as the name but at least it's guaranteed to be present

support/db/metrics.go
```diff
@@ -356,6 +369,55 @@ func getQueryType(ctx context.Context, query squirrel.Sqlizer) QueryType {
...
+	}
+
+	switch {
+	// check the context first, if was canceled or timed out,
```
tamirms, line 391: I think we actually want to do the opposite. First we check the postgres error code if it exists. Otherwise we check the context. The reasoning for that is we primarily want insight into how postgres is terminating our queries and the postgres error codes provide the most certainty about that.

support/db/metrics.go
tamirms: or you can use the error name if it is present and otherwise default to the error code

support/db/metrics.go
sreuland: 'cancel' and 'timeout' don't seem to fit as well under 'error' categorization, since they are intentional aspects, i'll change it, unless 'abnormalConditionsCounter' , idk, can't find a better synonym...

support/db/metrics.go
sreuland: thx, I'll remove, I overrode them earlier to apply the metrics scrape fn on their err's ,but have since changed that to be done just in Session.handleError()

support/db/metrics.go
tamirms: actually, now that I think about it, I don't think there's a need to check the context at all. Instead you can use https://pkg.go.dev/errors#Is to check if the returned error is `context.Cancelled` or `context.DeadlineExceeded`. So, I think the new workflow should be:

check if the error is pq.Error, if so include those details in the metric. If it's not a pq.Error check if it's a context error. Finally, failing all those options, report the error in the metric as unknown

tamirms: this basically changes the metric so that it's classifying all the DB errors we encounter. I know that's a bit from what is described in https://github.com/stellar/go/issues/5217 but I think from the DB error we could infer how the connection was closed (if at all).

support/db/session.go
sreuland: that will still be the outcome, ErrTimeout is asserted in [TestContextTimeoutDuringSql](https://github.com/stellar/go/pull/5225/files/fd3e0722fa1266c0176c804b954548e789293017#diff-b642ef4516e14debf2fefed8f54fdbf6a5f1f33f1e4b23d76dd19da5feba0274R42) and  [TestContextTimeoutBeforeSql](https://github.com/stellar/go/pull/5225/files/fd3e0722fa1266c0176c804b954548e789293017#diff-b642ef4516e14debf2fefed8f54fdbf6a5f1f33f1e4b23d76dd19da5feba0274R66), comments there provide more detail on ctx lifecycle aspects, i.e. notice how the 'condition' label for db server code gets populated only on the 'during' case.

there was subtle overlap on mapping of these conditions and what source generated them, i.e. libpq generated the error `"pq: canceling statement due to user request"` when the go context deadline is exceeded, so ErrTimeout was really equated to the root fact of context deadline.

support/db/session.go
sreuland: ErrStatementTimeout case is still present on [line 272](https://github.com/stellar/go/pull/5225/files/fd3e0722fa1266c0176c804b954548e789293017#diff-bc24bbc29b516b4d305accf90f8846637f0ff0e13f100aefc72c95f309dc7b32R272), and asserted on [TestStatementTimeout](https://github.com/stellar/go/pull/5225/files/fd3e0722fa1266c0176c804b954548e789293017#diff-b642ef4516e14debf2fefed8f54fdbf6a5f1f33f1e4b23d76dd19da5feba0274R127)

wanted to derive it more directly from the source being pg server side error, sqlstate=57014 - `query_canceled` rather than the libpq wrapped error "pq: canceling statement due to statement timeout"

support/db/metrics.go
sreuland: doh, great catch, thanks.

support/db/metrics.go
sreuland: done, https://github.com/stellar/go/pull/5225/commits/26c3316ab8f1a1acdbc17706821139f22ad8820b

support/db/metrics.go
sreuland: https://github.com/stellar/go/pull/5225/commits/26c3316ab8f1a1acdbc17706821139f22ad8820b

support/db/metrics.go
sreuland: yeah, so, rather than check `callerContext` for cancel/deadline, check the err returned from libpq calls instead, since libpq will wrapped the same ctx err's if any present into the final err emitted from a method call, and then I flip the order of precedence for interrogating the final err to derive the metric via (1) check for db-side errs, (2) check for ctx err's, (3) use the err string as-is

tamirms: @sreuland yeah, that's right! But, in case (3) I think we should just set the label to "unknown". There are two reasons for doing this:

1. having a high cardinality for label values can impact prometheus performance
2. prometheus label values cannot be free form strings. there are restrictions on what characters can appear in a label value https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels

support/db/metrics.go
sreuland: prefer to use one source consistently for searching aspects later, so will go with the sql-state error code value.

support/db/metrics.go
sreuland: yes, that case(3) data could be really insightful, but no way to sanitize the err strings, however, rather than `unknown` which infers something was there and the service didn't understand it, how about instead just empty string or `n/a` ? it tends to infer more of general consistent semantic for these cases where no db server code is available, like client side only ctx err's and any non-server libpq err's

tamirms: I'm not sure if prometheus will allow the empty string. `n/a` is fine by me

sreuland: [committed this change](https://github.com/stellar/go/pull/5225/commits/8aabea605b1de3555723c8dea77a406a971b3fd1), altho, i had to retain inclusion of the caller context during SessionWithMetrics.handlerErrorEvent(), observed libpq will not include wrapped err which has both server side pg.Error and ctx err that may have occurred in libpq during the server round trip, explained it in [comments here](https://github.com/stellar/go/pull/5225/files#diff-bc24bbc29b516b4d305accf90f8846637f0ff0e13f100aefc72c95f309dc7b32R260-R263).

the session_test.go has some decent coverage that demonstrates the final metrics output from these various cases, and I'm going to add one more to assert this feature that any pg errors will be mapped into the metric, by triggering a non-57014 sql state error code and confirming a metric is generated with condition=code

support/db/session.go
```diff
@@ -232,29 +234,60 @@ func (s *Session) NoRows(err error) bool {
...
+
+	// if libpql sends to server, and then any server side error is reported,
+	// libpq passes back only an pq.ErrorCode from method call
+	// even if the caller context generates a cancel/deadline error during the server trip,
+	// libpq will only return an instance of pq.ErrorCode as a non-wrapped error
+	if go_errors.As(dbErr, &pqErr) {
+		abendDbErrorCode = pqErr.Code
+	}
+
 	switch {
```
tamirms, line 268: We are not changing the behavior of how we replace db errors with horizon errors, right? If that is the case then I think we should leave this switch statement unchanged so we can minimize the code review diff and reduce the chance of unintentionally introducing behavioral changes to the `replaceWithKnownError()` functionality

support/db/session.go
```diff
@@ -3,13 +3,15 @@ package db
 import (
 	"context"
 	"database/sql"
+	go_errors "errors"
```
tamirms, line 6: nit: I suggest using the name `goerrors` . According to https://go.dev/blog/package-names underscores and mixed caps should be avoided in package names

support/db/metrics.go
```diff
@@ -356,6 +369,52 @@ func getQueryType(ctx context.Context, query squirrel.Sqlizer) QueryType {
...
+		// if getting here, no context deadline happened, but
+		// the db reported query_canceled, which leaves only the possibility of
+		// db-side statement timeout was triggered
+		abendType = "timeout"
+	}
+
+	s.abendCounter.With(prometheus.Labels{
+		"origin":    abendOrigin,
+		"condition": abendCondition,
+		"type":      abendType,
```
tamirms, line 414: I thought about this some more and I think we should avoid trying to derive `origin` and `type` because it is coupled to implementation details about how horizon terminates any context which can be connected to a db operation. If those implementation details change then these metric labels might not apply anymore.

For example, this code assumes that when the context is canceled that means the http client which is being served by horizon disconnected. But this is actually not true in all cases because when horizon shuts down it cancels the horizon context and any running db operations will terminate because of that context cancellation. In the future we might add other cases where we cancel contexts that are not related to the http client disconnecting from horizon.

Instead of trying to infer `origin` and `type` from db / context errors, I think we should try to report something closer to the raw error values with little interpretation. What do you think of the following proposal?

Instead of `origin`, `condition`, and `type`, we use the following labels: `ctx_error`, `db_error`, and `db_error_extra`.

- `ctx_error` -
  - `canceled` if `ctx.Err() == context.Canceled`
  - `deadline_exceeded` if `ctx.Err() == context.DeadlineExceeded`
  - `other` as a default catchall
- `db_error` - 
  - `pqErr.Code`  (e.g. 57014) if `dbErr` is a `*pq.Error` value
  - `"driver_bad_connection"` if `dbErr` is a `driver: bad connection` error 
  - `"other"` as a default catchall
- `db_error_extra` - this field can be used to add extra context to `db_error`. For now, we will populate this value to:
  -  `"user_request"` when `pqErr.Message == "canceling statement due to user request"`
  - `"statement_timeout"` when `pqErr.Message == "canceling statement due to statement timeout"`
  - `"n/a"` as a default catchall

If this proposal makes sense then I think we should also rename `abendCounter` to `errorCounter` since now the metric is focused on reporting db / context errors.

support/db/metrics.go
tamirms: @sreuland you're right, I think it is necessary to include the context error in these metrics. sorry for the back and forth on this topic. I made one (hopefully!) last proposal in https://github.com/stellar/go/pull/5225#discussion_r1508651977

support/db/session.go
sreuland: correct, the outcome/behavior on mapping of db errors to horizon errors has not been changed, and the tests in session_test I think assert the same horizon errors are being emitted. this is internal refactor to make it cleaner, the prior case on `"pq: canceling statement due to user request")` was redundant and like never triggered since the context deadline/cancel cases get hit first and those trigger libpq to generate that error string.

support/db/metrics.go
sreuland: @tamirms , I can proceed on this, I feel like need to add some tweak to the taxonomy to avoid overlapping semantics, based on libpq behavior for errors it emits and demonstrated in session_test.go cases:

* ctx_error -
  * "canceled" if goerrors.Is(caller's ctx.Err(), context.Canceled)
  * "deadline_exceeded" if goerrors.Is(caller's ctx.Err(), context.DeadlineExceeded)
  * "n/a" default if not cancel/deadline

* db_error -
  * the sql state error code, only available when libpsql performed a network trip to server and server emitted a sql state error code, in that case, libpq only emits a non-wrapped dbErr as *pq.Error value which has code=sql state error
  * <del>"canceled" if goerrors.Is(dbErr, context.Canceled)</del> will already be represented on the same metric key occurrence with label `ctx_error=canceled`
  * <del>"deadline_exceeded" if goerrors.Is(dbErr, context.DeadlineExceeded)</del> will already be represented on the same metric key occurrence with label `ctx_error=deadline_exceeded`
  * "n/a" when no sql state was available

* libpq_error -   
  * "driver_bad_connection" libpq emits a go Error immediately which contains "driver: bad connection", it doesn't make a network trip, caller requested to do something on a session which had been prior opened by libpq but later server closed the session for whatever reason and sent a signal back to libpq which then places a `syncErr` on the libpq session which gets checked locally by libpq before attempting network trip.
  * "n/a" for all other err's emitted by libpq calls, the err string has great data, but we can't stuff the opaque error string into metric label due to format limits, so, just need to truncate to "n/a
  
* error_extra - this field can be set with optional extra context for the metric. For now, we will populate this value to:
  * <del>"user_request" when pqErr.Message == "canceling statement due to user request"</del> - this is duplicate, it is already represented by `ctx_error=canceled`
  * <del>"statement_timeout" when pqErr.Message == "canceling statement due to statement timeout"</del> this is duplicate, it is already represented by `ctx_error!=nil` && `db_error=57014`, we could still emit "statement_timeout" for shorthand convenience/de-normalization in terms of querying for statement timeouts if desired?
  * "n/a" as a default catchall

let me know if this sounds inline with your expectation, can then proceed, thanks!

sreuland: @tamirms , I folded in the new metrics label model as discussed, your thoughts on finer grained, unique error messages from the db server being carried forward in the `pq.Error` from libpq for the same error code of 57014 was correct, so, was able to proceed with `db_error_extra` based on that. 

I recommend looking at session_test.go and lines with`assertDbErrorMetrics` to see how the final metrics output looks for the various error outcomes we have discussed, it lines up very closely to the [revised labels suggestion](https://github.com/stellar/go/pull/5225#discussion_r1508651977)

support/db/metrics.go
```diff
@@ -356,6 +369,53 @@ func getQueryType(ctx context.Context, query squirrel.Sqlizer) QueryType {
...
+		dbError = "canceled"
+	case errors.Is(dbErr, context.DeadlineExceeded):
+		dbError = "deadline_exceeded"
+	}
+
+	switch {
+	case errors.Is(ctx.Err(), context.Canceled):
+		ctxError = "canceled"
+	case errors.Is(ctx.Err(), context.DeadlineExceeded):
+		ctxError = "deadline_exceeded"
```
tamirms, line 409: could we add a case for nil to distinguish between when `ctx.Err()` is nil vs when `ctx.Err()` is non-nil and also not `context.Canceled` or `context.DeadlineExceeded`?

tamirms (approved):
looks great! I just left one comment about capturing the context error nil case https://github.com/stellar/go/pull/5225#discussion_r1512245336

support/db/metrics.go
sreuland: yes will do:
 nil="n/a", "canceled", "deadline_exceeded" and lastly "other" if it was non-nil but neither of those

tamirms: thanks, that sounds good!

sreuland: a little bit more investigation, the Context.Err() can only be set by invoking cancel func first, CancelWithCause allows an error to be included as `cancel(err)` , but that usage would still equate first to a 'canceled', so I don't think there is a possibility for an 'other' case, it will just be nil="n/a", or "canceled", or "deadline_exceeded"

tamirms: yes, you're right. the documentation for the context interface states that `Err()` returns either canceled or deadline exceeded if it's non-nil:

https://pkg.go.dev/context#Context

PR #5228, opened 2024-03-04
services/horizon: return null txmeta in api model when SKIP_TXMETA enabled

### LINKED ISSUE #5221: horizon: meta should be returned as null rather than empty when `SKIP_TXMETA` enabled
Original issue here https://github.com/stellar/go/issues/5189

Our original implementation of this had the API returning valid but empty metas. This seems to have been a miscommunication as the product request was for the API to return `null`. We should update the API responses to return null rather than empty metas.

sreuland:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

When SKIP_TXMETA is enabled the api resource model for `Transaction.result_meta_xdr` will be set to null.

### Why
Closes #5221
prior work for eliding tx meta on SKIP_TXMETA for context - https://github.com/stellar/go/pull/5208

### Known limitations
this will require a new horizon patch release 2.28.4
this will require a new go client sdk releast 0.16.0

To provide nullability on `Transaction.result_meta_xdr`, had to make breaking change on [Transaction resource model](https://github.com/stellar/go/pull/5228/files#diff-1fd45f9debb71699157f646e42df74d6bc214e6d16631c55507ec9158990b15eR9), this will require propagating the nullable aspect to other SDK's to handle potential null's.


protocols/horizon/main.go
```diff
@@ -518,7 +518,7 @@ type Transaction struct {
 	OperationCount    int32     `json:"operation_count"`
 	EnvelopeXdr       string    `json:"envelope_xdr"`
 	ResultXdr         string    `json:"result_xdr"`
-	ResultMetaXdr     string    `json:"result_meta_xdr"`
+	ResultMetaXdr     *string   `json:"result_meta_xdr"`
```
sreuland, line 521: to meet the requirements for null, need a pointer here can't use empty string.

tamirms: could you use an empty string with the omitempty tag?

```suggestion
	ResultMetaXdr     string    `json:"result_meta_xdr,omitempty"`
```

services/horizon/internal/resourceadapter/transaction.go
```diff
@@ -43,7 +43,11 @@ func PopulateTransaction(
 	dest.OperationCount = row.OperationCount
 	dest.EnvelopeXdr = row.TxEnvelope
 	dest.ResultXdr = row.TxResult
-	dest.ResultMetaXdr = row.TxMeta
+	if GetResourceAdapter().SkipTxmeta {
```
tamirms, line 46: I think it would be cleaner if you added a `skipTxMeta` parameter to the `PopulateTransaction()` function.  Then, you can add a `SkipTxMeta` field to the `GetTransactionByHashHandler` and `GetTransactionsHandler` structs in the actions package which will be propagated to the `resourceadapter.PopulateTransaction()` invocations

sreuland: ya, I was trying for a ioc/dependency injection approach, to avoid too much ripple up the stack, but I will go this pattern, will require adding RouterConfig.SkipTxMeta so the handler factories can get it, thx!

protocols/horizon/main.go
sreuland: yes, good suggestion, the absence of the field vs. the explicit null of the field has some semantic difference in json parsing, but for the requirements here of 'null', I'm not sure if maybe the absence of field could suffice for same meaning or be too strong? 

There seems to be precedence for omission, i.e. absence of other fields in here, so seems like could follow this pattern for `result_meta_xdr` and avoid the extra noise of compile time breaking change in the go client sdk, it would entail an empty string be present in the `Transaction.ResultMetaXdr` rather than nil, which is more just go/json aspect. I think given that, will move forward with using the `omitempty` and absence of field.

on the wire, the absence of the field is logically equivalent to the field being present and set to null, and still equates to an api breaking change for being an optional field in json parsing terms for any other sdk/client.

sreuland: @mollykarcher , do you see any concern with the criteria of null on meta for  https://github.com/stellar/go/issues/5221 if we make the api json tx meta field `result_meta_xdr` be absent instead of present but set to `result_meta_xdr:null`? as outlined in comments here, there's a slight nuance in the go client sdk for empty string representation in that case, but does avoid an overall breaking compile time change on the go client sdk.

services/horizon/internal/resourceadapter/transaction.go
sreuland: I redid the flag resolution from dependency injection to functional composition, interesting to see the difference in code/footprint between the patterns - https://github.com/stellar/go/pull/5228/commits/0e767ad005bf438ce3efa2a459081fd6b7365557, more boilerplate to propagate it through the call stack, not excessive, dependency injection can sometimes be viable as an option where component needs reference to another component or data which is many hops away from nearest reference.

tamirms (approved):

protocols/horizon/main.go
mollykarcher: I think that's alright. It gets at the same intent of, clients expecting it to be there and trying to grab it by key or something would break.

Issue #5230, opened 2024-03-04
update go version to 1.22


mollykarcher:
### What problem does your feature solve?

The current/most recent version of golang is 1.22, and we are currently testing and building with version 1.20. Golang will only issue security patches/releases for the two most recent versions of golang. So if any are released, we are missing out on patches for those due to the older version we're using.

### What would you like to see?

Update all relevant repos and built packages/images to use go 1.22
- [x] go monorepo (github CI workflows, Dockerfiles, go.mod, etc etc)
- [x] soroban-rpc (github CI workflows, Dockerfiles, go.mod, etc etc)
- [x] pipelines/package builders (https://github.com/stellar/pipelines) for both Horizon and RPC
- [x] quickstart https://github.com/stellar/quickstart

PR #5232, opened 2024-03-04
Bump go version to 1.22

urvisavla:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Bump go version to 1.22 

### Why

https://github.com/stellar/go/issues/5230

### Known limitations

N/A


sreuland:
@urvisavla , can you target this pr at new [2.28.4 rel branch](https://github.com/stellar/go/tree/release-horizon-v2.28.4)

.github/workflows/go.yml
```diff
@@ -18,7 +18,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-22.04]
-        go: ["1.21"]
+        go: ["1.22.1"]
```
sreuland, line 21: is there a reason to pin the minor ver`1.22.1` or can it be `1.22` ?

services/horizon/docker/verify-range/Dockerfile
```diff
@@ -1,5 +1,6 @@
 FROM ubuntu:22.04
 
+ARG GO_VERSION
```
sreuland, line 3: not seeing a reference to this ARG, can it be removed?

.github/workflows/go.yml
urvisavla: Like I mentioned in the slack I'm not entirely sure why that was required but this the "check" action was failing without it.

urvisavla:
> @urvisavla , can you target this pr at new [2.28.4 rel branch](https://github.com/stellar/go/tree/release-horizon-v2.28.4), thanks!

Sure if we're planning to get  2.28.4 out in a day or two; otherwise it might be better to include this in 2.28.3 considering it's related to a CVE?

PR #5244, opened 2024-03-13
services/horizon/internal: Add limit on number of horizon requests in flight

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add limit on number of horizon requests in flight.

### Why

We observed a case where a staging horizon node received too many requests which caused it to OOM. See https://stellarfoundation.slack.com/archives/C02B04RMK/p1710326856362359?thread_ts=1710235095.516249&cid=C02B04RMK for more details. 

### Known limitations

[N/A]

### DIFF
--- a/services/horizon/internal/flags.go
+++ b/services/horizon/internal/flags.go
@@ -69,8 +69,9 @@ const (
 	// StellarTestnet is a constant representing the Stellar test network
 	StellarTestnet = "testnet"
 
-	defaultMaxHTTPRequestSize = uint(200 * 1024)
-	clientQueryTimeoutNotSet  = -1
+	defaultMaxConcurrentRequests = 1000
+	defaultMaxHTTPRequestSize    = uint(200 * 1024)
+	clientQueryTimeoutNotSet     = -1
 )
 
 var (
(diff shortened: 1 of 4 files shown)


services/horizon/internal/flags.go
```diff
@@ -68,7 +68,8 @@ const (
 	// StellarTestnet is a constant representing the Stellar test network
 	StellarTestnet = "testnet"
 
-	defaultMaxHTTPRequestSize = uint(200 * 1024)
+	defaultMaxConcurrentRequests = 1000
```
mollykarcher, line 71: Do you know what the number was at when we OOMed?

mollykarcher (approved):

services/horizon/internal/flags.go
```diff
@@ -445,6 +446,15 @@ func Flags() (*Config, support.ConfigOptions) {
 			Usage:          "sets the limit on the maximum allowed http request payload size, default is 200kb, to disable the limit check, set to 0, only do so if you acknowledge the implications of accepting unbounded http request payload sizes.",
 			UsedInCommands: ApiServerCommands,
 		},
+		&support.ConfigOption{
+			Name:        "max-concurrent-requests",
+			ConfigKey:   &config.MaxConcurrentRequests,
+			OptType:     types.Uint,
+			FlagDefault: defaultMaxConcurrentRequests,
+			Usage: "sets the limit on the maximum number of concurrent http requests, default is 1000, to disable the limit set to 0. " +
```
sreuland, line 454: just for non-breaking backwards compatibility, should the default be `0` ?

sreuland (approved):
lgtm, left one suggestion for consideration on default value

services/horizon/internal/flags.go
tamirms: it was a little over 3000 when the instance OOMed as can be observed in this [graph](https://grafana.stellar-ops.com/explore?left=%7B%22datasource%22:%22Prometheus%22,%22queries%22:%5B%7B%22expr%22:%22sum%20by%20(instance)%20(horizon_http_requests_in_flight%7Benvironment%3D~%5C%22stg%5C%22,%20horizon_network%3D~%5C%22pubnet%5C%22,%20horizon_cluster%3D~%5C%22.*%5C%22%7D)%22,%22format%22:%22time_series%22,%22interval%22:%22%22,%22intervalFactor%22:1,%22refId%22:%22A%22,%22datasource%22:%7B%22type%22:%22prometheus%22,%22uid%22:%22000000001%22%7D%7D%5D,%22range%22:%7B%22from%22:%221710278724000%22,%22to%22:%221710279504000%22%7D%7D&orgId=1)

services/horizon/internal/flags.go
tamirms: I think the default limit is really high that it should never be observed during normal usage. so, although it is technically a breaking change, I think it would be better to have this default so we can protect against OOMs by default

Shaptic (approved):

services/horizon/internal/flags.go
Shaptic: looks like we get ~2k/rps on the whole cluster, so yeah 1k/machine should be way more than plenty

PR #5246, opened 2024-03-14
services/horizon/internal/ingest/processors: Fix bug in claimable balance change processor

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

In https://github.com/stellar/go/issues/5086 we made some optimizations to the claimable change processor based on the assumption that claimable balances are always created or removed but never updated. It turns out that assumption is incorrect because it's possible for the sponsor of a claimable balance to change. This commit corrects that assumption and allows the change processor to ingest updated claimable balance ledger entries. 

### Known limitations

[N/A]

### DIFF
--- a/services/horizon/internal/ingest/processors/claimable_balances_change_processor.go
+++ b/services/horizon/internal/ingest/processors/claimable_balances_change_processor.go
@@ -1,8 +1,10 @@
 package processors
 
 import (
+	"bytes"
 	"context"
 	"fmt"
+	"sort"
 
 	"github.com/stellar/go/ingest"
 	"github.com/stellar/go/services/horizon/internal/db2/history"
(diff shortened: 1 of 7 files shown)


leighmcculloch (approved):

services/horizon/internal/ingest/processors/claimable_balances_change_processor.go
```diff
@@ -112,6 +117,12 @@ func (p *ClaimableBalancesChangeProcessor) Commit(ctx context.Context) error {
 		return errors.Wrap(err, "error executing ClaimableBalanceBatchInsertBuilder")
 	}
 
+	if len(updatedBalances) > 0 {
+		if err = p.qClaimableBalances.UpsertClaimableBalances(ctx, updatedBalances); err != nil {
```
urvisavla, line 121: Do we also need to restore the BatchInsertBuilder to handle upserts for claimants? Like what happens if new claimants are added? Will we skip adding those to the claimants table?

tamirms: `UpsertClaimableBalances()` will upsert to both the claimable_balances and the  claimable_balance_claimants tables

tamirms: I updated the function in a later commit:

https://github.com/stellar/go/pull/5246/commits/672d637f6c13979c12da6b63446a2a2aa77515ed

PR #5250, opened 2024-03-15
services/horizon: Avoid combining SQL clauses when filtering is disabled

Shaptic:
### What
Stop adding the `UNION ALL` clause alongside a filter for transaction hashes in the `history_transactions_filtered_tmp` table when ingestion filtering is disabled.

Previously, the final SQL statement would look something like this:

```sql
ListOfPendingTxs = [ ... ]

SELECT `...[fields from ht and hl]...`
    FROM
        history_transactions ht LEFT JOIN history_ledgers hl 
        ON ht.ledger_sequence = hl.sequence 
    WHERE 
        (ht.ledger_sequence >= $1 AND 
        (ht.transaction_hash IN ListOfPendingTxs) OR 
        (ht.inner_transaction_hash ListOfPendingTxs))
UNION ALL 
SELECT `...[fields from ht and hl]...`
    FROM
        history_transactions_filtered_tmp ht LEFT JOIN history_ledgers hl 
        ON ht.ledger_sequence = hl.sequence 
    WHERE 
        (ht.ledger_sequence >= $1 AND 
        (ht.transaction_hash IN ListOfPendingTxs) OR 
        (ht.inner_transaction_hash ListOfPendingTxs))
```

whereas now it will only include the portion prior to the `UNION ALL` clause.

### Why
There's no reason to have the additional clause when the `history_transactions_filtered_tmp` table is guaranteed to be empty if the `EXP_ENABLE_INGESTION_FILTERING` flag is set to `false`.

### Known limitations
n/a

### DIFF
--- a/services/horizon/internal/db2/history/transaction.go
+++ b/services/horizon/internal/db2/history/transaction.go
@@ -30,21 +30,34 @@ func (q *Q) PreFilteredTransactionByHash(ctx context.Context, dest interface{}, 
 
 // TransactionsByHashesSinceLedger fetches transactions from `history_transactions_filtered_tmp`
 // table which match the given hash since the given ledger sequence (for perf reasons).
-func (q *Q) AllTransactionsByHashesSinceLedger(ctx context.Context, hashes []string, sinceLedgerSeq uint32) ([]Transaction, error) {
+func (q *Q) AllTransactionsByHashesSinceLedger(
+	ctx context.Context,
+	hashes []string,
+	sinceLedgerSeq uint32,
+	includeFiltered bool,
+) ([]Transaction, error) {
 	var dest []Transaction
 	innerOrOuterAndSeqGtEq :=
-		sq.And{sq.GtOrEq{"ht.ledger_sequence": sinceLedgerSeq}, sq.Or{sq.Eq{"ht.transaction_hash": hashes}, sq.Eq{"ht.inner_transaction_hash": hashes}}}
+		sq.And{
+			sq.GtOrEq{"ht.ledger_sequence": sinceLedgerSeq},
+			sq.Or{
+				sq.Eq{"ht.transaction_hash": hashes},
+				sq.Eq{"ht.inner_transaction_hash": hashes}}}
 
-	preFilteredTxs := selectTransactionPreFilteredTmp.Where(innerOrOuterAndSeqGtEq)
 	historyTxs := selectTransactionHistory.Where(innerOrOuterAndSeqGtEq)
+	query := historyTxs
 
-	preFilteredTxsString, args, err := preFilteredTxs.ToSql()
-	if err != nil {
-		return nil, errors.Wrap(err, "could not get string for un filtered sql query")
+	if includeFiltered {
+		preFilteredTxs := selectTransactionPreFilteredTmp.Where(innerOrOuterAndSeqGtEq)
+		preFilteredTxsString, args, err := preFilteredTxs.ToSql()
+		if err != nil {
+			return nil, errors.Wrap(err, "could not get string for un filtered sql query")
+		}
+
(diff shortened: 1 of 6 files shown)


services/horizon/internal/init.go
```diff
@@ -236,6 +236,7 @@ func initSubmissionSystem(app *App) {
 		DB: func(ctx context.Context) txsub.HorizonDB {
 			return &history.Q{SessionInterface: app.HorizonSession()}
 		},
-		LedgerState: app.ledgerState,
+		LedgerState:       app.ledgerState,
+		FilteringDisabled: !app.config.EnableIngestionFiltering,
```
tamirms, line 240: I think it would be easier to read if we had a `FilteringEnabled` field. That way you could pass in the field directly to `AllTransactionsByHashesSinceLedger()` without inverting it

services/horizon/internal/db2/history/transaction.go
```diff
@@ -30,21 +30,34 @@ func (q *Q) PreFilteredTransactionByHash(ctx context.Context, dest interface{},
 
 // TransactionsByHashesSinceLedger fetches transactions from `history_transactions_filtered_tmp`
 // table which match the given hash since the given ledger sequence (for perf reasons).
-func (q *Q) AllTransactionsByHashesSinceLedger(ctx context.Context, hashes []string, sinceLedgerSeq uint32) ([]Transaction, error) {
+func (q *Q) AllTransactionsByHashesSinceLedger(
```
tamirms, line 33: could you add a test which exercises both code paths where `includeFiltered` is true / false

services/horizon/internal/init.go
sreuland: the [ingestion filtering enabled config flag is true by default](https://github.com/Shaptic/go/blob/no-union-all/services/horizon/internal/flags.go#L280), it's marked deprecated, won't be override-able soon. for most deployments, it's probably not being set, and defaulting to true. 

may need to continue to check the config flag for now but also check that at least one filter rule is enabled to determine if effectively disabled, can probably wrap that as new method on the [QFilter interface](https://github.com/Shaptic/go/blob/no-union-all/services/horizon/internal/db2/history/filter_rules.go#L31):

```
func (q *Q) IsFilteringEffective(ctx context.Context) (bool, error) {
	accountFilter, err := q.GetAccountFilterConfig(ctx)
	if err != nil {
		return false, err
	}

	assetFilter, err := q.GetAssetFilterConfig(ctx)
	if err != nil {
		return false, err
	}

	return accountFilter.Enabled || assetFilter.Enabled, nil
}

```

in db migrations, [both rules were seeded in the db as disabled by default](https://github.com/Shaptic/go/blob/no-union-all/services/horizon/internal/db2/schema/migrations/55_filter_rules.sql#L16-L17).

tamirms: if we're going to deprecate the flag and always enable filtering then I think we should leave the pending transactions query as is because, the over head of having to query the db on whether filtering rules are present is probably more inefficient than getting rid of the union in the pending transactions query. There is also the problem that querying the db on whether filtering rules are present is still vulnerable to a race condition where filtering rules are added right after we query the db.

sreuland: yes, agree on that summary, 👍

Shaptic: ah wow, I didn't realize we were deprecating the flag entirely! in that case there's no benefit to having this piece of code, since as you said @tamirms we'd need to actively query for filters to avoid a(n empty) subquery which defeats the purpose. thanks for the discussion!

Shaptic:
counterproductive per the above discussion: https://github.com/stellar/go/pull/5250#discussion_r1532984343

PR #5252, opened 2024-03-20
services/horizon/internal/ingest: Remove unnecessary use of ChangeCompactor to reduce memory bloat during ingestion

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Part of #5258 

Horizon users have reported that state rebuilds have caused Horizon's memory usage to exceed 32 GB. After profiling Horizon we discovered that most memory is being retained in the `ChangeCompactor`:

profile001

The `ChangeCompactor` is described in detail here https://github.com/stellar/go/blob/master/ingest/change_compactor.go#L49 . It turns out that there is no need to use the `ChangeCompactor`when ingesting from the history archives because all ledger entries changes obtained from the history archives are of the created type which means that we will never compact any ledger entries.

By removing the `ChangeCompactor` from all the state processors we avoid creating additional copies of ledger entries obtained from the history archives. This has resulted in a substantial reduction in memory usage. Previously we used ~32 GB of RAM to do a state rebuild, now, we are able to do a state rebuild in ~5G:

Screenshot 2024-03-20 at 5 02 50 PM

The only place where it makes sense to use a `ChangeCompactor` is when processing ledger entry changes extracted from a `LedgerCloseMeta` payload. It is possible for a ledger entry to be created and then modified or removed within the same ledger. 

### Known limitations

[N/A]


ire-and-curses:
> The only place where it makes sense to use a ChangeCompactor is when processing ledger entry changes extracted from a LedgerCloseMeta payload. It is possible for a ledger entry to be created and then modified or removed within the same ledger.

It's probably worth adding a comment to ChangeCompactor saying something like this, to prevent incorrect usage in future.

services/horizon/internal/ingest/processor_runner.go
```diff
@@ -308,8 +308,29 @@ func (s *ProcessorRunner) runChangeProcessorOnLedger(
 		logFrequency,
 		s.logMemoryStats,
 	)
-	if err = processors.StreamChanges(s.ctx, changeProcessor, changeReader); err != nil {
-		return errors.Wrap(err, "Error streaming changes from ledger")
+
+	compactor := ingest.NewChangeCompactor()
+	for {
```
sreuland, line 313: for consideration, rather than having compactor specifics, ln 313-333, could that block be refactored into a new  `ChangeCompactor.Apply(changeReader, changeProcessor)` ?

services/horizon/internal/ingest/processors/asset_stats_processor.go
```diff
@@ -68,11 +79,14 @@ func (p *AssetStatsProcessor) ProcessChange(ctx context.Context, change ingest.C
...
+		p.contractDataChanges = append(p.contractDataChanges, change)
+	case xdr.LedgerEntryTypeTtl:
+		err = p.addExpirationChange(change)
+	default:
+		return errors.Errorf("Change type %v is unexpected", change.Type)
 	}
-	if err := p.cache.AddChange(change); err != nil {
-		return errors.Wrap(err, "error adding to ledgerCache")
-	}
-	return nil
```
sreuland, line 75: I noticed this doesn't have a maxBatchSize to check after each change, which intermittently purges RAM like the other change processors, it would be holding larger cache and longer on RAM across all tx changes for a ledger, is it worthwhile to include?

services/horizon/internal/ingest/processors/account_data_processor.go
```diff
@@ -23,8 +24,9 @@ func NewAccountDataProcessor(dataQ history.QData) *AccountDataProcessor {
 }
 
 func (p *AccountDataProcessor) reset() {
-	p.cache = ingest.NewChangeCompactor()
 	p.batchInsertBuilder = p.dataQ.NewAccountDataBatchInsertBuilder()
+	p.dataToUpsert = []history.Data{}
```
sreuland, line 28: very interesting, this solution applied on every change processor, will still take RAM per change, but stores a more specific application struct derived from `ingest.Change` rather than the `ingest.Change` itself which holds two `LedgerEntry`, and that difference alone is what accounts for the bulk of memory reduction?

services/horizon/internal/ingest/processors/signers_processor.go
```diff
@@ -36,37 +31,73 @@ func (p *SignersProcessor) Name() string {
...
+		accountAddress := change.Post.Data.MustAccount().AccountId.Address()
+		if err := p.addAccountSigners(accountAddress, signerSummary, sponsersPerSigner); err != nil {
 			return err
 		}
-	} else {
-		return errors.New("SignersProcessor is in insert only mode")
+	}
+
+	if p.batchInsertBuilder.Len() > maxBatchSize {
+		if err := p.Commit(ctx); err != nil {
```
sreuland, line 98: using maxBatchSize is new behavior for this processor, do any test cases in signer_processor_test.go exercise this new path by default to verify?

sreuland (approved):
lgtm, great write-up as well on problem/solution in pr description, left some minor comments for consideration.

services/horizon/internal/ingest/processors/asset_stats_processor.go
tamirms: yeah, unfortunately the asset stats processor cannot be batched because for the every contract data asset balance we need to know its corresponding expiration ledger entry. There is no guarantee about the ordering of ledger entry changes we process while ingesting from the history archives. So in the worst case it is possible that the first ledger entry change we process is a contract asset balance and the very last ledger entry change we process is the contract asset balance's expiration ledger entry. This means we need to hold all the contract data asset balances and all the expiration ledger entries in memory until we have finished processing all ledger entry changes

services/horizon/internal/ingest/processors/account_data_processor.go
tamirms: during state rebuilds we only process created ledger entry changes so we can ignore the update case. The reason why this PR is effective in reducing memory is because the change compactor will allocate a copy of the ledger key and store it:

https://github.com/stellar/go/blob/master/ingest/change_compactor.go#L87-L98

This allocation is also amplified by the fact that every processor was having its own dedicated change compactor. Consider that the accounts processor, accounts data processor, and the account signers processor all consume account ledger entry changes. That means, during a state rebuild, up to `3 * maxBatchSize` account ledger keys are getting stored by the change compactors for the processors which operate on account ledger entry changes.

In this PR, we remove the change compactor entirely from state rebuilds so we are no longer storing any copy of the ledger keys.

services/horizon/internal/ingest/processors/signers_processor.go
tamirms: the `removed` and `signerSummary` entries are immediately applied in the processor. They are not persisted beyond the function call. Only the entries added to `p.batchInsertBuilder` are accumulated so that is why we only check `p.batchInsertBuilder.Len()` against `maxBatchSize`.

PR #5253, opened 2024-03-20
services/horizon/internal/ingest: Add missing tables to TruncateIngestStateTables()

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

The `contract_asset_balances` and `contract_asset_stats` tables were missing from `TruncateIngestStateTables()` which means that those tables would not get cleared during state rebuilds. It is necessary for all tables which hold ledger entries to be truncated in `TruncateIngestStateTables()` otherwise state rebuilds may fail because Horizon can try to insert rows that already exist in the state tables which would result in a duplicate key errors from postgres.

### Known limitations

[N/A]

### DIFF
--- a/services/horizon/internal/db2/history/ingestion.go
+++ b/services/horizon/internal/db2/history/ingestion.go
@@ -17,6 +17,8 @@ func (q *Q) TruncateIngestStateTables(ctx context.Context) error {
 		"claimable_balances",
 		"claimable_balance_claimants",
 		"exp_asset_stats",
+		"contract_asset_balances",
+		"contract_asset_stats",
 		"liquidity_pools",
 		"offers",
 		"trust_lines",
--- a/services/horizon/internal/ingest/verify_test.go
+++ b/services/horizon/internal/ingest/verify_test.go
@@ -283,6 +283,32 @@ func ttlForContractData(tt *test.T, gen randxdr.Generator, contractData xdr.Ledg
 	return ttl
 }
 
+func TestTruncateIngestStateTables(t *testing.T) {
+	tt := test.Start(t)
+	defer tt.Finish()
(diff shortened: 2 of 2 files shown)


services/horizon/internal/db2/history/ingestion.go
```diff
@@ -17,6 +17,8 @@ func (q *Q) TruncateIngestStateTables(ctx context.Context) error {
 		"claimable_balances",
 		"claimable_balance_claimants",
 		"exp_asset_stats",
+		"contract_asset_balances",
+		"contract_asset_stats",
```
Shaptic, line 21: is there any way for us to catch forgetting to add tables here going forward?

services/horizon/internal/ingest/verify_test.go
```diff
@@ -402,3 +399,23 @@ func TestStateVerifier(t *testing.T) {
...
+			genClaimableBalance(tt, gen),
+			genOffer(tt, gen),
+			genTrustLine(tt, gen),
+			genAccount(tt, gen),
+			genAccountData(tt, gen),
+			genContractCode(tt, gen),
+			genConfigSetting(tt, gen),
+			genTTL(tt, gen),
+		)
+		changes = append(changes, genAssetContractMetadata(tt, gen)...)
```
Shaptic, line 418: nit: why isn't this part of the above `append` line? can you not mix with the `...` operator?

Shaptic (approved):

services/horizon/internal/ingest/verify_test.go
tamirms: I tried that but it doesn't seem possible in Go https://github.com/golang/go/issues/18605

PR #5254, opened 2024-03-21
exp/services/ledgerexporter: Implement Docker based deployment

urvisavla:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

- Created a Dockerfile to build the Ledger Exporter image. 
    - Currently, there is no versioning scheme for the Ledge Exporter, so the Docker image is created with the `latest` tag. 
    - Command line parameters such as `start`, `end`, and `from_last` are passed as env variables when running the container. For configuring the Captive Core, the Ledger Exporter only accepts the `NETWORK` configuration parameter which can be set to 'testnet' or 'pubnet'. The destination URL is provided as `ARCHIVE_TARGET` env variable and it is currently limited to GCS only. These config params are also passed as env variables in Docker.
  For example:
```sh     
docker run --platform linux/amd64 -t -v "$HOME/.config/gcloud/application_default_credentials.json":/gcp/creds.json:ro -e GOOGLE_APPLICATION_CREDENTIALS=/gcp/creds.json -e NETWORK=testnet -e ARCHIVE_TARGET=gcs://exporter-test -e START=100  -e END=200 stellar/ledger-exporter:latest
```

Since there is no CLI support for config params in the Ledger Exporter, a startup script creates a configuration file based on the provided parameters. 
There are two additional optional configuration params, 'ledgers_per_file' and 'files_per_partition', which default to 1 and 64000. These can be overridden by passing the corresponding env variables when running the docker container.

- Created a Makefile with the following commands:
    1. `docker-build`: Builds a Docker image from the Dockerfile.
    2. `docker-test`: Tests the Ledger Exporter Docker image for a specific range of ledgers using the GCP emulator for storage. It spins up two containers: one for the GCP emulator and the other for the Ledger Exporter.
    3. `docker-push`: Pushes the Docker image to the Stellar Docker registry.

- The above commands are integrated into the `horizon.yml` GitHub workflow. 
    - The first two are executed on pull requests and `docker-push` is executed on merging into the master.

### Why

https://stellarorg.atlassian.net/browse/HUBBLE-270

### Known limitations
No deb package, only Docker.
Doesn't have versioning. Each push to the master will update the Docker image.

### DIFF
--- a/exp/services/ledgerexporter/Makefile
+++ b/exp/services/ledgerexporter/Makefile
@@ -3,22 +3,16 @@ SUDO := $(shell docker version >/dev/null 2>&1 || echo "sudo")
 # https://github.com/opencontainers/image-spec/blob/master/annotations.md
 BUILD_DATE := $(shell date -u +%FT%TZ)
 VERSION ?= $(shell git rev-parse --short HEAD)
-TAG ?= stellar/ledger-exporter:$(VERSION)
+DOCKER_IMAGE := stellar/ledger-exporter
 
 docker-build:
-ifndef STELLAR_CORE_VERSION
 	cd ../../../ && \
 	$(SUDO) docker build --platform linux/amd64 --pull --label org.opencontainers.image.created="$(BUILD_DATE)" \
 	--build-arg VERSION=$(VERSION) \
+$(if $(STELLAR_CORE_VERSION), --build-arg STELLAR_CORE_VERSION=$(STELLAR_CORE_VERSION)) \
 	-f exp/services/ledgerexporter/docker/Dockerfile \
-	-t $(TAG) .
-else
-	cd ../../../ && \
-	$(SUDO) docker build --platform linux/amd64 --pull --label org.opencontainers.image.created="$(BUILD_DATE)" \
-	--build-arg VERSION=$(VERSION) --build-arg STELLAR_CORE_VERSION=$(STELLAR_CORE_VERSION) \
-	-f exp/services/ledgerexporter/docker/Dockerfile \
-	-t $(TAG) .
-endif
+	-t $(DOCKER_IMAGE):$(VERSION) \
+	-t $(DOCKER_IMAGE):latest .
 
 docker-test:
 	# Create temp storage dir
@@ -38,7 +32,7 @@ docker-test:
 		 -e START=1000 \
 		 -e END=2000 \
 		 -e STORAGE_EMULATOR_HOST=http://fake-gcs-server:4443 \
-		 $(TAG)
+		 $(DOCKER_IMAGE):$(VERSION)
 
 	# Cleanup
 	$(SUDO) docker stop fake-gcs-server
@@ -47,4 +41,5 @@ docker-test:
 	$(SUDO) docker network rm test-network
 
 docker-push:
-	$(SUDO) docker push $(TAG)
+	$(SUDO) docker push $(DOCKER_IMAGE):$(VERSION)
+	$(SUDO) docker push $(DOCKER_IMAGE):latest
--- a/exp/services/ledgerexporter/Makefile
+++ b/exp/services/ledgerexporter/Makefile
@@ -1,0 +1,25 @@
+SUDO := $(shell docker version >/dev/null 2>&1 || echo "sudo")
+
+# https://github.com/opencontainers/image-spec/blob/master/annotations.md
+BUILD_DATE := $(shell date -u +%FT%TZ)
+VERSION ?= latest
+TAG ?= stellar/ledger-exporter:$(VERSION)
+
+docker-build:
+ifndef STELLAR_CORE_VERSION
+	cd ../../../ && \
+	$(SUDO) docker build --pull --label org.opencontainers.image.created="$(BUILD_DATE)" \
+	--build-arg VERSION=$(VERSION) \
+	-f exp/services/ledgerexporter/docker/Dockerfile \
+	-t $(TAG) .
+else
+	cd ../../../ && \
+	$(SUDO) docker build --pull --label org.opencontainers.image.created="$(BUILD_DATE)" \
+	--build-arg VERSION=$(VERSION) --build-arg STELLAR_CORE_VERSION=$(STELLAR_CORE_VERSION) \
+	-f exp/services/ledgerexporter/docker/Dockerfile \
+	-t $(TAG) .
+endif
+
+docker-push:
+	cd ../../../ && \
+	$(SUDO) docker push $(TAG)
--- a/exp/services/ledgerexporter/Makefile
+++ b/exp/services/ledgerexporter/Makefile
@@ -20,6 +20,32 @@ else
 	-t $(TAG) .
 endif
 
+docker-test:
+	# Create temp storage dir
+	$(SUDO) mkdir -p ${PWD}/storage/exporter-test
+
+	# Create test network for docker
+	$(SUDO) docker network create test-network
+
+	# Run the fake GCS server
+	$(SUDO) docker run -d --name fake-gcs-server -p 4443:4443 \
+		 -v ${PWD}/storage:/data --network test-network fsouza/fake-gcs-server -scheme http
+
+	# Run the ledger-exporter
+	$(SUDO) docker run --platform linux/amd64 -t --network test-network\
+		 -e NETWORK=pubnet \
+		 -e ARCHIVE_TARGET=gcs://exporter-test \
+		 -e START=1000 \
+		 -e END=2000 \
+		 -e STORAGE_EMULATOR_HOST=http://fake-gcs-server:4443 \
+		 stellar/ledger-exporter
+
+	# Cleanup
+	$(SUDO) docker stop fake-gcs-server
+	$(SUDO) docker rm fake-gcs-server
+	$(SUDO) rm -rf ${PWD}/storage
+	$(SUDO) docker network rm test-network
+
 docker-push:
 	cd ../../../ && \
 	$(SUDO) docker push $(TAG)
--- a/exp/services/ledgerexporter/Makefile
+++ b/exp/services/ledgerexporter/Makefile
@@ -2,19 +2,19 @@ SUDO := $(shell docker version >/dev/null 2>&1 || echo "sudo")
 
 # https://github.com/opencontainers/image-spec/blob/master/annotations.md
 BUILD_DATE := $(shell date -u +%FT%TZ)
-VERSION ?= latest
+VERSION ?= $(shell git rev-parse --short HEAD)
 TAG ?= stellar/ledger-exporter:$(VERSION)
 
 docker-build:
 ifndef STELLAR_CORE_VERSION
 	cd ../../../ && \
-	$(SUDO) docker build --pull --label org.opencontainers.image.created="$(BUILD_DATE)" \
+	$(SUDO) docker build --platform linux/amd64 --pull --label org.opencontainers.image.created="$(BUILD_DATE)" \
 	--build-arg VERSION=$(VERSION) \
 	-f exp/services/ledgerexporter/docker/Dockerfile \
 	-t $(TAG) .
 else
 	cd ../../../ && \
-	$(SUDO) docker build --pull --label org.opencontainers.image.created="$(BUILD_DATE)" \
+	$(SUDO) docker build --platform linux/amd64 --pull --label org.opencontainers.image.created="$(BUILD_DATE)" \
 	--build-arg VERSION=$(VERSION) --build-arg STELLAR_CORE_VERSION=$(STELLAR_CORE_VERSION) \
 	-f exp/services/ledgerexporter/docker/Dockerfile \
 	-t $(TAG) .
@@ -38,7 +38,7 @@ docker-test:
 		 -e START=1000 \
 		 -e END=2000 \
 		 -e STORAGE_EMULATOR_HOST=http://fake-gcs-server:4443 \
-		 stellar/ledger-exporter
+		 $(TAG)
 
 	# Cleanup
 	$(SUDO) docker stop fake-gcs-server
@@ -47,5 +47,4 @@ docker-test:
 	$(SUDO) docker network rm test-network
 
 docker-push:
-	cd ../../../ && \
 	$(SUDO) docker push $(TAG)
--- a/exp/services/ledgerexporter/Makefile
+++ b/exp/services/ledgerexporter/Makefile
@@ -34,7 +34,6 @@ docker-test:
 		 -e STORAGE_EMULATOR_HOST=http://fake-gcs-server:4443 \
 		 $(DOCKER_IMAGE):$(VERSION)
 
-	# Cleanup
 	$(SUDO) docker stop fake-gcs-server
 	$(SUDO) docker rm fake-gcs-server
 	$(SUDO) rm -rf ${PWD}/storage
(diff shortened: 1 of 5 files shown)


exp/services/ledgerexporter/docker/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+ENV STELLAR_CORE_VERSION=${STELLAR_CORE_VERSION:-*}
+ENV STELLAR_CORE_BINARY_PATH /usr/bin/stellar-core
+
+ENV DEBIAN_FRONTEND=noninteractive
+# ca-certificates are required to make tls connections
+RUN apt-get update && apt-get install -y --no-install-recommends ca-certificates curl wget gnupg apt-utils
+RUN wget -qO - https://apt.stellar.org/SDF.asc | APT_KEY_DONT_WARN_ON_DANGEROUS_USAGE=true apt-key add -
+RUN echo "deb https://apt.stellar.org focal stable" >/etc/apt/sources.list.d/SDF.list
+RUN echo "deb https://apt.stellar.org focal unstable" >/etc/apt/sources.list.d/SDF-unstable.list
+RUN apt-get update && apt-get install -y stellar-core=${STELLAR_CORE_VERSION}
```
chowbao, line 25: Is there an advantage to doing an install of stellar-core?
For stellar-etl we kinda cheat and just use the image built from core/ops team
```
FROM stellar/stellar-core:20.2.0-1716.rc3.34d82fc00.focal

WORKDIR /etl

COPY --from=build /usr/local/bin/stellar-etl /usr/local/bin/stellar-etl
COPY --from=build /usr/src/etl/docker docker
```

urvisavla: The only advantage I can think of is that we won't have to update our Dockerfile when the core gets an update.

exp/services/ledgerexporter/Makefile
```diff
@@ -0,0 +1,51 @@
+SUDO := $(shell docker version >/dev/null 2>&1 || echo "sudo")
+
+# https://github.com/opencontainers/image-spec/blob/master/annotations.md
+BUILD_DATE := $(shell date -u +%FT%TZ)
+VERSION ?= latest
```
chowbao, line 5: Because we don't have a defined version we want to use right now. Would it make sense to push both a :latest and `git rev-parse --short HEAD` as a temporary way to version control?

exp/services/ledgerexporter/Makefile
```diff
@@ -0,0 +1,51 @@
...
+	-t $(TAG) .
+else
+	cd ../../../ && \
+	$(SUDO) docker build --pull --label org.opencontainers.image.created="$(BUILD_DATE)" \
+	--build-arg VERSION=$(VERSION) --build-arg STELLAR_CORE_VERSION=$(STELLAR_CORE_VERSION) \
+	-f exp/services/ledgerexporter/docker/Dockerfile \
+	-t $(TAG) .
+endif
+
+docker-test:
```
tamirms, line 23: I think this way of testing is fine for now. But, we may want eventually want to adopt the same integration testing style as horizon and soroban rpc where we setup a docker-compose file with the dependencies (in this case the fake gcs service) and launch the service in a go test. Once https://github.com/stellar/go/pull/5260 is merged, we can also rerun captive core over the same range and assert that the ledgers exported to the fake GCS service match what is returned by captive core.

tamirms (approved):
I agree with @chowbao's comment for including the git sha in the docker image tag. Apart from that, this looks good!

exp/services/ledgerexporter/Makefile
```diff
@@ -0,0 +1,51 @@
+SUDO := $(shell docker version >/dev/null 2>&1 || echo "sudo")
+
+# https://github.com/opencontainers/image-spec/blob/master/annotations.md
+BUILD_DATE := $(shell date -u +%FT%TZ)
+VERSION ?= latest
+TAG ?= stellar/ledger-exporter:$(VERSION)
+
+docker-build:
+ifndef STELLAR_CORE_VERSION
+	cd ../../../ && \
```
sreuland, line 10: minor, could do the directory change to set initial docker context path once before `ifndef`

exp/services/ledgerexporter/Makefile
```diff
@@ -0,0 +1,51 @@
...
+
+docker-test:
+	# Create temp storage dir
+	$(SUDO) mkdir -p ${PWD}/storage/exporter-test
+
+	# Create test network for docker
+	$(SUDO) docker network create test-network
+
+	# Run the fake GCS server
+	$(SUDO) docker run -d --name fake-gcs-server -p 4443:4443 \
```
sreuland, line 31: nice!

exp/services/ledgerexporter/docker/start
```diff
@@ -0,0 +1,44 @@
...
+network = "${NETWORK}"
+destination_url = "${ARCHIVE_TARGET}"
+
+[exporter_config]
+  ledgers_per_file = $ledgers_per_file
+  files_per_partition = $files_per_partition
+EOF
+
+# Check if START or END variables are set
+if [[ -n "$START" || -n "$END" ]]; then
```
sreuland, line 34: could this script just pass all parameters to ledgerexporter bin and let [validateAndSetLedgerRange()](https://github.com/urvisavla/go/blob/ledger-exporter-packaging/exp/services/ledgerexporter/internal/config.go#L88) sort it out and report error if needed, rather than replicate similar validation in this layer also?

sreuland (approved):
nice addition, left minor comments for consideration.

chowbao (approved):

exp/services/ledgerexporter/docker/start
urvisavla: Good point. Done.

exp/services/ledgerexporter/Makefile
urvisavla: Done.

exp/services/ledgerexporter/Makefile
urvisavla: Done.

jacekn (changes requested):
This will create divergence and inconsistency between how horizon is being build and published and how ledgerexporter would be. For horizon we support both debs and docker images to ease of deployment and I think we should be consistent that that approach given that one day we'll ask people to replace their full history horizon with ledgerexporter.
Ledgerexporter is a golang app so it's super easy to use build process consistent with horizon, I can help you with that.

jacekn:
@urvisavla something else I wanted to add - this PR would make it impossible to people to cryptographically verify binaries coming from SDF. This is one of the big security wins of deb packages and this benefit propagates to docker images build from those debs. For smaller projects we accepted the security tradeoff but for something at the core of ingestions, such as this project, this tradeoff might be unacceptable.

mollykarcher:
> this PR would make it impossible to people to cryptographically verify binaries coming from SDF

@jacekn can you tell me more about this? As in, is this possible today and how? We actually have a couple partners (at minimum, Coinbase and Circle) that have chosen not to use our debian packages (and instead build from source), due to their own internal compliance standards and not wanting to use pre-built binaries.

urvisavla:
> This will create divergence and inconsistency between how horizon is being build and published and how ledgerexporter would be. For horizon we support both debs and docker images to ease of deployment and I think we should be consistent that that approach given that one day we'll ask people to replace their full history horizon with ledgerexporter. Ledgerexporter is a golang app so it's super easy to use build process consistent with horizon, I can help you with that.

That's a valid concern. The deb package is absent in this PR because this is the first iteration and it will be added in a [later iteration.](https://stellarorg.atlassian.net/browse/HUBBLE-369) We can update the Dockerfile later to install from the deb package instead of building it from source. 

The reason the deb package is not implemented in the first iteration is because we're prioritizing the creation of a Docker image and putting it somewhere the data team can access and use it to test it. This is defined in the [ticket](https://stellarorg.atlassian.net/browse/HUBBLE-270). 

>Something else I wanted to add - this PR would make it impossible to people to cryptographically verify binaries coming from SDF. This is one of the big security wins of deb packages and this benefit propagates to docker images build from those debs. For smaller projects we accepted the security tradeoff but for something at the core of ingestions, such as this project, this tradeoff might be unacceptable.

If security around building from source in the first iteration is a major concern, can we push the image to a private registry for internal use only? This may allow us to make progress without publishing it to the public registry.

exp/services/ledgerexporter/docker/start
```diff
@@ -0,0 +1,39 @@
+#! /usr/bin/env bash
```
jacekn, line 1: This wrapper adds complexity and will require maintenance. I think would be good to remove it and instead just call ledgerexporter directly for simplicity.

With this wrapper it's not easy to add arguments to the ledgerexporter call - users would need to bypass the script completely and use custom entrypoint but this requires deeper understanding of how the image is constructed.
Also part of the script's functionality is to render config on disk from env variables. This normally isn't necessary in docker or k8s world - it's very easy to consume configs both with docker and in k8s (via ConfigMaps). Rendering config like this also means that support for every new config option needs to be baked into the image before those options can be used which would add maintenance burden. On top of that users who decide to use ConfigMaps in k8s will see the entrypoint try to write data to the config.toml file which might clash with the location for the config they choose.

The wrapper does two main things:
1. It passes arguments to the ledgerexporter call
2. It renders config
 
To address point 1 we should be able to rely on docker and k8s directly. Both make it very easy to pass arguments to the images. For example with the horizon image the following work as expected:
```
docker run stellar/stellar-horizon:stable --help
docker run stellar/stellar-horizon:stable version
```

We should be able to make the following work with the ledgerexporter image:
```
docker run stellar/ledgerexporter --start 10 --end 20
```

I can think of two solutions to address item 2.
First option is to just rely on users providing config. In k8s we'd require this via ConfigMap and for docker using volume mount. This will require no changes to ledgerexporter at all.

If we want to add flexibility we could also update ledgerexporter to allow it to consume environment variables thus not require config at all. We only set the following: `NETWORK` , `ARCHIVE_TARGET`, `ledgers_per_file` and  `files_per_partition`. If we do this we will of course still retain capability to use ConfigMaps so this would be completely optional method.

jacekn:
> > This will create divergence and inconsistency between how horizon is being build and published and how ledgerexporter would be. For horizon we support both debs and docker images to ease of deployment and I think we should be consistent that that approach given that one day we'll ask people to replace their full history horizon with ledgerexporter. Ledgerexporter is a golang app so it's super easy to use build process consistent with horizon, I can help you with that.
> 
> That's a valid concern. The deb package is absent in this PR because this is the first iteration and it will be added in a [later iteration.](https://stellarorg.atlassian.net/browse/HUBBLE-369) We can update the Dockerfile later to install from the deb package instead of building it from source.

Thank you for context, this sounds fair to me. I can also see that the Dockerfile is Ubuntu based which will make it easy to switch method of binary install.

I added a few more comments. If we use different binary path we will be able to make it even easier to switch to debs in the future because path wouldn't change.

I also added comment about the `start` wrapper, I think we may be able to remove it to simplify the image.

exp/services/ledgerexporter/docker/start
jacekn: > We haven't yet implemented the "CLI" yet, which will include both the command line flags as well as environment variables to configure the Ledger Exporter, similar to how it's done in Horizon. Here is the [ticket](https://stellarorg.atlassian.net/browse/HUBBLE-271) for that.
> 
> The start script has been added to simplify testing the image for the team so that they can specify the configuration like the `NETWORK` and `ARCHIVE_TARGET` without hardcoding them or manually modifying the container's config file, until the CLI/config is ready.

Could you describe workflow that you think will be used during testing? I'm wondering if there are some assumptions I'm making because I _think_ using config file should be very simple. If you create `./config/config.toml` then something like this should work:
```
docker run -v ./config:/etc/ledgerexporter/ stellar/ledgerexporter --start 10 --end 20
```

if you use env variables you presumably need to run something like this:
```
docker run --env-file ./settings.env stellar/ledgerexporter
```

In either case there is no need to hardcode anything or modify anything inside container when settings change.

The only difference I can see is that the first method doesn't allow you to edit config on the fly using env variables passed as docker CLI flags but this can be easily achieved using simple wrapper script run outside of docker. In fact you should be able to take the `start` script and modify it to write config on disk before running docker to achieve the same result without baking this logic inside the image.
This way you'd keep the docker image closer to best practices and there would be no need to modify it when ledgerexporter CLI is changed.

> Regarding configuration options, we're also planning to add additional command line options such as 'resume.' Plus, we'll be adding configuration options to set up captive-core in the traditional manner, using settings like `CAPTIVE_CORE_CONFIG_PATH` and `CAPTIVE_CORE_BINARY_PATH` instead of relying on the default configuration associated with the "NETWORK" parameter. Here is the [ticket](HUBBLE-278: https://stellarorg.atlassian.net/browse/HUBBLE-27) for that.
> So while we currently have a basic CLI that takes `start` and `end` parameters from the command line which will allow a command like `docker run stellar/ledgerexporter --start 10 --end 20`, I'd rather hold off on making changes until we have completed the CLI/config implementation. It might be more efficient to implement all the changes at once.

So what I'm proposing means that you will not have to change the image at all when you make CLI changes to ledgerexporter. The only place where you'd be adding/changing/removing flags would be the ledgerexporter itself.
If it starts using env variables natively then you'll be able to pass those to the container and they'll be available for ledgerexporter to read and act on. If you add new CLI flags you'll be able to pass them to the image without need to change any wrappers:
```
docker run -v ./config:/etc/ledgerexporter/ stellar/ledgerexporter --start 10 --end 20 --new-flag foo
# or
docker run -v ./config:/etc/ledgerexporter/ stellar/ledgerexporter resume
```
if you change any default values for CLI flags those will take effect as soon as you rebuild the image without the need to edit the `start` wrapper. For the most part the Docker packaging config will become maintenance free, you'll only have to worry about ledgerexporter itself.

urvisavla: Once the cli is complete, the config file and the start script won't be needed anymore. The current method of wrapping the config is simpler for testing purpose because it doesn't require the data team to be familiar with the config file format.  Also, once the "start" is removed, there is no other way to run the container than by providing ledger exporter command line parameters in the docker run command.

jacekn: > Once the cli is complete, the config file and the start script won't be needed anymore. The current method of wrapping the config is simpler for testing purpose because it doesn't require the data team to be familiar with the config file format.

Maybe we could help the data team get up to speed on the config format? I'd be happy to help @sydneynotthecity out with this if this. The format is very simple so hopefully it won't take long.

Another option I can think of to avoid baking custom wrappers in this repo would be to build standard image here and have the data team build their own based on one from this repo. They should be able to start with `FROM stellar/ledgerexporter:latest` and add custom wrapper on top with minimal effort. I can also help with that if needed.

> Also, once the "start" is removed, there is no other way to run the container than by providing ledger exporter command line parameters in the docker run command.

Yes this is great. I was thinking that if we do it in this PR then there will be no need to change anything when you update ledgerexporter

jacekn (approved):

exp/services/ledgerexporter/docker/start
urvisavla: @jacekn, The team discussed this during our standup today and prioritized addressing these concerns. Since I'll be on vacation for the next couple of weeks, I'll merge this PR for now but I have created a [ticket](https://stellarorg.atlassian.net/browse/HUBBLE-369) to address these and will resume work on this as well as implementing CLI after I return. Thanks!

PR #5260, opened 2024-03-24
ledgerbackend: Add cloud_storage_ledger backend

chowbao:
<details>
  <summary>PR Checklist</summary>
  
### PR Structure

* [ ] This PR has reasonably narrow scope (if not, break it down into smaller PRs).
* [ ] This PR avoids mixing refactoring changes with feature changes (split into two PRs
  otherwise).
* [ ] This PR's title starts with name of package that is most changed in the PR, ex.
  `services/friendbot`, or `all` or `doc` if the changes are broad or impact many
  packages.

### Thoroughness

* [ ] This PR adds tests for the most critical parts of the new functionality or fixes.
* [ ] I've updated any docs ([developer docs](https://developers.stellar.org/api/), `.md`
  files, etc... affected by this change). Take a look in the `docs` folder for a given service,
  like [this one](https://github.com/stellar/go/tree/master/services/horizon/internal/docs).

### Release planning

* [ ] I've updated the relevant CHANGELOG ([here](services/horizon/CHANGELOG.md) for Horizon) if
  needed with deprecations, added features, breaking changes, and DB schema changes.
* [ ] I've decided if this PR requires a new major/minor version according to
  [semver](https://semver.org/), or if it's mainly a patch change. The PR is targeted at the next
  release branch if it's not a patch change.
</details>

### What

BufferedStorageBackend adds in a new ledgerBackend reader for processing txmeta files from a storage service (the only datastore currently implemented is for GCS)

### Why

The new ledgerbackend is created to allow reading txmeta from a cloud storage service where the ledgerexporter is writing txmeta files

### Known limitations

* Adding this new backend in did require pulling out some shared logic with LedgerExporter and moving it to support in order to prevent a circular dependencies error. For instance ledgerexporter imports ledgerbackend but also contains the logic for datastore which is also used in ledgerbackend

PR #5264, opened 2024-03-28
/exp/services/ledgerexporter: resumable export, check data storage for optimal starting point

sreuland:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

* Added `ResumableManager`, app will call this during init to apply a best effort attempt to check data storage layer and increase the effective `start` value for captive core ingestion to the boundary aligned starting ledger of the nearest absent object key of data store relative to requested `start`. 
* Enhanced the DataStoreConfig toml model to be more generic key/value params to allow for holding different data store configuration values going forward, the app checks `datastore_config.type` to know which datastore instance to create, passing the `datastore_config.params` to it:
  ```
  [datastore_config]
  type = "GCS"
  
  [datastore_config.params]
  destination_url = "gcs://your-bucket-name"
  ```

### Why

Efficiency of exporter, avoiding exporting files to data store that may already be present.

Closes: https://stellarorg.atlassian.net/browse/HUBBLE-272


sreuland:
I need to run some live integration tests against GCS.

exp/services/ledgerexporter/internal/app.go
```diff
@@ -19,30 +19,66 @@ var (
...
+	if a.dataStore, err = NewDataStore(ctx, config.DataStoreConfig, config.Network, config.ExporterConfig); err != nil {
+		return errors.Wrap(err, "Could not connect to destination data store")
+	}
+
 	a.uploader = NewUploader(a.dataStore, a.exportManager.GetMetaArchiveChannel())
+
+	resumableManager := NewResumableManager(a.dataStore, a.config.ExporterConfig, NetworkManagerService, config.Network)
+	resumableStartLedger := resumableManager.FindStartBoundary(ctx, config.StartLedger, config.EndLedger)
+	if config.EndLedger > 0 && resumableStartLedger > config.EndLedger {
+		return &DataAlreadyExported{Start: config.StartLedger, End: config.EndLedger}
```
sreuland, line 64: optimization, since resume has determined that the entire requested range is already on data store, there is no need to run the export, short circuit here so app can gracefully shut down.

sreuland:
> I need to run some live integration tests, building ledgerexporter and running it against GCS bucket.

I ran ledgerexporter on testnet against a real gcs bucket doing a sequence of commands that requested ranges overlapping, to confirm the expected resumable starts were calculated, using `ledgers_per_file=1`:

```
ledgerexporter --config-file config.toml --start 842267 --end 844267
... started core at 842267, uploaded all 2000

ledgerexporter --config-file config.toml --start 842267 --end 844267
... log output stating full range existed on data store already, no core or upload attempted

ledgerexporter --config-file config.toml --start 842267 --end 844277
... log output stating start was adjusted to 844268, uploaded next 10

 ledgerexporter --config-file config.toml --start 842267 --end 844287
... log output stating start was adjusted to 844278, uploaded next 10

 ledgerexporter --config-file config.toml --start 842267 --end 0
... log output stating start was adjusted to 844288, and continued on unbounded upload from there
```

exp/services/ledgerexporter/Makefile
```diff
@@ -14,12 +14,18 @@ $(if $(STELLAR_CORE_VERSION), --build-arg STELLAR_CORE_VERSION=$(STELLAR_CORE_VE
...
+	$(SUDO) docker network rm test-network || true
+    
+docker-test: docker-clean
 	# Create temp storage dir
-	$(SUDO) mkdir -p ${PWD}/storage/exporter-test
+	$(SUDO) mkdir -p ${PWD}/storage/exporter-test 
 
 	# Create test network for docker
-	$(SUDO) docker network create test-network
+	$(SUDO) docker network create test-network
```
tamirms, line 28: trailing whitespace

exp/services/ledgerexporter/internal/resumablemanager.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	binarySearchStart := start
+	binarySearchStop := end
+	nearestAbsentLedger := uint32(0)
+	lookupCache := map[string]bool{}
+
+	logger.Infof("Resumability searching datastore for next absent object key between ledgers %d and %d", start, end)
+
+	for binarySearchStart <= binarySearchStop {
```
tamirms, line 70: there is a binary search function in the go standard library:

https://pkg.go.dev/sort#Search

I recommend using that because implementing binary search without bugs is [difficult](https://stackoverflow.com/questions/504335/what-are-the-pitfalls-in-implementing-binary-search)

exp/services/ledgerexporter/internal/resumablemanager.go
```diff
@@ -0,0 +1,110 @@
...
+// most recent archived ledger number.
+//
+// return - the next bounded start ledger position
+//
+// Will be non-zero If able to identify the nearest "LedgersPerFile" starting boundary ledger number
+// which is absent on datastore given start and end.
+// If data store has all files up to end, then returns the next "LedgersPerFile" starting boundary ledger
+// If not able to identify next boundary ledger due to any type of error, returns 0.
+func (rm resumableManagerService) FindStartBoundary(ctx context.Context, start, end uint32) uint32 {
+	if ctx.Err() != nil {
```
tamirms, line 38: I don't think it's necessary to check context here or in the binary search loop so long as the context is propagated in all the long running operations like the http calls. the time spent in the code which runs in between the long running operations is very very small

exp/services/ledgerexporter/internal/resumablemanager.go
```diff
@@ -0,0 +1,110 @@
...
+// If end=0, meaning unbounded, this will substitute an effective end value of the
+// most recent archived ledger number.
+//
+// return - the next bounded start ledger position
+//
+// Will be non-zero If able to identify the nearest "LedgersPerFile" starting boundary ledger number
+// which is absent on datastore given start and end.
+// If data store has all files up to end, then returns the next "LedgersPerFile" starting boundary ledger
+// If not able to identify next boundary ledger due to any type of error, returns 0.
+func (rm resumableManagerService) FindStartBoundary(ctx context.Context, start, end uint32) uint32 {
```
tamirms, line 37: I think returning `(uint32, error)` would be a more intuitive API and it would allow the caller to respond differently based on the type of error

exp/services/ledgerexporter/internal/resumablemanager.go
```diff
@@ -0,0 +1,110 @@
...
+
+	// streaming mode for start, no historical point to resume from
+	if start < 1 {
+		return 0
+	}
+
+	// streaming mode for end, get current ledger to use for a sane bounded range during resumability check
+	if end < 1 {
+		var latestErr error
+		end, latestErr = rm.networkManager.GetLatestLedgerSequenceFromHistoryArchives(ctx, rm.network)
```
tamirms, line 50: checkpoint ledgers occur once every 64 ledgers, so it's possible that the last exported ledger occurs after the latest checkpoint ledger. for that reason, I would suggest using the latest checkpoint ledger + 128 as the upper bound of the search because we know that it's possible that the last exported ledger could be greater than the latest checkpoint ledger, however, we are very confident that it will be less than the latest checkpoint ledger + 128

exp/services/ledgerexporter/internal/resumablemanager.go
```diff
@@ -0,0 +1,110 @@
...
+		}
+		logger.Infof("Resumability resovled unbounded to latest ledger =%d for network=%v", end, rm.network)
+
+		if start > end {
+			// requested to start at a point beyond the latest network, resume not applicable.
+			return 0
+		}
+	}
+
+	binarySearchStart := start
```
tamirms, line 63: if we have a buckets db from the previous run we should [query](https://github.com/stellar/go/blob/master/ingest/ledgerbackend/stellar_core_runner.go#L371-L379) it to find the last closed ledger because it's very likely that the last uploaded ledger is close to the LCL.

I think we should use LCL - bucketSize *2 as the first guess to narrow down the search space

exp/services/ledgerexporter/internal/networkmanager.go
```diff
@@ -0,0 +1,54 @@
...
+
+var (
+	NetworkManagerService = &networkManagerService{}
+)
+
+type NetworkManager interface {
+	GetLatestLedgerSequenceFromHistoryArchives(ctx context.Context, networkName string) (uint32, error)
+}
+
+type networkManagerService struct{}
```
tamirms, line 20: I think this struct is unnecessary. we can use an instance of [ArchiveInterface](https://github.com/stellar/go/blob/master/historyarchive/archive.go#L63) to find the latest checkpoint ledger. There is also a [mock implementation](https://github.com/stellar/go/blob/master/historyarchive/mocks.go#L9) in case we need to use it in unit tests

exp/services/ledgerexporter/internal/mocks.go
```diff
@@ -0,0 +1,87 @@
...
+type MockNetworkManager struct {
+	mock.Mock
+}
+
+func (m *MockNetworkManager) GetLatestLedgerSequenceFromHistoryArchives(ctx context.Context, networkName string) (uint32, error) {
+	a := m.Called(ctx, networkName)
+	return a.Get(0).(uint32), a.Error(1)
+}
+
+type MockResumableManager struct {
```
tamirms, line 74: this doesn't seemed to be used anywhere

exp/services/ledgerexporter/docker/start
```diff
@@ -29,10 +34,6 @@ EOF
 if [[ -n "$START" || -n "$END" ]]; then
     echo "START: $START END: $END"
     /usr/bin/ledgerexporter --config-file config.toml --start $START --end $END
-# Check if FROM_LAST variable is set
```
tamirms, line 32: I think there is still value in having the `--from-last` flag. Consider the use case of a data lake which has a retention period of 1 week.

exp/services/ledgerexporter/README.md
```diff
@@ -33,16 +33,15 @@ Exports ledgers continuously starting from --start. In this mode, the end ledger
...
-ledgerexporter --from-last <number_of_ledgers> --config-file <config_file_path>
-```
-
 ### Configuration (toml):
 
 ```toml
 network = "testnet"  # Options: `testnet` or `pubnet`
+
+[datastore_config]
+type = "GCS"
```
tamirms, line 42: if we are introducing a type field then we should probably remove the destination url scheme where we use the `gcs://` prefix for gcs objects. perhaps this type of refactor should be done in a separate PR since it isn't necessary for resumability ?

exp/services/ledgerexporter/internal/app.go
```diff
@@ -58,7 +94,16 @@ func (a *App) Run() {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
-	a.init(ctx)
+	if err := a.init(ctx); err != nil {
+		switch err.(type) {
```
tamirms, line 98: better to use https://pkg.go.dev/errors#As in case the error is wrapped

exp/services/ledgerexporter/internal/app.go
```diff
@@ -19,30 +19,66 @@ var (
 	logger = log.New().WithField("service", "ledger-exporter")
 )
 
+type DataAlreadyExported struct {
+	Start uint32
+	End   uint32
+}
+
+func (m *DataAlreadyExported) Error() string {
```
tamirms, line 27: I think we should use the value receive instead of the pointer receive here because DataAlreadyExported is a small struct which is supposed to be immutable

exp/services/ledgerexporter/internal/config.go
```diff
@@ -17,97 +17,71 @@ import (
 const Pubnet = "pubnet"
 const Testnet = "testnet"
 
+type Flags struct {
```
tamirms, line 20: I think the separation of flags from other config would be best handled in the PR which will implement https://stellarorg.atlassian.net/browse/HUBBLE-271 and https://stellarorg.atlassian.net/browse/HUBBLE-374 since these changes are not strictly necessary for resumability and the ops team will probably want to weigh in on any changes which impact the usability of the cli

exp/services/ledgerexporter/internal/exportmanager.go
```diff
@@ -2,17 +2,54 @@ package ledgerexporter
...
 	"github.com/pkg/errors"
 	"github.com/stellar/go/ingest/ledgerbackend"
 	"github.com/stellar/go/xdr"
 )
 
+const (
+	fileSuffix = ".xdr.gz"
+)
+
 type ExporterConfig struct {
```
tamirms, line 16: with the functions defined on `ExporterConfig` and how it's used in the resumability code, it seems `ExporterConfig` should be renamed to `LedgerBatchConfig` or something similar because this struct is describing how many ledgers are included in each tx meta file and the naming scheme of the tx meta files. The ExportManager relies on this functionality but the ExportManager may evolve to include more configuration properties that are not relevant to resumability

exp/services/ledgerexporter/internal/resumablemanager.go
sreuland: ok, yes, if it's not optimal, then it's just noise, will remove.

exp/services/ledgerexporter/README.md
sreuland: yes, I agree, It seemed like a quick win to demonstrate how this is a pluggable framework for datastores, I can pull it out, but before doing so, I could also remove that one last reference to `gcs://` in gcs_datastore.go to avoid that overlap and if that is clean, can go with it?

exp/services/ledgerexporter/docker/start
sreuland: we had a [discussion](https://stellarorg.atlassian.net/browse/HUBBLE-272?focusedCommentId=16465) on `--from-last` flag on hubble-272, as it introduces potential for gaps when used in conjuction with resumability, there were a couple options suggested, the consensus was to remove the flag. This use case your raise is a good one, it could also be achieved by using existing `--start` and we assume the caller of ledgerexporter can externally obtain/generate the equivalent ledger value?

tamirms: I didn't realize you had already discussed removing the flag. In that case you can ignore my comment and proceed with the removal. We can always add it back in if there's a demand for that specific use case

exp/services/ledgerexporter/internal/config.go
sreuland: the flags are still separate from 'other config' such as the Config struct and config.toml, they were just internally refactored to an aggregate`Flags` model for single reference to all of them, the flags retrieval was shifted to the app which seems ideal place as it knows best where to source this info in any case, rather than prior, this was hardcoding the retrieval of these from command line internally within config creation, the app would have no way to override. Net effect outwardly is no functional change. I think this is a small non-functional refactor which still provides the same starting point that  https://stellarorg.atlassian.net/browse/HUBBLE-271 and https://stellarorg.atlassian.net/browse/HUBBLE-374 can proceed?

exp/services/ledgerexporter/README.md
urvisavla: I think we should defer adding any config parameters that aren't absolutely necessary to the [CLI](https://stellarorg.atlassian.net/browse/HUBBLE-271) ticket.  Because each parameter  added to the config file will need to be supported via command line (and environment variables) as well. So may be we can take a bit more time and design the CLI/config.

exp/services/ledgerexporter/Makefile
```diff
@@ -14,12 +14,18 @@ $(if $(STELLAR_CORE_VERSION), --build-arg STELLAR_CORE_VERSION=$(STELLAR_CORE_VE
 	-t $(DOCKER_IMAGE):$(VERSION) \
 	-t $(DOCKER_IMAGE):latest .
 
-docker-test:
+docker-clean:
```
urvisavla, line 17: Thanks for adding this.

exp/services/ledgerexporter/internal/datastore.go
```diff
@@ -21,37 +17,12 @@ type DataStore interface {
...
 }
 
-// NewDataStore creates a new DataStore based on the destination URL.
-// Currently, only accepts GCS URLs.
-func NewDataStore(ctx context.Context, destinationURL string) (DataStore, error) {
-	parsed, err := url.Parse(destinationURL)
-	if err != nil {
-		return nil, err
+// NewDataStore factory, it creates a new DataStore based on the config type
+func NewDataStore(ctx context.Context, datastoreConfig DataStoreConfig, network string, exporterConfig ExporterConfig) (DataStore, error) {
```
urvisavla, line 21: `exporterConfig ExporterConfig` parameter is unused.

exp/services/ledgerexporter/internal/networkmanager.go
sreuland: the NetworkManager abstraction also encapsulates obtaining the archive server urls per the config, w/o this abstraction callers will end up duplicating this to create ArchiveInterface instance or it gets refactored into a shared function somewhere, in either case, I think the result is no longer late dispatch, i.e. does not leave a way to mock out in unit tests on config.NewConfig() and ResumableManager.FindStartBoundary()

exp/services/ledgerexporter/internal/resumablemanager.go
```diff
@@ -0,0 +1,110 @@
...
+
+		if middleFoundOnStore {
+			binarySearchStart = binarySearchMiddle + 1
+		} else {
+			nearestAbsentLedger = binarySearchMiddle
+			binarySearchStop = binarySearchMiddle - 1
+		}
+	}
+
+	if nearestAbsentLedger > 0 {
```
urvisavla, line 100: Should we consider asserting here or at the caller that the nearestAbsentLedger falls on the ledgers-per-file boundary? It could help catch any changes in the ledger-per-file config which would result in data inconsistency on the specified destination?

exp/services/ledgerexporter/internal/exportmanager.go
```diff
@@ -2,17 +2,54 @@ package ledgerexporter
...
 	LedgersPerFile    uint32 `toml:"ledgers_per_file"`
 	FilesPerPartition uint32 `toml:"files_per_partition"`
 }
 
+func (ec ExporterConfig) GetSequenceNumberStartBoundary(ledgerSeq uint32) uint32 {
+	return (ledgerSeq / ec.LedgersPerFile) * ec.LedgersPerFile
```
urvisavla, line 22: Should we validate that ec.LegersPerFile is non-zero here?

exp/services/ledgerexporter/internal/resumablemanager.go
sreuland: done - https://github.com/stellar/go/pull/5264/commits/decdcd3e6348293aff4f3194cfd4a0d997836043

exp/services/ledgerexporter/internal/resumablemanager.go
sreuland: done - https://github.com/stellar/go/pull/5264/commits/decdcd3e6348293aff4f3194cfd4a0d997836043

exp/services/ledgerexporter/README.md
sreuland: after reading [CLI](https://stellarorg.atlassian.net/browse/HUBBLE-271), the scope doesn't touch on potential datastore config, but it may actually be beneficial to have the more pluggable key/value params schema present before going into that effort as it would then get woven into cli expectations from the beginning. Let me know if that makes a difference here, otherwise I will remove it, thanks!

exp/services/ledgerexporter/internal/app.go
sreuland: good point, done, - https://github.com/stellar/go/pull/5264/commits/decdcd3e6348293aff4f3194cfd4a0d997836043

exp/services/ledgerexporter/internal/exportmanager.go
sreuland: agreed, pulled it up to ledgerbatch_config.go - https://github.com/stellar/go/pull/5264/commits/32302d4d4cb719085801a87c6318f1a9a380d8df

exp/services/ledgerexporter/internal/resumablemanager.go
sreuland: is this condition of detecting changes of `ledgers-per-file` done anywhere else yet for reference? I don't understand the implications enough yet, are you suggesting `nearestAbsentLedger % ledger-per-file = 0` ? that seems to only catch one value-based case of changing `ledgers-per-file`?

exp/services/ledgerexporter/internal/app.go
```diff
@@ -19,30 +19,66 @@ var (
...
+	if config, err = NewConfig(ctx, NetworkManagerService, a.flags); err != nil {
+		return errors.Wrap(err, "Could not load configuration")
+	}
+	a.config = *config
+
+	if a.dataStore, err = NewDataStore(ctx, a.config.DataStoreConfig, a.config.Network, a.config.ExporterConfig); err != nil {
+		return errors.Wrap(err, "Could not connect to destination data store")
+	}
+
+	resumableManager := NewResumableManager(a.dataStore, a.config.ExporterConfig, NetworkManagerService, a.config.Network)
```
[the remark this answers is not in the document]
sreuland: what's the situation that it would be used in? after partner experience effort, I lean towards not adding a flag such as this which majority of time the default will be used and may indirectly be a coverup for something else that is awry, like a datastore problem?  

One alternative to this flag is for user to just delete the start+1 object on the data store, that absence would trigger exporter's resumability search algorithm to find start+1 for the resumable start position, which is pretty close to `--resume false`.

exp/services/ledgerexporter/internal/exportmanager.go
sreuland: great catch! thanks, fixed - https://github.com/stellar/go/pull/5264/commits/ebe8bc95744e6e8d3725be42cd8f3c9aebe9a3d5

exp/services/ledgerexporter/internal/resumablemanager.go
sreuland: sounds good, wired that in  - https://github.com/stellar/go/pull/5264/commits/f8aae6f975be1b8f3eee2308124cfed8adb809ae

exp/services/ledgerexporter/internal/resumablemanager.go
sreuland: Can this assumption of buckets db having accurate LCL left on disk from any prior run of core with `catchup` or `run` be safely done? Is it ok to intentionally override the user's requested range? maybe the user knows their remote storage has a gap or maybe the user just wants the remote storage verified from the starting range they requested.

this seems like it could introduce more complexity related to LCL retrieval, requiring further knowledge specific to LedgerBackend and potential DX(developer experience) confusion by overriding the user's request and ensuing expectation in exchange for realizing a minimal gain in performance, as this is binary search, the shortened range realized by this may only result in eliminating a couple data storage binary search 'middle' queries at best, is this worth it?

exp/services/ledgerexporter/internal/resumablemanager.go
sreuland: nice! [refactored to use sort.Search](https://github.com/stellar/go/pull/5264/commits/4151d9c2f73d66acd26ad539c66e81ec31905633), was cool to see the test cases including mocked expectations for when data store queries are performed during binary search didn't require much of any change, almost exactly the same 'middle' queries.

exp/services/ledgerexporter/internal/networkmanager.go
tamirms: it seems to me that NetworkManager is just a wrapper around the archive interface. How we construct the archive instance isn't really important. Instead of passing in a NetworkManager to config.NewConfig() or ResumableManager.FindStartBoundary() you can pass in an archive instance.  If you want to mock out the archive functionality you can pass in a MockArchive into config.NewConfig() or ResumableManager.FindStartBoundary().

The reason why I prefer using an archive interface is because every abstraction we introduce adds extra cognitive load to the reader because you have to understand what the abstraction is representing. So when I see a new interface which only contains one function, `GetLatestLedgerSequenceFromHistoryArchives(ctx context.Context, networkName string) (uint32, error)`, the first question that pops in my head is: how is this different from the pre-existing ArchiveInterface? It is not apparent what value the NetworkManager interface is providing to justify this extra cognitive load

exp/services/ledgerexporter/README.md
urvisavla: Having separate flags is fine 👍 but I think having a single flag that includes destination type in the destination URL might be simpler because I can't think of a scenario where having separate flags would be beneficial but that's just my initial thought..

exp/services/ledgerexporter/internal/app.go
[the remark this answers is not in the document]
urvisavla: It could be useful if we [suspect any gaps](https://github.com/stellar/go/pull/5264#discussion_r1567215185) and want to fill them for a given range?

exp/services/ledgerexporter/internal/resumablemanager.go
urvisavla: That's a good point. We're not checking it elsewhere because of lack of info about what has been exported. In this case, not only there is no assert on it but also the code also returns the `nearestAbsentBoundaryLedger` instead of `nearestAbsentLedgerSequence` which will prevent the caller to performing the check. May be we simply return the `nearestAbsentLedgerSequence` and let the caller crosscheck with the `ledgers_per_file` config, wdyt?

exp/services/ledgerexporter/internal/app.go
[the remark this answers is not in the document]
sreuland: ah, got it, resume would be too eager and short circuit that gap fill by-default behavior for a given range requested, will add a new `--resume`, thanks!

exp/services/ledgerexporter/internal/resumablemanager.go
sreuland: to maintain good dev experiencer I will need to add [--resume=true|false](https://github.com/stellar/go/pull/5264#discussion_r1544855413) here in pr so users can disable resumption if they have known gaps on requested export range,  but, if they don't disable resume, then we are in the `monotonic-safe-to-do-binary` case.

I see a 'paranoid' edge case here if user has prior run `ledgerexporter` against a different remote data store, then the left over state in the captive core buckets/LCL isn't relevant when the same `ledgerexporter`on machine is run against a different remote data store correct? If we can dismiss that case though, is the extra complexity to query LCL worth the net potential gain on performance? Also, just to confirm, your suggestion is to use `binarySearchStart=LCL - bucketSize *2` or am I mis-understanding intent?

exp/services/ledgerexporter/README.md
sreuland: is reference to 'flags' meaning a toml config parameter as shown here? This encodes config parameters for data store types as toml key/values rather than encoding the same into a single urn/url. The 'gcs' data store currently requires one config parameter for `destination_bucket_path`. That single value approach is possible because it is not including other required config params for gcs, such as the authentication creds, since we are assuming the operator will provide those settings via [GOOGLE_APPLICATION_CREDENTIALS](https://cloud.google.com/docs/authentication/application-default-credentials#GAC) in the environment for implicit authentication by gcp sdk. But, the auth config could be modeled as more config params also like`credentials=/path/to/local/credentials.json` or `token=xyz` just to illustrate the potential for more data store params.

The type and key/value format in toml, should provide a flexible model now that handles the single-value config of gcs data store, but at same time can demonstrate how ledgerexporter is enabled to use different data store types, with different config models, ensuring the type specific config gets passed to the associated constructor for the matching Datastore type implementation at ledgerexporter init time.

exp/services/ledgerexporter/internal/resumablemanager.go
```diff
@@ -0,0 +1,118 @@
...
+		return rm.ledgerBatchConfig.GetSequenceNumberStartBoundary(networkLatest), false
+	}
+
+	// data store had all ledgers for requested range, no resumability needed.
+	logger.Infof("Resumability found no absent object start keys between ledgers %d and %d", start, end)
+	return 0, true
+}
+
+func binarySearchCallbackFn(rm *resumableManagerService, ctx context.Context, start, end uint32) func(ledgerSequence int) bool {
+	lookupCache := map[string]bool{}
```
Shaptic, line 98: `Set[T]` will help here (from our [support package](https://github.com/stellar/go/blob/master/support/collections/set/set.go)).

exp/services/ledgerexporter/internal/resumablemanager.go
```diff
@@ -0,0 +1,118 @@
...
+		if start > networkLatest {
+			// requested to start at a point beyond the latest network, resume not applicable.
+			return 0, false
+		}
+	}
+
+	binarySearchStop := end
+	if networkLatest > 0 {
+		binarySearchStop = networkLatest
+	}
```
Shaptic, line 73: this is the same as `binarySearchStop := max(end, networkLatest)` since `networkLatest` is only set if `end == 0`, right?

exp/services/ledgerexporter/internal/resumablemanager.go
```diff
@@ -0,0 +1,118 @@
...
+	// streaming mode for end, get latest network ledger to use for a sane bounded range during resumability check
+	// this will assume a padding of network latest = network latest + 2 checkpoint_frequency,
+	// since the latest network will be some number of ledgers past the last archive checkpoint
+	// this lets the search be a little more greedy on finding a potential empty object key towards the end of range on data store.
+	networkLatest := uint32(0)
+	if end < 1 {
+		var latestErr error
+		networkLatest, latestErr = rm.networkManager.GetLatestLedgerSequenceFromHistoryArchives(ctx, rm.network)
+		if latestErr != nil {
+			logger.WithError(latestErr).Infof("Resumability of requested export ledger range start=%d, end=%d, was not able to get latest ledger from network %v", start, end, rm.network)
```
Shaptic, line 57: nit: log level
```suggestion
			logger.WithError(latestErr).Errorf("Resumability of requested export ledger range start=%d, end=%d, was not able to get latest ledger from network %v", start, end, rm.network)
```

exp/services/ledgerexporter/internal/resumablemanager.go
```diff
@@ -0,0 +1,118 @@
...
+		return nearestAbsentBoundaryLedger, false
+	}
+
+	// unbounded, and datastore had up to latest network, return the start for youngest ledger on data store
+	if networkLatest > 0 {
+		return rm.ledgerBatchConfig.GetSequenceNumberStartBoundary(networkLatest), false
+	}
+
+	// data store had all ledgers for requested range, no resumability needed.
+	logger.Infof("Resumability found no absent object start keys between ledgers %d and %d", start, end)
```
Shaptic, line 93: might be mildly cleaner to do something like `L := logger.WithField("start", start).WithField("end", end)` at the beginning and just reuse `L` for these log lines instead of doing the `%d` thing for each log line

exp/services/ledgerexporter/internal/encoder.go
```diff
@@ -0,0 +1,39 @@
...
+		return 0, err
+	}
+	defer gr.Close()
+
+	n, err := xdr3.Unmarshal(gr, d.XdrPayload)
+	if err != nil {
+		return int64(n), err
+	}
+	return int64(n), nil
+}
```
Shaptic, line 39: nice! would be really good to make explicit which interfaces this is intended to support

Shaptic: also could integrate this into `ingest` I bet

exp/services/ledgerexporter/internal/gcs_datastore.go
```diff
@@ -20,6 +24,42 @@ type GCSDataStore struct {
...
+	gcsBucketURL := fmt.Sprintf("gcs://%s/%s", destinationBucketPath, network)
+	parsed, err := url.Parse(gcsBucketURL)
+	if err != nil {
+		return nil, err
+	}
+
+	// Inside gcs, all paths start _without_ the leading /
+	pth := strings.TrimPrefix(parsed.Path, "/")
+	bucketName := parsed.Host
+	prefix := pth
```
Shaptic, line 44: nit: 

```suggestion
	// Inside gcs, all paths start _without_ the leading /
	prefix := strings.TrimPrefix(parsed.Path, "/")
	bucketName := parsed.Host
```

exp/services/ledgerexporter/internal/config.go
sreuland: does my explanation adequately explain reasons for leaving the internal-only refactor of flags in? just need a thumbs up/down, to confirm next step, thanks!

exp/services/ledgerexporter/internal/resumablemanager.go
sreuland: thanks!

exp/services/ledgerexporter/internal/resumablemanager.go
sreuland: yes, that approach has better readability,  updated, thanks!

exp/services/ledgerexporter/internal/resumablemanager.go
sreuland: Ah. ok, thanks for clarification, I will change this to return the raw data, i.e. the non-boundary absent ledger, and avoid the resumemanager from having knowledge of boundary'ness. I intentionally applied rounding up to nearest boundary ledger thinking it would encapsulate/enforce consistency and caller would not have to juggle it, but it sounds like the expectation is the opposite, with calller knowing best what to do with this data as we are still wrangling on approach for handling `ledgers-per-file` aspects of existing remote data. btw, this same notion of [capturing ledgers-per-file on remote data store came up on a separate go/pull/5260 recently also](https://github.com/stellar/go/pull/5260#discussion_r1572876523).

exp/services/ledgerexporter/internal/resumablemanager.go
sreuland: nice suggestion, done - https://github.com/stellar/go/pull/5264/commits/e0013133e5db9f8d0a60c1d92e46c0bf233225ed

exp/services/ledgerexporter/internal/app.go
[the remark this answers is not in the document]
sreuland: added `--resume` - https://github.com/stellar/go/pull/5264/commits/73028494374fc59dd5ee8d3c16565e1ab731a655

exp/services/ledgerexporter/internal/resumablemanager.go
sreuland: changed the resume manager to return non-boundary, lowest absent ledger number from data store, and added validation on that ledger sequence against app's ledgers-per-file config value during app.init - https://github.com/stellar/go/pull/5264/commits/57e5cea1035aa78880e971b589b02666bc230521

exp/services/ledgerexporter/internal/resumablemanager.go
tamirms: If the captive core LCL isn't then we end up wasting a request to GCS. However, the code should still return the correct result.

> is the extra complexity to query LCL worth the net potential gain on performance?

I think it's worth it when you have a data lake with full history. Currently, there are ~50 million ledgers in full history. To find the latest ledger in the data lake, it would take log_2(50 * 10^6) which comes out to about 25 requests to GCS. If we use the LCL from captive core, we could potentially reduce the search space to ~ 100s of ledgers or less . log_2(100) comes out to ~ 7 requests to GCS.

But if you think there is significant complexity in using the LCL from captive core, we can always create a follow up issue and defer this optimization until later.

> Also, just to confirm, your suggestion is to use binarySearchStart=LCL - bucketSize *2 or am I mis-understanding intent?

yes, that's right. but anything around that ballpark should work

exp/services/ledgerexporter/internal/config.go
tamirms: sorry for the delay! yes, that makes sense. I think we can leave the flags refactor in the pr

exp/services/ledgerexporter/internal/networkmanager.go
sreuland: I refactored out the network manager in favor of archive interface direct, tweaked the app init routine, as config needed to be parsed first to know target network before archive interface can be created - https://github.com/stellar/go/pull/5264/commits/ef4a400bbacc239ee47794c76b5dbfd0e7f4ba39

exp/services/ledgerexporter/internal/resumablemanager.go
sreuland: ok, I carved this out to a new ticket  - https://stellarorg.atlassian.net/browse/HUBBLE-391, thanks!

exp/services/ledgerexporter/internal/app.go
```diff
@@ -24,41 +25,117 @@ var (
 	logger = log.New().WithField("service", "ledger-exporter")
 )
 
+func NewDataAlreadyExported(Start uint32, End uint32) *DataAlreadyExported {
```
tamirms, line 28: Error types end in "Error"

https://go.dev/wiki/Errors#naming

exp/services/ledgerexporter/internal/app.go
```diff
@@ -24,41 +25,117 @@ var (
...
+}
+
+func NewInvalidDataStore(LedgerSequence uint32, LedgersPerFile uint32) *InvalidDataStore {
+	return &InvalidDataStore{
+		LedgerSequence: LedgerSequence,
+		LedgersPerFile: LedgersPerFile,
+	}
+}
+
+type InvalidDataStore struct {
```
tamirms, line 51: Error types end in "Error"

https://go.dev/wiki/Errors#naming

exp/services/ledgerexporter/internal/app.go
```diff
@@ -24,41 +25,117 @@ var (
...
+func (a *App) init(ctx context.Context) error {
+	var config *Config
+	var err error
+	var archive historyarchive.ArchiveInterface
+
+	if config, err = NewConfig(ctx, a.flags); err != nil {
+		return errors.Wrap(err, "Could not load configuration")
+	}
+
+	if archive, err = CreateHistoryArchiveFromNetworkName(ctx, config.Network); err != nil {
```
tamirms, line 93: `CreateHistoryArchiveFromNetworkName()` seems like a private helper function which will not be used by any other package. If that's the case it should not be exported and would be better renamed to `createHistoryArchiveFromNetworkName()`

exp/services/ledgerexporter/internal/app.go
```diff
@@ -24,41 +25,117 @@ var (
...
+		return err
+	}
+	config.ValidateAndSetLedgerRange(ctx, archive)
+	a.config = *config
```
tamirms, line 97: a.config and config should have the same type. so either a.config should be a pointer or `NewConfig()` should return a value. I would go with the option of making a.config a pointer

exp/services/ledgerexporter/internal/config.go
```diff
@@ -13,106 +13,117 @@ import (
...
+			UserAgent: "ledger-exporter",
+			Context:   ctx,
+		},
+	})
+}
 
-	// Validate config params
-	err = config.validateAndSetLedgerRange(latestNetworkLedger)
-	logFatalIf(err, "Error validating config params.")
+func GetLatestLedgerSequenceFromHistoryArchives(ctx context.Context, archive historyarchive.ArchiveInterface) (uint32, error) {
```
tamirms, line 74: if this is a private helper function which will not be used outside of this package then this function should not be exported

exp/services/ledgerexporter/internal/config.go
```diff
@@ -13,106 +13,117 @@ import (
...
+	}
 
-	// Validate and build the appropriate range
-	// TODO: Make it configurable
-	config.adjustLedgerRange()
+	return has.CurrentLedger, nil
+}
 
-	return nil
+func GetHistoryArchivesCheckPointFrequency() uint32 {
```
tamirms, line 84: why not use historyarchive.DefaultCheckpointFrequency directly instead of creating a `GetHistoryArchivesCheckPointFrequency()` wrapper?

exp/services/ledgerexporter/internal/exportmanager.go
```diff
@@ -11,45 +11,51 @@ import (
...
 
-type ExporterConfig struct {
-	LedgersPerFile    uint32 `toml:"ledgers_per_file"`
-	FilesPerPartition uint32 `toml:"files_per_partition"`
-}
-
 // ExportManager manages the creation and handling of export objects.
-type ExportManager struct {
-	config             ExporterConfig
+type ExportManager interface {
```
tamirms, line 15: I think it's better to remove this interface and make `NewExportManager()` return the concrete type like how it was before. 

It is generally recommended not to define interfaces before they are used:

https://go.dev/wiki/CodeReviewComments#interfaces

exp/services/ledgerexporter/internal/exportmanager.go
```diff
@@ -11,45 +11,51 @@ import (
...
+	Run(ctx context.Context, startLedger uint32, endLedger uint32) error
+	AddLedgerCloseMeta(ctx context.Context, ledgerCloseMeta xdr.LedgerCloseMeta) error
+	GetLatestLedgerMetric() *prometheus.GaugeVec
```
tamirms, line 18: I don't think we need to publicly expose the latest ledger metric

exp/services/ledgerexporter/internal/mocks.go
```diff
@@ -0,0 +1,86 @@
...
+	args := m.Called(ctx, path, in)
+	return args.Get(0).(bool), args.Error(1)
+}
+
+func (m *MockDataStore) Close() error {
+	args := m.Called()
+	return args.Error(0)
+}
+
+type MockExportManager struct {
```
tamirms, line 47: this type does not seem to be used anywhere so we can delete it

exp/services/ledgerexporter/internal/resumablemanager.go
```diff
@@ -0,0 +1,121 @@
...
+	//
+	// If end=0, meaning unbounded, this will substitute an effective end value of the
+	// most recent archived ledger number.
+	//
+	// return:
+	// resumableLedger - if > 0, will be the next ledger that is not populated on data store.
+	// dataStoreComplete - if true, there was no gaps on data store for bounded range requested
+	//
+	// if resumableLedger is 0 and dataStoreComplete is false, no resumability was applicable.
+	FindStart(ctx context.Context, start, end uint32) (resumableLedger uint32, dataStoreComplete bool)
```
tamirms, line 25: I think `FindStart()` should return an error instead of logging the error and swallowing it. That will give the caller of `FindStart()` more flexibility on how it can handle the error

exp/services/ledgerexporter/internal/resumablemanager.go
```diff
@@ -0,0 +1,121 @@
...
+		dataStore: dataStore,
+		archive:   archive,
+	}
+}
+
+func (rm resumableManagerService) FindStart(ctx context.Context, start, end uint32) (resumableLedger uint32, dataStoreComplete bool) {
+	// start < 1 means streaming mode, no historical point to resume from
+	if start < 1 || !rm.config.Resume {
+		return 0, false
+	}
```
tamirms, line 46: should this be an error case? we require a start parameter in order to do the search

exp/services/ledgerexporter/internal/resumablemanager.go
```diff
@@ -0,0 +1,121 @@
...
+	// return:
+	// resumableLedger - if > 0, will be the next ledger that is not populated on data store.
+	// dataStoreComplete - if true, there was no gaps on data store for bounded range requested
+	//
+	// if resumableLedger is 0 and dataStoreComplete is false, no resumability was applicable.
+	FindStart(ctx context.Context, start, end uint32) (resumableLedger uint32, dataStoreComplete bool)
+}
+
+type resumableManagerService struct {
+	config    *Config
```
tamirms, line 29: I don't think we require the entire config. Consider if @chowbao uses this library for the ledger backend he is implementing, that use case is outside of ledger exporter

exp/services/ledgerexporter/internal/resumablemanager.go
```diff
@@ -0,0 +1,121 @@
...
+		objectKeyMiddle := rm.config.LedgerBatchConfig.GetObjectKeyFromSequenceNumber(start + uint32(binarySearchIndex))
+
+		// there may be small occurrence of repeated queries on same object key once
+		// search narrows down to a range that fits within the ledgers per file
+		// worst case being 'log of ledgers_per_file' queries.
+		middleFoundOnStore, foundInCache := lookupCache[objectKeyMiddle]
+		if !foundInCache {
+			var datastoreErr error
+			middleFoundOnStore, datastoreErr = rm.dataStore.Exists(ctx, objectKeyMiddle)
+			if datastoreErr != nil {
```
tamirms, line 113: If we encounter an error we should have a retry policy. If after retrying the request still fails we should propagate that error back to the caller of `FindStart()`.

exp/services/ledgerexporter/internal/config.go
sreuland: yes, I wasn't sure if that was the best source to get the checkpoint frequency, so abstracted it, maybe it comes from somewhere else later, I might remove this for now.

exp/services/ledgerexporter/internal/resumablemanager.go
sreuland: the intent was to have resumability be best effort and seamless to caller by not propagating the error. seems like only other action to take with error is to stop the export, which I thought was too much interruption, do we want to trigger that now? otherwise `0` allows caller to proceed with original requested export as-is.

exp/services/ledgerexporter/internal/resumablemanager.go
sreuland: wouldn't that client sdk use case entail importing the`ledgerexporter` go package which makes all the public components available? 

Related to that, providing the binary search aspect for his sdk to use, that is what I'm changing at present, am refactoring FindStart out of here into a common interface and ResumeManager and other sdks can use it directly as:

```
type Search interface {
   // start=0, end=0 means no bounded range search the entire datastore
   FindOldestLedger(start, end) (uint32, error)
   FindYoungestLedger(start, end) uint32, error)
}
```

exp/services/ledgerexporter/internal/resumablemanager.go
sreuland: I don't think it requires an err, was returning `0` instead to indicate no resumption was done as part of [best effort, seamless intent](https://github.com/stellar/go/pull/5264/files/7ddc4a0dc7ea8e9248b6f46e003db3a90b6af877#r1581254966), I'll add a line in docs here to mention start=0 means no-op in terms of resumption. If there is objection, I can add error response, but wanted to mention the reasoning behind this first.

exp/services/ledgerexporter/internal/resumablemanager.go
```diff
@@ -0,0 +1,121 @@
...
+	binarySearchStop := max(end, networkLatest)
+	binarySearchStart := start
+
+	log.Infof("Resumability is searching datastore for next absent object key of requested export ledger range")
+
+	rangeSize := max(int(binarySearchStop-binarySearchStart), 1)
+	lowestAbsentIndex := sort.Search(rangeSize, binarySearchCallbackFn(&rm, ctx, binarySearchStart, binarySearchStop))
+	if lowestAbsentIndex < 1 {
+		// data store had no data within search range
+		return 0, false
```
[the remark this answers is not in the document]
sreuland: the response value of this function is meant to indicate a start value determined based on remote data or `0` if no start value was found or applicable. It doesn't make that next value based decision. This allows the caller to decide how they want to continue given the response, i.e. if `0` then caller can choose 'I will resume from the very beginning of my requested range since resumability gave me no new info'

exp/services/ledgerexporter/internal/resumablemanager.go
sreuland: ok, this touches on same regards to propagating errors from resumability rather than swallowing as part of a [best effort, seamless intent](https://github.com/stellar/go/pull/5264/files/7ddc4a0dc7ea8e9248b6f46e003db3a90b6af877#r1581254966), I don't want to add retry logic here for now, more complexity, instead I will add error to the return value here now, as there are a few new comments here which all converge on this, thanks!

exp/services/ledgerexporter/internal/resumablemanager.go
tamirms: if you need to find the latest ledger in the data lake you shouldn't need to provide the full ledger exporter config. The ledger exporter config contains info (such as the captive core binary path and captive core toml path) that is irrelevant for this specific functionality. The library for finding the latest ledger should only take the minimum set of dependencies that are necessary to implement the search

exp/services/ledgerexporter/internal/resumablemanager.go
tamirms: > seems like only other action to take with error is to stop the export, which I thought was too much interruption, do we want to trigger that now?

I think that could be valid option. By returning the error it gives the option to the caller and the caller can decide to abort or just start from the beginning. The caller can also keep track of how often certain errors occur.

Ledger exporter may want to vary its behavior depending on how big the ledger range is. For a full history data lake, it could be better to just abort ledger exporter which will be an immediate signal to the operator that something is wrong and the issue should be investigated. Whereas if ledger exporter swallowed the error and just started ingesting from the genesis ledger, the operator might be surprised  by that behavior.

Then there is also the use case of the GCS ledger backend which is currently in progress in https://github.com/stellar/go/pull/5260 . The ledger backend interface defines the following function:

```go
GetLatestLedgerSequence(ctx context.Context) (uint32, error)
```

So there is the expectation there that errors are propagated rather than swallowed.

exp/services/ledgerexporter/internal/resumablemanager.go
tamirms: sure, that sounds good. please create a new ticket for adding retry logic because, given that a full history data lake will require about 25 GCS requests to find the latest ledger, it can be likely that we encounter a transient error on occasion.

exp/services/ledgerexporter/internal/config.go
sreuland: for now, only have internal usage, I thought about making these public in `historyarchive` package as a client convenience, but yes, will just make private and not try to be anymore reaching than that.

exp/services/ledgerexporter/internal/config.go
sreuland: I commented inline on method with same reasoning.

exp/services/ledgerexporter/internal/exportmanager.go
sreuland: removed - https://github.com/stellar/go/pull/5264/commits/afac63cc83d545981da847835849ab5454185ccb

I think this idiom is bit harder to embrace, interpreting what usage should be, a counter point being its good to program to a contract/interface for encapsulation. The ExportManager interface was used by App to `Run()`, wherein it did not need to know what struct the thing came from.

exp/services/ledgerexporter/internal/resumablemanager.go
sreuland: ah, got it, good point, removed dependency on Config.

exp/services/ledgerexporter/internal/resumablemanager.go
sreuland: since adding error return value, included this as well - https://github.com/stellar/go/pull/5264/commits/7a1231bba2655eb4fb44b654f0d2b18a98551634

exp/services/ledgerexporter/internal/resumablemanager.go
sreuland: https://stellarorg.atlassian.net/browse/HUBBLE-392

exp/services/ledgerexporter/internal/resumablemanager.go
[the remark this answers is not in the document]
tamirms: I think that returning 0 is a bit unintuitive given that the documentation states that FindSearch returns the first absent ledger. Returning 0 is also inconsistent with the behavior of the sort.Search function in the go standard library which always returns a value in the given range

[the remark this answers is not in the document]
sreuland: I can adjust docs to better state intention of the function is to provide facts about the remote data store given requested inputs, but not advice:

```
// Given a requested ledger range consisting of a start and end value, identify the closest ledger number to start but no greater which does not exist on datastore and is also less than or equal to end value.
//
// start - begin search inclusively from this ledger, must be greater than 0.
// end   - stop search inclusively up to this ledger.
//
// If start=0, invalid, error will be returned.
//
// If end=0, is a convenience, to allow requesting an effectively dynamic end value for the range, which will be an approximation of the network's most recent checkpointed ledger + (2 * checkpoint_frequency).
//
// return:
// resumableLedger    - if > 0, will be the next ledger that is not populated on data store within requested range.
// dataStoreComplete - if true, there was no missing ledgers found on data store within the requested range.
// err                              - the search was cancelled due to this error, resumableLedger and dataStoreComplete should be ignored.
//
// when no err, the two return values will compose the following truth table:
//    1. datastore had no data in the requested range: resumableLedger=0, dataStoreComplete=false
//    2. datastore had partial data in the requested range: resumableLedger > 0, dataStoreComplete=false
//    3. datastore had all data in the requested range: resumableLedger=0, dataStoreComplete=true
```
 
with the additional docs, I think `FindStart` defines it's purpose more uniquely to that of remote data conditions, and avoid unintentional semantics to `sort.Search`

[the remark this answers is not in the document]
tamirms: I think you could succinctly describe `FindStart(ctx context.Context, start, end uint32) (absentLedger uint32, ok bool, error)` as:

returns the first absent ledger in the range of `[start, end]`. If `ok` is false then there are no absent ledgers in the range of `[start, end]`.

This spec implies that if the entire `[start, end]` range is absent then the first absent ledger is `start`. I don't see how that is not a fact. But maybe this detail of the API is a matter of taste. I think it's worth getting some other opinions from the platform team

exp/services/ledgerexporter/internal/resumablemanager.go
```diff
@@ -0,0 +1,142 @@
...
+
+	log := logger.WithField("start", start).WithField("end", end).WithField("network", rm.network)
+
+	networkLatest := uint32(0)
+	if end < 1 {
+		var latestErr error
+		networkLatest, latestErr = getLatestLedgerSequenceFromHistoryArchives(rm.archive)
+		if latestErr != nil {
+			err := errors.Wrap(latestErr, "Resumability of requested export ledger range, was not able to get latest ledger from network")
+			log.WithError(err)
```
tamirms, line 71: this line doesn't do anything. I think it should be deleted

exp/services/ledgerexporter/internal/resumablemanager.go
```diff
@@ -0,0 +1,142 @@
...
+			return 0, false, err
+		}
+		logger.Infof("Resumability acquired latest archived network ledger =%d + for network=%v", networkLatest, rm.network)
+		networkLatest = networkLatest + (getHistoryArchivesCheckPointFrequency() * 2)
+		logger.Infof("Resumability computed effective latest network ledger including padding of checkpoint frequency to be %d + for network=%v", networkLatest, rm.network)
```
tamirms, line 76: I don't think we need to log both of these messages. Logging only one of these message should be sufficient

exp/services/ledgerexporter/internal/resumablemanager.go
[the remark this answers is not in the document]
sreuland: seems like we have agreement on functional, just need to close on semantics, using the newly suggested names for the return parameters:

```
FindStart(ctx context.Context, start, end uint32) (absentLedger uint32, ok bool, err error) 
```

@tamirms @urvisavla @chowbao, which sounds best for return values when case of `datastore had no data in the requested range`:

_option (1)_
FindStart returns `absentLedger=0, ok=false, err=nil`

or 

_option (2)_
FindStart returns `absentLedger=<start>, ok=false, err=nil`


the other return value cases are same in either semantic:

`datastore had partial data in the requested range` 
FindStart returns  `absentLedger > 0, ok=false, err=nil`

`datastore had all data in the requested range`
FindStart returns `absentLedger=0, ok=true, err=nil`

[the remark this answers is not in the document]
urvisavla: Another possibility could be to to have `FindStart(start, end uint32) (lastAbsentLedger uint32, err error)`:
1. If there are no ledgers in the specified range, it could `return start, nil`.
2. If all ledgers in the range are present, it could `return end + 1, nil`. The caller would then decide what to do if the  returned value is outside the requested range. 
3. If only a partial range is uploaded, it could `return lastAbsentLedger, nil` where  `lastAbsentLedger` is found by resumability logic.
4. It can `return 0, error` for any error conditions such as a connection error to GCS or history archives (in the case of an unbounded range).
This way we reduce the number of return parameters which could be less confusing.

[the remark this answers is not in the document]
tamirms: @sreuland I would prefer if the ok bool was true that meant the absent ledger was found. This is consistent with the OK comma idiom in golang https://dev.to/saurabh975/comma-ok-in-go-l4f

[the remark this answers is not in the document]
tamirms: @urvisavla I like your suggestion. The only issue I see is that end can be -1 which signals that we don’t know what the end of the range is. In that scenario, according to your spec we should return 0 if there are no absent ledgers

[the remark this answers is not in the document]
urvisavla: In the case where the `end` = -1 (unbounded), since we use the (latest checkpoint ledger on HA * 2) as the max for our binary search, the `lastAbsentLedger` should effectively be the last ledger that was exported so the caller can proceed exporting from the next ledger?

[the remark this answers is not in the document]
sreuland: @urvisavla , thanks for suggestion, I tried that approach on FindStart originally using format similar to your suggestion, noticed similar edge cases like @tamirms  mentioned which results in client code having to do more parsing logic on the various values. whereas the additional bool idiom seems to allow leaner client parsing of results, it sounds like we converge on that and make this return 'absentLedger=start' when no data is present:

```
FindStart(ctx context.Context, start, end uint32) (absentLedger uint32, ok bool, err error) 
```

datastore had partial data in the requested range:
FindStart returns `absentLedger={some value > start}, ok=true, err=nil`

datastore had all data in the requested range:
FindStart returns `absentLedger=0, ok=false, err=nil`

datastore had no data in the requested range:
FindStart returns `absentLedger={start}, ok=true, err=nil`

just give a thumbs up if sounds good, or more context if feel not resolved, thanks!

exp/services/ledgerexporter/internal/resumablemanager.go
```diff
@@ -0,0 +1,139 @@
...
+	// If start=0, invalid, error will be returned.
+	//
+	// If end=0, is provided as a convenience, to allow requesting an effectively
+	// dynamic end value for the range, which will be an approximation of the network's
+	// most recent checkpointed ledger + (2 * checkpoint_frequency).
+	//
+	// return:
+	// absentLedger      - if > 0, will be the oldest ledger sequence between inclusive range of [start, end]
+	//                     which is not populated on data store
+	// ok                - if true, there was no absent ledgers found on data store inclusive to the requested [start, end] range.
```
tamirms, line 28:
```suggestion
	// ok                - if false, there was no absent ledgers found on data store inclusive to the requested [start, end] range.
```

exp/services/ledgerexporter/internal/resumablemanager.go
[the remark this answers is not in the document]
tamirms: @sreuland 👍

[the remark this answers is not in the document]
sreuland: done -  https://github.com/stellar/go/pull/5264/commits/f26af676fadfffaedbd2370951f661b6bca4018e

exp/services/ledgerexporter/internal/resumablemanager.go
```diff
@@ -0,0 +1,139 @@
...
+	// err               - the search was cancelled due to this error, 'absentLedger' and 'ok' return values should be ignored.
+	//
+	// When no error, the two return values will compose the following truth table:
+	//    1. datastore had no data in the requested range: absentLedger=0, ok=false
+	//    2. datastore had partial data in the requested range: absentLedger > 0, ok=false
+	//    3. datastore had all data in the requested range: absentLedger=0, ok=true
```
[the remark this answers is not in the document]
sreuland: thanks, done - https://github.com/stellar/go/pull/5264/commits/f26af676fadfffaedbd2370951f661b6bca4018e

urvisavla (approved):
lgtm!

exp/services/ledgerexporter/internal/queue_test.go
```diff
@@ -56,7 +56,7 @@ func TestQueue(t *testing.T) {
 	require.Equal(t, float64(0), getMetricValue(queue.queueLengthMetric).GetGauge().GetValue())
 	require.Equal(t, uint32(3), l.GetStartLedgerSequence())
 
-	l, ok, err = queue.Dequeue(context.Background())
+	l, _, err = queue.Dequeue(context.Background())
 	require.NoError(t, err)
 	require.False(t, false)
```
tamirms, line 61:
```suggestion
	l, ok, err = queue.Dequeue(context.Background())
	require.NoError(t, err)
	require.False(t, ok)
```

looks like a prexisting bug

exp/services/ledgerexporter/internal/resumablemanager.go
```diff
@@ -0,0 +1,138 @@
...
+	// If start=0, invalid, error will be returned.
+	//
+	// If end=0, is provided as a convenience, to allow requesting an effectively
+	// dynamic end value for the range, which will be an approximation of the network's
+	// most recent checkpointed ledger + (2 * checkpoint_frequency).
+	//
+	// return:
+	// absentLedger      - will be non-zero, the oldest ledger sequence between range of [start, end]
+	//                     which is not populated on data store.
+	// ok                - if true, 'absentLedger' has a usable non-zero value, if false, 'absentLedger' will not be usable, and is set to zero.
```
tamirms, line 27:
```suggestion
	// ok                - if true, 'absentLedger' has a usable non-zero value, if false, there is no absent ledger in the requested range and 'absentLedger' is set to zero.
```

exp/services/ledgerexporter/internal/resumablemanager.go
```diff
@@ -0,0 +1,138 @@
...
+	if lowestAbsentIndex < 1 {
+		// data store had no data within search range
+		return start, true, nil
+	}
+
+	if lowestAbsentIndex < int(rangeSize) {
+		nearestAbsentLedgerSequence := binarySearchStart + uint32(lowestAbsentIndex)
+		log.Infof("Resumability determined next absent object start key of %d for requested export ledger range", nearestAbsentLedgerSequence)
+		return nearestAbsentLedgerSequence, true, nil
+	}
```
tamirms, line 101:
```suggestion
	if lowestAbsentIndex < rangeSize {
		nearestAbsentLedgerSequence := binarySearchStart + uint32(lowestAbsentIndex)
		log.Infof("Resumability determined next absent object start key of %d for requested export ledger range", nearestAbsentLedgerSequence)
		return nearestAbsentLedgerSequence, true, nil
	}
```

no need to handle the `lowestAbsentIndex < 1` case separately. also, I think `rangeSize` is already an int so no need to cast it

exp/services/ledgerexporter/README.md
```diff
@@ -33,17 +33,24 @@ Exports ledgers continuously starting from --start. In this mode, the end ledger
...
+Exporting a ledger range can be optimized further by enabling resumability if the remote data store supports it.
 
-Starts exporting from a specified number of ledgers before the latest ledger sequence number on the network.
-```bash
-ledgerexporter --from-last <number_of_ledgers> --config-file <config_file_path>
-```
+By default, resumability is disabled, `--resume false`
+
+When enabled, `--resume true`, ledgerexporter will search the remote data store within the requested range, looking for the oldest absent ledger sequence number within range. If abscence is detected, the export range is narrowed to `--start <absent_ledger_sequence>`. 
+This feature requires any existing data within the requested range on remote data store to be in an ordered and consistent state, beginning at `--start` and populalated contiguously in increasing order of sequence numbers, without gaps, up to point of abscence or `--end <end>`, whichever comes first or if streaming with `--end 0`.
```
tamirms, line 42:
```suggestion
This feature requires all ledgers to be present on the remote data store for some (possibly empty) prefix of the requested range and then absent for the (possibly empty) remainder.
```

Borrowing some language from https://pkg.go.dev/sort#Search to make the description more concise

tamirms (approved):
looks goof! I left a few minor suggestions, feel free to apply them as you see fit

exp/services/ledgerexporter/internal/resumablemanager.go
sreuland: good catch, cleaned it up more - https://github.com/stellar/go/pull/5264/commits/a60db1963e7ccc03e495da1e99581dd0424499da

exp/services/ledgerexporter/README.md
sreuland: yes, succinct

PR #5265, opened 2024-03-28
exp/services/ledgerexporter: Add prometheus metrics for ledger exporter

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Adds prometheus metrics to the ledger exporter service. The metrics are served via HTTP through an admin port which is specified in the ledger exporter configuration. Some notable metrics included:

- latest ledger sequence consumed by the exporter
- upload duration
- tx meta object size (both compressed and uncompressed)
- number of pending upload tasks

### Why

https://stellarorg.atlassian.net/browse/HUBBLE-277

### Known limitations

[N/A]


exp/services/ledgerexporter/internal/app.go
```diff
@@ -62,17 +95,19 @@ func (a *App) Run() {
...
-
-		err := a.uploader.Run(ctx)
-		if err != nil && !errors.Is(err, context.Canceled) {
-			logger.WithError(err).Error("Error executing Uploader")
-			cancel()
-		}
-	}()
+	wg.Add(1 + a.config.UploadWorkers)
+
+	for i := 0; i < a.config.UploadWorkers; i++ {
```
sreuland, line 100: this wasn't required for metrics correct, is just some tech debt on the side?

tamirms: correct, this change is a side effect of making the queue size configurable. I refactored the queue into its own type to make it easier to instrument metrics and that lead to having upload workers configurable

exp/services/ledgerexporter/internal/exportmanager.go
```diff
@@ -14,35 +17,32 @@ type ExporterConfig struct {
 }
 
 // ExportManager manages the creation and handling of export objects.
-type ExportManager interface {
```
sreuland, line 17: why get rid of the Interface? there was some consistency with Managers expressed as types rather than structs, also, it helps for ongoing unit tests, to be able to mock out.

exp/services/ledgerexporter/internal/queue.go
```diff
@@ -0,0 +1,58 @@
+package ledgerexporter
+
+import (
+	"context"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+// UploadQueue is a queue of LedgerMetaArchive objects which are scheduled for upload
+type UploadQueue struct {
```
sreuland, line 10: this is nice addition, but not sure why is it being added as feature scope to this pr's intent of adding metrics to the existing app, the concerns seem unrelated. I noticed https://stellarorg.atlassian.net/browse/HUBBLE-344 which seem to be the applicable ticket for this parallelization of uploaders enhancement to be done under?

tamirms: I wanted to expose a metric for the size of the queue and it was easier to do that after refactoring the queue to be its own component

exp/services/ledgerexporter/internal/exportmanager.go
tamirms: the export manager interface was not used. 

"Do not define interfaces before they are used" is a recommendation found in the golang style guide https://go.dev/wiki/CodeReviewComments#interfaces

exp/services/ledgerexporter/internal/queue.go
tamirms: the UploadQueue refactoring still does address https://stellarorg.atlassian.net/browse/HUBBLE-344 because I have not done any sort of benchmarking to asses whether having multiple upload workers will be effective when reingesting large ranges. IMO, https://stellarorg.atlassian.net/browse/HUBBLE-344 is still valid even after merging this pr

exp/services/ledgerexporter/internal/uploader.go
```diff
@@ -2,47 +2,116 @@ package ledgerexporter
...
+		},
+		[]string{"already_exists", "ledgers"},
+	)
+	objectSizeMetrics := prometheus.NewSummaryVec(
+		prometheus.SummaryOpts{
+			Namespace: "ledger_exporter", Subsystem: "uploader", Name: "object_size_bytes",
+			Help:       "size of a ledger batch in bytes, sliding window = 10m",
+			Objectives: map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.99: 0.001},
+		},
+		[]string{"ledgers", "already_exists", "compression"},
```
urvisavla, line 41: `already_exists` doesn't seem  to be relevant for `objectSizeMetrics`?

exp/services/ledgerexporter/internal/uploader.go
```diff
@@ -2,47 +2,116 @@ package ledgerexporter
...
+	)
+	prometheusRegistry.MustRegister(uploadDurationMetric)
```
urvisavla, line 43: `objectSizeMetrics` need to be registered as well.

exp/services/ledgerexporter/internal/queue.go
```diff
@@ -0,0 +1,58 @@
...
+	prometheusRegistry.MustRegister(queueLengthMetric)
+	return UploadQueue{
+		metaArchiveCh:     make(chan *LedgerMetaArchive, size),
+		queueLengthMetric: queueLengthMetric,
+	}
+}
+
+// Enqueue will add an upload task to the queue. Enqueue may block if the queue is full.
+func (u UploadQueue) Enqueue(ctx context.Context, archive *LedgerMetaArchive) error {
+	u.queueLengthMetric.Inc()
```
urvisavla, line 32: Should we increment only after the object has been added to the queue on line 34? This is to account queue is full case and adding to the queue is blocked, similar to how it's handled in Dequeue.

exp/services/ledgerexporter/internal/uploader.go
```diff
@@ -2,47 +2,116 @@ package ledgerexporter
...
+		count:  &r.totalCompressed,
+	})
+	r.totalUncompressed += uncompressedCount
+	return uncompressedCount, err
 }
 
 // Upload uploads the serialized binary data of ledger TxMeta to the specified destination.
-// TODO: Add retry logic.
-func (u *uploader) Upload(ctx context.Context, metaArchive *LedgerMetaArchive) error {
+func (u Uploader) Upload(ctx context.Context, metaArchive *LedgerMetaArchive) error {
```
urvisavla, line 79: In addition to the `export_manager_latest_ledger` metric for the last ingested ledger in the export manager it might also be useful to have a metric for the last uploaded ledger sequence number here in the uploader?

exp/services/ledgerexporter/internal/uploader.go
tamirms: if the file already exists the request may not need to upload the entire file to GCS because GCS can detect if the file already exists early on. So this tag is necessary to determine if we are spending full bandwidth to upload files that already exist

exp/services/ledgerexporter/internal/queue.go
tamirms: Yeah, this behavior is actually intentional.

If the channel is full there is still an implicit queue of excess archives that are blocked in this function trying to push to the channel. It is good that this metric will still give us visibility into that case.

exp/services/ledgerexporter/internal/uploader.go
tamirms: I didn't want to implement that in case there were multiple upload workers. With multiple upload workers, each worker would have a different latest ledger. There are different ways to resolve that issue but I think it's better to do it as a part of https://stellarorg.atlassian.net/browse/HUBBLE-344

exp/services/ledgerexporter/internal/app.go
tamirms: I removed the upload workers configuration and hard coded the queue size to 1

exp/services/ledgerexporter/internal/queue.go
tamirms: I removed the upload workers configuration and hard coded the queue size to 1

tamirms:
@sreuland @urvisavla PTAL I believe all the feedback has been addressed

exp/services/ledgerexporter/internal/uploader.go
urvisavla: That's a valid point. Another option would be to add the metric now assuming a single upload worker (as currently implemented) and handle the multiple upload workers in https://stellarorg.atlassian.net/browse/HUBBLE-344.

exp/services/ledgerexporter/internal/app.go
```diff
@@ -97,20 +133,23 @@ func (a *App) Run() {
...
+func mustNewLedgerBackend(ctx context.Context, config Config, prometheusRegistry *prometheus.Registry) ledgerbackend.LedgerBackend {
 	captiveConfig := config.GenerateCaptiveCoreConfig()
 
+	var backend ledgerbackend.LedgerBackend
+	var err error
 	// Create a new captive core backend
-	backend, err := ledgerbackend.NewCaptive(captiveConfig)
+	backend, err = ledgerbackend.NewCaptive(captiveConfig)
 	logFatalIf(err, "Failed to create captive-core instance")
+	backend = ledgerbackend.WithMetrics(backend, prometheusRegistry, "ledger_exporter")
```
urvisavla, line 152: I'm not sure if this is expected or its an issue here or in captive-core, but when I ran ledger exporter, the ledger_exporter_ingest_captive_stellar_core_supported_protocol_version always shows as -1.

```
# HELP ledger_exporter_ingest_captive_stellar_core_supported_protocol_version determines the supported version of the protocol by Captive-Core
# TYPE ledger_exporter_ingest_captive_stellar_core_supported_protocol_version gauge
ledger_exporter_ingest_captive_stellar_core_supported_protocol_version -1

```

urvisavla (approved):
LGTM! Just added a couple of notes for consideration.

exp/services/ledgerexporter/internal/uploader.go
tamirms: I think the metric definition for the most recently uploaded ledger will change in https://stellarorg.atlassian.net/browse/HUBBLE-344 because it will likely require a metric label for the thread / worker id.

exp/services/ledgerexporter/internal/app.go
tamirms: the version is set to -1 when we cannot probe the captive core port:

https://github.com/stellar/go/blob/master/ingest/ledgerbackend/captive_core_backend.go#L241-L243

It looks like this is occurring because we do not configure the captive core http port:

https://github.com/stellar/go/blob/master/exp/services/ledgerexporter/internal/config.go#L182-L191

We should configure the http port as a part of https://stellarorg.atlassian.net/browse/HUBBLE-292 or https://stellarorg.atlassian.net/browse/HUBBLE-271

PR #5270, opened 2024-03-31
ingest: reduce memory consumption of CheckpointChangeReader

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Reduce memory consumption by about 1GB when iterating through ledger entries during state verification and state rebuilds. This reduction in memory is achieved by:

1. Removing claimable balance and offer ledger keys from the visited set when encountering the init entry.
2. Using `map[string]struct{}` to store ledger keys instead of `map[string]bool{}` (see https://stackoverflow.com/questions/57620170/can-you-have-a-map-with-no-value-in-go )

The first optimization takes advantage of the fact that claimable balance and offer ids are globally unique. Once a claimable balance or offer is deleted it is impossible to restore it or create a new claimable balance / offer with the same id. 

### Known limitations

[N/A]


ingest/checkpoint_change_reader.go
```diff
@@ -483,6 +414,14 @@ LoopBucketEntry:
 						return false
 					}
 				}
+			} else if entry.Type == xdr.BucketEntryTypeInitentry && unique {
```
sreuland, line 417: just for understanding of xdr model, how did you know not to include `xdr.BucketEntryTypeLiveentry` in this case?

ingest/checkpoint_change_reader.go
```diff
@@ -483,6 +414,14 @@ LoopBucketEntry:
...
+				// we can remove the ledger key because we know that it's unique in the ledger
+				// and cannot be recreated
+				err := r.tempStore.Remove(h)
+				if err != nil {
+					r.readChan <- r.error(errors.Wrap(err, "Error removing key from tempStore"))
```
sreuland, line 422: is it worth err'ing out, could it suffice to silently continue and log instead?

ingest/checkpoint_change_reader.go
```diff
@@ -483,6 +414,14 @@ LoopBucketEntry:
 						return false
 					}
 				}
+			} else if entry.Type == xdr.BucketEntryTypeInitentry && unique {
+				// we can remove the ledger key because we know that it's unique in the ledger
+				// and cannot be recreated
+				err := r.tempStore.Remove(h)
```
sreuland, line 420: can there be a new assertion added in checkpoint_change_reader_test.go to verify this is being done only for the specific case it relates to?

sreuland (approved):
lgtm, left comment for consideration in adding test coverage if possible

ingest/memory_temp_set.go
```diff
@@ -5,30 +5,31 @@ package ingest
 // state processing. The internal structure is dereferenced after the
 // store is closed.
 type memoryTempSet struct {
-	m map[string]bool
+	m map[string]struct{}
 }
```
Shaptic, line 9: This is what our [`Set[T]` implementation](https://github.com/stellar/go/blob/master/support/collections/set/set.go) is for; I think it can replace this entirely?

ingest/checkpoint_change_reader.go
tamirms: `xdr.BucketEntryTypeLiveentry` implies the ledger entry was updated and therefore there must be a corresponding `xdr.BucketEntryTypeInitentry` for the same ledger entry in an older bucket. So, we cannot remove live entries from the visited set because then we will end up adding them again to the visited set when we encounter the init entry

Issue #5236, opened 2024-03-07
[epic] dry-run horizon database truncation


mollykarcher:
We'll be truncating SDF Horizon's history retention to 1 year later this year. To our knowledge, most partners that enable history retention use 1-3 months of history, so it's possible that there could be issues that only present with this data profile (not retaining full history, but large retention window) that we simply haven't seen or heard of yet.

We should dry-run the truncation and mirror traffic to it for some amount of time, observing the performance impact and resolving any issues that arise from this process. Given the timing and the need to continue using staging to test/issue releases of Horizon prior to the truncation, we should not be doing this on the staging cluster and will need to spin up a new/independent one. 

At a minimum:
- [x] Spin up another staging-like cluster of Horizon (https://github.com/stellar/ops/issues/2900)
- [x] ~Upgrade PostgreSQL 12 :arrow_right: 16 (https://github.com/stellar/go/issues/4831)~
- [x] Enabling reaping on that instance and set the retention to 1 year
  * There are different ways this can be accomplished and we need time to evaluate that. For example, we could turn on reaping on the whole DB and see what happens (which may result in a lockup due to the massive amount of data that needs to be reaped, plus a possible full vacuum) or we could start from scratch and ingest a year+ of data and then enable reaping, or there may be other options.
  * After discussion, it appears we **must** approach this by reaping the whole DB, because reingestion may take on the order of months. The hope :crossed_fingers: is that because the database will be much smaller, the full vacuum will be feasible without any extra operational concerns.
  * some of reaping performance epic, #4870, may be applicable and needed here to ensure reaping exhibits acceptable performance on the truncated 1 year pubnet history db.
  * [x] The periodic reaping frequency should be configurable https://github.com/stellar/go/pull/3823
- [x] Document the operational plan that should be performed to repeat this process for live prd on blue/green prod clusters eventually https://github.com/stellar/go-internal/issues/18
- [ ] Mirror traffic from production to this cluster
- [ ] Brainstorm how we could (or if we need to) simulate load from transaction submission 
- [ ] Observe, identify, and resolve (or defer/prioritize) any performance degradation

sreuland:
@tamirms , what is the latest status on spinning up the test db cluster for this reaping test effort? I think you've mentioned it was in progress but pending due to PG16 issues?

[...]

PR #5272, opened 2024-04-05
services/horizon: Make reaping batch sizes configurable via `--history-retention-reap-count`.

Shaptic:
### What
Introduce a new `--history-retention-reap-count` parameter which will control how many ledgers are reaped at a time when retention (`--history-retention-count`) is enabled.

### Why
Related to #5236: we want to be able to configure this value in case truncation locks up the system with the existing batch sizes.

### Known limitations
TODO: Add a test.


services/horizon/internal/reap/system.go
```diff
@@ -70,19 +70,23 @@ func (r *System) runOnce(ctx context.Context) {
...
-var batchSize = int32(100_000)
+// Current ledger at 2024-04-04s is 51,092,283, so 50k means 1021 batches. At 1
+// batch/second, that seems like a reasonable balance between running under an
+// hour, and slowing it down enough to leave some CPU for other processes.
 var sleep = 1 * time.Second
 
 func (r *System) clearBefore(ctx context.Context, startSeq, endSeq int32) error {
+	batchSize := int32(r.RetentionBatch)
+	if batchSize <= 0 {
+		batchSize = int32(50_000)
```
sreuland, line 87: minor nit for consideration, pull this up to a package const, `reap_batch_default_size = 50_000`

services/horizon/internal/flags.go
```diff
@@ -654,7 +654,15 @@ func Flags() (*Config, support.ConfigOptions) {
...
-			Usage:          "the minimum number of ledgers to maintain within horizon's history tables.  0 signifies an unlimited number of ledgers will be retained",
+			Usage:          "the minimum number of ledgers to maintain within Horizon's history tables (0 = retain an unlimited number of ledgers)",
+			UsedInCommands: IngestionCommands,
+		},
+		&support.ConfigOption{
+			Name:           "history-retention-reap-count",
+			ConfigKey:      &config.HistoryRetentionReapCount,
+			OptType:        types.Uint,
+			FlagDefault:    uint(50_000),
+			Usage:          "the number of ledgers worth of history data to remove per second from the Horizon database",
```
sreuland, line 665: is this accurate description, the reap batch execution time could be greater than 1 second dependent on sql performance, and the reap thread just pauses 1 second in between running each batch sql?

sreuland: should we add some type of validation for a sane maximum value here, 1M?

services/horizon/internal/reap/system.go
tamirms: since you have already configured the default in flags.go I would just return an error in case `batchSize <= 0` because that should be unexpected. you can also add validation in flags.go to ensure the batch size must be at least 1

services/horizon/internal/flags.go
Shaptic: description fixed up, good call! and eh, seems fine to leave it up to the user - you really shouldn't use this unless you know what you're doing

services/horizon/internal/reap/system.go
Shaptic: yeah fair enough, done :+1:

tamirms (approved):

PR #5273, opened 2024-04-08
Bump core to stable v20.4.0

marta-lokhova:


marta-lokhova:
I noticed that a few places are referencing v19 versions of core:
https://github.com/stellar/go/blob/a30c441b6cd6cbcd8eb3749412094422b1a4ea49/exp/tools/dump-ledger-state/Dockerfile#L3
https://github.com/stellar/go/blob/a30c441b6cd6cbcd8eb3749412094422b1a4ea49/.github/workflows/horizon.yml#L130

Assuming it's intended, so I didn't update those since it'd be a major version bump. If I should, please let me know.

tamirms:
@marta-lokhova could you also update https://github.com/stellar/go/blob/a30c441b6cd6cbcd8eb3749412094422b1a4ea49/.github/workflows/horizon.yml#L130
 ?

PR #5275, opened 2024-04-11
Bump XDR for protocol 21

2opremio:


leighmcculloch (approved):

mollykarcher:
Shouldn't we also be updating the integration test [workflows](.github/workflows/horizon.yml) to now test with protocols 20 and 21?

xdr/Stellar-ledger.x
```diff
@@ -484,11 +526,23 @@ struct LedgerCloseMetaV0
...
+
+union LedgerCloseMetaExt switch (int v)
+{
+case 0:
+    void;
+case 1:
+    LedgerCloseMetaExtV1 v1;
+};
+
+struct LedgerCloseMetaV1
```
Shaptic, line 543: Do we need to add support for these new cases in `xdr/ledger_close_meta.go`?

2opremio:
> Shouldn't we also be updating the integration test [workflows](.github/workflows/horizon.yml) to now test with protocols 20 and 21?

Yep, good point. Now that we have a new core build, we can

xdr/Stellar-ledger.x
Shaptic: Ah good point - it's additive so we'd need new wrappers around anything we'd want to extract from the new extension but otherwise nothing needs to be done.

PR #5278, opened 2024-04-16
don't run filtered tmp processor when reingestion is enabled

### LINKED ISSUE #5277: history_transactions_filtered_tmp is updated when not needed, during reingestion

2.29.0

### What did you do?

`stellar horizon db reingest <range>`

during the reingestion runtime, but doesn't add/delete on`history_transactions_filtered_tmp` as that table is only needed for live/forward ingestion to support tx-sub queue processing

### What did you see instead?
reingestion spends cycles on add/delete to`history_transactions_filtered_tmp` during `stellar horizon db reingest <range>` invocations

sreuland:
when a reingestion range is being run, the processor runner was still running `TransactionFilteredTmpProcessor` which is inserting and deleting tx's to `history_transactions_filtered_tmp` on every flush. This step is un-necessary during reingestion since TransactionFilteredTmpProcessor` only needs to be maintained with transactions from 'forward' ingestion to support tx sub queue processing. 

In addtion, during some reingestion runs with parallel workers, observed some long running/hanging DELETE stmts left on the db server which were slowing down overall reingestion workers.

```
postgres=# SELECT
  pid,
  now() - pg_stat_activity.query_start AS duration,
  query,
  state
FROM pg_stat_activity
WHERE (now() - pg_stat_activity.query_start) > interval '5 minutes';
  pid   |    duration     |                                                query                                                 | state
--------+-----------------+------------------------------------------------------------------------------------------------------+--------
 954110 | 17:10:23.48803  | DELETE FROM history_transactions_filtered_tmp WHERE now() >= (created_at + interval '1 second' * $1) | active
 954342 | 17:08:26.893771 | DELETE FROM history_transactions_filtered_tmp WHERE now() >= (created_at + interval '1 second' * $1) | active
 954457 | 17:09:19.293392 | DELETE FROM history_transactions_filtered_tmp WHERE now() >= (created_at + interval '1 second' * $1) | active
 954539 | 17:07:52.870689 | DELETE FROM history_transactions_filtered_tmp WHERE now() >= (created_at + interval '1 second' * $1) | active
 954540 | 17:08:40.926155 | DELETE FROM history_transactions_filtered_tmp WHERE now() >= (created_at + interval '1 second' * $1) | active
 954543 | 17:08:24.030046 | DELETE FROM history_transactions_filtered_tmp WHERE now() >= (created_at + interval '1 second' * $1) | active
 954564 | 17:07:58.474278 | DELETE FROM history_transactions_filtered_tmp WHERE now() >= (created_at + interval '1 second' * $1) | active
 954565 | 17:07:59.878985 | DELETE FROM history_transactions_filtered_tmp WHERE now() >= (created_at + interval '1 second' * $1) | active
```

this behavior may be related to the deadlocks situation that was noted and resolved by https://github.com/stellar/go/pull/5263.

This pr proposes additional changes related to reingestion sql performance on top of the deadlock fix in https://github.com/stellar/go/pull/5263:
* Removed `EnableIngestionFiltering` flag and global config variable, it is obsolete, as ingestion filter was permanently enabled, the flag provides no value.
* repoint conditional checks on running `TransactionFilteredTmpProcessor` to reference `ReingestionEnabled` instead of `EnableIngestionFiltering`, this allows the parallel reingestion workers to skip doing that, thus avoiding that additional sql activity

Closes #5277


services/horizon/internal/ingest/processor_runner.go
```diff
@@ -165,17 +165,15 @@ func (s *ProcessorRunner) buildTransactionProcessor(ledgersProcessor *processors
...
 	return newGroupTransactionFilterers(f)
 }
 
 func (s *ProcessorRunner) buildFilteredOutProcessor() *groupTransactionProcessors {
-	// when in online mode, the submission result processor must always run (regardless of filtering)
 	var p []horizonTransactionProcessor
-	if s.config.EnableIngestionFiltering {
+
+	// when in online mode, the submission result processor must always run (regardless of whether filter rules exist or not)
+	if !s.config.ReingestEnabled {
```
tamirms, line 176: It seems like this is the only place where `ReingestEnabled` is ever used. I think it would be better to delete the `ReingestEnabled` field and instead only include `buildFilteredOutProcessor()` when ingesting through the `RunAllProcessorsOnLedger()` code path. This will allow us to avoid running the the tmp transaction processor in both the `historyRangeState` and the `reingestHistoryRangeState` states.

services/horizon/internal/ingest/processor_runner.go
```diff
@@ -577,8 +575,7 @@ func (s *ProcessorRunner) flushProcessors(groupFilteredOutProcessors *groupTrans
 		defer s.session.Rollback()
 	}
 
-	if s.config.EnableIngestionFiltering {
-
+	if !s.config.ReingestEnabled {
 		if err := groupFilteredOutProcessors.Flush(s.ctx, s.session); err != nil {
 			return errors.Wrap(err, "Error flushing temp filtered tx from processor")
 		}
```
tamirms, line 581: it looks like `s.lastTransactionsTmpGC` is never updated. So we will be executing the delete statement during every call to `flushProcessors()`

sreuland: doh! great catch, it will now be updated once per the interval expiration - https://github.com/stellar/go/pull/5278/commits/4f4a2f092587bd426619a63c31be295914bd9af4

services/horizon/internal/ingest/processor_runner.go
sreuland: yes, good reduction, I added some comments to RunTransactionProcessorsOnLedgers to highlight it's difference now, since it won't update filtered tmp - RunTransactionProcessorsOnLedgers

PR #5280, opened 2024-04-17
ingest: Verify the bucket list hash in CheckpointChangeReader

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

- Refactor the bucket list hash validation when ingesting from history archives so that it occurs within the CheckpointChangeReader
- Validate the bucket list hash when performing state verification

### Why

We currently validate the bucket list hash when ingesting from the history archives in horizon. However, there is a flaw in the validation logic. This is the workflow we were using:

1. In `ProcessRunner` before creating a checkpoint change reader, we [validate](https://github.com/stellar/go/blob/horizon-v2.29.0/services/horizon/internal/ingest/processor_runner.go#L201-L223) the bucket list hash by requesting the [HistoryArchiveState](https://github.com/stellar/go/blob/horizon-v2.29.0/historyarchive/history_archive_state.go#L18-L33) for a given checkpoint ledger from the history archives.
2. In `CheckpointChangeReader` we again send a request to the history archives for the [HistoryArchiveState](https://github.com/stellar/go/blob/horizon-v2.29.0/historyarchive/history_archive_state.go#L18-L33) on the same checkpoint ledger. Then we use [HistoryArchiveState](https://github.com/stellar/go/blob/horizon-v2.29.0/historyarchive/history_archive_state.go#L18-L33) to iterate through all the bucket entries from the history archives.

The problem is that the [HistoryArchiveState](https://github.com/stellar/go/blob/horizon-v2.29.0/historyarchive/history_archive_state.go#L18-L33) response might vary between the two requests even though we are using the same checkpoint ledger. An attacker could theoretically gain control of the history archives and respond with a valid [HistoryArchiveState](https://github.com/stellar/go/blob/horizon-v2.29.0/historyarchive/history_archive_state.go#L18-L33) in the first request and then it could respond with a malicious [HistoryArchiveState](https://github.com/stellar/go/blob/horizon-v2.29.0/historyarchive/history_archive_state.go#L18-L33)  in the second request. In that scenario our bucket list validation would pass and yet we would still ingest malicious ledger entries.

This PR fixes this flaw by ensuring we validate the bucket list hash against the same [HistoryArchiveState](https://github.com/stellar/go/blob/horizon-v2.29.0/historyarchive/history_archive_state.go#L18-L33) used by the `CheckpointChangeReader`.
 

Another benefit of this PR is that we now validate the bucket list hash when performing state verification which allows us to detect if the history archives have been corrupted during state verification.

### Known limitations

[N/A]


ingest/checkpoint_change_reader.go
```diff
@@ -97,6 +99,29 @@ func NewCheckpointChangeReader(
 	}, nil
 }
 
+// VerifyBucketList verifies that the bucket list hash computed from the history archive snapshot
+// associated with the CheckpointChangeReader matches the expectedHash.
+// Assuming expectedHash comes from a trusted source (captive-core running in unbounded mode), this
+// check will give you full security that the data returned by the CheckpointChangeReader can be trusted.
+// Note that XdrStream will verify all the ledger entries from an individual bucket and
```
Shaptic, line 106: hmm :thinking: with this change, is there a need to do entry verification if the bucket is known to be valid?

services/horizon/internal/ingest/history_archive_adapter.go
```diff
@@ -14,10 +14,14 @@ type historyArchiveAdapter struct {
 	archive historyarchive.ArchiveInterface
 }
 
+type verifiableChangeReader interface {
+	ingest.ChangeReader
+	VerifyBucketList(expectedHash xdr.Hash) error
+}
```
Shaptic, line 20: Should this be part of `ingest`?

services/horizon/internal/ingest/verify.go
```diff
@@ -168,6 +174,9 @@ func (s *system) verifyState(verifyAgainstLatestCheckpoint bool) error {
 		return errors.Wrap(err, "Error running GetState")
 	}
 	defer stateReader.Close()
+	if err = stateReader.VerifyBucketList(expectedBucketListHash); err != nil {
+		return ingest.NewStateError(err)
+	}
```
Shaptic, line 179: If bucket list hashes match, is there even a reason to do the remaining state verification? :thinking: or would we need to actually _recreate_ the bucket list and check its hash to be certain of the validity?

Shaptic: Ah, I see this line:

> Another benefit of this PR is that we now validate the bucket list hash when performing state verification which allows us to detect if the history archives have been corrupted during state verification.

so it's like a bidirectional verification now; interesting!

services/horizon/internal/ingest/history_archive_adapter.go
tamirms: this interface is only used by horizon unit tests so there's no need to move it to the public ingest package

ingest/checkpoint_change_reader.go
tamirms: Here's a concrete example, let's say we have two buckets:

Bucket 1 has hash `abcdef` and the following bucket entries:
`e1`, `e2`, `e3`, `e4`


Bucket 2 has has `ghijklm` and the following bucket entries:
`e5`, `e6`, `e7

To verify that the history archive ingestion is correct we need to ensure that:

`hash(e1 + e2 + e3 + e4)` = `abcdef`

AND

`hash(e5 + e6 + e7)` = `ghijklm`

AND

`hash(`abcdef` + `ghijklm`)` = `expectedBucketListHash`

The first two checks encapsulated in `XdrStream` and the last check is encapsulated in the `VerifyBucketList()` function

Shaptic (approved):

PR #5288, opened 2024-04-21
horizon: Clean up integration test variables

2opremio:
All tests use captive-core so there is no need to specify it explicitly.

Soroban-rpc is enabled on a test by test basis so there is no need to guard it through its own env variable.


Shaptic (approved):
deleting code :clap: :clap: :clap:

Issue #5256, opened 2024-03-21
services/horizon: add metrics for ingestions failures and alert on them


mollykarcher:
### What problem does your feature solve?

There's been 2 incidents in the past year where ingestion in horizon halted (for different reasons):
- https://docs.google.com/document/d/1rzuvcXFhXHVf_L_1UBxmCbADD7reEKxNhPnZOj2_ZDg/edit#heading=h.b5xdh739mlq
- https://docs.google.com/document/d/1gnjeNOiQ8W6DNYicT5nkZ_X6f4Zr0Uw6Tdom9Ab6RHw/edit

During the resolution process, we had a lot of general health metric alerts firing but it wasn't immediately clear it was an ingestion halt until we looked at error logs.

### What would you like to see?

An explicit metric that tracks ingestion failures. When a single instance of this occurs, we should alert (critical) on it. This will make it clearer to the responding engineer _immediately_ what the root cause of the issue is.

PR #5302, opened 2024-05-07
services/horizon/ingest: added 'horizon_ingest_errors_total' metric key

sreuland:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

added a new metrics counter key 'horizon_ingest_errors_total' which ingest fsm will emit anytime it traps an error response from in ingestion state run. 

### Why

to configure prometheus alert against the new metric for early visibility on when an ingestion halt starts to form - https://github.com/stellar/prometheus/pull/243

Closes: #5256


tamirms:
We log ingestion errors here:

https://github.com/stellar/go/blob/master/services/horizon/internal/ingest/main.go#L646

If you search on kibana you can see that these logs occur very rarely. I think it would make sense to have an ingestion error counter metric which is basically triggered every time we emit the log in https://github.com/stellar/go/blob/master/services/horizon/internal/ingest/main.go#L646. The metric definition should mirror the log message, so it should be a counter with 2 labels: current_state and next_state.

We can the create the alert using a very simple heuristic: if ingestion error counter is incremented more than `n` times in the last `m` minutes fire the alert

sreuland:
> I think it would make sense to have an ingestion error counter metric which is basically triggered every time we emit the log in https://github.com/stellar/go/blob/master/services/horizon/internal/ingest/main.go#L646.

ah, thanks for pointer to that, updated - https://github.com/stellar/go/pull/5302/commits/3319801ce7c8370db5564811c431e49732fca702

go.mod
```diff
@@ -49,7 +49,7 @@ require (
 	github.com/spf13/viper v1.17.0
 	github.com/stellar/go-xdr v0.0.0-20231122183749-b53fb00bcac2
 	github.com/stellar/throttled v2.2.3-0.20190823235211-89d75816f59d+incompatible
```
tamirms, line 51: are the go mod changes intentional?

services/horizon/internal/ingest/fsm.go
```diff
@@ -405,7 +431,7 @@ func (resumeState) GetState() State {
 	return Resume
 }
 
-func (r resumeState) run(s *system) (transition, error) {
+func (r resumeState) run(s *system) (transitionResult transition, errorResult error) {
```
tamirms, line 434: can you revert this change given that we don't need the named return values anymore?

services/horizon/internal/ingest/main.go
```diff
@@ -443,6 +447,16 @@ func (s *system) initMetrics() {
 		},
 		[]string{"source", "type"},
 	)
+
+	s.metrics.IngestionErrorCounter = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: "horizon", Subsystem: "ingest", Name: "errors_total",
+			Help: "Counters of the number of times the live/forward ingestion state machine encountered an error. " +
+				"'current_state' label has the name of the state that error occurred. " +
```
tamirms, line 455:
```suggestion
				"'current_state' label has the name of the state where the error occurred. " +
```

tamirms (approved):
I left a few minor comments but overall looks good!

go.mod
sreuland: yes, pulled in a race bug fix on `EventuallyWithT`

PR #5303, opened 2024-05-07
Ingestion filtering should use OR logic for rules rather than AND

aditya1702:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

#5304 

### Why

We want transactions to be stored if it matches atleast one of the ingestion filtering rules.

### Known limitations

[TODO or N/A]


services/horizon/internal/ingest/filters/account.go
```diff
@@ -64,3 +63,8 @@ func (f *accountFilter) FilterTransaction(ctx context.Context, transaction inges
 	}
 	return false, nil
 }
+
+func (f accountFilter) IsEmpty(ctx context.Context) bool {
+	// filtering is disabled if the whitelist is empty for now, as that is the only filter rule
+	return len(f.whitelistedAccountsSet) == 0 || !f.enabled
+}
```
JakeUrban, line 70: I don't think we should have a check for whether the filter is enabled within a function called `IsEmpty()`. Is there a name we can use that better reflects the logic it executes?

JakeUrban: Maybe this should just be `isEnabled()`, and we remove the length check on `whitelistedAccountsSet` or `canonicalAssetsLookup`. Length checks can be implemented in the `FilterTransaction()` function.

services/horizon/internal/ingest/group_processors.go
```diff
@@ -179,20 +179,38 @@ func (g *groupTransactionFilterers) Name() string {
 }
 
 func (g *groupTransactionFilterers) FilterTransaction(ctx context.Context, tx ingest.LedgerTransaction) (bool, error) {
+	matchAtleastOneFilter := false
```
sreuland, line 182: for consideration, can condense boolean logic and loop down to this pseudo:

```
filtersEnabled = false

for filter in range filters {
    if !filter.isEnabled()
          continue
    filtersEnabled = true
    include, err := filter.FilterTransaction(tx)
    if err
          return error
    g.AddRunDuration(f.Name(), startTime)
    if include
        return true, nil
}

if filtersEnabled
    g.droppedTransactions++
    return false, nil

return true, nil
```

services/horizon/internal/ingest/group_processors.go
```diff
@@ -179,22 +179,36 @@ func (g *groupTransactionFilterers) Name() string {
...
 	}
+
+	if filtersEnabled {
+		g.droppedTransactions++
+		return false, nil
+	}
 	return true, nil
 }
 
+func (g *groupTransactionFilterers) IsEnabled() bool {
```
sreuland, line 208: ouch, didn't realize the `IsEnabled` pull-up refactor would trickle into here, as this method isn't applicable at the group filter level because the filters are always instantiated and added to it...could just return `true` here and leave it at that, but, beyond this, the method is unused which is more of code smell. 

I have a suggestion to avoid it and still allow callers to reference `IsEnabled` aspect on filters polymorphically:
```
LedgerTransactionFilterer interface {
         Name() string
	FilterTransaction(ctx context.Context, transaction ingest.LedgerTransaction) (include bool, enabled bool, err error)
}
```

which means revert the enabled logic back into `FilterTransaction()` and then callers use the new`enabled` return value from `FilterTransaction()` in place of proposed `filter.IsEnabled()` usage. how does that sound?

services/horizon/internal/ingest/filters/asset_test.go
```diff
@@ -25,11 +25,11 @@ func TestAssetFilterAllowsOnMatch(t *testing.T) {
 	err := filter.RefreshAssetFilter(filterConfig)
 	tt.NoError(err)
 
-	result, err := filter.FilterTransaction(ctx, getAssetTestV1Tx(t, "GD6WNNTW664WH7FXC5RUMUTF7P5QSURC2IT36VOQEEGFZ4UWUEQGECAL"))
+	_, result, err := filter.FilterTransaction(ctx, getAssetTestV1Tx(t, "GD6WNNTW664WH7FXC5RUMUTF7P5QSURC2IT36VOQEEGFZ4UWUEQGECAL"))
```
sreuland, line 28: I think would be worthwhile to assert on `enabled` return value for true?

services/horizon/internal/ingest/filters/asset_test.go
```diff
@@ -47,11 +47,11 @@ func TestAssetFilterAllowsWhenEmptyWhitelist(t *testing.T) {
...
```
sreuland, line 50: same suggestion here, would be worthwhile to assert that `enabled` return value was false since the `Whitelist` is empty.

services/horizon/internal/ingest/filters/asset_test.go
```diff
@@ -69,7 +69,7 @@ func TestAssetFilterAllowsWhenDisabled(t *testing.T) {
...
```
sreuland, line 72: minor suggestion on additional coverage, assert that `enabled` return value is false.

services/horizon/internal/ingest/filters/asset_test.go
aditya1702: Ahhh, I added it to account_test and thought I also added it here

services/horizon/internal/integration/ingestion_filtering_test.go
```diff
@@ -168,3 +169,67 @@ func TestFilteringAssetWhiteList(t *testing.T) {
 	_, err = itest.Client().TransactionDetail(txResp.Hash)
 	tt.NoError(err)
 }
+
+func TestFilteringAssetAndAccountFilters(t *testing.T) {
```
sreuland, line 173: great coverage!

sreuland (approved):
lgtm. left some comments on additional test coverage for consideration.

wanted to also suggest for consideration, would be worthwhile to add a [changelog](https://github.com/aditya1702/go/blob/ingestion-filtering-bug/services/horizon/CHANGELOG.md) entry under `Unreleased` for this so that you can provide a more detailed description for it, as this is changing the inclusion behavior of filters and so, can highlight this as a 'breaking' change under `Unreleased`, otherwise the next release process would probably not pick up on this meaning, and it would just be added under general changes with the pr title.

PR #5305, opened 2024-05-08
ledgerexporter: Add lexicographical ordering token to filenames

chowbao:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

For [HUBBLE-395](https://stellarorg.atlassian.net/browse/HUBBLE-395)
Add a lexicographical ordering token as a prefix to the partition and file names so that the first file within a cloud storage bucket (e.g. GCS) is the largest ledger sequence number.

Example: `/bucket/FFFF05FE-64001-128000/FFFF05FE-64001-64064.xdr.gz`

### Why

The current ordering of partitions and files are integers but the sorting in GCS is alphanumeric making some files appear out of order within GCS.

### Known limitations

This does add more text to the filenames which make it slightly less human readable.

### DIFF
--- a/exp/services/ledgerexporter/internal/exportmanager_test.go
+++ b/exp/services/ledgerexporter/internal/exportmanager_test.go
@@ -2,7 +2,6 @@ package ledgerexporter
 
 import (
 	"context"
-	"fmt"
 	"sync"
 	"testing"
 	"time"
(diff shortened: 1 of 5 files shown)


exp/services/ledgerexporter/internal/exportmanager_test.go
```diff
@@ -138,43 +137,14 @@ func (s *ExportManagerSuite) TestRunWithCanceledContext() {
 	require.EqualError(s.T(), err, "context canceled")
 }
 
-func (s *ExportManagerSuite) TestGetObjectKeyFromSequenceNumber() {
```
chowbao, line 141: Duplicate test: Test was moved to support/datastore/ledgerbatch_config_test.go in a previous PR

support/datastore/ledgerbatch_config_test.go
```diff
@@ -15,16 +15,16 @@ func TestGetObjectKeyFromSequenceNumber(t *testing.T) {
 		fileSuffix        string
```
tamirms, line 15: could you add a test a test where you create a lot of [random](https://pkg.go.dev/math/rand#Uint32) sequence numbers and then assert that the lexicographic order of the object keys matches the descending order of the sequence numbers?

chowbao: Added in [d877c41](https://github.com/stellar/go/pull/5305/commits/d877c41707995951481c7a5cc63839bc92df7a81)

support/datastore/ledgerbatch_config_test.go
```diff
@@ -35,3 +37,41 @@ func TestGetObjectKeyFromSequenceNumber(t *testing.T) {
...
+	}
+
+	sort.Slice(keys, func(i, j int) bool {
+		return keys[i] < keys[j]
+	})
+
+	prev := sequenceMap[keys[0]]
+	for i := 1; i < sequenceCount; i++ {
+		curr := sequenceMap[keys[i]]
+		if prev < curr {
```
tamirms, line 72:
```suggestion
		if prev <= curr {
```

on the off chance that the random generator returns the same value more than once

support/datastore/ledgerbatch_config_test.go
```diff
@@ -35,3 +37,41 @@ func TestGetObjectKeyFromSequenceNumber(t *testing.T) {
...
+			break
+		}
+		randSequence := rand.Uint32()
+		if _, ok := sequenceMap[randSequence]; ok {
+			continue
+		}
+		sequenceMap[randSequence] = config.GetObjectKeyFromSequenceNumber(randSequence)
+		keys = append(keys, randSequence)
+		count++
+	}
```
tamirms, line 63: could you also add 0 and math.MaxUint32 to the set so we cover edge cases?

tamirms (approved):
looks good! I left a couple of minor suggestions for your consideration

PR #5306, opened 2024-05-09
exp/services/ledgerexporter: Change default compression algorithm from gzip to zstd

urvisavla:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
- Set `zstd` as the default compression
- Removed the `filesuffix` configuration parameter and introduced a `GetName()` in the `Compressor` interface and using it for file suffix.

Also, [here](https://stellarorg.atlassian.net/browse/HUBBLE-368?focusedCommentId=16938) is the comparison of evaluated compression algos. 

### Why
Addresses [HUBBLE-368](https://stellarorg.atlassian.net/browse/HUBBLE-368)

### Known limitations
N/A

[HUBBLE-368]: https://stellarorg.atlassian.net/browse/HUBBLE-368?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


support/compressxdr/compressor.go
```diff
@@ -0,0 +1,37 @@
+package compressxdr
+
+import (
+	"github.com/klauspost/compress/zstd"
+	"io"
```
tamirms, line 5: the import order is not correct. here's a link to how to configure the correct order in Goland:

https://github.com/stellar/go/pull/5004#discussion_r1289005979

support/compressxdr/compressor.go
```diff
@@ -0,0 +1,37 @@
...
+	NewWriter(w io.Writer) (io.WriteCloser, error)
+	NewReader(r io.Reader) (io.ReadCloser, error)
+	GetName() string
+}
+
+// ZstdCompressor is an implementation of the Compressor interface for Zstd compression.
+type ZstdCompressor struct{}
+
+// GetName returns the name of the compression algorithm.
+func (z *ZstdCompressor) GetName() string {
```
tamirms, line 21: better to avoid the "Get" prefix https://google.github.io/styleguide/go/decisions#getters

support/compressxdr/compressor.go
```diff
@@ -0,0 +1,37 @@
...
+
+// Compressor represents a compression algorithm.
+type Compressor interface {
+	NewWriter(w io.Writer) (io.WriteCloser, error)
+	NewReader(r io.Reader) (io.ReadCloser, error)
+	GetName() string
+}
+
+// ZstdCompressor is an implementation of the Compressor interface for Zstd compression.
+type ZstdCompressor struct{}
```
tamirms, line 18: since `ZstdCompressor` has no state it's better to make all the functions take value receivers instead of pointer receivers

support/compressxdr/compress_xdr.go
```diff
@@ -3,36 +3,67 @@ package compressxdr
...
 }
 
-func NewXDRDecoder(compressionType string, xdrPayload interface{}) (XDRDecoder, error) {
-	switch compressionType {
-	case GZIP:
-		return &XDRGzipDecoder{XdrPayload: xdrPayload}, nil
-	default:
-		return nil, errors.Errorf("invalid compression type %s, not supported", compressionType)
+// Unzip returns the decompressed bytes.
+func (d *XDRDecoder) Unzip(r io.Reader) ([]byte, error) {
```
tamirms, line 55: I think you can get rid of this function because it's only used in ingest/ledgerbackend/ledger_buffer.go . Instead of calling Unzip, I think you can refactor `getFromLedgerQueue()` to return a  xdr.LedgerCloseMetaBatch instance and you can use a XDRDecoder with a non nil payload to derive the xdr.LedgerCloseMetaBatch value from the compressed bytes

urvisavla: Sure, I'd be happy to. In fact, I made the change initially but I removed it to avoid refactoring to keep the PR short.

exp/services/ledgerexporter/internal/uploader_test.go
```diff
@@ -172,6 +171,7 @@ func (s *UploaderSuite) testUploadPutError(putOkReturnVal bool) {
 	dataUploader := NewUploader(&s.mockDataStore, queue, registry)
 	err := dataUploader.Upload(context.Background(), archive)
 	require.Equal(s.T(), fmt.Sprintf("error uploading %s: error in PutFileIfNotExists", key), err.Error())
+	decoder := compressxdr.NewXDRDecoder(compressxdr.DefaultCompressor, nil)
```
tamirms, line 174: you don't need to create a decoder here. Whenever `nil` is passed into `compressxdr.NewXDRDecoder()` it indicates that the decoder is not necessary. A few lines below instead of having:

```go
for _, compression := range []string{decoder.Compressor.Name(), "none"} { 
```

you can do:

```go
for _, compression := range []string{compressxdr.DefaultCompressor.Name(), "none"} {
```

ingest/ledgerbackend/ledger_buffer.go
```diff
@@ -203,7 +202,7 @@ func (lb *ledgerBuffer) storeObject(ledgerObject []byte, sequence uint32) {
 	}
 }
 
-func (lb *ledgerBuffer) getFromLedgerQueue(ctx context.Context) ([]byte, error) {
+func (lb *ledgerBuffer) getFromLedgerQueue(ctx context.Context) (*xdr.LedgerCloseMetaBatch, error) {
```
tamirms, line 205: this should return a `xdr.LedgerCloseMetaBatch` value instead of a pointer

ingest/ledgerbackend/ledger_buffer.go
```diff
@@ -218,13 +217,14 @@ func (lb *ledgerBuffer) getFromLedgerQueue(ctx context.Context) ([]byte, error)
 			// len(taskQueue) + len(ledgerQueue) + ledgerPriorityQueue.Len() <= bsb.config.BufferSize
 			lb.pushTaskQueue()
 
-			reader := bytes.NewReader(compressedBinary)
-			lcmBinary, err := lb.decoder.Unzip(reader)
+			lcmBatch := &xdr.LedgerCloseMetaBatch{}
+			decoder := compressxdr.NewXDRDecoder(compressxdr.DefaultCompressor, lcmBatch)
```
tamirms, line 221:
```suggestion
			lcmBatch := xdr.LedgerCloseMetaBatch{}
			decoder := compressxdr.NewXDRDecoder(compressxdr.DefaultCompressor, &lcmBatch)
```

support/compressxdr/compress_xdr.go
```diff
@@ -3,36 +3,50 @@ package compressxdr
...
 
-const (
-	GZIP = "gzip"
-)
+func NewXDREncoder(compressor Compressor, xdrPayload interface{}) XDREncoder {
+	return XDREncoder{Compressor: compressor, XdrPayload: xdrPayload}
+}
+
+func NewXDRDecoder(compressor Compressor, xdrPayload interface{}) XDRDecoder {
+	return XDRDecoder{Compressor: compressor, XdrPayload: xdrPayload}
```
tamirms, line 14:
```suggestion
func NewXDRDecoder(compressor Compressor, xdrPayload interface{}) *XDRDecoder {
	return &XDRDecoder{Compressor: compressor, XdrPayload: xdrPayload}
```

`NewXDRDecoder()` should return a pointer because `ReadFrom(r io.Reader)` modifies `XdrPayload`

support/compressxdr/compress_xdr.go
```diff
@@ -3,36 +3,50 @@ package compressxdr
...
 }
 
-func NewXDREncoder(compressionType string, xdrPayload interface{}) (XDREncoder, error) {
-	switch compressionType {
-	case GZIP:
-		return &XDRGzipEncoder{XdrPayload: xdrPayload}, nil
-	default:
-		return nil, errors.Errorf("invalid compression type %s, not supported", compressionType)
+// WriteTo writes the XDR compressed encoded data
+func (e *XDREncoder) WriteTo(w io.Writer) (int64, error) {
```
tamirms, line 25:
```suggestion
func (e XDREncoder) WriteTo(w io.Writer) (int64, error) {
```
`xdr3.Marshal()` should not modify `e.XdrPayload` so this function does not require a pointer receiver

support/compressxdr/compress_xdr.go
tamirms: actually, I'm sorry I think `NewXDRDecoder()` should remain as is (return a value instead of a pointer). I didn't consider that `XdrPayload` is actually a pointer value. However, the `ReadFrom()` function on `XDRDecoder` should be a value receiver and not a pointer receiver

exp/services/ledgerexporter/internal/uploader.go
```diff
@@ -83,15 +83,10 @@ func (u Uploader) Upload(ctx context.Context, metaArchive *datastore.LedgerMetaA
...
-	// JIRA https://stellarorg.atlassian.net/browse/HUBBLE-368
-	xdrEncoder, err := compressxdr.NewXDREncoder(compressxdr.GZIP, &metaArchive.Data)
-	if err != nil {
-		return err
-	}
+	xdrEncoder := compressxdr.NewXDREncoder(compressxdr.DefaultCompressor, &metaArchive.Data)
 
 	writerTo := &writerToRecorder{
-		WriterTo: xdrEncoder,
+		WriterTo: &xdrEncoder,
```
tamirms, line 89:
```suggestion
		WriterTo: xdrEncoder,
```
passing in a pointer is not necessary here

support/compressxdr/compress_xdr.go
```diff
@@ -3,36 +3,50 @@ package compressxdr
...
-	default:
-		return nil, errors.Errorf("invalid compression type %s, not supported", compressionType)
+// XDRDecoder combines decompression with XDR decoding
+type XDRDecoder struct {
+	Compressor Compressor
+	XdrPayload interface{}
+}
+
+// ReadFrom reads XDR compressed encoded data
+func (d *XDRDecoder) ReadFrom(r io.Reader) (int64, error) {
```
tamirms, line 43:
```suggestion
func (d XDRDecoder) ReadFrom(r io.Reader) (int64, error) {
```
this doesn't need to be a pointer receiver because it does not modify the XDRDecoder  struct

tamirms (approved):
there are just 2 small issues to fix but looks good apart from that!

PR #5307, opened 2024-05-09
ingest/ledgerbackend: Improve thread-safety of stellarCoreRunner.close()

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

While looking at ingestion logs I observed an unexpected error `error preparing range: error starting prepare range: the previous Stellar-Core instance is still running` which occurred when we rolled out a new version of core:

```
 level=warning msg="detected new version of captive core binary /usr/bin/stellar-core , aborting session." service=ingest subservice=stellar-core
 level=info msg="default: got signal 2, shutting down" service=ingest subservice=stellar-core
 level=info msg="Processed ledger" commit=true duration=0.732382788 ledger=true sequence=51570776 service=ingest state=true
 level=info msg="Ingestion system state machine transition" current_state="resume(latestSuccessfullyProcessedLedger=51570775)" next_state="resume(latestSuccessfullyProcessedLedger=51570776)" service=ingest level=error msg="Error in ingestion state machine" current_state="resume(latestSuccessfullyProcessedLedger=51570776)" error="error preparing range: error starting prepare range: the previous Stellar-Core instance is still running" next_state=start service=ingest level=info msg="default: Application destructing" service=ingest subservice=stellar-core
 level=info msg="default: Application destroyed" pid=31974 service=ingest subservice=stellar-core
```

The error occurred in this block of code:

https://github.com/stellar/go/blob/ab3a9265a31d7a86a794d447061652ab53cc3237/ingest/ledgerbackend/captive_core_backend.go#L456-L460

I was surprised that stellar-core was still running even though we just called `c.stellarCoreRunner.close()`. I expected the `close()` function on stellarCoreRunner to block until the core process was reaped. However, it turns out that is not the case in this scenario because the `close()` function is called by multiple go routines.

Once the new core binary is detected, the file watcher go routine will first call `close()` https://github.com/stellar/go/blob/ab3a9265a31d7a86a794d447061652ab53cc3237/ingest/ledgerbackend/file_watcher.go#L66 . This invocation will block until the core process is reaped. However, it takes a few seconds for that to happen. So, while that is occurring PrepareRange() calls `close()` and that returns immediately based on this condition:

https://github.com/stellar/go/blob/ab3a9265a31d7a86a794d447061652ab53cc3237/ingest/ledgerbackend/stellar_core_runner.go#L555-L559

This PR fixes this issue by guaranteeing that the core process will be reaped after any invocation of `close()` , even if `close()` is called concurrently. This behavior is implemented by using https://pkg.go.dev/sync#Once

### Why

This bug does not prevent ingestion from progressing because upon errors we will still retry to ingest and eventually `PrepareRange()` is able to succeed once the core process is reaped. However, it is still good to get rid of this error case for the following reasons:

- reduce unnecessary delays caused by ingestion retries
- reduce error messages in the ingestion logs
- we will soon alert on ingestion errors so it's good to remove cases which should not actually be errors

### Known limitations

[N/A]


ingest/ledgerbackend/stellar_core_runner_test.go
```diff
@@ -149,6 +150,61 @@ func TestCloseOnlineWithError(t *testing.T) {
 	assert.NoError(t, runner.close())
 }
 
+func TestCloseConcurrency(t *testing.T) {
```
sreuland, line 153: nice test coverage

sreuland (approved):

PR #5310, opened 2024-05-10
support/datastore: Add CRC32C validation when exporting and consuming tx meta files

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add CRC32C validation validation when uploading and downloading tx meta files.

<del>https://stellarorg.atlassian.net/browse/HUBBLE-271</del>
https://stellarorg.atlassian.net/browse/HUBBLE-278

### Why

From https://cloud.google.com/storage/docs/hashes-etags

> There are a variety of ways that data can be corrupted while uploading to or downloading from the Cloud:
 > - Memory errors on client or server computers, or routers along the path
 > - Software bugs (e.g., in a library that customers use)
> - Changes to the source file when an upload occurs over an extended period of time

> Cloud Storage supports two types of hashes you can use to check the integrity of your data: CRC32C and MD5. CRC32C is the [recommended validation method](https://cloud.google.com/storage/docs/hashes-etags#validation) for performing integrity checks.

### Known limitations

Unfortunately, the CRC32 hash must be computed at the beginning of the HTTP request which means we must write out the payload to a temporary buffer instead of computing the hash on the fly.

### DIFF
--- a/support/datastore/gcs_datastore.go
+++ b/support/datastore/gcs_datastore.go
@@ -66,11 +66,11 @@ func NewGCSDataStore(ctx context.Context, params map[string]string, network stri
 // GetFile retrieves a file from the GCS bucket.
 func (b GCSDataStore) GetFile(ctx context.Context, filePath string) (io.ReadCloser, error) {
 	filePath = path.Join(b.prefix, filePath)
-	// setting ReadCompressed(true) will avoid transcoding of compressed files
-	// by including an "Accept-Encoding: gzip" header to the request:
+	// setting ReadCompressed(true) will avoid transcoding of compressed files by including
+	// an "Accept-Encoding: gzip" header in the request:
 	// https://github.com/googleapis/google-cloud-go/blob/main/storage/http_client.go#L1307-L1309
 	// https://cloud.google.com/storage/docs/transcoding#decompressive_transcoding
-	// This will ensure that the reader performs CRC validation when finishing the download:
+	// This will ensure that the reader performs CRC validation upon finishing the download:
 	// https://pkg.go.dev/cloud.google.com/go/storage#Reader
 	r, err := b.bucket.Object(filePath).ReadCompressed(true).NewReader(ctx)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


2opremio (approved):

sreuland:
@tamirms , I made small edit on description, I think the intention was to reference [HUBBLE-278](https://stellarorg.atlassian.net/browse/HUBBLE-278)  for crc check.

tamirms:
@sreuland my bad, thanks for correcting the mistake!

support/datastore/gcs_datastore.go
```diff
@@ -64,7 +66,13 @@ func NewGCSDataStore(ctx context.Context, params map[string]string, network stri
...
 func (b GCSDataStore) GetFile(ctx context.Context, filePath string) (io.ReadCloser, error) {
 	filePath = path.Join(b.prefix, filePath)
-	r, err := b.bucket.Object(filePath).NewReader(ctx)
+	// setting ReadCompressed(true) will avoid transcoding of compressed files
+	// by including an "Accept-Encoding: gzip" header to the request:
+	// https://github.com/googleapis/google-cloud-go/blob/main/storage/http_client.go#L1307-L1309
+	// https://cloud.google.com/storage/docs/transcoding#decompressive_transcoding
+	// This will ensure that the reader performs CRC validation when finishing the download:
+	// https://pkg.go.dev/cloud.google.com/go/storage#Reader
+	r, err := b.bucket.Object(filePath).ReadCompressed(true).NewReader(ctx)
```
sreuland, line 75: since change means will not be compatible with existing exported files, it's still pre-release, but should this breaking change be mentioned through any dev channels for those that have generated exports, i.e. need to rerun ledgerexporter correct?

sreuland (approved):

support/datastore/gcs_datastore.go
tamirms: @sreuland I don't think this is a breaking change because the documentation on https://cloud.google.com/storage/docs/hashes-etags#crc32c states:

> All Cloud Storage objects have a [CRC32C hash](https://datatracker.ietf.org/doc/html/rfc4960#appendix-B).

So, even if we don't send a CRC32C during the upload process, GCS will still compute the CRC32C hash and record the result as part of its metadata. This can be verified by uploading an object to a GCS bucket wiithout including the CRC32C hash in the upload request and then verifying that there is still a CRC32C hash in the GCS object's metadata.

The reason we include the CRC32C in the upload request in `putFile()` is so that we can validate that the object we upload on GCS has the same serverside hash that we computed client side

sreuland: ah, got it, thanks for clarification, didn't realize gcs was doing crc by default already, so `ReadCompressed` will work on any existing exported files.

PR #5311, opened 2024-05-13
exp/services/ledgerexporter: refactoring LedgerMetaArchive

urvisavla:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Refactoring and enhancements in LedgerMetaArchive:
-  Moved `createLedgerCloseMeta` test helper function out of LedgerCloseMetaArchive.
- Added unit tests for LedgerCloseMetaArchive `GetLedger` method.
- Removed getter methods from LedgerCloseMetaArchive to align with go convention.


exp/services/ledgerexporter/internal/exportmanager_test.go
```diff
@@ -15,8 +15,27 @@ import (
 	"github.com/stellar/go/ingest/ledgerbackend"
 	"github.com/stellar/go/support/collections/set"
 	"github.com/stellar/go/support/datastore"
+	"github.com/stellar/go/xdr"
 )
 
+func createLedgerCloseMeta(ledgerSeq uint32) xdr.LedgerCloseMeta {
```
chowbao, line 21: Just curious. When is it better to create public/private helper function for testing? Like in this case is creating `createLedgerCloseMeta` here and in the ledgerbackend test preferred? Or should could have it just been moved to like support/datastore/mocks.go

sreuland: that's a good call out, I don't have ref to any posts that suggest idiom or convention on that, but over time, you can see how shared infrastructure in tests like utility funcs lead to tighter coupling between tests, i.e brittleness, and you end up having to change unrelated tests occasionally b/c your changing some unrelated but shared test infra instead and it starts to get complicated and become an app of it's own, so the tradeoff on some duplicate test setup code but encapsulated to only the test that needs is ok.

sreuland (approved):

exp/services/ledgerexporter/internal/exportmanager_test.go
urvisavla: Yeah basically, as Shawn mentioned, a tradeoff to minimize coupling.

support/datastore/ledger_meta_archive.go
```diff
@@ -44,53 +44,19 @@ func (f *LedgerMetaArchive) AddLedger(ledgerCloseMeta xdr.LedgerCloseMeta) error
...
-}
-
-// GetObjectKey returns the object key of the archive.
-func (f *LedgerMetaArchive) GetObjectKey() string {
-	return f.ObjectKey
-}
-
+// GetLedger retrieves the LedgerCloseMeta for a given sequence number.
+// It returns an error if the sequence number is invalid or if the LedgerCloseMeta for the sequence number is not found.
 func (f *LedgerMetaArchive) GetLedger(sequence uint32) (xdr.LedgerCloseMeta, error) {
```
tamirms, line 49: I think it is more appropriate for this function to be defined on `xdr.LedgerCloseMetaBatch` instead of `LedgerMetaArchive`. There are several cases of implementing helper functions on xdr types, see xdr/account_id.go for example. So you could add a xdr/ledger_close_meta_batch.go file and include  a `GetLedger(sequence uint32) (xdr.LedgerCloseMeta, error)` implementation there.

tamirms: Also, if you implement `GetLedger()` function on `xdr.LedgerCloseMetaBatch` then you should change the buffered storage backend to store a `xdr.LedgerCloseMetaBatch`  instance instead of a `LedgerMetaArchive` field. The buffered storage backend ignores the `ObjectKey` field completely and does all of its operations on the `Data` field on its `LedgerMetaArchive` instance.

urvisavla: Thanks for the suggestion. I didn't realize extending the xdr was an option... duh. I'll moved both `AddLedger` and `GetLedger` to xdr.LedgerCloseMeta which makes `LedgerMetaArchive` unnecessary. But we still need a structure to encapsulate filename, object data and metadata(?) for the upload queue, so I'll moved it back to the ledger exporter/internal for now. 

Also, I'm thinking if we should store compressed binary data in `LedgerMetaArchive` instead of xdr for efficient memory usage. This will be particularly useful if we queue up multiple objects for upload. That is a bigger change though so it can be done it in a separate PR.

tamirms: > But we still need a structure to encapsulate filename, object data and metadata(?) for the upload queue, so I'll moved it back to the ledger exporter/internal for now.

that makes sense to me. I don't anticipate any package except for ledger exporter using `LedgerMetaArchive`. When consuming the tx meta files you'll be parsing the file contents into xdr.LedgerCloseMeta instances.

> Also, I'm thinking if we should store compressed binary data in LedgerMetaArchive instead of xdr for efficient memory usage. This will be particularly useful if we queue up multiple objects for upload. 

this sounds like a good idea, 👍

tamirms (approved):

PR #5312, opened 2024-05-14
support/datastore: Fix putFile in gcs_datastore  to correctly upload to the bucket

urvisavla:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Fix issue in `putFile` of `gcs_datastore` to correctly write the object to the bucket.

### Why
Without this fix, empty files were uploaded to the datastore.

### Known limitations
N/A

### DIFF
--- a/support/datastore/gcs_datastore.go
+++ b/support/datastore/gcs_datastore.go
@@ -162,7 +162,7 @@ func (b GCSDataStore) putFile(ctx context.Context, filePath string, in io.Writer
 	w.SendCRC32C = true
 	// we must set CRC32C before invoking w.Write() for the first time
 	w.CRC32C = crc32.Checksum(buf.Bytes(), crc32.MakeTable(crc32.Castagnoli))
-	if _, err := in.WriteTo(w); err != nil {
+	if _, err := io.Copy(w, buf); err != nil {
 		return errors.Wrapf(err, "failed to put file: %s", filePath)
 	}
 	return w.Close()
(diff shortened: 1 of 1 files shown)


support/datastore/gcs_datastore.go
```diff
@@ -162,7 +162,7 @@ func (b GCSDataStore) putFile(ctx context.Context, filePath string, in io.Writer
 	w.SendCRC32C = true
 	// we must set CRC32C before invoking w.Write() for the first time
 	w.CRC32C = crc32.Checksum(buf.Bytes(), crc32.MakeTable(crc32.Castagnoli))
-	if _, err := in.WriteTo(buf); err != nil {
+	if _, err := in.WriteTo(w); err != nil {
```
tamirms, line 165:
```suggestion
	if _, err := io.Copy(w, buf); err != nil {
```

Another issue is that `in.WriteTo()` is executed twice. that has a side effect of over counting the number of bytes / compressed bytes in the metrics for the uploader:

https://github.com/stellar/go/blob/master/exp/services/ledgerexporter/internal/uploader.go#L65-L78

If we copy from `buf` to `w` that should fix the bug without calling `in.WriteTo()` twice

tamirms:
Sorry, I introduced this bug in https://github.com/stellar/go/pull/5310 . I can work on a test to catch this regression

support/datastore/gcs_datastore.go
tamirms: In general, it seems like we shouldn't assume that `in.WriteTo()` can be called consecutively without some call to rewind the WriterTo instance back to the beginning:

https://groups.google.com/g/golang-nuts/c/cHOwJJMZfyo

sreuland: given this type of regression that wasn't caught, this may be a good time to consider adding gcs_datastore_test.go and at least one unit test on PutFile?

tamirms: @sreuland yeah, I completely agree. I created https://github.com/stellar/go/pull/5313 to address the lack of testing on the GCS datastore.

tamirms (approved):

urvisavla:
Closing this PR since the followup [PR](https://github.com/stellar/go/pull/5313) also contains the fix.

PR #5313, opened 2024-05-14
support/datastore: Add tests for GCS datastore

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add unit tests for the GCS datastore using a GCS emulator.

Note this PR requires that https://github.com/stellar/go/pull/5312 is merged first.

### Why

The GCS datastore is completely untested and https://github.com/stellar/go/pull/5310 introduced a [bug](https://github.com/stellar/go/pull/5312).

### Known limitations

[N/A]


support/datastore/datastore.go
```diff
@@ -26,7 +26,11 @@ type DataStore interface {
 func NewDataStore(ctx context.Context, datastoreConfig DataStoreConfig, network string) (DataStore, error) {
 	switch datastoreConfig.Type {
 	case "GCS":
-		return NewGCSDataStore(ctx, datastoreConfig.Params, network)
+		destinationBucketPath, ok := datastoreConfig.Params["destination_bucket_path"]
```
sreuland, line 29: what was reason to move gcs config handling up into the factory rather then leaving it encapsulated in  `NewGCSDataStore()` ? ideally, each type specific datastore impl, such as gcs_datastore.go maintains the know-how for parsing/validating it's own config values from generic key-values of `datastoreConfig.Params`, such as `destination_bucket_path` for gcs.

tamirms: There are several reasons motivating this change:

- it's not clear to the caller of `NewGCSDataStore()` what parameters it expects because you have a map of params which makes it opaque. In order for me to understand what key / values I need to include in the map I need to look inside the implementation of `NewGCSDataStore()`.
- If you make a typo and misspell "destination_bucket_path" or some other key required in the parameters map you will get a runtime error. In this PR we eliminate this class of error.
- if we ever change `NewGCSDataStore()` to require configuration parameters which aren't naturally strings we end up losing type safety because every configuration parameter has to fit in the params map[string]string container. Also if you encode the configuration parameter into a string in the wrong way that will trigger a runtime error. If you instead represent the configuration parameter using its natural type you eliminate this class of error.

I understand that the motivation for using the parameters map container was to make it easy to pass through configuration from the ledger exporter toml file directly to the constructor of the data store. However, that optimizes the developer experience of `NewGCSDataStore()` for the ledger exporter toml parsing use case at the expense of other use cases (e.g. unit tests or constructing a GCS data store for the buffered ledger backend).

I think we can structure the code so that it suits all use cases. We can make `NewGCSDataStore()` take well defined positional arguments with proper types (e.g. what it looks like in this PR). Then for the ledger exporter toml file use case we can have a function which translates the configuration embedded in the toml file into expected arguments and then create the data store instance. Ideally this function should live in the ledger exporter package.

sreuland (approved):
great test coverage!

Issue #5295, opened 2024-04-26
services/horizon: update bucketlistdb config sent to core


mollykarcher:
### What problem does your feature solve?

Core is [upgrading bucketlist db](https://github.com/stellar/stellar-core/issues/4202) from experimental status into default backend status (see [blog draft](https://docs.google.com/document/d/1yoJ06YUYj40CXO_-e-M-zXtHMUMa1w4is7UwPmnfGSs/edit) for details). This means that they are going to be removing flag `EXPERIMENTAL_BUCKETLIST_DB` and replacing it with a new/non-experimental flag. At the time of writing this issue there are a couple different proposed names for this new flag (`USE_BUCKETLIST_DB` and `USE_DEPRECATED_SQL`), so confirm with core on the name before implementing. 

### What would you like to see?

We should to send the new field, and stop sending the old one. This is debatable, but I don't think we should expose a flag for this Horizon-side at all, non-bucketlist will eventually be removed core-side, and bucketlist has been default enabled in captive core for over a year. We've tended to overindex on choice/configurability, and this has proved to only confuse operators, so I'd prefer to limit options going forward. This would mean that we should:
- [ ] Remove Horizon's `EXPERIMENTAL_BUCKETLIST_DB` configuration option
    - One could argue that we should deprecate this first, but it has "experimental" in the name so I actually think it's okay to just kill it off without warning. Open to differing opinions though
- [ ] When `CAPTIVE_CORE_USE_DB=true`, send `USE_DEPRECATED_SQL=false`. 
    - We deprecated `CAPTIVE_CORE_USE_DB` in the 2.30 release, so we'll eventually remove it entirely (but wait for a protocol boundary) and the `USE_DEPRECTED_SQL=false` will be hardcoded/not configurable
    - Figure out if version-gating, similar to what currently exists for `EXPERIMENTAL_BUCKETLIST_DB`, is necessary. Core exits when it sees config flags it doesn't recognize. The first stable version of core for p21 will include this new flag, _but_ that won't be available until the testnet vote is successful, so it's possible for people to run core v20.* with 2.30.* of Horizon and see this issue.

The captive core toml creation + default setting is shared between Horizon + RPC, so the same fix should work for RPC but we should be mindful that we'll need to update/release both.

PR #5323, opened 2024-05-21
Bump core to the stable v21 build.

dmkozh:
I've also updated some v19 images which I suspect weren't being used (or at least the tests were set up with protocol 19).


sreuland (approved):
this required more changes than probably was desired, as the core toml config bumped into larger scope of #5295 for bucketlist config cleanup, had to do some adjustments in here related to that scope to get integration passing, this seems ok in terms of limiting this scope to just getting the newer version in and #5295 still needs to be done.

PR #5324, opened 2024-05-28
exp/services/ledgerexporter: Add metadata to exported files

urvisavla:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

The following metadata added to the exported files in GCS:

```
	StartLedger          uint32
	EndLedger            uint32
	StartLedgerCloseTime int64
	EndLedgerCloseTime   int64
	ProtocolVersion      uint32
	CoreVersion          string
	Network              string
	CompressionType      string
	Version              string
```

- `CoreVersion` and `ProtocolVersion` are parsed from the output of the `stellar-core version` command.
- Build process does not include versioning yet, so lexie `Version` will default to "develop". Versioning of Lexie is out of scope for this PR.
- `Network` will be either "pubnet" or "testnet".
- `CompressionType` is the name of the compression algorithm used which also used as the file extension.

Datatstore interface changes:
- Put interface modified to accept a map of strings as input which holds the metadata.
- Get interface returns metadata as a map of strings. The usage of this metadata is not yet defined so the ledgerbackend currently ignores it.

Field names:
```
start-ledger
end-ledger
start-ledger-close-time
end-ledger-close-time
protocol-version
core-version
network
compression-type
version
```

### Why

Addresses [HUBBLE-279](https://stellarorg.atlassian.net/browse/HUBBLE-279)

### Known limitations

[HUBBLE-279]: https://stellarorg.atlassian.net/browse/HUBBLE-279?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


support/datastore/datastore.go
```diff
@@ -14,9 +14,9 @@ type DataStoreConfig struct {
 
 // DataStore defines an interface for interacting with data storage
 type DataStore interface {
-	GetFile(ctx context.Context, path string) (io.ReadCloser, error)
-	PutFile(ctx context.Context, path string, in io.WriterTo) error
-	PutFileIfNotExists(ctx context.Context, path string, in io.WriterTo) (bool, error)
+	GetFile(ctx context.Context, path string) (io.ReadCloser, map[string]string, error)
```
urvisavla, line 17: I've modified the datastore Get and Put interfaces to use map[string]string to keep to generic. But I'm wondering if it would be better to pass `Metadata` (or  `LedgerMetaArchive`) instead  to encapsulate the metadata structure?

exp/services/ledgerexporter/internal/app.go
```diff
@@ -91,6 +91,16 @@ func NewApp(flags Flags) *App {
 	return app
 }
 
+// Version gets the version of the app from environment variable "LEXIE_VERSION" which is set during the build process.
+// If the environment variable is not set, it defaults to "develop".
+func Version() string {
```
tamirms, line 96: for horizon and soroban-rpc we use go linker flags to set the version string:

https://stackoverflow.com/questions/47509272/how-to-set-package-variable-using-ldflags-x-in-golang-build

Here's how it's done in soroban-rpc:

https://github.com/stellar/soroban-rpc/blob/main/Makefile#L17-L20

exp/services/ledgerexporter/internal/config.go
```diff
@@ -151,6 +161,54 @@ func (config *Config) GenerateCaptiveCoreConfig() (ledgerbackend.CaptiveCoreConf
...
+	}
+
+	re := regexp.MustCompile(`ledger protocol version: (\d+)`)
+	for _, row := range rows[1:] {
+		if matches := re.FindStringSubmatch(row); len(matches) > 1 {
+			c.ProtocolVersion = matches[1]
+			break
+		}
+	}
+	if c.ProtocolVersion == "" {
```
tamirms, line 205: the protocol version should be extracted from the ledger:

https://github.com/stellar/soroban-rpc/blob/main/cmd/soroban-rpc/internal/methods/simulate_transaction.go#L289

There is a corner case where a ledger file can span more than protocol version if it contains ledgers both before and after a protocol upgrade. I think in that scenario you can just use the protocol version of the latest ledger contained within the ledger file

support/datastore/gcs_datastore.go
```diff
@@ -59,8 +59,14 @@ func FromGCSClient(ctx context.Context, client *storage.Client, bucketPath strin
 }
 
 // GetFile retrieves a file from the GCS bucket.
-func (b GCSDataStore) GetFile(ctx context.Context, filePath string) (io.ReadCloser, error) {
+func (b GCSDataStore) GetFile(ctx context.Context, filePath string) (io.ReadCloser, map[string]string, error) {
```
tamirms, line 62: I don't think we should include metadata in `GetFile()` because that will incur an extra HTTP request. For users which ignore metadata, this extra request will increase latency. I think it would be better to add a separate function just for querying metadata.

exp/services/ledgerexporter/internal/app.go
urvisavla: Yes, I'm familiar with this but as mentioned in the description, I'm leaving out build change for a separate PR.

tamirms: ok but it seems like nothing is setting the  `LEXIE_VERSION` environment variable currently so I am not sure why we read the variable here. The workflow for soroban-rpc does not use environment variables

support/datastore/metadata.go
```diff
@@ -0,0 +1,79 @@
...
+	if val, ok := data["x-lexie-end-ledger-close-time"]; ok {
+		endLedgerCloseTime, err := strconv.ParseInt(val, 10, 64)
+		if err != nil {
+			return metaData, err
+		}
+		metaData.EndLedgerCloseTime = endLedgerCloseTime
+	}
+
+	if val, ok := data["x-lexie-protocol-version"]; ok {
+		protocolVersion, err := strconv.ParseInt(val, 10, 64)
```
tamirms, line 66:
```suggestion
		protocolVersion, err := strconv.ParseInt(val, 10, 32)
```

support/datastore/metadata.go
```diff
@@ -0,0 +1,79 @@
...
+	ProtocolVersion      uint32
+	CoreVersion          string
+	Network              string
+	CompressionType      string
+	Version              string
+}
+
+func (m MetaData) ToMap() map[string]string {
+	return map[string]string{
+		"x-lexie-start-ledger":            strconv.FormatUint(uint64(m.StartLedger), 10),
```
tamirms, line 19: the `x-lexie-` prefix is unnecessary because we don't need to disambiguate our metadata from some other metadata which is present on the object. The only custom metadata set on the object is the one we provide in the upload request.

exp/services/ledgerexporter/internal/config.go
```diff
@@ -151,6 +160,35 @@ func (config *Config) GenerateCaptiveCoreConfig() (ledgerbackend.CaptiveCoreConf
 	}, nil
 }
 
+// By default, it points to exec.Command, overridden for testing purpose
+var execCommand = exec.Command
+
+// Executes the "stellar-core version" command and parses its output to extract
+// the core version
+// The output of the "version" command is expected to be a multi-line string where the
```
chowbao, line 168: Do you think core would ever provide a json or other format of the stellar-core version output? Doing the regex below feels bad but there isn't an alternative right now.

exp/services/ledgerexporter/internal/uploader_test.go
```diff
@@ -45,6 +46,39 @@ func (s *UploaderSuite) TestUpload() {
...
+	}
+	metadata := datastore.MetaData{
+		StartLedger:          start,
+		EndLedger:            end,
+		StartLedgerCloseTime: 123456789,
+		EndLedgerCloseTime:   987654321,
+		ProtocolVersion:      3,
+		CoreVersion:          "v1.2.3",
+		Network:              "testnet",
+		CompressionType:      "gzip",
```
chowbao, line 63: Should CompressionType be `zstd`?

support/datastore/metadata.go
```diff
@@ -0,0 +1,79 @@
+package datastore
+
+import "strconv"
+
+type MetaData struct {
+	StartLedger          uint32
+	EndLedger            uint32
+	StartLedgerCloseTime int64
+	EndLedgerCloseTime   int64
+	ProtocolVersion      uint32
```
chowbao, line 10: nit: ProtocolVersion in your PR description says it is a string. Just confirming you do want it as uint32

exp/services/ledgerexporter/internal/config.go
urvisavla: > Do you think core would ever provide a json or other format of the stellar-core version output?

Don't know, but that would be nice..

exp/services/ledgerexporter/internal/uploader_test.go
urvisavla: The test data is completely random, but I tried to keep it meaningful.

support/datastore/metadata.go
urvisavla: Thanks. Updated the PR description. Since it is read from XDR, it will be an int, so I had changed it from a string to an int.

exp/services/ledgerexporter/internal/app.go
urvisavla: Added versioning  for docker build in eb8f17bb92a3dabc7c09a503d7f4a588d2788fcc. I wasn't sure about the release strategy for the exporter so for now prefixing the git hash with "1.0.0-".  @chowbao  mentioned that he wants to start filling the data lake soon so this will ensure that the exported files have a version that is easy to track.

exp/services/ledgerexporter/Makefile
```diff
@@ -2,12 +2,13 @@ SUDO := $(shell docker version >/dev/null 2>&1 || echo "sudo")
...
 BUILD_DATE := $(shell date -u +%FT%TZ)
-VERSION ?= $(shell git rev-parse --short HEAD)
+VERSION ?= 1.0.0-$(shell git rev-parse --short HEAD)
 DOCKER_IMAGE := stellar/ledger-exporter
 
 docker-build:
 	cd ../../../ && \
 	$(SUDO) docker build --platform linux/amd64 --pull --label org.opencontainers.image.created="$(BUILD_DATE)" \
+    --build-arg GOFLAGS="-ldflags=-X=github.com/stellar/go/exp/services/ledgerexporter/internal.version=$(VERSION)" \
 	--build-arg VERSION=$(VERSION) \
```
tamirms, line 12: is `--build-arg VERSION=$(VERSION)` used anywhere? I can't find any references in the docker file

tamirms (approved):

PR #5326, opened 2024-05-30
support/datastore: Make resumability robust to unexpected overlaps in adjacent ranges

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Adjacent ranges may end up overlapping due to the clamping behavior in [`adjustLedgerRange()`](https://github.com/stellar/go/blob/fff01229a5af77dee170a37bf0c71b2ce8bb8474/exp/services/ledgerexporter/internal/config.go#L173-L192)

For example, assuming 64 ledgers per file, [2, 100] and [101, 150] get adjusted to [2, 127] and [64, 191]

If we export [64, 191] and then try to resume on [2, 127], the binary search logic will determine that [2, 127] is fully exported because:

1. the midpoint [2, 127] is 64
2. ledger 64 is present on the data store given that we already exported [64, 191]

This behavior will therefore return an incorrect result if ledgers [2, 63] are missing from the data store.

We can fix this issue by querying the end ledger and, if it is present, only do the binary search on all the preceding files. So, in the example above, when exporting [2, 127], we will:

1. check if ledger 127 is present on the data store
2. since the ledger is present we will only do the binary search on the range of [2, 63] instead of [2, 127]
3. If the ledger file corresponding to [2, 63] is missing, resumability will correctly determine we must start from ledger 2

Note that if there is an overlap in adjacent ranges caused by [`adjustLedgerRange()`](https://github.com/stellar/go/blob/fff01229a5af77dee170a37bf0c71b2ce8bb8474/exp/services/ledgerexporter/internal/config.go#L173-L192), the size of the overlap will never be larger than the number of files per partition and that is why it is sufficient to only check if the end ledger is present.

### Known limitations

[N/A]

### DIFF
--- a/support/datastore/resumablemanager_test.go
+++ b/support/datastore/resumablemanager_test.go
@@ -5,8 +5,9 @@ import (
 	"testing"
 
 	"github.com/pkg/errors"
-	"github.com/stellar/go/historyarchive"
 	"github.com/stretchr/testify/require"
+
+	"github.com/stellar/go/historyarchive"
 )
 
 func TestResumability(t *testing.T) {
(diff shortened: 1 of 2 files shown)


support/datastore/resumeablemanager.go
```diff
@@ -77,6 +78,29 @@ func (rm resumableManagerService) FindStart(ctx context.Context, start, end uint
 			return 0, false, errors.Errorf("Invalid start value of %v, it is greater than network's latest ledger of %v", start, networkLatest)
 		}
 		end = networkLatest
+	} else if end >= rm.ledgerBatchConfig.LedgersPerFile {
+		// Adjacent ranges may end up overlapping due to the clamping behavior in adjustLedgerRange()
+		// https://github.com/stellar/go/blob/fff01229a5af77dee170a37bf0c71b2ce8bb8474/exp/services/ledgerexporter/internal/config.go#L173-L192
+		// For example, assuming 64 ledgers per file, [2, 100] and [101, 150] get adjusted to [2, 127] and [64, 191]
+		// If we export [64, 191] and then try to resume on [2, 127], the binary search logic will determine that
+		// [2, 127] is fully exported because the midpoint of [2, 127] is present.
```
chowbao, line 86: What happens if we only export [64, 127] and then try to resume on [2, 191]? This would skip [2, 63] right?
Or this this edge case out of scope as it'd probably be better suited for a future `detect_gaps` feature?

tamirms: Resuming on [2, 191] would lead to undefined behavior if we have already exported [64, 127] . In this case the ledger ranges are not really adjacent since [64, 127] is a subset of [2, 191]. The aim of this PR is to allow people to export adjacent ranges without unexpected behaviors. After this PR is merged, resumability will still not work when there are gaps

support/datastore/resumablemanager_test.go
```diff
@@ -61,6 +62,42 @@ func TestResumability(t *testing.T) {
...
+			networkName: "test",
+		},
+		{
+			name:         "ledger range overlaps with an range which is already exported",
+			startLedger:  2,
+			endLedger:    127,
+			absentLedger: 2,
+			findStartOk:  true,
+			ledgerBatchConfig: LedgerBatchConfig{
+				FilesPerPartition: uint32(1000),
```
chowbao, line 96: nit: FilesPerPartition should be 100 to be consistent with the other tests

chowbao (approved):

PR #5328, opened 2024-05-31
services/horizon: Add hash to the txsub timeout response

aditya1702:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

- Change the txsub timeout description to suggest polling the GET endpoint instead of resubmission.
- Add the transaction hash to the response

### Why

We want users to favor polling instead of resubmitting on a timeout. This is because the `504 Timeout` does not necessarily mean the transaction has failed and resubmission can lead to `tx_bad_seq` errors, which has happened with horizon operators before.

### Known limitations

NA


services/horizon/internal/actions/submit_transaction.go
```diff
@@ -191,6 +191,17 @@ func (handler SubmitTransactionHandler) GetResource(w HeaderWriter, r *http.Requ
 		if r.Context().Err() == context.Canceled {
 			return nil, hProblem.ClientDisconnected
 		}
-		return nil, hProblem.Timeout
+		return nil, &problem.P{
```
JakeUrban, line 194: Shouldn't we be updating the `hProblem.Timeout` rather than defining a custom `problem.P`? Or is that not possible since the response includes the specific hash and envelope?

services/horizon/internal/actions/submit_transaction.go
```diff
@@ -191,6 +191,17 @@ func (handler SubmitTransactionHandler) GetResource(w HeaderWriter, r *http.Requ
...
+			Type:   "transaction_submission_timeout",
+			Title:  "Transaction Submission Timeout",
+			Status: http.StatusGatewayTimeout,
+			Detail: "Your transaction submission request has timed out. This does not necessarily mean the submission has failed. " +
+				"Before resubmitting, please use the transaction hash provided in `extras.hash` to poll the GET /transactions endpoint for sometime and " +
+				"check if it was included in a ledger.",
+			Extras: map[string]interface{}{
+				"hash":         info.hash,
+				"envelope_xdr": raw,
```
JakeUrban, line 203: Whats the rationale for providing the envelope in the response? I think we can assume the client has the envelope since it was required to get this response in the first place.

services/horizon/internal/actions/submit_transaction.go
aditya1702: @JakeUrban The `hProblem.Timeout` is also used in other places. I think a more customised txsub timeout response is better suited here. Since we can update the description to direct users to poll the hash.

services/horizon/internal/actions/submit_transaction.go
aditya1702: @JakeUrban Yes that is true, however I wanted to keep it similar to the other problem responses we send for txsub. Then we should also remove it from the other responses too I think.

tamirms (approved):

PR #5332, opened 2024-06-03
Expose captive core version details

psheth9:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Exposes captive core version details in captive-core-backend. This change is prerequisite for fixing https://github.com/stellar/soroban-rpc/issues/161

### Why

Currently getNetowrk and getVersionInfo endpoint calls /info on core client, which in theory can semi-block the core thread. We need to refactor captive core implementation to expose captive core version so that soroban-rpc does not fetch it using http client. (i.e cache it at core process level)

### Known limitations

N/A


ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -245,6 +246,21 @@ func (c *CaptiveStellarCore) coreVersionMetric() float64 {
 	return float64(info.Info.ProtocolVersion)
 }
 
+func (c *CaptiveStellarCore) setCoreVersion() {
+	if c.stellarCoreClient == nil {
+		log.WithField("err", errors.New("HTTP stellarCoreClient is not initialized")).Error("cannot fetch captive core version details")
+		return
+	}
+
+	info, err := c.stellarCoreClient.Info(c.config.Context)
```
psheth9, line 255: I think it is safe to add exponential backoff here but wanted to confirm with you @tamirms

tamirms, line 255: it would be better to obtain the version by executing the core binary with the version command. You can see https://github.com/stellar/go/pull/5324 for an example of how that's done. Also the version should be refreshed every time the stellarCoreRunnerFactory() is invoked

ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -211,7 +214,7 @@ func NewCaptive(config CaptiveCoreConfig) (*CaptiveStellarCore, error) {
 			URL: fmt.Sprintf("http://localhost:%d", config.Toml.HTTPPort),
 		}
 	}
-
+	c.setCoreVersion()
```
tamirms, line 217: this also needs to be called when captive core restarts. If captive core restarts we will create a new stellar core runner instance

tamirms:
There is no function for accessing the version.

psheth9:
Fixed the review comments Thank you @tamirms :)

ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -245,6 +249,35 @@ func (c *CaptiveStellarCore) coreVersionMetric() float64 {
...
+
+// Executes the "stellar-core version" command and parses its output to extract
+// the core version
+// The output of the "version" command is expected to be a multi-line string where the
+// first line is the core version in format "vX.Y.Z-*".
+func (c *CaptiveStellarCore) setCoreVersion() {
+	versionCmd := execCommand(c.config.BinaryPath, "version")
+	versionOutput, err := versionCmd.Output()
+	if err != nil {
+		log.Errorf("failed to execute stellar-core version command: %s", err)
```
tamirms, line 263: we should use the logger in `c.config`

tamirms (approved):

PR #5333, opened 2024-06-06
ingest/ledgerbackend: Update captive-core config with new BucketlistDB parameters

urvisavla:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
- Adds the new stellar-core [BucketListDB](https://github.com/stellar/stellar-core/issues/4202) related config parameter DEPRECATED_SQL_LEDGER_STATE. It also addresses the EXPERIMENTAL_BUCKETLIST_DB parameter, which will be completely removed in the next Stellar-core release, making DEPRECATED_SQL_LEDGER_STATE mandatory.

- Horizon currently has a an option called CAPTIVE_CORE_USE_DB (to be deprecated) to use captive-core in in-memory mode. The default value of DEPRECATED_SQL_LEDGER_STATE is determined based on this flag: true for in-memory mode and false for the on-disk (preferred) mode. This check will not be necessary once the CAPTIVE_CORE_USE_DB flag is fully removed.

- It also updates the parameter names: EXPERIMENTAL_BUCKETLIST_DB_INDEX_PAGE_SIZE_EXPONENT to BUCKETLIST_DB_INDEX_PAGE_SIZE_EXPONENT and EXPERIMENTAL_BUCKETLIST_DB_INDEX_CUTOFF to BUCKETLIST_DB_INDEX_CUTOFF

*This should not be merged until protocol 21 has been voted on and everyone has had the chance to update to stellar-core 21.*

### Why

Addresses https://github.com/stellar/go/issues/5295

### Known limitations

*This PR completely removes the EXPERIMENTAL_BUCKETLIST_DB flag, which we currently use in our deployment. This means we will need to update our deployment scripts as part of the releasing this change.*


ingest/ledgerbackend/toml.go
```diff
@@ -538,17 +538,16 @@ func (c *CaptiveCoreToml) setDefaults(params CaptiveCoreTomlParams) {
 		checkCoreVersionF = checkCoreVersion
 	}
 	currentCoreVersion := checkCoreVersionF(params.CoreBinaryPath)
```
tamirms, line 540: do we still need to check the core version when generating the captive core toml file?

ingest/ledgerbackend/toml.go
```diff
@@ -642,10 +641,8 @@ func (c *CaptiveCoreToml) validate(params CaptiveCoreTomlParams) error {
 		)
 	}
 
-	if def := c.tree.Has("EXPERIMENTAL_BUCKETLIST_DB"); def && !params.UseDB {
-		return fmt.Errorf(
-			"BucketListDB enabled in captive core config file, requires Horizon flag --captive-core-use-db",
-		)
+	if def := c.tree.Has("DEPRECATED_SQL_LEDGER_STATE"); def && params.UseDB && *c.DeprecatedSqlLedgerState {
```
tamirms, line 644: what if `DEPRECATED_SQL_LEDGER_STATE` is set to false and UseDB is also false?

ingest/ledgerbackend/toml.go
urvisavla: We decided not to maintain backward compatibility for the new parameters, so I didn't add a version check but the core version check is still used [here](https://github.com/stellar/go/blob/97a97d88d32b8b1f86e8accbf903fb0465370928/ingest/ledgerbackend/toml.go#L585), which is why this code is not removed.

ingest/ledgerbackend/toml.go
```diff
@@ -23,6 +23,8 @@ var (
 
 	//go:embed configs/captive-core-testnet.cfg
 	TestnetDefaultConfig []byte
+
+	DefaultBucketListDBPageSize uint = 12
```
sreuland, line 27: this could be private correct?

ingest/ledgerbackend/toml.go
sreuland: good catch, that case is another error that should be trapped, it could be slipping by now even with current EXPERIMENTAL_BUCKETLIST_DB

ingest/ledgerbackend/toml.go
sreuland: yes, looks like it's still used to determine soroban diagnostic event enablement , [ln 585](https://github.com/stellar/go/pull/5333/files#diff-c7702ef819e52917ed3868588dff5273774c61a6aa8d36c1644d7ba251fb7756R585)

tamirms: @urvisavla if we are not maintaining backwards compatibility with protocol versions older than 21 then I don't think we need to check the core version in the code you linked to because we can assume that the core must be >= 21.0.0

services/horizon/CHANGELOG.md
```diff
@@ -10,8 +10,12 @@ file. This project adheres to [Semantic Versioning](http://semver.org/).
 - Change ingestion filtering logic to store transactions if any filter matches on it. ([5303](https://github.com/stellar/go/pull/5303))
   - The previous behaviour was to store a tx only if both asset and account filters match together. So even if a tx matched an account filter but failed to match an asset filter, it would not be stored by Horizon.  
 
-## 2.30.0
+- Captive-core configuration parameters updated to align with [stellar-core v21](https://github.com/stellar/stellar-core/issues/3811) ([5333](https://github.com/stellar/go/pull/5333))
+  - BucketlistDB is now the default database for stellar-core, deprecating `EXPERIMENTAL_BUCKETLIST_DB`.
+  - A new mandatory parameter `DEPRECATED_SQL_LEDGER_STATE` (default: false) will be added automatically, no action required.
```
sreuland, line 15:
```suggestion
  - A new mandatory parameter `DEPRECATED_SQL_LEDGER_STATE` (default: false) is required by stellar-core on its configuration toml file. if the toml provided by `CAPTIVE_CORE_CONFIG_PATH` does not have this new setting, Horizon will add it automatically, therefore, no action required.
```

sreuland (approved):
nice work, left minor comments for consideration.

ingest/ledgerbackend/toml.go
tamirms: @urvisavla I think it makes sense to return an error so the operator is aware that explicitly setting `DEPRECATED_SQL_LEDGER_STATE` to false is not compatible with in-memory mode

ingest/ledgerbackend/toml.go
urvisavla: I was a bit hesitant to remove this check for soroban diagnostics because I'm unsure of the possible impact on rpc and rpc tests..

tamirms: I don't think there would be a negative impact on rpc because once protocol 21 is live it will not be possible to run rpc with a version of core older than 21.0.0

urvisavla: Ok, that makes sense. There are a few other things that can be cleaned up when we remove the version check, so I'm creating a separate PR https://github.com/stellar/go/pull/5346.

tamirms (approved):

sreuland:
since `EXPERIMENTAL_BUCKETLIST_DB` is now removed, it effectively means any horizon ingestion node will only work when captive core bin on host is min v21 or higher. So, should this include a sanity check do be run as part of captive core init of horizon ingestion startup, to fail fast with error on command line, if captive core bin version is less than v21? via [toml.go coreVersion.IsEqualOrAbove](https://github.com/stellar/go/blob/827c6b04d37446fe8081fbd0844dc13ebc8d89a1/ingest/ledgerbackend/toml.go#L462)

PR #5335, opened 2024-06-07
exp/services/ledgerexporter: refactor ledgerexporter cli into sub-commands

sreuland:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

* changed the cli interface to accept new sub-commands and removed --resume flag:

  `ledgerexporter scan-and-fill --start --end --config-file` - exports the entire bounded requested range, only the ledgers which are missing are uploaded to the datastore.

  `ledgerexporter append --start --end --config-file` - export ledgers beginning with dynamically determined starting point based on binary search across `--start` and `--end` for the first absent ledger with that range.

* updated config settings now allow for disabling testnet/pubnet preconfigured settings for captive core and providing a custom external captive core config file instead via `stellar_core_config.captive_core_toml_path`

* this also includes a breaking change on toml config:
  ```
  network_name = "testnet"  
  
  [datastore_config]
  type = "GCS"
  
  [datastore_config.params]
  destination_bucket_path = "your-bucket-name/<optional_subpaths>"
  
  [datastore_config.schema]
  ledgers_per_file = 64
  files_per_partition = 10
  file_suffix = ".xdr.gz"
  
  # preconfigured captive config
  [stellar_core_config]
    stellar_core_binary_path = "/my/path/to/stellar-core"
    preconfigured_network = "testnet" # Options: `testnet` or `pubnet`
  
  or
  
  # manual  captive config
  [stellar_core_config]
    captive_core_toml_path = "my-captive-core.cfg"
    stellar_core_binary_path = "/my/path/to/stellar-core"
    history_archive_urls = ["http://testarchiveurl1", "http://testarchiveurl2"]
    network_passphrase = "test"
  
  ```

### Why

* sub-commands conveniently encapsulate lower level aspects of export outcomes and performance to make it best outcome for user automatically, such as using 'resume'. 

* operators needed ability to override the preconfigured testnet/pubnet captive core settings.

Closes [hubble-271](https://stellarorg.atlassian.net/browse/HUBBLE-271)


exp/services/ledgerexporter/README.md
```diff
@@ -21,44 +21,54 @@ type LedgerCloseMetaBatch struct {
...
 type = "GCS"
 
 [datastore_config.params]
 destination_bucket_path = "your-bucket-name/<optional_subpaths>"
 
-[exporter_config]
+[datastore_config.schema]
 ledgers_per_file = 64
 files_per_partition = 10
+file_suffix = ".xdr.gz"
```
chowbao, line 53: I thought we got rid of the `file_suffix` in favor of only allowing one compression alg? @urvisavla to confirm?

exp/services/ledgerexporter/README.md
```diff
@@ -21,44 +21,54 @@ type LedgerCloseMetaBatch struct {
...
+  stellar_core_binary_path = "/my/path/to/stellar-core"
+  preconfigured_network = "testnet" # Options: `testnet` or `pubnet`
+
+or
+
+[stellar_core_config]
+  captive_core_toml_path = "my-captive-core.cfg"
+  stellar_core_binary_path = "/my/path/to/stellar-core"
+  history_archive_urls = ["http://testarchiveurl1", "http://testarchiveurl2"]
+  network_passphrase = "test"
```
chowbao, line 65: If you pass the `captive-core.cfg` you can ignore setting the other vars right?

exp/services/ledgerexporter/README.md
sreuland: oh, this might have been a bad merge conflict resolution I did earlier when pulling latest master in. looks like I should remove this.

exp/services/ledgerexporter/README.md
```diff
@@ -21,44 +21,54 @@ type LedgerCloseMetaBatch struct {
...
+[datastore_config.schema]
 ledgers_per_file = 64
 files_per_partition = 10
+file_suffix = ".xdr.gz"
+
+[stellar_core_config]
+  stellar_core_binary_path = "/my/path/to/stellar-core"
+  preconfigured_network = "testnet" # Options: `testnet` or `pubnet`
+
+or
```
tamirms, line 59: to the reader the "or" appears to be part of the toml configuration. I think you should end the code block here and then add some text explaining that instead of using preconfigured_network you can specify a captive_core_toml_path, history_archive_urls, and network_passphrase.

exp/services/ledgerexporter/config.toml
```diff
@@ -1,15 +1,17 @@
-network = "testnet"
+network_name = "testnet"
```
tamirms, line 1: so long as we're exposing a network field here, could we get rid of preconfigured_network and instead use this network_name filed?

exp/services/ledgerexporter/README.md
sreuland: no, following same pattern as horizon ingest config for captive which requires these values be separately defined also and I think it has been driven that way through underlying ingest sdk's ledgerbackend captive core config params interface  [ledgerbackend.CaptiveCoreTomlParams](https://github.com/sreuland/go/blob/lex_config/exp/services/ledgerexporter/internal/config.go#L189) which needs these values separately from the core cfg, as it internally creates a connection to archive server urls as part of init/range setup before it even attempts to run a captive core with the cfg.

the same parameters are also used directly within ledgerexporter to configure history archive client so that it can query archive server urls for init aspects, such as latest checkpointed ledger to adjust search ranges for resume.

I think your assumption does highlight a nice optimization, to glean the passphrase and history archive urls from the supplied captive core file instead, but that would not be trivial, as would need to parse the `Validators` toml array in cfg, and then further parse the 'curl' command strings on the `Validator.History` param to get just the archive url sub-string, which is all that's needed for the go archive pool config. which is probably why the ingest sdk went with a more simplistic approach to get it done.

exp/services/ledgerexporter/internal/config.go
```diff
@@ -16,20 +17,31 @@ import (
...
+const (
+	Pubnet    = "pubnet"
+	Testnet   = "testnet"
+	UserAgent = "ledgerexporter"
+)
+
+type Mode int
+
+const (
+	ScanFill Mode = iota
```
tamirms, line 32:
```suggestion
	_ Mode = iota
	ScanFill Mode = iota
```

it's better to start at 1 so that the default value (0) can be invalid. This will allow us to distinguish the case where a Mode was explicitly set to `ScanFill` vs being unassigned completely

exp/services/ledgerexporter/internal/main.go
```diff
@@ -0,0 +1,94 @@
...
+
+	rootCmd.AddCommand(scanAndFillCmd)
+	rootCmd.AddCommand(appendCmd)
+
+	scanAndFillCmd.PersistentFlags().Uint32("start", 2, "Starting ledger (inclusive), must be set to a value greater than 1")
+	scanAndFillCmd.PersistentFlags().Uint32("end", 0, "Ending ledger (inclusive), must be set to value greater than 'start' and less than 'latest checkpoint ledger from network history archives + (2 * checkpoint ledger frequency)'")
+	scanAndFillCmd.PersistentFlags().String("config-file", "config.toml", "Path to the TOML config file. Defaults to 'config.toml' on runtime working directory path.")
+	viper.BindPFlags(scanAndFillCmd.PersistentFlags())
+
+	appendCmd.PersistentFlags().Uint32("start", 2, "Starting ledger (inclusive), must be set to a value greater than 1")
```
tamirms, line 70: can we make the start parameter required?

exp/services/ledgerexporter/internal/main.go
```diff
@@ -0,0 +1,94 @@
...
+			)
+			settings.Mode = Append
+			return ledgerExporterCmdRunner(settings)
+		},
+	}
+
+	rootCmd.AddCommand(scanAndFillCmd)
+	rootCmd.AddCommand(appendCmd)
+
+	scanAndFillCmd.PersistentFlags().Uint32("start", 2, "Starting ledger (inclusive), must be set to a value greater than 1")
```
tamirms, line 65: There is a ConfigOption type declared in the support package which is used by horizon and other services in the go monorepo:

https://github.com/stellar/go/blob/f30d11432e81c7a7cbb739a694520f729bbb31dd/support/config/config_option.go#L22

Would we benefit from using ConfigOptions in ledger exporter?

exp/services/ledgerexporter/config.toml
sreuland: this is an opaque label which user can set to anything. since `stellar_core_config.preconfigured_network` is optional now, and could only have been testnet/pubnet, but with external captive cfg now, the network could be anything like futurenet, standalone. And the notion of a network name was still needed within ledgerexporter, as it's used by Datastore config such as for how GCSDataStore uses network name to build bucket paths, and ResumeManager uses in minor way for logging contexts. I think it could be organized better, what if I move it out of general config and into the stellar network config:

```
[stellar_core_config]
  network_name = "whatever"
  ... 
```

tamirms: I think having network_name and preconfigured_network both part of `[stellar_core_config]` would be confusing. If the only use case for `network_name` is to build bucket paths, I think it would be better to remove that config field entirely and the operator can decide whether they want to include the network name in the `destination_bucket_path` field

exp/services/ledgerexporter/internal/main.go
sreuland: `end` and `start` are validated specific to the mode(append or scan-and-fill) further down in app init [ValidateAndSetLedgerRange](https://github.com/sreuland/go/blob/lex_config/exp/services/ledgerexporter/internal/config.go#L97), and would bubble back up to here as an error to trigger cmd to stop. was trying to keep the flags layer thin, no logic.

exp/services/ledgerexporter/internal/main.go
sreuland: I thought about pulling that in, but decided in favor of (re)writing to just what's needed for simplicity at this point, ConfigOptions is a pretty intense abstraction with many features that likely not used.

exp/services/ledgerexporter/config.toml
sreuland: ok, yeah, I like the idea of pushing that down into gcs config parameter only then, will do. thanks.

exp/services/ledgerexporter/internal/main.go
tamirms: I think that validation would work for end. But, if `start` is omitted the default value would be 2 which would not trigger an error in [ValidateAndSetLedgerRange](https://github.com/sreuland/go/blob/lex_config/exp/services/ledgerexporter/internal/config.go#L97)

exp/services/ledgerexporter/README.md
sreuland: yes, cleaned that up - https://github.com/stellar/go/pull/5335/commits/9d98a7c35985de5d03a273aee38c0888b9948f48, thx

exp/services/ledgerexporter/internal/main.go
sreuland: ah, so subtle, nice catch! fixed https://github.com/stellar/go/pull/5335/commits/f2de0140fef30385bb2d1d61ea49533441481558

exp/services/ledgerexporter/internal/config.go
sreuland: yes, that is a good convention to follow, done - https://github.com/stellar/go/pull/5335/commits/d4aada61ec331c11741e7e1eae931ebe13e0f1db

exp/services/ledgerexporter/config.toml
urvisavla: Also, `preconfigured_network` seems a little confusing. Perhaps we should rename it to `network_name`?

exp/services/ledgerexporter/README.md
```diff
@@ -21,45 +21,57 @@ type LedgerCloseMetaBatch struct {
...
+  - manually configure captive core by supplying external core config file and archive urls and passphrase:
+    `stellar_core_config.captive_core_toml_path`
+    `stellar_core_config.history_archive_urls`
+    `stellar_core_config.network_passphrase`
+  - Ensure you have stellar-core installed and accessible in your system's $PATH.
+
+An example config, demonstrating preconfigured captive core settings.
 ```toml
-network = "testnet"  # Options: `testnet` or `pubnet`
+network_name = "testnet"
```
tamirms, line 51: could you add documentation for the optional admin_port field which can be included in the toml config?

exp/services/ledgerexporter/README.md
```diff
@@ -21,45 +21,57 @@ type LedgerCloseMetaBatch struct {
...
 ```bash
-ledgerexporter --start <start_ledger> --end <end_ledger> --config-file <config_file_path>
+ledgerexporter scan-and-fill --start <start_ledger> --end <end_ledger> --config-file <config_file_path>
 ```
 
-#### Unbounded Mode:
-Exports ledgers continuously starting from --start. In this mode, the end ledger is either not provided or set to 0.
+#### Append Mode:
+Exports ledgers initially searching from --start, looking for the next absent ledger sequence number proceeding --start on the data store. If abscence is detected, the export range is narrowed to `--start <absent_ledger_sequence>`. 
+This feature requires ledgers to be present on the remote data store for some (possibly empty) prefix of the requested range and then absent for the (possibly empty) remainder.
```
tamirms, line 32: I think somewhere in the Append Mode section we should mention that ledger exporter will never leave any gaps (assuming that there were no pre-existing gaps). In the [ledger exporter spec
](https://docs.google.com/document/d/1t9bKbzW6QtbnRCrGv4sUstJtK_DFN0kXv75kYZ8j6h0/edit) I phrased this behavior as:

>If Ledger Exporter is terminated early (gracefully or via a crash), the export command will have made partial progress by exporting a (possibly empty) prefix of the requested range.
In other words, if n is the highest ledger sequence exported to the data lake during the execution of the program then it’s guaranteed that all ledgers preceding n in the requested range were also exported to the data lake during the execution of the program.

But feel free to write your own description if you prefer

exp/services/ledgerexporter/README.md
urvisavla: How about a middle ground where it uses the default parameters for captive-core config based on the `network_name` but provide an option to override just the captive-core config file? This option would be useful in cases like Simon's, where you might want to set/overwrite stellar-core parameters (e.g. ENABLE_SOROBAN_DIAGNOSTIC_EVENTS) without modifying validator/history archive settings.

exp/services/ledgerexporter/internal/config.go
```diff
@@ -39,105 +52,143 @@ type StellarCoreConfig struct {
...
+			config.EndLedger, latestNetworkLedger)
 	}
 
 	config.adjustLedgerRange()
 	return nil
 }
 
-func (config *Config) generateCaptiveCoreConfig() (ledgerbackend.CaptiveCoreConfig, error) {
-	coreConfig := &config.StellarCoreConfig
+func (config *Config) GenerateHistoryArchive(ctx context.Context) (historyarchive.ArchiveInterface, error) {
```
urvisavla, line 132: Should the methods on the config start with a lowercase letter since the `Config` is internal to the package and not exported? I'm not sure what the Go conventions say about methods that are not necessarily internal to the struct (called from other structs like `App` in this case) but are also not exported outside the package?

exp/services/ledgerexporter/internal/main.go
```diff
@@ -0,0 +1,94 @@
...
+			)
+			settings.Mode = Append
+			return ledgerExporterCmdRunner(settings)
+		},
+	}
+
+	rootCmd.AddCommand(scanAndFillCmd)
+	rootCmd.AddCommand(appendCmd)
+
+	scanAndFillCmd.PersistentFlags().Uint32("start", 0, "Starting ledger (inclusive), must be set to a value greater than 1")
```
urvisavla, line 65: Do we want to have shorthand options for the start, end, and config-file flags?

chowbao: Ooo +1

exp/services/ledgerexporter/internal/main.go
```diff
@@ -0,0 +1,94 @@
...
+	scanAndFillCmd.PersistentFlags().Uint32("end", 0, "Ending ledger (inclusive), must be set to value greater than 'start' and less than 'latest checkpoint ledger from network history archives + (2 * checkpoint ledger frequency)'")
```
urvisavla, line 66: Is it expected that the operator should know how to get the latest checkpoint ledger and checkpoint ledger frequency from the history archives? If not, would it not be simpler to just mention that the end ledger should be less than the latest ledger on the Stellar network?

exp/services/ledgerexporter/internal/main.go
```diff
@@ -0,0 +1,94 @@
...
+	scanAndFillCmd.PersistentFlags().String("config-file", "config.toml", "Path to the TOML config file. Defaults to 'config.toml' on runtime working directory path.")
+	viper.BindPFlags(scanAndFillCmd.PersistentFlags())
+
+	appendCmd.PersistentFlags().Uint32("start", 0, "Starting ledger (inclusive), must be set to a value greater than 1")
+	appendCmd.PersistentFlags().Uint32("end", 0, "Ending ledger, optional, setting to non-zero means bounded mode, "+
```
urvisavla, line 71:
```suggestion
	appendCmd.PersistentFlags().Uint32("end", 0, "Ending ledger (inclusive), optional, setting to non-zero means bounded mode, "+
```

exp/services/ledgerexporter/internal/main.go
```diff
@@ -0,0 +1,94 @@
...
+				cmd.PersistentFlags().Lookup("config-file"),
+			)
+			settings.Mode = ScanFill
+			return ledgerExporterCmdRunner(settings)
+		},
+	}
+	appendCmd = &cobra.Command{
+		Use:   "append",
+		Short: "export ledgers beginning with a dynamically determined starting point based on binary search for first absent ledger on datastore after the 'start' ledger",
+		Long:  "export ledgers beginning with a dynamically determined starting point based on binary search for first absent ledger on datastore after the 'start' ledger",
```
urvisavla, line 51: Do we need to clarify that the reason for doing this is to resume from where it left off? And may be we could simplify the wording to something like `finds the first missing ledger after the specified 'start' ledger and resumes exporting from there`?

exp/services/ledgerexporter/config.toml
sreuland: will consolidate to `stellar_core_config.network` which is then same as horizon `NETWORK`, valid only for 'testnet|pubnet'

exp/services/ledgerexporter/README.md
sreuland: ok, that sounds like a nice convenience, it infers all `stellar_core_config` params are optional, and would relax validation. maybe that introduces a bit more complexity, with room for users to inadverently override their own intent with the permutation/combos of network(preconfigured) and manual configs that overlap. I think as part of doing this will make sure to print a log line of the final computed network config state at runtime so users can see confirmation.

urvisavla: There is also a validation [here](https://github.com/stellar/go/blob/bba243eafaaede0b70686220daabf5091c4d9656/ingest/ledgerbackend/toml.go#L613) that would catch if the network specified in the exporter config file differs from the one in the captive-core config file.

sreuland: done - https://github.com/stellar/go/pull/5335/commits/bfc25f48916a08c9961fad5bfae82983f74a06b3

exp/services/ledgerexporter/config.toml
sreuland: done - https://github.com/stellar/go/pull/5335/commits/bfc25f48916a08c9961fad5bfae82983f74a06b3

exp/services/ledgerexporter/README.md
sreuland: yes - https://github.com/stellar/go/pull/5335/commits/6d269e2601705418a206a5282c2c404c5bffdac9

exp/services/ledgerexporter/internal/main.go
sreuland: nice suggestion - https://github.com/stellar/go/pull/5335/commits/e1836a12ad392efeaaafd61600bb5200cbb768fd

exp/services/ledgerexporter/internal/main.go
sreuland: yeah, that is not even readable as well, reworded, thanks

exp/services/ledgerexporter/internal/main.go
sreuland: yes, like the simplified suggestion, i don't think we need to go into any prior reasons of why - https://github.com/stellar/go/pull/5335/commits/ed16b8be99e4ca454ec75e4b8e000267afa47f83

exp/services/ledgerexporter/internal/config.go
sreuland: yes, i like the way it indicates that type of struct level visibility since these are all under internal package already, kinda like oo with class level access modifiers on methods, i think go limits to package level convention - https://stackoverflow.com/questions/22148143/private-fields-and-methods-for-a-struct

exp/services/ledgerexporter/README.md
```diff
@@ -21,45 +21,56 @@ type LedgerCloseMetaBatch struct {
...
 
-#### Stellar-core configuration:
-- The exporter automatically configures stellar-core based on the network specified in the config.
-- Ensure you have stellar-core installed and accessible in your system's $PATH.
+[stellar_core_config]
+  network = "testnet"
+  stellar_core_binary_path = "/my/path/to/stellar-core"
+  captive_core_toml_path = "my-captive-core.cfg"
+  history_archive_urls = ["http://testarchiveurl1", "http://testarchiveurl2"]
+  network_passphrase = "test"
```
tamirms, line 72: if this example is meant to demonstrate preconfigured captive core settings then I think we can remove these fields

exp/services/ledgerexporter/internal/config.go
```diff
@@ -39,115 +52,126 @@ type StellarCoreConfig struct {
...
 
-	config.StartLedger = uint32(flags.StartLedger)
-	config.EndLedger = uint32(flags.EndLedger)
-	config.Resume = flags.Resume
+	config.StartLedger = uint32(settings.StartLedger)
+	config.EndLedger = uint32(settings.EndLedger)
+	config.Mode = settings.Mode
 
-	logger.Infof("Requested ledger range start=%d, end=%d, resume=%v", config.StartLedger, config.EndLedger, config.Resume)
+	logger.Infof("Requested export mode of %v with start=%d, end=%d", settings.Mode, config.StartLedger, config.EndLedger)
```
tamirms, line 79: when you print settings.Mode do you get an integer or  is there a meaningful string?

exp/services/ledgerexporter/ledgerexporter
tamirms, line 1: I'm guessing this was not intended to be checked in?

exp/services/ledgerexporter/internal/config.go
sreuland: integer, added string convert - https://github.com/stellar/go/pull/5335/commits/328aff03a5d1396a0726ae7aea5dd8baf51e909d

exp/services/ledgerexporter/ledgerexporter
sreuland: doh! removed, thanks!

tamirms (approved):

PR #5337, opened 2024-06-07
services/horizon: Update guide for developers

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

- Improved services/horizon/docker/start.sh script so that it could be executed from any directory. Previously, the script only worked if you executed it in the services/horizon/docker directory
- Updated the configuration in the developer docs to use the protocol 21 stellar core version
- Fixed the docker compose commands in the "Database Setup" so that they reference the docker compose files from the repository root.
- Added " --remove-orphans -v" flags to the docker-compose down command so that the db volumes are wiped out
- Updated the environment variables in the IDE configuration


services/horizon/internal/docs/GUIDE_FOR_DEVELOPERS.md
```diff
@@ -94,18 +68,43 @@ Add a debug configuration in your IDE to attach a debugger to the local Horizon
      "program": "${workspaceRoot}/services/horizon/main.go",
      "env": {
          "DATABASE_URL": "postgres://postgres@localhost:5432/horizon?sslmode=disable",
-         "CAPTIVE_CORE_CONFIG_APPEND_PATH": "./ingest/ledgerbackend/configs/captive-core-testnet.cfg",
-         "HISTORY_ARCHIVE_URLS": "https://history.stellar.org/prd/core-testnet/core_testnet_001,https://history.stellar.org/prd/core-testnet/core_testnet_002",
-         "NETWORK_PASSPHRASE": "Test SDF Network ; September 2015",
```
tamirms, line 99: these env variables can be replaced by `NETWORK=testnet`

services/horizon/internal/docs/GUIDE_FOR_DEVELOPERS.md
```diff
@@ -94,18 +68,43 @@ Add a debug configuration in your IDE to attach a debugger to the local Horizon
      "program": "${workspaceRoot}/services/horizon/main.go",
      "env": {
          "DATABASE_URL": "postgres://postgres@localhost:5432/horizon?sslmode=disable",
```
chowbao, line 70: To confirm when do we want the `horizon` part in the URL or not?

services/horizon/internal/docs/GUIDE_FOR_DEVELOPERS.md
```diff
@@ -94,18 +68,43 @@ Add a debug configuration in your IDE to attach a debugger to the local Horizon
...
-         "CAPTIVE_CORE_CONFIG_APPEND_PATH": "./ingest/ledgerbackend/configs/captive-core-testnet.cfg",
-         "HISTORY_ARCHIVE_URLS": "https://history.stellar.org/prd/core-testnet/core_testnet_001,https://history.stellar.org/prd/core-testnet/core_testnet_002",
-         "NETWORK_PASSPHRASE": "Test SDF Network ; September 2015",
+         "APPLY_MIGRATIONS": "true",
```
tamirms, line 71: this flag is required so that migrations are automatically applied when horizon starts up

services/horizon/internal/docs/GUIDE_FOR_DEVELOPERS.md
tamirms: this should work if you spin up the horizon db using docker-compose. If you install / run postgres some other way then this db url will not work unless you explicitly create the horizon database.

services/horizon/internal/docs/GUIDE_FOR_DEVELOPERS.md
```diff
@@ -42,47 +42,21 @@ We will now configure a development environment to run Horizon service locally w
...
 
-### Building Horizon
-
-1. Change to the horizon services directory - `cd go/services/horizon/`.
-2. Compile the Horizon binary: `go build -o stellar-horizon && go install`. You should see the resulting `stellar-horizon` executable in `go/services/horizon`.
-3. Add the executable to your PATH in your `~/.bashrc` or equivalent, for easy access: `export PATH=$PATH:{absolute-path-to-horizon-folder}`
-
-Open a new terminal. Confirm everything worked by running `stellar-horizon --help` successfully. You should see an informative message listing the command line options supported by Horizon.
-
 ### Database Setup
```
chowbao, line 45: Should it be more explicit that the developer will probably need to do the database setup to run with their local horizon build?

services/horizon/internal/docs/GUIDE_FOR_DEVELOPERS.md
sydneynotthecity, line 1: nit: the `Go Monorepo` section has a bad URL to clone the go monorepo. Should be:

```
git clone https://github.com/stellar/go.git
```

services/horizon/internal/docs/GUIDE_FOR_DEVELOPERS.md
mollykarcher: I think the point is that you don't actually have to do local psql setup if you use the docker compose. Are you saying we should document how you locally install psql+create the dbs if you wanna go that route?

chowbao: Hmm probably not. Maybe it's just worth removing the "you can install locally" from the text block below. Like if the expectation is to always use the docker-compose option, maybe we should only offer the docker-compose option without any alternatives for local development

services/horizon/internal/docs/GUIDE_FOR_DEVELOPERS.md
```diff
@@ -94,18 +68,43 @@ Add a debug configuration in your IDE to attach a debugger to the local Horizon
...
+Open a new terminal. Confirm everything worked by running `stellar-horizon --help` successfully. You should see an informative message listing the command line options supported by Horizon.
+
+### Run tests
+At this point you should be able to run Horizon's unit tests:
+```bash
+cd go/services/horizon/
+go test ./...
+```
+
+To run the integration tests, you need to set some environment variables:
```
sreuland, line 95: in this `Run tests` scenario, since is outside of docker image, running directly on host, should we mention pre-requirements for running integration tests such as having a captive core binary installed on the o/s PATH?

services/horizon/internal/docs/GUIDE_FOR_DEVELOPERS.md
sreuland, line 1: I think `Code Formatting` section at bottom, line 234 can be reduced to just a link to the top level repo [style guidelines](https://github.com/tamirms/go/blob/update-horizon-developer-docs/CONTRIBUTING.md#style-guides) which repeat much of the same.

sreuland (approved):
lgtm!

services/horizon/internal/docs/GUIDE_FOR_DEVELOPERS.md
tamirms: I took a look at the [style guidelines](https://github.com/tamirms/go/blob/update-horizon-developer-docs/CONTRIBUTING.md#style-guides) but there is no mention of the following scripts:

```
./gofmt.sh
./govet.sh
./staticcheck.sh
```

tamirms:
@chowbao can you take another look at the PR? I think I have addressed all your code review feedback

chowbao:
> @chowbao can you take another look at the PR? I think I have addressed all your code review feedback

I think all the content looks good. I still wonder if the docs are easy enough for a new user to use though. But I don't think that should block the PR

nit: it would be nice if there was a `table of contents`.

chowbao (approved):

PR #5346, opened 2024-06-14
ingest/ledgerbackend: Add version check for protocol 21 

urvisavla:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

- Added shared functions to get the stellar-core build version and protocol version:
`GetCoreBuildVersion`
`GetCoreProtocolVersion`

- Updated ledgerexporter and captivecorebackend to use these shared functions to get stellar-core build version. 
- Implemented a version check for protocol 21 when creating a new captive-core instance.

### Why

We previously had the core build for Soroban in several places. Now that we're on protocol 21, we should check the protocol version instead of the core build version. For more details, see https://github.com/stellar/go/pull/5333.

### Known limitations

Any system using captive-core will not be compatible with stellar-core versions older than 21.


tamirms:
you can also remove protocol 20 from the integration tests matrix https://github.com/stellar/go/blob/master/.github/workflows/horizon.yml#L16

tamirms (approved):

tamirms:
@urvisavla I think you meant to say "Removing checks for core versions older than protocol 21," right?

urvisavla:
> @urvisavla I think you meant to say "Removing checks for core versions older than protocol 21," right?

yes, thank you.

ingest/ledgerbackend/toml.go
```diff
@@ -445,104 +441,13 @@ func (c *CaptiveCoreToml) CatchupToml() (*CaptiveCoreToml, error) {
 	return offline, nil
 }
 
-// coreVersion helper struct identify a core version and provides the
```
sreuland, line 448: is it worth considering leaving this core version check capability in but re-using it in a runtime validation check when ingest mode is enabled(INGEST=true), to allow Horizon to warn user if they are running Horizon against a now invalid older core version < v21? same suggestion [mentioned on go/5333](https://github.com/stellar/go/pull/5333#issuecomment-2168482703). rather than letting horizon ingestion engine try to initialize and run and then getting some various combo of lower level failures logged out related to toml or xdr, or scp error outputs, which users will have to decipher to realize they have the wrong core version?

urvisavla: I agree, that would be a better ux. I will add a check for the core version right at the startup for Horizon.

exp/services/ledgerexporter/internal/config_test.go
```diff
@@ -95,7 +93,7 @@ func TestDefaultCaptiveCoreBin(t *testing.T) {
 		RuntimeSettings{ConfigFilePath: "test/no_core_bin.toml"})
 	require.NoError(t, err)
 
-	cmdOut = "v20.2.0-2-g6e73c0a88\n"
+	ledgerbackend.GetCoreBuildVersionFunc = func(string) (string, error) { return "v20.2.0-2-g6e73c0a88", nil }
```
sreuland, line 96: this *Func override convention could be avoided by using a dependency injection pattern on Config instead, thereby enabling ledgerexporter callers(tests) to mock it and ledgerbackend doesn't need to provide any additional conventions:

```
// add this in config.go

type GetCoreBuildVersion func(coreBinaryPath string) (string, error) 

Config struct {
   ...
   GetCoreBuildVersionFunc GetCoreBuildVersion    
}

func NewConfig(settings RuntimeSettings, getCoreVer GetCoreBuildVersion) {
    ...
    config.GetCoreBuildVersionFunc = ledgerbackend.GetCoreBuildVersionFunc
    if getCoreVer != nil {
       config.GetCoreBuildVersionFunc = getCoreVer
   }
   ...
}
```

ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -22,6 +20,8 @@ import (
 	"github.com/stellar/go/xdr"
 )
 
+var minProtocolVersionSupported uint = 21
```
sreuland, line 23: const ?

sreuland: after reading more, what do you think about not hard coding here, and the threshold is passed by caller instead on`NewCaptive()`, since there's nothing inside captive_core that drives this protocol version, only the caller knows what protocol compatability is needed correct?

ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -250,28 +261,12 @@ func (c *CaptiveStellarCore) coreVersionMetric() float64 {
...
-
-// Executes the "stellar-core version" command and parses its output to extract
-// the core version
-// The output of the "version" command is expected to be a multi-line string where the
-// first line is the core version in format "vX.Y.Z-*".
 func (c *CaptiveStellarCore) setCoreVersion() {
-	versionCmd := execCommand(c.config.BinaryPath, "version")
-	versionOutput, err := versionCmd.Output()
+	var err error
+	c.captiveCoreVersion, err = GetCoreBuildVersionFunc(c.config.BinaryPath)
```
sreuland, line 266: could follow the same dependency injection pattern to allow caller/environment to set the reference for `GetCoreBuildVersionFunc` as [recommended on here](https://github.com/stellar/go/pull/5346/files#r1644823760)

ingest/ledgerbackend/captive_core_backend_test.go
```diff
@@ -169,6 +171,35 @@ func TestCaptiveNew(t *testing.T) {
...
+
+	GetCoreProtocolVersionFunc = func(string) (uint, error) { return 20, nil }
+
+	_, err = NewCaptive(
+		CaptiveCoreConfig{
+			BinaryPath:         executablePath,
+			NetworkPassphrase:  networkPassphrase,
+			HistoryArchiveURLs: historyURLs,
+			StoragePath:        storagePath,
+			UserAgent:          "uatest",
```
sreuland, line 196: if following a dependency injection pattern, this would add `GetCoreBuildVersionFunc : GetCoreProtocolVersionFunc` rather than overriding a package singleton variable.

ingest/ledgerbackend/stellar_core_version.go
```diff
@@ -0,0 +1,63 @@
...
+import (
+	"fmt"
+	"os/exec"
+	"regexp"
+	"strconv"
+	"strings"
+)
+
+// By default, it points to exec.Command, overridden for testing purpose
+var execCommand = exec.Command
```
sreuland, line 12: if callers use [dependency injection pattern](https://github.com/stellar/go/pull/5346/files#r1644823760), can remove these vars, ln 12-14

ingest/ledgerbackend/stellar_core_version_test.go
sreuland, line 1: awesome test coverage!

ingest/ledgerbackend/captive_core_backend.go
urvisavla: The captive-core configuration in ingest package (toml.go) is only compatible with version 21 and above so it is appropriate to specify the min supported version here.

exp/services/ledgerexporter/internal/config_test.go
urvisavla: Addressed in https://github.com/stellar/go/pull/5346/commits/c43e98114da8a0a727b9a4362b31d96cadb4b88f

ingest/ledgerbackend/captive_core_backend_test.go
urvisavla: Addressed in https://github.com/stellar/go/commit/c43e98114da8a0a727b9a4362b31d96cadb4b88f

ingest/ledgerbackend/stellar_core_version.go
urvisavla: Addressed in https://github.com/stellar/go/commit/c43e98114da8a0a727b9a4362b31d96cadb4b88f

ingest/ledgerbackend/captive_core_backend.go
urvisavla: Addressed in https://github.com/stellar/go/commit/c43e98114da8a0a727b9a4362b31d96cadb4b88f

sreuland (approved):
lgtm, thanks for considering the suggeestions.

Issue #5342, opened 2024-06-13
ingest/ledgerbackend: getLogLineWriter() leaks goroutines and pipes


2opremio:
`getLogLineWriter()` creates a pipe and a goroutine, but the goroutine sometimes doesn't exit (and the pipe is never released).

Every time captive core is restarted this causes a leak.

The easiest fix is probably returning a `CloseWriter` and making sure that the writer is closed (releasing the pipe and the goroutine).

2opremio:
Discovered while investigating https://github.com/stellar/soroban-rpc/pull/207

2opremio:
It turns out the goroutines are closed in many places but something must be amiss, since they are still left behind in some cases. All the `(*pipe).read` goroutines in this dump are from `Closed()` core ledger backends:


<img width="700" alt="Screenshot 2024-06-13 at 17 30 56" src="https://github.com/stellar/go/assets/2362916/43558679-1295-4b62-ada5-913cb47aad03">

2opremio:
I think `stellarCoreRunner.handleExit()` may not be closing the pipes in some cases

2opremio:
Here's another dump:

<img width="691" alt="Screenshot 2024-06-13 at 18 28 57" src="https://github.com/stellar/go/assets/2362916/02109d27-01c5-401c-92f5-d7707996325e">

tamirms:
It looks like there are createCmd calls where we do not close the log line writers:

https://github.com/stellar/go/blob/a4e5a3fc6664f7cf6b26fb3196b366910563636f/ingest/ledgerbackend/stellar_core_runner.go#L311

https://github.com/stellar/go/blob/a4e5a3fc6664f7cf6b26fb3196b366910563636f/ingest/ledgerbackend/stellar_core_runner.go#L400-L404

https://github.com/stellar/go/blob/a4e5a3fc6664f7cf6b26fb3196b366910563636f/ingest/ledgerbackend/stellar_core_runner.go#L311

Issue #5349, opened 2024-06-16
ingest/ledgerbackend: Stopping captive core can be slow


2opremio:
We only cancel the running processes after acquiring the lock (which can be really slow in some cases)

```go
func (c *CaptiveStellarCore) Close() error {
	c.stellarCoreLock.RLock()
	defer c.stellarCoreLock.RUnlock()

	c.closed = true

	// after the CaptiveStellarCore context is canceled all subsequent calls to PrepareRange() will fail
	c.cancel()
	[...]
}
```

I don't know if it's safe to cancel before acquiring the lock (I bet it is) or if we can reduce the critical sections around the lock.

Related to https://github.com/stellar/go/issues/5347 ?

2opremio:
For context: I found this during an integration test which spends most of the time waiting for captive core to close

2opremio:
Uhm, actually moving the `cancel()` out of the criticial section didn't help.

For more context, core was trying to catchup when closing it. It took 20 seconds to close it.

```
time="2024-06-16T16:45:17.338+02:00" level=warning msg="Process: process 65163 exited 22: curl -sf http://localhost:57043/history/00/00/00/history-00000017.json -o buckets/tmp/history-2ea087fae299fb7b/4abab04a1c772310-stellar-history.json" pid=65148 subservice=stellar-core
time="2024-06-16T16:45:17.338+02:00" level=error msg="History: Could not download file: archive h0 maybe missing file history/00/00/00/history-00000017.json" pid=65148 subservice=stellar-core
time="2024-06-16T16:45:17.338+02:00" level=error msg="History: Missing HAS for ledger 23: maybe stale archive h0" pid=65148 subservice=stellar-core
time="2024-06-16T16:45:21.717+02:00" level=info msg="History: Downloading history archive state: history/00/00/00/history-00000017.json" pid=65148 subservice=stellar-core
time="2024-06-16T16:45:21.717+02:00" level=warning msg="Process: process 65166 exited 7: curl -sf http://localhost:57043/history/00/00/00/history-00000017.json -o buckets/tmp/history-2ea087fae299fb7b/4abab04a1c772310-stellar-history.json" pid=65148 subservice=stellar-core
time="2024-06-16T16:45:21.717+02:00" level=error msg="History: Could not download file: archive h0 maybe missing file history/00/00/00/history-00000017.json" pid=65148 subservice=stellar-core
[...]

PR #5352, opened 2024-06-16
Misc fixes

### LINKED ISSUE #5350: ingest/ledgerbackend: Captive core's LogLineWriter should be closed synchronously
We just close the writer side of the pipe but don't wait for the goroutine to exit:

https://github.com/stellar/go/blob/a4e5a3fc6664f7cf6b26fb3196b366910563636f/ingest/ledgerbackend/stellar_core_runner.go#L197-L209

This can cause an extra log line to be written after closing the backend. Which causes problems in the new Soroban RPC integration tests. See https://github.com/stellar/soroban-rpc/actions/runs/9538438244/job/26287637351?pr=216#step:10:50

### LINKED ISSUE #5351: historyarchive: should not use default logger
For instance:

https://github.com/stellar/go/blob/a4e5a3fc6664f7cf6b26fb3196b366910563636f/historyarchive/archive_pool.go#L109-L114

Causes a log to happen without any control. A logger, controlled by the caller, should be used instead or no logging should happen.

For instance, this causes untrackable logs to happen in tests . e.g: https://github.com/stellar/soroban-rpc/actions/runs/9538438244/job/26287637351?pr=216#step:10:22

The log cannot be attributed to a particular test, which is problematic (and noisy)

2opremio:
Closes #5351 
Closes #5350 (including the removal of a nasty cast)
Addresses part of #5349 

Remove redundant code of `(*Session) Begin()`

### DIFF
--- a/historyarchive/archive_pool.go
+++ b/historyarchive/archive_pool.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
+
 	log "github.com/stellar/go/support/log"
 	"github.com/stellar/go/xdr"
 
@@ -19,6 +20,7 @@ import (
 // An ArchivePool is just a collection of `ArchiveInterface`s so that we can
 // distribute requests fairly throughout the pool.
 type ArchivePool struct {
+	logger  *log.Entry
 	backoff backoff.BackOff
 	pool    []ArchiveInterface
 	curr    int
@@ -46,6 +48,7 @@ func NewArchivePoolWithBackoff(archiveURLs []string, opts ArchiveOptions, strate
 	ap := ArchivePool{
 		pool:    make([]ArchiveInterface, 0, len(archiveURLs)),
 		backoff: strategy,
+		logger:  opts.Logger,
 	}
 	var lastErr error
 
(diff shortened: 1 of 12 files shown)


historyarchive/archive_pool.go
```diff
@@ -107,8 +110,8 @@ func (pa *ArchivePool) runRoundRobin(runner func(ai ArchiveInterface) error) err
 		}
 
 		// Intentionally avoid logging context errors
-		if stats := ai.GetStats(); len(stats) > 0 {
-			log.WithField("error", err).Warnf(
+		if stats := ai.GetStats(); len(stats) > 0 && pa.logger != nil {
```
sreuland, line 113: for consider, short circuit on logger presence instead:
```
if pa.logger == nil {
    return err
}
```

2opremio: I didn't want to modify how the library behaved

ingest/ledgerbackend/main.go
```diff
@@ -49,36 +48,39 @@ type cmdI interface {
...
+	getStdout() *logLineWriter
+	setStderr(stderr *logLineWriter)
+	getStderr() *logLineWriter
 	getProcess() *os.Process
 	setExtraFiles([]*os.File)
 }
 
 type realCmd struct {
 	*exec.Cmd
+	stdout, stderr *logLineWriter
```
sreuland, line 61: can this continue to use abstraction with an interface still rather direct struct ref, seems like it could get by with `io.PipeWriter`? and then instances of logLineWriter or mocks can be used by callers/tests.

sreuland (approved):
I left comment for consideration, using [interface vs struct for log line writer](https://github.com/stellar/go/pull/5352/files#r1643206851). your choice, otherwise, lgtm!

ingest/ledgerbackend/main.go
2opremio: I considered it, but I didn’t see a need for it.

2opremio:
Unfortunately this didn't fully fix the logging race condition, we should fix https://github.com/stellar/go/issues/5342

PR #5340, opened 2024-06-11
exp/services/ledgerexporter: Add guide for contributors/developers

urvisavla:
<details>
  <summary>PR Checklist</summary>
  
### PR Structure

* [ ] This PR has reasonably narrow scope (if not, break it down into smaller PRs).
* [ ] This PR avoids mixing refactoring changes with feature changes (split into two PRs
  otherwise).
* [ ] This PR's title starts with name of package that is most changed in the PR, ex.
  `services/friendbot`, or `all` or `doc` if the changes are broad or impact many
  packages.

### Thoroughness

* [ ] This PR adds tests for the most critical parts of the new functionality or fixes.
* [ ] I've updated any docs ([developer docs](https://developers.stellar.org/api/), `.md`
  files, etc... affected by this change). Take a look in the `docs` folder for a given service,
  like [this one](https://github.com/stellar/go/tree/master/services/horizon/internal/docs).

### Release planning

* [ ] I've updated the relevant CHANGELOG ([here](services/horizon/CHANGELOG.md) for Horizon) if
  needed with deprecations, added features, breaking changes, and DB schema changes.
* [ ] I've decided if this PR requires a new major/minor version according to
  [semver](https://semver.org/), or if it's mainly a patch change. The PR is targeted at the next
  release branch if it's not a patch change.
</details>

### What

Add a guide for developers and contributors to complement the [Installation and Usage](https://github.com/stellar/go/blob/master/exp/services/ledgerexporter/README.md) guide.

### Why
Fixes [HUBBLE-457](https://stellarorg.atlassian.net/browse/HUBBLE-457)

### Known limitations

[HUBBLE-457]: https://stellarorg.atlassian.net/browse/HUBBLE-457?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #5355, opened 2024-06-20
exp/services/ledgerexporter: Guide to installing and running ledger exporter

urvisavla:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This PR is part of [hubble-382](https://stellarorg.atlassian.net/browse/HUBBLE-382) which includes adding instructions on how to run the ledger exporter and explaining the command line options to the public docs. Currently, I am adding the installation and running guide to the README with the goal of keeping all necessary documentation for setting up and running the ledger exporter close to the code. 

Additional docs, such as a developer guide is included in a separate [PR](https://github.com/stellar/go/pull/5340). There will also be a guide for creating consumer apps using the data exported by the ledger exporter. Once all the documentation is in place, we can discuss and decide what information should be included in the public documentation and what can remain in the github repo.

### Why

[HUBBLE-382](https://stellarorg.atlassian.net/browse/HUBBLE-382)

### Known limitations
N/A


exp/services/ledgerexporter/README.md
```diff
@@ -1,101 +1,124 @@
...
+### 3. Run the Exporter
 
-Enable web service that will be bound to localhost post and publishes metrics by including `admin_port = {port}`
+The following command demonstrates how to run the Ledger Exporter:
 
-An example config, demonstrating preconfigured captive core settings and gcs data store config.
-```toml
-admin_port = 6061
+```bash
+docker run --platform linux/amd64 -d \
```
sreuland, line 63: since using `-d` for daemon mode, none of the console logging output from ledgerexporter is making it out. Should we recommend how uses can access the log output of ledger exporter when running docker? I think it would be another host `-v` volume share to redirect the logs into a file that gets stored on host machine since daemon prevents console aspect?

exp/services/ledgerexporter/config.example.toml
```diff
@@ -0,0 +1,42 @@
...
+admin_port = 6061
+
+# Datastore Configuration
+[datastore]
+# Specifies the type of datastore. Currently, only Google Cloud Storage (GCS) is supported.
+type = "GCS"
+
+[datastore.parameters]
+# The Google Cloud Storage bucket path for storing data, with optional subpaths for organization.
+bucket_path = "your-bucket-name/<optional_subpath1>/<optional_subpath2>/"
```
chowbao, line 15: Is this supposed to be `destination_bucket_path` like in the normal `config.toml`?

exp/services/ledgerexporter/config.example.toml
```diff
@@ -0,0 +1,42 @@
...
+
+[datastore.schema]
+# Configuration for data organization
+ledgers_per_file = 64      # Number of ledgers stored in each file.
+files_per_partition = 10   # Number of files per partition/directory.
```
chowbao, line 20: nit: I know this is an example file but should these values be the same as the normal `config.toml`?

exp/services/ledgerexporter/README.md
```diff
@@ -1,101 +1,124 @@
...
-    `stellar_core_config.captive_core_toml_path`
-    `stellar_core_config.history_archive_urls`
-    `stellar_core_config.network_passphrase`
+### 2. Configure the Exporter (config.toml)
+The Ledger Exporter relies on a configuration file (config.toml) to connect to your specific environment. This file defines details like:
+- Your Google Cloud Storage (GCS) bucket where exported ledger data will be stored.
+- Stellar network settings, such as the network you're using (testnet or pubnet).
+- Datastore schema to control data organization.
+
+A sample configuration file [config.example.toml](config.example.toml) is provided. Copy and rename it to config.toml for customization. Edit the copied file (config.toml) to replace placeholders with your specific details.
```
chowbao, line 56: nit: should we also mention they could use/edit the existing `config.toml` file as well?
Or maybe this can be worded differently. I think it is a little odd to tell the user user copy and rename to `config.toml` when a `config.toml` file already exists

exp/services/ledgerexporter/README.md
```diff
@@ -1,101 +1,124 @@
...
 
-#### Example:
-with `ledgers_per_file = 64` and `files_per_partition = 10`:
-- Partition names: `/0-639`, `/640-1279`, ...
-- Filenames: `/0-639/0-63.xdr.zstd`, `/0-639/64-127.xdr.zstd`, ...
+Arguments:
+- `--start <start_ledger>` (required): The starting ledger sequence number in the range to export.
+- `--end <end_ledger>` (required): The ending ledger sequence number in the range.
+- `--config-file <config_file_path>` (optional): The path to your configuration file, containing details like GCS bucket information. If not provided, the exporter will look for config.toml in the directory where you run the command.
+### 2. append: Continuously Export New Data
```
chowbao, line 104: nit: I think append should be 1. and scan-and-fill should be 2. because append is probably going to be the more common case right?

tamirms (commented):
In the CLI section I like how you have added sections describing the usage and arguments for each command. However, I think the descriptions for the fill-gaps and append commands would be better left as is. There is a lot of detail in the command descriptions that @sreuland wrote in https://github.com/stellar/go/pull/5335 that appears to be edited away in this PR.

For example, the original description for append mode mentions the precondition for resumability:

> This feature requires ledgers to be present on the remote data store for some (possibly empty) prefix of the requested range and then absent for the (possibly empty) remainder.

I think it's important we keep this info in the readme because that is the only place where we communicate the prerequisite

exp/services/ledgerexporter/config.example.toml
urvisavla: The config.toml file is not required, so it was meant to be deleted and replaced by config.example.toml.

urvisavla:
> In the CLI section I like how you have added sections describing the usage and arguments for each command. However, I think the descriptions for the fill-gaps and append commands would be better left as is. There is a lot of detail in the command descriptions that @sreuland wrote in #5335 that appears to be edited away in this PR.
> 
> For example, the original description for append mode mentions the precondition for resumability:
> 
> > This feature requires ledgers to be present on the remote data store for some (possibly empty) prefix of the requested range and then absent for the (possibly empty) remainder.
> 
> I think it's important we keep this info in the readme because that is the only place where we communicate the prerequisite


You're right, It does leave out the following points for the `append`  mode because and mainly because I was unsure of its meaning:

`- This feature requires ledgers to be present on the remote data store for some (possibly empty) prefix of the requested range and then absent for the (possibly empty) remainder.
`

What does this precondition mean? Does it mean that append mode cannot be used if there is no pre-existing data?

`- It’s guaranteed that ledgers exported during `append` mode from `start` and up to the last logged ledger file `Uploaded {ledger file name}` were contiguous, meaning all ledgers within that range were exported to the data lake with no gaps or missing ledgers in between.` 

Regarding the phrase “and up to the last logged ledger file Uploaded {ledger file name},” does this refer to the console logs? Also, if there are no gaps then why is the scan-and-fill command required?


It might be more useful to provide guidelines on when to use each mode for the user but I am not entirely clear on them myself.  Perhaps you or @sreuland  could provide more insights into this? Thanks!

exp/services/ledgerexporter/README.md
urvisavla: That's a very good point. I think it would be best to remove the -d flag from the instructions rather than adding more instructions.

tamirms:
> What does this precondition mean? Does it mean that append mode cannot be used if there is no pre-existing data?

Running `append` on the range `[s, e]` (where `s <= e`) will efficiently export all the ledgers in the range by skipping ahead to the last exported ledger in the range. For example if the range is [2, 100] and we have already exported [2, 50], `append` will skip ahead to ledger 51 and start the export process from that point.

However, `append` is only guaranteed to export export all the ledgers in the requested range if the data lake satisfies the following precondition:

>This feature requires ledgers to be present on the remote data store for some (possibly empty) prefix of the requested range and then absent for the (possibly empty) remainder.

I mentioned an example above where the requested range is [2, 100] and we have already exported [2, 50]. In this example, the already exported prefix is [2, 50] and the absent remainder is [51, 100], thus the precondition is satisfied.

Here are some other scenarios where the precondition is satisfied:

- the data lake is empty. The already exported prefix is empty because the data lake has no ledgers. The absent remainder is [2, 100] because all the ledgers from the requested range are missing from the data lake.
- the data lake already contains ledgers [2, 100]. The already exported prefix is the entire requested range [2, 100]. The absent remainder is empty because all the ledgers have already been exported.

Here are some scenarios where the precondition is not satisfied:

- the data lake contains ledgers [10, 70]. [10, 70] is not a prefix of [2, 100]. A prefix of [2, 100] must start at ledger 2.
- the data lake contains ledgers [2, 70] and [95, 100]. [2, 70] is a prefix of [2, 100]. However, all the ledgers following the prefix must be absent in order for the precondition to be satisfied ("...and then absent for the (possibly empty) remainder"). The remainder [71, 100] is not absent because the data lake contains ledgers [95, 100].

Because the precondition is not satisfied in the 2 cases above, we cannot guarantee that `append` will export all the ledgers in the requested [2, 100] range. However, `scan-and-fill` will be able to guarantee that all the ledgers in the requested [2, 100] range will be exported because it will always iterate through every ledger in the requested range checking to see if each one is present.

---

> Regarding the phrase “and up to the last logged ledger file Uploaded {ledger file name},” does this refer to the console logs? Also, if there are no gaps then why is the scan-and-fill command required?

That part of the documentation is essentially stating that, given a requested ledger range `[s, e]`, if the `append` precondition was satisfied on `[s, e]` before running `append --start s --end e`, it is guaranteed that `[s, e]` will still satisfy the  `append` precondition after running `append` no matter how ledger exporter terminates. In other words, even if `append` crashes unexpectedly at any point during the export process, it will never introduce any gaps that would violate the `append` precondition.

> Also, if there are no gaps then why is the scan-and-fill command required?

This property of `append` only guarantees that the `append` precondition is maintained if the `append` precondition was satisfied before running `append`. There are many scenarios where scan-and-fill is not needed. But it's still possible to modify a data lake in such a way that the `append` precondition is not satisfied. For example:

- Manually deleting a prefix of `[s, e]` ledgers from the data lake would break the precondition.
- Manually deleting ledger(s) in the middle of `[s, e]` would break the precondition.
- If you run `append` on [2, 100] and then on [150, 200], we know that the precondition is satisfied on  [2, 100] and [150, 200]. However, if you consider the range [2, 200] the precondition is not satisfied because there is going to be a gap in the middle.

urvisavla:
> > What does this precondition mean? Does it mean that append mode cannot be used if there is no pre-existing data?
> 
> Running `append` on the range `[s, e]` (where `s <= e`) will efficiently export all the ledgers in the range by skipping ahead to the last exported ledger in the range. For example if the range is [2, 100] and we have already exported [2, 50], `append` will skip ahead to ledger 51 and start the export process from that point.
> 
> However, `append` is only guaranteed to export all the ledgers in the requested range if the data lake satisfies the following precondition:
> 
> > This feature requires ledgers to be present on the remote data store for some (possibly empty) prefix of the requested range and then absent for the (possibly empty) remainder.
> 
> I mentioned an example above where the requested range is [2, 100] and we have already exported [2, 50]. In this example, the already exported prefix is [2, 50] and the absent remainder is [51, 100], thus the precondition is satisfied.
> 
> Here are some other scenarios where the precondition is satisfied:
> 
> * the data lake is empty. The already exported prefix is empty because the data lake has no ledgers. The absent remainder is [2, 100] because all the ledgers from the requested range are missing from the data lake.
> * the data lake already contains ledgers [2, 100]. The already exported prefix is the entire requested range [2, 100]. The absent remainder is empty because all the ledgers have already been exported.
> 
> Here are some scenarios where the precondition is not satisfied:
> 
> * the data lake contains ledgers [10, 70]. [10, 70] is not a prefix of [2, 100]. A prefix of [2, 100] must start at ledger 2.
> * the data lake contains ledgers [2, 70] and [95, 100]. [2, 70] is a prefix of [2, 100]. However, all the ledgers following the prefix must be absent in order for the precondition to be satisfied ("...and then absent for the (possibly empty) remainder"). The remainder [71, 100] is not absent because the data lake contains ledgers [95, 100].
> 
> Because the precondition is not satisfied in the 2 cases above, we cannot guarantee that `append` will export all the ledgers in the requested [2, 100] range. However, `scan-and-fill` will be able to guarantee that all the ledgers in the requested [2, 100] range will be exported because it will always iterate through every ledger in the requested range checking to see if each one is present.
> 
> > Regarding the phrase “and up to the last logged ledger file Uploaded {ledger file name},” does this refer to the console logs? Also, if there are no gaps then why is the scan-and-fill command required?
> 
> That part of the documentation is essentially stating that, given a requested ledger range `[s, e]`, if the `append` precondition was satisfied on `[s, e]` before running `append --start s --end e`, it is guaranteed that `[s, e]` will still satisfy the `append` precondition after running `append` no matter how ledger exporter terminates. In other words, even if `append` crashes unexpectedly at any point during the export process, it will never introduce any gaps that would violate the `append` precondition.
> 
> > Also, if there are no gaps then why is the scan-and-fill command required?
> 
> This property of `append` only guarantees that the `append` precondition is maintained if the `append` precondition was satisfied before running `append`. There are many scenarios where scan-and-fill is not needed. But it's still possible to modify a data lake in such a way that the `append` precondition is not satisfied. For example:
> 
> * Manually deleting a prefix of `[s, e]` ledgers from the data lake would break the precondition.
> * Manually deleting ledger(s) in the middle of `[s, e]` would break the precondition.
> * If you run `append` on [2, 100] and then on [150, 200], we know that the precondition is satisfied on  [2, 100] and [150, 200]. However, if you consider the range [2, 200] the precondition is not satisfied because there is going to be a gap in the middle.

Oh, so the precondition is not necessarily for running in append mode; it's a precondition for guaranteeing an export without gaps. The phrase "This feature requires ledgers to be present" confused me into thinking that unless the condition is met, you cannot run in append mode. Now it makes sense. Thanks for explaining it in depth 🙏  I'll add the original description back.

tamirms (approved):

PR #5359, opened 2024-06-26
sevices/friendbot: Allow friendbot to include cloudflare derived IP address in request logs

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Friendbot uses a logging middleware which is used to log incoming requests. Those logs include the client's IP address. However, that ip address is incorrect if the friendbot http server is load balanced by cloudflare. This PR implements a configuration option which enables friendbot to include the real client ip derived from cloudflare in the request logs.

### Why

It is useful to see real ips in the logs so we can see observe traffic patterns

### Known limitations

[N/A]


services/friendbot/main.go
```diff
@@ -84,8 +86,12 @@ func run(cmd *cobra.Command, args []string) {
 	})
 }
 
-func initRouter(fb *internal.Bot) *chi.Mux {
-	mux := http.NewAPIMux(log.DefaultLogger)
+func initRouter(cfg Config, fb *internal.Bot) *chi.Mux {
+	mux := chi.NewRouter()
+	// first apply XFFMiddleware so we can have the real ip in the subsequent
+	// middlewares
+	mux.Use(http.XFFMiddleware(http.XFFMiddlewareConfig{BehindCloudflare: cfg.UseCloudflareIP}))
```
sreuland, line 93: this is fine for now to get unstuck, but could this be done in a more general non-vendor way to configure getting source ip addresses through a configurable http header name instead? it usually resolves down to an http header, but different name specific to vendor on the lb/firewall, etc.

sreuland (approved):
are there any integration tests in the repo that bring up friendbot to confirm this config change works?

services/friendbot/main.go
tamirms: yes, I think this could be a future improvement for the XFFMiddleware

tamirms:
@sreuland I added a unit test. Does that look good to you?

PR #5366, opened 2024-07-01
services/horizon/internal: Add flag to enable / disable reaping of lookup tables

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add flag to enable / disable reaping of lookup tables.

### Why

Reaping of lookup tables is an operation which blocks the ingestion thread and it can be a lengthy operation. We need to have a mechanism to disable reaping of lookup tables while still enabling reaping of history so we can prevent ingestion lag from occurring when reaping of lookup tables takes too long.

Note that it is safe to run horizon with reaping of history enabled and reaping of lookup tables disabled. The only negative consequence of such a configuration is that there may be rows in the lookup tables which are not referenced anymore (e.g. a history account id which is no longer used in any of the history tables). But, once reaping of lookup tables is re-enabled those rows should eventually be deleted.

### Known limitations

[N/A]

### DIFF
--- a/services/horizon/internal/flags.go
+++ b/services/horizon/internal/flags.go
@@ -687,6 +687,14 @@ func Flags() (*Config, support.ConfigOptions) {
 				return nil
 			},
 		},
+		{
+			Name:           "reap-lookup-tables",
+			ConfigKey:      &config.ReapLookupTables,
+			OptType:        types.Bool,
+			FlagDefault:    true,
+			Usage:          "enables the reaping of history lookup tables.",
+			UsedInCommands: IngestionCommands,
+		},
 		&support.ConfigOption{
 			Name:           "history-stale-threshold",
 			ConfigKey:      &config.StaleThreshold,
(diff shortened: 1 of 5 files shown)


services/horizon/internal/flags.go
```diff
@@ -687,6 +687,14 @@ func Flags() (*Config, support.ConfigOptions) {
 				return nil
 			},
 		},
+		{
+			Name:           "reap-lookup-tables",
```
sreuland, line 691: can this configuration be passed by allowing presence of `--reap-frequency 0` rather than a new flag?

tamirms: @sreuland no because reaping lookup tables is separate and independent from reaping history. reap-frequency is configuration specific to reaping history which does not apply to reaping lookup tables.

sreuland: ok, thanks for clarification. out of scope for this task, but, maybe for consideration later, consolidate reaper configs to represent single whole and under the hood, horizon can optimize the lookup and history settings from there. few external users probably grasp this distinction or intentionally tune these differently?

sreuland (approved):

PR #5368, opened 2024-07-01
Create CI workflow for ledger exporter release

urvisavla:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Two new git CI workflows for the ledger exporter:

1. PR Workflow: Runs whenever a pull request is merged to the master branch and does the following:
- Builds the ledger exporter docker image.
- Tests the image using a fake-gcs-storage emulator.

2. Release Workflow: Runs whenever a release tag is created for the ledger exporter and does the following:
- Extracts the version number from the branch name (e.g., "ledgerexporter-v1.2").
- Builds a docker image tagged with stellar/ledger-exporter:<version_number> (e.g., "stellar/ledger-exporter:1.2").
- Pushes the built image to the Stellar docker repo.

### Why

Fixes [HUBBLE-492
](https://stellarorg.atlassian.net/browse/HUBBLE-492)

### Known limitations

N/A

### DIFF
--- a/.github/workflows/ledgerexporter-release.yml
+++ b/.github/workflows/ledgerexporter-release.yml
@@ -1,0 +1,36 @@
+name: Ledger Exporter release
+
+on:
+  push:
+    tags: ['ledgerexporter-v*']
+
+jobs:
+
+  publish-docker:
+    name: Test and push the Ledger Exporter images
+    runs-on: ubuntu-latest
+    env:
+      STELLAR_CORE_VERSION: 21.0.0-1872.c6f474133.focal
+      VERSION: ${GITHUB_REF_NAME#ledgerexporter-v}
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          # For pull requests, build and test the PR head not a merge of the PR with the destination.
+          ref:  ${{ github.event.pull_request.head.sha || github.ref }}
+      - name: Build Ledger Exporter docker
+        run: make -C exp/services/ledgerexporter docker-build
+
+      - name: Run Ledger Exporter test
+        run: make -C exp/services/ledgerexporter docker-test
+
+      # Push images
+      - if: github.ref == 'refs/heads/master'
+        name: Login to DockerHub
+        uses: docker/login-action@bb984efc561711aaa26e433c32c3521176eae55b
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - if: github.ref == 'refs/heads/master'
+        name: Push to DockerHub
+        run: make -C exp/services/ledgerexporter docker-push
--- a/.github/workflows/ledgerexporter-release.yml
+++ b/.github/workflows/ledgerexporter-release.yml
@@ -16,7 +16,7 @@ jobs:
       - uses: actions/checkout@v3
         with:
           # For pull requests, build and test the PR head not a merge of the PR with the destination.
-          ref:  ${{ github.event.pull_request.head.sha || github.ref }}
+          ref: github.sha
       - name: Build Ledger Exporter docker
         run: make -C exp/services/ledgerexporter docker-build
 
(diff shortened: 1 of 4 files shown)


exp/services/ledgerexporter/Makefile
```diff
@@ -2,7 +2,7 @@ SUDO := $(shell docker version >/dev/null 2>&1 || echo "sudo")
 
 # https://github.com/opencontainers/image-spec/blob/master/annotations.md
 BUILD_DATE := $(shell date -u +%FT%TZ)
-VERSION ?= 1.0.0-$(shell git rev-parse --short HEAD)
+VERSION ?= $(shell git rev-parse --short HEAD)
```
chowbao, line 5: So ledger exporter will no longer use SDF's modified semantic versioning? Version will only be the git commit hash?
I'm fine with this. I just want to confirm.

urvisavla: This line sets the version to the git hash only if the VERSION environment variable is not set, which will not be the case during the release. The version is extracted from the branch name [here](https://github.com/stellar/go/pull/5368/files#diff-05abd859457f4ed37b8116b1827a87518a8bfd6cd23bed26d87b1359ed7fd3c7R14), so the branch name should follow the semver as with other projects.

chowbao (approved):

.github/workflows/ledgerexporter.yml
```diff
@@ -0,0 +1,24 @@
...
+  push:
+    branches: [master]
+  pull_request:
+
+jobs:
+  ledger-exporter:
+    name: Build and test Ledger Exporter image
+    runs-on: ubuntu-latest
+    env:
+      STELLAR_CORE_VERSION: 21.0.0-1872.c6f474133.focal
```
sreuland, line 13: this is for captive core bin and it config'd to testnet correct? should this be bumped to latest 21.1.0, `21.1.0-1921.b3aeb14cc.focal`

urvisavla: This is for installing a specific version of stellar-core inside the container. Do the testnet and pubnet require different versions of stellar-core?

.github/workflows/ledgerexporter-release.yml
```diff
@@ -0,0 +1,36 @@
...
+  push:
+    tags: ['ledgerexporter-v*']
+
+jobs:
+
+  publish-docker:
+    name: Test and push the Ledger Exporter images
+    runs-on: ubuntu-latest
+    env:
+      STELLAR_CORE_VERSION: 21.0.0-1872.c6f474133.focal
```
sreuland, line 13: same here, core version bump? `21.1.0-1921.b3aeb14cc.focal`

.github/workflows/ledgerexporter-release.yml
```diff
@@ -0,0 +1,36 @@
...
+      VERSION: ${GITHUB_REF_NAME#ledgerexporter-v}
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          # For pull requests, build and test the PR head not a merge of the PR with the destination.
+          ref:  ${{ github.event.pull_request.head.sha || github.ref }}
```
sreuland, line 19: conditional check on pull request attributes not necessary as this only triggers on tags correct?  i think this can be simplified to `ref: github.sha `

.github/workflows/ledgerexporter-release.yml
```diff
@@ -0,0 +1,36 @@
...
+      - name: Build Ledger Exporter docker
+        run: make -C exp/services/ledgerexporter docker-build
+
+      - name: Run Ledger Exporter test
+        run: make -C exp/services/ledgerexporter docker-test
+
+      # Push images
+      - if: github.ref == 'refs/heads/master'
```
sreuland: is the intent only to push image if the tag was against a commit on `master` branch? can you create tags to confirm this will happen? would filtering out forked repos suffice, i.e. only push image if the tag was done on stellar/go repo, with`github.repository_owner == 'stellar'`

sreuland: this code changed, comment not applicable anymore.

.github/workflows/ledgerexporter-release.yml
```diff
@@ -0,0 +1,34 @@
+name: Ledger Exporter release
+
+on:
+  push:
+    tags: ['ledgerexporter-v*']
+
+jobs:
+
+  publish-docker:
```
sreuland, line 9: is it worthwhile to add job filter to prevent forked repos, i.e. only push if tag is on stellar/go:

```
if: github.repository_owner == 'stellar'
```

sreuland (approved):

.github/workflows/ledgerexporter-release.yml
urvisavla: This will fail for forked repos because they won't have access to GH secrets. While we could add an if condition here but someone with a forked repo could still remove the if condition in the fork.

.github/workflows/ledgerexporter.yml
sreuland: was referring to the config of docker tests initially which run on testnet, I think 21.1.0 can be used for captive bin against either network now, we are running that version for captive core with horizon on pubnet.

PR #5374, opened 2024-07-05
services/horizon/ingest: Use buffered storage backend for reingest command

### LINKED ISSUE #4911: Add support for ingesting via precomputed TxMeta in Horizon
Horizon currently has two modes of ingestion:

1. The deprecated [DatabaseBackend](https://github.com/stellar/go/blob/master/ingest/ledgerbackend/database_backend.go) which extracts ledgers from Stellar Core's postgres DB
2. Captive Core

We need to add a third mode which is ingesting from the [precomputed TxMeta backend](https://github.com/stellar/go/issues/4910). This will require adding new [LedgerBackend](https://github.com/stellar/go/blob/master/ingest/ledgerbackend/ledger_backend.go#L10) implementation and configuration flags for Horizon that allow operators to select the new ledger backend. 

Initial scope is to enable precomputed for reingestion only, not live/forward ingestion. We will implement the following changes to the Horizon reingest command to support the new mode of ingestion:

- [ ] <del>When using precomputed TxMeta we may need to adjust the minimum [batch size](https://github.com/stellar/go/blob/master/services/horizon/internal/ingest/parallel.go#L14) and we may not need to [round the batch sizes](https://github.com/stellar/go/blob/master/services/horizon/internal/ingest/parallel.go#L108-L109) to multiples of 64.</del> - deferring this optimization for initial pass.
- [x] The default value for the [`parallel-job-size` parameter](https://github.com/stellar/go/blob/master/services/horizon/cmd/db.go#L275) needs to be reduced. (experimentally I found 100 to be the most efficient on production hw).
- [x]  add command line / env flags to allow horizon to toggle between ingestion via captive core and ingestion via precomputed Tx Meta for just reingestion commands.
- [x]  add command line / env flags for horizon to configure the [BufferedStorageBackend](https://github.com/stellar/go/blob/master/ingest/ledgerbackend/buffered_storage_backend.go#L20-L27), i.e. gcs bucket, schema aspects like ledgers-per-file, etc
- [ ] update reingestion integration tests so we exercise reingestion via both captive core and the BufferedStorageBackend

As part of scope considerations, we will not address the following on first deliverable, and these can be done in later iterations:

- cli command flag validations related to captive core vs. the new pre-computed, several captive core settings are global.
- updating any ops deployed environments or ci jobs(jenkins) that run `horizon db reingest`

sreuland:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Change `db reingest` command to accept additional configuration parameters to use a new buffered storage ledger backend(consumes precomputed tx meta files generated from ledger exporter) instead of using captive core backend. There is currently one implementation of the buffered storage reader available, so will see reference on that often which is for Google Cloud Storage(GCS) backend.

Refer to new [Reingestion readme docs](https://github.com/stellar/go/blob/reingest_precomputed/services/horizon/internal/ingest/README.md#reingestion) for details on the invoking `db reingest range <from> <to>` with new configuration via `--ledgerbackend` and `--datastore-config` parameters

Captive core remains the default ledger backend in all cases still, reingestion commands or on live/forward ingestion. Only reingestion allows config to override the ledger backend type.

Coauthors: @urvisavla 

### Why

Faster reingestion speeds. Ingestion from precomputed tx meta files is faster than consuming same tx meta from captive core pipe.
Closes #4911
### Known limitations

only affects reingestion command. Requires an authenticatable GCP cloud connection and the cloud data store must be maintained(receiving data files) from a running [ledger exporter](https://github.com/stellar/go/tree/master/exp/services/ledgerexporter) process, which is out of band from here.


ingest/ledgerbackend/buffered_storage_backend.go
```diff
@@ -54,17 +54,17 @@ func NewBufferedStorageBackend(ctx context.Context, config BufferedStorageBacken
 		return nil, errors.New("number of workers must be <= BufferSize")
 	}
 
-	if config.DataStore == nil {
+	if dataStore == nil {
```
tamirms, line 57: I think we can assume that dataStore must be non-nil and remove this check.

services/horizon/internal/ingest/README.md
```diff
@@ -140,8 +140,46 @@ This pauses the state machine for 10 seconds then tries again, in hopes that a n
...
+		   --datastore-config "config.storagebackend.toml" \
+       100 200
+  ```  
+Notice, even though we no longer need to provide stellar-core related config for binary or config file, we do still need to provide network related config, using convenience parameter `NETWORK=testnet|pubnet` or directly with `NETWORK_PASSPHRASE` and `HISTORY_ARCHIVE_URLS`
+
+The `--datastore-config` must point to a new toml config file that will provide the necessary parameters for ingestion to work with remote GCS storage.
+
+example config toml:
+```
+# Datastore Configuration
```
urvisavla, line 167: Add buffered_storage_backend_config section.

sreuland: intentionally left that out at this level, as those are low level perf settings, and figured didn't need that noise up here at the intro, user intro level.

services/horizon/cmd/db.go
sreuland, line 1: the main reason for the large diff on this file is refactoring the cobra command definitions from one time static package initializers into a repeatable method, `DefineDBComands`

ingest/ledgerbackend/buffered_storage_backend.go
```diff
@@ -45,7 +45,7 @@ type BufferedStorageBackend struct {
 }
 
 // NewBufferedStorageBackend returns a new BufferedStorageBackend instance.
-func NewBufferedStorageBackend(ctx context.Context, config BufferedStorageBackendConfig) (*BufferedStorageBackend, error) {
+func NewBufferedStorageBackend(ctx context.Context, config BufferedStorageBackendConfig, dataStore datastore.DataStore) (*BufferedStorageBackend, error) {
```
chowbao, line 48: Just confirming that the only changes I'll need to update Hubble with are 
* `NewBufferedStorageBackend` gets `dataStore datastore.DataStore` directly now instead of through `BufferedStorageBackendConfig`
* Pass `LedgersPerFile:    1`, `FilesPerPartition: 64000,` in `DataStoreConfig.schema` when calling `NewDataStore` instead of `LedgerBatchConfig`

or I could implement using a `config.toml` like services/horizon/internal/ingest/testdata/config.storagebackend.toml

ingest/ledgerbackend/buffered_storage_backend.go
```diff
@@ -54,17 +54,17 @@ func NewBufferedStorageBackend(ctx context.Context, config BufferedStorageBacken
 		return nil, errors.New("number of workers must be <= BufferSize")
 	}
 
-	if config.DataStore == nil {
+	if dataStore == nil {
 		return nil, errors.New("no DataStore provided")
 	}
 
-	if config.LedgerBatchConfig.LedgersPerFile <= 0 {
+	if dataStore.GetSchema(ctx).LedgersPerFile <= 0 {
```
tamirms, line 61: why does `GetSchema()` require a context parameter?

services/horizon/internal/flags.go
```diff
@@ -850,30 +850,10 @@ func NewAppFromFlags(config *Config, flags support.ConfigOptions) (*App, error)
 }
 
 type ApplyOptions struct {
-	AlwaysIngest                 bool
 	RequireCaptiveCoreFullConfig bool
+	NoCaptiveCore                bool
```
tamirms, line 854: where is this ever set to true?

ingest/ledgerbackend/buffered_storage_backend.go
```diff
@@ -17,13 +17,13 @@ import (
 // Ensure BufferedStorageBackend implements LedgerBackend
 var _ LedgerBackend = (*BufferedStorageBackend)(nil)
 
+type BufferedStorageBackendFactory func(ctx context.Context, config BufferedStorageBackendConfig, dataStore datastore.DataStore) (*BufferedStorageBackend, error)
```
tamirms, line 20: I don't see the benefit of introducing a factory here. We don't use this pattern for the captive core ledgerbackend so I'm not sure why it is necessary for us in this use case.

services/horizon/cmd/db.go
```diff
@@ -307,108 +144,49 @@ func ingestRangeCmdOpts() support.ConfigOptions {
 			FlagDefault: uint(5),
 			Usage:       "[optional] backoff seconds between reingest retries",
 		},
+		{
+			Name:        "ledgerbackend",
+			ConfigKey:   &ledgerBackendStr,
+			OptType:     types.String,
+			Required:    false,
+			FlagDefault: ingest.CaptiveCoreBackend.String(),
+			Usage:       "[optional] Specify the ledger backend type: 'captive-core' (default) or 'datastore'",
```
tamirms, line 153: ingest.CaptiveCoreBackend.String() and ingest.BufferedStorageBackend.String() should be included in the string instead of hard coding 'captive-core' / 'datastore'

services/horizon/cmd/db.go
```diff
@@ -307,108 +144,49 @@ func ingestRangeCmdOpts() support.ConfigOptions {
...
+		},
+		{
+			Name:      "datastore-config",
+			ConfigKey: &storageBackendConfigPath,
+			OptType:   types.String,
+			Required:  false,
+			Usage:     "[optional] Specify the path to the datastore config file (required for datastore backend)",
+			CustomSetValue: func(co *support.ConfigOption) error {
+				val := viper.GetString(co.Name)
+				if ledgerBackendType == ingest.BufferedStorageBackend && val == "" {
```
tamirms, line 176: how do we know that the "ledgerbackend" flag will be evaluated before the "datastore-config" flag?

services/horizon/internal/integration/db_test.go
```diff
@@ -537,16 +547,144 @@ func TestReingestDB(t *testing.T) {
...
+	if err := os.MkdirAll(tempSeedBucketFolder, 0777); err != nil {
+		t.Fatalf("unable to create seed data in temp path, %v", err)
+	}
+
+	err := cp.Copy("./testdata/testbucket", tempSeedBucketFolder)
+	if err != nil {
+		t.Fatalf("unable to copy seed data files for fake gcs, %v", err)
+	}
+
+	testWriter := &testWriter{test: t}
```
tamirms, line 588: instead of creating all these files, I think a better approach would be to read the testdata into a `[]fakestorage.Object` slice and then pass that into the  `fakestorage.NewServer()` constructor:

https://github.com/stellar/go/blob/master/support/datastore/gcs_test.go#L16-L25

services/horizon/internal/integration/db_test.go
```diff
@@ -537,16 +547,144 @@ func TestReingestDB(t *testing.T) {
...
+	webAppDone := make(chan struct{})
+	go func() {
+		defer close(webAppDone)
+		if err = listener.Close(); err != nil {
+			return
+		}
+		webApp.Serve()
+	}()
+
+	defer func() {
```
tamirms, line 637: a lot of this code already exists in the integration test framework. I think you can replace this code by calling integration.NewTest() and modifying it so that it runs horizon with ingest set to false

services/horizon/cmd/db.go
sreuland: yes, I think the safer approach would be to parse the `--ledgerbackend` value from command line string here instead, doing same as the [CustomSetValue](https://github.com/stellar/go/pull/5374/files/efc8da9eb7838929995eb5480c2ca18ec04eefa2#diff-2594a13a4fe6a9473afdc79639c36e7dde657c3e183f33afe8c7b1283eacf8a2R156) does, pull it up to a shared func so not repeated.

ingest/ledgerbackend/buffered_storage_backend.go
sreuland: https://github.com/stellar/go/pull/5374/commits/96284d056dfe93e2d4c626b383ccd82121479ca7

ingest/ledgerbackend/buffered_storage_backend.go
sreuland: yes, the deployment might benefit from externalizing overall for `datastore.DataStoreConfig` with a single external toml file. example of the file [was added to readme](https://github.com/stellar/go/blob/reingest_precomputed/services/horizon/internal/ingest/README.md#reingestion)

ingest/ledgerbackend/buffered_storage_backend.go
sreuland: I think the interface was setting up for upcoming feature..a request to have the [schema be remotely hosted on the datastore, hubble-397 ](https://stellarorg.atlassian.net/browse/HUBBLE-397)- which would leverage context-awareness.

ingest/ledgerbackend/buffered_storage_backend.go
sreuland: it's enables unit testing the ingest `NewSystem` with mock outs on bufferedstorage and datastore's , otherwise becomes hardwired to creating real instances.

services/horizon/internal/flags.go
sreuland: [it is set true](https://github.com/stellar/go/blob/reingest_precomputed/services/horizon/cmd/db.go#L507) when `db reingest --ledgerbackend datastore` is used

services/horizon/cmd/db.go
sreuland: https://github.com/stellar/go/pull/5374/commits/8d543ba21cc4181f11290547696bc7f3311a7876

services/horizon/internal/integration/db_test.go
sreuland: yes,`integration.StartHorizon() `always runs an ingest process and a web process, so, initially went this route to stay clear of it, I'll change `StartHorizon()` to skip creating ingest process when `Config{HorizonEnvironment: map[string]string{"INGEST": "FALSE"}}`

services/horizon/internal/integration/db_test.go
sreuland: yes, refactored from seed path to InitalObjects - https://github.com/stellar/go/pull/5374/commits/d8110f98e3ee9e803e2a238d5ce9d9e3f94f35a5

services/horizon/cmd/db.go
urvisavla: Addressed in 7f1fe6427dd6a1722c434daf730a96301888c287

ingest/ledgerbackend/buffered_storage_backend.go
tamirms: @sreuland `GetSchema()` is called multiple times in the buffered storage backend. It doesn't make sense to do an http request to fetch that info every time when the schema never changes. I think it would be better if the datastore fetched the schema in the constructor and then stored the result in the datastore struct.

ingest/ledgerbackend/buffered_storage_backend.go
tamirms: I looked at the test and it doesn't do anything with the mocked ledger backend. The test just asserts the type of the ledger backend which is constructed. I don't think this test justifies the use of a factory here and in the datastore

urvisavla: That's a good point. addressed in 7f1fe6427dd6a1722c434daf730a96301888c287.

ingest/ledgerbackend/ledger_buffer.go
```diff
@@ -83,19 +83,19 @@ func (bsb *BufferedStorageBackend) newLedgerBuffer(ledgerRange Range) (*ledgerBu
...
 	for i := 0; i <= int(bsb.config.BufferSize); i++ {
-		ledgerBuffer.pushTaskQueue()
+		ledgerBuffer.pushTaskQueue(ctx)
 	}
 
 	return ledgerBuffer, nil
 }
 
-func (lb *ledgerBuffer) pushTaskQueue() {
+func (lb *ledgerBuffer) pushTaskQueue(ctx context.Context) {
```
tamirms, line 92: ctx is not required anymore

ingest/ledgerbackend/ledger_buffer.go
```diff
@@ -180,7 +180,7 @@ func (lb *ledgerBuffer) downloadLedgerObject(ctx context.Context, sequence uint3
 	return objectBytes, nil
 }
 
-func (lb *ledgerBuffer) storeObject(ledgerObject []byte, sequence uint32) {
+func (lb *ledgerBuffer) storeObject(ctx context.Context, ledgerObject []byte, sequence uint32) {
```
tamirms, line 183: ctx is not required here anymore

ingest/ledgerbackend/buffered_storage_backend.go
sreuland: yes, that was the intent of [TestNewSystemBuffered](https://github.com/stellar/go/pull/5374/files#diff-df55dc5a92854f0a58f52c7ff9d81cf2888ed56e94b6aa6ebb3e4a6a6999bc08R114) as unit test, to just exercise NewSystem() code, assert it performs path to create expected buffered backend since we added the [creation of buffered storage backend as a conditional branch](https://github.com/stellar/go/pull/5374/files#diff-933b9cab99da778822646614e836dc6c6b0cae788d582b977e896c86a67bd5d8R295). 

Without the factory abstraction, NewSystem ultimately attempts `NewDataStore`, which fails as it tries to create real gcs connection. If still concern, I will remove the factory abstraction on DataStore and BufferedStorageBackend, and also remove `TestNewSystemBuffered` as a unit test as I think it becomes untenable and we defer to integration tests to exercise NewSystem for buffered backend. In general, am trying to apply an IoC pattern of injection to allow resolution of dependency at runtime rather than compile time.

services/horizon/internal/integration/db_test.go
sreuland: @tamirms , integration.NewTest() is quite 'opinionated' on how a test should have horizon environment provisioned with separate processes for web and ro db conn and ingest and rw db conn. For this test of reingest with datastore it needs a horizon web so it can run a client to verify reingestion data landed in history by using the api, and it needs a rw db conn(with db having full migrations) to pass to the reingestion command, but no ingest process. So, I tried out a change on NewTest/StartHorizon to provide ability for test to have this layout - https://github.com/sreuland/go/commit/58609a20ac9b5843942bf84588058217a458c501

if you think the changeset is not too disruptive of refactor or scope creep, let me know, I'll push it into here, thanks!

ingest/ledgerbackend/buffered_storage_backend.go
tamirms: > I will remove the factory abstraction on DataStore and BufferedStorageBackend, and also remove TestNewSystemBuffered as a unit test

That sounds good to me. As you said, we can defer to the integration tests to exercise that code path

sreuland: https://github.com/stellar/go/pull/5374/commits/08f3ab29ef263c0b3fa3217368d0dd6c3783185b

services/horizon/internal/integration/db_test.go
tamirms: that commit looks good, you can include it in this PR!

sreuland: I stumbled into a subtle aspect of integration.StartHorizon, some tests such a parameters_test, intentionally send mix of params that will result in an err on horizon cmd execute, but in that case an App instance is still created and it has a db conn established, i found a more direct way to clean that up from test framework side, and remove unnecessary Shutdown and sleep calls from tests - https://github.com/stellar/go/pull/5374/commits/1642a1ae76ea12c8a22b913f135f38957f5ff60e

PR #5377, opened 2024-07-08
add unit test coverage on buffered backend ingest system creation

sreuland:
adding unit test to confirm NewSystem works with buffered ledger backend config.


urvisavla (approved):
lgtm!

PR #5383, opened 2024-07-11
horizon/integration/test: added test for reingest with --ledgerbackend datastore

sreuland:
uses fakegcs, seeded with some prebuilt export files, ingests that from gcs using `db reingest --ledgerbackend datastore` and then verifies through horizon web api that the ledger exists in history db.

this took a bit of detour to provide an instance safe version of RootCmd also for testing purposes, avoiding flag/arg collisions during tests.


services/horizon/internal/integration/db_test.go
```diff
@@ -549,6 +558,131 @@ func TestReingestDB(t *testing.T) {
...
+
+	testTempDir := t.TempDir()
+	tempSeedDataPath := filepath.Join(testTempDir, "data")
+	tempSeedBucketPath := filepath.Join(tempSeedDataPath, "path", "to", "my", "bucket")
+	tempSeedBucketFolder := filepath.Join(tempSeedBucketPath, "FFFFFFFF--0-63999")
+	if err := os.MkdirAll(tempSeedBucketFolder, 0777); err != nil {
+		t.Fatalf("unable to create seed data in temp path, %v", err)
+	}
+
+	err := cp.Copy("./testdata/testbucket", tempSeedBucketFolder)
```
urvisavla, line 580: Does this need to be cleaned up before after the test is complete?

sreuland: it's a sub folder on the `testing.TempDir()`, so it should be cleaned up automatically after the test ends, which that is supposed to ensure.

urvisavla (approved):

PR #5385, opened 2024-07-11
services/horizon/internal/ingest: Add more fine-grained reap metrics

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add duration metrics for reaping of lookup tables where we can see how much time is spent reaping each table. See link below which displays the new metrics on staging:

https://grafana.stellar-ops.com/d/x8xDSQQIk/stellar-horizon?orgId=1&viewPanel=2531

Also, add metrics which count how many rows were deleted from each lookup table. See link below which displays the new metrics on staging:

https://grafana.stellar-ops.com/d/x8xDSQQIk/stellar-horizon?orgId=1&viewPanel=2541
 
### Why

These fine-grained metrics are required so that we can analyze which tables are the most expensive to reap.

### Known limitations

[N/A]

### DIFF
--- a/services/horizon/internal/ingest/main.go
+++ b/services/horizon/internal/ingest/main.go
@@ -145,9 +145,8 @@ type Metrics struct {
 	// duration of rebuilding trade aggregation buckets.
 	LedgerIngestionTradeAggregationDuration prometheus.Summary
 
-	// LedgerIngestionReapLookupTablesDuration exposes timing metrics about the rate and
-	// duration of reaping lookup tables.
-	LedgerIngestionReapLookupTablesDuration prometheus.Summary
+	ReapDurationByLookupTable *prometheus.SummaryVec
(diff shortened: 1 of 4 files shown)


services/horizon/internal/ingest/main.go
```diff
@@ -327,6 +326,7 @@ func NewSystem(config Config) (System, error) {
 			config.ReapConfig,
 			config.HistorySession,
 		),
+		reapOffsets: map[string]int64{},
```
sreuland, line 329: minor, for consideration, make this a type, like:
```
type ReapOffsetByTable map[string][]int64
```

or change the name to indicate key expectation also, i.e. it's a table name correct?

sreuland (approved):

PR #5398, opened 2024-07-19
Fix CI test workflow errors in RecoverySigner unit tests 

urvisavla:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Recently, we've been seeing the following error in the test workflow `Received unexpected error: pq: duplicate key value violates unique constraint 'pg_authid_rolname_index'` in the RecoverySigner unit test. The first occurrence of this error in recent history was [here](https://github.com/stellar/go/commit/01dc1762d6be40d84829476ee154834d88490155). After committing the [PR](https://github.com/stellar/go/commit/a3fae02c3b634b353dc4c37597ccc1da78f46317) that moves Galexie out of experimental, the issue became exacerbated and started failing every time. The RecoverySigner unit test always fails, but not always in the same test, though the error remains the same.

This issue arises because multiple concurrent tests are trying to create the same role, resulting in a race condition. It is unclear why we consistently see this issue now; it may be due to the order in which the tests are executed changing after the relocation of Galexie out of experimental.

The solution is to check for the specific error and if it is a duplicate key violation, we know that the role has been created by another concurrent transaction in which case we ignore the error. Any other type of error during role creation will result in the tests failing.

### Why
Resolve CI test workflow errors.

### DIFF
--- a/support/db/dbtest/db.go
+++ b/support/db/dbtest/db.go
@@ -127,6 +127,12 @@ func checkReadOnly(t testing.TB, DSN string) {
 
 	if !rows.Next() {
 		_, err = tx.Exec("CREATE ROLE user_ro WITH LOGIN PASSWORD 'user_ro';")
+		if err != nil {
+			// Handle race condition by ignoring the error if it's a duplicate key violation
+			if pqErr, ok := err.(*pq.Error); ok && pqErr.Code == "23505" {
+				return
+			}
+		}
 		require.NoError(t, err)
 	}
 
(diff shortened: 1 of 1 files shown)


support/db/dbtest/db.go
```diff
@@ -127,6 +127,12 @@ func checkReadOnly(t testing.TB, DSN string) {
```
tamirms, line 127: I think this function used a serializable transaction with the goal of avoiding this exact race conditoin. However, it seems that serializable transactions behave differently with catalog tables (see: https://stackoverflow.com/questions/47400084/postgresql-serializable-transactions-with-grant-drop-role ).

I think your solution is good but the only change I would add is to get rid of the serializable transaction since I don't think it's necessary given that you already check for the duplicate key violation.

tamirms (approved):

PR #5399, opened 2024-07-24
services/friendbot: support funding existing accounts

leighmcculloch:
### What
Support funding existing accounts in Friendbot.

If an account does not exist, do as done today, create the account with the starting balance.

If an account exists, and balance is equal to or above the starting balance, error as the account is funded sufficiently.

If an account exists, and balance is below the starting balance, send a payment to the account of the starting balance.

### Why
The stellar-cli supports funding accounts and there's a request to support funding to top up existing accounts that have spent funds (https://github.com/stellar/stellar-cli/issues/1389).

There are reasons why this shouldn't be done. It's possible for someone to more easily exhaust the testnet native balance available. However, it is really already very easy to do that by simply passing random addresses. And because this change doesn't allow unlimited funding of a single account without moving funds out of the account the same amount of friction will exist as it does today to someone wishing to accumulate lots of testnet native.

Close https://github.com/stellar/stellar-cli/issues/1389

### DIFF
--- a/services/friendbot/internal/minion.go
+++ b/services/friendbot/internal/minion.go
@@ -3,6 +3,7 @@ package internal
 import (
 	"fmt"
 
+	"github.com/stellar/go/amount"
 	"github.com/stellar/go/clients/horizonclient"
 	"github.com/stellar/go/keypair"
 	hProtocol "github.com/stellar/go/protocols/horizon"
@@ -28,7 +29,7 @@ type Minion struct {
 	// Mockable functions
 	SubmitTransaction    func(minion *Minion, hclient horizonclient.ClientInterface, tx string) (*hProtocol.Transaction, error)
 	CheckSequenceRefresh func(minion *Minion, hclient horizonclient.ClientInterface) error
-	CheckAccountExists   func(minion *Minion, hclient horizonclient.ClientInterface, destAddress string) (bool, error)
+	CheckAccountExists   func(minion *Minion, hclient horizonclient.ClientInterface, destAddress string) (bool, string, error)
 
 	// Uninitialized.
 	forceRefreshSequence bool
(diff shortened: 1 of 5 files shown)


chadoh (approved):
Love it, thanks!

services/friendbot/internal/minion.go
```diff
@@ -138,11 +213,43 @@ func (minion *Minion) makeTx(destAddress string) ([32]byte, string, error) {
 	if err != nil {
 		return [32]byte{}, "", errors.Wrap(err, "unable to hash")
 	}
+	return txh, txe, err
+}
 
-	// Increment the in-memory sequence number, since the tx will be submitted.
-	_, err = minion.Account.IncrementSequenceNumber()
+func (minion *Minion) makePaymentTx(destAddress string) ([32]byte, string, error) {
+	createAccountOp := txnbuild.Payment{
```
Shaptic, line 220: nit: typo? threw me off heh
```suggestion
	paymentOp := txnbuild.Payment{
```

services/friendbot/internal/minion.go
```diff
@@ -138,11 +213,43 @@ func (minion *Minion) makeTx(destAddress string) ([32]byte, string, error) {
...
+
+	tx, err = tx.Sign(minion.Network, minion.Keypair, minion.BotKeypair)
+	if err != nil {
+		return [32]byte{}, "", errors.Wrap(err, "unable to sign tx")
+	}
+
+	txe, err := tx.Base64()
+	if err != nil {
+		return [32]byte{}, "", errors.Wrap(err, "unable to serialize")
+	}
```
Shaptic, line 247: <rant
It's so annoying to have to handle errors that are completely out of the realm of possibility of occurring lol
/rant>

Shaptic (commented):
Coupla comments, just waiting on the Slack discussion to :heavy_check_mark:

services/friendbot/internal/minion.go
leighmcculloch: 💯

services/friendbot/internal/minion.go
```diff
@@ -105,7 +157,30 @@ func (minion *Minion) checkHandleBadSequence(err *horizonclient.Error) {
...
-func (minion *Minion) makeTx(destAddress string) ([32]byte, string, error) {
+func (minion *Minion) checkBalance(balance string) error {
+	bal, err := amount.ParseInt64(balance)
+	if err != nil {
+		return errors.Wrap(err, "cannot parse account balance")
+	}
+	starting, err := amount.ParseInt64(minion.StartingBalance)
+	if err != nil {
+		return errors.Wrap(err, "cannot parse starting balance")
+	}
```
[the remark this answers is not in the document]
leighmcculloch: I originally wrote it that way. I looked too at if the starting balance could be converted once and stored as int64. But then I was also converting back to strings at the entry/exit points of the SDK and it felt like I was fighting the SDKs decision that amounts should always be strings.

services/friendbot/internal/minion.go
leighmcculloch: fc14d1bd

Shaptic (approved):

Issue #4870, opened 2023-05-04
[Epic] Improving Reap Performance of History Lookup Tables


tamirms:
This is an epic tracking the overall implementation of the reaping solution proposed in [Improving Reap Performance of History Lookup Tables](https://docs.google.com/document/d/1RiW5vCqEAy-evteAc5Ap5NRZsJ16jN7nJn4igUazMJs/edit#)

1. https://github.com/stellar/go/issues/4864
2. https://github.com/stellar/go/issues/4865
3. https://github.com/stellar/go/issues/4866
4. https://github.com/stellar/go/issues/4867
5. https://github.com/stellar/go/issues/4868
6. https://github.com/stellar/go/issues/4869

PR #5405, opened 2024-07-30
services/horizon/internal/ingest: reap lookup tables without blocking ingestion

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Close https://github.com/stellar/go/issues/4870

This PR improves reaping of history lookup tables (e.g. history_accounts, history_claimable_balances) so that it can run safely in parallel with ingestion. Currently, reaping of history lookup tables is a blocking operation for ingestion so if the queries to reap history lookup tables take too long that can result in ingestion lag. With this PR, reaping of history lookup tables will be able to run concurrently to ingestion with minimal contention. Also, it is important to note that this PR  does not add any performance degradation for either reingestion or live ingestion.

When reviewing this PR it would be helpful to read this design doc:

https://docs.google.com/document/d/1CGfBCS99MTEZDP4mMhV1o6Z5NE_Tlg7ENCcWTwzhlio/edit 

### Known limitations

After running a full vacuum on history_accounts, the reaping query sped up dramatically. Previously, the duration of reaping the history_accounts table peaked at ~1.9 seconds:

https://grafana.stellar-ops.com/d/x8xDSQQIk/stellar-horizon?orgId=1&from=1722295775773&to=1722400061302&var-environment=stg&var-cluster=pubnet&var-network=All&var-route=All&viewPanel=2531

Screenshot 2024-08-28 at 7 23 34 PM

After the vacuum, the average duration for reaping history_accounts is ~20 ms and the peak duration was ~400 ms:

https://grafana.stellar-ops.com/d/x8xDSQQIk/stellar-horizon?orgId=1&from=1724782666959&to=1724869066959&var-environment=stg&var-cluster=pubnet&var-network=All&var-route=All&viewPanel=2531

Screenshot 2024-08-28 at 7 18 34 PM

This means that the risk that reaping of history lookup tables taking so long that it introduces ingestion lag is a lot less of a concern.

Update:

After running reaping of history lookup tables on staging for 24 hours I have observed that the peak duration actually reaches 600 ms. 

https://grafana.stellar-ops.com/d/x8xDSQQIk/stellar-horizon?orgId=1&from=1724866821793&to=1724953221793&var-environment=stg&var-cluster=pubnet&var-network=All&var-route=All&viewPanel=2531

Screenshot 2024-08-29 at 6 40 17 PM


services/horizon/internal/db2/history/account_loader.go
```diff
@@ -134,17 +165,34 @@ func (l *loader[K, T]) Exec(ctx context.Context, session db.SessionInterface) er
...
-		return err
+		if count, err := l.insert(ctx, q, keys); err != nil {
+			return err
+		} else {
+			l.stats.Total += count
+			l.stats.Inserted += count
+		}
 	} else {
-		l.stats.Total += count
+		return fmt.Errorf("concurrency mode %v is invalid", l.concurrencyMode)
```
sreuland, line 195: can this be caught earlier in NewAccountLoader instead?

tamirms: yes, but I would prefer that this check is performed when concurrencyMode is used. In theory if we check in the constructor that should be sufficient but the type system does not guarantee that concurrencyMode cannot take on invalid values even though we validated it in the constructor

services/horizon/internal/db2/history/account_loader.go
```diff
@@ -17,6 +17,35 @@ import (
 
 var errSealed = errors.New("cannot register more entries to Loader after calling Exec()")
 
+// ConcurrencyMode is used to configure the level of thread-safety for a loader
+type ConcurrencyMode int
```
sreuland, line 21: nit, this is used through package, not specific to account loader, maybe relocate to ./main.go instead.

services/horizon/internal/db2/history/account_loader.go
```diff
@@ -204,11 +252,15 @@ func (l *loader[K, T]) insert(ctx context.Context, q *Q, keys []K) (int, error)
 	return len(rows), nil
 }
 
-func (l *loader[K, T]) query(ctx context.Context, q *Q, keys []K) (int, error) {
+func (l *loader[K, T]) query(ctx context.Context, q *Q, keys []K, lockRows bool) (int, error) {
```
sreuland, line 255: for consideration, if this is the real place that knows how concurrency mode affects row locking behavior, then encapsulate that locking decision in here and remove `lockRows` parameter so caller's don't need that know how.

services/horizon/internal/db2/history/key_value.go
```diff
@@ -207,41 +206,26 @@ func (q *Q) getValueFromStore(ctx context.Context, key string, forUpdate bool) (
...
+		if errors.Cause(err) == sql.ErrNoRows {
+			return 0, nil
 		}
-		offsets[table] = offset
+		return 0, err
+	}
+	var offset int64
+	offset, err = strconv.ParseInt(text, 10, 64)
+	if err != nil {
+		return 0, fmt.Errorf("invalid offset: %s", text)
```
sreuland, line 226:
```suggestion
		return 0, fmt.Errorf("invalid offset: %s for table %s", text, table)
```

services/horizon/internal/db2/history/loader_concurrency_test.go
```diff
@@ -0,0 +1,189 @@
...
+	s2 := s1.Clone()
+
+	for _, testCase := range []struct {
+		mode ConcurrencyMode
+		pass bool
+	}{
+		{ConcurrentInserts, true},
+		{ConcurrentDeletes, false},
+	} {
+		t.Failed()
```
sreuland, line 31: what is this meant for, debug code?

services/horizon/internal/db2/history/loader_concurrency_test.go
```diff
@@ -0,0 +1,189 @@
...
+			var wg sync.WaitGroup
+			wg.Add(1)
+			go func() {
+				defer wg.Done()
+				<-time.After(time.Second * 3)
+				assert.NoError(t, s1.Commit())
+			}()
+			assert.NoError(t, l2.Exec(context.Background(), s2))
+			assert.NoError(t, s2.Commit())
+			wg.Wait()
```
sreuland, line 65: what is the significance of doing s1.Commit() from a separate go routine? would doing`s1.Commit()` on ln 65 w/o the async wait equate to the same coverage? it appears the intent is to have second loader tx always commit before first loader tx, correct?

services/horizon/internal/db2/history/loader_concurrency_test.go
```diff
@@ -0,0 +1,189 @@
...
+				Inserted: 10,
+			}, l1.Stats())
+
+			if testCase.pass {
+				assert.Equal(t, LoaderStats{
+					Total:    15,
+					Inserted: 5,
+				}, l2.Stats())
+			} else {
+				assert.NotEqual(t, LoaderStats{
```
sreuland, line 78: are the second loader stats deterministic in this case? if so, would be interesting to demonstrate what expected value is here.

services/horizon/internal/db2/history/loader_concurrency_test.go
```diff
@@ -0,0 +1,189 @@
...
+				assert.Equal(t, l1Id, l2Id)
+
+				var account Account
+				assert.NoError(t, q.AccountByAddress(context.Background(), &account, address))
+				assert.Equal(t, account.ID, l1Id)
+				assert.Equal(t, account.Address, address)
+			}
+
+			for _, address := range addresses[10:] {
+				l2Id, err := l2.GetNow(address)
```
sreuland, line 101: great coverage here! is it worthwhile to assert the first loader, `l1.GetNow()` expectation in this case also, i.e. should it return err on these addresses, as loader instances shouldn't share any state?

services/horizon/internal/db2/history/main.go
```diff
@@ -977,63 +980,63 @@ type LookupTableReapResult struct {
...
+func (q *Q) FindLookupTableRowsToReap(ctx context.Context, table string, batchSize int) ([]int64, int64, error) {
+	offset, err := q.getLookupTableReapOffset(ctx, table)
+	if err != nil {
+		return nil, 0, fmt.Errorf("could not obtain offsets: %w", err)
 	}
 
-	offsets, err := q.getLookupTableReapOffsets(ctx)
+	// Find new offset before removing the rows
+	var newOffset int64
+	err = q.GetRaw(ctx, &newOffset, fmt.Sprintf("SELECT id FROM %s where id >= %d limit 1 offset %d", table, offset, batchSize))
```
sreuland, line 991: should `ORDER BY id` be included here since limit/offset usage, it seems PG [docs](https://www.postgresql.org/docs/current/queries-order.html#QUERIES-ORDER) mention indeterminate if `ORDER BY` absent

sreuland: also, the usage of `OFFSET=config_provided_batchsize` might lead to some performance degradation if the batchsize configured is large, do you think the same result can be achieved without OFFSET such as with `SELECT id FROM %s where id >= {offset+batchSize} limit 1` ?

services/horizon/internal/db2/history/main.go
```diff
@@ -977,63 +980,63 @@ type LookupTableReapResult struct {
...
+	return rowsDeleted, nil
+}
 
-		results[table] = LookupTableReapResult{
-			Offset:      newOffset,
-			RowsDeleted: rows,
-			Duration:    time.Since(startTime),
+func (q *Q) reapLookupTable(ctx context.Context, table string, ids []int64, newOffset int64) (int64, error) {
+	if err := q.updateLookupTableReapOffset(ctx, table, newOffset); err != nil {
+		return 0, fmt.Errorf("error updating offset: %w", err)
```
sreuland, line 1028:
```suggestion
		return 0, fmt.Errorf("error updating offset for table %s: %w ", table, err)
```

services/horizon/internal/db2/history/main.go
```diff
@@ -1125,29 +1128,77 @@ var historyLookupTables = map[string][]tableObjectFieldPair{
 // possible that rows will be skipped from deletion. But offset is reset
```
sreuland, line 1128: may need to refresh the docs here to reflect name/functionality change.

services/horizon/internal/db2/history/main.go
```diff
@@ -1125,29 +1128,77 @@ var historyLookupTables = map[string][]tableObjectFieldPair{
...
 // deleted.
-func constructReapLookupTablesQuery(table string, historyTables []tableObjectFieldPair, batchSize int, offset int64) string {
+func (q *Q) deleteLookupTableRows(ctx context.Context, table string, ids []int64) (int64, error) {
+	deleteQuery := constructDeleteLookupTableRowsQuery(table, ids)
+	result, err := q.ExecRaw(
+		context.WithValue(ctx, &db.QueryTypeContextKey, db.DeleteQueryType),
+		deleteQuery,
+	)
+	if err != nil {
+		return 0, fmt.Errorf("error running query: %w", err)
```
sreuland, line 1138: may want to include the full sql in the err message from `deleteQuery` for debugging.

services/horizon/internal/db2/history/verify_lock.go
```diff
@@ -15,7 +15,8 @@ const (
 	// reaperLockId is the objid for the advisory lock acquired during
```
sreuland, line 15:
```suggestion
	// reaper lock ids are unique objid for the advisory lock acquired during
```

services/horizon/internal/ingest/main.go
```diff
@@ -369,6 +366,7 @@ func NewSystem(config Config) (System, error) {
 			config.ReapConfig,
 			config.HistorySession,
 		),
+		lookupTableReaper: newLookupTableReaper(config.HistorySession),
```
sreuland, line 369: can `lookupTableReaper` be a field on `Reaper` and constructed internally by NewReaper rather than separate?

sreuland:
one edge case wanted to check on, if a user reingests an older range which goes further back than the retention period cutoff, and reaping for data and lookup tables has already completed for that retention period, will the next iteration of lookup reaper sense those and delete the qualified(orhpaned) lookup ids in that case? I ask b/c of the offsets for reapers that are stored in key-value, it seems like once those advance, the reaper won't inspect that older id range anymore?

services/horizon/internal/db2/history/account_loader.go
tamirms: concurrency mode affects the loaders in 2 ways:

1. we lock rows when querying if we allow concurrent deletes
2. the ordering changes depending on the concurrency mode. if we allow concurrent deletes we first query and then insert. whereas if we allow concurrent inserts, we first insert and then we query.

I prefer that both of those behaviors are encapsulated in Exec()

services/horizon/internal/ingest/main.go
tamirms: the two reaping workflows are independent so I'm not sure what the advantage is of coupling them

services/horizon/internal/db2/history/main.go
tamirms: > do you think the same result can be achieved without OFFSET such as with SELECT id FROM %s where id >= {offset+batchSize} limit 1 ?

unfortunately, I don't think that will always work because the ids are not guaranteed to be consecutive. In any case, even discarding the OFFSET issue, performance of reaping will degrade if the batch size is too large because the queries to identify the orphaned rows are expensive

services/horizon/internal/db2/history/key_value.go
```diff
@@ -207,41 +206,26 @@ func (q *Q) getValueFromStore(ctx context.Context, key string, forUpdate bool) (
...
-		if _, ok := historyLookupTables[table]; !ok {
-			return nil, fmt.Errorf("invalid key: %s", pair.Key)
-		}
-
-		var offset int64
-		offset, err = strconv.ParseInt(pair.Value, 10, 64)
-		if err != nil {
-			return nil, fmt.Errorf("invalid offset: %s", pair.Value)
+		if errors.Cause(err) == sql.ErrNoRows {
+			return 0, nil
```
tamirms, line 219: @sreuland this is the code where we detect that we have searched through all the rows of the history lookup table. So, we return 0 to start from the very beginning again.

tamirms:
@sreuland I believe I have addressed your feedback. PTAL, thanks!

services/horizon/internal/db2/history/key_value.go
sreuland: ah, cool, that was my follow up question, thanks for clarification!

sreuland (approved):
looks great, nice work!

PR #5410, opened 2024-08-05
services/horizon/internal/actions: populate default cursor value for history endpoints

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

The following endpoints use a [TOID](https://github.com/stellar/go/blob/master/toid/main.go#L8) cursor value derived from the ledger sequence number:

/effects
/operations
/transactions
/ledgers
/trades
/payments

When a cursor is omitted the query to fetch the history data will look like:

`select * from history_transactions order by id asc` (if the order is asc which is by default))

or 

`select * from history_transactions order by id desc` (if the order is desc)

However, the query performs better with a paging clause, eg:

`select * from history_transactions where id >= 200076760811839488 order by id asc`

 
This commit sets a default cursor value based on the paging order. If order is set to asc, we will use the oldest ledger -1 to create the default cursor value. Similarly, if the order is set to desc, we will use the latest ledger + 100 to create the default cursor value.
 
### Known limitations

We obtain the oldest and latest ledger using `ledgerState *ledger.State` which is refreshed periodically. It is possible that `ledgerState` is out of date but that is why we add a buffer of 100 extra ledgers to `ledgerState.CurrentStatus().HistoryLatest`.

### DIFF
--- a/services/horizon/internal/actions/helpers.go
+++ b/services/horizon/internal/actions/helpers.go
@@ -226,7 +226,7 @@ func GetPageQuery(ledgerState *ledger.State, r *http.Request, opts ...Opt) (db2.
 	if cursor == "" && defaultTOID {
 		if pageQuery.Order == db2.OrderAscending {
 			pageQuery.Cursor = toid.AfterLedger(
-				ordered.Max(1, ledgerState.CurrentStatus().HistoryElder-1),
+				ordered.Max(0, ledgerState.CurrentStatus().HistoryElder-1),
 			).String()
 		} else if pageQuery.Order == db2.OrderDescending {
 			pageQuery.Cursor = toid.AfterLedger(
--- a/services/horizon/internal/actions/helpers.go
+++ b/services/horizon/internal/actions/helpers.go
@@ -21,6 +21,7 @@ import (
 	"github.com/stellar/go/services/horizon/internal/ledger"
 	hProblem "github.com/stellar/go/services/horizon/internal/render/problem"
 	"github.com/stellar/go/support/errors"
+	"github.com/stellar/go/support/ordered"
 	"github.com/stellar/go/support/render/problem"
 	"github.com/stellar/go/toid"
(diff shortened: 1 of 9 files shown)


urvisavla (approved):
LGTM!  Nice solution to addresses the issue of 503 timeouts for ASC. The same default cursor logic isn't required for DESC but I'm assuming there's no downside to using it for DESC as well.

tamirms:
@urvisavla good point regarding the order=desc behavior. I've decided to restore the original behavior for desc to limit the footprint of this PR

services/horizon/internal/actions/helpers.go
```diff
@@ -214,6 +221,13 @@ func GetPageQuery(ledgerState *ledger.State, r *http.Request, opts ...Opt) (db2.
 
 		return db2.PageQuery{}, err
 	}
+	if cursor == "" && defaultTOID {
```
sreuland, line 224: is it worth encapsulating this cursor behavior internally in `db2.NewPageQuery(...)` ? it looks like the method processes the other cursor 'disabled' behavior already.

services/horizon/internal/actions/helpers_test.go
```diff
@@ -290,6 +291,63 @@ func TestGetPageQuery(t *testing.T) {
 	tt.Assert.Error(err)
 }
 
+func TestGetPageQueryCursorDefaultTOID(t *testing.T) {
```
sreuland, line 294: great coverage.

sreuland (approved):
looks good, left one minor comment for consideration.

services/horizon/internal/actions/helpers.go
tamirms: I thought about it but `db2.NewPageQuery()` doesn't have access to the ledger state which is required to compute the default cursor value

Issue #4864, opened 2023-05-04
Simplify history.CreateAccounts() to avoid select statement


tamirms:
The [CreateAccounts](https://github.com/stellar/go/blob/master/services/horizon/internal/db2/history/account.go#L27-L80) , [CreateHistoryClaimableBalances](https://github.com/stellar/go/blob/master/services/horizon/internal/db2/history/history_claimable_balances.go#L22-L68) ,  [CreateAssets](https://github.com/stellar/go/blob/master/services/horizon/internal/db2/history/asset.go#L36), and [CreateHistoryLiquidityPools](https://github.com/stellar/go/blob/master/services/horizon/internal/db2/history/history_liquidity_pools.go#L21-L77) functions are used to insert records into the `history_accounts`, `history_claimable_balances`, `history_assets`, and `history_liquidity_pools` tables. The functions are implemented by first executing insert statements into the respective tables followed by executing select statements to lookup the rows which where just inserted.

However, the select statements can be avoided because we can add a [`RETURNING`](https://www.postgresql.org/docs/current/dml-returning.html) clause to the insert statements which will return the row which was just inserted.

urvisavla:
The `RETURNING` clause in an `INSERT` statement only returns the newly inserted rows. To return all rows, including existing ones, we can force an update using the `ON CONFLICT UPDATE SET... ` clause. However this additional update will likely have a negative impact on performance.

To improve reap performance, we plan to add a [`last_Reference_ledger`](https://github.com/stellar/go/issues/4867) column to these tables. This column will be updated every time which will allow us to use the `RETURNING` clause. However, when reingesting older ledgers, we may not want to update the `last_Reference_ledger` if it is greater than the one we are reingesting. So at this point I'm not sure that we can combine the two statements.

PR #5415, opened 2024-08-07
services/horizon/internal/db2/history: Insert and query rows from history lookup tables with one query

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

fix https://github.com/stellar/go/issues/4864

Currently, the loaders for the history lookup tables attempt to get or create history ids by issuing up to 3 SQL statements. First, we query for any existing rows. Then we insert the rows that are not present. Finally, we query the rows that were just inserted.

In https://github.com/stellar/go/issues/4864 we discussed that if we included a [RETURNING](https://www.postgresql.org/docs/current/dml-returning.html) clause to the insert statements we could obtain all the rows both present and inserted in a single query.

However, @urvisavla pointed out:

>The RETURNING clause in an INSERT statement only returns the newly inserted rows. To return all rows, including existing ones, we can force an update using the ON CONFLICT UPDATE SET...  clause. However this additional update will likely have a negative impact on performance.

So that means the [RETURNING](https://www.postgresql.org/docs/current/dml-returning.html) clause could reduce the number of SQL statements from 3 to 2 but we would still need to do a SQL statement to find all the pre-existing rows.
 
~But, it turns out we can use a [UNION](https://www.postgresql.org/docs/current/queries-union.html) clause to combine the `INSERT ... RETURNING *` statement with another query which just selects all the pre-existing rows.  This technique is described in more detail here:~

~https://hakibenita.com/postgresql-get-or-create~

### Known limitations

Unfortunately when deploying this PR to staging I did not observe any significant improvements to the "Ingestion Loaders Duration" metric. ~~Also, with this change we lose insight into the number of rows which were inserted vs queried.~~ Nevertheless, I do think it's still worth including this change as it reduces code complexity


services/horizon/internal/db2/history/account_loader.go
```diff
@@ -122,47 +101,32 @@ func (a *AccountLoader) Exec(ctx context.Context, session db.SessionInterface) e
 	for address := range a.set {
 		addresses = append(addresses, address)
 	}
-
-	if err := a.lookupKeys(ctx, q, addresses); err != nil {
```
urvisavla, line 126: Can you remind me why the original code used an initial select query instead of just doing an insert followed by a select to fetch both existing and new records?

services/horizon/internal/db2/history/account_loader.go
```diff
@@ -201,20 +165,28 @@ func bulkInsert(ctx context.Context, q *Q, table string, conflictFields []string
 		)
 	}
 
+	columns := strings.Join(insertFieldsPart, ",")
```
urvisavla, line 168: Do you think it would be helpful to add a comment explaining in short what the SQL statement does?

services/horizon/internal/db2/history/account_loader.go
```diff
@@ -201,20 +165,28 @@ func bulkInsert(ctx context.Context, q *Q, table string, conflictFields []string
...
+	WITH rows AS
+		(SELECT ` + strings.Join(unnestPart, ",") + `),
+	inserted_rows AS (
+		INSERT INTO ` + table + `
+			(` + columns + `)
+		SELECT * FROM rows
+		ON CONFLICT (` + columns + `) DO NOTHING
+		RETURNING *
+	)
+	SELECT * FROM inserted_rows
```
urvisavla, line 179: The description mentions that we lose the count of rows inserted versus queried. Can we not use COUNT(*) here to get the number of rows inserted?

services/horizon/internal/db2/history/account_loader.go
tamirms: previously I thought doing an insert followed by a select would lock up all the rows which would cause contention when doing parallel ingestion (imagine two separate workers are ingesting a ledger range which have common account ids). However, it turns out that the `on conflict do nothing` clause will prevent pre-existing rows from being locked in the insert statement. so in summary, there was no good reason to do a select, followed by insert, followed by select

tamirms:
@urvisavla can you take another look? I believe I've addressed all the feedback

tamirms:
Actually it turns out there's a possible race condition in the get or create query where the query can return 0 rows if there are 2 transactions attempting to insert the same data concurrently. See https://stackoverflow.com/a/15950324 for more details. I will need to update this PR

tamirms:
@urvisavla now it's ready for review again. I fixed the race condition but unfortunately we cannot do both insertion and querying within a single sql statement

urvisavla (approved):
So we're essentially removing the initial select, updating the insert to return the inserted rows, and then adding a select to fetch only the remaining rows, correct? 

Also, I'm assuming there's no need to check the performance of the change on staging since we're no longer combining the queries, right?

Changes look good to me. Left a couple of comments for consideration. Also the description needs update.

services/horizon/internal/db2/history/account_loader.go
```diff
@@ -24,8 +23,6 @@ type FutureAccountID struct {
 	loader  *AccountLoader
 }
 
-const loaderLookupBatchSize = 50000
```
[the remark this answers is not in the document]
tamirms: previously we didn't use `UNNEST` to get around the [postgres parameters limit](https://klotzandrew.com/blog/postgres-passing-65535-parameter-limit) and so we had to limit the sizes of how many values we could place in the `WHERE (...) IN (...)` clause

tamirms:
@urvisavla 

> So we're essentially removing the initial select, updating the insert to return the inserted rows, and then adding a select to fetch only the remaining rows, correct?

yes, that's correct

> Also, I'm assuming there's no need to check the performance of the change on staging since we're no longer combining the queries, right?

I'm still checking the performance of staging to make sure there are no regressions

I also updated the PR to reduce code duplication between all 4 loaders. PTAL, thanks!

PR #5428, opened 2024-08-15
clients/stellarcore: Add support for Core's new `/getledgerentryraw`

### LINKED ISSUE #5426: clients/stellarcore: Add support for Core's new HTTP endpoints
# Protocol 23: Core Endpoints

**Epics**: https://github.com/stellar/go/issues/5433, stellar/soroban-rpc#267

---------------------------

Stellar Core will have one new high-performance HTTP endpoint as part of their v22 in stellar/stellar-core#4350 which is supported as part of https://github.com/stellar/go/pull/5428. It will have three more in v23. We need to support them. The endpoint schemas are defined in [this document](https://docs.google.com/document/d/1BXXVtYkSjN3lsZkrGdnCw2WjtYND4JD7f5Ak0zdHC4o/edit) and are replicated here (please check the document to ensure the following schemas are not outdated).

### Request format
Each endpoint needs to be implemented into the `stellarcore` client as its described here. The format for every endpoint is the same POST request:

```
ledgerSeq=N&key=base64&key=base64...
```

Here, `ledgerSeq` is an optional `uint32` number and each `key` is a base64-encoded `xdr.LedgerKey`. If the `ledgerSeq` is not available on the given instance, a 404 will be returned.

#### **Protocol 23**: `/getledgerentry`
JSON response (as a TypeScript interface):

```typescript
interface Response {
  ledger: number; // uint32
  entries: {
    e: string; // base64-encoded xdr.LedgerEntry or xdr.LedgerKey (if state == new)
    state: "live" | "new" | "archived" | "evicted"
  }[]
```

Note that this is REPLACING the existing endpoint, meaning we need to ensure that all code paths that use the current variant of the endpoint are updated correctly.

---------------------------

# Protocol 24: Core Endpoints

The details are omitted here so they can be moved to a new issue when appropriate, since the endpoint schemas might change.

<details>

### Request format
Each endpoint needs to be implemented into the `stellarcore` client as its described here. The format for every endpoint is the same POST request:

```
ledgerSeq=N&key=base64&key=base64...
```

Here, `ledgerSeq` is an optional `uint32` number and each `key` is a base64-encoded `xdr.LedgerKey`. If the `ledgerSeq` is not available on the given instance, a 404 will be returned.

#### Protocol 24: `/getinvocationproof`
JSON response (as a TypeScript interface):

```typescript
interface Response {
  ledger: number; // uint32
  proof?: string; // base64-encoded xdr.ArchivalProof
}
```

Note that if a proof is not required for any of the given entries, `proof` will be omitted.

#### Protocol 24: `/getrestorationproof`
JSON response (as a TypeScript interface):

```typescript
interface Response {
  ledger: number; // uint32
  proof?: string; // base64-encoded xdr.ArchivalProof
}
```

Note that if a proof is not required for any of the given entries, `proof` will be omitted.

## Notes
Core does not have a build ready with the Protocol 24 endpoints so they cannot be tested end-to-end.

</details>

### LINKED ISSUE #5455: clients/stellarcore: Add support for Core's `getledgerentryraw` endpoint
Stellar core recently introduced a new HTTP endpoint `getledgerentryraw` in the p22 release: https://github.com/stellar/stellar-core/pull/4350.

## Request Format

`getledgerentryraw` needs to be implemented into the `stellarcore` client as its described here. The format is a POST request as follows:
```
ledgerSeq=NUM&key=Base64&key=Base64...
```
  * `ledgerSeq`: An optional parameter, specifying the ledger snapshot to base the query on.
  If the specified ledger in not available, a 404 error will be returned. If this parameter
  is not set, the current ledger is used.
  * `key`: A series of Base64 encoded XDR strings specifying the `LedgerKey` to query.
  
## Response Format

`/getledgerentryraw`

JSON response (as a TypeScript interface):

```ts
interface Response {
  ledger: number; // uint32
  entries: {
    le: string; // base64-encoded xdr.LedgerEntry
  }[]
```

Note that this replaces the preexisting `getledgerentry` endpoint, which no longer exists.

## Notes

While there is no release yet with the new endpoint, stellar-core master currently has the endpoint and can be built from source. Note that the `getledgerentryraw` is disabled by default and must be enabled explicitly via the new `HTTP_QUERY_PORT` config option. See "Query Commands" section in stellar-core `docs/software/commands.md`.

Shaptic:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Adds support for a new endpoint: `/getledgerentryraw`.

### Why
Closes #5455.

Note that this endpoint will have the same request format (see #5426 or the PR itself for details) as a bunch of upcoming future endpoints, so for now there is a generic invocation method that can be shared. I reverted their implementation in https://github.com/stellar/go/pull/5428/commits/cd811da44b2ad52751b00a9621a8483626681575, so you can see that adding this commit back would be all we need to do to support them.

### Known limitations
Needs end-to-end testing, though basic unit tests are present.


clients/stellarcore/client.go
```diff
@@ -75,54 +75,12 @@ func (c *Client) Upgrade(ctx context.Context, version int) (err error) {
...
-		return
+	if hresp.StatusCode < 200 || hresp.StatusCode >= 300 {
+		return errors.New("http request failed with non-200 status code")
 	}
 	return nil
 }
 
-// GetLedgerEntry submits a request to the stellar core instance to get the latest
-// state of a given ledger entry.
-func (c *Client) GetLedgerEntry(ctx context.Context, ledgerKey xdr.LedgerKey) (proto.GetLedgerEntryResponse, error) {
```
sreuland, line 87: is this breaking change needed, could the underlying response of /getledgerentryraw be transformed internally here to retain same API? was the multi key request/response format originated from need as user feature?

Shaptic: Unfortunately, the new endpoint doesn't provide the `state = "live" | "dead"` field on the old response so we can't fill that in even if we wanted to (cc @SirTyson, maybe it's possible? maybe it's always `"live"`?).

The multi-key approach is a precedent for bulk-fetches of ledger entries for both simulation and state archival purposes, this is just the preface of getting a high-performance HTTP server in!

SirTyson: >  is it always "live" and implied "dead" if no entry is returned?).

Yeah, this is correct, all returned keys are "live". If a queried key isn't returned, it's "dead." A future release will have a `getledgerentry` endpoint that returns the "state" of a given key. This will cover both the live and dead cases like before, but also be expanded to include archival state information. 

Unfortunately I think this breaking change is necessary. Simply returning `live` or `dead` doesn't make sense in a state archival world. In order to return useful state archival information, the queries are significantly more expensive. This endpoint isn't really intended to be called by wallets or user facing applications, but for more intensive applications like stellar expert that need a simple key-value lookup of raw ledger entries (thus the "raw" in the name).

Shaptic: Yep, this all makes sense. 

> Simply returning live or dead doesn't make sense in a state archival world.

But aren't we only _really_ in that world in P23? So for this release we could feasibly maintain compatibility by doing the whole "line up ledger key existence in the response" thing to mimic a `state` field and break it harder in the next release.

I don't think that'd be super worth it, though: FWIW @sreuland, there were zero side effects to the monorepo by removing this method, so I wouldn't be surprised if this had very minimal impact. Also, it's good to inform users that state management will get more complicated very soon.

clients/stellarcore/client.go
```diff
@@ -332,3 +289,100 @@ func (c *Client) simpleGet(
...
 	return req, nil
 }
+
+// rawPost returns a new POST request to the connected stellar-core using the
+// provided path and the params values encoded as the request body to construct
+// the result.
+func (c *Client) rawPost(
+	ctx context.Context,
+	newPath string,
+	params string,
```
tamirms, line 299: nit: at first I thought `params` were query params and then after reading the code I saw that it's actually the body of the post request. I suggest renaming `params` to `body`

clients/stellarcore/client.go
```diff
@@ -332,3 +289,100 @@ func (c *Client) simpleGet(
...
+	target interface{},
+	endpoint string,
+	ledgerSeq uint32,
+	keys ...xdr.LedgerKey,
+) error {
+	q, err := buildMultiKeyRequest(keys...)
+	if err != nil {
+		return err
+	} else if ledgerSeq >= 2 { // optional param
+		q += fmt.Sprintf("ledgerSeq=%d", ledgerSeq)
```
tamirms, line 336: I thought `buildMultiKeyRequest()` strips the trailing `&` which means that there is no `&` separating the `ledgerSeq` param from the last ledger key query param

clients/stellarcore/client.go
```diff
@@ -332,3 +289,100 @@ func (c *Client) simpleGet(
...
+	ctx context.Context,
+	target interface{},
+	endpoint string,
+	ledgerSeq uint32,
+	keys ...xdr.LedgerKey,
+) error {
+	q, err := buildMultiKeyRequest(keys...)
+	if err != nil {
+		return err
+	} else if ledgerSeq >= 2 { // optional param
```
tamirms, line 335: nit: no need for else if since we return in the if statement above
```suggestion
	if ledgerSeq >= 2 { // optional param
```

clients/stellarcore/client.go
```diff
@@ -332,3 +289,100 @@ func (c *Client) simpleGet(
...
+	//     SELECT AVG(LENGTH(HEX(key))) / 2 FROM ledger_entries;
+	//
+	// on a pubnet RPC instance is ~57.6. We can use this to preallocate a
+	// string buffer for performance.
+	//
+	// We know that these endpoints will almost exclusively be used for
+	// ContractData and the like, so we could optimize the buffer further for
+	// that, but that data is harder to query since it'd involve parsing the XDR
+	// from the DB to check the key type.
+	q := strings.Builder{}
```
tamirms, line 375: I suspect that using `strings.Join()` will be just as fast or faster than concatenating the string manually:

https://stackoverflow.com/questions/60688270/performance-comparison-for-concatenation-of-strings

clients/stellarcore/client.go
```diff
@@ -332,3 +289,100 @@ func (c *Client) simpleGet(
...
+
+// makeLedgerKeyRequest is a generic method to perform a request in the form
+// `key=...&key=...&ledgerSeq=...` which is useful because several Stellar Core
+// endpoints all use this request format. Be sure to pass `target` by reference.
+func (c *Client) makeLedgerKeyRequest(
+	ctx context.Context,
+	target interface{},
+	endpoint string,
+	ledgerSeq uint32,
+	keys ...xdr.LedgerKey,
```
tamirms, line 330: should we assert that keys is non-empty?

protocols/stellarcore/getledgerentryraw_response.go
```diff
@@ -0,0 +1,11 @@
+package stellarcore
+
+// GetLedgerEntryRawResponse is the structure of Stellar Core's /getledgerentryraw
+type GetLedgerEntryRawResponse struct {
+	Ledger  uint32                   `json:"ledger"`
+	Entries []RawLedgerEntryResponse `json:"entries"`
```
tamirms, line 6: out of curiosity why is each entry a json blob containing one string field instead of simply being a string?

clients/stellarcore/client.go
sreuland: got it, thanks for clarification, wasn't aware of the live/dead significance either, thanks!

protocols/stellarcore/getledgerentryraw_response.go
Shaptic: Not my schema, heh :shrug: @SirTyson?

clients/stellarcore/client.go
Shaptic: good idea! done in fe204eb2

clients/stellarcore/client.go
Shaptic: wow great catch, fixed in fe204eb2

clients/stellarcore/client.go
Shaptic: much cleaner! done in fe204eb2

clients/stellarcore/client.go
Shaptic: yeah, good idea :+1: fe204eb2

clients/stellarcore/client_test.go
```diff
@@ -75,3 +78,39 @@ func TestManualClose_NotAvailable(t *testing.T) {
...
+				Entry: "pretend this is XDR lol",
+			},
+			{
+				Entry: "pretend this is another XDR lol",
+			},
+		},
+	}
+
+	// happy path - fetch an entry
+	hmock.On("POST", "http://localhost:11626/getledgerentryraw").
```
tamirms, line 100: can you add a check that the request body is correct? the motivation behind having such an assertion is to prevent regressions where, for example, there is no & precending the ledgerSeq param

Shaptic: I wanted to (see the commit message of 7c8a69335647bbc621051e4edfbde4d1b7e07c0f), but I couldn't figure out how to do that with our mock [expectations library](https://github.com/stellar/go/blob/6fbba33f8123b69f398ac3cd7166ebac9311fde1/support/http/httptest/main.go#L41). After digging into it more, I just pushed 4da6f4ce which does it!

tamirms (approved):

PR #5431, opened 2024-08-19
ingest/ledgerbackend: add trusted hash to captive core catchup

### LINKED ISSUE #4538: Horizon: missing information passed to captive-core when configured to run "on disk"

v19.02.19.0

## What did you do?

We had a testnet outage that revealed that captive-core reconstructed a ledger and replayed transactions from an untrusted archive.

We should always verify data before replaying. In this case we should always pass down proofs thatt were acquired from the network.

### What did you see instead?

captive-core replayed transactions from a bad archive, horizon ingested this data and corrupted itself. This required clearing core and horizon.

### Additional information

the root cause of the issue is that:
* when running using the `--in-memory` flag, core uses `--start-at-ledger` and `--start-at-hash` to bootstrap trust (ledger information comes from the latest ledger that Horizon ingested). This works as expected
* when running "on disk", Horizon invokes core in two steps. The first one is the problematic one: when running `catchup` (offline catchup), core needs to "anchor" trust somewhere. The way to do it is by storing trusted hashes in a file and passing it down to core using the `--trusted-checkpoint-hashes` option. Note that unlike `--start-at-ledger` the hashes that can be passed to core this way must be "checkpoint ledgers" *and* catchup has to be invoked to reconstruct that checkpoint (ie: the "to" ledger must be a checkpoint ledger).

Related core issue https://github.com/stellar/stellar-core/issues/3503 to surface this better in the future as it's a footgun for sure.

sreuland:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

The Captive Core backend now performs 'online' stellar-core `run` for bounded modes of tx-meta retrieval, which will be used for Horizon's `db reingest range` and `ingest verify-range` commands. 

Using `run` enables core to build, validate, and emit trusted ledger hashes into the tx-meta stream for the requested ledger range. The bounded range commands will no longer do the 'offline' mode of running core `catchup` for generating tx-meta from just history archives, which does not guarantee verification of the ledger hashes to that of live network as mentioned in ([#4538](https://github.com/stellar/go/pull/4538)).

Due to the usage of `run` with LCL set to the `from` , there is now potential for longer run time `reingest` and `verify-range` execution durations due to core having to perform online replay upfront from network latest ledger back to `from`. The longer runtime duration will be proportional to the older age of the `from` ledger. 

### Why

Prevents consuming tx-meta from a potentially tampered history archives server. 

Closes #4538 

### Known limitations

The execution duration will be longer for bounded ranges with relatively old 'from' ledger sequence compared to latest network age(ledger sequence). 

Evaluated other paths for obtaining trusted hashes, by enhancing the usage of existing `catchup` first, what was discovered is that command usage with trusted hash input flags leads to complex scenarios for obtaining the hash(es) in deployments and out-of-band from the horizon command(`reingest`, `verify-state`) invocation:
* `--trusted-checkpoint-hashes` - to use this with `catchup`, need to first run `verify-checkpoints` from core to build the file, this will take just as long as `run` with LCL=1 at first time. After that, need to find a persistent disk storage path to keep the file, which contradicts the `reingest` cli expectations, which allows (re)running from any host/path, and always creates a new storage path for parallel purposes, and avoiding conflicts in general with multiple captive instances on same host. This doesn't scale for the range of uses cases that `reingest` can be used and would impinge on it. 

* `--trusted-hash` - to use this with `catchup`, requires first getting the live network side trusted hash for the intended 'to' ledger first, this can only be done via `run` with LCL={to-1} and ingesting the 'to' ledger and then stopping core. 
  * The horizon db could have 'to' ledger previously ingested from prior, but that probably shouldn't be considered a trusted source as the db table is no different than a HA server. 
  * This approach is similar to that of current proposed in pr, the difference being the PR uses `run` with  LCL={from-1} and collects the meta for range from the online 'internal catchup/replay' mode that core runs regardless, whereas this would generate the meta for range from offline `catchup` command.


ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -327,6 +327,49 @@ func (c *CaptiveStellarCore) getLatestCheckpointSequence() (uint32, error) {
 	return has.CurrentLedger, nil
 }
 
+// Obtain hash for a ledger sequence from a live connection with network.
+// Will NOT use history archives for hash retrieval.
+// Returns the requested ledger sequence or the closest one to it that could be replayed
+// and its associated trusted hash.
+func (c *CaptiveStellarCore) getTrustedHashForLedger(sequence uint32) (uint32, string, error) {
```
sreuland, line 334: internal/integration/db_test.go already had several test cases that will exercise this code by running the known uses cases for `db reingest`, no additional test coverage seemed worthwhile.

services/horizon/internal/test/integration/integration.go
```diff
@@ -566,6 +567,13 @@ func (i *Test) waitForCore() {
 			if durationSince := time.Since(infoTime); durationSince < coreStartupPingInterval {
 				time.Sleep(coreStartupPingInterval - durationSince)
 			}
+			cmdline := []string{"-f", integrationYaml, "logs", "core"}
```
sreuland, line 570: added this for visibility on core startup during test execution, as the docker compose container output is detached, this helped identify a core dump that was happening in the container as the horizon test logs just repeat 'cannot connect to localhost:11626'

ingest/ledgerbackend/captive_core_backend.go
```diff
@@ -388,9 +340,15 @@ func (c *CaptiveStellarCore) openOnlineReplaySubprocess(ctx context.Context, fro
...
 	c.nextLedger = 0
-	ran := UnboundedRange(from)
+	ran := ledgerRange
+	var last *uint32
+	if ledgerRange.bounded {
+		boundedTo := ledgerRange.to
+		last = &boundedTo
+	}
+
+	c.lastLedger = last
```
sreuland, line 350: this is the key to being able to (re)use live `runFrom` with bounded range, trigger the captive core to be stopped once it emits ledger on pipe that matched up to the bounded `to` if exists, which would have been same as core `catchup` process terminating at point of same `to` ledger emitted.

.github/workflows/horizon.yml
```diff
@@ -134,8 +138,8 @@ jobs:
       - name: Build and test the Verify Range Docker image
         run: |
           docker build --build-arg="GO_VERSION=$(sed -En 's/^toolchain[[:space:]]+go([[:digit:].]+)$/\1/p' go.mod)" -f services/horizon/docker/verify-range/Dockerfile -t stellar/horizon-verify-range services/horizon/docker/verify-range/
-          # Any range should do for basic testing, this range was chosen pretty early in history so that it only takes a few mins to run
-          docker run -e BRANCH=$(git rev-parse HEAD) -e FROM=10000063 -e TO=10000127 stellar/horizon-verify-range
+          # Use small default range of two most recent checkpoints back from latest archived checkpoint. 
+          docker run -e TESTNET=true -e BRANCH=$(git rev-parse HEAD) -e FROM=0 -e TO=0 stellar/horizon-verify-range
```
sreuland, line 142: using latest checkpoint on testnet for verify-range, tried similar on pubnet, takes 45+ minutes, mostly on change processor ledger entry processing.

tamirms:
@sreuland could you provide some data on how much slower `db reingest range` now that we use `stellar-core run` instead of `stellar-core catchup`?

sreuland:
> @sreuland could you provide some data on how much slower `db reingest range` now that we use `stellar-core run` instead of `stellar-core catchup`?

On mac/m1, with core v21.3.1, pubnet network:
| Range              |                this pr    |           horizon v2.32.0 | notes             |
| ---------------| ----------------- | --------------------- |--------------|
| 53136575-53136639| 53 minutes |  27 minutes  |     recent 64 ledger range   |

i'm running older/larger ranges, will post results once complete.

tamirms:
@sreuland I think we need to take a step back and consider alternative implementations.

The 26 minute overhead on a recent 64 ledger range is pretty significant. Assuming the overhead increases linearly as you go further back in time, it seems likely that this issue becomes worse when reingesting older ledger ranges. The overhead of using `stellar-core run` instead of `stellar-core catchup` is also compounded when doing parallel reingestion with several workers on a large ledger range.

You mentioned `--trusted-checkpoint-hashes` as one possible alternative implementation. Could you measure how long it takes to generate the trusted hashes file by running `stellar-core verify-checkpoints` ? Generating the file once and then using it during parallel reingestion of large ledger ranges seems appealing because the overhead to generate the file is just a one time cost compared to the overhead of using `stellar-core run` on every [task](https://github.com/stellar/go/blob/f10ea0f79143c1b5594f5a91c5aeb4a8ae39f2cc/services/horizon/internal/ingest/parallel.go#L64-L72) in the parallel reingestion task queue.

Another idea to consider is that we can verify the ledger chain at the end of a running reingestion task. After ingesting a bounded range with captive core we can check the ledgers at the boundaries of that range in the horizon db.

For example if we reingest the bounded range `[start, end]` we would query the horizon db for ledger `end+1` and, if it exists, we check that previous ledger hash of ledger `end+1` matches the ledger hash of `end`. Similarly, we can query the horizon db for ledger `start-1` and, if it exists, we check that the previous ledger hash of ledger `start` matches the ledger hash of `start-1`. If there is a mismatch in ledger hashes we can exit with an error.

Also, when ingesting an unbounded range with captive core we would need check that the start ledger's previous ledger hash matches whatever we have in the horizon db. Howevr, one of the downsides of this approach is that if there is a ledger mismatch we only find out at the very end.

It would be helpful to have a design document which outlines different possible implementations with their pros / cons to make it easier to analyze which is the best solution.

sreuland:
> It would be helpful to have a design document which outlines different possible implementations with their pros / cons to make it easier to analyze which is the best solution.

yes, good suggestion, I've started [design doc](https://docs.google.com/document/d/1891A0NQZ13vgCiRH51UIhken4my4xfHTthrGRLTYXDA/edit#heading=h.pwg50z56u42u) to continue discussion on options.

sreuland:
> Also, when ingesting an unbounded range with captive core we would need check that the start ledger's previous ledger hash matches whatever we have in the horizon db. Howevr, one of the downsides of this approach is that if there is a ledger mismatch we only find out at the very end.

Ah, good call out, I think this could be a separate scope/ticket, to essentially add current network validation via ledger hash to horizon's db as part of forward 'live' ingestion mode? This ticket was focused on hash validation specific to reingestion of bounded offline.

sreuland:
closing this pr as obsolete now in favor of a [different design approach mentioned for repairing skip list usage](https://github.com/stellar/go/issues/4538#issuecomment-2387008117) in captive core binary instead, once this repair propagates, it will enable existing 'catchup' to emit anchored(to current network) trusted hashes within ledger metadata.

PR #5442, opened 2024-08-21
ingest: Add verbosity to missing file error for `BufferedStorageBackend`

Shaptic:
### What
Specify the object key when requesting a particular ledger batch.

### Why
This makes it easier to debug your configuration because you can see the exact object and compare it against your actual data source more easily.

### Known limitations
n/a


ingest/ledgerbackend/ledger_buffer.go
```diff
@@ -167,7 +167,7 @@ func (lb *ledgerBuffer) downloadLedgerObject(ctx context.Context, sequence uint3
 
 	reader, err := lb.dataStore.GetFile(ctx, objectKey)
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrapf(err, "failed finding file: %s", objectKey)
```
urvisavla, line 170: "unable to retrieve file" or "file not found"?

urvisavla (approved):

ingest/ledgerbackend/ledger_buffer.go
Shaptic: Oh I phrased it this way as a parallel to "failed reading file" below :laughing: but it sounds awkward. Fixed!

PR #5454, opened 2024-09-03
Send an error response for Horizon.Problem

aditya1702:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Change the decoding method for non-2xx codes in `async-txsub` endpoint. Current code decodes all error responses to `horizon.Problem` object. This change adds a separate decoding method specific for the async-txsub endpoint responses.

### Why

For non-2xx codes, the `async-txsub` endpoint not only sends a `horizon.Problem` response, but also sends `AsyncTransactionSubmissionResponse` object depending on certain core responses.

### Known limitations

NA

### DIFF
--- a/clients/horizonclient/internal.go
+++ b/clients/horizonclient/internal.go
@@ -70,10 +70,7 @@ func decodeAsyncTxSubResponse(resp *http.Response, object interface{}) error {
 	// the hash of the transaction. If the response was not a valid AsyncTransactionSubmissionResponse object,
 	// the hash of the converted object will be empty.
 	asyncRespDecoder := json.NewDecoder(bytes.NewReader(bodyBytes))
-	err = asyncRespDecoder.Decode(&object)
-	if err != nil {
-		return errors.Wrap(err, "error decoding response")
-	}
+	asyncRespDecoder.Decode(&object)
 	if asyncResp, ok := object.(*horizon.AsyncTransactionSubmissionResponse); ok && asyncResp.Hash != "" {
 		return nil
 	}
--- a/clients/horizonclient/internal.go
+++ b/clients/horizonclient/internal.go
@@ -67,7 +67,8 @@ func decodeAsyncTxSubResponse(resp *http.Response, object interface{}) error {
 	}
 
 	// The first decoder converts the response to AsyncTransactionSubmissionResponse and checks
-	// the hash of the transaction. If the response was not a valid one, the hash will be empty.
+	// the hash of the transaction. If the response was not a valid AsyncTransactionSubmissionResponse object,
+	// the hash of the converted object will be empty.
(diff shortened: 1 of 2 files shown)


clients/horizonclient/internal.go
```diff
@@ -47,6 +51,45 @@ func decodeResponse(resp *http.Response, object interface{}, horizonUrl string,
...
+	}
+
+	// The first decoder converts the response to AsyncTransactionSubmissionResponse and checks
+	// the hash of the transaction. If the response was not a valid AsyncTransactionSubmissionResponse object,
+	// the hash of the converted object will be empty.
+	asyncRespDecoder := json.NewDecoder(bytes.NewReader(bodyBytes))
+	err = asyncRespDecoder.Decode(&object)
+	if err != nil {
+		return errors.Wrap(err, "error decoding response")
+	}
```
tamirms, line 76: if the response doesn't contain a AsyncTransactionSubmissionResponse and instead contains a problem, will there be an error returned by `asyncRespDecoder.Decode(&object)`?

aditya1702: @tamirms No, it will still decode without any errors but all the fields of the `AsyncTransactionSubmissionResponse` object will be empty. Hence the condition `asyncResp.Hash != ""` through which I check which object is it actually.

tamirms: @aditya1702 I think this solution is a bit brittle. I'm not sure if the assumption that we can decode the response into a `AsyncTransactionSubmissionResponse` without errors will always hold in the future. I think it would be better to first attempt to decode the response into `AsyncTransactionSubmissionResponse` and if there are no decode errors and the `AsyncTransactionSubmissionResponse` is valid you can return it. Otherwise, you can attempt to decode the response into a problem

clients/horizonclient/internal.go
```diff
@@ -47,6 +51,42 @@ func decodeResponse(resp *http.Response, object interface{}, horizonUrl string,
...
+	// the hash of the transaction. If the response was not a valid AsyncTransactionSubmissionResponse object,
+	// the hash of the converted object will be empty.
+	asyncRespDecoder := json.NewDecoder(bytes.NewReader(bodyBytes))
+	err = asyncRespDecoder.Decode(&object)
+	if asyncResp, ok := object.(*horizon.AsyncTransactionSubmissionResponse); err == nil && ok && asyncResp.Hash != "" {
+		return nil
+	}
+
+	// Create a new reader for the second decoding. The second decoder decodes to Horizon.Problem object.
+	errorDecoder := json.NewDecoder(bytes.NewReader(bodyBytes))
```
tamirms, line 79: nit: I would rename to problemDecoder

tamirms (approved):

Issue #5412, opened 2024-08-06
ingest/pipeline: Create functional producer for BufferedStorageBackend


sreuland:
### What problem does your feature solve?
BufferedStorageBackend provides ledger close meta(LCM) individually per `GetLedger`, but there is no more efficient way to participate as a streaming producer of LCM's.

### What would you like to see?
Follow [design proposal on Functional Processors Lib](https://docs.google.com/document/d/17wy1r5l5cUWxAj5Qty_faXni4tsXER4pAdvTo05nnO0/edit#heading=h.3dy6vkm)
Provide a ‘Producer’ function for BufferedStorage Backend. 
The function will be used as ‘producer’ operator in a pipeline, emitting tx-meta LCM over a callback fn and acts as a closures to encapsulate private instance of BufferedStorageBackend and avoid any unintended side effects.

```
// the returned channel will either return one error or it will be closed when the publishing is finished for bounded range
// if it's unbounded, then chan would never be closed.

PublishFromBufferedStorageBackend(ledgerRange ledgerbackend.Range,
                                            bufferedConfig ledgerbackend.BufferedStorageBackendConfig,
                                            dataStoreConfig datastore.DataStoreConfig 
                                            ctx context.Context, 
                                            callback func (xdr.LedgerCloseMeta) error) chan error
    
}
```

The method will return immediate, creating an async worker routine in background to continue processing.
* it uses whatever prepared ledger range was configured for the backend.
* it goes into a loop of invoking internal LedgerBacked.GetLedger to iterate each LedgerCloseMeta retrieved for the given prepared ledger range. 
* It invokes the callback func per each LedgerCloseMeta. 
* If the range was unbounded then it runs infinite until ctx is canceled.

Visualization of where the producer function fits in the larger CDP design for data transformation pipeline:

![dual_func_general](https://github.com/user-attachments/assets/86d118c7-01cc-419b-b88f-df64e58cfc05)

Relates to:
* #5451 
* https://github.com/stellar/stellar-docs/issues/889
* #5413 
* #5414 

### What alternatives are there?
new streaming ingestion app use cases would have to implement the same locally.

tamirms:
I think there are some subtleties to this interface change that we should consider:

- If captive-core terminates unexpectedly this interface does not allow us to communicate that error to the caller 
- What happens if you call `Publish()` and while that go routine is running you call `GetLedger()` concurrently?
- What happens if you call `Publish()`multiple times before the go routine finishes running?
[...]

PR #5462, opened 2024-09-17
ingest/ledgerbackend: Create functional producer for BufferedStorageBackend

sreuland:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Added new `PublishFromBufferedStorageBackend` to provide a functional origin for streaming data development as part of [CDP processors library](https://github.com/stellar/go/labels/cdp-processors-lib).

### Why

Provide efficient tooling in the sdk to produce a stream of tx-meta from a remote Datastore that has pre-computed tx-meta files from Galexie. The stream is established as a callback function which caller passes, it will be asynchronously invoked and serves as origin in a data transformation pipeline, providing source of network tx-meta to which pipeline transform operators can derive from.

Closes: #5412


ingest/ledgerbackend/buffered_storage_backend.go
```diff
@@ -64,6 +105,121 @@ func NewBufferedStorageBackend(config BufferedStorageBackendConfig, dataStore da
...
+				"duration": time.Since(startTime).Seconds(),
+			}).Info("Ledger returned from the backend")
+
+			err = callback(ledgerCloseMeta)
+			if err != nil {
+				resultCh <- errors.Wrap(err, "received an error from callback invocation")
+				return
+			}
+		}
+		close(resultCh)
```
urvisavla, line 217: Why isn't the channel closed on early return due to an error?

ingest/ledgerbackend/buffered_storage_backend.go
```diff
@@ -64,6 +105,121 @@ func NewBufferedStorageBackend(config BufferedStorageBackendConfig, dataStore da
...
+	ctx context.Context,
+	callback func(xdr.LedgerCloseMeta) error) chan error {
+
+	logger := publisherConfig.Log
+	if logger == nil {
+		logger = log.DefaultLogger
+	}
+	resultCh := make(chan error, 1)
+
+	go func() {
```
urvisavla, line 158: Should the context cancellation be handled in this goroutine to handle caller cancelling the context?

ingest/ledgerbackend/buffered_storage_backend.go
```diff
@@ -64,6 +105,121 @@ func NewBufferedStorageBackend(config BufferedStorageBackendConfig, dataStore da
...
+			}
+
+			log.WithFields(log.F{
+				"sequence": ledgerSeq,
+				"duration": time.Since(startTime).Seconds(),
+			}).Info("Ledger returned from the backend")
+
+			err = callback(ledgerCloseMeta)
+			if err != nil {
+				resultCh <- errors.Wrap(err, "received an error from callback invocation")
```
urvisavla, line 213: nit: should we consistently use fmt.Errorf ?

ingest/ledgerbackend/buffered_storage_backend_test.go
```diff
@@ -302,6 +303,286 @@ func TestBSBPrepareRange(t *testing.T) {
...
+	},
+		time.Second*3,
+		time.Millisecond*50)
+
+	resultCh = PublishFromBufferedStorageBackend(Range{from: uint32(2), to: uint32(3), bounded: false}, pubConfig, ctx, appCallback)
+	assert.Eventually(t, func() bool {
+		select {
+		case chErr, ok := <-resultCh:
+			if ok {
+				assert.ErrorContains(t, chErr, "invalid end value for unbounded ranged, must be zero")
```
urvisavla, line 470:
```suggestion
				assert.ErrorContains(t, chErr, "invalid end value for unbounded range, must be zero")
```

ingest/ledgerbackend/buffered_storage_backend.go
```diff
@@ -64,6 +105,121 @@ func NewBufferedStorageBackend(config BufferedStorageBackendConfig, dataStore da
...
+			ledgerBackend = WithMetrics(ledgerBackend, publisherConfig.Registry, publisherConfig.RegistryNamespace)
+		}
+
+		if ledgerRange.bounded && ledgerRange.to <= ledgerRange.from {
+			resultCh <- errors.New("invalid end value for bounded range, must be greater than start")
+			return
+		}
+
+		if !ledgerRange.bounded && ledgerRange.to > 0 {
+			resultCh <- errors.New("invalid end value for unbounded ranged, must be zero")
```
urvisavla, line 182:
```suggestion
			resultCh <- errors.New("invalid end value for unbounded range, must be zero")
```

ingest/ledgerbackend/buffered_storage_backend.go
tamirms, line 158: It seems to be a best practice to prefer asynchronous APIs in Go, see:

https://google.github.io/styleguide/go/decisions.html#synchronous-functions
https://stackoverflow.com/a/28510173
page 9 of https://drive.google.com/file/d/1nPdvhB0PutEJzdCq5ms6UI58dp50fcAN/view

sreuland: I don't think so, the loop below calls GetLedger(ctx) which detects ctx cancellation and throws error to allow this fn to wind down, propogating the err onto the returned channel

ingest/ledgerbackend/buffered_storage_backend.go
sreuland: nice catch, missed that, had doc'd the behavior as always closing this channel even when an err is sent.

ingest/ledgerbackend/buffered_storage_backend.go
```diff
@@ -64,6 +105,121 @@ func NewBufferedStorageBackend(config BufferedStorageBackendConfig, dataStore da
...
+			return
+		}
+
+		from := ordered.Max(2, ledgerRange.from)
+		to := ledgerRange.to
+		if !ledgerRange.bounded {
+			to = math.MaxUint32
+		}
```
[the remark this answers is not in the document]
sreuland: less code, thanks!

ingest/ledgerbackend/buffered_storage_backend.go
sreuland: yes, I waffled on the pattern, was reaching for what I thought would be easiest client usage, but, good references with justifications for sync behavior for clients to have control. This fn is actually doing a blend of Callback API(page 9)  with Future' API(page 10). In any case, definitely on board to change the signature.

sreuland: done - https://github.com/stellar/go/pull/5462/commits/8c3f694b3eda98f4116eb17dbbf90b8c686f9ad2

sreuland: good call out, It made me realize needed to add test coverage to assert this outcome does happen - [TestBSBProducerCallerCancelsCtx](https://github.com/stellar/go/pull/5462/commits/f1e9d27fe4f7f3b55d1b96a61067f4b226a37248)

ingest/cdp/producer.go
```diff
@@ -0,0 +1,147 @@
+package cdp
```
urvisavla, line 1: I think this is the first time mention of CDP in our repo. Do you think we should add a brief readme or changelog for the package? Alternatively, we could move this to ledgerbackend package. No strong preference either way.

sreuland: yes, I added a second changelog for intro of new cdp package - https://github.com/stellar/go/pull/5462/commits/d25a506ea0b31072c5f871604252ab08fd5bf76e

urvisavla (approved):
lgtm!

ingest/cdp/producer.go
```diff
@@ -0,0 +1,147 @@
...
+// ctx - the context. Caller uses this to cancel the internal ledger processing,
+// when canceled, the function will return asap with that error.
+//
+// callback - function. Invoked for every LedgerCloseMeta. If callback invocation
+// returns an error, the processing will stop and return an error asap.
+//
+// return - error, function only returns if requested range is bounded or an error occured.
+// nil will be returned only if bounded range requested and completed processing with no errors.
+// otherwise return will always be an error.
+func PublishFromBufferedStorageBackend(ledgerRange ledgerbackend.Range,
```
tamirms, line 89: I'm not sure if `PublishFromBufferedStorageBackend` is the best name for this function. Here are some suggestions of other names for you to consider:

IngestFromBufferedStorageBackend
ApplyLedgersFromBufferedStorageBackend
MapLedgersFromBufferedStorageBackend

tamirms: I also wonder if this function should be moved to the ledgerbackend package since I associate cdp with extracting meaningful information out of the ledger close meta xdr objects

sreuland: I was trying to convey the intended usage of this as a 'source' operator in streams processing(data pipeline), it's an origin of data.

How about `LedgerMetadataSource`? 

The name also incorporates new colloquialism for tx-meta and avoids any direct reference to ledger backend's as that aspect is purely internal. I think with this terminology it aligns better as a widget in the `cdp` package, used by developers as an operator in their ingestion stream pipelines.

tamirms: I think the ledger backend is actually the source of the ledgers. This function is similar to the concept of [map()](https://en.wikipedia.org/wiki/Map_(higher-order_function))in functional programming because the function is simply applying a callback on all the ledgers emitted by the ledger backend. That's why I thought including "ingest", "apply", or "map" in the name of the function could be a good descriptor.

sreuland: ok, how about `ApplyLedgerMetadata` ? this fn represents applying an app callback to a source or origin of ledger metadata, the type of backend being used internally doesn't need to be reflected in name.

tamirms (approved):

PR #5465, opened 2024-09-18
services/horizon/internal/db2/history: Include lower bound on descending history queries

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Include a lower bound on queries for the history endpoints which fetch rows in descending order. The following endpoints have been modified to include the lower bound:

/transactions
/ledgers
/effects
/operations
/trades

### Why

In the case where a lot of history has been reaped when running a horizon instance with limited history, there will be a large amount of rows which will be deleted from the history tables. These rows will still be present in Postgres until they get vacuumed at which point they will be removed from the tables and indexes.

So, for example, if you query for transactions in sorted order, postgres will be need to sift through and discard rows which have been deleted. This can result in the query timing out if there is a large amount of deleted rows in the search. We can avoid having to sift through deleted rows by including a lower bound beyond which there should be no deleted rows. We can use the oldest ledger in the horizon db as the lower bound.

### Known limitations

[N/A]


services/horizon/internal/actions/effects.go
```diff
@@ -94,20 +94,20 @@ func (handler GetEffectsHandler) GetResourcePage(w HeaderWriter, r *http.Request
...
-func loadEffectRecords(ctx context.Context, hq *history.Q, qp EffectsQuery, pq db2.PageQuery) ([]history.Effect, error) {
+func loadEffectRecords(ctx context.Context, hq *history.Q, qp EffectsQuery, pq db2.PageQuery, oldestLedger int32) ([]history.Effect, error) {
 	switch {
 	case qp.AccountID != "":
-		return hq.EffectsForAccount(ctx, qp.AccountID, pq)
+		return hq.EffectsForAccount(ctx, qp.AccountID, pq, oldestLedger)
 	case qp.LiquidityPoolID != "":
-		return hq.EffectsForLiquidityPool(ctx, qp.LiquidityPoolID, pq)
+		return hq.EffectsForLiquidityPool(ctx, qp.LiquidityPoolID, pq, oldestLedger)
 	case qp.OperationID > 0:
```
urvisavla, line 103: Just want to confirm that these EffectsForOperation, EffectsForLedger and EffectsForTransaction don’t require a lower bound because the queries are for a specific ledger number, correct?

tamirms: yes, when you specify a specific ledger, operation, or transaction that restricts the query to a tighter bound than the entire history retention in the horizon db

karthikiyer56:
As a part of this cleanup/deprecation task, I think the fields in Db structure `ExpAssetStat` also need to be removed.
I think that means the 2 rows in `expat_asset_stat` table need to be dropped.
can someone confirm if that is true?

urvisavla (approved):
Looks good!

PR #5466, opened 2024-09-19
updates to horizon docker builds

mwtzzz:
### What
- a new Dockerfile to install Core from the "testing" pool an update to the Makefile specifically for this case
- a new Dockerfile to install Core and Horizon from their "stable" pools
- additional comments to clarify what the different files do

### Why
this is needed to support Horizon's new workflow

### Testing
this has been tested successfully

### Issue addressed by this PR
https://github.com/stellar/ops/issues/3124

### DIFF
--- a/services/horizon/docker/Dockerfile.core-testing
+++ b/services/horizon/docker/Dockerfile.core-testing
@@ -1,0 +1,19 @@
+# install both horizon and core from apt "testing" pool
+FROM ubuntu:focal
+
+ARG VERSION
+ARG STELLAR_CORE_VERSION
+ARG DEBIAN_FRONTEND=noninteractive
+ARG ALLOW_CORE_UNSTABLE=no
+
+RUN apt-get update && apt-get install -y wget apt-transport-https gnupg2 && \
+    wget -qO /etc/apt/trusted.gpg.d/SDF.asc https://apt.stellar.org/SDF.asc && \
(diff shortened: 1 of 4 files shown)


services/horizon/docker/Dockerfile.core-testing
```diff
@@ -0,0 +1,19 @@
+# install both horizon and core from apt "testing" pool
+FROM ubuntu:focal
```
jacekn, line 2: I wonder if we can use the same method as for the `ALLOW_CORE_UNSTABLE` to avoid 2nd Dockerfile. If we pass version explicitly that may work?

mwtzzz: I already thought that and decided against one Dockerfile because it will be much easier to read and know what's going on by splitting if off to a 2nd one.

jacekn: The reason I pointed it out is because after the change we'd have one dockerfile for the testing repo and one for stable + unstable. This feel inconsistent to me.
I think two best options would be to:
1. Use one Dockerfile per repo
2. Use one Dockerfile for all repos

mwtzzz: I can split unstable/stable into two. So let's keep this PR, if you can approve it, and I will do that split in a separate PR because that's unrelated to the workflow addressed in this one.

jacekn (approved):

Issue #5438, opened 2024-08-20
protocols/horizon: Remove deprecated API fields


Shaptic:
**Epic**: #5433 

-----------------------

There are several API fields that have been deprecated for a very long time. We should remove them as part of the Protocol 22 release in which we will change to the protocol-specific versioning scheme (i.e. `horizon-v22.0.0`).

The following API fields are deprecated:

* In `/assets`, the `num_accounts` and `amount` fields:

https://github.com/stellar/go/blob/dfabe31b5b89cb235e81732d708a3f5bdeff507a/protocols/horizon/main.go#L180-L187

* In `/transactions/:id` (or any paginating variant), the `valid_before` and `valid_after` fields, because users should instead check the transaction's preconditions which can have way more than time bounds:

https://github.com/stellar/go/blob/dfabe31b5b89cb235e81732d708a3f5bdeff507a/protocols/horizon/main.go#L536-L538

sreuland:
should this effort include a coordinated outreach/blast to sdk's for notice? a separate ticket for cleanup of the same in js-sdk?

PR #5478, opened 2024-09-27
Karthik/5438/deprecate horizon fields for protocol 22

karthikiyer56:
<details>
  <summary>Remove deprecated fields from horizon in protocol 22 release as a part of issue 5438 (https://github.com/stellar/go/issues/5438)</summary>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
* This PR removes 4 fields that have been deprecated, and were waiting for the next protocol release to be rolled out.
They are `numAccounts`, `amount`, `validBefore`, `validAfter` from the `AssetStat` structure in services/horizon/protocols/main.go
* Additionally, the `num_accounts` and `amount` columns are also dropped from the `exp_asset_stat` tables and the DB bindings and code are removed

The PR that updates/removes the references for these deprecated fields from the stellar-docs is: https://github.com/stellar/stellar-docs/pull/1014


karthikiyer56:
As a part of this cleanup/deprecation task, I think the fields in Db structure ExpAssetStat also need to be removed.
I think that means the 2 rows in expat_asset_stat table need to be dropped.
can someone confirm if that is true?

services/ticker/internal/scraper/asset_scraper.go
```diff
@@ -143,14 +145,28 @@ func isDomainVerified(orgURL string, tomlURL string, hasCurrency bool) bool {
 	}
 	return true
 }
+func AddStringsToSum(values ...string) (float64, error) {
```
karthikiyer56, line 148: I was looking for a better place to put these helper functions.
can someone point out where that might be?
If there already exists such a varadic function for add, Ima delete this.

services/ticker/internal/scraper/asset_scraper.go
```diff
@@ -143,14 +145,28 @@ func isDomainVerified(orgURL string, tomlURL string, hasCurrency bool) bool {
...
+}
 
 // makeTomlAsset aggregates Horizon Data with TOML Data
 func makeFinalAsset(
 	asset hProtocol.AssetStat,
 	issuer TOMLIssuer,
 	errors []error,
 ) (t FinalAsset, err error) {
-	amount, err := strconv.ParseFloat(asset.Amount, 64)
+	amount, err := AddStringsToSum(asset.Balances.Authorized + asset.Balances.Unauthorized + asset.Balances.Unauthorized)
```
karthikiyer56, line 169: The code in this ticker service is a moot point, since this entire service will be pulled out soon form the monorepo and moved to the archives repo.

At any rate, I am making this change for self-consistency purposes.

karthikiyer56, line 169: Is this `if` condition correct?
Does this need to be simply `if asset.Balances.authorized`, instead of adding all 3?

.gitignore
```diff
@@ -5,7 +5,7 @@
 /services/horizon/captive-core
 /services/horizon/horizon
 /services/horizon/stellar-horizon
-/bucket-cache
+**bucket-cache/
```
tamirms, line 8: did you intend to include this change in this PR?

services/ticker/internal/scraper/asset_scraper.go
```diff
@@ -23,23 +23,25 @@ import (
...
 		return true
 	}
-	f, _ := strconv.ParseFloat(asset.Amount, 64)
-	if f == 0.0 {
+	if amt == 0.0 {
 		return true
 	}
 	// [StellarX Ticker]: assets need at least some adoption to show up
-	if asset.NumAccounts < 10 {
+	if asset.Accounts.Unauthorized+asset.Accounts.Authorized+asset.Accounts.AuthorizedToMaintainLiabilities < 10 {
```
tamirms, line 36: I think it would be easier to read if you stored asset.Accounts.Unauthorized+asset.Accounts.Authorized+asset.Accounts.AuthorizedToMaintainLiabilities in a new variable

services/ticker/internal/scraper/asset_scraper_test.go
```diff
@@ -14,69 +14,6 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestShouldDiscardAsset(t *testing.T) {
```
tamirms, line 17: what's the motivation behind removing this test entirely?

tamirms:
@karthikiyer56 

> As a part of this cleanup/deprecation task, I think the fields in Db structure ExpAssetStat also need to be removed.
I think that means the 2 rows in expat_asset_stat table need to be dropped.
can someone confirm if that is true?

Yes, you're correct. You will need to add a new [migration](https://github.com/stellar/go/tree/master/services/horizon/internal/db2/schema/migrations) which removes the `amount` and `num_accounts` columns from the exp_asset_stats table. Here is an example of a migration which does something similar that you can use as a reference point:

https://github.com/stellar/go/blob/master/services/horizon/internal/db2/schema/migrations/35_drop_participant_id.sql

services/ticker/internal/scraper/asset_scraper.go
tamirms: > Does this need to be simply if asset.Balances.authorized, instead of adding all 3?

yes, only the authorized accounts / balances need to be considered:

https://github.com/stellar/go/blob/master/services/horizon/internal/ingest/processors/asset_stats_processor.go#L459-L460

.gitignore
karthikiyer56: Yes,  I did intend to add this.
without this change, my `git status` shows up like so
```
On branch karthik/5438/deprecate-horizon-fields-for-protocol-22
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   .gitignore

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	services/horizon/bucket-cache/

no changes added to commit (use "git add" and/or "git commit -a")
```

Maybe, because I ran tests from within `services/horizon` ?
When I see the files there, it looks like xdr generated code?
Is it even supposed to be inside services/horizon, or only at top level?

My intention was to make it so that I dont accidentally add it with `git add`

services/ticker/internal/scraper/asset_scraper.go
karthikiyer56: Looks like I might not need this after all, but for future use
 can someone point me to a place where I can park or even find helper utils like this?

I tried finding something called `utils` directory (couldnt think of any other name), but the occurences that I did find didnt seem to indicate  a place where common utils used across the entire go monorepo might be stored.

services/ticker/internal/scraper/asset_scraper_test.go
karthikiyer56: For the most part, this test was creating fixtures with the to-be-removed fields `Amount` and `NumAccounts`, so I ended up deleting them all.

Now that I look, I think the test fixtures need to be modified to add the new `Accounts` and `Balances` fields 

will make the change

services/ticker/internal/scraper/asset_scraper.go
sreuland: global shared common/static utilities are probably regarded more an anti-pattern, as it would lead to some brittle coupling between unrelated packages, i.e. if one client package needed to change some aspect or behavior, it would break others, ideally, functionality is encapsulated within its package at least, you can see some common functions that are attached as receiver methods to public structs in a package which does promote shared usage of those from all clients.

Shaptic: Probably somewhere in [go/support](https://github.com/stellar/go/tree/master/support)?

services/horizon/internal/db2/schema/migrations/68_remove_deprecated_fields_from_exp_asset_stats.sql
```diff
@@ -0,0 +1,18 @@
+-- +migrate Up
+
+ALTER TABLE exp_asset_stats DROP CONSTRAINT valid_num_accounts;
+
+ALTER TABLE exp_asset_stats
+    DROP COLUMN num_accounts;
```
karthikiyer56, line 6: @stellar/platform-data-committers : Requesting clarification from you folks.

Will the removal of these fields from `exp_asset_stats` break any of your workflows/pipelines, per chance?

Broader question:
do any of the ETL pipelines in the data team even read, in any way/shape/form, from horizon related tables, i.e the RDS instances run by SDF?

services/ticker/internal/scraper/asset_scraper.go
karthikiyer56: done.
removed the others from the list

services/ticker/internal/scraper/asset_scraper.go
karthikiyer56: retaining only `authorized`

services/ticker/internal/scraper/asset_scraper_test.go
karthikiyer56: done.
updated the test fixture, instead of removing it altogether

karthikiyer56:
> @karthikiyer56
> 
> > As a part of this cleanup/deprecation task, I think the fields in Db structure ExpAssetStat also need to be removed.
> > I think that means the 2 rows in expat_asset_stat table need to be dropped.
> > can someone confirm if that is true?
> 
> Yes, you're correct. You will need to add a new [migration](https://github.com/stellar/go/tree/master/services/horizon/internal/db2/schema/migrations) which removes the `amount` and `num_accounts` columns from the exp_asset_stats table. Here is an example of a migration which does something similar that you can use as a reference point:
> 
> https://github.com/stellar/go/blob/master/services/horizon/internal/db2/schema/migrations/35_drop_participant_id.sql

@tamirms 
done

services/horizon/internal/db2/schema/migrations/68_remove_deprecated_fields_from_exp_asset_stats.sql
tamirms: The ETL pipelines do not ingest anything from the Horizon DB tables. Instead they ingest directly from the ledger close meta objects emitted by stellar-core. So, this schema change should not impact any of the data pipelines

tamirms (approved):

PR #5484, opened 2024-10-03
services/horizon: Remove --parallel-job-size config parameter used for reingestion.

### LINKED ISSUE #5468: services/horizon: adjust ParallelJobSize parameter when reingesting from BufferedStorageBackend (datastore)
TLDR;
For reingestion using the BufferedStorageBackend, the [ParallelJobSize](https://github.com/stellar/go/blob/f668d5b831715d0b55010c4f5b8d7ea2d6313065/services/horizon/cmd/db.go#L491-L493)
 parameter is currently set to a default of 100. This creates a conflict with the buffer_size parameter in the BufferedStorageBackend configuration. To fix this, we should either remove the ParallelJobSize parameter for reingestion using BufferedStorageBackend or set it to a larger value by default.

More info:
The ParallelJobSize parameter determines how many ledgers are processed in each task during parallel ingestion. It is set to 100k by default for Captive Core ingestion. This number was determined  based on how many ledgers Captive Core can efficiently replay at once. However, when using the BufferedStorageBackend for reingestion, ParallelJobSize doesn’t really matter because the BufferedStorageBackend is stateless and isn’t affected by the number of ledgers assigned to it.

The BufferedStorageBackend has [buffer_size](https://github.com/stellar/go/blob/master/ingest/ledgerbackend/buffered_storage_backend.go#L21) config parameter which determines how many files are kept in memory. By setting ParallelJobSize to 100, we limit the usefulness of buffer_size. So for example, even if buffer_size is set to 1000, each task will still only process 100 ledgers due to the ParallelJobSize setting.

While ParallelJobSize doesn’t directly impact performance with the BufferedStorageBackend, it could be helpful to set it to a larger valu like 100k for more efficient reingestion of large ledger ranges. When reingesting larger ledger range, the older ledgers are  smaller and faster to process. This means some workers might finish their tasks faster than others. By dividing the workload into smaller tasks, those faster workers can take on additional work, resulting in better efficiency.

urvisavla:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

- Removed the `--parallel-job-size` config parameter. 
- buffer_size parameter is capped to the job/range size.

### Why
Fixes #5468 

### Known limitations
This could potentially disrupt any automation scripts using `--parallel-job-size` parameter, although it's unlikely since reingestion is generally run as a batch job on as needed basis.

### DIFF
--- a/ingest/ledgerbackend/buffered_storage_backend_test.go
+++ b/ingest/ledgerbackend/buffered_storage_backend_test.go
@@ -196,10 +196,10 @@ func TestNewLedgerBufferSizeLargerThanRangeSize(t *testing.T) {
 	assert.Eventually(t, func() bool { return len(ledgerBuffer.ledgerQueue) == 15 }, time.Second*1, time.Millisecond*50)
 	assert.NoError(t, err)
 
-	for i := uint32(0); i < endLedger; i++ {
+	for i := startLedger; i < endLedger; i++ {
 		lcm, err := ledgerBuffer.getFromLedgerQueue(context.Background())
 		assert.NoError(t, err)
-		assert.Equal(t, xdr.Uint32(startLedger+i), lcm.StartSequence)
+		assert.Equal(t, xdr.Uint32(i), lcm.StartSequence)
 	}
 	assert.Equal(t, ledgerRange, ledgerBuffer.ledgerRange)
 }
--- a/ingest/ledgerbackend/buffered_storage_backend_test.go
+++ b/ingest/ledgerbackend/buffered_storage_backend_test.go
@@ -160,6 +160,50 @@ func TestNewLedgerBuffer(t *testing.T) {
 	assert.Equal(t, ledgerRange, ledgerBuffer.ledgerRange)
 }
 
+func TestNewLedgerBufferSizeLessThanRangeSize(t *testing.T) {
+	startLedger := uint32(10)
+	endLedger := uint32(30)
+	bsb := createBufferedStorageBackendForTesting()
+	bsb.config.NumWorkers = 2
+	bsb.config.BufferSize = 10
+	ledgerRange := BoundedRange(startLedger, endLedger)
+	mockDataStore := createMockdataStore(t, startLedger, endLedger, partitionSize, ledgerPerFileCount)
+	bsb.dataStore = mockDataStore
+
+	ledgerBuffer, err := bsb.newLedgerBuffer(ledgerRange)
+	assert.Eventually(t, func() bool { return len(ledgerBuffer.ledgerQueue) == 10 }, time.Second*1, time.Millisecond*50)
+	assert.NoError(t, err)
+
+	for i := startLedger; i < endLedger; i++ {
+		lcm, err := ledgerBuffer.getFromLedgerQueue(context.Background())
+		assert.NoError(t, err)
+		assert.Equal(t, xdr.Uint32(i), lcm.StartSequence)
+	}
+	assert.Equal(t, ledgerRange, ledgerBuffer.ledgerRange)
+}
+
+func TestNewLedgerBufferSizeLargerThanRangeSize(t *testing.T) {
+	startLedger := uint32(1)
+	endLedger := uint32(15)
+	bsb := createBufferedStorageBackendForTesting()
+	bsb.config.NumWorkers = 2
+	bsb.config.BufferSize = 100
+	ledgerRange := BoundedRange(startLedger, endLedger)
+	mockDataStore := createMockdataStore(t, startLedger, endLedger, partitionSize, ledgerPerFileCount)
+	bsb.dataStore = mockDataStore
+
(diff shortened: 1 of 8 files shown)


services/horizon/cmd/db.go
```diff
@@ -118,14 +117,6 @@ func ingestRangeCmdOpts() support.ConfigOptions {
 			FlagDefault: uint(1),
 			Usage:       "[optional] if this flag is set to > 1, horizon will parallelize reingestion using the supplied number of workers",
 		},
-		{
```
urvisavla, line 121: I was waffling between just removing the parameter which results in an 'unknown flag' error if the parameter is supplied and explicitly handling the config parameter and throwing an error stating that this particular parameter has been removed. For now, I've just removed but I'm open to other suggestions.

tamirms: I think removing it is fine since we mention this in the changelog

ingest/ledgerbackend/buffered_storage_backend_test.go
```diff
@@ -160,6 +160,50 @@ func TestNewLedgerBuffer(t *testing.T) {
...
+	bsb.config.BufferSize = 100
+	ledgerRange := BoundedRange(startLedger, endLedger)
+	mockDataStore := createMockdataStore(t, startLedger, endLedger, partitionSize, ledgerPerFileCount)
+	bsb.dataStore = mockDataStore
+
+	ledgerBuffer, err := bsb.newLedgerBuffer(ledgerRange)
+	assert.Eventually(t, func() bool { return len(ledgerBuffer.ledgerQueue) == 15 }, time.Second*1, time.Millisecond*50)
+	assert.NoError(t, err)
+
+	for i := uint32(0); i < endLedger; i++ {
```
tamirms, line 199: why does i start at 0 instead of startLedger?

services/horizon/internal/ingest/parallel.go
```diff
@@ -113,12 +113,10 @@ func enqueueReingestTasks(ledgerRanges []history.LedgerRange, batchSize uint32,
...
-	batchSize := batchSizeSuggestion
-	if batchSize == 0 || rangeSize/batchSize < uint32(workerCount) {
-		// let's try to make use of all the workers
-		batchSize = rangeSize / uint32(workerCount)
-	}
+func calculateParallelLedgerBatchSize(rangeSize uint32, workerCount uint) uint32 {
+	// let's try to make use of all the workers
+	batchSize := rangeSize / uint32(workerCount)
+
 	// Use a minimum batch size to make it worth it in terms of overhead
```
tamirms, line 120: should there be a maximum batch size as well? I remember you mentioned that it could be helpful to have a maximum batch size because the ledger ingestion time varies based on whether the ledger is from recent history vs very old history. So, in the scenario where you want to reingest full history with, for example, 4 workers, the workers which handle the first half of history will be idle for a long time.

sreuland: it seems like adding max batch size here would be trying to address a larger concern of maximizing worker pool throughput(minimizing idle workers) which seems like more scope than this function is meant for, it's a good point on performance, wondering if it warrants a separate feature ticket to investigate how constant worker throughput could be accomplished?
 
one wild thought, workers could be interruptable, so that an idle worker could interrupt one that is running, and take some of it's upper range, the worker would stop it's captive core when it receives lcm for it's adjusted 'to' range in any case.

ingest/ledgerbackend/buffered_storage_backend_test.go
urvisavla: fixed

services/horizon/internal/ingest/parallel.go
urvisavla: Yes, I did mention that having a max batch size could result in better throughput especially when reingesting large ranges but I left it out of this implementation because we need to account for both Captive Core and BufferedStorage when determining the max batch size. For BufferedStorage, a smaller batch size would improves throughput but for Captive Core the overhead per worker is significant so a smaller batch size could reduce the  throughput. I'm not sure  if we should have separate batch sizes for each backend but a batch size of 10k or 20k might work well for both. I can either add it to the current PR or create a separate one to explore other options like Shawn’s suggested.

tamirms: > a batch size of 10k or 20k might work well for both

I think that's fine. We can start with a simple heuristic and later on improve upon that if necessary.

tamirms: I think you could also only apply the max batch size if we're using a BufferedStorage ledger backend. Captive core should be fine without a max batch size

sreuland (approved):
lgtm

services/horizon/internal/ingest/parallel.go
urvisavla: Updated the code to include maximum batch size limits based on our discussion.

tamirms (approved):

Issue #5437, opened 2024-08-20
Update Horizon ingestion and tests to support the new host function


Shaptic:
**Epic**: https://github.com/stellar/go/issues/5433
**Note**: All work should target a feature branch: [`protocol-22`](https://github.com/stellar/go/tree/protocol-22).
**Depends On**: #5423 

-------------------------

* Update Horizon's [operations_processor.go](https://github.com/stellar/go/blob/master/services/horizon/internal/ingest/processors/operations_processor.go#L672) to inspect the new type
* Ensure that Horizon's [invocation](https://github.com/stellar/go/blob/master/services/horizon/internal/integration/invokehostfunction_test.go) and [SAC tests](https://github.com/stellar/go/blob/master/services/horizon/internal/integration/sac_test.go) use / validate the new type

tamirms:
implemented in https://github.com/stellar/go/pull/5488

PR #5488, opened 2024-10-07
services/horizon/internal/ingest/processors: Ingest new invoke host operation type

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Fixes https://github.com/stellar/go/issues/5437

Update Horizon ingestion and tests to support creating a contract with a constructor.

The test case included in this PR covers invoking a constructor which has multiple arguments, requires auth, and performs asset balance changes.

### Known limitations

[N/A]


services/horizon/internal/integration/testdata/soroban_constructor_contract.wasm
Shaptic, line 1: I'm not sure about the precedent here but does it make sense for these to be compiled into the repo vs. a separate `make` command to run alongside the integration tests? Then we can be sure that we have all of the WASMs. And can run [optimization](https://developers.stellar.org/docs/build/smart-contracts/getting-started/hello-world#optimizing-builds) on them, too.

services/horizon/internal/integration/invokehostfunction_test.go
```diff
@@ -126,6 +130,105 @@ func TestContractInvokeHostFunctionCreateContractByAddress(t *testing.T) {
...
+
+	issuer := itest.Master().Address()
+	code := "USD"
+	asset := xdr.MustNewCreditAsset(code, issuer)
+	createSAC(itest, asset)
+
+	// establish which account will be contract owner, and load it's current seq
+	sourceAccount, err := itest.Client().AccountDetail(horizonclient.AccountRequest{
+		AccountID: itest.Master().Address(),
+	})
```
Shaptic, line 150: You can use `itest.MasterAccount()` for this, I think?

services/horizon/internal/ingest/processors/operations_processor.go
```diff
@@ -717,6 +702,32 @@ func (operation *transactionOperationWrapper) Details() (map[string]interface{},
...
 			}
+		case xdr.HostFunctionTypeHostFunctionTypeCreateContractV2:
+			args := op.HostFunction.MustCreateContractV2()
+			switch args.ContractIdPreimage.Type {
+			case xdr.ContractIdPreimageTypeContractIdPreimageFromAddress:
+				fromAddress := args.ContractIdPreimage.MustFromAddress()
+				address, err := fromAddress.Address.String()
+				if err != nil {
+					panic(fmt.Errorf("error obtaining address for: %s", args.ContractIdPreimage.Type))
+				}
```
Shaptic, line 713: Do we want `panic`s here vs. returning errors? I see both in this code but it'd be good to unify.

services/horizon/internal/ingest/processors/operations_processor.go
```diff
@@ -740,6 +751,24 @@ func (operation *transactionOperationWrapper) Details() (map[string]interface{},
...
 }
 
+func extractFunctionArgs(args []xdr.ScVal) []map[string]string {
+	params := make([]map[string]string, 0, len(args))
+	for _, param := range args {
+		serializedParam := map[string]string{}
+		serializedParam["value"] = "n/a"
+		serializedParam["type"] = "n/a"
+
+		if scValTypeName, ok := param.ArmForSwitch(int32(param.Type)); ok {
```
Shaptic, line 761: When would this not be `ok`? Or is this just defensive?

Shaptic (commented):
Coupla minor comments but LGTM otherwise!

services/horizon/internal/integration/contracts/constructor/src/lib.rs
```diff
@@ -0,0 +1,24 @@
...
+impl Contract {
+    pub fn __constructor(env: Env, sender: Address, token_contract: Address, amount: i128) {
+        let client = token::Client::new(&env, &token_contract);
+        client.transfer(&sender, &env.current_contract_address(), &amount);
+        env.storage().persistent().set(&KEY, &1_u32);
+        env.storage().instance().set(&KEY, &2_u32);
+        env.storage().temporary().set(&KEY, &3_u32);
+    }
+
+    pub fn get_data(env: Env, key: Symbol) -> u32 {
```
sreuland, line 19: could probably delete this fn and the env usage in constructor? the contract storage doesn't appear to be part of test verification.

sreuland (approved):
lgtm, nice work!

services/horizon/internal/ingest/processors/operations_processor.go
sreuland: good point, there was prior usage there with panic, but returning err seems more idiomatic for error management, i.e. it would be rough for all of ingestion to halt due to this?

Shaptic: Ingestion would still halt on an error since it'd just enter an infinite retry loop, but the rest of the service would at least still be alive?

services/horizon/internal/ingest/processors/operations_processor.go
tamirms: I don't ever expect `ok` to be false but there is no `MustArmForSwitch()` function

Shaptic: Perfect, just asking for my own edification

tamirms:
@Shaptic can you take another look? I think all code review feedback has been addressed

Shaptic (approved):
LGTM :clap:

PR #5489, opened 2024-10-08
ingest/cdp: fixed producer test for caller ctx cancel to be consistent

sreuland:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

fixed inconsistent test on producer caller context cancelled.

### Why

the test had inconsistent pass/fail results due to race conditions

### DIFF
--- a/ingest/cdp/producer_test.go
+++ b/ingest/cdp/producer_test.go
@@ -196,6 +196,13 @@ func TestBSBProducerCallerCancelsCtx(t *testing.T) {
 		BufferedStorageConfig: DefaultBufferedStorageBackendConfig(1),
 	}
 
+	// prevent the ledgerbuffer from being eagerly async loaded initially up to
+	// buffer-size ledgers and the requested test range, forces new load attempt
+	// during GetLedger which is when ctx is checked, test asks for 3 ledger range
+	// since 2 ledgers are by default eagerly loaded.
+	pubConfig.BufferedStorageConfig.BufferSize = 1
+	pubConfig.BufferedStorageConfig.NumWorkers = 1
+
 	// the buffering runs async, test needs to stub datastore methods for potential invocation,
 	// but is race, since test also cancels the backend context which started the buffer,
 	// so, not deterministic, no assert on these.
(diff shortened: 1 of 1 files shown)


ingest/cdp/producer_test.go
```diff
@@ -196,6 +196,13 @@ func TestBSBProducerCallerCancelsCtx(t *testing.T) {
 		BufferedStorageConfig: DefaultBufferedStorageBackendConfig(1),
 	}
 
+	// prevent the ledgerbuffer from being eagerly async loaded initially up to
```
sreuland, line 199: one thing I noticed while exercising a test where caller cancels context to BufferedStorageBackend, the [PrepareRange()](https://github.com/stellar/go/blob/master/ingest/ledgerbackend/buffered_storage_backend.go#L170) doesn't check the context, should it be as general convention? the context is actually unused there. If it did check, then this test would not need to manipuate the buffer size and range aspect in order to get the cancelled context to be acted upon from BSB.

ingest/cdp/producer_test.go
```diff
@@ -205,7 +207,11 @@ func TestBSBProducerCallerCancelsCtx(t *testing.T) {
 		FilesPerPartition: 1,
 	})
 
-	mockDataStore.On("GetFile", mock.Anything, "FFFFFFFD--2.xdr.zstd").Return(makeSingleLCMBatch(2), nil)
+	// artifically stall the ledgerbuffer worker so nothing new arriving on ledgerqueue
```
sreuland, line 210: @tamirms , while the ledger buffer eagerly loads up to buffer size in background, I noticed getting a context cancelled in there was a bit delicate, this forced it for test concerns, noticed it [selects on caller ctx and the queue](https://github.com/sreuland/go/blob/fix_producer_test/ingest/ledgerbackend/ledger_buffer.go#L207), and if the ctx is cancelled, the behavior there would be somewhat random when that gets noticed, i.e. as long as the queue channel is receiving more files, it may preclude the select from choosing context done signal right away, is that intended, i.e. to let 'draining' like behavior, rather than aborting that as soon as context canceled?

sreuland: I can move the cancel invocation to the suggested callback, which is nice, but need to keep the mock on `GetFile("FFFFFFFC--3.xdr.zstd")` due to the [ledger buffer eager fetch behavior in getFromLedgerQueue()](https://github.com/stellar/go/pull/5489/commits/37eb785cb065fde659b832702a67a3ab4f89dbcf#diff-f2385e028548bd0adeee892f65c691a907b5d2117f24ccca571560f22d5be54aR215-R217).

tamirms (approved):

Issue #5414, opened 2024-08-06
services/horizon/ingest: express parsing routines for Offers as public, reusable functions


sreuland:
### What problem does your feature solve?
New application development that needs to run ingestion pipeline will tend to implement(repeat) many similar parsing routines to generate derived 'Offer' models that Horizon ingestion does internally. 

Horizon has many routines that are package private, internal to ingestion which parse and derive 'Offer' from stellar tx meta, so, they can't be re-used externally as-is.

### What would you like to see?
Follows [same design proposal for reusable transformer function for a common model](https://docs.google.com/document/d/17wy1r5l5cUWxAj5Qty_faXni4tsXER4pAdvTo05nnO0/edit#heading=h.ck7b7cn6zl5v) 

A few of these private ingestion parsing routines in Horizon exposed as pure(no side effects) functions and package public so they can be re-used by external apps to derive the same 'Offer' data models quickly in their ingestion pipelines.

The presence of public functions for parsing 'Offer' model can further encourage external adoption of [proposed streaming ingestion pipeline](https://docs.google.com/document/d/17wy1r5l5cUWxAj5Qty_faXni4tsXER4pAdvTo05nnO0/edit#heading=h.3dy6vkm) of which operators within a stream invoke these functions directly to include offers data in their stream processing.

Create new pure function and model for 'Offer' in the shared [ingest](https://github.com/stellar/go/tree/master/ingest) package. Horizon internal ingestion should not be changed or refactored, later this may be revisited.

* `ingest.HasOffer(op xdr.Operation) bool` 
* for all operations that qualified as offer in `HasOffer()`, pull up the 'details' parsing routines of these operations from `transactionOperationWrapper.Details()` and fold into -> `ingest.DeriveOffer(tx xdr.Transaction, op xdr.Operation) []OfferEvent`:
  * managebuy
  * managesell
  * createsell 
  
* Offer data is modeled as a derived event representing the life cycle events of an offer, this is initial estimate of event schema, during effort, can change these:

```  
  type OfferPartiallyFilledEvent struct {    
        Price:                         "8.0000000"
        Amount:                    "89.9342000"
        PriceAssetRatioD:    "1"
        PriceAssetRatioN:    "8"
        BuyingAsset:            "native"
        SellingAsset:             "XCZ:GC2BTZPHJEVOGPW6MJUVUPANWLRGK3JMD26ZMP64ZRMYXBRSW5CIKEW5"
        Timestamp:               "1723153694"
   }
```

```
[...]

PR #5501, opened 2024-11-04
XDRill ledger low level helpers example

chowbao:
Example ledgers code for https://github.com/stellar/platform-design-docs/pull/2

### DIFF
--- a/exp/xdrill/ledgers/ledgers.go
+++ b/exp/xdrill/ledgers/ledgers.go
@@ -1,0 +1,191 @@
+package ledgers
+
+import (
+	"encoding/base64"
+	"fmt"
+	"time"
+
+	"github.com/stellar/go/exp/xdrill/utils"
+	"github.com/stellar/go/xdr"
+)
+
+type Ledgers struct {
+	xdr.LedgerCloseMeta
+}
+
+func (l Ledgers) Sequence() uint32 {
+	return uint32(l.LedgerHeaderHistoryEntry().Header.LedgerSeq)
+}
+
+func (l Ledgers) ID() int64 {
+	return utils.NewID(int32(l.LedgerSequence()), 0, 0).ToInt64()
+}
+
+func (l Ledgers) Hash() string {
+	return utils.HashToHexString(l.LedgerHeaderHistoryEntry().Hash)
+}
+
+func (l Ledgers) PreviousHash() string {
+	return utils.HashToHexString(l.PreviousLedgerHash())
+}
+
+func (l Ledgers) CloseTime() int64 {
+	return l.LedgerCloseTime()
+}
+
+func (l Ledgers) ClosedAt() time.Time {
+	return time.Unix(l.CloseTime(), 0).UTC()
+}
+
+func (l Ledgers) TotalCoins() int64 {
+	return int64(l.LedgerHeaderHistoryEntry().Header.TotalCoins)
+}
+
+func (l Ledgers) FeePool() int64 {
+	return int64(l.LedgerHeaderHistoryEntry().Header.FeePool)
+}
+
+func (l Ledgers) BaseFee() uint32 {
+	return uint32(l.LedgerHeaderHistoryEntry().Header.BaseFee)
+}
+
+func (l Ledgers) BaseReserve() uint32 {
+	return uint32(l.LedgerHeaderHistoryEntry().Header.BaseReserve)
+}
+
+func (l Ledgers) MaxTxSetSize() uint32 {
+	return uint32(l.LedgerHeaderHistoryEntry().Header.MaxTxSetSize)
+}
+
+func (l Ledgers) LedgerVersion() uint32 {
+	return uint32(l.LedgerHeaderHistoryEntry().Header.LedgerVersion)
+}
+
+func (l Ledgers) GetSorobanFeeWrite1Kb() (int64, bool) {
+	lcmV1, ok := l.GetV1()
+	if ok {
+		extV1, ok := lcmV1.Ext.GetV1()
+		if ok {
+			return int64(extV1.SorobanFeeWrite1Kb), true
+		}
+	}
+
+	return 0, false
+}
+
+func (l Ledgers) GetTotalByteSizeOfBucketList() (uint64, bool) {
+	lcmV1, ok := l.GetV1()
+	if ok {
+		return uint64(lcmV1.TotalByteSizeOfBucketList), true
+	}
+
+	return 0, false
+}
+
+func (l Ledgers) GetNodeID() (string, bool) {
+	LedgerCloseValueSignature, ok := l.LedgerHeaderHistoryEntry().Header.ScpValue.Ext.GetLcValueSignature()
+	if ok {
+		nodeID, ok := utils.GetAddress(LedgerCloseValueSignature.NodeId)
+		if ok {
+			return nodeID, true
+		}
+	}
+
+	return "", false
+}
+
+func (l Ledgers) GetSignature() (string, bool) {
+	LedgerCloseValueSignature, ok := l.LedgerHeaderHistoryEntry().Header.ScpValue.Ext.GetLcValueSignature()
+	if ok {
+		return base64.StdEncoding.EncodeToString(LedgerCloseValueSignature.Signature), true
+	}
+
+	return "", false
+}
+
+func (l Ledgers) GetTransactionCounts() (successTxCount, failedTxCount int32, ok bool) {
+	transactions := getTransactionSet(l)
+	results := l.V0.TxProcessing
+	txCount := len(transactions)
+	if txCount != len(results) {
+		return 0, 0, false
+	}
+
+	for i := 0; i < txCount; i++ {
+		if results[i].Result.Successful() {
+			successTxCount++
+		} else {
+			failedTxCount++
+		}
+	}
+
+	return successTxCount, failedTxCount, true
+}
+
+func (l Ledgers) GetOperationCounts() (operationCount, txSetOperationCount int32, ok bool) {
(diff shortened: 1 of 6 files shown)


exp/xdrill/ledgers/ledgers.go
```diff
@@ -0,0 +1,191 @@
...
+
+func (l Ledgers) MaxTxSetSize() uint32 {
+	return uint32(l.LedgerHeaderHistoryEntry().Header.MaxTxSetSize)
+}
+
+func (l Ledgers) LedgerVersion() uint32 {
+	return uint32(l.LedgerHeaderHistoryEntry().Header.LedgerVersion)
+}
+
+func (l Ledgers) GetSorobanFeeWrite1Kb() (int64, bool) {
```
sreuland, line 64: can probably drop the `Get` prefix in favor of convention, i.e. any fn here will return a value.

exp/xdrill/ledgers/ledgers.go
```diff
@@ -0,0 +1,191 @@
...
+			successTxCount++
+		} else {
+			failedTxCount++
+		}
+	}
+
+	return successTxCount, failedTxCount, true
+}
+
+func (l Ledgers) GetOperationCounts() (operationCount, txSetOperationCount int32, ok bool) {
```
sreuland, line 125: docs on each method should help promote these also, I'm not even sure what is the significance of `txSetOperationCount` ?

exp/xdrill/ledgers/ledgers.go
```diff
@@ -0,0 +1,191 @@
...
+
+			operationCount += int32(len(operationResults))
+		}
+
+	}
+
+	return operationCount, txSetOperationCount, true
+}
+
+func getTransactionSet(l Ledgers) (transactionProcessing []xdr.TransactionEnvelope) {
```
sreuland, line 153: would this be where new layer can reference the new non-xdr model instead, i.e. [[]transactions.Transactions](https://github.com/stellar/go/pull/5501/files#diff-755c45b8f4393e62acc70fb5570d1cb9d608194ea00fa6e102d2b966b02073d9R7) ?

exp/xdrill/ledgers/ledgers.go
chowbao: Yeah docs will help. I'll definitely add them in the real PR/implementation.

As for the difference between `operationCount`, `txSetOperationCount` is the total operation count (successful + failed) and only successful operation count.

The names are definitely confusing and unclear but I copied these from stellar-etl code that predates me lol
I will definitely rename/refactor these functions in the real PR/implementation

exp/xdrill/ledgers/ledgers.go
chowbao: That wasn't the intent of this function. This was just carryover from stellar-etl to count the transactions/operations.

I think you're right though that it would be nice to include a function for users to call to get the non-xdr model instead of [manually doing it](https://github.com/stellar/go/pull/5501/files#diff-66f0fa4c407fa259fe465008660d9f4595050522c1fd7254dfc7f2924ad9b671R36-R38)

exp/xdrill/transform_ledger.go
```diff
@@ -0,0 +1,104 @@
...
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/stellar/go/exp/xdrill/ledgers"
+	"github.com/stellar/go/xdr"
+)
+
+type LedgerOutput struct {
```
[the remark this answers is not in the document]
chowbao: Ah yeah makes sense

exp/xdrill/ledgers/ledgers.go
chowbao: Updated in 146dd525e7f8816819508cdcc2709fc6de3cdbf8

exp/xdrill/transform_ledger.go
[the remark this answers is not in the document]
chowbao: Updated in 146dd525e7f8816819508cdcc2709fc6de3cdbf8

exp/xdrill/transform_ledger.go
```diff
@@ -0,0 +1,104 @@
...
+	MaxTxSetSize               uint32    `json:"max_tx_set_size"`
+	ProtocolVersion            uint32    `json:"protocol_version"`
+	LedgerID                   int64     `json:"id"`
+	SorobanFeeWrite1Kb         int64     `json:"soroban_fee_write_1kb"`
+	NodeID                     string    `json:"node_id"`
+	Signature                  string    `json:"signature"`
+	TotalByteSizeOfBucketList  uint64    `json:"total_byte_size_of_bucket_list"`
+}
+
+func TransformLedger(lcm xdr.LedgerCloseMeta) (LedgerOutput, error) {
```
sreuland: will this pattern scale well for the sdk, i.e. what will the next fn be named that transforms to a different model, the pattern implies something like `TransformLedgerTo<DerivedModel>() <DerivedModel> ` which is probably fine, this fn is specifically  `TransformLedgerToOutput` correct?

chowbao: > this fn is specifically TransformLedgerToOutput correct?

Yes

I didn't think of the pattern for how to name custom processors. This is just how we define/call it in stellar-etl and was mostly used to serve as one possible way to call the low level helper functions.

It makes sense to standardize the name/pattern though. I think it would be hard though.
I think there's currently the idea from like 
* [OffersProcessor()](https://github.com/stellar/go/issues/5414) with `HasOffer` and `DeriveOffer`
* stellar-etl has an `input/*.go` functions to read the ledgers --> transform/*.go like the above `TransformLedger` to process the data
* `TransformLedgerTo<DerivedModel>() <DerivedModel>` `TransformLedgerToOutput` as noted above

And I think there'd then be another layer depending on what kind of output/data model the user wants for their application. Like a user could rightfully/wrongfully do:
```
func ProcessLedgerMetadata(lcm xdr.LedgerCloseMeta) {
  TransformLedgerToOutput()
  TransformOfferToOutput()
  
  if ingest.HasPayment(lcm) {
    ingest.DerivePayment(lcm)
  }
}
```

This might be worth bringing back to the [design doc](https://github.com/stellar/platform-design-docs/pull/2) or in the team meeting to discuss

exp/xdrill/ledgers/ledgers.go
```diff
@@ -0,0 +1,193 @@
...
+import (
+	"encoding/base64"
+	"fmt"
+	"time"
+
+	"github.com/stellar/go/exp/xdrill/utils"
+	"github.com/stellar/go/xdr"
+)
+
+type Ledgers struct {
```
sreuland, line 12: should this be singular,`Ledger`, all the receiver methods imply such.

chowbao: Ah yeah that's a good catch. I'll make the packages, structs, and functions all singular instead of plural

exp/xdrill/transform_ledger.go
sreuland: yes, good suggestion on opening up to further design discussion, another option to consider may be to remove the `transformer` package to avoid grey area where app domain opinions inevitably start leaking in as the derived model, refactor this transform as a more factual attribute retrieval like `ledgers.Ledgers.Closed() ClosedOutput`, ideally this new sdk of low level helpers should give developers enough tooling that they can quickly write their own lcm-to-model transforms, rather than the sdk codifying transforms?

exp/xdrill/ledgers/ledgers.go
chowbao: Updated in 3b90190992bc678d96888168c8c11bdaaa457f9c

exp/xdrill/transform_ledger.go
chowbao: > ideally this new sdk of low level helpers should give developers enough tooling that they can quickly write their own lcm-to-model transforms, rather than the sdk codifying transforms?

Yes 100%

exp/xdrill/transform_ledger_xdr.go
chowbao, line 1: Example using helper functions only in xdr. I think for the LedgerClosed case this works really well

Type casting should be removed and just added to the functions themselves if possible

xdr/account_entry.go
chowbao, line 1: Didn't look into other ledgerEntries but it seems like adding helper functions to the xdr seems good. The only issue I found was that for a ledgerEntry like AccountEntry there are existing functions and fields that are of `xdr.Int64` or `xdr.Uint32` which would force the user to type cast in their application. It's just a small annoying thing but I wonder if it would be worthwhile to figure out a way to remove the casting while still using a reasonable function name

Another thing with ledgerEntries that I found was it's awkward to get to a specific entry type like AccountEntry before calling these functions. I couldn't figure out a way to allow the user to use one of `Changes` or `LedgerEntry` themselves to get specific data because some EntryTypes have similarly named fields like Balance. Is it possible to make LedgerEntry functions smart enough to return the correct field from the correct entryType?

```
// This could work but I feel like it might be confusing to users where the balance actually came from if the user doesn't realize they need to look at the LedgerEntry.Type
func (l LedgerEntry) Balance() float64 {
  switch l.Type {
  case LedgerEntryTypeAccount:
    return l.Account.Balance
  case LedgerEntryTypeTrustline:
    return l.TrustLine.Balance
  }
}
```

xdr/ledger_close_meta.go
chowbao, line 1: Added helper functions to LCM

mollykarcher:
This pull request has been inactive for an extended period, and we’re performing routine maintenance to close out long-standing open PRs.

If this work remains relevant or you would like to continue it, please feel free to reopen the PR or submit a new one with updated information.

PR #5523, opened 2024-11-11
services/horizon: Add missing skip-txmeta flag in the /transactions/ endpoint.

urvisavla:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add missing `skip-txmeta` flag in the `/transactions/{tx-id}` endpoint.

### Why
`skip-txmeta` flag was implemented in https://github.com/stellar/go/pull/5228. However, the `/transactions/` endpoint was missed. As a result, the 22.0.0-rc2 horizon deployed on the testnet still includes `result_meta_xdr` in its response.

### Known limitations
N/A

### DIFF
--- a/services/horizon/internal/httpx/router.go
+++ b/services/horizon/internal/httpx/router.go
@@ -295,7 +295,7 @@ func (r *Router) addRoutes(config *RouterConfig, rateLimiter *throttled.HTTPRate
 	r.Route("/transactions", func(r chi.Router) {
 		r.With(historyMiddleware).Method(http.MethodGet, "/", streamableHistoryPageHandler(ledgerState, actions.GetTransactionsHandler{LedgerState: ledgerState, SkipTxMeta: config.SkipTxMeta}, streamHandler))
 		r.Route("/{tx_id}", func(r chi.Router) {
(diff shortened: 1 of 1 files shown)


mollykarcher:
It was the overall intent of this config to prevent/help data growth, so we wanted to stop saving txmeta xdr to the database when it was enabled. 

My question is, if all that's needed to fix this issue is this filtering change at the API layer, that still seems like a bug, because there shouldn't even have been xdr/meta saved in the database for the API to serve. Am I misunderstanding something?

urvisavla:
> It was the overall intent of this config to prevent/help data growth, so we wanted to stop saving txmeta xdr to the database when it was enabled.
> 
> My question is, if all that's needed to fix this issue is this filtering change at the API layer, that still seems like a bug, because there shouldn't even have been xdr/meta saved in the database for the API to serve. Am I misunderstanding something?

I’ll need to verify but my guess is that it’s likely the older ledgers where the flag wasn’t enabled. I believe it should be empty for the most recent ledgers but i'll verify.

mollykarcher:
Ah yeah, you're right. Looking through `/transactions?order=desc` and it's absent for all recent ones. 👍

mollykarcher (approved):

mollykarcher:
Another note, we might consider dropping the data in these columns for all historical stuff, via a manual SQL command.

PR #5531, opened 2024-11-14
Karthik/5614/generic offers parsing

karthikiyer56:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

### Why

### Known limitations

[TODO or N/A]

### DIFF
--- a/ingest/offers/main.go
+++ b/ingest/offers/main.go
@@ -14,7 +14,7 @@ const (
 	EventTypeOfferClosed  = "OfferClosed"
 )
 
-// Base struct with common fields for all offer events
+// Base struct with common fields for all offer events.
 type OfferEventData struct {
 	SellerId string
 	OfferID  int64
@@ -22,7 +22,7 @@ type OfferEventData struct {
 	BuyingAsset  xdr.Asset
 	SellingAsset xdr.Asset
 
-	RemainingAmount int64
+	RemainingAmount int64 // Remaining amount that still needs to be filled for this offer
 	PriceN          int32
 	PriceD          int32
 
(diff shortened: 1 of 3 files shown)


ingest/offers/main.go
```diff
@@ -0,0 +1,108 @@
...
+		PriceN:             int32(offer.Price.N),
+		PriceD:             int32(offer.Price.D),
+		Flags:              int32(offer.Flags),
+		IsPassive:          int32(offer.Flags) == int32(xdr.OfferEntryFlagsPassiveFlag),
+		LastModifiedLedger: uint32(e.LastModifiedLedgerSeq),
+		Sponsor:            utils.LedgerEntrySponsorToNullString(*e),
+	}
+}
+
+func ProcessOffer(change ingest.Change) OfferEvent {
```
karthikiyer56, line 79: This ProcessOffer has been pulled out into the top level `/ingest` package.
now horizon and stellar-etl can leverage this ProcessOffer function and thus eliminate the code duplication.

Further down in  this PR in `offer_processor.go`, I have updated the horizon code to use this common function
next step: make stellar-etl use this `ProcesOffer` function [here](https://github.com/stellar/stellar-etl/blob/master/internal/transform/offer.go#L13)   Cc: @chowbao

ingest/offers/main.go
```diff
@@ -0,0 +1,108 @@
...
+
+// Constants for event types
+const (
+	EventTypeOfferCreated = "OfferCreated"
+	EventTypeOfferFill    = "OfferUpdated"
+	EventTypeOfferClosed  = "OfferClosed"
+)
+
+// Base struct with common fields for all offer events.
+type OfferEventData struct {
```
tamirms, line 18: I think it would be better to just use xdr.OfferEntry instead of this because it contains the same data so there is no need to introduce another struct

PR #5534, opened 2024-11-14
Add General purpose API wrapper util

amishas157:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This PR adds a general purpose API wrapper which could be used to call any API endpoint with given query params and headers.
There are two modules added:
- Client: Contains function to call an API. This supports retries in case there are throttling or service unavailable.
- Request: Contains function to build requestBody with given inputs such request type, headers etc

Also updated httptest library to support mocking http request when we want to see multiple results on subsequent api calls. This is particularly helpful when we want to test whether API retries work as expected or not.

### Why

This will be used in stellar-etl to invoke various APIs-retool, github etc to fetch data and store in bigquery for analytics.

### Known limitations

I did not add support for pagination as we don't have a use case as of now. However, this util can be extended in future as needed.

### Questions

I am not sure if this is right place to add this library. I am open to relocating it where it finds better fit.

### DIFF
--- a/utils/apiclient/client.go
+++ b/utils/apiclient/client.go
@@ -1,98 +1,85 @@
 package apiclient
 
 import (
-	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"net/http"
 	"net/url"
+	"time"
 
 	"github.com/pkg/errors"
 )
 
-func (c *APIClient) createRequestBody(endpoint string, queryParams url.Values) (*http.Request, error) {
-	fullURL := c.url(endpoint, queryParams)
-	req, err := http.NewRequest("GET", fullURL, nil)
-	if err != nil {
-		return nil, errors.Wrap(err, "http GET request creation failed")
-	}
-	return req, nil
+const (
+	maxRetries     = 5
+	initialBackoff = 1 * time.Second
+)
+
+func isRetryableStatusCode(statusCode int) bool {
+	return statusCode == http.StatusTooManyRequests || statusCode == http.StatusServiceUnavailable
 }
 
-func (c *APIClient) callAPI(req *http.Request) (interface{}, error) {
+func (c *APIClient) GetURL(endpoint string, qstr url.Values) string {
+	return fmt.Sprintf("%s/%s?%s", c.BaseURL, endpoint, qstr.Encode())
+}
+
+func (c *APIClient) CallAPI(reqParams RequestParams) (interface{}, error) {
+	if reqParams.QueryParams == nil {
+		reqParams.QueryParams = url.Values{}
+	}
+
+	if reqParams.Headers == nil {
+		reqParams.Headers = map[string]interface{}{}
+	}
+
+	url := c.GetURL(reqParams.Endpoint, reqParams.QueryParams)
+	reqBody, err := CreateRequestBody(reqParams.RequestType, url)
+	if err != nil {
+		return nil, errors.Wrap(err, "http request creation failed")
+	}
+
+	SetAuthHeaders(reqBody, c.authType, c.authHeaders)
+	SetHeaders(reqBody, reqParams.Headers)
 	client := c.HTTP
 	if client == nil {
 		client = &http.Client{}
 	}
 
-	resp, err := client.Do(req)
-	if err != nil {
-		return nil, errors.Wrap(err, "http GET request failed")
-	}
-	defer resp.Body.Close()
-
-	if resp.StatusCode != http.StatusOK {
-		return nil, fmt.Errorf("API request failed with status %d", resp.StatusCode)
-	}
-
-	body, err := ioutil.ReadAll(resp.Body)
-	if err != nil {
-		return nil, fmt.Errorf("failed to read response body: %w", err)
-	}
-
 	var result interface{}
-	if err := json.Unmarshal(body, &result); err != nil {
-		return nil, fmt.Errorf("failed to unmarshal JSON: %w", err)
+	retries := 0
+
+	for retries <= maxRetries {
+		resp, err := client.Do(reqBody)
+		if err != nil {
+			return nil, errors.Wrap(err, "http request failed")
+		}
+		defer resp.Body.Close()
+
+		if resp.StatusCode == http.StatusOK {
+			body, err := ioutil.ReadAll(resp.Body)
+			if err != nil {
+				return nil, fmt.Errorf("failed to read response body: %w", err)
+			}
+
+			if err := json.Unmarshal(body, &result); err != nil {
+				return nil, fmt.Errorf("failed to unmarshal JSON: %w", err)
+			}
+
+			return result, nil
+		} else if isRetryableStatusCode(resp.StatusCode) {
+			retries++
+			backoffDuration := initialBackoff * time.Duration(1<<retries)
+			if retries <= maxRetries {
+				fmt.Printf("Received retryable status %d. Retrying in %v...\n", resp.StatusCode, backoffDuration)
+				time.Sleep(backoffDuration)
+			} else {
+				return nil, fmt.Errorf("Maximum retries reached after receiving status %d", resp.StatusCode)
+			}
+		} else {
+			return nil, fmt.Errorf("API request failed with status %d", resp.StatusCode)
+		}
 	}
 
-	return result, nil
-}
-
-func setHeaders(req *http.Request, args map[string]interface{}) {
-	for key, value := range args {
-		strValue, ok := value.(string)
-		if !ok {
-			fmt.Printf("Skipping non-string value for header %s\n", key)
-			continue
-		}
-
-		req.Header.Set(key, strValue)
-	}
-}
-
-func setAuthHeaders(req *http.Request, authType string, args map[string]interface{}) error {
-	switch authType {
-	case "basic":
-		username, ok := args["username"].(string)
-		if !ok {
-			return fmt.Errorf("missing or invalid username")
-		}
-		password, ok := args["password"].(string)
-		if !ok {
-			return fmt.Errorf("missing or invalid password")
-		}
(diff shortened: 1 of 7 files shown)


utils/apiclient/client.go
```diff
@@ -0,0 +1,85 @@
...
+	"net/http"
+	"net/url"
+	"time"
+
+	"github.com/pkg/errors"
+)
+
+const (
+	maxRetries     = 5
+	initialBackoff = 1 * time.Second
```
chowbao, line 16: Should these be configurable?

utils/apiclient/client.go
```diff
@@ -0,0 +1,85 @@
...
+)
+
+func isRetryableStatusCode(statusCode int) bool {
+	return statusCode == http.StatusTooManyRequests || statusCode == http.StatusServiceUnavailable
+}
+
+func (c *APIClient) GetURL(endpoint string, qstr url.Values) string {
```
chowbao, line 23: nit: Is `qstr` a misnomer? Because it isn't a string? It's probably fine because it's a "short lived" variable but might be worth renaming if you can think of a better name

utils/apiclient/client.go
```diff
@@ -0,0 +1,85 @@
...
+func (c *APIClient) CallAPI(reqParams RequestParams) (interface{}, error) {
+	if reqParams.QueryParams == nil {
+		reqParams.QueryParams = url.Values{}
+	}
+
+	if reqParams.Headers == nil {
+		reqParams.Headers = map[string]interface{}{}
+	}
+
+	url := c.GetURL(reqParams.Endpoint, reqParams.QueryParams)
```
chowbao, line 36: Would be good to check/error on `Endpoint` because it is required in `RequestParams`

utils/apiclient/client.go
amishas157: Yes, configurable makes sense 👍

utils/apiclient/client.go
```diff
@@ -0,0 +1,85 @@
...
+		defer resp.Body.Close()
+
+		if resp.StatusCode == http.StatusOK {
+			body, err := ioutil.ReadAll(resp.Body)
+			if err != nil {
+				return nil, fmt.Errorf("failed to read response body: %w", err)
+			}
+
+			if err := json.Unmarshal(body, &result); err != nil {
+				return nil, fmt.Errorf("failed to unmarshal JSON: %w", err)
```
chowbao, line 66: nit: I think these two errors can use the same `errors.Wrap()` as above instead of `fmt.Errorf()` and passing it the err

utils/apiclient/client.go
```diff
@@ -0,0 +1,85 @@
...
+			}
+
+			return result, nil
+		} else if isRetryableStatusCode(resp.StatusCode) {
+			retries++
+			backoffDuration := initialBackoff * time.Duration(1<<retries)
+			if retries <= maxRetries {
+				fmt.Printf("Received retryable status %d. Retrying in %v...\n", resp.StatusCode, backoffDuration)
```
chowbao, line 74: There is a stellar/go `log` package to write logs instead of fmt.Printf
https://github.com/stellar/go/blob/master/support/log/main.go

An [example](https://github.com/stellar/go/blob/master/ingest/ledgerbackend/buffered_meta_pipe_reader.go#L121)

utils/apiclient/client_test.go
```diff
@@ -0,0 +1,108 @@
...
+	assert.Equal(t, "https://stellar.org/federation?acct=2382376&federation_type=bank_account&swift=BOPBPHMM&type=forward", furl)
+}
+
+func TestCallAPI(t *testing.T) {
+	testCases := []struct {
+		name          string
+		mockResponses []httptest.ResponseData
+		expected      interface{}
+		expectedError string
+	}{
```
chowbao, line 32: nit: you can pull out the testCase into its own struct
```
type testCase struct{
    name          string
    mockResponses []httptest.ResponseData
    expected      interface{}
    expectedError string
}
```
then use the struct in your list of testCases
```
testCases := []testCase{
  ...
}
```

This just makes it easier to reuse `testCase` in case there needs to be another unit test added that would use the same testCase format.

utils/apiclient/request.go
```diff
@@ -0,0 +1,61 @@
...
+		return nil, errors.Wrap(err, "http GET request creation failed")
+	}
+	return req, nil
+}
+
+func SetHeaders(req *http.Request, args map[string]interface{}) {
+	for key, value := range args {
+		strValue, ok := value.(string)
+		if !ok {
+			fmt.Printf("Skipping non-string value for header %s\n", key)
```
chowbao, line 23: Same comment about using the stellar/go `log` "github.com/stellar/go/support/log"

utils/apiclient/client.go
amishas157: Do you mean following change?
```suggestion
				return nil, fmt.Errorf("failed to read response body: %w", err)
			}

			if err := json.Unmarshal(body, &result); err != nil {
				return nil, err
```

utils/apiclient/request.go
```diff
@@ -0,0 +1,61 @@
...
+		if !ok {
+			return fmt.Errorf("missing or invalid API key")
+		}
+		SetHeaders(req, map[string]interface{}{
+			"Authorization": apiKey,
+		})
+
+	default:
+		return fmt.Errorf("unsupported auth type: %s", authType)
+	}
```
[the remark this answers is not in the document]
amishas157: What do you mean? My understanding is that if authType is not api_key or basic, it should raise an error for unsupported auth type

utils/apiclient/client.go
amishas157: Addressed in https://github.com/stellar/go/pull/5534/commits/f5792cd3a40b6d35cab7367424685caa69fa6931

utils/apiclient/client.go
amishas157: Addressed in https://github.com/stellar/go/pull/5534/commits/f5792cd3a40b6d35cab7367424685caa69fa6931

utils/apiclient/client.go
amishas157: Addressed in https://github.com/stellar/go/pull/5534/commits/f5792cd3a40b6d35cab7367424685caa69fa6931

utils/apiclient/client.go
amishas157: Addressed in https://github.com/stellar/go/pull/5534/commits/f5792cd3a40b6d35cab7367424685caa69fa6931

utils/apiclient/client_test.go
amishas157: Addressed in https://github.com/stellar/go/pull/5534/commits/f5792cd3a40b6d35cab7367424685caa69fa6931

utils/apiclient/client_test.go
```diff
@@ -0,0 +1,108 @@
...
+			if tc.expectedError != "" {
+				if err == nil {
+					t.Fatalf("expected error %q, got nil", tc.expectedError)
+				}
+				if err.Error() != tc.expectedError {
+					t.Fatalf("expected error %q, got %q", tc.expectedError, err.Error())
+				}
+			} else if err != nil {
+				t.Fatal(err)
+			}
```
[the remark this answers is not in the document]
amishas157: Addressed in https://github.com/stellar/go/pull/5534/commits/f5792cd3a40b6d35cab7367424685caa69fa6931

utils/apiclient/request.go
amishas157: Addressed in https://github.com/stellar/go/pull/5534/commits/f5792cd3a40b6d35cab7367424685caa69fa6931

utils/apiclient/request.go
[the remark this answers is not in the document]
chowbao: Ah you're right. That's my misunderstanding

utils/apiclient/client.go
amishas157: Addressed in https://github.com/stellar/go/pull/5534/commits/f4588532c1994e8fa7f54108c5a4c721594238c4

utils/apiclient/client.go
chowbao: Sorry I meant that instead of doing `fmt.Errorf("failed to read response body: %w", err)` you could just do `errors.Wrap(err, "supplemental error text")`

But it looks like I got the recommendation wrong. It seems like we want to avoid using `errors.Wrap()` in favor of using `fmt.Errorf()`

Two notable things from this stack overflow question:
https://stackoverflow.com/questions/61933650/whats-the-difference-between-errors-wrapf-errors-errorf-and-fmt-errorf
"*The github.com/pkg/errors package is no longer actively maintained. The repo has been archived, and is no longer accepting updates, but remains available for use."
"One difference between Wrapf(err, "") and Errorf("%w", err) is that wrapf would return nil if err is nil, but errorf would return an error regardless."

amishas157: Hrm. This is interesting. It makes sense to use `fmt.errorf` to get the error regardless.

amishas157: Addressed in https://github.com/stellar/go/pull/5534/commits/66f236c8f1c2fb2a72c43ae20a315c44b8d15291

utils/apiclient/client_test.go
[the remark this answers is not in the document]
Shaptic: You should use `require` over `assert` unless you want the test to continue if the assertion fails (require will stop test, assert keeps going)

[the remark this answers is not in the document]
amishas157: I think it's okay to continue to check other asserts (and report at the end) in this case.

chowbao (approved):
LGTM

Issue #5143, opened 2023-12-17
Strict-receive pathfinder returns invalid paths


mikhail-dvorkin:
Consider finding a strict-receive path to swap AQUA to 1 XLM. Here's the link to Horizon:

https://horizon.stellar.org/paths/strict-receive?source_assets=AQUA%3AGBNZILSTVQZ4R7IKQDGHYGY2QXL5QOFJYQMXPKWRRM5PAV7Y4M67AQUA&destination_asset_type=native&destination_amount=1

For many weeks it shows a path through assets SGB and IBMWorldWire, proposing a price of < 150:

```
{
        "source_asset_type": "credit_alphanum4",
        "source_asset_code": "AQUA",
        "source_asset_issuer": "GBNZILSTVQZ4R7IKQDGHYGY2QXL5QOFJYQMXPKWRRM5PAV7Y4M67AQUA",
        "source_amount": "125.1023169",
        "destination_asset_type": "native",
        "destination_amount": "1.0000000",
        "path": [
          {
            "asset_type": "credit_alphanum4",
            "asset_code": "SGB",
            "asset_issuer": "GAME5YNKHIPRSMN2BI3HLS3OYD4NE6FQ2WKUUJPYYM6GU4MRDUBPHN6B"
          },
          {
            "asset_type": "credit_alphanum12",
            "asset_code": "IBMWorldWire",
            "asset_issuer": "GB6KHHHS52HXXFRG5V5QKF2D2DMYHW6GICHPH3FJGJS5MYIPVXDL74TY"
          }
        ]
}
```

However, when I try to use this path in a pathPaymentStrictReceive operation, I constantly receive
`{"transaction":"tx_failed","operations":["op_too_few_offers"]}`

The horizon's alleged path always shows a price < 150, and in my operation I set 175 as sendMax, so this is not a small fluctuation issue.

### What version are you using?

"horizon_version": "2.27.0-208a73f6165f59c4eeacf672b6d24659fe54d256",
"core_version": "stellar-core 20.0.0 (6177299100b114aa108584053414371f38aebf53)"

### To reproduce

1. Find the best strict-receive path to swap AQUA:GBNZILSTVQZ4R7IKQDGHYGY2QXL5QOFJYQMXPKWRRM5PAV7Y4M67AQUA to 1 XLM.
2. Take the path through SGB and IBMWorldWire, it proposes a price of < 150.
3. Try to execute strict-receive transaction over that path with sendMax = 175.

PR #5541, opened 2024-12-03
exp/orderbook: Fix bug in CalculatePoolExpectation()

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Fixes https://github.com/stellar/go/issues/5143

This commit fixes three inconsistencies with how liquidity pool payouts / expectations are calculated in core vs horizon:

1. When calculating how much we need to deposit into a pool in order to receive a desired payout we need to consider that the calculated deposit may overflow the reserves, see https://github.com/stellar/stellar-core/blob/fdd833d57c86cfe0c5057da5b2319953ab841de0/src/transactions/OfferExchange.cpp#L1361-L1363
2. We should rule out trades that will result in a payout of 0, see https://github.com/stellar/stellar-core/blob/fdd833d57c86cfe0c5057da5b2319953ab841de0/src/transactions/OfferExchange.cpp#L1325-L1327
3. Fees that are greater than or equal to 100% are illegal, see https://github.com/stellar/stellar-core/blob/fdd833d57c86cfe0c5057da5b2319953ab841de0/src/transactions/OfferExchange.cpp#L1254-L1259

(2) and (3) do not affect the output of the horizon path finding endpoint because those corner cases are excluded from consideration in the path finding algorithm.

However, (1) results in horizon returning incorrect paths which attempt to trade against liquidity pools in a way that would overflow the liquidity pool reserves.  When users try to submit path payments using these incorrect paths, core is not able to execute the trades and that results in a [`op_too_few_offers`](https://developers.stellar.org/docs/data/horizon/api-reference/errors/result-codes/operation-specific/path-payment-strict-receive) error.

### Known limitations

[N/A]

### DIFF
--- a/exp/orderbook/pools.go
+++ b/exp/orderbook/pools.go
@@ -4,6 +4,7 @@ import (
 	"math"
 
 	"github.com/holiman/uint256"
+
 	"github.com/stellar/go/support/errors"
 	"github.com/stellar/go/xdr"
 )
@@ -20,6 +21,7 @@ import (
 const (
 	tradeTypeDeposit     = iota // deposit into pool, what's the payout?
 	tradeTypeExpectation = iota // expect payout, what to deposit?
+	maxBasisPoints       = 10_000
 )
 
 var (
@@ -91,6 +93,9 @@ func makeTrade(
 //
 // It returns false if the calculation overflows.
 func CalculatePoolPayout(reserveA, reserveB, received xdr.Int64, feeBips xdr.Int32, calculateRoundingSlippage bool) (xdr.Int64, xdr.Int64, bool) {
+	if feeBips < 0 || feeBips >= maxBasisPoints {
+		return 0, 0, false
+	}
 	X, Y := uint256.NewInt(uint64(reserveA)), uint256.NewInt(uint64(reserveB))
 	F, x := uint256.NewInt(uint64(feeBips)), uint256.NewInt(uint64(received))
 
@@ -101,7 +106,7 @@ func CalculatePoolPayout(reserveA, reserveB, received xdr.Int64, feeBips xdr.Int
 
 	// We do all of the math with 4 extra decimal places of precision, so it's
 	// all upscaled by this value.
-	maxBips := uint256.NewInt(10000)
+	maxBips := uint256.NewInt(maxBasisPoints)
 	f := new(uint256.Int).Sub(maxBips, F) // upscaled 1 - F
 
 	// right half: X + (1 - F)x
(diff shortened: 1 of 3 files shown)


exp/orderbook/pools.go
```diff
@@ -91,6 +93,9 @@ func makeTrade(
 //
 // It returns false if the calculation overflows.
 func CalculatePoolPayout(reserveA, reserveB, received xdr.Int64, feeBips xdr.Int32, calculateRoundingSlippage bool) (xdr.Int64, xdr.Int64, bool) {
+	if feeBips < 0 || feeBips >= maxBasisPoints {
+		return 0, 0, false
+	}
```
Shaptic, line 98: Iirc `feeBips` was just a way to make this configurable for future AMMs, but the value will always be 0.3% per [the CAP](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0038.md#future-work-support-fees-other-than-03), so this check seems unnecessary as the code stands today? But it's good defensive programming for the future so we should keep it.

exp/orderbook/pools.go
```diff
@@ -231,7 +239,11 @@ func CalculatePoolExpectation(
 	}
 
 	val := xdr.Int64(result.Uint64())
-	ok = ok && result.IsUint64() && val >= 0
+	ok = ok &&
+		result.IsUint64() &&
+		val >= 0 &&
+		// check that the calculated deposit would not overflow the reserve
+		val <= math.MaxInt64-reserveA
```
Shaptic, line 246: Nice sleuthing; bet we didn't expect pools to ever hit `MaxInt64`...

exp/orderbook/pools.go
```diff
@@ -153,7 +158,7 @@ func CalculatePoolPayout(reserveA, reserveB, received xdr.Int64, feeBips xdr.Int
 	}
 
 	val := xdr.Int64(result.Uint64())
-	ok = ok && result.IsUint64() && val >= 0
+	ok = ok && result.IsUint64() && val > 0
```
Shaptic, line 161: This isn't an error, is it? It's just saying that you'd get nothing in return if you deposited `received` :thinking: 

I'm fine with doing it this way if it makes the code cleaner so we don't have to special-case `val == 0` at the caller level, but maybe a comment here clarifying, then?

Shaptic (approved):
LGTM! Just a coupla comments below :arrow_down: really thorough investigation and fix, though :clap:

exp/orderbook/pools.go
Shaptic: Ah, I just saw the code you linked in the Core repo: if they treat it as an error we should too!

PR #5546, opened 2024-12-10
clients/stellarcore: Support more parameters from the upgrades endpoint

2opremio:
See https://github.com/stellar/stellar-core/blob/0241e79f74dc017f20e190abd3825873222c5ca5/src/main/CommandHandler.cpp#L568 for alll the supported parameters.

### DIFF
--- a/clients/stellarcore/client.go
+++ b/clients/stellarcore/client.go
@@ -56,12 +56,55 @@ func drainReponse(hresp *http.Response, close bool, err *error) (outerror error)
 }
 
 // Upgrade upgrades the protocol version running on the stellar core instance
+//
+// Deprecated: use UpgradeProtocol instead
 func (c *Client) Upgrade(ctx context.Context, version int) (err error) {
-	queryParams := url.Values{}
-	queryParams.Add("mode", "set")
-	queryParams.Add("upgradetime", "1970-01-01T00:00:00Z")
-	queryParams.Add("protocolversion", strconv.Itoa(version))
+	return c.UpgradeProtocol(ctx, version, time.Unix(int64(0), 0))
+}
 
+// UpgradeProtocol upgrades the protocol version running on the stellar core instance
+func (c *Client) UpgradeProtocol(ctx context.Context, protocolVersion int, at time.Time) (err error) {
+	queryParams := url.Values{}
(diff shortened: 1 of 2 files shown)


clients/stellarcore/client.go
```diff
@@ -56,12 +58,105 @@ func drainReponse(hresp *http.Response, close bool, err *error) (outerror error)
...
+	}
+	cmd := exec.Command(corePath, "get-settings-upgrade-txs",
+		config.SigningKey.Address(),
+		strconv.FormatUint(uint64(config.BaseSeqNum), 10),
+		config.NetworkPassphrase,
+		"--xdr", upgradeConfigB64,
+		"--signtxs")
+	inputStr := config.SigningKey.Seed()
+	cmd.Stdin = strings.NewReader(inputStr)
+	out, err := cmd.CombinedOutput()
```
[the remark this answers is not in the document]
2opremio: Makes sense. I will fix it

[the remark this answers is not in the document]
2opremio: Looking at the code, it seems that `Output` also captures the sdterr and appends it any error it finds. So, simply replacing it was good enough.

clients/stellarcore/client.go
```diff
@@ -56,12 +58,105 @@ func drainReponse(hresp *http.Response, close bool, err *error) (outerror error)
...
+	queryParams := url.Values{}
+	queryParams.Add("configupgradesetkey", keyB64)
+	return c.setUpgradesAt(ctx, at, queryParams)
+}
+
+type GenSorobanConfig struct {
+	BaseSeqNum        uint32
+	NetworkPassphrase string
+	// TODO: Should we ask core to sign the tx or should we do it ourselves?
+	SigningKey *keypair.Full
```
2opremio, line 89: @tamirms thoughts about this? Should we have Core sign the transactions or would we be better off doing ti ourselves?

tamirms: I think either way is ok. We are using this for load testing purposes so the signing key is not sensitive

2opremio: Alright, I will remove the TODO then

tamirms (approved):

PR #5563, opened 2025-01-02
ingest/ledgerbackend: fix to ensure that the ledger buffer properly queues the last batch of ledgers (LCM) within the specified range

urvisavla:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
The fix updates the boundary check in bounded mode to ensure the `BufferedStorageBackend`/`LedgerBuffer` queues ledgers up to the end *boundary* ledger of the specified range. This fixes an issue where the final batch of ledgers is skipped in scenarios where multiple ledgers are stored per file and the `from` ledger does not align with the *start boundary* of the stored file.

### Why

This issue occurs only in bounded mode where multiple ledgers are stored per file.

For example:  
When`ledgers_per_file` is set to 4, each file contains 4 consecutive ledgers, resulting in files like:  
`0–3.xdr`, `4–7.xdr`, `8–11.xdr`, `12–15.xdr`, `16–19.xdr` and so on.

Suppose we need to fetch ledgers `6 to 17` using the `BufferedStorageBackend`, so the `from` ledger is `6` and `to` ledger as `17`:  

1. The `LedgerBuffer` starts by downloading the file containing ledger `6` (`4–7.xdr`).  
2. It increments the next ledger to fetch by 4 (equal to `ledgers_per_file`), fetching the batch containing ledger `10` and `14` resulting in the download of files `8–11.xdr` and `12–15.xdr`.  
3. Now when the next ledger to fetch is incremented to `18`, it exceeds the `to` ledger (`17`).   
At this point, the buffer stops queuing, skipping file `16–19.xdr` even though it contains ledger `17` which is part of the requested range.  

When `GetLedger` is called, the `BufferedStorageBackend` first checks if the current LCM object contains the requested ledger. If not, it tries to fetch the next LCM object from the `LedgerBuffer`. Since `16–19.xdr` was never queued due to the boundary check, `GetLedger` blocks indefinitely when called for ledger `16` or later, waiting for a ledger that is never queued.

This issue does not trigger an error log because it is not treated as a failure which causing the system to remain in wait state.

### Known limitations
N/A

### DIFF
--- a/ingest/ledgerbackend/ledger_buffer.go
+++ b/ingest/ledgerbackend/ledger_buffer.go
@@ -96,8 +96,8 @@ func (bsb *BufferedStorageBackend) newLedgerBuffer(ledgerRange Range) (*ledgerBu
 }
 
 func (lb *ledgerBuffer) pushTaskQueue() {
-	// In bounded mode, don't queue past the end ledger
-	if lb.nextTaskLedger > lb.ledgerRange.to && lb.ledgerRange.bounded {
+	// In bounded mode, don't queue past the end boundary ledger for the specified range.
+	if lb.ledgerRange.bounded && lb.nextTaskLedger > lb.dataStore.GetSchema().GetSequenceNumberEndBoundary(lb.ledgerRange.to) {
 		return
 	}
 	lb.taskQueue <- lb.nextTaskLedger
--- a/ingest/CHANGELOG.md
+++ b/ingest/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable changes to this project will be documented in this file. This projec
 
 ## Pending
 
+### Bug Fixes
+* Update the boundary check in `BufferedStorageBackend` to queue ledgers up to the end boundary, resolving skipped final batch when the `from` ledger doesn't align with file boundary [5563](https://github.com/stellar/go/pull/5563).
+
 ### New Features
 * Create new package `ingest/cdp` for new components which will assist towards writing data transformation pipelines as part of [Composable Data Platform](https://stellar.org/blog/developers/composable-data-platform). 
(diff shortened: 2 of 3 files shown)


ingest/ledgerbackend/ledger_buffer.go
```diff
@@ -96,8 +96,8 @@ func (bsb *BufferedStorageBackend) newLedgerBuffer(ledgerRange Range) (*ledgerBu
 }
 
 func (lb *ledgerBuffer) pushTaskQueue() {
-	// In bounded mode, don't queue past the end ledger
-	if lb.nextTaskLedger > lb.ledgerRange.to && lb.ledgerRange.bounded {
+	// In bounded mode, don't queue past the end boundary ledger for the specified range.
+	if lb.ledgerRange.bounded && lb.nextTaskLedger > lb.dataStore.GetSchema().GetSequenceNumberEndBoundary(lb.ledgerRange.to) {
```
urvisavla, line 100: Alternatively, this issue could be fixed by adjusting the `to` ledger to align with the end ledger boundary in [BufferedStorageBackend::PrepareRange](https://github.com/stellar/go/blob/d5767df5b21a1c34a5cb05b158d9b0962c94e7bf/ingest/ledgerbackend/buffered_storage_backend.go#L170-L187) but not sure if that  has any advantage over this solution.

tamirms: another alternative is that you could initialize `nextTaskLedger`  to `dataStore. GetSchema().GetSequenceNumberStartBoundary(ledgerRange.from)`. This would align `nextTaskLedger` to the ledger boundary. I think that would be better than modifying `PrepareRange()`.

But I think the current solution you have is good and perhaps more robust because it does not assume any alignment on `nextTaskLedger`

tamirms (approved):

tamirms:
Could you update the changelog as well?

ingest/ledgerbackend/ledger_buffer.go
urvisavla: I attempted to modify `nextLedger` in the LedgerBuffer without adjusting the range but that required changes to `nextExpectedSequence()` which assumes the ledger specified in `GetLedger` always matches `nextLedger` and that assumption would no longer hold with that approach.

Issue #5549, opened 2024-12-10
XDRill: LedgerCloseMeta level functions


chowbao:
The LedgerCloseMeta level functions is similar to the [ledger.go](https://github.com/stellar/go/pull/5501/files#diff-9dd3fcb834482ae6c797b5df230f53df362b99dc1b9431ae79c4da609e0a13cd) example code from the design doc

* Create new struct for the LedgerCloseMeta level functions
  * Don't wrap xdr.LedgerCloseMeta
  ```
  type Ledgers struct {
    ledger xdr.LedgerCloseMeta
  }
  ```
* Define functions needed for the `history_ledgers` transforms in [stellar-etl](https://github.com/stellar/stellar-etl/blob/master/internal/transform/ledger.go)

PR #5568, opened 2025-01-08
xdrill for ledgerCloseMeta

chowbao:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

[github issue](https://github.com/stellar/go/issues/5549)

xdrill functions for LedgerCloseMeta 

### Why

First commit for xdrill with the low level helper functions from the processors/transforms from stellar-etl for `history_ledgers` found [here](https://github.com/stellar/stellar-etl/blob/master/internal/transform/ledger.go)

### Known limitations

* Refactor of the processor/transforms to be done in separate ticket/pr


exp/xdrill/ledger_test.go
chowbao, line 1: How exhaustive do we want the testing to be?

exp/xdrill/ledger.go
```diff
@@ -0,0 +1,208 @@
...
+	"fmt"
+	"time"
+
+	"github.com/stellar/go/exp/xdrill/utils"
+	"github.com/stellar/go/xdr"
+)
+
+type Ledger struct {
+	ledger *xdr.LedgerCloseMeta
+}
```
tamirms, line 14: this struct definition would not work outside of the xdrill package. Although the `Ledger` struct is exported, the `ledger` field remains private to external packages. So, if you would not be able to construct a `Ledger` instance outside of the xdrill package because you would not be able to reference the `ledger` field.

To fix this issue you can make the `ledger` field public:

```go
type Ledger struct {
	Ledger *xdr.LedgerCloseMeta
}
```

Alternatively, you could create a type definition:

```golang
type Ledger xdr.LedgerCloseMeta

func (l Ledger) Sequence() uint32 {
	return uint32(l.ledger.LedgerHeaderHistoryEntry().Header.LedgerSeq)
}
```

Then to call the helper function you would cast your `xdr.LedgerCloseMeta` value into an `xdrill.Ledger` value so you could call the helper function:

```golang

package app

import (
	"github.com/stellar/go/exp/xdrill"
	"github.com/stellar/go/xdr"
)

func doSomething(ledger xdr.LedgerCloseMeta) {
  l := xrill.Ledger(ledger)
  seq := l.Sequence()
  ...
}
```

Or you could make the helper functions take the ledger parameter as the first argument:

```go

func LedgerSequence(ledger xdr.LedgerCloseMeta) uint32 {
	return uint32(ledger.LedgerHeaderHistoryEntry().Header.LedgerSeq)
}
 ```

IMO the cleanest API would be define helper functions like these in the xdr package directly on the `LedgerCloseMeta` type. I think I missed the discussion about this decision. What was the motivation for defining a type equivalent to xdr.LedgerCloseMeta in the xdrill package?

Comment (chowbao) [reply exp/xdrill/ledger.go:14]: Ah oops yeah it should be public

Comment (chowbao) [reply exp/xdrill/ledger.go]: > What was the motivation for defining a type equivalent to xdr.LedgerCloseMeta in the xdrill package?

Talked in slack. tl;dr - it's nicer for users as a separate package because they would be given a subset of only XDRill functions instead of all the functions/fields within xdr and ingest

Comment (chowbao) [reply exp/xdrill/ledger.go]: Updated functions to accept lcm as a param instead of the Ledger struct
d45b0fd0b35943076ea44efa733dd95c65d0a61c

File: exp/xdrill/ledger/ledger.go [hunk 2]
```diff
@@ -0,0 +1,204 @@
...
+
+func MaxTxSetSize(l xdr.LedgerCloseMeta) uint32 {
+	return uint32(l.LedgerHeaderHistoryEntry().Header.MaxTxSetSize)
+}
+
+func LedgerVersion(l xdr.LedgerCloseMeta) uint32 {
+	return uint32(l.LedgerHeaderHistoryEntry().Header.LedgerVersion)
+}
+
+func SorobanFeeWrite1Kb(l xdr.LedgerCloseMeta) *int64 {
```
tamirms, line 60: I think returning `(int64, bool)` is easier to work with compared to returning a pointer.

exp/xdrill/ledger/ledger.go
```diff
@@ -0,0 +1,204 @@
...
+	if !ok {
+		return nil
+	}
+
+	result := uint64(lcmV1.TotalByteSizeOfBucketList)
+
+	return &result
+}
+
+func NodeID(l xdr.LedgerCloseMeta) *string {
```
tamirms, line 87: similar to the other comment, I think returning a bool value to indicate whether the node id is present or not is better than returning a pointer

exp/xdrill/ledger/ledger.go
```diff
@@ -0,0 +1,204 @@
...
+// Add docstring to larger, more complicated functions
+func TransactionCounts(l xdr.LedgerCloseMeta) (successTxCount, failedTxCount int32, ok bool) {
+	var results []xdr.TransactionResultMeta
+
+	transactions := getTransactionSet(l)
+	results = l.TxProcessing()
+	txCount := len(transactions)
+	if txCount != len(results) {
+		return 0, 0, false
+	}
```
tamirms, line 121: in which scenario can there be a mismatch between `txCount` and `len(results)`?

exp/xdrill/ledger/ledger.go
```diff
@@ -0,0 +1,204 @@
...
+
+			operationCount += int32(len(operationResults))
+		}
+
+	}
+
+	return operationCount, txSetOperationCount, true
+}
+
+func getTransactionSet(l xdr.LedgerCloseMeta) (transactionProcessing []xdr.TransactionEnvelope) {
```
tamirms, line 166: there is a TransactionEnvelopes() function defined on xdr.LedgerCloseMeta which returns all the transaction envelopes in a ledger

exp/xdrill/utils/utils.go
```diff
@@ -0,0 +1,128 @@
...
+	// TransactionShift is the number of bits to shift an int64 to
+	// target the transaction component
+	TransactionShift = 12
+	// OperationShift is the number of bits to shift an int64 to target
+	// the operation component
+	OperationShift = 0
+)
+
+// New creates a new total order ID
+func NewID(ledger int32, tx int32, op int32) *ID {
```
tamirms, line 46: there is already a toid package which has this code

exp/xdrill/utils/utils.go
```diff
@@ -0,0 +1,128 @@
+package utils
```
tamirms, line 1: utils packages are generally discouraged in go:

https://google.github.io/styleguide/go/decisions.html#package-names

chowbao: Ah ok. What is the correct way to organize shared "utils"/"common" functions?

Would just renaming the package to `xdrillutils` be good?
Or should I split it out into more descriptive packages like `xdrillschemas`, `xdrillconversions`?

exp/xdrill/ledger/ledger.go
chowbao: I made these pointers because nil/null is a useful value for representing data in database tables. Like in hubble/bigquery there is a distinction between int64 null VS 0

I see [this article](https://medium.com/@dual/ways-to-handle-sql-null-values-in-go-e237f31b82fc)
Would it be better practice to use `github.com/guregu/null` or `sql.Null*` to represent nulls as a value?

Or is it still better to have the `(result int64, ok bool)` and handle the `null` in my application (stellar-etl) instead?

exp/xdrill/utils/utils.go
tamirms: I think `xdrillutils` is basically equivalent to `utils`. In general, package names should be related to what the code actually does and if there is no unifying theme for all the code in the package then it could be that the code should be separated into multiple packages

for example, `CreateSampleTx()` is used to generate to testing fixtures. you could argue that it doesn't need to even be exported outside of the testing files which use it.

exp/xdrill/ledger/ledger.go
tamirms: yeah, I would suggest removing it and only iterating over results

exp/xdrill/ledger/ledger.go
tamirms: > Or is it still better to have the (result int64, ok bool) and handle the null in my application (stellar-etl) instead?

yeah, I think that is a cleaner option because we don't want to couple the implementation to SQL or some other database. It's better if xdrill remains agnostic to whichever database is used to store the data extracted out of the XDR objects

exp/xdrill/utils/utils.go
chowbao: Yeah makes sense. I'll try splitting them up into more relevantly named packages. Maybe ChatGPT can offer some suggestions lol

exp/xdrill/utils/utils.go
```diff
@@ -0,0 +1,128 @@
...
+		},
+	)
+	PanicOnError(err)
+
+	env := tx.ToXDR()
+	return env
+}
+
+// PanicOnError is a function that panics if the provided error is not nil
+func PanicOnError(err error) {
```
tamirms, line 124: I think this function can remain private. it isn't used anywhere except in CreateSampleTx()

chowbao: I'll take a look. I believe it is used in future functions and depending on how I split up and rename the utils package it might need to be public

exp/xdrill/ledger/ledger.go
chowbao: Yup sounds good to me

exp/xdrill/ledger.go
chowbao: Moved xdrill ledger functions to go/ingest/ledger
40d505759d70f72b5937f53447da774105c3bff6

exp/xdrill/ledger/ledger.go
chowbao: Updated to use (result, bool)
40d505759d70f72b5937f53447da774105c3bff6

exp/xdrill/ledger/ledger.go
chowbao: Updated to use (result, bool)
40d505759d70f72b5937f53447da774105c3bff6

exp/xdrill/ledger/ledger.go
chowbao: * Removed the useless check
* Use range on results
40d505759d70f72b5937f53447da774105c3bff6

exp/xdrill/ledger/ledger.go
chowbao: Updated to use existing xdr function
40d505759d70f72b5937f53447da774105c3bff6

exp/xdrill/utils/utils.go
chowbao: Updated to use existing toid
40d505759d70f72b5937f53447da774105c3bff6

xdr/node_id.go
```diff
@@ -0,0 +1,22 @@
...
+		}
+		raw := make([]byte, 32)
+		copy(raw, ed[:])
+		encodedAddress, err := strkey.Encode(strkey.VersionByteAccountID, raw)
+		if err != nil {
+			return "", false
+		}
+		return encodedAddress, true
+	default:
+		return "", false
```
sydneynotthecity, line 20: is there a reason you don't return an error for unknown id types?

ingest/ledger/ledger.go
```diff
@@ -0,0 +1,146 @@
...
+			operationResults, ok := result.Result.OperationResults()
+			if !ok {
+				return 0, 0, false
+			}
+
+			operationCount += int32(len(operationResults))
+		}
+	}
+
+	return operationCount, txSetOperationCount, true
```
sydneynotthecity, line 145: `operationCount` vs `txSetOperationCount` is confusing (I also find it confusing in stellar-etl). Can we update to `successfulOperationCount` or something more descriptive that indicates what operations are counted?

xdr/ledger_close_meta.go
```diff
@@ -156,3 +156,15 @@ func (l LedgerCloseMeta) EvictedPersistentLedgerEntries() ([]LedgerEntry, error)
...
+
+// TxProcessing returns the TransactionResultMeta in this ledger
+func (l LedgerCloseMeta) TxProcessing() []TransactionResultMeta {
+	switch l.V {
+	case 0:
+		return l.MustV0().TxProcessing
+	case 1:
+		return l.MustV1().TxProcessing
+	default:
+		panic(fmt.Sprintf("Unsupported LedgerCloseMeta.V: %d", l.V))
```
sydneynotthecity, line 168: This is the only function where you panic if the XDR is not parseable. imo, we should be consistent with either using `Get<Thing>()` vs `Must()`. thoughts?

sydneynotthecity (commented):
Overall looks like a good start, just had a couple comments about consistency within the package to ensure that we're not carrying over oddities from stellar-etl that we should fix now.

ingest/ledger/ledger.go
```diff
@@ -0,0 +1,146 @@
...
+// Add docstring to larger, more complicated functions
+func TransactionCounts(l xdr.LedgerCloseMeta) (successTxCount, failedTxCount int32, ok bool) {
+	var results []xdr.TransactionResultMeta
+
+	transactions := l.TransactionEnvelopes()
+	results = l.TxProcessing()
+	txCount := len(transactions)
+	if txCount != len(results) {
+		return 0, 0, false
+	}
```
tamirms, line 109: I don't know that this case can ever happen. I think we should panic if there is a length mismatch

ingest/ledger/ledger.go
```diff
@@ -0,0 +1,146 @@
...
+		operations := transactions[i].Operations()
+		numberOfOps := int32(len(operations))
+		txSetOperationCount += numberOfOps
+
+		// for successful transactions, the operation count is based on the operations results slice
+		if result.Result.Successful() {
+			operationResults, ok := result.Result.OperationResults()
+			if !ok {
+				return 0, 0, false
+			}
```
tamirms, line 139: if the operation result is successful there will always be a results slice. so there is no need to return the ok boolean

xdr/ledger_close_meta.go
chowbao: Hmm that's true. I'll make ok, err, and panic more consistent

ingest/ledger/ledger.go
```diff
@@ -0,0 +1,146 @@
...
+)
+
+func Sequence(l xdr.LedgerCloseMeta) uint32 {
+	return uint32(l.LedgerHeaderHistoryEntry().Header.LedgerSeq)
+}
+
+func ID(l xdr.LedgerCloseMeta) int64 {
+	return toid.New(int32(l.LedgerSequence()), 0, 0).ToInt64()
+}
+
```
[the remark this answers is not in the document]
chowbao: Removed in be020c9be95a0f4d7688409139b86b2fc3390642

xdr/node_id.go
chowbao: Updated to return error in be020c9be95a0f4d7688409139b86b2fc3390642

ingest/ledger/ledger.go
chowbao: Renamed output to successfulOperationCount and totalOperationCount in be020c9be95a0f4d7688409139b86b2fc3390642

ingest/ledger/ledger.go
chowbao: Updated in be020c9be95a0f4d7688409139b86b2fc3390642

ingest/ledger/ledger.go
chowbao: Updated in be020c9be95a0f4d7688409139b86b2fc3390642

Removed the ok bool from the function returns. Will now panic in this case instead

ingest/ledger/ledger.go
```diff
@@ -0,0 +1,145 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	for i, result := range results {
+		operations := uint32(len(transactions[i].Operations()))
+		totalOperationCount += operations
+
+		// for successful transactions, the operation count is based on the operations results slice
+		if result.Result.Successful() {
```
tamirms, line 134: is it possible for a transaction to be successful but still have some operations which failed? I think that's not possible but it's worth double checking

tamirms (approved):
looks good to me!

ingest/ledger/ledger.go
sydneynotthecity: so much better, thank you

ingest/ledger/ledger.go
sydneynotthecity: transaction success is all or nothing. Either the full transaction succeeds, or it fails. It's not possible for only some operations to fail within a transaction

xdr/ledger_close_meta.go
sydneynotthecity: agreed that updating and fixing existing code is outside the scope of this PR. I do think we should adhere to the principles outlined here for XDRill.

> Did you want Get/Must to be appended to all functions within xdr/leger_close_meta.go or everywhere?

within `ledger_close_meta.go` for now. if we decide to go back and fix some of the consistency problems within the code, we can address then. I also think it's nice to have the Get/Must functions, but unnecessary as long as the data transformation itself follows ok, err and panic principles that make sense.

ingest/ledger/ledger.go
```diff
@@ -0,0 +1,145 @@
...
+	LedgerCloseValueSignature, ok := l.LedgerHeaderHistoryEntry().Header.ScpValue.Ext.GetLcValueSignature()
+	if !ok {
+		return "", false
+	}
+
+	return base64.StdEncoding.EncodeToString(LedgerCloseValueSignature.Signature), true
+}
+
+// TransactionCounts calculates and returns the number of successful and failed transactions
+func TransactionCounts(l xdr.LedgerCloseMeta) (successTxCount, failedTxCount uint32) {
```
sydneynotthecity, line 97: super nit picky: `TransactionCounts()` returns successTxCount and failedTxCount

This is different from `OperationCounts()` which returns successOpCount and totalOpCount

this may be a departure from stellar-etl or Horizon today, but I always found the inconsistency in the return values of these two functions annoying. Take it or leave it on whether you would like to adjust.

sydneynotthecity (approved):
lgtm

ingest/ledger/ledger.go
chowbao: Hmm now that you've pointed it out it is annoying lol

Updated in 21dd54ee561185e3cf8f8f9ccc2e745e07f833fd

Issue #5434, opened 2024-08-19
Update XDR definitions for Protocol 23


Shaptic:
**Epic**: https://github.com/stellar/soroban-rpc/issues/267
**Note**: All work should target a feature branch: TBD.

-----------------------------------

When the new XDR definitions for Protocol 23 are available we will need to update the go monorepo to include these defintions along with the generated Go bindings

PR #5581, opened 2025-01-28
5434/p23 xdr update

urvisavla:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Update XDR definitions for Protocol 23.

### Why

#5434 

### Known limitations
N/A


txnbuild/invoke_host_function_test.go
```diff
@@ -144,7 +144,7 @@ func TestInvokeHostFunctionRoundTrip(t *testing.T) {
 					WriteBytes:   0,
 				},
 				ResourceFee: 1,
-				Ext: xdr.ExtensionPoint{
+				Ext: xdr.SorobanTransactionDataExt{
```
Shaptic, line 147: @SirTyson clarifying question:

The proofs aren't required for invocation unless there's a filter miss, right? So this code should continue to work, though we may see [Heisenbugs](https://en.wikipedia.org/wiki/Heisenbug) in integration testing once in a blue moon if the miss occurs. Right? I think it's fine (we need to do the work to move integration tests to the new simulation response once that's done, anyway), just raising this for visibility.

Shaptic (approved):
Nice! I'd like @SirTyson to answer my Q but also don't want to block this PR so :heavy_check_mark: from me

txnbuild/invoke_host_function_test.go
SirTyson: Not necessarily. During invocation, creating new entries will occasionally require proofs due to bloom filter misses. However, restorations will often require proofs. Now that we have auto restore in p23, this means proofs will be required for invocations.

That being said, proofs (and there associated XDR structs) should not be required until protocol 24. I think next has some State Archival stuff from before the switch to partial state archival. I'll push an update to next tomorrow, but Proofs aren't necessary for p23, sorry.

Shaptic: Right, so it's likely (since these run in a standalone network, filter misses on new become **extremely** unlikely, I imagine) only restoration integration tests will need to be updated, and only for P24.

tamirms (approved):

PR #5582, opened 2025-01-29
PoC PR -Sample protos build (Not to be merged)

karthikiyer56:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

### Why

### Known limitations

[TODO or N/A]


protos/address/address_helpers.go
```diff
@@ -0,0 +1,21 @@
+package address
+
+import (
+	"fmt"
+)
+
+// String returns a human-readable representation of the Address.
+func (a *Address) Strings() string {
```
leighmcculloch, line 8: Typo? Comment says String, function says Strings. Typically the interface to implement is "String".
```suggestion
func (a *Address) String() string {
```

protos/address/address.proto
```diff
@@ -0,0 +1,15 @@
...
+package address;
+option go_package = "github.com/stellar/go/protos/address;address";
+
+// Address message with oneof for different address types
+message Address {
+  oneof address_type {
+    string smartContractAddress = 1; // Smart Contract address
+    string accountAddress = 2;       // Account address
+    string liquidityPoolHash = 3;    // Liquidity Pool hash
+    string claimableBalanceId = 4;   // Claimable Balance ID
```
leighmcculloch, line 12: Make these a strkey instead? Liquidity Pools and Claimable Balances are likely to get strkeys as part of the work on CAP-67 and be rendered as strkeys in the XDR-JSON, so you could use the same format here.

We can move along strkey definitions for them to unblock any work that might be dependent on them.

protos/address/address.proto
```diff
@@ -0,0 +1,15 @@
+syntax = "proto3";
+
+package address;
+option go_package = "github.com/stellar/go/protos/address;address";
+
+// Address message with oneof for different address types
+message Address {
+  oneof address_type {
+    string smartContractAddress = 1; // Smart Contract address
```
leighmcculloch, line 9: Suggestion, the longer form is unnecessary. Elsewhere in code we simply refer to them as contract addresses:
```suggestion
    string contractAddress = 1; // Contract address
```

protos/address/address.pb.go
leighmcculloch, line 1: It's undesirable to group all types into a functional package like "proto" rather than a domain focused package which will be the location where they'll be used.

For example, if as a dev using the ingest SDK I'm browsing the ingest package docs, and am looking for related and interesting types, I won't see any there because they're all defined outside the ingest package in the 'protos' package. When I look at the protos package I'll see an assortment of types that relate to the ingest SDK, and potentially others that don't assuming it becomes a general place to place any protos for anything in the monorepo.

Is it possible to generate the protos within each application or package where they'll be used? So that the proto types for a specific processor are grouped with that processor?

leighmcculloch (commented):
Couple thoughts inline.

protos/address/address.proto
chowbao: Hmm is there a reasonable way to calculate the strkey for lp and claimable balances? I would say if it's added to the strkey lib then yes I would agree with using them.

One thing to note is that would be net new and would need existing systems to know to use the strkey lp instead of hash going forward (not blocking but would "break" things like stellar-etl)

karthikiyer56: I am not even sure what you guys mean when you say `strkey`.

The reason the `Address` is structured as a `oneof` is to align with the `enum SCAddressType` - the one that will be extended to add Liquiditypools and CBs as a part of CAP-67

protos/address/address.pb.go
karthikiyer56: I could very well move this entire `protos` inside the `ingest/` directory
That way it is all under ingest, if thats what you were suggesting?

protos/address/address.proto
karthikiyer56: so, basically, you are saying that the Address can just be a  proto string then?

protos/address/address.proto
karthikiyer56: Noted.
I want to point out that I want this PR to be used to decide on structure of the code, more than what goes inside the eventual proto data model will look like
That will be worked upon in a separte PR. review comments from here will be addressed there.

protos/address/address.proto
leighmcculloch: Strkeys is the format of Stellar Addresses, they're defined in SEP-23:
- https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0023.md

protos/address/address.pb.go
leighmcculloch: > The option go_package field that I specify in the .proto files already separates the resulting golang code logically, if that is what your concern was.

Yes this was my concern because I saw .pb.go files ending up in a `protos/` directory. Ideally the go types live with their respective SDKs / directories.

Ideally .proto files would live there too, but I understand that's not a desirable tradeoff.

karthikiyer56: @leighmcculloch - I made it so that the data model for tokenTransferEvent is generated under where the code will potentially reside. 

And I am hoping that it is under `/ingest/processors`

Assuming that the new place for these processors in under `ingest/processors/<processor_name>/`,
the as long as the corresponding `.proto` files follow the same directory structure under `/protos`

i.e 
1. `/protos/address/address.proto` --> `/ingest/address/address.pb.go` (Additionally, added a helper file with receiver functions in `/ingest/address`)
2. `/protos/asset/asset.proto` --> `/ingest/asset/asset.pb.go` (Additionally, added a helper file with receiver functions in `/ingest/asset/`)
3. The main one
`/protos/processors/token_transfer/token_transfer_event.proto` --> `/ingest/processors/token_transfer/token_transfer_event.pb.go`

There are a few important points to note here:
- We should actually **not be using** the `option go_path` in the `.proto` files themselves
- Instead, I got creative with the Makefile target, and achieved the same sentiment using `-go_opts` and the `-M` flag. Checkout the commit message, where I explain why I did that
cc @tamirms

protos/processors/token_transfer/token_transfer_event.proto
```diff
@@ -1,7 +1,7 @@
 syntax = "proto3";
 
 package token_transfer;
-option go_package = "github.com/stellar/go/protos/processors/token_transfer;token_transfer";
+//option go_package = "github.com/stellar/go/protos/processors/token_transfer;token_transfer";
```
karthikiyer56, line 4: This is a big no-no
I have left it commented out, instead of deleting, for the sake of clarity, in this PoC PR

Makefile
```diff
@@ -73,3 +73,40 @@ xdr-clean:
...
+    	done); \
+    	MAP_OPTS=$${MAP_OPTS%,}; \
+    	echo "Running protoc with options:"; \
+    	echo "  --go_out=ingest"; \
+    	echo "  --go_opt=paths=source_relative"; \
+    	echo "  --go_opt=$$MAP_OPTS"; \
+    	echo "  $(PROTO_FILES)"; \
+    	protoc -I=$(PROTO_DIR) \
+    	       --go_out=ingest --go_opt=paths=source_relative \
+    	       --go_opt=$$MAP_OPTS \
```
karthikiyer56, line 105: Use `-go_opt` with the `M` flag to indicate what the package location/path inside the generated proto should look like

Issue #5480, opened 2024-09-30
Implement a LedgerBackend for benchmarking ingestion performance as a function of ledger size


tamirms:
Given the following configuration options:

- minimum number of transactions / operations per ledger
- ledger close time
- queue capacity

The LedgerBackend will start prefetching ledgers from the data lake and, using https://github.com/stellar/go/issues/5479 , it will combine consecutive ledgers until we have a ledger which meets the minimum number of transactions / operations per ledger. Then the resulting "fat" ledger will be emitted at a rate matching the configured ledger close time (we can default to 5 seconds per ledger). If the fat ledgers are emitted at a rate faster than they can be consumed, the LedgerBackend will push the fat ledgers onto a queue. Once the queue has exceeded its capacity we can terminate with an error because we know that the consumer of the LedgerBackend is not able to keep up with the current throughput settings.

PR #5556, opened 2024-12-20
services/horizon/internal/integration: Add load test which generates ledgers loaded with soroban Tx Meta

tamirms:
<details>
  <summary>PR Checklist</summary>
  
### PR Structure

* [ ] This PR has reasonably narrow scope (if not, break it down into smaller PRs).
* [ ] This PR avoids mixing refactoring changes with feature changes (split into two PRs
  otherwise).
* [ ] This PR's title starts with name of package that is most changed in the PR, ex.
  `services/friendbot`, or `all` or `doc` if the changes are broad or impact many
  packages.

### Thoroughness

* [ ] This PR adds tests for the most critical parts of the new functionality or fixes.
* [ ] I've updated any docs ([developer docs](https://developers.stellar.org/api/), `.md`
  files, etc... affected by this change). Take a look in the `docs` folder for a given service,
  like [this one](https://github.com/stellar/go/tree/master/services/horizon/internal/docs).

### Release planning

* [ ] I've reviewed the changes in this PR and if I consider them worthwhile for being mentioned on release notes then I have updated the relevant `CHANGELOG.md` within the  component folder structure. For example, if I changed horizon, then I updated ([services/horizon/CHANGELOG.md](services/horizon/CHANGELOG.md). I add a new line item describing the change and reference to this PR. If I don't update a CHANGELOG, I acknowledge this PR's change may not be mentioned in future release notes.  
* [ ] I've decided if this PR requires a new major/minor version according to
  [semver](https://semver.org/), or if it's mainly a patch change. The PR is targeted at the next
  release branch if it's not a patch change.
</details>

### What

Close https://github.com/stellar/stellar-rpc/issues/330 

This PR implements an integration load test to generate ledgers with soroban transactions. The test can be configured using the following parameters:

- `transactions-per-ledger` number of transactions per ledger
- `transfers-per-tx` number of asset transfers for each transaction
- `ledgers` number of ledgers to generate 

The test will output `ledgers` number of ledgers each containing  `transactions-per-ledger` transactions. Each transaction will have `transfers-per-tx` asset transfers. Half the asset transfers will result in movement among accounts and the other half of the transfers will result in movement between accounts and contracts.

The test is implemented as follows:

1. We use https://github.com/stellar/go/pull/5546 to setup a private stellar network with unlimited soroban TPS.
2. We use a bulk transfer soroban contract which will allow us to have a single invocation which transfers assets to multiple destination accounts / contracts
3. We submit many soroban transactions to core
[...]

PR #5584, opened 2025-01-30
ingest/ledgerbackend: Add ledger backend which replays ledgers

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Close https://github.com/stellar/go/issues/5480

Following https://github.com/stellar/go/pull/5556 , this PR implements a ledger backend which takes synthetically generated ledgers, merges them with real ledgers, and then replays them at a configurable rate.

### Why

This ledger backend will be used to benchmark ingestion duration for horizon / rpc when ingesting ledgers with a high volume of soroban transactions. These benchmarks will inform us of horizon and rpc's ceiling when it comes to ingestion throughput.

### Known limitations

[N/A]


services/horizon/internal/integration/change_test.go
```diff
@@ -150,11 +151,11 @@ func getLedgers(itest *integration.Test, startingLedger uint32, endLedger uint32
...
-	if err != nil {
-		t.Fatalf("unable to create captive core config: %v", err)
-	}
+	require.NoError(t, err)
+
+	captiveCore, err := ledgerbackend.NewCaptive(ccConfig)
+	require.NoError(t, err)
 
-	captiveCore, err := ledgerbackend.NewCaptive(*ccConfig)
 	if err != nil {
```
karthikiyer56, line 159: this 2nd `if err!= nil` can also be removed then

ingest/replay_backend.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+	var changes xdr.LedgerEntryChanges
+	for i := 0; i < len(r.generatedLedgerEntries); i++ {
+		changes = append(changes, xdr.LedgerEntryChange{
+			Type:    xdr.LedgerEntryChangeTypeLedgerEntryCreated,
+			Created: &r.generatedLedgerEntries[i],
+		})
+	}
+	var flag xdr.Uint32 = 1
+	ledger.V1.UpgradesProcessing = append(ledger.V1.UpgradesProcessing, xdr.UpgradeEntryMeta{
```
karthikiyer56, line 97: it looks like you are adding already buffered entries in `ReplayBackend.generatedEntries[i]` to the incoming ledger's UpgradeChanges? 

Why is that?

ingest/replay_backend.go
```diff
@@ -0,0 +1,275 @@
...
+		if err != nil {
+			return err
+		}
+		if err = MergeLedgers(r.networkPassphrase, &ledger, r.generatedLedgers[0]); err != nil {
+			return err
+		}
+		r.mergedLedgers = append(r.mergedLedgers, ledger)
+		r.generatedLedgers = r.generatedLedgers[1:]
+	}
+	r.startTime = time.Now()
```
karthikiyer56, line 120: Is this correct?
shouldnt `r.startTime` be the time consistent with the close time of the `ledger` fetched from `GetLedger()`
i.e shudnt `r.startTime = ledger.LedgerCloseTime()`

ingest/replay_backend.go
```diff
@@ -0,0 +1,275 @@
...
+	return nil
+}
+
+func (r *ReplayBackend) IsPrepared(ctx context.Context, ledgerRange ledgerbackend.Range) (bool, error) {
+	return r.ledgerBackend.IsPrepared(ctx, ledgerRange)
+}
+
+func (r *ReplayBackend) GetLedger(ctx context.Context, sequence uint32) (xdr.LedgerCloseMeta, error) {
+	if sequence < r.startLedger {
+		return xdr.LedgerCloseMeta{}, fmt.Errorf("sequence number %v out of range", sequence)
```
karthikiyer56, line 131: Perhaps the error can be more descriptive. - `sequence number is less than the earliest ledger available`?

ingest/replay_backend.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+	i := int(sequence - r.startLedger)
+	if i >= len(r.mergedLedgers) {
```
karthikiyer56, line 134: Perhaps the error can be more descriptive - `sequence number is greater than the latest ledger available`?

ingest/replay_backend.go
```diff
@@ -0,0 +1,275 @@
...
+		return false, err
+	}
+	otherSerialized, err := b.MarshalBinary()
+	if err != nil {
+		return false, err
+	}
+	return bytes.Equal(serialized, otherSerialized), nil
+}
+
+func changeIsEqual(a, b Change) (bool, error) {
```
karthikiyer56, line 196: IMO this shud be moved from here to be a receiver function in `ingest/change.go`
Something like `func (a Change)IsEqual(b Change)`
I remember that I had some usecase for comparing changes when I was working on some change related issues, and I was needing such a compare function

tamirms: This function deliberately doesn't compare all the fields in the change struct so in that sense it is tailored specifically to the replay backend's needs. That's why I did not expose it as a public utility function.

The changeIsEqual function compares the xdr values of the pre and post but it does not compare the the transaction, operation, or ledger because it is used to determine that the changes from merging two input ledgers is equal to changes from the first of the input ledgers followed by the changes from the second input ledgers.

ingest/replay_backend.go
tamirms: no, this is used to throttle the rate at which ledgers are obtained from the replay backend. I will try to add some comments to make that clear

tamirms:
@karthikiyer56 I think I have addressed your code review feedback, could you take another look?

karthikiyer56 (approved):

ingest/replay_backend.go
```diff
@@ -0,0 +1,296 @@
...
+	"os"
+	"time"
+
+	"github.com/klauspost/compress/zstd"
+
+	"github.com/stellar/go/ingest/ledgerbackend"
+	"github.com/stellar/go/xdr"
+)
+
+type ReplayBackend struct {
```
2opremio, line 19: I would probably call this LocalFileBackend or FsBackend to indicate it comes from a local file.

ingest/replay_backend.go
```diff
@@ -0,0 +1,296 @@
...
+func NewReplayBackend(config ReplayBackendConfig, ledgerBackend ledgerbackend.LedgerBackend) (*ReplayBackend, error) {
+	var generatedLedgers []xdr.LedgerCloseMeta
+	var generatedLedgerEntries []xdr.LedgerEntry
+
+	if err := unmarshallCompressedXDRFile(config.LedgerEntriesFilePath, &generatedLedgerEntries); err != nil {
+		return nil, err
+	}
+	if err := unmarshallCompressedXDRFile(config.LedgersFilePath, &generatedLedgers); err != nil {
+		return nil, err
+	}
```
2opremio, line 66: It may be better to stream the decoding instead of doing it upfront. It allows for larger files not fitting memory and allows parelellizing the decompression and decoding in the future.

ingest/replay_backend.go
```diff
@@ -0,0 +1,296 @@
...
+		b64, err = ledgerKey.MarshalBinaryBase64()
+		if err != nil {
+			return err
+		}
+		changeMap[b64] = append(changeMap[b64], change)
+	}
+	return nil
+}
+
+func xdrEquals(a, b encoding.BinaryMarshaler) (bool, error) {
```
2opremio, line 198: Shouldn't this live in the XDR package?

ingest/replay_backend.go
```diff
@@ -0,0 +1,296 @@
...
+		}
+	} else {
+		if ok, err := xdrEquals(a.Post, b.Post); err != nil || !ok {
+			return ok, err
+		}
+	}
+	return true, nil
+}
+
+func changesAreEqual(a, b map[string][]Change) (bool, error) {
```
2opremio, line 235: Shouldn't this live Change is defined?

ingest/replay_backend.go
```diff
@@ -0,0 +1,296 @@
...
+			if ok, err := changeIsEqual(aChange, bChange); !ok || err != nil {
+				return ok, err
+			}
+		}
+	}
+	return true, nil
+}
+
+// MergeLedgers merges two xdr.LedgerCloseMeta instances.
+func MergeLedgers(networkPassphrase string, dst *xdr.LedgerCloseMeta, src xdr.LedgerCloseMeta) error {
```
2opremio, line 255: Shouldn't this live in the XDR package?

ingest/replay_backend.go
```diff
@@ -0,0 +1,296 @@
...
+	"time"
+
+	"github.com/klauspost/compress/zstd"
+
+	"github.com/stellar/go/ingest/ledgerbackend"
+	"github.com/stellar/go/xdr"
+)
+
+type ReplayBackend struct {
+	ledgerBackend          ledgerbackend.LedgerBackend
```
2opremio, line 20: This ledger backend has a strange role. It composes in a strange way with the actual backend, implementing only certain operations. Also in order for it to work, the passphrase of the passed backend need to be the same, so it's duplicated.

I wonder if we can simply mock some parameters so that passing the backend isn't necessary.

2opremio:
All my comments are optional. I am mostly concerned about the role of the passed ledger backend, which is a bit unintuitive to me.

2opremio (approved):

ingest/replay_backend.go
tamirms: the ledger backend is necessary because we want to merge the synthetic ledgers with real ledgers in order to have a mix of soroban activity and stellar classic activity. We could do the merging process offline and supply a file of ledgers that already have been merged with real ledgers but I think there's an advantage to merging on the fly. If we merge on the fly we can ingest any ledger range at runtime. However, if we merge the ledgers offline then we need to predetermine the ledger range we will be ingesting in the benchmarks ahead of time.

2opremio: Can you add a comment explaining this?

ingest/replay_backend.go
tamirms: I tried to move it to the xdr package but I could not because it would introduce a circular dependency with the ingest package

ingest/replay_backend.go
tamirms: This function deliberately doesn't compare all the fields in the change struct so in that sense it is tailored specifically to the replay backend's needs. That's why I did not expose it as a public utility function.

PR #5586, opened 2025-02-03
ingest: Allow enabling Captive Core's fast HTTP server

2opremio:


ingest/ledgerbackend/toml.go
```diff
@@ -346,6 +355,8 @@ type CaptiveCoreTomlParams struct {
 	EnforceSorobanDiagnosticEvents bool
 	// Enfore EnableSorobanTransactionMetaExtV1 when not disabled explicitly
 	EnforceSorobanTransactionMetaExtV1 bool
+	// Enable the Fast HTTP server with these parameters
```
leighmcculloch, line 358: What does it mean to enable the fast http server?

Shaptic: @leighmcculloch see https://github.com/stellar/stellar-rpc/issues/269 and

ingest/ledgerbackend/toml.go
```diff
@@ -498,6 +509,15 @@ func (c *CaptiveCoreToml) setDefaults(params CaptiveCoreTomlParams) {
...
 		enforceOption(&c.EnableEmitSorobanTransactionMetaExtV1)
 	}
+
+	if params.FastHTTPServerParams != nil {
+		port := uint(params.FastHTTPServerParams.Port)
+		c.HTTPQueryPort = &port
+		ledgers := uint(params.FastHTTPServerParams.SnapshotLedgers)
+		c.QuerySnapshotLedgers = &ledgers
+		poolSize := uint(params.FastHTTPServerParams.ThreadPoolSize)
+		c.QueryThreadPoolSize = &poolSize
```
Shaptic, line 519: These should have reasonable defaults, right? We should make it easy to set this up from a configuration standpoint.

ingest/ledgerbackend/toml.go
```diff
@@ -321,6 +324,12 @@ func (c *CaptiveCoreToml) unmarshal(data []byte, strict bool) error {
 	return nil
 }
 
+type FastHTTPServerParams struct {
```
tamirms, line 327:
```suggestion
type HTTPQueryServerParams struct {
```

It seems that core is naming this server as the http query server since the toml fields all include "query" as a substring. So, I think a better name  would be HTTPQueryServerParams. If you accept this name suggestion then the comment in CaptiveCoreTomlParams  can be updated to:

```go
	// Enable the HTTP query server with these parameters
	HTTPQueryServerParams *HTTPQueryServerParams
```

ingest/ledgerbackend/toml.go
```diff
@@ -498,6 +509,15 @@ func (c *CaptiveCoreToml) setDefaults(params CaptiveCoreTomlParams) {
 	if params.EnforceSorobanTransactionMetaExtV1 {
 		enforceOption(&c.EnableEmitSorobanTransactionMetaExtV1)
 	}
+
+	if params.FastHTTPServerParams != nil {
+		port := uint(params.FastHTTPServerParams.Port)
+		c.HTTPQueryPort = &port
```
tamirms, line 515: it seems that these will override any values which are present in the toml file. It seems that usually we only set values if they are not already defined in the toml

tamirms: I think we might need to update `validate()` to assert that the values in `FastHTTPServerParams` are consistent with the values provided in the toml file

ingest/ledgerbackend/toml.go
2opremio: I think that the default is correct (for now at least) which is to omit the parameters so that the server is not enabled.

ingest/ledgerbackend/toml.go
2opremio: I have now modified the code so that the values are only overriden if not set in the file.

Can you provide an example of a possible inconsistency?

tamirms: In the `validate()` function there are checks to see that the config values match the toml fields in the case that both are configured. For example, this block of code checks that if the toml port is configured and the params port is configured both values must match:

https://github.com/stellar/go/blob/6e4c56eeea5534f70bcfdb7d94aa5ed3af3ceb5b/ingest/ledgerbackend/toml.go#L552-L558

I think we need to do the same for all of the new params introduced to support the query server. And we should also add some test cases which exercise the validation logic

tamirms (approved):

ingest/ledgerbackend/toml.go
2opremio: We discussed it offline and decided to make the query parameters consistent with the others:
* Not make them overrides (just optional)
* Validate them if present

PR #5536, opened 2024-11-20
Include operation information in the ingest.Change struct

karthikiyer56:
<details>
  <summary>PR Checklist</summary>
  
### PR Structure

* [x] This PR has reasonably narrow scope (if not, break it down into smaller PRs).
* [x] This PR avoids mixing refactoring changes with feature changes (split into two PRs
  otherwise).
* [x] This PR's title starts with name of package that is most changed in the PR, ex.
  `services/friendbot`, or `all` or `doc` if the changes are broad or impact many
  packages.

### Thoroughness

* [x] This PR adds tests for the most critical parts of the new functionality or fixes.
* [] I've updated any docs ([developer docs](https://developers.stellar.org/api/), `.md`
  files, etc... affected by this change). Take a look in the `docs` folder for a given service,
  like [this one](https://github.com/stellar/go/tree/master/services/horizon/internal/docs).

### Release planning

* [x] I've reviewed the changes in this PR and if I consider them worthwhile for being mentioned on release notes then I have updated the relevant `CHANGELOG.md` within the  component folder structure. For example, if I changed horizon, then I updated ([services/horizon/CHANGELOG.md](services/horizon/CHANGELOG.md). I add a new line item describing the change and reference to this PR. If I don't update a CHANGELOG, I acknowledge this PR's change may not be mentioned in future release notes.  
* [x] I've decided if this PR requires a new major/minor version according to
  [semver](https://semver.org/), or if it's mainly a patch change. The PR is targeted at the next
  release branch if it's not a patch change.
</details>

### What

This PR deals with this github issue - https://github.com/stellar/go/issues/5535

### Why
Details in the associated github issue -  github issue - https://github.com/stellar/go/issues/5535

Tl;DR:
- This serves as as the basis for identifying exactly what operation/activity caused the change in ledgreEntry
- This change serves as a precursor for writing several generic change/transaction processors, primarily offer+trade processor

### Known limitations
These extensions to the `Change` struct do not work when the compacting change reader is used.

[TODO or N/A]

PR #5587, opened 2025-02-04
(WIP) ingest: Add support for ledger entry change type "restore"

urvisavla:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

### Why

### Known limitations

[TODO or N/A]


ingest/change.go
```diff
@@ -144,24 +147,46 @@ func GetChangesFromLedgerEntryChanges(ledgerEntryChanges xdr.LedgerEntryChanges)
...
-				Pre:  &state,
-				Post: nil,
+				Type:       state.Data.Type,
+				Pre:        &state,
+				Post:       nil,
+				ChangeType: entryChange.Type,
+			})
+		case xdr.LedgerEntryChangeTypeLedgerEntryRestored:
+			restored := entryChange.MustRestored()
+			if i > 0 {
```
sreuland, line 174: maybe a comment or link to cap explaining the reason for this sequence of changes, does `i > 0` represent detection of a Restored TTL case, the new integration tests which are great, seem to imply that?

ingest/ledger_change_reader_test.go
```diff
@@ -598,6 +599,281 @@ func TestLedgerChangeLedgerCloseMetaV2(t *testing.T) {
...
+												},
+											},
+										},
+										xdr.LedgerEntryChange{
+											Type: xdr.LedgerEntryChangeTypeLedgerEntryRestored,
+											Restored: &xdr.LedgerEntry{
+												Data: xdr.LedgerEntryData{
+													Type: xdr.LedgerEntryTypeTtl,
+													Ttl: &xdr.TtlEntry{
+														KeyHash:            keyHash,
```
sreuland, line 805: for test this is ok, just for own understanding in real cases of the new archival protocol, the KeyHash here should refer to the hash of the LedgerEntry which is Created in the preceeding Change correct?

ingest/change.go
```diff
@@ -223,16 +248,7 @@ func sortChanges(changes []Change) {
 
 // LedgerEntryChangeType returns type in terms of LedgerEntryChangeType.
 func (c Change) LedgerEntryChangeType() xdr.LedgerEntryChangeType {
```
tamirms, line 250: I think it makes sense to remove this function entirely since `ChangeType` is a public field and can be accessed by anyone who would otherwise call the `LedgerEntryChangeType()` function

ingest/change_compactor.go
```diff
@@ -210,22 +235,109 @@ func (c *ChangeCompactor) addRemovedChange(change Change) error {
...
+				"can't restore an entry that already exists (ledger key = %s)",
+				base64.StdEncoding.EncodeToString(ledgerKey),
+			))
+		case xdr.LedgerEntryChangeTypeLedgerEntryRemoved:
+			c.cache[ledgerKeyString] = Change{
+				Type:       key.Type,
+				Pre:        existingChange.Pre,
+				Post:       change.Post,
+				ChangeType: change.ChangeType,
+			}
```
Shaptic, line 306: I'm probably misunderstanding the behavior here. If the change being inserted is a restore, then shouldn't we set pre = nil, post = change. And if it's being removed, then shouldn't it be a no-op (i.e. pre and post set, so no-op)?

ingest/change_compactor.go
```diff
@@ -210,22 +235,109 @@ func (c *ChangeCompactor) addRemovedChange(change Change) error {
...
+		default:
+			return errors.Errorf("Unknown LedgerEntryChangeType: %d", existingChange.LedgerEntryChangeType())
+		}
+	} else {
+		// If 'Pre' is not nil, it indicates that an item previously *archived* is being restored.
+		switch existingChange.ChangeType {
+		case xdr.LedgerEntryChangeTypeLedgerEntryCreated:
+			c.cache[ledgerKeyString] = Change{
+				Type:       key.Type,
+				Pre:        nil,
```
Shaptic, line 316: Why not set `change.Pre` here?

ingest/change.go
Shaptic: as long as we mark the breaking API change in the changelog nobody updates :wink:

urvisavla: That would be a breaking change which I’d prefer to avoid if possible. 
Noted changelog update @Shaptic

ingest/change_compactor.go
```diff
@@ -25,34 +25,58 @@ import (
...
-//     existing change is:
-//     a. CREATED it means that due to previous transitions we want to create
-//     this in a DB what means that it doesn't exist in a DB. If it was
-//     created and removed in the same ledger it's a noop so we remove entry
-//     from the cache.
-//     b. UPDATED we simply update it to be a REMOVE change because the UPDATE
-//     change means the entry exists in a DB.
-//     c. REMOVED it returns error because we can't remove an entry that was
-//     already removed.
+//     d. RESTORED: update it with the new value but keep the type as RESTORED.
```
karthikiyer56, line 44: Can someone help break case 2.d down for me.
we are compacting changes in a ledger.
we come across an entry that has changeType set as updated.
Now we come across the same entry,  with a changeType set as restored. Shudnt this be an error?
What does it mean to have an entry show up as restored when it was updated (and not evicted)?

urvisavla: 2.d is about "updating" an entry that was previously "restored" not the other way around.

karthikiyer56: Argh.
sorry, you are right.

ingest/change.go
tamirms: it's ok to make breaking changes along protocol boundaries as long as it's possible for consumers of the ingestion library to upgrade ahead of the protocol switch. I believe that is the case here. Also, anyone who is ingesting data from Stellar will have to make code changes to support the restoration change type so we might as remove `LedgerEntryChangeType()` to make the API cleaner.

Alternatively, we could make `ChangeType` a private field and still keep `LedgerEntryChangeType()`. I think having both a public `ChangeType` field and a public  `LedgerEntryChangeType()` function can be confusing

urvisavla: @tamirms that's a good point since it's a protocol boundary, it's ok to make a breaking change. I would have preferred making the `ChangeType` private but the ingest/processors tests rely on setting this field. So I'll go ahead and just remove the getter function as you suggested.

ingest/change_compactor.go
```diff
@@ -203,29 +228,116 @@ func (c *ChangeCompactor) addRemovedChange(change Change) error {
...
+			ChangeType: change.ChangeType,
 		}
 	case xdr.LedgerEntryChangeTypeLedgerEntryRemoved:
 		return NewStateError(errors.Errorf(
 			"can't remove an entry that was previously removed (ledger key = %s)",
 			base64.StdEncoding.EncodeToString(ledgerKey),
 		))
+	case xdr.LedgerEntryChangeTypeLedgerEntryRestored:
+		if existingChange.Pre == nil {
+			// Entry was created and removed in the same ledger; deleting it is effectively a noop.
```
tamirms, line 250: this only true for ttl ledger entries. other ledger entries which can be restored (e.g. contract data) will always have a Pre set to nil. So it is possible for an archived but not yet evicted contract data ledger entry to be restored and then be removed.  In that case, the ledger entry will be completely removed from the ledger and we should not be losing that information during compaction

ingest/change_compactor.go
```diff
@@ -203,29 +228,116 @@ func (c *ChangeCompactor) addRemovedChange(change Change) error {
...
-		return errors.Errorf("Unknown LedgerEntryChangeType: %d", existingChange.Type)
+		return errors.Errorf("Unknown LedgerEntryChangeType: %d", existingChange.ChangeType)
 	}
 
 	return nil
 }
 
+// addRestoredChange adds a change to the cache, but returns an error if the restore
+// change is unexpected.
+func (c *ChangeCompactor) addRestoredChange(change Change) error {
```
tamirms, line 270: the documentation states the change compactor can only be used to squash changes within a single ledger. I think it's easy to overlook that so perhaps we should enforce that assumption by returning an error if the change compactor is fed changes from multiple ledgers.

assuming all the changes come from a single ledger. I do not think it is possible to restore a ledger entry which was created, updated, removed, or restored within the same ledger.

You can only restore something which is in an archived state. If you create a ledger entry at ledger n it cannot be archived in that same ledger. Therefore, you cannot restore a ledger entry which was created in that same ledger.

Similarly, only active ledger entries can be updated. So if a ledger entry at ledger n was updated then it could not be restored within the same ledger because the ledger entry must have been already active at ledger n.

If a ledger entry is removed then there are two cases. If the ledger entry was archived and then evicted it is not possible for the restoration to occur within the same ledger since eviction occurs at the very end when all transactions / operations have been executed. If the ledger entry was completely removed from the network then it is not possible to restore the ledger entry because it does not exist.

finally, when a ledger entry is restored it becomes active and it cannot be restored again in the same ledger because restoration can only be done on archived ledger entries.

In summary, as long as we're only compacting changes from a single ledger, I think it's impossible for a restoration change to occur after any other type of change for the same ledger entry.

urvisavla: You're right. I've updated the code to return a state error

ingest/change_compactor.go
```diff
@@ -75,13 +104,15 @@ func NewChangeCompactor() *ChangeCompactor {
 // cache takes too much memory, you apply changes returned by GetChanges and
 // create a new ChangeCompactor object to continue ingestion.
 func (c *ChangeCompactor) AddChange(change Change) error {
-	switch {
-	case change.Pre == nil && change.Post != nil:
+	switch change.ChangeType {
```
urvisavla, line 107: I need to add a check here to ensure that the Change Compactor instance is only used to compact changes from a single ledger.

ingest/change_compactor.go
```diff
@@ -203,26 +233,58 @@ func (c *ChangeCompactor) addRemovedChange(change Change) error {
...
+			ChangeType: change.ChangeType,
 		}
 	case xdr.LedgerEntryChangeTypeLedgerEntryRemoved:
 		return NewStateError(errors.Errorf(
 			"can't remove an entry that was previously removed (ledger key = %s)",
 			base64.StdEncoding.EncodeToString(ledgerKey),
 		))
+	case xdr.LedgerEntryChangeTypeLedgerEntryRestored:
+		if c.config.EmitExpiredEntryRemovedChange {
+			c.cache[ledgerKeyString] = Change{
```
urvisavla, line 255: Still considering whether to add a flag to indicate that this removed change is for a restored (previously expired) entry.

ingest/change_compactor.go
```diff
@@ -53,17 +57,42 @@ import (
...
 //     c. REMOVED it returns error because we can't remove an entry that was
 //     already removed.
+//     d. RESTORED depending on the change compactor's configuration, we may or
+//     may not emit a REMOVE change type for an entry that was restored earlier
+//     in the ledger.
+//  4. If the change is RESTORED it checks if any change related to the given
+//     entry already exists in the cache. If not, it adds the RESTORED change.
+//     Otherwise, it returns an error because only expired entries can be
+//     restored. If the entry was created, updated or removed in the same
+//     ledger, it's not expired.
```
tamirms, line 67:
```suggestion
//     ledger, the entry must be active and not archived.
```

ingest/change_compactor.go
```diff
@@ -53,17 +57,42 @@ import (
...
+	return &ChangeCompactorConfig{
+		// By default, set it to true to enable the change compactor
+		// to emit REMOVED change for expired entries.
+		EmitExpiredEntryRemovedChange: true,
+	}
 }
 
 // NewChangeCompactor returns a new ChangeCompactor.
-func NewChangeCompactor() *ChangeCompactor {
+func NewChangeCompactor(config *ChangeCompactorConfig) *ChangeCompactor {
```
tamirms, line 91:
```suggestion
func NewChangeCompactor(config ChangeCompactorConfig) *ChangeCompactor {
```

I think we should pass the config struct by value since it is immutable and not so large

ingest/change_compactor.go
```diff
@@ -53,17 +57,42 @@ import (
...
+}
+
+type ChangeCompactorConfig struct {
+	// Determines whether the change compactor emits a REMOVED change when an expired entry
+	// is restored and then removed within the same ledger.
+	// If set to true, a REMOVED change is emitted; if false, REMOVED change is suppressed.
+	EmitExpiredEntryRemovedChange bool
+}
+
+func NewChangeCompactorDefaultConfig() *ChangeCompactorConfig {
```
tamirms, line 82: I think we should get rid of this function and instead make the empty `ChangeCompactorConfig{}` value align with what we think the defaults should be

ingest/change_compactor.go
```diff
@@ -203,26 +233,58 @@ func (c *ChangeCompactor) addRemovedChange(change Change) error {
...
+func (c *ChangeCompactor) addRestoredChange(change Change) error {
+	ledgerKey, err := c.getLedgerKey(change.Post)
+	if err != nil {
+		return err
 	}
+	ledgerKeyString := string(ledgerKey)
 
+	if _, exist := c.cache[ledgerKeyString]; exist {
+		return NewStateError(errors.Errorf(
+			"can't restore an entry that already exists (ledger key = %s)",
```
tamirms, line 283:
```suggestion
			"can't restore an entry that is already active (ledger key = %s)",
```

ingest/change_compactor.go
urvisavla: The main reason for this approach is that we want to emit the removal change by default. To use ChangeCompactorConfig{} with default value (false), we would need to name the variable something like `skipArchivedEntryRemovedChange` or `doNotEmitArchivedEntryRemovedChange` and I wanted to avoid using negative named config option for Bool as it can be less intuitive to understand.

ingest/change_compactor.go
urvisavla: As discussed this in the planning meeting, added `EmitArchivedEntryRemovalChange` flag to manager this behavior.

ingest/change_compactor.go
```diff
@@ -53,17 +57,42 @@ import (
...
 	cache          map[string]Change
 	encodingBuffer *xdr.EncodingBuffer
+	config         ChangeCompactorConfig
+}
+
+type ChangeCompactorConfig struct {
+	// Determines whether the change compactor emits a REMOVED change when an archived entry
+	// is restored and then removed within the same ledger.
+	// If set to true, a REMOVED change is emitted; if false, REMOVED change is suppressed.
+	EmitArchivedEntryRemovedChange bool
```
tamirms, line 79: If you want to keep to align the default behavior with the empty struct you could rename the field to `SupressRemoveAfterRestoreChange` . When `SupressRemoveAfterRestoreChange` is set to true a REMOVE following a RESTORE results in an no-op

tamirms: not sure if `SupressRemoveAfterRestoreChange` is any better than `EmitArchivedEntryRemovedChange` tbh so I'll let you decide on which you prefer

tamirms:
@urvisavla the PR looks good, I think the only remaining task is to add a check to ensure that the Change Compactor instance is only used to compact changes from a single ledger

urvisavla:
> @urvisavla the PR looks good, I think the only remaining task is to add a check to ensure that the Change Compactor instance is only used to compact changes from a single ledger

@tamirms 
To ensure the change compactor processes changes from only one ledger sequence, I initially tried checking for mismatches like this:

```
for each existingChange in cache:
    if existingChange.LedgerSequenceNumber ≠ newChange.LedgerSequenceNumber:
        return "Ledger sequence mismatch: Cannot compact changes from different ledgers"

    break // Only check the first entry
```

However, I'm having trouble reliably getting the ledger sequence number from a `Change`. I've looked at `Change::LedgerEntry::LastModifiedLedgerSeq` (Pre & Post), `Change::Transaction`, `Change::Ledger` etc but none of them consistently provide the correct sequence number for all LedgerEntry changes.

What's the most reliable way to determine the current ledger sequence number for a Change without storing it in the `ChangeCompactor` object and then passing it with every `AddChange` call?

Also, it seems the new Change struct fields added in https://github.com/stellar/go/pull/5536 are not designed to be used with the Change Compactor which wasn't immediately obvious to me and caused some confusion.

ingest/change_compactor.go
urvisavla: done.

urvisavla:
> The fields are not preserved after the compaction process but the changes fed into the change reader should still have valid fields

These fields must be preserved during compaction if we are to compare the ledger sequence number of incoming changes with existing ones in the cache. However, there are alternative solutions, such as storing the ledger sequence in the ChangeCompactor object. I have created #5607 to evaluate potential solutions and implement it separately to unblock this PR.

tamirms (approved):

ingest/change_compactor.go
urvisavla: https://github.com/stellar/go/issues/5607

Issue #5425, opened 2024-08-13
Update CheckpointChangeReader to handle bucketlist entry types


tamirms:
We will need to update the checkpoint change reader which is used by both state verification and ingesting ledger entries from history archives when bootstrapping from an empty horizon db. The checkpoint change reader scans through the bucket list in the history archives to find all ledger entries which have not been evicted:

[https://github.com/stellar/go/blob/cabbd30486ee22c50648166040dfb311e4912c7e/ingest/checkpoint_change_reader.go#L129-L151](https://github.com/stellar/go/blob/cabbd30486ee22c50648166040dfb311e4912c7e/ingest/checkpoint_change_reader.go#L129-L151)

Since there will be structural changes to the bucket list and bucket list entries to support state archival, we will need to update the checkpoint change reader so that it handles the new bucket list design correctly

tamirms:
this issue is mostly obsolete because it turns out there will be no changes to how bucket entries will be stored in the history archives for the live state bucket list.

From https://github.com/stellar/stellar-protocol/blob/master/core/cap-0062.md#live-state-bucketlist :

>The Live State BucketList most closely resembles the current BucketList. It will contain all “live” state of the ledger, including Stellar classic entries, live Soroban entries, network config settings, etc. The Live State BucketList is published to the History Archive on every checkpoint ledger via the "history" category. The Live BucketList functions identically to the current BucketList.

However, there will be a change to how the bucket list hash is computed in the ledger header:


>While the XDR structure of LedgerHeader remains unchanged, bucketListHash will be changed to the following:

>header.bucketListHash = SHA256(liveStateBucketListHash, hotArchiveHash)

This will require an update to the checkpoint change reader which validates the bucket list hash:

https://github.com/stellar/go/blob/6dc6d2f3dfb4b37c2dedbed4f4124cffa24ac469/ingest/checkpoint_change_reader.go#L102-L108

PR #5590, opened 2025-02-06
historyarchive: Update bucket list hash validation for protocol 23

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Close https://github.com/stellar/go/issues/5425

In protocol 23 there will be two bucketlists: one for the live ledger entries and one for evicted archived ledger entries. Consequently the bucketlist hash in the ledger header will change to (see [CAP-62](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0062.md#changes-to-ledgerheader)):

```
header.bucketListHash = SHA256(liveStateBucketListHash, hotArchiveHash)
```

The checkpoint change reader will need to be updated to calculate the bucket list hash in this way. The [live bucketlist](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0062.md#live-state-bucketlist) will still be the same as the current bucketlist we ingest in the checkpoint change reader so there will not be any changes required there. However, I did add an assertion to ensure the checkpoint change reader is ingesting from the live bucket list instead of the bucket list for evicted ledger entries.

### Known limitations

[N/A]

### DIFF
--- a/historyarchive/history_archive_state_test.go
+++ b/historyarchive/history_archive_state_test.go
@@ -7,6 +7,7 @@ package historyarchive
 import (
 	"encoding/hex"
 	"encoding/json"
+	"os"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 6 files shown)


ingest/checkpoint_change_reader_test.go
```diff
@@ -399,6 +405,32 @@ func (s *SingleLedgerStateReaderTestSuite) TestMalformedProtocol11BucketNoMeta()
 	s.Assert().Equal("Error while reading from buckets: Read INITENTRY from version <11 bucket: 0@517bea4c6627a688a8ce501febd8c562e737e3d86b29689d9956217640f3c74b", err.Error())
 }
 
+// TestMalformedProtocol11Bucket ensures the checkpoint change reader asserts its reading from the live bucketlist
```
urvisavla, line 408:
```suggestion
// TestMalformedBucketListType ensures the checkpoint change reader asserts its reading from the live bucketlist
```

historyarchive/history_archive_state_test.go
```diff
@@ -47,112 +48,36 @@ func TestUnmarshalState(t *testing.T) {
...
-            "snap": "0000000000000000000000000000000000000000000000000000000000000000"
-        }
-    ]
-}`)
+	for _, testCase := range []struct {
+		inputFile    string
+		expectedHash string
+	}{
+		{
+			// This is real bucket hash list for pubnet's ledger: 24088895
```
sreuland, line 56: nice!

ingest/checkpoint_change_reader.go
```diff
@@ -325,9 +325,18 @@ func (r *CheckpointChangeReader) streamBucketContents(hash historyarchive.Hash,
 				)
 				return false
 			}
-			// We can't use MustMetaEntry() here. Check:
-			// https://github.com/golang/go/issues/32560
-			bucketProtocolVersion = uint32(entry.MetaEntry.LedgerVersion)
+			metaEntry := entry.MustMetaEntry()
+			bucketProtocolVersion = uint32(metaEntry.LedgerVersion)
+			bucketListType, ok := metaEntry.Ext.GetBucketListType()
+			if ok && bucketListType != xdr.BucketListTypeLive {
```
sreuland, line 331: what is the real life possibility of this occurrence, can see that only streamBuckets() calls this while parsing currentBuckets, of which that list should only continue to have live buckets correct, if this got triggered it means the remote archive is corrupt, as core would not emit this way, at least that's how it appears in the historyV1.json and historyv2.json test cases and cap-62 didn't mention a behavior change that could lead to this corner case, just trying to understand why this detection is necessary?

tamirms: this code makes sure we are reading the live bucketlist instead of the hot archive bucketlist. in practice this error should never occur unless we are reading off the wrong location in the history archives or if, as you mentioned, the history archives are corrupted.

here is a description of the bucketlists in cap-62 https://github.com/stellar/stellar-protocol/blob/master/core/cap-0062.md#live-state-bucketlist

sreuland (approved):

Issue #5591, opened 2025-02-06
Protos for TokenTransferProccessor


karthikiyer56:
- Generate appropriate protos for the data model defined in #5580 
- Specify code organization guidelines for protos and generated golang code
- Makefile targets for generating/regenerating go bindings from protos

PR #5596, opened 2025-02-08
Karthik/5591/protos for ttp

karthikiyer56:
### What

This PR deals with implementing protos as defined in  [issue 5591](https://github.com/stellar/go/issues/5591)


ingest/address/address.go
sreuland, line 1: these helper functions seem to be extra code to maintain with minimal value-add for clients, it should be intuitive to achieve same direct with the gen code? 
for suggestion, would be good to add some unit tests which demonstrate/assert pb usage like this for client examples and a new section in [ingest/README.md](https://github.com/stellar/go/blob/master/ingest/README.md) that introduces these, their purpose, etc?

karthikiyer56: I agree. 
The helpers, specifically in address.go, are overkill. 
I will remove them.
The helpers in token_transfer event l,on the other hand, help reduce boiler plate for what i intend to do in my next body of work.

protos/processors/token_transfer/token_transfer_event.proto
```diff
@@ -0,0 +1,56 @@
...
+
+import "address/address.proto";
+import "asset/asset.proto";
+
+// EventMeta message
+message EventMeta {
+  uint32 ledgerSequence = 1;
+  google.protobuf.Timestamp closedAt = 2;
+  string txHash = 3;
+  google.protobuf.UInt32Value operationIndex = 4; // Optional for fee events or smart contract transactions
```
sreuland, line 16: as of proto3.15 the `optional` was added back in and is considered stable across all language bindings, so, could go back to that form rather than use the wrappers, it does have slightly better readability. Assuming don't need the  corner case of explicitly setting to 'null', which that is only supported by using the wrapper.

ingest/address/address.go
sreuland: sounds good!

protos/processors/token_transfer/token_transfer_event.proto
karthikiyer56: made it optional instead of using `google.protobuf.Uint32Value`

ingest/address/address.go
karthikiyer56: Done.
Added README in the protos repo. That, imo, is a more logical place to talk about code structuere, where gen files will go etc
Added unit tests for usage and ser-deser for `asset.proto`

protos/ingest/address/address.proto
```diff
@@ -0,0 +1,16 @@
...
+
+enum AddressType {
+  ADDRESS_TYPE_ACCOUNT = 0;
+  ADDRESS_TYPE_CONTRACT = 1;
+  ADDRESS_TYPE_LIQUIDITY_POOL = 2;
+  ADDRESS_TYPE_CLAIMABLE_BALANCE = 3;
+  ADDRESS_TYPE_MUXED_ACCOUNT = 4;
+}
+
+message Address {
```
karthikiyer56, line 13: This is deliberate
I realize that we are going to be using the SEP-23 strkey representation of these 5 kinds of addresses that will show up in unified strings. So all addresses could very well have been strings

however, the reason I want to have an enum in addition to the `str_key` in the `message Address` is to drive home the exact type of address, without having to refer to SEP-23.

I was tempted to simply include the hash emitted by core and have message formats for all the 5 kind of addresses.
something like
```
message AccountAddress {
    bytes hash 1;
  }

message MuxedAccount {
    uint64 id 1;
    bytes hash 2;
 }
 ....
 ....
 ```
 but that might be overkill

File: protos/ingest/processors/token_transfer/token_transfer_event.proto [hunk 3]
```diff
@@ -0,0 +1,56 @@
...
+import "google/protobuf/timestamp.proto";
+
+// Import are relative to the top level directory -  `/protos`
+import "ingest/address/address.proto";
+import "ingest/asset/asset.proto";
+
+// EventMeta message
+message EventMeta {
+  uint32 ledger_sequence = 1;
+  google.protobuf.Timestamp closed_at = 2;
```
sreuland, line 14: may be worthwhile to rename this message to `LedgerMeta` to avoid ambiguity, as all these fields are referring to attributes of ledger not event, and `closed_at` under current naming, could be interpreted as different things. maybe rename that as well to `closed`?

is `google.protobuf.Timestamp` usage to provide more client code gen date helpers for convenience? otherwise, the ledger close time in xdr only has precision down to second correct, so that could be simplified to just uint32 for unix timestsamp.

karthikiyer56: >may be worthwhile to rename this message to LedgerMeta to avoid ambiguity
Its kinda of a mix-bag, tbh. It has ledger #, txHash, operation index, contract address etc.
The reason i named it `EventMeta` was to indicate that this struct has _additional_ information to the asset movement main stuff which is in the event body.
In that sense, imo, it is aptly named.

Re choice of `google.protobuf.Timestamp` vs just uInt64 as represented in ledgerHeaderEntry....
Yes, it is for convenience.
It is also in keeping with the fact that, in all downstream code (including in ingest package), we store the ledgerCloseTime/closedAt as a `time.Time`

protos/ingest/processors/token_transfer/token_transfer_event.proto
```diff
@@ -0,0 +1,56 @@
...
+}
+
+message Fee {
+  address.Address from = 1;
+  string amount = 2;
+}
+
+message TokenTransferEvent {
+  EventMeta meta = 1;
+  asset.Asset asset = 2; // Asset can be native or issued. For custom tokens, it will be absent
```
tamirms, line 48: should this field be optional if it can be absent for custom tokens?

protos/ingest/processors/token_transfer/token_transfer_event.proto
```diff
@@ -0,0 +1,56 @@
...
+
+// Import are relative to the top level directory -  `/protos`
+import "ingest/address/address.proto";
+import "ingest/asset/asset.proto";
+
+// EventMeta message
+message EventMeta {
+  uint32 ledger_sequence = 1;
+  google.protobuf.Timestamp closed_at = 2;
+  string tx_hash = 3;
```
tamirms, line 15: I think it's unlikely but it's still worth asking the core team if it is possible for an asset to be moved outside the context of a transaction. If it turns out to be possible then tx_hash should be an optional field

tamirms (approved):

protos/ingest/processors/token_transfer/token_transfer_event.proto
karthikiyer56: We dont need to explicitly specify the `optional` prefix for custom message types
any non-primitive data type is implicitly optional 
https://github.com/stellar/go/actions/runs/13253119213/workflow?pr=5596

protos/ingest/processors/token_transfer/token_transfer_event.proto
karthikiyer56: @sisuresh -  could you help answer this question? 
Specifically - `if it is possible for an asset to be moved outside the context of a transaction`

PR #5609, opened 2025-02-19
Initial boiler plate commit for token transfer processor

karthikiyer56:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

### Why

### Known limitations

[TODO or N/A]

### DIFF
--- a/ingest/processors/token_transfer/token_transfer_processor.go
+++ b/ingest/processors/token_transfer/token_transfer_processor.go
@@ -1,0 +1,171 @@
+package token_transfer
+
+import (
+	"github.com/stellar/go/ingest"
+	"github.com/stellar/go/support/errors"
+	"github.com/stellar/go/xdr"
+	"io"
+)
+
+func ProcessTokenTransferEventsFromLedger(lcm xdr.LedgerCloseMeta, networkPassPhrase string) ([]*TokenTransferEvent, error) {
+	var events []*TokenTransferEvent
+	txReader, err := ingest.NewLedgerTransactionReaderFromLedgerCloseMeta(networkPassPhrase, lcm)
+	if err != nil {
+		return nil, errors.Wrap(err, "error creating transaction reader")
+	}
+
+	for {
+		var tx ingest.LedgerTransaction
+		var txEvents []*TokenTransferEvent
+		tx, err = txReader.Read()
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			return nil, errors.Wrap(err, "error reading transaction")
+		}
+		txEvents, err = ProcessTokenTransferEventsFromTransaction(tx)
+		if err != nil {
+			return nil, errors.Wrap(err, "error processing token transfer events from transaction")
+		}
+		events = append(events, txEvents...)
+	}
+	return events, nil
+}
+
+func ProcessTokenTransferEventsFromTransaction(tx ingest.LedgerTransaction) ([]*TokenTransferEvent, error) {
+	var events []*TokenTransferEvent
+	feeEvent := generateFeeEvent(tx)
+	events = append(events, feeEvent)
+	operations := tx.Envelope.Operations()
+	operationResults, _ := tx.Result.OperationResults()
+
+	// Ensure we only process operations if the transaction was successful
+	if tx.Result.Successful() {
+		for i := range operations {
+			op := operations[i]
+			opResult := operationResults[i]
+
+			// Process the operation and collect events
+			opEvents, err := ProcessTokenTransferEventsFromOperation(uint32(i), op, opResult, tx)
+			if err != nil {
+				return nil,
+					errors.Wrapf(err, "error processing token transfer events from operation, index: %d,  %s", i, op.Body.Type.String())
+			}
+
+			events = append(events, opEvents...)
+		}
+	}
+
+	return events, nil
+}
+
+// opIndex will be needed, on the offchance that we need to fetch ledgerEntry changes
+func ProcessTokenTransferEventsFromOperation(opIndex uint32, op xdr.Operation, opResult xdr.OperationResult, tx ingest.LedgerTransaction) ([]*TokenTransferEvent, error) {
+	switch op.Body.Type {
+	case xdr.OperationTypeCreateAccount:
+		return accountCreateEvents(op.Body.MustCreateAccountOp(), opResult.Tr.MustCreateAccountResult(), tx)
+	// AccountMerge is wierd
+	case xdr.OperationTypeAccountMerge:
+		return mergeAccountEvents(op, opResult, tx)
+	case xdr.OperationTypePayment:
+		return paymentEvents(op.Body.MustPaymentOp(), opResult.Tr.MustPaymentResult(), tx)
+	case xdr.OperationTypeCreateClaimableBalance:
+		return createClaimableBalanceEvents(op.Body.MustCreateClaimableBalanceOp(), opResult.Tr.MustCreateClaimableBalanceResult(), tx)
+	case xdr.OperationTypeClaimClaimableBalance:
+		return claimClaimableBalanceEvents(op.Body.MustClaimClaimableBalanceOp(), opResult.Tr.MustClaimClaimableBalanceResult(), tx)
+	case xdr.OperationTypeClawback:
+		return clawbackEvents(op.Body.MustClawbackOp(), opResult.Tr.MustClawbackResult(), tx)
+	case xdr.OperationTypeClawbackClaimableBalance:
+		return clawbackClaimableBalanceEvents(op.Body.MustClawbackClaimableBalanceOp(), opResult.Tr.MustClawbackClaimableBalanceResult(), tx)
+	case xdr.OperationTypeAllowTrust:
+		return allowTrustEvents(op.Body.MustAllowTrustOp(), opResult.Tr.MustAllowTrustResult(), tx)
+	case xdr.OperationTypeSetTrustLineFlags:
+		return setTrustLineFlagsEvents(op.Body.MustSetTrustLineFlagsOp(), opResult.Tr.MustSetTrustLineFlagsResult(), tx)
+	case xdr.OperationTypeLiquidityPoolDeposit:
+		return liquidityPoolDepositEvents(op.Body.MustLiquidityPoolDepositOp(), opResult.Tr.MustLiquidityPoolDepositResult(), tx)
+	case xdr.OperationTypeLiquidityPoolWithdraw:
+		return liquidityPoolWithdrawEvents(op.Body.MustLiquidityPoolWithdrawOp(), opResult.Tr.MustLiquidityPoolWithdrawResult(), tx)
+	case xdr.OperationTypeManageBuyOffer:
+		return manageBuyOfferEvents(op.Body.MustManageBuyOfferOp(), opResult.Tr.MustManageBuyOfferResult(), tx)
+	case xdr.OperationTypeManageSellOffer:
+		return manageSellOfferEvents(op.Body.MustManageSellOfferOp(), opResult.Tr.MustManageSellOfferResult(), tx)
+	case xdr.OperationTypeCreatePassiveSellOffer:
+		return createPassiveSellOfferEvents(op.Body.MustCreatePassiveSellOfferOp(), opResult.Tr.MustCreatePassiveSellOfferResult(), tx)
+	case xdr.OperationTypePathPaymentStrictSend:
+		return pathPaymentStrictSendEvents(op.Body.MustPathPaymentStrictSendOp(), opResult.Tr.MustPathPaymentStrictSendResult(), tx)
+	case xdr.OperationTypePathPaymentStrictReceive:
+		return pathPaymentStrictReceiveEvents(op.Body.MustPathPaymentStrictReceiveOp(), opResult.Tr.MustPathPaymentStrictReceiveResult(), tx)
+	default:
+		return nil, nil
+	}
+}
+
+func generateFeeEvent(tx ingest.LedgerTransaction) *TokenTransferEvent {
+	return nil
+}
+
+// Function stubs
+func accountCreateEvents(op xdr.CreateAccountOp, result xdr.CreateAccountResult, tx ingest.LedgerTransaction) ([]*TokenTransferEvent, error) {
+	return nil, nil
+}
+
+func mergeAccountEvents(op xdr.Operation, result xdr.OperationResult, tx ingest.LedgerTransaction) ([]*TokenTransferEvent, error) {
+	return nil, nil
+}
+
+func paymentEvents(op xdr.PaymentOp, result xdr.PaymentResult, tx ingest.LedgerTransaction) ([]*TokenTransferEvent, error) {
+	return nil, nil
+}
+
+func createClaimableBalanceEvents(op xdr.CreateClaimableBalanceOp, result xdr.CreateClaimableBalanceResult, tx ingest.LedgerTransaction) ([]*TokenTransferEvent, error) {
+	return nil, nil
+}
+
+func claimClaimableBalanceEvents(op xdr.ClaimClaimableBalanceOp, result xdr.ClaimClaimableBalanceResult, tx ingest.LedgerTransaction) ([]*TokenTransferEvent, error) {
+	return nil, nil
+}
+
+func clawbackEvents(op xdr.ClawbackOp, result xdr.ClawbackResult, tx ingest.LedgerTransaction) ([]*TokenTransferEvent, error) {
+	return nil, nil
+}
+
+func clawbackClaimableBalanceEvents(op xdr.ClawbackClaimableBalanceOp, result xdr.ClawbackClaimableBalanceResult, tx ingest.LedgerTransaction) ([]*TokenTransferEvent, error) {
+	return nil, nil
+}
+
+func allowTrustEvents(op xdr.AllowTrustOp, result xdr.AllowTrustResult, tx ingest.LedgerTransaction) ([]*TokenTransferEvent, error) {
+	return nil, nil
+}
+
+func setTrustLineFlagsEvents(op xdr.SetTrustLineFlagsOp, result xdr.SetTrustLineFlagsResult, tx ingest.LedgerTransaction) ([]*TokenTransferEvent, error) {
+	return nil, nil
+}
+
+func liquidityPoolDepositEvents(op xdr.LiquidityPoolDepositOp, result xdr.LiquidityPoolDepositResult, tx ingest.LedgerTransaction) ([]*TokenTransferEvent, error) {
+	return nil, nil
+}
+
+func liquidityPoolWithdrawEvents(op xdr.LiquidityPoolWithdrawOp, result xdr.LiquidityPoolWithdrawResult, tx ingest.LedgerTransaction) ([]*TokenTransferEvent, error) {
+	return nil, nil
+}
+
+func manageBuyOfferEvents(op xdr.ManageBuyOfferOp, result xdr.ManageBuyOfferResult, tx ingest.LedgerTransaction) ([]*TokenTransferEvent, error) {
+	return nil, nil
+}
+
+func manageSellOfferEvents(op xdr.ManageSellOfferOp, result xdr.ManageSellOfferResult, tx ingest.LedgerTransaction) ([]*TokenTransferEvent, error) {
+	return nil, nil
+}
+
+func createPassiveSellOfferEvents(op xdr.CreatePassiveSellOfferOp, result xdr.ManageSellOfferResult, tx ingest.LedgerTransaction) ([]*TokenTransferEvent, error) {
+	return nil, nil
+}
+
+func pathPaymentStrictSendEvents(op xdr.PathPaymentStrictSendOp, result xdr.PathPaymentStrictSendResult, tx ingest.LedgerTransaction) ([]*TokenTransferEvent, error) {
+	return nil, nil
+}
+
+func pathPaymentStrictReceiveEvents(op xdr.PathPaymentStrictReceiveOp, result xdr.PathPaymentStrictReceiveResult, tx ingest.LedgerTransaction) ([]*TokenTransferEvent, error) {
+	return nil, nil
+}


amount/main.go
```diff
@@ -133,6 +133,15 @@ func String128(v xdr.Int128Parts) string {
 	return rat.FloatString(7)
 }
 
+func String128Raw(v xdr.Int128Parts) string {
```
karthikiyer56, line 136: Ignore for now

ingest/processors/token_transfer/token_transfer_processor.go
```diff
@@ -0,0 +1,331 @@
...
+	for _, change := range feeChanges {
+		if change.Type != xdr.LedgerEntryTypeAccount {
+			return nil, errors.Errorf("invalid ledgerEntryType for fee change: %s", change.Type.String())
+		}
+
+		// Do I need to do all this? Can I not simply use tx.Result.Result.FeeCharged
+		preBalance := change.Pre.Data.MustAccount().Balance
+		postBalance := change.Post.Data.MustAccount().Balance
+		accId := change.Pre.Data.MustAccount().AccountId
+		amt := amount.String(postBalance - preBalance)
```
tamirms, line 122: I think this will be negative

ingest/processors/token_transfer/token_transfer_processor.go
```diff
@@ -0,0 +1,331 @@
...
+	return events, nil
+}
+
+func ProcessTokenTransferEventsFromTransaction(tx ingest.LedgerTransaction) ([]*TokenTransferEvent, error) {
+	var events []*TokenTransferEvent
+	feeEvents, err := generateFeeEvent(tx)
+	if err != nil {
+		return nil, errors.Wrap(err, "error generating fee event")
+	}
+	events = append(events, feeEvents...)
```
tamirms, line 45: you should confirm the ordering of the fee event with Siddharth. We will want to be consistent with cap-67 as much as possible and I'm not sure if in cap 67 the fee event occurs first or after all the events emitted by the operation

ingest/processors/token_transfer/token_transfer_processor.go
karthikiyer56: Argh.. you are right
I didnt write a unit test for this yet. Otherwise it wud have been caught. Ima write it soon

ingest/ledger_transaction.go
```diff
@@ -592,31 +596,6 @@ func (t *LedgerTransaction) AccountMuxed() (string, bool) {
 
 }
 
-func (t *LedgerTransaction) FeeAccount() (string, bool) {
```
karthikiyer56, line 595: The name of the function seems to suggest that this function will give the underlying fee account - be it a regular transaction or a regular transaction.
IMO, this might be quite misleading. I found that behaved contrary to what the name suggested when my code was not working as expected.

Besides, I dont see it being used anywhere other than in a sample unit test.
I am removing this function and the function below so as to not cause further confusion

ingest/ledger_transaction_test.go
```diff
@@ -928,16 +928,6 @@ func TestTransactionHelperFunctions(t *testing.T) {
 	assert.Equal(t, true, ok)
 	assert.Equal(t, "MAISEMYAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAPMJ2I", accountMuxed)
 
-	var feeAccount string
```
karthikiyer56, line 931: Tests removed for the now removed FeeAccount() function, from above comment

ingest/processors/token_transfer/token_transfer_processor.go
karthikiyer56: Isnt that a non-issue?
I'd imagine that the "feeEvent"  will be a a top transaction-level field - something like TransactionMeta
whereas operationEvents will be in the in the OperationMeta. 

Which means - once CAP-67 goes out, we (processor code_ will still need to parse the events for fee and per operation. 
Even there, we can ensure that we follow the same order -  fee event first and then operation events.

In that sense,from core's perpective  - they dont impose any ordering on feeEvent

ingest/processors/token_transfer/token_transfer_processor_test.go
```diff
@@ -0,0 +1,562 @@
+package token_transfer
+
+import (
+	"github.com/stellar/go/ingest"
+	addressProto "github.com/stellar/go/ingest/address"
```
karthikiyer56, line 5: This file is the real deal and needs to be reviewed thoroughly.
Adding a comment to ensure that this doesnt slip up - since the load diff might not render the file.

ingest/processors/token_transfer/token_transfer_processor_test.go
```diff
@@ -0,0 +1,562 @@
...
+	"github.com/stellar/go/xdr"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/protobuf/proto"
+	"strings"
+	"testing"
+)
+
+var (
+	someTxAccount = xdr.MustMuxedAddress("GBF3XFXGBGNQDN3HOSZ7NVRF6TJ2JOD5U6ELIWJOOEI6T5WKMQT2YSXQ")
```
karthikiyer56, line 16: There are several variables here that I am using throughout the test, and so they have been clubbed here at the top.
They, IMHO, improve the code readability when reviewing the tests, and I am not inclined to move them to individual tests, but rather have them at the top-level here

ingest/processors/token_transfer/token_transfer_processor.go
karthikiyer56: never mind this bug.
reworked logic to use `FeeCharged in64` in the TxResult to generate the feeEvent

ingest/processors/token_transfer/token_transfer_processor.go
```diff
@@ -0,0 +1,485 @@
...
+	for _, change := range changes {
+		if change.Type != xdr.LedgerEntryTypeClaimableBalance {
+			continue
+		}
+		if change.Pre != nil && change.Post == nil {
+			cb = change.Pre.Data.MustClaimableBalance()
+
+			if cb.BalanceId.MustV0().Equals(cbId.MustV0()) {
+				found = true
+				break
```
tamirms, line 252: you can return the claimable balance here and that way you don't have to maintain the `found` variable

services/horizon/internal/integration/change_test.go
```diff
@@ -18,6 +22,240 @@ import (
 	"github.com/stellar/go/xdr"
 )
 
+func TestLiquidityPoolHappyPath2(t *testing.T) {
```
karthikiyer56, line 25: the changes in this test are purely for PoC.
will be removed later

ingest/processors/token_transfer/token_transfer_processor.go
karthikiyer56: this has been reworked completely

ingest/processors/token_transfer/token_transfer_processor.go
```diff
@@ -0,0 +1,743 @@
...
+	event := mintOrBurnOrTransferEvent(cb.Asset, from, to, amount.String(cb.Amount), meta)
+	return []*TokenTransferEvent{event}, nil
+}
+
+func clawbackEvents(tx ingest.LedgerTransaction, opIndex uint32, op xdr.Operation) ([]*TokenTransferEvent, error) {
+	clawbackOp := op.Body.MustClawbackOp()
+	meta := NewEventMeta(tx, &opIndex, nil)
+
+	// fromAddress is NOT the operationSourceAccount.
+	// It is the account specified in the operation from whom you want money to be clawed back
```
tamirms, line 338: in [cap-67](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0067.md#clawback--clawback-claimable-balance) it says:

> `from` is the account or claimable balance being credited.

I think that's probably a typo but worth double checking with the core team

ingest/processors/token_transfer/token_transfer_processor.go
```diff
@@ -0,0 +1,743 @@
...
+			Type: xdr.ClaimableBalanceIdTypeClaimableBalanceIdTypeV0,
+			V0:   &sha256hash,
+		}
+		generatedClaimableBalanceIds = append(generatedClaimableBalanceIds, cbId)
+	}
+
+	return generatedClaimableBalanceIds, nil
+}
+
+func getClaimableBalanceEntriesFromOperationChanges(tx ingest.LedgerTransaction, opIndex uint32) ([]xdr.ClaimableBalanceEntry, error) {
```
tamirms, line 277: this function is used to find claimable balances that were either created or removed. however, the caller expects the returned list to never have a mix between the two types. I think it would be better to add a `xdr.LedgerEntryChangeType` parameter which can be used to filter for the exact change types desired by the caller (e.g. only created claimable balances or only removed claimable balances)

ingest/processors/token_transfer/token_transfer_processor.go
```diff
@@ -0,0 +1,743 @@
...
+		There is no concrete way to tie which claimable balances(either 1 or 2) were created in response to which liquidity pool.
+		So, it is difficult to generate the from, to in the transfer event.
+		So, we need to re-implement the code logic to derive claimableBalanceIds from liquidity pool
+		and then do additional logic to compare it with the claimableBalances that were actually created (i,e the ones that show up in the operation changes)
+
+		It would be nice, if somehow, core could trickle this information up in the ledger entry changes for the claimable balances created - something like createdBy in the extension field.
+	*/
+	createdCbIdToCreatorLpId := make(map[string]string)
+	for _, lp := range impactedLiquidityPools {
+		// This `claimableBalancesCreated` slice will have exactly 2 entries - one for each asset in the LP
```
tamirms, line 412: will it always have 2 entries? I thought it was possible for there to be only 1 entry?

karthikiyer56: this is the code path where I am decidedly creating 2 CbIds (artificially) for me to compare the actually created ones from the ledgerEntryChanges.

so, yes, there will be exactly 2 CBs here always.
The actual check for 1 vs 2, created in real life is about 20 lines below this LoC

ingest/processors/token_transfer/token_transfer_processor.go
```diff
@@ -0,0 +1,743 @@
...
+		}
+
+		entry.amountChangeForAssetA = abs64(postA - preA)
+		entry.amountChangeForAssetB = abs64(postB - preB)
+		entries = append(entries, entry)
+	}
+
+	if len(entries) == 0 {
+		return nil, ErrNoLiquidityPoolEntryFound
+	}
```
tamirms, line 553: nit: I think it's better style to return the empty list without an error so the caller can decide how the empty list should be handled. it's not obvious from the name of the function that at least 1 liquidity pool should be present and anything otherwise would be an error

ingest/processors/token_transfer/token_transfer_processor.go
```diff
@@ -0,0 +1,743 @@
...
+			entries = append(entries, cb)
+		} else if change.Post != nil && change.Pre == nil { // check if claimable balance entry is created
+			cb = change.Post.Data.MustClaimableBalance()
+			entries = append(entries, cb)
+		}
+	}
+
+	if len(entries) == 0 {
+		return nil, ErrNoClaimableBalanceEntryFound
+	}
```
tamirms, line 307: nit, same comment as ErrNoLiquidityPoolEntryFound:

I think it's better style to return the empty list without an error so the caller can decide how the empty list should be handled.

ingest/processors/token_transfer/token_transfer_processor.go
karthikiyer56: marking this as closed, based on my response.
please do re-open if you disagree or if there is more to it than what I mentioned in my previous comment

ingest/processors/token_transfer/token_transfer_processor.go
karthikiyer56: yes, it is a typo in cap-67 writeup.
this is being addressed by sid in his most recent pr

ingest/processors/token_transfer/token_transfer_processor.go
karthikiyer56: Seeing that this is an internal function, I see little chance for misuse/mis-representation of what this function is intended to do.
at any rate, I have changed the function to accept a `changeType xdr.LedgerEntryChangeType` and added a comment at the top.
callers of the function will either pass - `LedgerEntryChangeTypeLedgerEntryRemoved` or `LedgerEntryChangeTypeLedgerEntryCreated`. It will fail when anything else is passed

ingest/processors/token_transfer/token_transfer_processor.go
karthikiyer56: reworked to do error/length check at call-site instead of inside the function

ingest/processors/token_transfer/token_transfer_processor.go
karthikiyer56: reworked to do error/length check at call-site instead of inside the function

ingest/processors/token_transfer/token_transfer_processor.go
```diff
@@ -0,0 +1,735 @@
...
+	return generateEventsForRevokedTrustlines(tx, opIndex)
+}
+
+func setTrustLineFlagsEvents(tx ingest.LedgerTransaction, opIndex uint32, op xdr.Operation) ([]*TokenTransferEvent, error) {
+	// ?? Should I be checking for generation of liquidity pools and CBs iff the flag is set to false?
+	// isAuthRevoked := op.Body.MustSetTrustLineFlagsOp().ClearFlags != 0
+	return generateEventsForRevokedTrustlines(tx, opIndex)
+}
+
+func generateEventsForRevokedTrustlines(tx ingest.LedgerTransaction, opIndex uint32) ([]*TokenTransferEvent, error) {
```
karthikiyer56, line 362: this is the dooziest logic of all the code in this file.
I have written unit tests that test this ad-nauseum
This is the only part of the code where I have felt the need to be over-liberal with comments to indicate what the code is doing, since it is very non-trivial how the Cbs are generated from LPs when trustline revocation of an asset happens for a trustor account

support/converters/main.go
```diff
@@ -0,0 +1,43 @@
+package converters
+
+import (
+	"crypto/sha256"
+	"github.com/stellar/go/support/errors"
+	"github.com/stellar/go/xdr"
+)
+
+/*
+Helper function to convert LiquidityPoolId to generate ClaimableBalanceIds in a deterministic way as per
```
karthikiyer56, line 10: I needed a place to park this generic logic that generates CBids from LPids.
I use this in the code and also in the unit tests.

It didnt make sense to subsume it under `token_transfer_processor`, so parking it here under a new directory  -`converters` under `support`
Happy to move this to another place, so long as this is a public function.

tamirms: I would put it either in the xdr package or along with the xdrill code in the ingest package. Also, I suggest using a more concise name for the function, maybe,  ClaimableBalanceIdFromRevocation or BalanceIdFromRevocation?

ingest/processors/token_transfer/token_transfer_processor.go
```diff
@@ -0,0 +1,736 @@
...
+		generatedClaimableBalanceIds = append(generatedClaimableBalanceIds, cbId)
+	}
+	return generatedClaimableBalanceIds, nil
+}
+
+// This operation is used to only find CB entries that are either created or deleted, not updated
+func getClaimableBalanceEntriesFromOperationChanges(changeType xdr.LedgerEntryChangeType, tx ingest.LedgerTransaction, opIndex uint32) ([]xdr.ClaimableBalanceEntry, error) {
+	if !(changeType == xdr.LedgerEntryChangeTypeLedgerEntryRemoved || changeType == xdr.LedgerEntryChangeTypeLedgerEntryCreated) {
+		return nil, fmt.Errorf("changeType: %v, not allowed", changeType)
+	}
```
tamirms, line 265: I don't think this error is necessary. The function will obtain all cb changes occurring from an operation filtered by a change type. The fact that the function happens to be called only for removed or created changes is something determined by the caller

ingest/processors/token_transfer/token_transfer_processor.go
```diff
@@ -0,0 +1,736 @@
...
+		}
+		// Check if claimable balance entry is deleted
+		//?? maybe it is not necessary to check change.Pre != nil && change.Post since that will be true for deleted entries?
+		if changeType == xdr.LedgerEntryChangeTypeLedgerEntryRemoved && change.Pre != nil && change.Post == nil {
+			cb = change.Pre.Data.MustClaimableBalance()
+			entries = append(entries, cb)
+		} else if changeType == xdr.LedgerEntryChangeTypeLedgerEntryCreated && change.Post != nil && change.Pre == nil { // check if claimable balance entry is created
+			cb = change.Post.Data.MustClaimableBalance()
+			entries = append(entries, cb)
+		}
```
tamirms, line 292:
```suggestion
		if change.Type != xdr.LedgerEntryTypeClaimableBalance || change.LedgerEntryChangeType() != changeType {
			continue
		}
		if change.Pre != nil {
			cb = change.Pre.Data.MustClaimableBalance()
			entries = append(entries, cb)
		} else if change.Post != nil {
			cb = change.Post.Data.MustClaimableBalance()
			entries = append(entries, cb)
		}
```

ingest/processors/token_transfer/token_transfer_processor.go
tamirms: you could simplify the if statement condition to:

```suggestion
	if changeType == xdr.LedgerEntryChangeTypeLedgerEntryUpdated {
		return nil, fmt.Errorf("LEDGER_ENTRY_UPDATED is not a valid filter")
	}
```

ingest/processors/token_transfer/token_transfer_processor.go
```diff
@@ -0,0 +1,736 @@
...
+		return nil, fmt.Errorf("more than one Liquidiy pool entry found for operation: %s", op.Body.Type.String())
+	}
+
+	delta := lpDeltas[0]
+	lpId := delta.liquidityPoolId
+	assetA, assetB := delta.assetA, delta.assetB
+	amtA, amtB := delta.amountChangeForAssetA, delta.amountChangeForAssetB
+	if amtA <= 0 {
+		return nil,
+			fmt.Errorf("deposited amount (%v) for asset: %v, cannot be negative in LiquidityPool: %v", amtA, assetA.String(), lpIdToStrkey(lpId))
```
tamirms, line 561: error string indicates negative values are not allowed but the if statement is actually including 0 values as well in the illegal state

ingest/processors/token_transfer/token_transfer_processor.go
```diff
@@ -0,0 +1,736 @@
...
+	"github.com/stellar/go/support/converters"
+	"github.com/stellar/go/support/errors"
+	"github.com/stellar/go/xdr"
+	"io"
+)
+
+var (
+	xlmProtoAsset                   = assetProto.NewNativeAsset()
+	ErrNoLiquidityPoolEntryFound    = errors.New("no liquidity pool entry found in operation changes")
+	ErrNoClaimableBalanceEntryFound = errors.New("no claimable balance entry found in operation changes")
```
tamirms, line 18: this is no longer used

ingest/processors/token_transfer/token_transfer_processor.go
tamirms: I reviewed the logic and it looks correct to me. I think the code would be easier to understand if you refactored it to avoid iterating over `impactedLiquidityPools` twice, but that's just my personal opinion. Here's my attempt to simplify the code:

https://github.com/tamirms/go/commit/9e14bdef0312042227a2655f6683e635c034905e

feel free to take or reject whichever parts you see fit

ingest/processors/token_transfer/token_transfer_processor.go
```diff
@@ -0,0 +1,736 @@
...
+	}
+}
+
+func cbIdToStrkey(cb xdr.ClaimableBalanceId) string {
+	return cb.MustV0().HexString()
+}
+
+func lpIdToStrkey(lpId xdr.PoolId) string {
+	return xdr.Hash(lpId).HexString()
+}
```
tamirms, line 736: this file is quite long, it would be easier to read the code if the code in this file was split into multiple files. maybe this could be refactored into three files:

- one for path payment and offer related operations
- one for revocation and liquidity pool operations
- one for everything else

ingest/ledger_transaction.go
tamirms: I agree with your assessment but we should sync with @chowbao to make sure etl isn't impacted by this change

karthikiyer56: @chowbao - i have removed 2 functions added by you in a previous pr. 
The rationale is ⬆️ 
Afaict, that code was added by you as a part of xdrill. 
I dont think it is used in stellar-etl yet right? Looking for confirmation on that part
I have removed the occurrences of this from the go monorepo.

chowbao: Yes removal of anything in the processors is good from the stellar-etl side right now.

Stellar-etl does not use the processors rgiht now

support/converters/main.go
karthikiyer56: Happy to rename the function. 
I don't quite agree about putting it under xdr. Somehow it doesn't feel right, since most of the code under xdr is receiver functions, and this function doesnt exactly lend itself to be a receiver function
If I have to, where would you recommend?
`xdr/claimable_balance_id.go` ?
There is no xdrill directory per se. 
Everything is under their own directory under `ingest/processors`, and i dont think it belongs there either. 

I am simply inclined to leave it here for now, if it's not too contentious

ingest/processors/token_transfer/token_transfer_processor.go
```diff
@@ -0,0 +1,736 @@
...
+		cbId, err := converters.ConvertLiquidityPoolIdToClaimableBalanceId(lpId, asset, seqNum, txSrcAccount, opIndex)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to generate claimable balance id from LiquidityPoolId: %v, for asset: %v", lpIdToStrkey(lpId), asset.String())
+		}
+		generatedClaimableBalanceIds = append(generatedClaimableBalanceIds, cbId)
+	}
+	return generatedClaimableBalanceIds, nil
+}
+
+// This operation is used to only find CB entries that are either created or deleted, not updated
```
sydneynotthecity, line 261: I don't think you can update claimable balances. I just checked Hubble and there are only create and delete ledger entry change types, there are no updates. I think it's safe to remove the initial update check because i don't think it'll ever trigger 🤷‍♀️

tamirms: a claimable balance update is theoretically possible if the sponsorship of a claimable balance is transferred from one account to another. This happened once on testnet 😅 https://stellarfoundation.slack.com/archives/C02B04RMK/p1710427955467749

ingest/ledger_transaction.go
karthikiyer56: sweet. marking this as resolved

ingest/processors/token_transfer/token_transfer_processor.go
karthikiyer56: yup, it still is
Checkout `claimClaimableBalanceEvents()` and `clawbackClaimableBalanceEvents()`
I moved the len(cbs) == 0 check at the callsite, and return this error there

ingest/processors/token_transfer/token_transfer_processor.go
karthikiyer56: marking this as resolved

ingest/processors/token_transfer/token_transfer_processor.go
karthikiyer56: this has been reworked based on the recommendations that appear across multiple review comments.
marking this as resolved

support/converters/main.go
karthikiyer56: > I'm curious why this wouldn't just live under processors/token_transfer?

because I did not think about it.  🙈 
Fair point, i can move it under `processors/token_transfer`

karthikiyer56: done.
moved it to under a `helper.go` in the same package

Also renamed it to `ClaimableBalanceIdFromRevocation`

ingest/processors/token_transfer/token_transfer_processor.go
karthikiyer56: done.
updated as per suggestion

ingest/processors/token_transfer/token_transfer_processor.go
karthikiyer56: I actually dont agree with breaking this file into multiple files, because it is long. 

More specifically, even if I were to do so, it is the `one for everything else` that I dont know how to club.

Agreed that this file is long. But, i'd argue that it is readable since it is quite modular in terms of - having a function for each operation and such.

Breaking the file along the lines of grouping functions like you suggested might be overkill, since there are helper functions too, and I am not sure where those would go.

protos/ingest/address/address.proto
```diff
@@ -2,15 +2,34 @@ syntax = "proto3";
...
-  ADDRESS_TYPE_LIQUIDITY_POOL = 2;
-  ADDRESS_TYPE_CLAIMABLE_BALANCE = 3;
-  ADDRESS_TYPE_MUXED_ACCOUNT = 4;
+message Account {
+  string str_key = 1;
 }
 
-message Address {
-  AddressType address_type = 1;
+message ClaimableBalance {
```
karthikiyer56, line 9: @tamirms - I want to draw your attention to this part and the generated `address.pb.go`
I changed this for good reason, and the reason was the latest way the MUXED addresses are going to be added in CAP-67. - https://github.com/stellar/stellar-xdr/commit/515cb63a99ad079519c59a35097b9c1e3d3deea1

I want to ensure that users of this library dont have to additionally unpack the M address to get the memoId
To that end, I had to redo the entire Address proto to be a oneOf, instead of just using strkeys and enum for AddressType.

Also ended up adding a few receiver functions in `address.go` to help with address generation and equality check for the unit test

tamirms: > I want to ensure that users of this library dont have to additionally unpack the M address to get the memoId

could that be achieved by implementing a receiver helper function in address.go ?

ingest/processors/token_transfer/token_transfer_processor.go
```diff
@@ -0,0 +1,734 @@
...
+
+func generateFeeEvent(tx ingest.LedgerTransaction) ([]*TokenTransferEvent, error) {
+	/*
+		For a feeBump transaction, this will be the outer transaction.
+		FeeAccount() gives the proper "muxed" account that paid the fees.
+		And we want the "muxed" Account, so that it can be passed directly to protoAddressFromAccount
+	*/
+	feeAccount := tx.FeeAccount()
+	// FeeCharged() takes care of a bug in an intermediate protocol release. So using that
+	feeAmt, _ := tx.FeeCharged()
```
tamirms, line 135: I'm not sure about the implication of the boolean return value being false. if `tx.FeeCharged()` returns false we might need return an error. We can confirm with @chowbao who implemented `FeeCharged()`

ingest/processors/token_transfer/verify_events.go
```diff
@@ -0,0 +1,353 @@
...
+
+type balanceKey struct {
+	holder string
+	asset  string
+}
+
+type keyAndAmount struct {
+	key balanceKey
+	amt int64
+}
```
tamirms, line 21: I think this intermediate representation is unnecessary. Instead, I would suggest  passing in the `map[balanceKey]int64` value to all the `fetch...FromChange()` functions and each function can update the map accordingly. This change would make the calculation of deltas from changes more similar to the  calculation of deltas from events which does not use the `keyAndAmount` intermediate representation.

I think the code would be easier to follow because removing `keyAndAmount` is one less abstraction you need to keep in your mind while reading the code. It is also easier to read the code when there is symmetry between the code which processes changes and the code which processes events

ingest/processors/token_transfer/verify_events.go
```diff
@@ -0,0 +1,353 @@
...
+	s[i], s[j] = s[j], s[i]
+}
+
+// Function to print map in sorted order of keys
+func printMap(m map[balanceKey]int64) {
+	// Extract the keys from the map and sort them
+	var keys []balanceKey
+	for key := range m {
+		keys = append(keys, key)
+	}
```
tamirms, line 344: in go 1.23 there is a `maps.Keys()` function which pulls out all the keys from a map:

https://pkg.go.dev/maps#Keys

you can use it here

tamirms: after you get the keys, you can use https://pkg.go.dev/slices#SortedFunc to sort them. I prefer passing in a comparison function because it's less verbose compared to implementing a `balanceKeySlice` type just for the purpose of sorting

ingest/processors/token_transfer/verify_events.go
```diff
@@ -0,0 +1,353 @@
...
+		}
+		if err != nil {
+			panic(errors.Wrapf(err, "unable to read from ledger"))
+		}
+		changes = append(changes, change)
+	}
+	return changes
+}
+
+func mapsEqual(map1, map2 map[balanceKey]int64) bool {
```
tamirms, line 283: you can use https://pkg.go.dev/maps#Equal

ingest/processors/token_transfer/verify_events.go
```diff
@@ -0,0 +1,353 @@
...
+
+	for _, event := range events {
+
+		if event.Asset == nil { // needed toKey check for custom token events which wont have an asset
+			continue
+		}
+
+		eventType := event.GetEventType()
+		switch eventType {
+		case "Fee":
```
tamirms, line 189: we should have constant variables for the event type strings so they can be referenced here and other places which reference event types

ingest/processors/token_transfer/token_transfer_processor.go
tamirms: since there is a helpers.go file in this package now, you can move all the helper functions in this file there and that would reduce the size of this file

ingest/processors/token_transfer/token_transfer_processor.go
karthikiyer56: reworked thist to check for the boolean and error out, if it isnt

protos/ingest/address/address.proto
karthikiyer56: reowrked this to just have one field - strkey

ingest/processors/token_transfer/verify_events.go
karthikiyer56: Yeah, I want to retain this, since I will use this to print a delta, assuming we find that verifcation fails for some ledgers

ingest/processors/token_transfer/verify_events.go
karthikiyer56: done.
reworked this as per recommendation

ingest/processors/token_transfer/verify_events.go
```diff
@@ -0,0 +1,353 @@
...
+			ev := event.GetFee()
+			address := ev.From.StrKey
+			asset := xlmAsset.StringCanonical()
+			amt := int64(amount.MustParse(ev.Amount))
+			// Address' balance reduces by amt in FEE
+			entry := balanceKey{holder: address, asset: asset}
+			hashmap[entry] += -amt
+			if hashmap[entry] == 0 {
+				delete(hashmap, entry)
+			}
```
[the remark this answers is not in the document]
karthikiyer56: done
pulled this out into a separate function

ingest/processors/token_transfer/verify_events.go
karthikiyer56: you are right.
i removed it

ingest/processors/token_transfer/token_transfer_processor.go
```diff
@@ -0,0 +1,839 @@
+package token_transfer
+
+import (
+	"fmt"
+	"github.com/stellar/go/amount"
+	"github.com/stellar/go/ingest"
+	addressProto "github.com/stellar/go/ingest/address"
+	assetProto "github.com/stellar/go/ingest/asset"
+	"github.com/stellar/go/support/errors"
```
tamirms, line 9: we should avoid using this package as the go standard library now has support for wrapping / unwrapping errors.


instead of errors.New() from github.com/stellar/go/support/errors we should prefer https://pkg.go.dev/errors#New and instead of errors.Wrap() from github.com/stellar/go/support/errors we should be wrapping errors using fmt.Errorf in the way described here, https://pkg.go.dev/errors#pkg-overview

ingest/processors/token_transfer/token_transfer_processor.go
```diff
@@ -0,0 +1,839 @@
...
+	return events, nil
+}
+
+// ProcessTokenTransferEventsFromTransaction processes token transfer events for all operations within a given transaction.
+//
+//	First, it generates a FeeEvent for the transaction
+//	If the transaction was successful, it processes all operations in the transaction by calling ProcessTokenTransferEventsFromOperationAndOperationResult for each operation in the transaction.
+//
+// If the transaction is unsuccessful, it only generates events for transaction fees.
+func (ttp *TokenTransferProcessor) ProcessTokenTransferEventsFromTransaction(tx ingest.LedgerTransaction) ([]*TokenTransferEvent, error) {
```
tamirms, line 86: nit: `EventsFromTransaction()` is a shorter name that communicates the same message imo

ingest/processors/token_transfer/token_transfer_processor.go
```diff
@@ -0,0 +1,839 @@
...
+}
+
+// ProcessTokenTransferEventsFromOperationAndOperationResult processes token transfer events for a given operation within a transaction.
+// It operates at the operation level, analyzing the operation type and generating corresponding token transfer events.
+// If the operation is successful, it processes the event based on the operation type (e.g., payment, account creation, etc.).
+// It handles various operation types like payments, account merges, trust line modifications, and more.
+// There is a separate private function to derive events for each classic operation.
+// It is implicitly assumed that the operation is successful, and thus will contribute towards generating events.
+// which is why we dont check for the success code in the OperationResult
+func (ttp *TokenTransferProcessor) ProcessTokenTransferEventsFromOperationAndOperationResult(tx ingest.LedgerTransaction, opIndex uint32, op xdr.Operation, opResult xdr.OperationResult) ([]*TokenTransferEvent, error) {
```
tamirms, line 125: nit: `EventsFromOperation()` is a shorter name that communicates the same message imo

ingest/processors/token_transfer/token_transfer_processor.go
```diff
@@ -0,0 +1,839 @@
...
+		}
+		return a
+	}
+)
+
+type TokenTransferProcessor struct {
+	networkPassphrase string
+}
+
+func NewTokenTransferProcessor(networkPassphrase string) *TokenTransferProcessor {
```
tamirms, line 45: since the package name is "token_transfer" it is better to omit "TokenTransfer" from these public types (see https://google.github.io/styleguide/go/decisions#repetitive-with-package ).

With that in mind I would suggest renaming `TokenTransferProcessor` to `Processor` or `EventsProcessor`

and then rename the `NewTokenTransferProcessor()` accordingly.

Also, the `TokenTransferEvent` proto message should be renamed to `Event`

ingest/processors/token_transfer/token_transfer_processor.go
```diff
@@ -0,0 +1,839 @@
...
+	}
+	opSrcAcc := operationSourceAccount(tx, op)
+	destAcc := op.Body.MustDestination()
+	amt := amount.String(*res.SourceAccountBalance)
+	meta := NewEventMeta(tx, tx.Index-1, &opIndex, nil)
+	event := NewTransferEvent(meta, protoAddressFromAccount(opSrcAcc), protoAddressFromAccount(destAcc), amt, xlmProtoAsset)
+	return []*TokenTransferEvent{event}, nil // Just one event will be generated
+}
+
+type addressWrapper struct {
```
tamirms, line 205: it seems like this type is used to store values before ultimately  converting them to `*addressProto.Address` instances. So, why not remove this intermediary and whenever you create an `addressWrapper` instance you instead create a `*addressProto.Address` instance?

ingest/processors/token_transfer/token_transfer_processor.go
```diff
@@ -0,0 +1,839 @@
...
+	}
+
+	meta := NewEventMeta(tx, tx.Index-1, &opIndex, nil)
+	opSrcAcc := operationSourceAccount(tx, op)
+	// From = operation source account, to = LP
+	from, to := addressWrapper{account: &opSrcAcc}, addressWrapper{liquidityPoolId: &delta.liquidityPoolId}
+
+	var events []*TokenTransferEvent
+	// I am not sure if it is possible for amtA or amtB to be ever 0, for e,g when LpDeposit updates the amount for just 1 asset in an already existing LP
+	// So, out of abundance of caution, I will generate the event only if the amounts are greater than 0
```
tamirms, line 635: we should confirm with the core team if this behavior will be consistent with their implementation of cap-67

ingest/processors/token_transfer/verify_events.go
```diff
@@ -0,0 +1,289 @@
...
+	for key, value := range map1 {
+		if val, exists := map2[key]; !exists || val != value {
+			return false
+		}
+	}
+
+	return true
+}
+
+func VerifyTtpOnLedger(ledger xdr.LedgerCloseMeta, passphrase string) bool {
```
tamirms, line 240: I suggest renaming this function to `VerifyEvents()` . TTP is implied by the package name. OnLedger is implied by the function arguments.

https://google.github.io/styleguide/go/best-practices#avoid-repetition

ingest/processors/token_transfer/verify_events.go
```diff
@@ -0,0 +1,289 @@
...
+	isSuccess := mapsEqual(eventsMap, changesMap)
+	if !isSuccess {
+		fmt.Println("----ChangeMap-----")
+		printMap(changesMap)
+		fmt.Println("------")
+		fmt.Println("----EventsMap-----")
+		printMap(eventsMap)
+		fmt.Println("------")
+	}
+	return isSuccess
```
tamirms, line 260: consider using this [library](https://pkg.go.dev/github.com/google/go-cmp/cmp#example-Diff-Testing) to get a diff which can then be printed by the caller of this function

ingest/processors/token_transfer/token_transfer_processor.go
karthikiyer56: Updated this after checking with Sid.
The amount can never be zero or negative.
Have updated the error message to reflect that

ingest/processors/token_transfer/token_transfer_processor.go
karthikiyer56: done

ingest/processors/token_transfer/token_transfer_processor.go
karthikiyer56: nope, this wrapper is very much required.
I use that to test for mint/burn conditions based on the passed input - addr vs lpId vs CbId

ingest/processors/token_transfer/token_transfer_processor.go
karthikiyer56: Actually, the `"github.com/pkg/errors"` itself has `errors.New` and `errors.WrapF` . I am just using that and removing `"github.com/stellar/go/support/errors"`

ingest/processors/token_transfer/verify_events.go
karthikiyer56: done

ingest/processors/token_transfer/token_transfer_processor.go
karthikiyer56: done

ingest/processors/token_transfer/token_transfer_processor.go
karthikiyer56: done

ingest/processors/token_transfer/token_transfer_processor.go
```diff
@@ -0,0 +1,839 @@
...
+func NewTokenTransferProcessor(networkPassphrase string) *TokenTransferProcessor {
+	return &TokenTransferProcessor{
+		networkPassphrase: networkPassphrase,
+	}
+}
+
+// ProcessTokenTransferEventsFromLedger processes token transfer events for all transactions in a given ledger.
+// This function operates at the ledger level, iterating over all transactions in the ledger.
+// it calls ProcessTokenTransferEventsFromTransaction to process token transfer events from each transaction within the ledger.
+func (ttp *TokenTransferProcessor) ProcessTokenTransferEventsFromLedger(lcm xdr.LedgerCloseMeta) ([]*TokenTransferEvent, error) {
```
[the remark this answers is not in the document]
karthikiyer56: done

ingest/processors/token_transfer/token_transfer_processor.go
tamirms: github.com/pkg/errors is not part of the standard library. github.com/pkg/errors Is currently in maintenance mode. the errors package in the standard library is preferred over github.com/pkg/errors

ingest/processors/token_transfer/token_transfer_processor.go
```diff
@@ -0,0 +1,927 @@
...
+func (p *EventsProcessor) maybeGenerateMintOrBurnEventsForReconciliation(tx ingest.LedgerTransaction, opIndex uint32, changesMap, eventsMap map[balanceKey]int64, account xdr.MuxedAccount, asset xdr.Asset) []*TokenTransferEvent {
+	var events []*TokenTransferEvent
+
+	accountStr := account.ToAccountId().Address()
+	// Create the balance key for this account and XLM asset
+	key := balanceKey{holder: accountStr, asset: asset.StringCanonical()}
+
+	// Get the balance changes from both maps
+	changesBalance, existsInChanges := changesMap[key]
+	eventsBalance, existsInEvents := eventsMap[key]
```
tamirms, line 822:
```suggestion
	changesBalance := changesMap[key]
	eventsBalance := eventsMap[key]
```

see comment below where I explain why the exists check is not necessary

ingest/processors/token_transfer/token_transfer_processor.go
```diff
@@ -0,0 +1,927 @@
...
+		burnEvent := NewBurnEvent(meta, accountStr, amount.String64Raw(xdr.Int64(-diff)), protoAsset)
+		events = append(events, burnEvent)
+	}
+
+	return events
+}
+
+func (p *EventsProcessor) generateXlmReconciliationEvents(tx ingest.LedgerTransaction, opIndex uint32, op xdr.Operation, operationEvents []*TokenTransferEvent) ([]*TokenTransferEvent, error) {
+	// DO not run this reconciliation check for ledgers > 8 or if operation is Inflation
+	if tx.Ledger.ProtocolVersion() > 8 {
```
tamirms, line 914: In https://github.com/stellar/stellar-protocol/blob/master/core/cap-0067.md#retroactively-emitting-events it states:
>Another relevant detail for replaying events is that prior to protocol 8, there was a bug that could result in the minting/burning of XLM. 

```suggestion
	// DO not run this reconciliation check for ledgers prior to protocol 8 or if operation is Inflation
	if tx.Ledger.ProtocolVersion() >= 8 {
```

tamirms: the if statement should also check that op.Body.Type is not xdr.OperationTypeInflation to be consistent with your comment

ingest/processors/token_transfer/helpers.go
```diff
@@ -0,0 +1,114 @@
...
+
+// TODO convert to strkey for LpId
+func lpIdToStrkey(lpId xdr.PoolId) string {
+	return xdr.Hash(lpId).HexString()
+}
+
+// TODO convert to strkey for CbId
+func cbIdToStrkey(cbId xdr.ClaimableBalanceId) string {
+	return cbId.MustV0().HexString()
+}
```
tamirms, line 79: it seems like these functions are duplicates of each other so we should probably be deleting `lpIdToStrkey()` and `cbIdToStrkey()` and replacing their uses with `protoAddressFromLpHash()` and  `protoAddressFromClaimableBalanceId()`

ingest/processors/token_transfer/verify_events.go
karthikiyer56: Thanks for the reminder here.
I had meant to change it to switch on types instead.
I will make the change

ingest/processors/token_transfer/token_transfer_processor.go
```diff
@@ -0,0 +1,927 @@
...
+	} else if existsInChanges {
+		// Only in changes map, not in events map
+		diff = changesBalance
+	} else if existsInEvents {
+		// Only in events map, not in changes map
+		diff = -eventsBalance
+	} else {
+		// Not in either map, no difference
+		return events
+	}
```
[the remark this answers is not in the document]
karthikiyer56: Yeah, I had a few debug breakpoints there, to see why my code wasnt working.
i wil clean it up.

ingest/processors/token_transfer/helpers.go
karthikiyer56: 🤦 yup, will make the change

ingest/processors/token_transfer/token_transfer_processor.go
karthikiyer56: nice catch. 
will make the change

karthikiyer56: removed the check altogether from this function.
The check is not at callsite, and at the callsite (`EventsFromOperation())`, reconciliation will be skipped for `tx.Ledger.ProtocolVersion() >= 8`

ingest/processors/token_transfer/token_transfer_processor.go
karthikiyer56: reworked this to remove checks

ingest/processors/token_transfer/token_transfer_processor.go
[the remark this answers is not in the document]
karthikiyer56: done

ingest/processors/token_transfer/token_transfer_processor.go
karthikiyer56: reworked all errors in this package to use the golang errors package or  `fmt.Errorf`

ingest/processors/token_transfer/helpers.go
karthikiyer56: done.
removed the extra functions for CbID and LpId

ingest/processors/token_transfer/verify_events.go
karthikiyer56: done.
I reworked this to check on the type of event instead

ingest/processors/token_transfer/token_transfer_processor.go
```diff
@@ -0,0 +1,927 @@
...
+		return nil, nil //TODO implement this
+	default:
+		return nil, nil
+	}
+
+	if err != nil {
+		return nil, err
+	}
+
+	// Run reconciliation for all operations except Inflation, InvokeHostFunction, and default cases
```
karthikiyer56: @tamirms  - previously, in our conversations, I had mentioned that this XLM recon for the bug shud not run for inflationOperation. That is not true. it shud run for every classic operation, since it is entirely possible that, even in the inflationOperation, the sourceAccount might have minted XLM by exercising the bug. Checked this with Siddharth already.

ingest/processors/token_transfer/verify_events.go
karthikiyer56: done.
removed the mapsEqual function and  reworked verifyEvents to return an error that shows the diff, if any

protos/ingest/processors/token_transfer/token_transfer_event.proto
```diff
@@ -5,41 +5,48 @@ package token_transfer;
...
 message Clawback {
-  address.Address from = 1;
+  string from = 1;
   string amount = 3;
 }
 
 message Fee {
-  address.Address from = 1;
+  string from = 1;
   string amount = 2;
```
tamirms, line 50: I just had a realization about the `asset.Asset asset` field.

I think it would be better to move that field inside all the event messages just like the `amount` field. Although the `amount` is common to all the events, we do not place at as a top level field in `TokenTransferEvent`.

The motivation for doing such a change is:

let's say you have some code which is examining only clawbacks and it is encapsulated into a function. Ideally, the function would look like:

```go
func handleClawback(clawback *token_transfer.Clawback) {
 // ....
}
```

however, if you need to know the asset, the function would be need to be:

 ```go
func handleClawback(clawback *token_transfer.Clawback, asset *assetProto.Asset) {
 // ....
}
```

Passing in both is ok but, IMO, it's preferable to have a single type which has all the info which is relevant for the clawback and you could argue that the asset is a relevant detail.

There is also the option of:

 ```go
func handleClawback(event *token_transfer.Event) {
 // ....
}
```
but I think this is not great either because it is not as type safe. You will need to unpack the clawback `event.GetClawback()`. it's possible that someone passes into the function a different event and it would be nice to eliminate that entire class of error.

I realize this could be an annoying change to make at this point as it would require fixing a lot of tests. I wish I had spotted earlier when we were reviewing the data model. Also, I would be ok with handling this in a separate PR if you prefer.

ingest/processors/token_transfer/token_transfer_processor.go
```diff
@@ -0,0 +1,936 @@
...
+
+	from, to := addressWrapper{account: &opSrcAcc}, addressWrapper{claimableBalanceId: &claimableBalanceId}
+	event, err := p.mintOrBurnOrTransferEvent(tx, &opIndex, createCbOp.Asset, from, to, amount.String64Raw(createCbOp.Amount))
+	if err != nil {
+		return nil, err
+	}
+	return []*TokenTransferEvent{event}, nil
+}
+
+func possibleClaimableBalanceIdsFromRevocation(lpEntry liquidityPoolEntryDelta, tx ingest.LedgerTransaction, txSrcAccount xdr.AccountId, opIndex uint32) ([]xdr.ClaimableBalanceId, error) {
```
tamirms, line 371: nit: this can be moved to helpers.go

protos/ingest/processors/token_transfer/token_transfer_event.proto
karthikiyer56: You are killing me 😄 
I'll make the change

ingest/processors/token_transfer/token_transfer_event.go
```diff
@@ -77,3 +71,31 @@ func NewFeeEvent(ledgerSequence uint32, closedAt time.Time, txHash string, from
...
+		TransactionIndex: tx.Index - 1,
+		OperationIndex:   operationIndex,
+		ContractAddress:  contractAddress,
+	}
+}
+
+func (event *TokenTransferEvent) GetEventType() string {
+	switch event.GetEvent().(type) {
+	case *TokenTransferEvent_Transfer:
+		return "Transfer"
```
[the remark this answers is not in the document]
karthikiyer56: done

ingest/processors/token_transfer/token_transfer_processor.go
karthikiyer56: done

ingest/processors/token_transfer/token_transfer_processor.go
```diff
@@ -0,0 +1,936 @@
...
+
+type liquidityPoolEntryDelta struct {
+	liquidityPoolId       xdr.PoolId
+	assetA                xdr.Asset
+	assetB                xdr.Asset
+	amountChangeForAssetA xdr.Int64
+	amountChangeForAssetB xdr.Int64
+}
+
+func getImpactedLiquidityPoolEntriesFromOperation(tx ingest.LedgerTransaction, opIndex uint32) ([]liquidityPoolEntryDelta, error) {
```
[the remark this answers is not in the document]
karthikiyer56: done

ingest/tutorial/extract-ledgers/extract_ledgers.go
```diff
@@ -0,0 +1,90 @@
...
+	backend, err := ledgerbackend.NewCaptive(conf)
+	panicIf(err)
+	defer backend.Close()
+
+	// Loop through the provided ledger numbers
+	for _, ledgerSeqStr := range ledgerNumbers {
+		// Convert ledger number to int
+		var ledgerSeq int
+		_, err := fmt.Sscanf(ledgerSeqStr, "%d", &ledgerSeq)
+		panicIf(err)
```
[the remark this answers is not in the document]
tamirms: 🤦 my bad, I misread this code. I thought it was writing the ledgers to the output file. you can ignore this comment

ingest/processors/token_transfer/verify_events.go
karthikiyer56: removed, since I am using diff now

tamirms (approved):
🎉 🎉 🎉

PR #5616, opened 2025-02-27
services/horizon/internal/ingest/processors: Fix ingestContractAssetBalance() restore case

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This PR is a follow up to https://github.com/stellar/go/pull/5611

In https://github.com/stellar/go/pull/5611 , I assumed that the structure of the restoration meta differed depending on whether the restoration was for an evicted ledger entry or a ledger entry which was archived but not yet evicted. But in the latest revision of [CAP-66](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0066.md#meta) it explicitly states that:

> Whenever an entry is restored, the data/code entry will be emitted as a LedgerEntryChangeType of type LEDGER_ENTRY_RESTORE. This will include the complete LedgerEntry of the restored entry. It does not matter if the entry has been evicted or not, the meta produced will be the same. Additionally, the corresponding TTL entry will also be emitted as a LedgerEntryChangeType of type LEDGER_ENTRY_RESTORE.

So, there is no case where the LEDGER_ENTRY_RESTORE meta resembles an update.

This PR removes code in ingestContractAssetBalance() which held the assumption that the LEDGER_ENTRY_RESTORE meta for TTL entries resembles an update.

### Known limitations

[N/A]

### DIFF
--- a/services/horizon/internal/ingest/processors/asset_stats_processor.go
+++ b/services/horizon/internal/ingest/processors/asset_stats_processor.go
@@ -196,7 +196,7 @@ func (p *AssetStatsProcessor) updateDB(
 		// When ingesting from the history archives we can take advantage of the fact
 		// that there are only created ledger entries. We don't need to execute any
 		// updates or removals on the asset stats tables. And we can also skip
-		// ingesting restored contract balances and expired contract balances.
+		// deleting expired contract balances.
 		assetStatsDeltas := p.assetStatSet.All()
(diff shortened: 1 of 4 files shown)


Shaptic (approved):
LGTM* 

Your inline comments + PR explanation + CAP xref all make sense to me and the code matches the intent, but I only have a cursory understanding of the asset processor.

Issue #5578, opened 2025-01-22
Update the ingest/change Change struct to always pass LCM if available


chowbao:
The ingest/change `Change` struct has a field for `Ledger *xdr.LedgerCloseMeta` found [here](https://github.com/stellar/go/blob/master/ingest/change.go#L65-L70)

It is not always populated depending on the Change.Reason (because LCM can be populated inside the change.Transaction). I believe it is more beneficial to always populate Change.Ledger no matter the change reason as it will always be able to be associated with a LedgerCloseMeta.

Supplementally we should evaluate whether or not it makes sense to always propagate OperationIndex and Transaction as well if they exist. This will depend on if they are only available for their correlated Change.Reason or if their current scope is not all inclusive of valid OperationIndex/Transaction data

Edit: OperationIndex and Transaction are always populated when relevant as it is already

Issue #5613, opened 2025-02-25
Add an eviction / restoration integration test


tamirms:
We have integration tests which exercise expiration and restoration of contract asset balances. However, these integration tests do not exercise eviction of contract data ledger entries because eviction is only implemented in protocol 23.

Once we have a protocol 23 enabled stellar rpc docker image we should add a horizon integration test which evicts a contract asset balance and asserts the /assets response works as expected. The test should also attempt to restore the evicted contract asset balance and assert the /assets response now includes the restored contract asset balance.

urvisavla:
I tried using p-23 stellar-rpc docker image @Shaptic shared internally but I wasn’t able to run the Horizon SAC integration tests for protocol 23.

After checking with @2opremio, it turns out the rpc  simulation library needs to be updated to support p-23 which depends on an updated soroban-rust-env and that isn’t available yet.

I’m moving this ticket to blocked for now, since we need p-23 support in RPC to create contracts with eviction metadata (which is only available starting in p-23).

PR #5619, opened 2025-03-04
ingest: Update ingestion library and horizon to adapt to protocol 23 representation of evictions

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

[CAP-62](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0062.md#meta) and [CAP-66](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0066.md#xdr-changes) updates the LedgerCloseMeta xdr to no longer populate  the  `evictedPersistentLedgerEntries` field. Instead, both persistent and temporary ledger keys will be included in the `evictedTemporaryLedgerKeys` field which will be renamed to `evictedKeys` in protocol 23.

The ingestion library previously presented evictions of persistent ledger entries as a [`Change`](https://github.com/stellar/go/blob/master/ingest/change.go#L41) instance with the `Reason` set to [`LedgerEntryChangeReasonEviction`](https://github.com/stellar/go/blob/master/ingest/change.go#L96-L97). This allowed downstream consumers of a [`Change`](https://github.com/stellar/go/blob/master/ingest/change.go#L41) stream to handle evictions as if they were equivalent to ledger entry deletions. However, in protocol 23 we can no longer maintain this strategy because the full contents of the evicted ledger entry will not be present in the LedgerCloseMeta xdr. We will only have the ledger key for the evicted ledger entry.

This means that ledger entry evictions will need to be handled outside of [`Change`](https://github.com/stellar/go/blob/master/ingest/change.go#L41) stream processing by downstream services. The second commit of this PR fixes the asset stats processor in Horizon to adapt to this new reality where evictions no longer appear in the [`Change`](https://github.com/stellar/go/blob/master/ingest/change.go#L41) stream.

The asset stats processor can ignore evicted contract balance ledger entries because contract balance ledger entries are deleted from the horizon db immediately upon expiration. However, the asset stats processor still needs to ingest the stellar asset contract instance storage so that we can link stellar classic assets with their corresponding contract id. So when the stellar asset contract instance storage is evicted we also need to handle that event so we can remove the link between stellar classic asset and the stellar asset contract id.

Also fixes https://github.com/stellar/go/issues/5578

### Known limitations

[N/A]

### DIFF
--- a/ingest/change.go
+++ b/ingest/change.go
@@ -66,7 +66,9 @@ type Change struct {
 	Transaction *LedgerTransaction
 
 	// The LedgerCloseMeta that precipitated the change.
-	Ledger xdr.LedgerCloseMeta
+	// This field is not populated when the Change is obtained from enumerating
+	// ledger entries from a history archive snapshot (e.g. via CheckpointChangeReader).
+	Ledger *xdr.LedgerCloseMeta
 
 	// Information about the upgrade, if the change occurred as part of an upgrade
 	// This field is relevant only when the Reason is LedgerEntryChangeReasonUpgrade
--- a/ingest/change.go
+++ b/ingest/change.go
@@ -66,11 +66,7 @@ type Change struct {
 	Transaction *LedgerTransaction
 
 	// The LedgerCloseMeta that precipitated the change.
-	// This is useful only when the Change is caused by an upgrade or by an eviction, i.e. outside a transaction
-	// This field is populated only when the Reason is one of:
-	// LedgerEntryChangeReasonUpgrade or LedgerEntryChangeReasonEviction
-	// For changes caused by transaction or operations, look at the Transaction field
-	Ledger *xdr.LedgerCloseMeta
+	Ledger xdr.LedgerCloseMeta
 
 	// Information about the upgrade, if the change occurred as part of an upgrade
 	// This field is relevant only when the Reason is LedgerEntryChangeReasonUpgrade
@@ -95,9 +91,6 @@ const (
 
 	// LedgerEntryChangeReasonUpgrade indicates a change caused by a ledger upgrade.
 	LedgerEntryChangeReasonUpgrade
-
-	// LedgerEntryChangeReasonEviction indicates a change caused by entry eviction.
-	LedgerEntryChangeReasonEviction
 )
 
 // String returns a best effort string representation of the change.
--- a/services/horizon/internal/ingest/processors/asset_stats_processor.go
+++ b/services/horizon/internal/ingest/processors/asset_stats_processor.go
@@ -4,11 +4,13 @@ import (
 	"context"
 	"database/sql"
 	"encoding/hex"
+	pkgerrors "errors"
 	"fmt"
 	"math/big"
 
 	"github.com/stellar/go/ingest"
 	"github.com/stellar/go/services/horizon/internal/db2/history"
+	"github.com/stellar/go/support/collections/set"
 	"github.com/stellar/go/support/errors"
 	"github.com/stellar/go/xdr"
 )
@@ -23,6 +25,9 @@ type AssetStatsProcessor struct {
 	updatedExpirationEntries map[xdr.Hash][2]uint32
 	ingestFromHistoryArchive bool
 	networkPassphrase        string
+	// evictedAssetContractCandidates is a list of evicted contracts
+	// that are possibly SACs
+	possibleEvictedAssetContracts set.Set[xdr.Hash]
 }
 
 // NewAssetStatsProcessor constructs a new AssetStatsProcessor instance.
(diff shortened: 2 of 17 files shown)


xdr/ledger_close_meta.go
```diff
@@ -157,19 +157,6 @@ func (l LedgerCloseMeta) EvictedTemporaryLedgerKeys() ([]LedgerKey, error) {
...
-func (l LedgerCloseMeta) EvictedPersistentLedgerEntries() ([]LedgerEntry, error) {
-	switch l.V {
-	case 0:
-		return nil, nil
-	case 1:
-		return l.MustV1().EvictedPersistentLedgerEntries, nil
-	default:
-		panic(fmt.Sprintf("Unsupported LedgerCloseMeta.V: %d", l.V))
-	}
-}
```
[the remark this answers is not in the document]
tamirms: pre-23 ledgers don't have any evictions for persistent ledger entries. so I think this field will always be empty for both pre 23 and post 23 ledgers

tamirms:
> Either way, though, integration tests will reveal issues once RPC gets tagged off, assuming we have some for SAC+eviction+asset stats?

@Shaptic yeah, we have a issue to capture writing the protocol 23 eviction / restoration integration tests: https://github.com/stellar/go/issues/5613

ingest/change.go
```diff
@@ -66,11 +66,7 @@ type Change struct {
 	Transaction *LedgerTransaction
 
 	// The LedgerCloseMeta that precipitated the change.
-	// This is useful only when the Change is caused by an upgrade or by an eviction, i.e. outside a transaction
-	// This field is populated only when the Reason is one of:
-	// LedgerEntryChangeReasonUpgrade or LedgerEntryChangeReasonEviction
-	// For changes caused by transaction or operations, look at the Transaction field
-	Ledger *xdr.LedgerCloseMeta
+	Ledger xdr.LedgerCloseMeta
```
chowbao, line 69: What's the reason for change this from pointer?

tamirms: I intended to fix https://github.com/stellar/go/issues/5578 and  make `Ledger` always be present

services/horizon/internal/ingest/processors/asset_stats_processor.go
```diff
@@ -31,21 +36,41 @@ func NewAssetStatsProcessor(
...
+		contractDataChanges:           []ingest.Change{},
+		removedExpirationEntries:      map[xdr.Hash]uint32{},
+		createdExpirationEntries:      map[xdr.Hash]uint32{},
+		updatedExpirationEntries:      map[xdr.Hash][2]uint32{},
+		possibleEvictedAssetContracts: findAssetContractCandidates(evictedLedgerKeys),
 	}
 	return p
 }
 
+func findAssetContractCandidates(evictedLedgerKeys []xdr.LedgerKey) set.Set[xdr.Hash] {
```
karthikiyer56, line 56: nit : `findAssetContractEvictionCandidates` might be more explicit

services/horizon/internal/ingest/processors/asset_stats_processor.go
```diff
@@ -512,98 +518,120 @@ func (p *AssetStatsProcessor) updateAssetStats(
...
-			// update the row to set the contract_id column to NULL
-			stat.ContractID = nil
-			rowsAffected, err = p.assetStatsQ.UpdateAssetStat(ctx, stat)
-			if err != nil {
-				return errors.Wrap(err, "could not update asset stat")
-			}
+	} else if err != nil {
+		return errors.Wrap(err, "error querying asset by asset code and issuer")
+	} else if dbContractID, ok := stat.GetContractID(); ok {
+		// the asset stat already has a column_id set which is unexpected (the column should be NULL)
```
karthikiyer56, line 568: is this condition likely to ever happen?

services/horizon/internal/ingest/processors/asset_stats_processor.go
```diff
@@ -512,98 +518,120 @@ func (p *AssetStatsProcessor) updateAssetStats(
...
+// removeContractID will update the asset stat row for the corresponding asset to
+// remove the given contract id
+func (p *AssetStatsProcessor) removeContractID(
+	ctx context.Context,
+	contractID xdr.Hash,
+) error {
+	var rowsAffected int64
+	stat, err := p.assetStatsQ.GetAssetStatByContract(ctx, contractID)
+	if pkgerrors.Is(err, sql.ErrNoRows) {
+		// we cannot know for sure if the contract id belongs to a legitimate SAC
```
karthikiyer56, line 604: if the contractId is even passed to this `removeContractId`, then it **must** be a SAC contract, right?
Meaning,  this condition `if pkgErrors.Is(err, sqlErrNoRows` shud never be triggered, right?
So, doesnt it make sense to return a meaningful error instead of nil?

services/horizon/internal/ingest/processors/asset_stats_processor.go
tamirms: no, it should never happen which is why this code path returns an error

services/horizon/internal/ingest/processors/asset_stats_processor.go
tamirms: unfortunately we cannot know if the contract id is a SAC unless we have the full ledger entry containing the asset metadata which is stored in persistent contract data. But, in protocol 23 we will not have the full ledger entry for evictions. We will only have the ledger keys which were evicted

urvisavla (approved):
Overall, the changes look good to me conceptually. This PR does a great job of refactoring/simplifying some of the existing logic around error handling. I spent a couple of days reviewing it but there are still many nuances to the AssetStatsProcessor that are difficult to fully grasp.

PR #5635, opened 2025-03-20
refactor: use slices.Equal to simplify code

linghuying:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

use slices.Equal to simplify code

### Why

In Go 1.21, a new method called `slices.Equal` was introduced. 

Since the current version of go being used in the project is 1.23, utilizing this method can significantly reduce the amount of code and improve readability.

### Known limitations

N/A


services/horizon/internal/actions/orderbook.go
```diff
@@ -24,17 +25,7 @@ type OrderBookResponse struct {
 }
 
 func priceLevelsEqual(a, b []protocol.PriceLevel) bool {
```
tamirms, line 27: the function is only used in the `Equals(other StreamableObjectResponse) bool` function below. I think it would be better to get rid of `priceLevelsEqual()` entirely and replace it with `slices.Equal()` at the call sites

linghuying: Good idea. Modified.

Please review again.

tamirms (approved):

PR #5636, opened 2025-03-20
.github: speedup go/horizon workflow tests by disabling postgres fsync

toqueteos:
## What

The CI PostgreSQL shouldn't write to disk, this is accomplished by disabling fsync.

## Why

The official Docker Hub [postgres](https://hub.docker.com/_/postgres) image doesn't expose an easy to use lever for this, [bitnami/postgresql](https://hub.docker.com/r/bitnami/postgresql) does.

This should speedup tests significantly, specially on GHA which has less than ideal I/O speeds.

## Known limitations

N/A

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -61,11 +61,12 @@ jobs:
     runs-on: ${{ matrix.os }}
     services:
       postgres:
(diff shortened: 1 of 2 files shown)


tamirms:
@toqueteos the CI checks were not successful

PR #5637, opened 2025-03-21
TTP as a Service PoC

sreuland:
### What

PoC of using the Token Transfer Processor [TTP](https://github.com/stellar/go/blob/master/ingest/processors/token_transfer/token_transfer_processor.go) as a compiled cli tool(micro-service) which emits token transfer events over streaming GRPC protocol. 

The sub-directory structure comprises two sample applications which demonstrate the TTP as a service using .protos for IDL and a Node javascript based consumer:

* `cli_tool` runs a GRPC service by using protobufs/GRPC generated code from the TTP .protos as the IDL for token transfer events. 
  * provides `GetTTPEvents` as a GRPC streaming endpoint. 
  * uses RPC serrver on testnet as the source of ledger metadata, 
  * runs an instance of TTP internally to derive token transfer events from ledger metadata
* A sample consumer app is provided in Node javascript. It uses client-side code generated from same TTP .protos. The app requests and consumes a stream of token transfer events over GRPC from a running instance of cli_tool.

ttp_poc

[demo recording](https://github.com/user-attachments/assets/fb951085-e7b2-4d1d-9c28-6fe130032b27) of building and running the processor as a GRPC service and then running a Node javascript consumer in command line terminal:

To try this at home, checkout the pr locally and follow the demo steps in [exp/tools/ttp_poc/README.md](https://github.com/sreuland/go/tree/ttp_poc/exp/tools/ttp_poc)

### Why
* Enables applications developed in non-go programming languages to consume Token Transfer events produced by TTP. 
* Composing ingestion processors as services with an accompanying IDL(.proto file) defining their service endpoint and output model facilitates distributed streaming topology. Processors interact with other Processors through the IDL only for loosely coupled systems.


sreuland:
as processors become event driven services, conceptually their usage could extend to dynamic ingestion pipelines as a topology of processors chained to each other only through IDL defined inbound/outbound event streams:

chained_proc

Issue #5626, opened 2025-03-17
txnbuild: add helpers for constructing transactions to send payments to contracts


tamirms:
Currently it's very easy to use txnbuild to send a payment to another account. However, it's not as straightforward to send payments to recipients which are smart contracts. 

We should update txnbuild to support constructing SAC payments. The auth and resource / fee values should be hardcoded so that we can construct the the invoke host operation without requiring RPC's simulateTransaction endpoint.

See https://github.com/lightsail-network/java-stellar-sdk/pull/677 and https://github.com/StellarCN/py-stellar-base/pull/1023 for examples of how this functionality was implemented for the java and python SDKs

tamirms:
cc @marcelosalloum

Shaptic:
Opened https://github.com/stellar/js-stellar-base/issues/792 to track JS work for this.

PR #5642, opened 2025-03-25
txnbuild: Add txnbuild helper for sending a payment to a smart contract

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Close https://github.com/stellar/go/issues/5626

Update txnbuild to support constructing SAC payments to contracts with default fees so that simulateTransaction calls to rpc are not necessary.

Also fixes a bug in txnbuild where the max fee in the transaction envelope did not include the additional soroban resource fee.
 
### Why

see https://github.com/stellar/go/issues/5626

### Known limitations

[N/A]


txnbuild/invoke_host_function.go
```diff
@@ -12,6 +15,208 @@ type InvokeHostFunction struct {
...
+	}
+	var destinationContractID xdr.Hash
+	copy(destinationContractID[:], decoded)
+
+	parsedAmount, err := amount.Parse(params.Amount)
+	if err != nil {
+		return InvokeHostFunction{}, err
+	}
+
+	transferArgs := xdr.ScVec{
```
Shaptic, line 81: Seems like the Go SDK could certainly use a [`nativeToScVal`](https://stellar.github.io/js-stellar-sdk/global.html#nativeToScVal) equivalent, but that'd be a separate PR+refactor of course.

txnbuild/invoke_host_function.go
```diff
@@ -12,6 +15,208 @@ type InvokeHostFunction struct {
...
+	ReadBytes    uint32
+	WriteBytes   uint32
+	ResourceFee  int64
+}
+
+var defaultPaymentToContractFees = SorobanFees{
+	Instructions: 400_000,
+	ReadBytes:    1_000,
+	WriteBytes:   1_000,
+	ResourceFee:  5_000_000,
```
Shaptic, line 29: Were these acquired from simulation + rounding up, or just "definitely large enough"?

txnbuild/restore_footprint.go
```diff
@@ -10,6 +12,77 @@ type RestoreFootprint struct {
...
+	ResourceFee:  4_000_000,
+}
+
+// AssetBalanceRestorationParams configures the restore footprint operation returned by
+// NewAssetBalanceRestoration
+type AssetBalanceRestorationParams struct {
+	// NetworkPassphrase is the passphrase for the Stellar network
+	NetworkPassphrase string
+	// Contract is the contract which holds the asset balance
+	Contract string
```
Shaptic, line 28: This could be an account, too, right? Or is that not a use-case we care about?

txnbuild/transaction.go
```diff
@@ -915,8 +914,16 @@ func NewTransaction(params TransactionParams) (*Transaction, error) {
 		if err != nil {
 			return nil, errors.Wrap(err, fmt.Sprintf("failed to build operation %T", sorobanOp))
 		}
+		if envelope.V1.Tx.Ext.SorobanData != nil {
```
Shaptic, line 917: I think either this should be done higher up or the uint32 check should be done after this, otherwise we miss the "proper" validation on L853 in this case.

txnbuild/invoke_host_function.go
tamirms: I got them from them java and python sdk implementations written by @overcat :

https://github.com/lightsail-network/java-stellar-sdk/pull/677

https://github.com/StellarCN/py-stellar-base/pull/1023

txnbuild/restore_footprint.go
tamirms: I think account to account payments are better served by the classic payment operation

txnbuild/invoke_host_function.go
```diff
@@ -12,6 +15,208 @@ type InvokeHostFunction struct {
...
+		}
+		footprint.ReadOnly = append(footprint.ReadOnly,
+			xdr.LedgerKey{
+				Type: xdr.LedgerEntryTypeAccount,
+				Account: &xdr.LedgerKeyAccount{
+					AccountId: issuer,
+				},
+			},
+		)
+		if !sourceAccount.Equals(issuer) {
```
tamirms, line 155: @overcat this check is missing from the java and python sdks. We should only include the trustline in the read write footprint if the sender is not the issuer. Otherwise, the transaction will fail because it is impossible for an issuer to have a trustline to its own asset.

@Shaptic this is something to keep in mind when implementing this feature in the javascript sdk

txnbuild/restore_footprint.go
Shaptic: Ah yeah - I meant here that you might want to restore an account's balance held by a contract but that wouldn't ever be the case since that's in the trustline, right?

txnbuild/invoke_host_function.go
Shaptic: Nice catch, I'll certainly keep that in mind :+1: '

txnbuild/invoke_host_function.go
overcat: It's definitely large enough. https://github.com/lightsail-network/java-stellar-sdk/pull/677/files#diff-338b40b80a77385ce8f94bb8c79269a588b87a112a370760184837e2a5345357R290-R293

txnbuild/invoke_host_function.go
overcat: Thank you, I will add the corresponding tests.

txnbuild/restore_footprint.go
overcat: Since we already have PaymentOp, I think supporting only contract address here will make things simpler, and clearly the costs required for PaymentOp are lower.

tamirms: > I meant here that you might want to restore an account's balance held by a contract but that wouldn't ever be the case since that's in the trustline, right?

sorry I thought you were commenting on the payment to contract helper but I see now that you're commenting about the restore footprint helper. you're right the account balance is held in a trustline and currently only contract data / code ledger entries can be archived. so there is no need currently to ever restore a trustline or any other ledger entry which isn't of a contract data / code ledger entry type.

txnbuild/transaction.go
tamirms: @Shaptic I just pushed a fix for this issue. can you take a look?

services/horizon/internal/test/integration/integration.go
```diff
@@ -214,6 +214,10 @@ func NewTest(t *testing.T, config Config) *Test {
 		}
 
 		i.WaitForHorizonIngest()
+
+		if !config.SkipCoreContainerCreation && !i.config.QuickExpiration {
```
sreuland, line 218: minor, could consolidate all condition checks to single point inside `upgradeLimits()` after existing check on protocolversion.

txnbuild/invoke_host_function.go
```diff
@@ -12,6 +15,208 @@ type InvokeHostFunction struct {
...
+type PaymentToContractParams struct {
+	// NetworkPassphrase is the passphrase for the Stellar network
+	NetworkPassphrase string
+	// Destination is the contract recipient of the payment
+	Destination string
+	// Amount is the amount being transferred
+	Amount string
+	// Asset is the asset being transferred
+	Asset Asset
+	// SourceAccount is the source account of the payment
```
sreuland, line 42:
```suggestion
	// SourceAccount is the source account of the payment, it must be a Stellar account in strkey encoded`VersionByteAccountID` format, i.e. a 'G' account.
```

txnbuild/invoke_host_function.go
```diff
@@ -12,6 +15,208 @@ type InvokeHostFunction struct {
...
+}
+
+// PaymentToContractParams configures the payment returned by NewPaymentToContract
+type PaymentToContractParams struct {
+	// NetworkPassphrase is the passphrase for the Stellar network
+	NetworkPassphrase string
+	// Destination is the contract recipient of the payment
+	Destination string
+	// Amount is the amount being transferred
+	Amount string
```
sreuland, line 39: could `Amount` be strongly typed to a float64 instead, more intuitive for callers?

txnbuild/invoke_host_function.go
```diff
@@ -12,6 +15,208 @@ type InvokeHostFunction struct {
...
+			Type: xdr.ScValTypeScvI128,
+			I128: &xdr.Int128Parts{
+				Hi: 0,
+				Lo: xdr.Uint64(parsedAmount),
+			},
+		},
+	}
+
+	resources := params.Fees
+	if resources.ResourceFee == 0 {
```
sreuland, line 106: could there ever be legit use case for setting fee to 0? If remotely plausible, it may be worthwhile to define `PaymentToContractParams.Fees` as a pointer `*SorobanFees` so can check for explicit absence before applying defaults.

txnbuild/invoke_host_function.go
```diff
@@ -12,6 +15,208 @@ type InvokeHostFunction struct {
...
+				},
+				FunctionName: "transfer",
+				Args:         transferArgs,
+			},
+		},
+		SourceAccount: params.SourceAccount,
+		Auth: []xdr.SorobanAuthorizationEntry{
+			{
+				Credentials: xdr.SorobanCredentials{
+					Type: xdr.SorobanCredentialsTypeSorobanCredentialsSourceAccount,
```
sreuland, line 188: Based on this hardcode, should source account authentication be stated as limitation in docs for`NewPaymentToContract` usage? 

Can auth aspect be configurable via something like `PaymentToContractParams.Auth SorobanAuth` with `SorobanAuth` as a new struct for defining auth type and sub-fields such as when type is `SOROBAN_CREDENTIALS_ADDRESS`? Maybe this latter address type auth requires client to assemble tx and invoke preflight manually, meaning they should not use `NewPaymentToContract` in that case?

txnbuild/transaction.go
```diff
@@ -997,6 +998,21 @@ func NewFeeBumpTransaction(params FeeBumpTransactionParams) (*FeeBumpTransaction
...
+
+	hi, lo := bits.Mul64(uint64(params.BaseFee), uint64(len(inner.operations)+1))
+	if hi > 0 || lo > math.MaxInt64 {
+		return nil, errors.Errorf("base fee %d results in an overflow of max fee", params.BaseFee)
+	}
+	totalFee := lo + sorobanFee
+	if totalFee < lo || totalFee > math.MaxInt64 {
+		return nil, fmt.Errorf("soroban fee %v results in an overflow of max fee", sorobanFee)
+	}
+	tx.maxFee = int64(totalFee)
```
sreuland, line 1014: minor, lines 1006-1014 looks like duplication of same code on lines 916-925, might be worth refactor to internal, reusable fn?

txnbuild/invoke_host_function.go
tamirms: the payment operation in the txnbuild package uses strings to represent amounts so I wanted to be consistent with that

txnbuild/transaction.go
tamirms: although this code appears to be identical to the one in lines 916-925, there are some subtle differences. For Fee bump transactions the fee value is limited to positive 64 bit integer values. However, for normal transactions, the fee has a lower ceiling as it is represented by positive 32 bit integer values.

tamirms:
@sreuland can you take another look at the PR?

txnbuild/invoke_host_function.go
sreuland: Yes, your reply answers my second question. 
Minor follow-up, In 'account to contract payment use case', this helper encodes source account authentication for the invocation. Do you think this nuance of auth behavior is worth mention on [public docs for the helper function](https://github.com/stellar/go/pull/5642/files#diff-934e4b7f5462fb8d3c53dc23d51db4114f25c8c184d4a2f86bf957f0b27de3a1R50-R51) ?

sreuland (approved):
lgtm, nice work!

Issue #5625, opened 2025-03-17
Implement verification of TTP events against SAC events


tamirms:
Verifying TTP events for contract balances is difficult because we cannot identify contract balance ledger entries which were generated by an SAC. In order to link a balance to an SAC we would need to know that the contract id of the balance corresponds to an SAC. However, while it is possible to go from asset to SAC id, the reverse (determining asset details from a contract id) is not possible. The asset code and issuer is not present in the [balance contract data](https://github.com/stellar/go/blob/385ac5aca1a4ace8b8c15ee067812de1f2fb9f6c/ingest/processors/contract_processor/contract_data.go#L309-L382) so we will not be able to determine if the contract id corresponds to an SAC.

To work around this problem we can exclude contract balances from the reconciliation logic outlined in https://gist.github.com/tamirms/aa80bf3d9f7c9ee6ff4fc6e13e054bf4 . Instead, we can filter all TTP events which were emitted by an invoke host operation and compare them against the soroban events found in the `SorobanTransactionMeta` for that operation. There should be a match between the TTP events and the soroban events because TTP events were designed to be isomorphic with SAC events.

PR #5648, opened 2025-03-27
Add support for contract events

karthikiyer56:
This PR unifies the work for parsing events custom tokens that are SEP-41 compliant, and SAC events.
SAC events are a superset of SEP-41. Meaning, Stellar Asset Contracts inherently conform to the interface as defined in SEP-41.
so, the logic to parse is as follows:
- Try to see if event conforms to SEP-41
- If not, error out
- If event conforms to SEP-41, then try to additionally parse it as a SAC event to get classic assetName.

This PR effectively closes out the following tickets:
#5593 and its 2 sub-tasks
and
the verification of smart contract events - #5625 

After much deliberation with @tamirms, we have concluded that the unit tests sufficiently cover the generation of TTP events from `xdr.ContractEvent`. There is nothing that can be verified against ledger entries, like how we do for accounts/trustlines/liquiditypools/cbs etc.
The only thing that is changed in the existing `VerifyEvents()` logic is that if the key is a C-address, we dont include it in the balanceChange delta reconciliation equation.
Previously, we were thinking of roundtripping like so - `list of xdr.ContractEvent --> parseEvent code which will generate list of TokenTransferEvents ---> regenerate xdr.ContractEvent`.
Except that is not possible, since the `parseEvent` will omit events and decidedly give you a smaller subset than the input `[]xdr.ContractEvent`


ingest/processors/token_transfer/contract_events.go
```diff
@@ -0,0 +1,156 @@
...
+				asset := assets[0]
+				// Verify contract ID matches expected asset contract ID
+				expectedId, idErr := asset.ContractID(p.networkPassphrase)
+				if idErr == nil && expectedId == *contractEvent.ContractId {
+					// If contract ID matches, update with validated asset
+					protoEvent.SetAsset(asset)
+				}
+			}
+		}
+	}
```
tamirms, line 67: do we need to return an error if `len(topics)` is not 3 or 4?

what if the last topic is not an asset, should we be returning an error?

finally, if in the code below, err is not nil or len(assets) != 1 should we be returning an error ?

```go
assets, err := xdr.BuildAssets(string(assetStr))
```

karthikiyer56: if you have passed the logic for `parseCustomTokenEvent`, it is a given that you are a sep-41 compliant event.
And so there will be a resulting proto event generated. After that point in the code, there wont be an error generated.

the remaining logic is to further check if the event is a SAC event, and if it is we extract the asset from it.

It could very well be the case that the event is like so
`["transfer", "someFromAddress", "someToAddress", "SomeExtra field here"]`
The "SomeExtra field here" could very well be a in a proper SEP-11 asset representation format or it could not.
In this case this is still a valid SEP-41 compliant event format, even if it is not a SAC event, and so we dont throw an error, but instead generate a transfer proto event

ingest/processors/token_transfer/contract_events_test.go
```diff
@@ -0,0 +1,967 @@
...
+	"fmt"
+	"github.com/stellar/go/keypair"
+	"github.com/stellar/go/strkey"
+	"github.com/stellar/go/xdr"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+var (
```
karthikiyer56, line 14: This is the most important file in this PR
I have covered all kinds of unit testing - both positive and negative.

ingest/processors/token_transfer/contract_events_test.go
```diff
@@ -0,0 +1,967 @@
...
+// Helper function to create a ScVal symbol
+func createSymbol(sym string) xdr.ScVal {
+	symStr := xdr.ScSymbol(sym)
+	return xdr.ScVal{
+		Type: xdr.ScValTypeScvSymbol,
+		Sym:  &symStr,
+	}
+}
+
+func contractIdToHash(contractId string) *xdr.Hash {
```
karthikiyer56, line 40: @Shaptic  - jfyi, quite a few of these helper functions are copied over from `support/contractevents` and tweaked a  little bit.

This duplication, imo, is ok and going forward perhaps, the code in support/contractEvents can be deprecated in lieu of this code, but thats not something we need to decide any time soon

ingest/processors/token_transfer/contract_events.go
```diff
@@ -0,0 +1,172 @@
...
+
+	// First, try parsing as a standard SEP41 token contractEvent
+	protoEvent, sepErr := parseCustomTokenEvent(string(fn), tx, opIndex, contractEvent)
+	if sepErr != nil {
+		return nil, sepErr
+	}
+
+	// This has passed validation for SEP-41 complaint token.
+	// At the very least, you will now emit an contractEvent.
```
chowbao, line 57:
```suggestion
	// At the very least, you will now emit a contractEvent.
```

ingest/processors/token_transfer/contract_events.go
```diff
@@ -0,0 +1,172 @@
...
+	// Attempt SAC validation if possible, to get asset name
+
+	// SAC validation requires a very strict check on len(topics)
+	// For transfer, mint and clawback - there will be exactly 4 elements
+	// For burn, there will be exactly 3 events
+	// transfer - "transfer", toAddr, fromAddr, sep11AssetString
+	// mint - "mint", admin, toAddr, sep11AssetString
+	// clawback - "clawback", admin, fromAddr, sep11AssetString
+	// burn - "burn", fromAddr, sep11AssetString
```
chowbao, line 66: nit: does it make sense to put the SAC validation in a separate function with the idea that SEP-41 compliant custom tokens won't follow the same validation logic?

For example, `admin` will be removed from SEP-41 custom tokens iirc

ingest/processors/token_transfer/contract_events.go
```diff
@@ -0,0 +1,172 @@
...
+			return nil, errNotSep41TokenFromError(fmt.Errorf("invalid fromAddress. error: %w", err))
+		}
+		to, err := extractAddress(topics[2])
+		if err != nil {
+			return nil, errNotSep41TokenFromError(fmt.Errorf("invalid toAddress. error: %w", err))
+		}
+		event = NewTransferEvent(meta, from, to, amtRaw128, nil)
+
+	case MintEvent:
+		// Mint requires MINIMUM 3 topics - event type, admin, toAddr
```
chowbao, line 125: I'm a little confused by the `// parseCustomTokenEvent attempts to parse a generic SEP41 token event` function
By including a check for `admin` this actually isn't a generic SEP-41 check anymore. Is this really a SAC event check?
```
The mint event is emitted when an amount of the token is minted.

The event has topics:

Symbol with value "mint"
Address the address to hold the newly minted tokens.
The event has data:

i128 the amount minted.
```

ingest/processors/token_transfer/contract_events.go
karthikiyer56: thats actually what I want to avoid doing
I dont want a separate function for parsing SAC events and a separate function for custom tokens becaue SAC is just a superset of SEP-41.
Having a separate function makes the logic hard to reason and test, as I found out when I was writing this code.

Re: the removal of `admin` in p23, this code will be reworked when P23 goes in,but not before

ingest/processors/token_transfer/contract_events.go
karthikiyer56: > Edit: NVM this point is moot because before P23 sep-41 did not even have mint/clawback events. After P23 admin will be removed from all these events anyways so this code will be updated

Yup, discussed this in person with Simon. This is a moot-point, since the [current version of SEP-41](https://github.com/stellar/stellar-protocol/blob/88342d515a6d73159617047aea936595a6061548/ecosystem/sep-0041.md) doesnt specify mint and clawback. It only specifies transfer and burn, and this code in the PR, inadvertenly, works corrrectly

ingest/processors/token_transfer/verify_events.go
```diff
@@ -15,8 +16,17 @@ type balanceKey struct {
 	asset  string
 }
 
+func isContractAddress(key balanceKey) bool {
+	_, err := strkey.Decode(strkey.VersionByteContract, key.holder)
```
Shaptic, line 20: Better would be to add a `IsValidContractId` to the `strkey` package to match the [other validation functions](https://github.com/stellar/go/blob/master/strkey/main.go#L285).

ingest/processors/token_transfer/token_transfer_processor.go
```diff
@@ -195,6 +195,24 @@ func (p *EventsProcessor) EventsFromOperation(tx ingest.LedgerTransaction, opInd
 	return events, nil
 }
 
+func (p *EventsProcessor) contractEvents(tx ingest.LedgerTransaction, opIndex uint32) ([]*TokenTransferEvent, error) {
+	diagnosticEvents, err := tx.GetDiagnosticEvents()
```
Shaptic, line 199: After our conversation I had a realization: rather than getting all of the diagnostic events, which are a superset of contract events, then filtering on having a contract ID + in a successful call, you can just fetch the contract events directly, i.e. some new helper `tx.GetContractEvents()` that doesn't do the diag. vs. contract coalescing we discussed in our last call. This will lead to faster processing (no filtering needed) and a smaller memory footprint (the slice capacity on L203 will be smaller).

ingest/processors/token_transfer/contract_events_test.go
Shaptic: This is a hex-encoded contract ID as opposed to a strkey-encoded contract ID, right? It'd be nice to avoid going to <-> from the binary representation entirely if we could and stick to strkeys.

ingest/processors/token_transfer/contract_events_test.go
```diff
@@ -0,0 +1,964 @@
...
+func createAddress(address string) xdr.ScVal {
+	scAddress := xdr.ScAddress{}
+
+	switch address[0] {
+	case 'C':
+		scAddress.Type = xdr.ScAddressTypeScAddressTypeContract
+		contractHash := strkey.MustDecode(strkey.VersionByteContract, address)
+		scAddress.ContractId = contractIdToHash(hex.EncodeToString(contractHash))
+
+	case 'G':
```
Shaptic, line 63: I know it's just a test but it'd be strictly more correct to use `case IsValidEd25519PublicKey(address)` here.

ingest/processors/token_transfer/contract_events_test.go
```diff
@@ -0,0 +1,964 @@
...
+	}
+	if copy(idBytes[:], rawBytes[:]) != 32 {
+		panic("couldn't copy 32 bytes to contract hash")
+	}
+
+	hash := xdr.Hash(idBytes)
+	return &hash
+}
+
+func createAddress(address string) xdr.ScVal {
```
Shaptic, line 54: This feels like this is worth sharing as part of `xdr/scval.go` due to its usefulness.

ingest/processors/token_transfer/contract_events_test.go
```diff
@@ -0,0 +1,964 @@
...
+	}
+}
+
+func TestValidContractEvents(t *testing.T) {
+	testCases := []struct {
+		name          string
+		eventType     string
+		addr1         string // meaning depends on event type (from/admin)
+		addr2         string // meaning depends on event type (to/from/empty)
+		amount        int64
```
Shaptic, line 141: It'd be good to throw an `i128` amount in here to validate the high-bits code path.

ingest/processors/token_transfer/contract_events_test.go
```diff
@@ -0,0 +1,964 @@
...
+	randomAccount     = keypair.MustRandom().Address()
+	someContractHash1 = xdr.Hash([32]byte{1, 2, 3, 4})
+	someContract1     = strkey.MustEncode(strkey.VersionByteContract, someContractHash1[:])
+	someContractHash2 = xdr.Hash([32]byte{4, 3, 2, 1})
+	someContract2     = strkey.MustEncode(strkey.VersionByteContract, someContractHash2[:])
+
+	processor = &EventsProcessor{
+		networkPassphrase: someNetworkPassphrase,
+	}
+	contractIdFromAsset = func(asset xdr.Asset, passphrase string) *xdr.Hash {
```
Shaptic, line 24: We can get away with hardcoding `passphrase` to `someNetworkPassphrase`, right?

ingest/processors/token_transfer/contract_events_test.go
```diff
@@ -0,0 +1,964 @@
...
+		name          string
+		eventType     string
+		addr1         string // meaning depends on event type (from/admin)
+		addr2         string // meaning depends on event type (to/from/empty)
+		amount        int64
+		isSacEvent    bool
+		validateEvent func(t *testing.T, event *TokenTransferEvent, addr1, addr2 string, amount string, assetItem interface{})
+	}{
+		{
+			name:       "Transfer SEP-41 Token Event",
```
Shaptic, line 146: Maybe I glossed over it but we should test [valid SEP-41] + [not a SAC event] + greater than expected topics

Shaptic (commented):
Really clean implementation and comprehensive test coverage :clap: handful of comments below but mostly LGTM.

sydneynotthecity (commented):
I don't have a solution for this but want to flag: it is possible for a subset of event types to be non-compliant with SEP-41, which will display an incomplete picture for a token. 

ie, it is possible for a `transfer` or `burn` event to be well formatted and included in the response of TTP, but `mint` events are thrown out because they're incorrectly formatted.  

In this case, that means it would not be possible to track circulating supply, but transactional volume _could_ be tracked. I think we just need to document this risk because it could be unintuitive that only a subset of custom token events are emitted.

karthikiyer56:
> I don't have a solution for this but want to flag: it is possible for a subset of event types to be non-compliant with SEP-41, which will display an incomplete picture for a token.
> 
> ie, it is possible for a `transfer` or `burn` event to be well formatted and included in the response of TTP, but `mint` events are thrown out because they're incorrectly formatted.
> 
> In this case, that means it would not be possible to track circulating supply, but transactional volume _could_ be tracked. I think we just need to document this risk because it could be unintuitive that only a subset of custom token events are emitted.

Ack.
I think the answer to that might depend on - what it means for a custom token contract to be sep 41 compliant. I tried to get some clarity here https://stellarfoundation.slack.com/archives/C030Z9EHVQE/p1743113563298039?thread_ts=1743113563.298039&cid=C030Z9EHVQE
The thread ended with us aligning on - ttp will parse and output an event if it passes the parsing logic, whether or not the token is sep41 complaint is simply a moot point from ttp's perspective. 

That being said,i want to point out that the original sep41 only mentions that transfer and burn events will be emitted by the custom token contract. 
The mint and clawback events and their signature were added relatively recently to sep41 as a part of unified events. 
This was something that @chowbao brought to my attention today. 
I don't know what that means for existing custom tokens already emitting mint events.

ingest/processors/token_transfer/contract_events_test.go
karthikiyer56: Yup.
those cases are being tested as part of `TestSacAssetValidation` mega test

ingest/processors/token_transfer/contract_events_test.go
karthikiyer56: done.
removed this function altogether

ingest/processors/token_transfer/verify_events.go
karthikiyer56: done.
Added a helper in `strkey`

ingest/processors/token_transfer/contract_events_test.go
karthikiyer56: done.

ingest/processors/token_transfer/contract_events_test.go
karthikiyer56: um, not so sure about that.
Everything in `scval.go` is receiver functions on Scval, more or less.
This is strictly used in the context of the test. I am inclined to leave it here

ingest/processors/token_transfer/contract_events_test.go
karthikiyer56: done

ingest/processors/token_transfer/token_transfer_processor.go
karthikiyer56: erm, I would have to do the same processing as in GetDiagnosticEvents(), right?
by same processing I mean - we have established that either diagnostic events or contract events will be present in sorobanMeta, but not both. So even in the case where diagnostic events exist. I'd have to have the `if` condition from`filterEvents`.
Basically, I am not sure if we get faster processing.

karthikiyer56: This has been reworked to call the `tx.GetContractEvents()` function as recommended

ingest/processors/token_transfer/contract_events_test.go
karthikiyer56: done

ingest/processors/token_transfer/contract_events_test.go
Shaptic: Ah nice, found it:

```
			name: "Valid SAC asset with correct format but not exactly 4 topics",
```

ingest/processors/token_transfer/verify_events.go
```diff
@@ -198,41 +203,42 @@ func findBalanceDeltasFromEvents(events []*TokenTransferEvent) map[balanceKey]in
...
-	eventsMap = findBalanceDeltasFromEvents(events)
+		txHash := tx.Hash.HexString()
+		txEvents, err = ttp.EventsFromTransaction(tx)
+		if err != nil {
+			return fmt.Errorf("verifyEventsError: %w", err)
+		}
+		feeChanges := tx.GetFeeChanges()
+		txChanges, err := tx.GetChanges()
+		if err != nil {
+			return fmt.Errorf("verifyEventsError: %w", err)
```
Shaptic, line 232: I _think_ [`errors.Join`](https://pkg.go.dev/errors#Join) is more canonical because then `errors.Is` works on the whole chain, but since these are just strings it doesn't really matter. Just wanted to point out that function's existence. It'd look something like `errors.Join(fmt.Errorf("verifyEvents failed"), err)`.

Shaptic (approved):
GREAT work :clap:

ingest/processors/token_transfer/verify_events.go
karthikiyer56: ack.

PR #5657, opened 2025-04-08
Karthik/add memo support for ttp

karthikiyer56:
Add muxed support in classic operation and events
Only `to_muxed_id`, if it exists will appear in transfer and mint events.


ingest/processors/token_transfer/token_transfer_processor.go
```diff
@@ -266,31 +311,43 @@ func (p *EventsProcessor) mintOrBurnOrTransferEvent(tx ingest.LedgerTransaction,
...
 
 		// Check for Mint Event
-		if to == "" {
+		if toAddress == "" {
 			return nil, NewEventError("mint event error: to address is nil")
 		}
-		return NewMintEvent(meta, to, amt, protoAsset), nil
+		mintEvent := NewMintEvent(meta, toAddress, amt, protoAsset)
+		// Add muxed information - this will only have `to_muxed_info`, if at all
+		err := mintEvent.addMuxedInfoForMintEvent(to, tx)
```
karthikiyer56, line 324: this is important and deliberate.
`toAddress` is the G-address that is passed to `NewMintEvent`
whereas `to` is pased to `.addMuxedInfo...` since `to` can be G-or-M address

Same with transferEvent a few lines down in the code

ingest/processors/token_transfer/token_transfer_processor_test.go
```diff
@@ -339,6 +370,64 @@ var (
 			},
 		}
 	}
+
+	strictSendOp = func(sourceAccount *xdr.MuxedAccount, destAccount xdr.MuxedAccount, sendAsset xdr.Asset, destAsset xdr.Asset) xdr.Operation {
```
karthikiyer56, line 374: This part is just code-reordering.
I am pulling these helpers out from the individual tests to the top-level, since I use them in the tests for mux-info

ingest/processors/token_transfer/token_transfer_processor_test.go
```diff
@@ -738,14 +902,6 @@ func TestPaymentEvents(t *testing.T) {
 				transferEvent(protoAddressFromAccount(accountA), protoAddressFromAccount(accountB), unitsToStr(100*oneUnit), usdcProtoAsset),
 			},
 		},
-		{
```
karthikiyer56, line 741: These tests dont make sense here anymore, and are pulled out into their own test `TestMuxedInformation`

ingest/processors/token_transfer/muxed_info.go
```diff
@@ -0,0 +1,41 @@
+package token_transfer
+
+import "github.com/stellar/go/xdr"
+
+func NewMemoFromXdrMemo(m *xdr.Memo) *MuxedInfo {
```
sreuland, line 5: minor nit, Is muxed and memo completely interchangeable? Would this be more technically accurate to be `NewMuxedFromXdrMemo`?

ingest/processors/token_transfer/muxed_info.go
```diff
@@ -0,0 +1,41 @@
...
+			Hash: hashSlice,
+		}
+	case xdr.MemoTypeMemoReturn:
+		hashSlice := make([]byte, 32)
+		copy(hashSlice, (*m.RetHash)[:])
+		protoMemo.Content = &MuxedInfo_Hash{
+			Hash: hashSlice,
+		}
+	}
+	return protoMemo
```
sreuland, line 32: just to confirm, is this the intended response when no type match and fall through?

ingest/processors/token_transfer/muxed_info.go
```diff
@@ -0,0 +1,41 @@
...
+
+	switch m.Type {
+	case xdr.MemoTypeMemoNone:
+		return nil
+	case xdr.MemoTypeMemoText:
+		protoMemo.Content = &MuxedInfo_Text{
+			Text: *m.Text,
+		}
+	case xdr.MemoTypeMemoId:
+		protoMemo.Content = &MuxedInfo_Id{
```
sreuland, line 16: can this use the helper function instead - `return NewMemoFromId(*m.Id)`

ingest/processors/token_transfer/token_transfer_processor.go
sreuland: Subtle, should this aspect be retained in code as a comment or naming convention  `to->toStrKey` and `toAddress->toGAddress`?

karthikiyer56: This naming convention- `toAddress renamed toGAddress` wont work, since this function is passed all kinds of strkeys - LiquidityPoolIds, claimableBalanceIds etc etc.

But, I can replace from and to to be `fromStrkey` and `toStrkey`. thats reasonable

ingest/processors/token_transfer/muxed_info.go
karthikiyer56: >  Is muxed and memo completely interchangeable? 
For the most part, yes.
`MuxedInfo` represents what can appear as `to_muxed_id` in the event data payload, post P23.
Memo.Id is mapped to MuxedInfo.Id
Memo.Text is mapped to MuxedInfo.text
Memo.Return and Memo.Hash are both mapped to MuxedInfo.Hash. as per CAP-67

So, yes, this shud really be `NewMuxedInfoFromXdrMemo`. I will make the change

karthikiyer56: done

ingest/processors/token_transfer/muxed_info.go
karthikiyer56: nope.
thats a miss.
Added a default panic

ingest/processors/token_transfer/muxed_info.go
karthikiyer56: done

ingest/processors/token_transfer/token_transfer_processor.go
karthikiyer56: done.

ingest/processors/token_transfer/token_transfer_processor_test.go
```diff
@@ -178,6 +195,20 @@ var (
 		}
 	}
 
+	memoFromMuxedAccount = func(acc xdr.MuxedAccount) *MuxedInfo {
```
sreuland, line 198: this could be more aptly named `muxedinfoFromMuxedAccount`?

ingest/processors/token_transfer/muxed_info.go
sreuland: is it worth panic or an error which caller can decide to handle differently?

protos/ingest/processors/token_transfer/token_transfer_event.proto
```diff
@@ -20,6 +20,14 @@ message EventMeta {
...
    */
   string contract_address = 6;
+
+  /*
+   The to_muxed_id will be either the 64-bit memo-id if the destination is a muxed account
+   OR
+   it will be the memo from the transaction
+   Refer this section is CAP-67 - https://github.com/stellar/stellar-protocol/blob/master/core/cap-0067.md#multiplexing-support
+   */
+  MuxedInfo to_muxed_id = 7;
```
sreuland, line 30:
```suggestion

  /*
   The to_muxed_info follows the spec outlined in section in CAP-67 - https://github.com/stellar/stellar-protocol/blob/master/core/cap-0067.md#multiplexing-support
 */  
  MuxedInfo to_muxed_info= 7;
```

protos/ingest/processors/token_transfer/token_transfer_event.proto
```diff
@@ -65,3 +73,11 @@ message TokenTransferEvent {
     Fee fee = 6;
   }
 }
+
+message MuxedInfo {
+  oneof content {
+    string text = 1;    // with max length of 28 characters
```
sreuland, line 79: since protos doesn't allow defining length constraints on strings this comment should probably just be removed as it has no effect and may trigger some clients to do un-necessary validation.

ingest/processors/token_transfer/token_transfer_processor_test.go
```diff
@@ -719,6 +808,81 @@ func TestMergeAccountEvents(t *testing.T) {
...
+			name: "Path Payment - G account to G Account with Tx Memo - BTC Transfer",
+			tx:   someTxWithMemo(someIdMemo),
+			op:   strictReceiveOp(&accountA, accountB, btcAsset, btcAsset, 100*oneUnit),
+			opResult: strictReceiveResult(
+				[]xdr.ClaimAtom{}, // empty path
+			),
+			expected: []*TokenTransferEvent{
+				transferEventWithDestMux(protoAddressFromAccount(accountA), protoAddressFromAccount(muxedAccountB), unitsToStr(100*oneUnit), btcProtoAsset, NewMuxedInfoFromMemo(&someIdMemo)),
+			},
+		},
```
sreuland, line 881: is it worthwhile to add one more test for `G account to M account with a Tx Memo` to assert outcome when both  muxed 'to' andr tx memo are present?

ingest/processors/token_transfer/token_transfer_processor.go
```diff
@@ -227,30 +227,75 @@ func (p *EventsProcessor) contractEvents(tx ingest.LedgerTransaction, opIndex ui
...
+	err := e.setDestinationMuxedInfo(to, tx)
+	if err != nil {
+		return fmt.Errorf("error setting toMuxedInfo for transfer event: %w", err)
+	}
+
+	return nil
+}
+
+func (e *TokenTransferEvent) setDestinationMuxedInfo(to string, tx ingest.LedgerTransaction) error {
+	if strkey.IsValidMuxedAccountEd25519PublicKey(to) {
```
sreuland, line 244: What if the 'to' account is muxed and the tx envelope also has a memo? Can this occur from a classic operation?
I see CAP-67 states memos are not allowed on soroban tx's and was curious if precendence towards the muxed 'to' account was the expected outcome if both present from classic?

tamirms:
@karthikiyer56 In https://github.com/stellar/stellar-protocol/blob/master/core/cap-0067.md#emit-a-map-as-the-data-field-in-the-transfer-event-if-muxed-information-is-being-emitted it states that both `from_muxed_id` and `to_muxed_id` can be present in the data payload of the transfer event

ingest/processors/token_transfer/token_transfer_processor_test.go
karthikiyer56: absolutely right.
I will add a test for when dest=mux and txMemo is present

ingest/processors/token_transfer/token_transfer_processor.go
karthikiyer56: The precedence, when both dest=Mux and txMemo=present, is defined in the CAP.
I will write a test case for it

ingest/processors/token_transfer/muxed_info.go
karthikiyer56: I prefer it this way, tbh.
I have seen it done both ways in a lot of the receiver functions everywhere.
IMO, a panic conveys the sentiment better that -  you dont expect a wierd `xxx.Type` to appear

protos/ingest/processors/token_transfer/token_transfer_event.proto
karthikiyer56: done

protos/ingest/processors/token_transfer/token_transfer_event.proto
karthikiyer56: done

ingest/processors/token_transfer/token_transfer_processor_test.go
karthikiyer56: done

ingest/processors/token_transfer/token_transfer_processor.go
karthikiyer56: Added UTs for this case

ingest/processors/token_transfer/token_transfer_processor_test.go
karthikiyer56: done

ingest/processors/token_transfer/token_transfer_processor.go
```diff
@@ -227,30 +227,75 @@ func (p *EventsProcessor) contractEvents(tx ingest.LedgerTransaction, opIndex ui
 	return events, nil
 }
 
+func (e *TokenTransferEvent) addMuxedInfoForTransferEvent(to string, tx ingest.LedgerTransaction) error {
```
tamirms, line 230: I think receiver functions on `TokenTransferEvent ` should be moved to ingest/processors/token_transfer/token_transfer_event.go

ingest/processors/token_transfer/token_transfer_processor.go
```diff
@@ -227,30 +227,75 @@ func (p *EventsProcessor) contractEvents(tx ingest.LedgerTransaction, opIndex ui
...
+			return fmt.Errorf("could not get muxed account id: %w", err)
+		}
+		e.Meta.ToMuxedInfo = NewMuxedInfoFromId(muxedId)
+		return nil
+	}
+
+	txMemo := tx.Envelope.Memo()
+	if txMemo.Type == xdr.MemoTypeMemoNone {
+		return nil
+	}
```
tamirms, line 257: I think this case is already handled in `NewMuxedInfoFromMemo()`

ingest/processors/token_transfer/token_transfer_processor.go
karthikiyer56: done

ingest/processors/token_transfer/token_transfer_processor.go
karthikiyer56: Agreed.
Changed the function `NewMuxedInfoFromMemo` to take in a `xdr.Memo` as well

sreuland (approved):
lgtm, nice work!

ingest/processors/token_transfer/token_transfer_processor.go
```diff
@@ -690,8 +704,9 @@ func (p *EventsProcessor) liquidityPoolWithdrawEvents(tx ingest.LedgerTransactio
 }
 
 func (p *EventsProcessor) generateEventsFromClaimAtoms(tx ingest.LedgerTransaction, opIndex uint32, opSrcAcc xdr.MuxedAccount, claims []xdr.ClaimAtom) ([]*TokenTransferEvent, error) {
+	// Converting MuxedAccount to strictly G-Addresss, since events from trading pair shouldnt have destination muxed_info set
+	operationSource := opSrcAcc.ToAccountId().Address()
 	var events []*TokenTransferEvent
-	operationSource := opSrcAcc.Address()
 	var seller string
 
 	for _, claim := range claims {
```
tamirms, line 712: a few lines below you have the following code:

```go
			sellerAccount := sellerId.ToMuxedAccount()
			seller = sellerAccount.Address()
```
I think the conversion to muxed account is unnecessary, right?

ingest/processors/token_transfer/token_transfer_event.go
```diff
@@ -156,3 +157,50 @@ func (event *TokenTransferEvent) SetAsset(asset xdr.Asset) {
...
+		muxedId, err := muxedAcc.GetId()
+		if err != nil {
+			return fmt.Errorf("could not get muxed account id: %w", err)
+		}
+		e.Meta.ToMuxedInfo = NewMuxedInfoFromId(muxedId)
+		return nil
+	}
+
+	txMemo := tx.Envelope.Memo()
+	e.Meta.ToMuxedInfo = NewMuxedInfoFromMemo(txMemo)
```
tamirms, line 192: in the case of a transaction with a memo, will all the transfer events emitted by that transaction (for example, events emitted during a path payment operation) include the muxed info? Is that the correct behavior?

tamirms:
@karthikiyer56 do we need to add any code to ingest/processors/token_transfer/contract_events.go to parse out muxed info from events?

karthikiyer56:
> @karthikiyer56 do we need to add any code to ingest/processors/token_transfer/contract_events.go to parse out muxed info from events?


Not at the moment, no.
There is no MUX anything support for smart contracts, so we dont need to do anything.
Once P23 goes out, all of this will have to change, and i will have to write new parsing logic, but nothing for now

ingest/processors/token_transfer/token_transfer_processor.go
karthikiyer56: yes, thats right
i wil fix it

ingest/processors/token_transfer/token_transfer_event.go
karthikiyer56: Good catch.

for path payments, if TxMemo is set, then only last event, if at all, will have dest mux set.
for manage offers, even if TxMemo is set, no events will have dest mux set.

I will need to make an ugly change to the `p.mintOrBurnOrTransferEvent`, sigh.

karthikiyer56: done.
@tamirms - as a part of doing this, the code has some additional complexity, and so I have added more UTs in the `TestMuxedInformation`

tamirms (approved):

PR #5691, opened 2025-05-08
Fee Event breakdown in token transfer processor

karthikiyer56:
Add support for fee refund events.
The comments in the `NewEventsFromLedger` function highlight the chronological order before P22 (i.e. the status quo)

I had to modify the signature of `NewEventsFromTransaction` to return FeeEvents and OperationEvents separately.
The coalescing is done in chronological order in `NewEventsFromLedger`

### DIFF
--- a/ingest/ledger_transaction.go
+++ b/ingest/ledger_transaction.go
@@ -224,10 +224,7 @@ func (t *LedgerTransaction) FeeCharged() (int64, bool) {
 		if uint32(t.Ledger.LedgerHeaderHistoryEntry().Header.LedgerVersion) < 21 && t.Envelope.Type == xdr.EnvelopeTypeEnvelopeTypeTxFeeBump {
 			var resourceFeeRefund int64
 
-			resourceFeeRefund, ok = t.SorobanResourceFeeRefund()
-			if !ok {
-				return 0, false
-			}
+			resourceFeeRefund = t.SorobanResourceFeeRefund()
 
 			return int64(t.Result.Result.FeeCharged) - resourceFeeRefund, true
 		}
@@ -345,6 +342,11 @@ func (t *LedgerTransaction) GetSorobanData() (xdr.SorobanTransactionData, bool) 
 	default:
 		panic(fmt.Errorf("unknown EnvelopeType %d", t.Envelope.Type))
 	}
+}
+
+func (t *LedgerTransaction) IsSorobanTx() bool {
+	_, res := t.GetSorobanData()
+	return res
 }
 
 func (t *LedgerTransaction) SorobanResourceFee() (int64, bool) {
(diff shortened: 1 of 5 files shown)


ingest/ledger_transaction.go
```diff
@@ -222,13 +222,7 @@ func (t *LedgerTransaction) FeeCharged() (int64, bool) {
...
 	if ok {
 		if uint32(t.Ledger.LedgerHeaderHistoryEntry().Header.LedgerVersion) < 21 && t.Envelope.Type == xdr.EnvelopeTypeEnvelopeTypeTxFeeBump {
-			var resourceFeeRefund int64
-
-			resourceFeeRefund, ok = t.SorobanResourceFeeRefund()
-			if !ok {
-				return 0, false
-			}
-
+			resourceFeeRefund := t.SorobanResourceFeeRefund()
```
karthikiyer56, line 225: A lot of these soroban fee related functions here can be simplified to return a value, instead of a `value, bool`.
Will tackle that separately

ingest/ledger_transaction.go
```diff
@@ -347,6 +341,11 @@ func (t *LedgerTransaction) GetSorobanData() (xdr.SorobanTransactionData, bool)
 	}
 }
 
+func (t *LedgerTransaction) IsSorobanTx() bool {
```
karthikiyer56, line 344: I could have sworn there was a function like this somewhere as a receiver method.
except, I couldnt find it

ingest/ledger_transaction.go
```diff
@@ -499,21 +498,23 @@ func getAccountBalanceFromLedgerEntryChanges(changes xdr.LedgerEntryChanges, sou
 	return accountBalanceStart, accountBalanceEnd
 }
 
-func (t *LedgerTransaction) SorobanResourceFeeRefund() (int64, bool) {
-	meta, ok := t.UnsafeMeta.GetV3()
-	if !ok {
-		return 0, false
+func (t *LedgerTransaction) OriginalFeeCharged() int64 {
+	startingBal, endingBal := getAccountBalanceFromLedgerEntryChanges(t.FeeChanges, t.FeeAccount().ToAccountId().Address())
```
karthikiyer56, line 502: Note that this function `OriginalFeeCharged` does not check for `isSorobanTx`, and rightly so.
For classic operations, you can rely on either this function OR on the `FeeCharged` field in the TxResult

ingest/processors/token_transfer/token_transfer_processor_test.go
```diff
@@ -234,6 +260,44 @@ var (
 		}
 	}
 
+	accountEntry = func(acc xdr.MuxedAccount, balance xdr.Int64) *xdr.AccountEntry {
```
karthikiyer56, line 263: moved around some helpers to generate ledgerEntryChanges to the  top level, for reuse. they were previously local to another test.

tamirms (approved):

PR #5696, opened 2025-05-15
Reorder code from under ingest/processors/ to top-level processors/

karthikiyer56:
Some additinal things that needed to happen:
- protos/ingest/processors/token_transfer  needed to be moved to protos/processors/token_transfer, so that the generated code wud come under proccessors/token_transfer. It is cleaner this way
- same for protos/ingest/asset. protos/ingest/asset was moved to protos/asset/. The resulting go genereated code is now under asset/


protos/README.md
```diff
@@ -61,7 +61,7 @@ This structure is maintained to ensure Go packages align correctly with their co
 
 - **Follow Directory Structure**
   - Organize `.proto` files into directories that match logical groupings.
-  - Keep related `.proto` files together (e.g., All data models related to token_tranfer should be under `protos/ingest/processors/token_transfer/`).
+  - Keep related `.proto` files together (e.g., All data models related to token_tranfer should be under `protos/processors/token_transfer/`).
```
Shaptic, line 64:
```suggestion
  - Keep related `.proto` files together (e.g., All data models related to token_transfer should be under `protos/processors/token_transfer/`).
```

processors/token_transfer/token_transfer_processor.go
```diff
@@ -6,8 +6,8 @@ import (
 	"io"
 
 	"github.com/stellar/go/amount"
+	assetProto "github.com/stellar/go/asset"
```
Shaptic, line 9: Should the generated code for protos also live under `processors/`? Or at least a single subdirectory `protosgen` (better name pls)? One could imagine we add tons more protos and they'd clutter the top-level folder significantly.

tamirms (approved):

processors/token_transfer/token_transfer_processor.go
karthikiyer56: in all fairness the asset proto message is independent of the processors.
The reason I put it _outside_ of ttp code was, if we build a few more processors using protos, then we can leverage the same `asset` message (or `address` message).

I knew that I wud have to contend with this when I started the `git mv` task.
I am still unsure about what to do here. That being said,  I am planning to have a 20 min working session with the team next week to discuss this further. But until then, perhaps we can let it be as is.
I want to get this in (and maybe support/contract move to horizon/internal) before xdr for p23 lands in

chowbao: > I want to get this in (and maybe support/contract move to horizon/internal) before xdr for p23 lands in

I also don't like having it at the top level but I think the quote from above is more important than figuring out where `protosgen` (lol at the name) should go

karthikiyer56: Agreed offline. 
Punting this convo for next week.

PR #5697, opened 2025-05-15
protocols: Update Stellar Core's `getledgerentry` endpoint schema

Shaptic:
### What
Update field names:
 * `state` is always present
 * `new` -> `not-found`
 * `ttl` -> `liveUntilLedgerSeq`
 * `e` -> `entry`, only present if `state == live|archived`

### Why
This matches the changes in https://github.com/stellar/stellar-core/pull/4721, cc @SirTyson.

### DIFF
--- a/protocols/stellarcore/getledgerentry_response.go
+++ b/protocols/stellarcore/getledgerentry_response.go
@@ -19,7 +19,7 @@ type GetLedgerEntryResponse struct {
 }
 
 type LedgerEntryResponse struct {
-	Entry string `json:"e"`             // base64-encoded xdr.LedgerEntry, or xdr.LedgerKey if state == new
-	State string `json:"state"`         // one of the above states
-	Ttl   uint32 `json:"ttl,omitempty"` // optionally, a Soroban entry's `liveUntilLedgerSeq`
+	State              string `json:"state"`                        // one of the above states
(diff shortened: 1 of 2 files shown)


protocols/stellarcore/getledgerentry_response.go
```diff
@@ -3,10 +3,9 @@ package stellarcore
 const (
 	// Indicates that the entry is live in the current state
 	LedgerEntryStateLive = "live"
-	// Indicates that the entry is proven to be brand new and will live in the
-	// current state when created. In this case, the `Entry` field will be an
-	// xdr.LedgerKey matching the one requested rather than an xdr.LedgerEntry.
-	LedgerEntryStateNew = "new"
+	// Indicates that the entry wasn't found (thus proven to be brand new) and
+	// will live in the current state if created.
+	LedgerEntryStateNotFound = "not-found"
```
urvisavla, line 8: Where is this used?

Shaptic: This endpoint is only used by RPC ([here](https://github.com/stellar/stellar-rpc/blob/protocol-23/cmd/stellar-rpc/internal/ledgerentries/main.go#L71-L75), today) for which I'll have an accompanying PR right after this gets merged!

urvisavla (approved):

PR #5698, opened 2025-05-16
Move stats_change processor from ingest/ to services/horizon/internal

karthikiyer56:
Additionally, the directories in top-level processors/ have been renamed and the _processor has been removed from the name, since it is redundant.
This also makes the dirname conform  with the name of the package that they house.


processors/account/account.go
```diff
@@ -7,7 +7,7 @@ import (
 	"github.com/guregu/null"
 	"github.com/guregu/null/zero"
 	"github.com/stellar/go/ingest"
-	utils "github.com/stellar/go/processors/processor_utils"
```
karthikiyer56, line 10: Changes like this are simply because the name of the directory under `processors/` has been renamed from `processor_utils` to simply `utils`

services/horizon/internal/ingest/processors/stats_change_processor.go
```diff
@@ -1,8 +1,9 @@
-package ingest
+package processors
 
 import (
```
karthikiyer56, line 3: This is the file that has been moved from `ingest/` to `horizon/internal/`
The original ticker mentions that this shud be moved to  the top-level `processors/` directory.
but on further digging, it makes sense to move it to horizon internal, since it is used only there.
Also this processor implements the `ProcessChange` function - something that is internal to horizon today

services/horizon/internal/ingest/verify_test.go
```diff
@@ -4,6 +4,7 @@ import (
 	"crypto/sha256"
 	"database/sql"
 	"fmt"
+	"github.com/stellar/go/services/horizon/internal/ingest/processors"
```
tamirms, line 7: nit: this should be grouped with the other stellar go imports

services/horizon/internal/ingest/processors/stats_change_processor_test.go
```diff
@@ -1,7 +1,8 @@
-package ingest
+package processors
 
 import (
 	"context"
+	"github.com/stellar/go/ingest"
```
tamirms, line 5: nit: this should be grouped with the other stellar go imports

services/horizon/internal/ingest/main_test.go
```diff
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"database/sql"
+	"github.com/stellar/go/services/horizon/internal/ingest/processors"
```
tamirms, line 7: nit: this should be grouped with the other stellar go imports

services/horizon/internal/ingest/fsm.go
```diff
@@ -3,13 +3,13 @@ package ingest
 import (
 	"context"
 	"fmt"
+	"github.com/stellar/go/services/horizon/internal/ingest/processors"
```
tamirms, line 6: nit: this should be grouped with the other stellar go imports

services/horizon/internal/ingest/build_state_test.go
```diff
@@ -4,9 +4,9 @@ package ingest
 
 import (
 	"context"
+	"github.com/stellar/go/services/horizon/internal/ingest/processors"
```
tamirms, line 7: nit: this should be grouped with the other stellar go imports

tamirms (approved):

services/horizon/internal/ingest/verify_test.go
karthikiyer56: done

services/horizon/internal/ingest/processors/stats_change_processor_test.go
karthikiyer56: done

services/horizon/internal/ingest/main_test.go
karthikiyer56: done

services/horizon/internal/ingest/build_state_test.go
karthikiyer56: done

services/horizon/internal/ingest/fsm.go
karthikiyer56: Done

Issue #5571, opened 2025-01-13
Enable RPC as a source of LCM in go/ingest


mollykarcher:
### What problem does your feature solve?

RPC is the recommended starting point for newcomers to Stellar, and we'd like to eventually migrate most people Horizon -> RPC. However, RPC (intentionally) does not have the same indexing capabilities or API niceties that Horizon does, which forces users to index their own data based on their own use cases. This means parsing and traversing raw XDR, as well as becoming intimately familiar with Stellar's underlying data model, which isn't the easiest for a lot of users. We want these kinds of users to be able to use the parsing/transforming componentry we've already built in go/ingest and go/processors so they don't need to reinvent the wheel or port over some parsing code.

### What would you like to see?

An RPC service can be swapped in as the source of LCM that can be parsed by the ingest library. Idea being, a golang service/project can:
- Run/deploy RPC
- Import go/ingest (+optionally go/processors) into their project
- Stream LCM from RPC via the same interaction model they would if they were pulling from any other LCM source (captive-core, Galexie-lake)

Something to keep is mind is that we are building a golang client for RPC - https://github.com/stellar/stellar-rpc/issues/326 - and it might be worthwhile to utilize that when building this. It's undecided/unclear where that client will live, but should it live in the stellar-rpc repo, this would create a circular dependency between stellar/go and stellar/stellar-rpc, so this perhaps could influence our current debate around repo organization. 

### Alternatives

Using a Galexie-exported data lake is an alternative. However, we're specifically thinking about use-cases where the operator/developer will still need to submit transactions, which is not possible via Galexie/data-lake. So while they _could_ use Galexie-lake as their txmeta source, that means they would need to still run both RPC and Galexie. Think wallet developers (ex. freighter), asset issuers (ex. circle), centralized exchanges (ex. coinbase), etc.

PR #5699, opened 2025-05-16
ingest: Rpc as a ledger backend

sreuland:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Added new `ledgerbackend.RPCLedgerBackend`

### Why

Provides RPC as another alternative source for users of the ingest sdk to acquire ledger metadata through the standard LedgerBackend interface.
Closes: #5571 

### Known limitations

The RPCLedgerBackend implementation has no-op's for range preparations of the `ledgerbackend.LedgerBackend` interface, `PrepareRange` and `IsPrepared`. Stateful range preparation cannot be deterministically done with RPC since it's ledger retention window is constantly sliding forward.


ingest/ledgerbackend/rpc_backend.go
```diff
@@ -0,0 +1,86 @@
...
+		return xdr.LedgerCloseMeta{}, fmt.Errorf("failed to unmarshal ledger close meta: %w", err)
+	}
+
+	return lcm, nil
+}
+
+// RPCLedgerBackend does not perform stateful range preparation.
+// This has no effect on the backend and is a no-op.
+func (b *RPCLedgerBackend) PrepareRange(ctx context.Context, ledgerRange Range) error {
+	return nil
```
Shaptic, line 75: We could use this to prefill/speed up the `GetLedger` response time by grabbing the first page of ledgers in `ledgerRange`?

Shaptic (approved):
LGTM as a starting point! Just one comment :arrow_down: as a suggestion for performance.

ingest/ledgerbackend/rpc_backend.go
sreuland: yes, neat, I'll weave that in.

tamirms:
this rpc backend implementation has a lot of differences in behavior compared to the buffered storage backend and captive core (e.g. rpc backend allows random access to ledgers whereas captive core / buffered storage backend only allows you to scan ledgers in ascending order). I think it would be good to make the rpc backend aligned with the buffered storage backend / captive core as well because it allows all the ledger backends to be interchangeable.

If someone were to use the rpc backend as it's implemented currently then there's a chance that their code would break when using the buffered storage backend or captive core and that would not be desirable.

tamirms:
I think there is an opportunity to share a lot of code between buffered storage backend and the rpc backend because conceptually the two are very similar. Perhaps the buffered storage backend could be refactored so that the implementation details that are specific to the data lake are abstracted into an interface. Then we can implement another instance of the interface for rpc and plug that into the buffered storage backend.

sreuland:
> I think there is an opportunity to share a lot of code between buffered storage backend and the rpc backend because conceptually the two are very similar. Perhaps the buffered storage backend could be refactored so that the implementation details that are specific to the data lake are abstracted into an interface. Then we can implement another instance of the interface for rpc and plug that into the buffered storage backend.

I wondered about BSB usage with RPC initially. There's impedance as RPC's sliding window of ledgers does not provide same guarantee of existence which BSB assumes with`Datastore`. Adapting RPC to BSB would significantly expand scope of effort into refactoring and regression testing afterward. The additional complexity may out-weigh the net result.

>  I think it would be good to make the rpc backend aligned with the buffered storage backend / captive core as well because it allows all the ledger backends to be interchangeable.

> If someone were to use the rpc backend as it's implemented currently then there's a chance that their code would break when using the buffered storage backend or captive core and that would not be desirable.

Can you briefly describe how it breaks interchangeability of using RPCLedgerBackend as a LedgerBackend? 
The `LedgerBackend.GetLedger` interface method throws Error and existing SDK clients of CaptiveCore or BSB backends would have error handling already. RPCLedgerBackend would throw error when not found. Is the concern related to these existing client usages ignore the error and re-try `GetLedger` for a sequence with the assumption of eventual arrival? Whereas with RPC, if it doesn't exist, it may never exist and client gets stuck in retries on the same not found error?

>  rpc backend allows random access to ledgers whereas captive core / buffered storage backend only allows you to scan ledgers in ascending order

That is internal behavior specific to a physical backend but the sequential ascending aspect is not documented externally as an expectation on the [LedgerBackend](https://github.com/stellar/go/blob/master/ingest/ledgerbackend/ledger_backend.go) interface in PrepareLedger or GetLedger usage. The interface docs do make a reference to Captive core as a specific impl with nuanced behavior.

I think the following options are present for RPCLedgerBackend:
*  1) RPCLedgerBackend uses BSB which encapsulates RPC usage.  
    * rewrite BSB to interact with RPCClient.   
      *  retrieve the entire requested range from RPC up front or throw error if RPC retention window doesn't have the requested range.  
      * eagerly long poll to fetch ledgers from RPC for unbounded. 
      * Store the results from RPC into mem or local files
    * Implement RPCLedgerBackend to just use BSB, no direct connection to RPC  
* 2) RPCLedgerBackend internally does same eager ranged buffering locally in its PrepareRange/GetLedger implementation.
* 3) RPCLedgerBackend in current PR format. Don't approximate sequential/ascending ledger behavior.

For first two options, range buffering on local client introduces need to address local storage, it probably should focus on disk rather than memory to be less intrusive on caller and some upper bound limit on ranges may need to be err'd during PrepareRange. Unbounded range implies a requirement for GetLedger to be called at same time in order to purge ledgers read from the buffer or it could max out local storage.

If we need to approximate the implied sequential/ascending behavior, I think option 2 may be less risk. If not, Option 3 could be viable.

tamirms:
>Can you briefly describe how it breaks interchangeability of using RPCLedgerBackend as a LedgerBackend?

the rpc backend allows random access of ledgers whereas captive core and buffered storage backend does not. BSB could allow random access, however, we deliberately tried to make the behavior consistent with captive core so that the two ledger backends can be interchangeable. 

> That is internal behavior specific to a physical backend but the sequential ascending aspect is not documented externally as an expectation on the [LedgerBackend](https://github.com/stellar/go/blob/master/ingest/ledgerbackend/ledger_backend.go) interface in PrepareLedger or GetLedger usage.

I agree that this behavior is not documented in the interface but the interface is hardly documented at all. I think the behavior of captive core is the de facto spec for how ledger backends should behave since captive core was the only implementation of the ledger backend interface up until we introduced BSB. The goal is to make the different ledger backends interchangeable so that workflows that run without errors using BSB can also execute correctly with captive core or vice versa.

However, we could definitely add another interface for retrieving ledgers that would allow random access and this could be implemented via rpc or the data lake.

Regarding the implementation options for rpc, I don't think it is absolutely necessary that rpc buffers ledgers. I think the MVP would just be that the rpc ledger backend fetches ledgers but restricts the ledger access in the same way it is restricted in captive core / BSB. In other words, we can do simple fetching of ledgers from rpc one by one but return the same errors as captive core / BSB if the ledger backend is used in an unexpected way (e.g. out of order ledger access).

sreuland:
> The goal is to make the different ledger backends interchangeable so that workflows that run without errors using BSB can also execute correctly with captive core or vice versa.
> I don't think it is absolutely necessary that rpc buffers ledgers.

Having RPCLedgerBackend interchangeable to this degree of `ingest.LedgerBackend` implies a guarantee that GetLedger will return all sequences requested from a prior PrepareRange invocation with no errors for missing ledgers. To meet this expectation I believe RPCLedgerBackend.PrepareRange would need to fetch that entire range from RPC up front then buffer it locally. Otherwise, there's race condition possible as time elapses after PrepareRange is called and the sliding retention window of RPC moves outside of a prepared range done by client earlier. 

> I think the MVP would just be that the rpc ledger backend fetches ledgers but restricts the ledger access in the same way it is restricted in captive core / BSB. 

Yes, this is doable, can retain state in RPCLedgerBackend.GetLedger so that it simulates same blocking and error behavior related to retrieval of forward-only, contiguously ascending sequence numbers.

> However, we could definitely add another interface for retrieving ledgers that would allow random access and this could be implemented via rpc or the data lake.

A new interface for this trait sounds good.
```
type RandomAccessLedgerBackend interface {
   FindLedger(ctx context.Context, sequence uint32) (xdr.LedgerCloseMeta, error)
```

Is it feasible for RPCLedgerBackend to just be a `ingest.RandomAccessLedgerBackend` rather than a `ingest.LedgerBackend`? In this way it's still provided as an abstraction in the ingest SDK for acquiring ledger metadata but avoids impedance with LedgerBackend and would update BSB to support FindLedger as well.

tamirms:
@sreuland I just realized that you can already obtain ledgers in a random access fashion using the rpc client. The rpc client exposes a `GetLedgers()` function which can be used to fetch arbitrary ledgers. So, I don't think there's any point in implementing a wrapper which more or less does the same thing.

> Otherwise, there's race condition possible as time elapses after PrepareRange is called and the sliding retention window of RPC moves outside of a prepared range done by client earlier.

I think we can live with that race condition for now. Obtaining all the ledgers during `PrepareRange()` would work but if the ledger range is large it would require a lot of memory to store everything

sreuland:
@tamirms per confirmation in chat related to [options](https://github.com/stellar/go/pull/5699#issuecomment-2905405646), I proceeded on Option 2. Applied the behaviors on the RPCLedgerBackend to mimic that of LedgerBackend expectation for sequential, increasing ledger retrieval only. Updated PrepareRange and IsPrepared to mimic expectations also. 

[410f9c9](https://github.com/stellar/go/pull/5699/commits/410f9c9ffa95b8e5c4ce673c68e83c7bed27c14f)

ingest/ledgerbackend/rpc_backend.go
```diff
@@ -0,0 +1,253 @@
...
+		return xdr.LedgerCloseMeta{}, err
+	}
+
+	if b.preparedRange == nil {
+		return xdr.LedgerCloseMeta{}, fmt.Errorf("RPCLedgerBackend must be prepared before calling GetLedger")
+	}
+
+	if b.closed {
+		return xdr.LedgerCloseMeta{}, fmt.Errorf("RPCLedgerBackend is closed; cannot GetLedger")
+	}
```
tamirms, line 115: I think this check is already included in `checkClosed()` which is invoked a few lines above

ingest/ledgerbackend/rpc_backend.go
```diff
@@ -0,0 +1,253 @@
...
+
+	rangesMatch := b.preparedRange.from == ledgerRange.from &&
+		b.preparedRange.bounded == ledgerRange.bounded &&
+		(!b.preparedRange.bounded || b.preparedRange.to == ledgerRange.to)
+
+	return rangesMatch, nil
+}
+
+func (b *RPCLedgerBackend) Close() error {
```
tamirms, line 193: Close() can be called in another thread while the main thread is calling PrepareRange() or GetLedger(). So I think we need to do some work to make this thread-safe and to make it possible for GetLedger() / PrepareRange() to be interrupted by Close()

ingest/ledgerbackend/rpc_backend.go
```diff
@@ -0,0 +1,253 @@
...
+}
+
+// Creates the RPCLedgerBackend with the given RPC URL and optional HTTP client.
+func NewRPCLedgerBackendFromURL(rpcURL string, httpClient *http.Client, bufferSize uint32) (*RPCLedgerBackend, error) {
+	return NewRPCLedgerBackend(rpc.NewClient(rpcURL, httpClient), bufferSize)
+}
+
+// GetLatestLedgerSequence queries the RPC server for the latest ledger sequence.
+func (b *RPCLedgerBackend) GetLatestLedgerSequence(ctx context.Context) (sequence uint32, err error) {
+	ledger, err := b.client.GetLatestLedger(ctx)
```
tamirms, line 76: to be consistent with BSB and CaptiveStellarCore, `GetLatestLedgerSequence()` should return the latest ledger available in the buffer. Also, `GetLatestLedgerSequence()` can be called concurrently with `GetLedger()` . `PrepareRange()`, and `Close()`

ingest/ledgerbackend/rpc_backend.go
```diff
@@ -0,0 +1,253 @@
...
+			return lcm, nil
+		}
+
+		_, isBeyondErr := err.(*RPCLedgerBeyondLatestError)
+		if !isBeyondErr {
+			return xdr.LedgerCloseMeta{}, err
+		}
+
+		select {
+		case <-ctx.Done():
```
tamirms, line 139: I think we'll need a third case which is if `Close()` interrupts the call to `GetLedger()`

ingest/ledgerbackend/rpc_backend.go
sreuland: doh, thanks!

ingest/ledgerbackend/rpc_backend.go
sreuland: I made public methods thread-safe per instance of the backend, this is same pattern as BSB employs. 

<del>Minor detail, `Close` method is not concurrently executable in regards to other methods on same instance to achieve interrupt aspect. When one thread already is executing a method on the instance, all other threads block to execute any method on same instance until first thread exits. So one thread looping on GetLedger invocations, would get interrupted by closed state but only after another thread called `Close` in between invocations of GetLedger. This has same behavior as BSB.</del>

[edit: - I misunderstood the concurrent interrupt expectation for Close. I. have resolved now so Close can be source of interrupt on the other backend methods. I put close aspect on a separate channel so it is sync'd separate from that of buffer locking, So, Close method can be invoked while other methods are running, [820bf790cecf](https://github.com/stellar/go/pull/5699/commits/820bf790cecfc84780d884eb79829a9788bfec8a)

ingest/ledgerbackend/rpc_backend.go
sreuland: oh, I misunderstood the expectation of LedgerBackend.GetLatestLedger then, the docs state it as 'latest available in the backend' but I see in BSB as well now, this should effectively return the last ledger sequence fetched by a prior GetLedger call, done - https://github.com/stellar/go/pull/5699/commits/d304ee13c462ce5d9ea8796cdc9403e82381eca3 , thx!

ingest/ledgerbackend/rpc_backend.go
```diff
@@ -0,0 +1,300 @@
...
+		return 0, fmt.Errorf("RPCLedgerBackend must be prepared before calling GetLatestLedgerSequence")
+	}
+
+	// mimic the expected behavior for backend, get latest returns 0 if prepared
+	// but GetLedger is not called yet, as this means tha backend has not advanced into any ledgers yet
+	if b.nextLedger == b.preparedRange.from {
+		return 0, nil
+	}
+
+	return b.nextLedger - 1, nil
```
tamirms, line 118: I think this should return the latest ledger available in the buffer like captive core and bsb

tamirms: also there is thread-safety issues with accessing b.nextLedger because GetLedger modifies b.nextLedger and GetLatestLedgerSequence() can be invoked concurrently with GetLedger() since they both use a read lock

ingest/ledgerbackend/rpc_backend.go
tamirms: ok, if we take that approach then I think we should ensure there are timeouts on the rpc client operations to remove the possibility of blocking for a prolonged period of time.

ingest/ledgerbackend/rpc_backend.go
```diff
@@ -0,0 +1,300 @@
...
+}
+
+// The minimum required RPC client interface required for usage by RPCLedgerBackend.
+type RPCClient interface {
+	GetLedgers(ctx context.Context, req protocol.GetLedgersRequest) (protocol.GetLedgersResponse, error)
+}
+
+// RPCLedgerBackend does not support stateful range preparations.
+// The rpc backend is composed of ephermeral sliding window of ledgers and therefore
+// connot prepare a range of ledgers which remains consistent over time.
```
tamirms, line 42: I'm not sure what this means. What is stateful range preparation? What does it mean for a range of ledgers to remain consistent over time?

ingest/ledgerbackend/rpc_backend.go
```diff
@@ -0,0 +1,300 @@
...
+//   - client: The RPC client implementation used to communicate with the server
+//   - bufferSize: Size of the ledger retrieval buffer (in number of ledgers).
+//     If 0, defaults to 10.
+//
+// Returns:
+//   - *RPCLedgerBackend: A new backend instance ready for use
+//   - error: If initialization fails
+//
+// The returned backend must be prepared with PrepareRange before GetLedger can be called.
+func NewRPCLedgerBackend(client RPCClient, bufferSize uint32) (*RPCLedgerBackend, error) {
```
tamirms, line 71: I think it's odd to explicitly pass in a buffer size of 0 as a positional argument. I think it would be better to either have a config struct where the buffer size could be unset or you could get rid of the constructor entirely and have the RPCLedgerBackend struct be configured directly similar to https://pkg.go.dev/net/http#Client

ingest/ledgerbackend/rpc_backend.go
```diff
@@ -0,0 +1,300 @@
...
+// Returns:
+//   - xdr.LedgerCloseMeta: The ledger meta data if found
+//   - error: One of:
+//   - nil                      if ledger is found
+//   - RPCLedgerMissingError    if sequence falls within the RPC's retention window,
+//     but the ledger for sequence is not in RPC's retained history
+//   - context.DeadlineExceeded if context times out
+//   - context.Canceled         if context is cancelled
+//   - Error                    if sequence falls outside of RPC's current retention window
+//     or due to any other error in general
```
tamirms, line 143: I don't think we need to document all the possible states for the error value because we don't know what type of transient error the rpc client can return. You can just mention that if the ledger cannot be found in RPC  RPCLedgerMissingError will be returned

ingest/ledgerbackend/rpc_backend.go
sreuland: I still misinterpreted this, ok corrected now, returns greatest sequence present in underlying buffer -https://github.com/stellar/go/commit/820bf790cecfc84780d884eb79829a9788bfec8a

ingest/ledgerbackend/rpc_backend.go
sreuland: No, i had this wrong and misunderstood the concurrency expectation with Close being a source of concurrent interrupt to other backend methods. I have fixed now. I put close aspect on a channel so it is sync'd separately from that of buffer access locking in other methods. So, the Close method can be invoked concurrently while other methods are running. - https://github.com/stellar/go/commit/820bf790cecfc84780d884eb79829a9788bfec8a

ingest/ledgerbackend/rpc_backend.go
```diff
@@ -0,0 +1,300 @@
...
+}
+
+// RPCLedgerBackend does not support stateful range preparations.
+// The rpc backend is composed of ephermeral sliding window of ledgers and therefore
+// connot prepare a range of ledgers which remains consistent over time.
+//
+// Callers should focus on using RPCLedgerBackend.GetLedger for the ledger range needed
+// and check the returned error for presence of a ledger.
+//
+// Instances of RPCLedgerBackend are thread-safe and can be used concurrently across goroutines.
```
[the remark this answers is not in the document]
sreuland: Yes, thanks for clarification I changed docs to reflect more narrowed usage for Close as interrupt. I have misunderstood the concurrency expectations but have cleaned them up [now](https://github.com/stellar/go/pull/5699/commits/820bf790cecfc84780d884eb79829a9788bfec8a) with regards to GetLedger and PrepareRange are on same write lock, rest of methods are on read lock and Close is on a separate channel.

ingest/ledgerbackend/rpc_backend.go
sreuland: Resolved. I changed GetLedger to be on write lock and GetLatest remains on read lock and checks for latest seq in buffer rather than last read seq - https://github.com/stellar/go/pull/5699/commits/820bf790cecfc84780d884eb79829a9788bfec8a

ingest/ledgerbackend/rpc_backend.go
sreuland: Was trying to explain how the historical state between prepare and get could change but this is too verbose in any case. I removed this content and focused on smaller comments specific to Prepare and Get methods - https://github.com/stellar/go/pull/5699/commits/581da0edfe4e0aa620077b8df8a97edecc07f56a

ingest/ledgerbackend/rpc_backend.go
sreuland: Yes, reduced the content.

ingest/ledgerbackend/rpc_backend.go
sreuland: agreed, I went with a config struct and reduced to single constructor fn - https://github.com/stellar/go/pull/5699/commits/81767a0d403934768a32611f2fe043d834fb3779

ingest/ledgerbackend/rpc_backend.go
```diff
@@ -0,0 +1,293 @@
...
+	Sequence     uint32
+	LatestLedger uint32
+}
+
+func (e *RPCLedgerBeyondLatestError) Error() string {
+	return fmt.Sprintf("ledger %d is beyond the RPC latest ledger is %d", e.Sequence, e.LatestLedger)
+}
+
+// The minimum required RPC client interface required for usage by RPCLedgerBackend.
+type RPCClient interface {
```
2opremio:
```suggestion
type LedgerGetter interface {
```

ingest/ledgerbackend/rpc_backend.go
```diff
@@ -0,0 +1,293 @@
...
+	GetLedgers(ctx context.Context, req protocol.GetLedgersRequest) (protocol.GetLedgersResponse, error)
+}
+
+type RPCLedgerBackend struct {
+	client        RPCClient
+	buffer        map[uint32]xdr.LedgerCloseMeta
+	bufferSize    uint32
```
2opremio:
```suggestion
	limit    uint32
```

the name `bufferSize` makes it seem that it would be equal to `len(buffer)`

ingest/ledgerbackend/rpc_backend.go
```diff
@@ -0,0 +1,293 @@
...
+	if len(b.buffer) == 0 {
+		return 0, nil
+	}
+
+	maxLedgerSequence := uint32(0)
+	for seq := range b.buffer {
+		if seq > maxLedgerSequence {
+			maxLedgerSequence = seq
+		}
+	}
```
2opremio: Why not keep the latest sequence pre-stored? Also, an atomic could be used to avoid using a lock for this

2opremio (approved):

ingest/ledgerbackend/rpc_backend.go
sreuland: nice reduction, done - https://github.com/stellar/go/pull/5699/commits/2229eb53227a0cd0029a567d94047798fd6b7ce0, thx!

PR #5708, opened 2025-05-22
services/galexie: add AWS S3 datastore plugin

hawkaa:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Add AWS S3 pluggable DataStore backend for Galexie, alongside existing GCS support.  
Update developer documentation and example config to reflect selection and configuration of both backends.

### Why

Adds broader cloud support for Galexie users and clarifies how to configure and extend storage backends with concise, improved documentation.

### Known limitations

- S3 feature set covers main workflow, but advanced options (e.g. custom endpoints, advanced KMS) aren’t comprehensively tested.
- Authored and contributed by a Go newcomer—please review for typical Go best practices and plugin idioms.


support/datastore/datastore.go
```diff
@@ -34,6 +34,16 @@ func NewDataStore(ctx context.Context, datastoreConfig DataStoreConfig) (DataSto
...
 		}
 		return NewGCSDataStore(ctx, destinationBucketPath, datastoreConfig.Schema)
+	case "S3":
+		// Basic validation in factory for clearer config errors
+		if _, ok := datastoreConfig.Params["destination_bucket"]; !ok {
+			return nil, errors.New("S3 config missing required parameter: destination_bucket")
+		}
+		if _, ok := datastoreConfig.Params["region"]; !ok {
+			return nil, errors.New("S3 config missing required parameter: region")
+		}
```
sreuland, line 44: for consideration, since passing the `Params` map to the factory `NewS3DataStore` these validation rules could be moved into there as well for consolidation of parameter parsing/validation to the s3 datastore layer.

support/datastore/s3_datastore.go
```diff
@@ -0,0 +1,350 @@
...
+	fullPath := s.fullPath(objectPath)
+	headInput := &s3.HeadObjectInput{
+		Bucket: aws.String(s.bucket),
+		Key:    aws.String(fullPath),
+	}
+
+	headOutput, err := s.s3Client.HeadObject(ctx, headInput)
+	if err != nil {
+		var apiErr smithy.APIError
+		if errors.As(err, &apiErr) && (apiErr.ErrorCode() == "NotFound" || apiErr.ErrorCode() == "NoSuchKey") {
```
urvisavla, line 197: Consider avoiding hardcoded strings for error handling and replace it with something like this for eg:
```
var noSuchKeyErr *types.NoSuchKey
if errors.As(err, &noSuchKeyErr)
```

support/datastore/s3_datastore.go
```diff
@@ -0,0 +1,350 @@
...
+	fullPath := s.fullPath(objectPath)
+	getInput := &s3.GetObjectInput{
+		Bucket: aws.String(s.bucket),
+		Key:    aws.String(fullPath),
+	}
+
+	getOutput, err := s.s3Client.GetObject(ctx, getInput)
+	if err != nil {
+		var apiErr smithy.APIError
+		if errors.As(err, &apiErr) && (apiErr.ErrorCode() == "NotFound" || apiErr.ErrorCode() == "NoSuchKey") {
```
urvisavla, line 217: Same consideration as above regarding error checking

support/datastore/s3_datastore.go
```diff
@@ -0,0 +1,350 @@
...
+	if err != nil {
+		return stellarErrors.Wrapf(err, "failed to upload object to S3: %s", fullPath)
+	}
+
+	return nil
+}
+
+// PutFileIfNotExists uploads data only if the object does not already exist in S3.
+// It returns true if the file was uploaded, false if it already existed.
+func (s *S3DataStore) PutFileIfNotExists(ctx context.Context, objectPath string, in io.WriterTo, metaData map[string]string) (bool, error) {
```
tamirms, line 268: S3 supports [conditional PUTs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/conditional-writes.html) which allows you to execute an upload only if the object does not already exists. We use the same concept in the GCS datastore because it is robust against  race conditions where multiple unloaders attempts to PUT the same object.

I think you can implement conditional PUTs using the go s3 client by configuring the `IfNoneMatch` field of the [PutObjectInput](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/s3#PutObjectInput) struct.

support/datastore/s3_datastore.go
```diff
@@ -0,0 +1,350 @@
...
+func (s *S3DataStore) PutFile(ctx context.Context, objectPath string, in io.WriterTo, metaData map[string]string) error {
+	fullPath := s.fullPath(objectPath)
+
+	buf := new(bytes.Buffer) // Buffer to hold data from io.WriterTo
+	_, err := in.WriteTo(buf)
+	if err != nil {
+		return stellarErrors.Wrap(err, "failed to write data to buffer for S3 upload")
+	}
+
+	uploadInput := &s3.PutObjectInput{
```
tamirms, line 238: it is also possible to include a checksum which is validated by s3 to ensure the object was updated properly:

https://docs.aws.amazon.com/AmazonS3/latest/userguide/checking-object-integrity.html

We use a crc32 checksum in the GCS datastore implementation to ensure the objects are uploaded properly. The checksum is also validated when fetching objects

support/datastore/s3_datastore.go
urvisavla, line 268: One consideration in the datastore implementation is enabling multiple Galexie instances to write to the same datastore without overwriting objects, which is achieved through `PutFileIfNotExists`. This implementation `PutFileIfNotExists` isn't atomic, meaning the file could be created between the existence check and upload.  In GCS, to address this, we use preconditions to ensure the upload only succeeds if the file doesn't exist. S3 offers a similar feature with preconditions, where setting `IfNoneMatch` in the `PutObject` request ensures the upload only happens if the file doesn't exist, performing the operation atomically.

You can refer to [S3 Conditional Uploads](https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObject.html) for more information. And here is the [GCS implementation](https://github.com/stellar/go/blob/d5767df5b21a1c34a5cb05b158d9b0962c94e7bf/support/datastore/gcs_datastore.go#L99-L115) for reference.

services/galexie/DEVELOPER_GUIDE.md
```diff
@@ -59,16 +67,16 @@ $ GALEXIE_INTEGRATION_TESTS_ENABLED=true go test -v -race -run TestGalexieTestSu
...
+- Add support for new datastore-specific features. Implement any datastore-specific custom logic. Different datastores have different ways of handling
   - race conditions
   - automatic retries
   - metadata storage, etc.
 - Add the new datastore to the factory function [NewDataStore](https://github.com/stellar/go/blob/master/support/datastore/datastore.go).
-- Add a [config](./config.example.toml) section for the new storage type. This may include configurations like destination, authentication information etc.
-- An emulator such as a GCS emulator [fake-gcs-server](https://github.com/fsouza/fake-gcs-server) can be used for testing without connecting to real cloud storage.
+- Add a [config](./config.example.toml) section for the new storage type. This may include configurations like destination, authentication information, etc.
+- You can use tools like [fake-gcs-server](https://github.com/fsouza/fake-gcs-server) (for GCS emulation) or [localstack](https://github.com/localstack/localstack) / [minio](https://min.io/) (for S3 emulation) to test your plugins without real cloud accounts.
```
urvisavla, line 79: nice!

support/datastore/s3_datastore.go
```diff
@@ -0,0 +1,350 @@
...
+	if err != nil {
+		return nil, stellarErrors.Wrap(err, "failed to load AWS config")
+	}
+
+	// --- Handle Path Style Addressing ---
+	s3ClientOptions := []func(*s3.Options){
+		func(o *s3.Options) {
+			// Check both the config param AND the standard env var AWS_S3_USE_PATH_STYLE
+			// Environment variable takes precedence if set.
+			usePathStyleEnv := os.Getenv("AWS_S3_USE_PATH_STYLE")
```
sreuland, line 146: what is driving the need to allow this optional override from env? It seems inconsistent in regards to the rest of optional s3 behavioral settings which are sourced only from config file params.

tamirms:
We have [tests](https://github.com/stellar/go/blob/master/support/datastore/gcs_test.go) for the GCS datastore which use a GCS emulator to test the end to end behavior of the datastore operations. I think we can take a similar approach and use [minio](https://github.com/minio/minio) as an s3 servicefor end to end testing.

support/datastore/s3_datastore.go
urvisavla: Ha, you beat me to it and are way more articulate 😄

services/galexie/config.example.toml
```diff
@@ -7,13 +7,33 @@ admin_port = 6061
...
 destination_bucket_path = "your-bucket-name/<optional_subpath1>/<optional_subpath2>/"
 
+# AWS S3 EXAMPLE CONFIGURATION:
+# Uncomment the below to use AWS S3 as the storage backend.
+# [datastore_config]
+# type = "S3"
+#
+# [datastore_config.params]
+# # Required for S3:
+# destination_bucket = "your-s3-bucket-name"
```
urvisavla, line 24: > # # destination_path_prefix = "ledgers/export"          # Prefix within the S3 bucket
> # destination_bucket = "your-s3-bucket-name"

Consider simplifying this to just one `destination_bucket_path` config variable, which includes the full path with the prefix and derive the bucket name and prefix internally. This way we can reduce the number of configurable parameters. Check the gcs implementation, it does something similar.

support/datastore/s3_datastore.go
```diff
@@ -0,0 +1,350 @@
...
+	if s.prefix == "" {
+		return trimmedObjectPath
+	}
+	// path.Join cleans multiple slashes and relative paths.
+	// If objectPath starts with "/", path.Join might treat it as absolute,
+	// so ensure objectPath is relative if prefix is used.
+	return path.Join(s.prefix, trimmedObjectPath)
+}
+
+// GetFileMetadata retrieves the custom user metadata associated with an S3 object.
```
sreuland, line 186: It might be helpful to mention in this doc comment for any of the Datastore method implementations which reference file metadata to further qualify the schema of the map will have known keys defined on the [datastore.Metadata](https://github.com/goldsky-io/stellar-go-public/blob/hakon/s3-datastore/support/datastore/metadata.go#L5)

btw @urvisavla, do you think longer term these `Datastore` methods which reference export file metadata could use the stronger typed `datastore.Metadata` instead of map?

support/datastore/s3_datastore_test.go
```diff
@@ -0,0 +1,103 @@
...
+	paramsMissingRegion := map[string]string{"destination_bucket": "test-bucket"}
+	_, err = NewS3DataStore(ctx, paramsMissingRegion, schema)
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "missing required parameter: region")
+}
+
+// Note: Testing successful creation requires mocking AWS config loading or having valid
+// dummy credentials available in the test environment (e.g., via env vars).
+// This basic test focuses on parameter validation.
+// More comprehensive tests would mock the S3 client interactions.
```
sreuland, line 31: +1 for mocking at some point to enable lightweight unit test coverage. 

since `manager.Uploader` and `s3.Client` are structs in the aws SDK, could define local interfaces in this package which just capture the methods used on those by S3DataStore and then tests can inject mocks of those into an S3DataStore instance.

support/datastore/s3_datastore.go
```diff
@@ -0,0 +1,350 @@
...
+	"strings"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/config"
+	"github.com/aws/aws-sdk-go-v2/feature/s3/manager"
+	"github.com/aws/aws-sdk-go-v2/service/s3"
+	"github.com/aws/aws-sdk-go-v2/service/s3/types"
+	"github.com/aws/smithy-go"
+	"github.com/aws/smithy-go/logging"
+	stellarErrors "github.com/stellar/go/support/errors"
```
Shaptic, line 22: We've been trying to move away from this package which is basically an ancient ancestor that was written prior to Go's native [`errors`](https://pkg.go.dev/errors) package being actually decent. Instead of `Wrapf` we've been using `errors.Join`, `fmt.Errorf`, etc.

support/datastore/s3_datastore.go
overcat: When uploading, the SDK will automatically add a CRC32 checksum. When using `GetObject`, setting `ChecksumMode: types.ChecksumModeEnabled` allows for verification of the CRC32 checksum during download.

overcat:
Hi @hawkaa, I saw you closed this PR yourself, and I really appreciate your understanding. 

I still want to sincerely apologize. I was in a rush for this feature and completely missed the great work you had already done here. I feel bad that your effort was duplicated.

Your approach was very insightful. Thanks again for everything! ♥️

PR #5710, opened 2025-05-23
Karthik/p23 sac event support

karthikiyer56:
This PR generates a new contractEvents package that is internal to horizon.
This new package accounts for events from txMetav3 and txMetav4 in accordance with the event format specified in cap-67.

The reason for the new package **that is internal to horizon** is because of https://github.com/stellar/go/issues/5669

### DIFF
--- a/ingest/ledger_transaction.go
+++ b/ingest/ledger_transaction.go
@@ -247,14 +247,18 @@ func (t *LedgerTransaction) GetContractEvents(opIndex uint32) ([]xdr.ContractEve
 }
 
 // GetDiagnosticEvents returns all contract events emitted by a given operation.
+// TODO: update comments
 func (t *LedgerTransaction) GetDiagnosticEvents() ([]xdr.DiagnosticEvent, error) {
 	switch t.UnsafeMeta.V {
 	case 1, 2:
 		return nil, nil
 	case 3:
-		return t.UnsafeMeta.GetDiagnosticEventsV3()
+		return t.UnsafeMeta.GetDiagnosticEventsV3(), nil
 	case 4:
-		return nil, nil
+		txMeta := t.UnsafeMeta.MustV4()
+		//TODO: is this correct?
+		return txMeta.DiagnosticEvents, nil
+
 	default:
 		return nil, fmt.Errorf("unsupported TransactionMeta version: %v", t.UnsafeMeta.V)
 	}
--- a/ingest/ledger_transaction.go
+++ b/ingest/ledger_transaction.go
@@ -242,26 +242,71 @@ func (t *LedgerTransaction) operationChanges(ops operationsMeta, index uint32) [
 	return changes
 }
 
-func (t *LedgerTransaction) GetContractEvents(opIndex uint32) ([]xdr.ContractEvent, error) {
-	return t.UnsafeMeta.GetContractEvents(opIndex)
+func (t *LedgerTransaction) GetContractEventsForOperation(opIndex uint32) ([]xdr.ContractEvent, error) {
+	return t.UnsafeMeta.GetContractEventsForOperation(opIndex)
 }
 
-// GetDiagnosticEvents returns all contract events emitted by a given operation.
-// TODO: update comments
+// GetSorobanContractEvents returns a []xdr.ContractEvent for the smart contract transaction
+// For getting soroban smart contract events,we rely on the fact that there will only be one operation present in the transaction
+func (t *LedgerTransaction) GetSorobanContractEvents() ([]xdr.ContractEvent, error) {
+	if !t.IsSorobanTx() {
+		return nil, errors.New("not a soroban transaction")
+	}
+	return t.GetContractEventsForOperation(0)
+}
+
+// GetDiagnosticEvents returns strictly diagnostic events emitted by a given transaction.
+/*
+	Please note that, depending on the configuration with which txMeta may be generated,
+	it is possible that, for smart contract transactions, the list of generated diagnostic events MAY include contract events as well
+	Users of this function (horizon, rpc, etc) should be careful not to double count diagnostic events and contract events in that case
+*/
 func (t *LedgerTransaction) GetDiagnosticEvents() ([]xdr.DiagnosticEvent, error) {
+	return t.UnsafeMeta.GetDiagnosticEvents()
+}
+
+// GetTransactionEvents gives the breakdown of xdr.ContractEvent, xdr.TransactionEvent, xdr.Disgnostic event as they appea in the TxMeta
+/*
+	In TransactionMetaV3, soroban transaction events and contract events appear in the SorobanMeta struct, i.e. at the top level
+	In TransactionMetaV4 and onwards, there is a more granular breakdown, because of CAP-67 unified events
+	- Contract events will also be present in the "operation []OperationMetaV2" in  structure.
+	- Classic operations will also have contract events.
+	- For smart contract transactions, contract events are going to show up in its respective operation index in []OperationMetaV2
+	- Additionally, if its a smart contract transaction, the contract events will also be included in the "DiagnosticEvents []DiagnosticEvent" structure
+
+	It is preferred to use this function in horizon and rpc
+*/
+func (t *LedgerTransaction) GetTransactionEvents() (TransactionEvents, error) {
+	txEvents := TransactionEvents{}
 	switch t.UnsafeMeta.V {
 	case 1, 2:
-		return nil, nil
+		return txEvents, nil
 	case 3:
-		return t.UnsafeMeta.GetDiagnosticEventsV3(), nil
+		contractEvents, err := t.GetSorobanContractEvents()
+		if err != nil {
+			return txEvents, err
+		}
+		diagnosticEvents, err := t.GetDiagnosticEvents()
+		if err != nil {
+			return txEvents, err
+		}
+		var opEvents [][]xdr.ContractEvent
+		opEvents = append(opEvents, contractEvents)
+		txEvents.OperationEvents = opEvents
+		txEvents.DiagnosticEvents = diagnosticEvents
 	case 4:
 		txMeta := t.UnsafeMeta.MustV4()
-		//TODO: is this correct?
-		return txMeta.DiagnosticEvents, nil
-
+		txEvents.TransactionEvents = txMeta.Events
+		txEvents.DiagnosticEvents = txMeta.DiagnosticEvents
+		txEvents.OperationEvents = make([][]xdr.ContractEvent, len(txMeta.Operations))
+		for _, op := range txMeta.Operations {
+			txEvents.OperationEvents = append(txEvents.OperationEvents, op.Events)
+		}
 	default:
-		return nil, fmt.Errorf("unsupported TransactionMeta version: %v", t.UnsafeMeta.V)
+		return txEvents, fmt.Errorf("unsupported TransactionMeta version: %v", t.UnsafeMeta.V)
 	}
+	return txEvents, nil
+
 }
 
 func (t *LedgerTransaction) ID() int64 {
--- a/ingest/ledger_transaction.go
+++ b/ingest/ledger_transaction.go
@@ -267,12 +267,13 @@ func (t *LedgerTransaction) GetDiagnosticEvents() ([]xdr.DiagnosticEvent, error)
 
 // GetTransactionEvents gives the breakdown of xdr.ContractEvent, xdr.TransactionEvent, xdr.Disgnostic event as they appea in the TxMeta
 /*
-	In TransactionMetaV3, soroban transaction events and contract events appear in the SorobanMeta struct, i.e. at the top level
+	In TransactionMetaV3, for soroban transactions, contract events and diagnostic events appear in the SorobanMeta struct in TransactionMetaV3, i.e. at the transaction level
 	In TransactionMetaV4 and onwards, there is a more granular breakdown, because of CAP-67 unified events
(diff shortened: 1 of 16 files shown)


xdr/transaction_meta.go
```diff
@@ -4,50 +4,52 @@ import (
...
-				// in tx meta's DiagnosticEvents, no need to look anywhere else for events
-				return diagnosticEvents, nil
-			}
-
-			contractEvents = sorobanMeta.Events
-			if len(contractEvents) == 0 {
-				// no events were present in this tx meta
-				return nil, nil
-			}
+		sorobanMeta := t.MustV3().SorobanMeta
```
karthikiyer56, line 32: This logic, imo, doesnt belong here.
This shud be accounted for in horizon/rpc code, where you call the function and then do "things" with it.
Actually, we shouldnt be using this function going forward in rpc or horizon.

We shud be using the new function I added in `ledger_transaction.go`. - `func (t *LedgerTransaction) GetTransactionEvents() (TransactionEvents, error)`

ingest/ledger_transaction.go
```diff
@@ -236,15 +242,73 @@ func (t *LedgerTransaction) operationChanges(ops operationsMeta, index uint32) [
...
+
+	It is preferred to use this function in horizon and rpc
+*/
+func (t *LedgerTransaction) GetTransactionEvents() (TransactionEvents, error) {
+	txEvents := TransactionEvents{}
+	switch t.UnsafeMeta.V {
+	case 1, 2:
+		return txEvents, nil
+	case 3:
+		contractEvents, err := t.GetSorobanContractEvents()
```
tamirms, line 285: if the transaction is not a soroban transaction you will return an error here. however, in other versions of transaction meta, there will be no error. To be consistent with the other versions of transaction meta, I think this function should return an empty list and no error if the transaction is not a soroban transaction.

ingest/ledger_transaction.go
```diff
@@ -236,15 +242,73 @@ func (t *LedgerTransaction) operationChanges(ops operationsMeta, index uint32) [
...
 }
 
-// GetDiagnosticEvents returns all contract events emitted by a given operation.
+// GetDiagnosticEvents returns strictly diagnostic events emitted by a given transaction.
+/*
+	Please note that, depending on the configuration with which txMeta may be generated,
+	it is possible that, for smart contract transactions, the list of generated diagnostic events MAY include contract events as well
+	Users of this function (horizon, rpc, etc) should be careful not to double count diagnostic events and contract events in that case
+*/
 func (t *LedgerTransaction) GetDiagnosticEvents() ([]xdr.DiagnosticEvent, error) {
```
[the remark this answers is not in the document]
karthikiyer56: I mean this thing has a place. 
If someone wants to use ingest for just getting diagnostic events, no?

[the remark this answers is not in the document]
Shaptic: Yeah actually if we leave this in place then RPC will work for Horizon integration testing as-is, i.e. no need for "proper" ordering since y'all don't care about it.

services/horizon/internal/ingest/contractevents/events.go
```diff
@@ -0,0 +1,376 @@
...
+)
+
+// StellarAssetContractEvent represents a parsed SAC event
+type StellarAssetContractEvent struct {
+	Type            EventType
+	Asset           xdr.Asset
+	From            string // For transfer, burn, clawback
+	To              string // For transfer, mint
+	Amount          xdr.Int128Parts
+	destinationMemo xdr.Memo // Can be uint64, []byte, or string (V4 only)
```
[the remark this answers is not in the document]
karthikiyer56: done.
Please note that I have not updated horizon code to surface this field as a part of GET endpoints.
I will do that separately. I want this to go out ASAP

xdr/transaction_meta.go
karthikiyer56: I have refactored horizon to use  the new function `GetSorobanContractEvents`.
Atleast for horizon's sake, its more apt, since we only carr for contract events for a successful smart contract operation

services/horizon/internal/ingest/processors/effects_processor_test.go
```diff
@@ -3768,6 +3785,11 @@ func makeInvocationTransaction(
 	envelope := xdr.TransactionV1Envelope{
 		Tx: xdr.Transaction{
 			// the rest doesn't matter for effect ingestion
+			Ext: xdr.TransactionExt{
+				V: 1,
+				// sorobanData is needed to pass the check for IsSorobanTx
+				SorobanData: &xdr.SorobanTransactionData{},
```
Shaptic, line 3791: This kind of workaround makes me wonder if `isSorobanTx` should check differently :thinking:

ingest/ledger_transaction.go
```diff
@@ -236,15 +242,78 @@ func (t *LedgerTransaction) operationChanges(ops operationsMeta, index uint32) [
...
 
-func (t *LedgerTransaction) GetContractEvents() ([]xdr.ContractEvent, error) {
-	return t.UnsafeMeta.GetContractEvents()
+func (t *LedgerTransaction) GetContractEventsForOperation(opIndex uint32) ([]xdr.ContractEvent, error) {
+	return t.UnsafeMeta.GetContractEventsForOperation(opIndex)
+}
+
+// GetSorobanContractEvents returns a []xdr.ContractEvent for the smart contract transaction
+// For getting soroban smart contract events,we rely on the fact that there will only be one operation present in the transaction
+func (t *LedgerTransaction) GetSorobanContractEvents() ([]xdr.ContractEvent, error) {
```
Shaptic, line 251: Suggestion: it should just be `GetContractEvents()`, since we're generally trying to move aware from using the Soroban nomenclature and this is an exported function.

ingest/ledger_transaction.go
```diff
@@ -236,15 +242,78 @@ func (t *LedgerTransaction) operationChanges(ops operationsMeta, index uint32) [
...
+	if !t.IsSorobanTx() {
+		return nil, errors.New("not a soroban transaction")
+	}
+	return t.GetContractEventsForOperation(0)
 }
 
-// GetDiagnosticEvents returns all contract events emitted by a given operation.
+// GetDiagnosticEvents returns strictly diagnostic events emitted by a given transaction.
```
Shaptic, line 258: Imo this and other docstrings should probably be reformatted to follow [godocs](https://tip.golang.org/doc/comment#func)

ingest/ledger_transaction.go
```diff
@@ -236,15 +242,78 @@ func (t *LedgerTransaction) operationChanges(ops operationsMeta, index uint32) [
...
+		if err != nil {
+			return txEvents, err
+		}
+		diagnosticEvents, err := t.GetDiagnosticEvents()
+		if err != nil {
+			return txEvents, err
+		}
+		var opEvents [][]xdr.ContractEvent
+		opEvents = append(opEvents, contractEvents)
+		txEvents.OperationEvents = opEvents
```
Shaptic, line 300: Isn't this just 
```suggestion
		txEvents.OperationEvents = contractEvents
```

processors/operation/operation.go
```diff
@@ -1854,14 +1843,14 @@ func FilterEvents(diagnosticEvents []xdr.DiagnosticEvent) []xdr.ContractEvent {
...
-	diagnosticEvents, err := operation.Transaction.GetDiagnosticEvents()
+	contractEvents, err := operation.Transaction.GetSorobanContractEvents()
 	if err != nil {
 		// this operation in this context must be an InvokeHostFunctionOp, therefore V3Meta should be present
 		// as it's in same soroban model, so if any err, it's real,
 		return nil, err
 	}
 
-	for _, contractEvent := range FilterEvents(diagnosticEvents) {
+	for _, contractEvent := range contractEvents {
```
Shaptic, line 1853: I think this is correct behavior, but I still think it's worth noting that this changes the behavior to consider events from all Soroban transactions rather than only successful ones. It's still correct because we want to (a) consider fee events and (b) are pulling from the event list directly rather than the flattened `DiagnosticEvent[]` list which precluded this filter, but just wanted to note it.

services/horizon/internal/ingest/contractevents/events.go
```diff
@@ -0,0 +1,376 @@
...
+		Type:   eventType,
+		Asset:  asset,
+		Amount: amount,
+	}
+
+	switch eventType {
+	case EventTypeTransfer:
+		if err := parseTransferEvent(topics, sacEvent); err != nil {
+			return nil, err
+		}
```
Shaptic, line 244: each of these could be 
```suggestion
		return sacEvent, parseTransferEvent(topics, sacEvent)
```
since `parse<Type>Event` modifies `sacEvent` by reference, right?

Shaptic: Alternatively, just assign `err` in each `case` and only `if err != nil` at the end.

services/horizon/internal/ingest/contractevents/events.go
```diff
@@ -0,0 +1,376 @@
...
+	case EventTypeClawback:
+		if err := parseClawbackEventFromTxMetaV3(topics, sacEvent); err != nil {
+			return nil, err
+		}
+	case EventTypeBurn:
+		if err := parseBurnEvent(topics, sacEvent); err != nil {
+			return nil, err
+		}
+	default:
+		return nil, fmt.Errorf("%w: %v", ErrEventUnsupported, eventType)
```
Shaptic, line 258: suggestion:
```suggestion
		return nil, errors.Join(ErrEventUnsupported, fmt.Errorf("%v", eventType))
```

though it only really matters if someone wants to `errors.Is(err, ErrEventUnsupported)` elsewhere.

services/horizon/internal/ingest/contractevents/events.go
```diff
@@ -0,0 +1,376 @@
...
+		switch string(key) {
+		case "amount":
+			amount, ok = entry.Val.GetI128()
+			if !ok {
+				return amount, memo, errors.New("amount field is not i128")
+			}
+			foundAmount = true
+
+		case "to_muxed_id":
+			// Handle different types for to_muxed_id.. do i need a default to error?
```
Shaptic, line 348: I think this comment isn't needed anymore? You do `default` to an error.

services/horizon/internal/ingest/contractevents/events.go
```diff
@@ -0,0 +1,376 @@
...
+					memo = xdr.MemoHash(xdr.Hash(val[:]))
+				}
+			case xdr.ScValTypeScvString:
+				if val, ok := entry.Val.GetStr(); ok {
+					memo = xdr.MemoText(string(val))
+				}
+			default:
+				return amount, memo, fmt.Errorf("invalid to_muxed_id type for data: %s", entry.Val.Type)
+			}
+		}
```
Shaptic, line 366: Should there be a `default` case here for any other `string(key)` values?

tamirms:
apart from @Shaptic 's comments, this PR looks good to me

services/horizon/internal/ingest/contractevents/events.go
karthikiyer56: it will have to be the latter, since the parseXXXEvent field return type is `error` instead of `(*StellarAssetContractEvent, error) `

services/horizon/internal/ingest/contractevents/events.go
karthikiyer56: erm, in the `errors.Join()`,  `fmt.Errorf("%v", eventType)` is not really indicating an error, is it?
As in, I am not joining 2 different kinds of errors. 
I am trying to add more context to the unsupported type of error by appending event type.
Hence the choice of `fmt.Errorf("%w:%v)`

It was my understanding that you would use errors.Join for something like this:
 `errors.Join(HasBadInput1Parameter, AlsoHasBadInput2Parameter,...)`

I am not particularly opionated on this, and am happy to make the change, just want to understand it, is all

services/horizon/internal/ingest/contractevents/events.go
karthikiyer56: not necessary, since between the `length == 2` check and the check for `amount` and `to_muxed_id`, we got it covered.
I will have to update more than a few test cases, if I make this change 🙈

services/horizon/internal/ingest/processors/effects_processor_test.go
karthikiyer56: I dont think so.
 If anything, the test fixtures shud be sanitized/refactored to account for smart contract data check if they are testing some path in the `InvokeHostFunction` switch case. 

 
We discussed this in the past and landed on --> the best way to check for `if its a smart contract` is to look inside the `envelope.SorobanData`  (during ingress) **as opposed** to looking at `TxMeta.SorobanMeta` (during egress), since it is possible that a smart contract invocation produces no meta at all(?)

ingest/ledger_transaction.go
karthikiyer56: done

services/horizon/internal/ingest/contractevents/events.go
karthikiyer56: removed

processors/operation/operation.go
Shaptic: Resolved offline: fee events aren't relevant at the operation level

ingest/ledger_transaction.go
Shaptic: oh derp nice

Shaptic (approved):
Really nice work and great test coverage.

PR #5713, opened 2025-05-28
Fix GetDiagnosticEvents() for transaction meta v3

2opremio:
https://github.com/stellar/go/pull/5710 changed the events obtained for meta v3 (breaking the API)

### DIFF
--- a/xdr/transaction_meta.go
+++ b/xdr/transaction_meta.go
@@ -33,11 +33,33 @@ func (t *TransactionMeta) GetDiagnosticEvents() ([]DiagnosticEvent, error) {
 	case 1, 2:
 		return nil, nil
 	case 3:
-		sorobanMeta := t.MustV3().SorobanMeta
-		if sorobanMeta == nil {
-			return nil, nil
+		var diagnosticEvents []DiagnosticEvent
+		var contractEvents []ContractEvent
+		if sorobanMeta := t.MustV3().SorobanMeta; sorobanMeta != nil {
+			diagnosticEvents = sorobanMeta.DiagnosticEvents
+			if len(diagnosticEvents) > 0 {
+				// all contract events and diag events for a single operation(by its index in the tx) were available
+				// in tx meta's DiagnosticEvents, no need to look anywhere else for events
+				return diagnosticEvents, nil
+			}
+
+			contractEvents = sorobanMeta.Events
+			if len(contractEvents) == 0 {
+				// no events were present in this tx meta
+				return nil, nil
+			}
 		}
-		return sorobanMeta.DiagnosticEvents, nil
+
+		// tx meta only provided contract events, no diagnostic events, we convert the contract
+		// event to a diagnostic event, to fit the response interface.
+		convertedDiagnosticEvents := make([]DiagnosticEvent, len(contractEvents))
+		for i, event := range contractEvents {
+			convertedDiagnosticEvents[i] = DiagnosticEvent{
+				InSuccessfulContractCall: true,
+				Event:                    event,
+			}
+		}
+		return convertedDiagnosticEvents, nil
 	case 4:
 		return t.MustV4().DiagnosticEvents, nil
 	default:


xdr/transaction_meta.go
```diff
@@ -33,11 +33,33 @@ func (t *TransactionMeta) GetDiagnosticEvents() ([]DiagnosticEvent, error) {
 	case 1, 2:
 		return nil, nil
 	case 3:
-		sorobanMeta := t.MustV3().SorobanMeta
-		if sorobanMeta == nil {
-			return nil, nil
+		var diagnosticEvents []DiagnosticEvent
+		var contractEvents []ContractEvent
```
karthikiyer56, line 37: @2opremio : this is deliberately changed.
The motivation was to get the diagnostic events to be strictly only showing diagnostic events as they appear in the txmeta.


This was discussed with @Shaptic  and @tamirms 
RPC will need to change the code to either use the LedgerTransaction.GetTransactionEvents

2opremio: At the very least we should add a comment to `GetDiagnosticEvents()` indicating that we the behavior has changed.

2opremio: I started using `GetTransactionEvents ` but it doesn't seem to be backwards compatible with meta v3 (there are missing events)

2opremio: I guess that `sorobanMeta.Events` is not added anywhere

karthikiyer56: when TxMetaVersion == 3:
`LedgerTransaction.GetTransactionEvents `code is like so
```
	case 3:
		// There wont be any events for classic operations in TxMetaV3
		if !t.IsSorobanTx() {
			return txEvents, nil
		}
		contractEvents, err := t.GetContractEvents()
		if err != nil {
			return txEvents, err
		}
		diagnosticEvents, err := t.GetDiagnosticEvents()
		if err != nil {
			return txEvents, err
		}
		// There  will only ever be 1 smart contract operation per tx.
		txEvents.OperationEvents = make([][]xdr.ContractEvent, 1)
		txEvents.OperationEvents[0] = contractEvents
		txEvents.DiagnosticEvents = diagnosticEvents
```
		
The t.getContractEvents is like so
```
// GetContractEvents returns a []xdr.ContractEvent for pnly smart contract transaction.
// If it is not a smart contract transaction, it throws an error
// For getting events from classic operations/transaction, use GetContractEventsForOperation
// For getting soroban smart contract events,we rely on the fact that there will only be one operation present in the transaction
func (t *LedgerTransaction) GetContractEvents() ([]xdr.ContractEvent, error) {
	if !t.IsSorobanTx() {
		return nil, errors.New("not a soroban transaction")
	}
	return t.GetContractEventsForOperation(0)
}
```

and `t.GetContractEventsForOperation()` is like so
```
func (t *LedgerTransaction) GetContractEventsForOperation(opIndex uint32) ([]xdr.ContractEvent, error) {
	return t.UnsafeMeta.GetContractEventsForOperation(opIndex)
}
```
and finally
```
func (t *TransactionMeta) GetContractEventsForOperation(opIndex uint32) ([]ContractEvent, error) {
	switch t.V {
	case 1, 2:
		return nil, nil
	// For TxMetaV3, events appear in the TxMetaV3.SorobanMeta.Events, and we dont need to rely on opIndex
	case 3:
		sorobanMeta := t.MustV3().SorobanMeta
		if sorobanMeta == nil {
			return nil, nil
		}
		return sorobanMeta.Events, nil
	// TxMetaV4 includes unified CAP-67 events that appear at the operation level
	// To fetch soroban contract events from TxMetaV4, you will need to pass in the operationIndex 0.
	case 4:
		return t.MustV4().Operations[opIndex].Events, nil
	default:
		return nil, fmt.Errorf("unsupported TransactionMeta version: %v", t.V)
	}
}
```

There is also a comment to that effect:
`// For TxMetaV3, events appear in the TxMetaV3.SorobanMeta.Events, and we dont need to rely on opIndex`
So, I am not quite following what is it that is failing?
Unit tests in rpc or something else?

PR #5718, opened 2025-06-10
ingest/ledgerbackend: fix rpc ledger backend retries

sreuland:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

Added new error code checking on GetLedgers response from RPC.

### Why

RPC may emit a jrpc2 Error from GetLedgers when a request exceeds internal duration limits configured in rpc, if this type of error does occur, the rpc ledger backend as client should retry, it's not a fatal error.

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -20,13 +20,12 @@ jobs:
           fetch-depth: 0 # required for new-from-rev option in .golangci.yml
       - uses: ./.github/actions/setup-go
       - name: Run golangci-lint
-        uses: golangci/golangci-lint-action@58eda26a511c265ee35b3ee4b101fb8adfd76480 # version v6.1.1
+        uses: golangci/golangci-lint-action@v7
         with:
-          version: v1.64.8 # this is the golangci-lint version
+          version: v2.1.6
           args: -v --issues-exit-code=0 --config=.golangci.yml # exit without errors for now - won't fail the build
           github-token: ${{ secrets.GITHUB_TOKEN }}
           only-new-issues: true
-        continue-on-error: true # continue on error, shows as warning
 
 
 
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -26,6 +26,7 @@ jobs:
           args: -v --issues-exit-code=0 --config=.golangci.yml # exit without errors for now - won't fail the build
           github-token: ${{ secrets.GITHUB_TOKEN }}
           only-new-issues: true
+        continue-on-error: true # continue on error, shows as warning
 
 
 
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -23,7 +23,7 @@ jobs:
         uses: golangci/golangci-lint-action@58eda26a511c265ee35b3ee4b101fb8adfd76480 # version v6.1.1
         with:
           version: v1.64.8 # this is the golangci-lint version
-          args: --issues-exit-code=0 # exit without errors for now - won't fail the build
+          args: -v --issues-exit-code=0 --config=.golangci.yml # exit without errors for now - won't fail the build
           github-token: ${{ secrets.GITHUB_TOKEN }}
           only-new-issues: true
 
--- a/ingest/ledgerbackend/rpc_backend.go
+++ b/ingest/ledgerbackend/rpc_backend.go
@@ -9,6 +9,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/creachadair/jrpc2"
 	"github.com/stellar/go/xdr"
 	rpc "github.com/stellar/stellar-rpc/client"
 	"github.com/stellar/stellar-rpc/protocol"
@@ -25,13 +26,10 @@ func (e *RPCLedgerMissingError) Error() string {
 	return fmt.Sprintf("ledger %d was not present on rpc", e.Sequence)
 }
 
-type RPCLedgerBeyondLatestError struct {
-	Sequence     uint32
-	LatestLedger uint32
-}
+type RPCLedgerBeyondLatestError struct{}
 
-func (e *RPCLedgerBeyondLatestError) Error() string {
-	return fmt.Sprintf("ledger %d is beyond the RPC latest ledger is %d", e.Sequence, e.LatestLedger)
+func (e RPCLedgerBeyondLatestError) Error() string {
+	return "ledger is not available on the RPC server yet"
 }
 
 // The minimum required RPC client methods used by RPCLedgerBackend.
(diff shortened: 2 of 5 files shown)


ingest/ledgerbackend/rpc_backend.go
```diff
@@ -9,13 +9,16 @@ import (
...
+	"github.com/creachadair/jrpc2"
 	"github.com/stellar/go/xdr"
 	rpc "github.com/stellar/stellar-rpc/client"
 	"github.com/stellar/stellar-rpc/protocol"
 )
 
 const rpcBackendDefaultBufferSize uint32 = 10
 const rpcBackendDefaultWaitIntervalSeconds uint32 = 2
+const rpcErrRequestExceededProcessingLimitThreshold = -32001
+const rpcErrFailToProcessDueToInternalIssue = -32003
```
Shaptic: I think an `internal issue` only occurs when the endpoint handler crashes, meaning there's a bug in RPC itself - can you point out how this happens? I thought we fixed that in https://github.com/stellar/stellar-rpc/pull/407.

Shaptic: Oh duh that isn't released yet :facepalm:

sreuland: I was seeing the -32003 error response from GetLedgers [when the client submits a request](https://github.com/stellar/go/pull/5718/files#diff-5615c809e96a8356619de492af2344b7b5693e8057c73959cb244d61ff456dc8R247) with start ledger number that was rpc's LatestLedger+1 and a limit of 10. That request wasn't triggering an out of range validation error from RPC, it appeared to be going further into the handler, but the [rpc duration limit handler kicked in](https://github.com/stellar/stellar-rpc/blob/6bb9d7a387d51e25c0e0d2c3417e31da78312f68/cmd/stellar-rpc/internal/network/requestdurationlimiter.go#L298) and triggered the -32003 response. If resubmit same request after the failure response, it then would work. This is trying to emulate unbounded range as close as possible to produce same stream of ongoing/latest ledgers from network and in that context this internal retry to smooth over an rpc idiosyncrasy seemed reasonable.

sreuland: I will look into logs from rpc instance for more info on error stack related to the internal failure(panic) which bubbled up to duration limiit handler in this case. It should have text like `Request for method GetLedgers resulted in an error...` as there is a deferred in that duration handler to log the panic case which is when the -32003 gets emitted.

sreuland: yes, see a couple issues reported on this behavior, https://github.com/stellar/stellar-rpc/issues/391 was actually filed by me. I found the logs showing the [panic cause it was index out of range](https://github.com/stellar/stellar-rpc/issues/454#issuecomment-2960229227) which [stellar-rpc/407](https://github.com/stellar/stellar-rpc/pull/407) should prevent now. So, i just need to change this pr to adjust to how https://github.com/stellar/stellar-rpc/pull/407 will make `getLedgers` behavior consistent, i.e. it will return the out of range error. thanks!

sreuland: I updated the [GetLedger usage to utilize](https://github.com/stellar/go/pull/5718/files#diff-5615c809e96a8356619de492af2344b7b5693e8057c73959cb244d61ff456dc8R251) the outcome from https://github.com/stellar/stellar-rpc/pull/407, which means invalid requested ranges will respond with InvalidRequest jrpc error.

.github/workflows/golangci-lint.yml
```diff
@@ -15,18 +15,18 @@ jobs:
...
-        uses: golangci/golangci-lint-action@537aa1903e5d359d0b27dbc19ddd22c5087f3fbc # version v3.2.0
+        uses: golangci/golangci-lint-action@58eda26a511c265ee35b3ee4b101fb8adfd76480 # version v6.1.1
         with:
-          version: v1.52.2 # this is the golangci-lint version
-          args: --issues-exit-code=0 # exit without errors for now - won't fail the build
+          version: v1.64.8 # this is the golangci-lint version
+          args: -v --issues-exit-code=0 --config=.golangci.yml # exit without errors for now - won't fail the build
           github-token: ${{ secrets.GITHUB_TOKEN }}
           only-new-issues: true
+        continue-on-error: true # continue on error, shows as warning
```
sreuland, line 29: golangci runs and outputs some errors on lints and these errors are already skipped with `--issues-exit-code=0` usage. This maintains the 'don't stop the build' by squashing  other exit errors not related to linting, for some reason it's returning exit code 7, but it still runs linting. gha installs setup-go with 1.24, checking versions compat, golangci-lint 1.64.2+ was recommended to work with go 1.24. not sure what is triggering the 7 code.

ingest/ledgerbackend/rpc_backend.go
```diff
@@ -9,6 +9,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/creachadair/jrpc2"
```
urvisavla: Do we need to update the depguard rules to fix the linter warning below?

ingest/ledgerbackend/rpc_backend.go
```diff
@@ -250,17 +248,21 @@ func (b *RPCLedgerBackend) getBufferedLedger(ctx context.Context, sequence uint3
 
 	ledgers, err := b.client.GetLedgers(ctx, req)
 	if err != nil {
+		// InvalidRequest code is the most specific error code provided for invalid range requests.
+		// https://github.com/stellar/stellar-rpc/pull/407/
+		// if received, assume it's range problem to enable retry.
+		var rpcErr *jrpc2.Error
+		if errors.As(err, &rpcErr) && rpcErr.Code == jrpc2.InvalidRequest {
+			return xdr.LedgerCloseMeta{}, &RPCLedgerBeyondLatestError{}
```
urvisavla: I think it could be helpful to include the ledger sequence number in the error and possibly even the ledger range available on rpc or  just include rpcerror.Message in the error returned here?

ingest/ledgerbackend/rpc_backend.go
sreuland: not with the current ci config which has been in place. The linter config on this ci step looks at only the delta of code related to a pr by using the `--new-from-patch`, so this warn won't appear after this pr merges. There are tons of depguard and other warnings generated by the linter on this repo if you run the linter w/o the delta limitation.  In relation to this aspect also is the ci config for linter is set to mute any linter errors with the `--issues-exit-code=0`, so, linting as far as ci is concerned is for info purposes only.

ingest/ledgerbackend/rpc_backend.go
sreuland: that is good call out, the error condition of requesting a ledger that is beyond rpc's latest is actually just internal concern which RPCLedgerBackend detects and is not an error propagated publicly, I cleaned this up further as I was able to use GetHealth to identify the beyond latest condition, and was able to then simplify the usage of rpc's GetLedgers to just relay any error condition as it means can't resolve the requested range.

PR #5721, opened 2025-06-12
Karthik/ttp for p23 events

karthikiyer56:


processors/token_transfer/contract_events.go
```diff
@@ -78,7 +88,12 @@ func (p *EventsProcessor) parseEvent(tx ingest.LedgerTransaction, opIndex *uint3
 					// If contract ID matches, update with validated asset
 					protoEvent.SetAsset(asset)
 
-					// This is tricky. Burn and mint events currently show up as transfer in SAC events
+					// For TxMetaV4, this is all that needs to be validated. You can simply return the event as is
+					if tx.UnsafeMeta.V == 4 {
```
Shaptic: might as well reuse it
```suggestion
					if txMetaVersion == 4 {
```

processors/token_transfer/contract_events.go
```diff
@@ -191,3 +235,162 @@ func parseCustomTokenEvent(
 
 	return event, nil
 }
+
+// parseCustomTokenEventV4 attempts to parse a generic SEP41 token event for V4 format
+func parseCustomTokenEventV4(
```
Shaptic, line 322: This whole function looks a **lot** like the new internal package... is it worth exposing it for reuse to eliminate this duplication?

processors/token_transfer/contract_events.go
```diff
@@ -191,3 +235,162 @@ func parseCustomTokenEvent(
...
+	// V4 data format can be:
+	// 1. Direct i128 (when there's no memo)
+	// 2. ScMap with exactly 2 fields: "amount" (i128) + "to_muxed_id" (u64/bytes/string)
+	// If it's a map, at the very least "amount" should be present.
+	if mapData, ok := value.GetMap(); ok {
+		if mapData == nil {
+			return nil, errNotSep41TokenFromMsg("map is empty")
+		}
+		var err error
+		amt, destinationMemo, err = parseV4MapDataForTokenEvents(*mapData)
```
Shaptic, line 341: is there any need to check `destinationMemo == nil` later in the "amount in map only" case?

processors/token_transfer/contract_events.go
karthikiyer56: nope, that the divergence between the horizon internal package vs this TTP code parsing logic.
horizon internal package errors strictly checks for SAC, and if it deduces that its not an SAC, it errors out 

this parsing logic does the opposite. it sees for a sep-41 token and if it passes that, then there is an event to be had, even if it fails SAC validation.

There is no way to unify the logic in horizon internal and this code

processors/token_transfer/contract_events.go
karthikiyer56: nope.
it is possible for it to be a SEP-41 token (and not a SAC token), have the data as the map and just have amount (and no muxed_info) in the map.
That is still a valid SEP-41 token based on our slack thread discussion from earlier this week

Ergo, no need to check for destinationMemo. not here nor in `parseV4MapDataForTokenEvents`

Shaptic: right right, I'm just talking about a code perspective to ensure that `desinationMemo == nil` (which I agree is valid per our discussion) is handled elsewhere - sounds like yes :+1:

processors/token_transfer/contract_events.go
Shaptic: Makes sense, thanks for the rundown :+1:

ingest/ledger_transaction.go
```diff
@@ -625,7 +625,17 @@ func (t *LedgerTransaction) SorobanResourceFeeRefund() int64 {
 	if !t.IsSorobanTx() {
 		return 0
 	}
-	startingBal, endingBal := getAccountBalanceFromLedgerEntryChanges(t.UnsafeMeta.MustV3().TxChangesAfter, t.FeeAccount().ToAccountId().Address())
+	var txChangesAfter xdr.LedgerEntryChanges
+	switch t.UnsafeMeta.V {
+	case 3:
+		txChangesAfter = t.UnsafeMeta.MustV3().TxChangesAfter
+	case 4:
+		txChangesAfter = t.UnsafeMeta.MustV4().TxChangesAfter
```
Shaptic, line 633: This would go well (incl. the `PostTxApplyFeeChanges` bit) to be refactored into a `GetRefundChanges` (or whatever) helper at the `xdr/ledger_close_meta` level, imo - we should try to minimize `switch`/`case` on `V`s in higher-level code.

processors/token_transfer/contract_events.go
```diff
@@ -24,6 +39,65 @@ func errNotSep41TokenFromError(err error) ErrNotSep41TokenEvent {
...
+	}
+
+	var feeEvents []*TokenTransferEvent
+	for _, ev := range txEvents.TransactionEvents {
+		contractEvent := ev.Event
+		// Validate basic contract contractEvent structure
+		if contractEvent.Type != xdr.ContractEventTypeContract ||
+			contractEvent.ContractId == nil ||
+			contractEvent.Body.V != 0 {
+			return nil, errInvalidFeeEvent(fmt.Sprintf("Invalid feeEvent format"))
```
Shaptic, line 56: Is this an error or should it just be ignored? It'd be annoying to update this code later if Core decides to add other events to this layer.

processors/token_transfer/contract_events.go
```diff
@@ -105,11 +193,41 @@ func (p *EventsProcessor) parseEvent(tx ingest.LedgerTransaction, opIndex *uint3
...
+		switch eventType {
+		case TransferEvent:
+			return 4 // ["transfer", from, to, asset]
+		case MintEvent:
+			return 4 // ["mint", admin, to, asset]
+		case ClawbackEvent:
+			return 4 // ["clawback", admin, from, asset]
+		case BurnEvent:
+			return 3 // ["burn", from, asset]
+		}
```
Shaptic: I understand the verbosity but would you be opposed to
```suggestion
		// V3 format includes admin addresses
		switch eventType {
		case BurnEvent:
			return 3 // ["burn", from, asset]
		default:
			// ["transfer", from, to, asset]
			// ["mint", admin, to, asset]
			// ["clawback", admin, from, asset]
			return 4
		}
```
?

processors/token_transfer/contract_events.go
```diff
@@ -191,3 +309,162 @@ func parseCustomTokenEvent(
...
+			return nil, errNotSep41TokenFromMsg(fmt.Sprintf("burn event requires minimum 2 topics, found: %v", lenTopics))
+		}
+		from, err := extractAddress(topics[1])
+		if err != nil {
+			return nil, errNotSep41TokenFromError(fmt.Errorf("invalid fromAddress error: %w", err))
+		}
+		event = NewBurnEvent(meta, from, amtRaw128, nil)
+
+	default:
+		return nil, errNotSep41TokenFromMsg(fmt.Sprintf("unsupported custom token event type: %v", eventType))
```
Shaptic, line 416: Is this an error, per se, or should you just skip trying to convert it? imo `return nil, nil` is fine here. Otherwise you'll crash on any non-SEP-41 events a token emits.

Shaptic: Actually I guess it depends how you handle `err` at a higher level - I guess just make sure you allow other events through without stopping TTP.

processors/token_transfer/token_transfer_processor.go
```diff
@@ -184,6 +282,9 @@ func (p *EventsProcessor) EventsFromTransaction(tx ingest.LedgerTransaction) (Tr
 func (p *EventsProcessor) EventsFromOperation(tx ingest.LedgerTransaction, opIndex uint32, op xdr.Operation, opResult xdr.OperationResult) ([]*TokenTransferEvent, error) {
 	var events []*TokenTransferEvent
 	var err error
+
+	// If
```
Shaptic: nit: leftover

processors/token_transfer/contract_events.go
karthikiyer56: > Actually I guess it depends how you handle err at a higher level 

This is what I am doing.
I mark it as error for the purpose of the function `parseCustomTokenEventV4 or V3`
And then at the call site in `parseEvent`, I omit emitting an event if its an error

processors/token_transfer/contract_events.go
karthikiyer56: I mean this is a valid `if` check for the core generated `xdr.TransactionEvent`, no?
You shudnt have an empty contractId and the eventType shud be strictly `ContractEventTypeContract`.
And if not, there is a parsing error?

What might be worthwhile instead is to relax this requirement
```

		topics := contractEvent.Body.V0.Topics
		value := contractEvent.Body.V0.Data
		if len(topics) != 2 {
			return nil, errInvalidFeeEvent(fmt.Sprintf("invalid topic length for fee event for txHash: %v, topicLength: %v", txHash, len(topics)))
		}

		// Extract the contractEvent function name
		fn, ok := topics[0].GetSym()
		if !ok {
			return nil, errInvalidFeeEvent("invalid function name")
		}
		if string(fn) != FeeEvent {
			return nil, errInvalidFeeEvent(fmt.Sprintf("invalid function name: %v", string(fn)))
		}
		
```

to something likes this instead

```
		topics := contractEvent.Body.V0.Topics
		value := contractEvent.Body.V0.Data

		// Extract the contractEvent function name
		fn, ok := topics[0].GetSym()
		if !ok {
			continue // this is to account for future proofing where xdr.TransactionEvents might be extended to include more than just fees.
		}
		if string(fn) != FeeEvent {
			continue
		}

		// Now that we have established it is a Fee event, it will need to undergo stricter checks
		if len(topics) != 2 {
			return nil, errInvalidFeeEvent(fmt.Sprintf("invalid topic length for fee event for txHash: %v, topicLength: %v", txHash, len(topics)))
		}
```

what say?

processors/token_transfer/contract_events.go
karthikiyer56: not at all.
I'll make the change

processors/token_transfer/contract_events.go
karthikiyer56: yup

ingest/ledger_transaction.go
karthikiyer56: I had similar thoughts and decided against it.

Some nuances here preclude it from being a receiver function on `xdr/ledger_close_meta`

I dont think `xdr/ledger_close_meta.go` is the place to add code to get feeRefund informatin for **a particular transaction**
realistically it shud be in `xdr/transaction_meta.go`, but the `postTxApplyFeeChanges` is populated from Lcm/TransactionResultV1 when the `ingest.LedgerTransaction` struct is created in `ledger_transaction_reader.go`
And so it cant be at `xdr/transaction_meta.go`

So, the only sane place is for it to be a receiver on `ingest.LedgerTransaction`

processors/token_transfer/contract_events.go
karthikiyer56: done

processors/token_transfer/contract_events.go
karthikiyer56: I have refactored it like i descibed ⬆️

processors/token_transfer/contract_events.go
karthikiyer56: done

processors/token_transfer/token_transfer_processor.go
karthikiyer56: removed

processors/token_transfer/contract_events.go
Shaptic: Yep that makes total sense, basically return error as soon as it doesn't fit your expected format and skip at the highest iteration level :+1:

processors/token_transfer/contract_events.go
Shaptic: Yeah I think that change makes sense: if more events get included, TTP will still find fee events because it isn't being as strict with the quantity but as strict on the format.

processors/token_transfer/token_transfer_processor.go
```diff
@@ -98,35 +117,97 @@ func (p *EventsProcessor) EventsFromLedger(lcm xdr.LedgerCloseMeta) ([]*TokenTra
...
-	return allEvents, nil
+
+	// Assemble final event order based on protocol
+	return p.assembleEventOrder(feeEvents, operationEvents, feeRefundEvents, isProtocol23Plus), nil
+}
+
+// assembleEventOrder creates the final ordered list of events based on protocol version
+func (p *EventsProcessor) assembleEventOrder(feeEvents, operationEvents, feeRefundEvents []*TokenTransferEvent, isProtocol23Plus bool) []*TokenTransferEvent {
+	allEvents := make([]*TokenTransferEvent, 0, len(feeEvents)+len(operationEvents)+len(feeRefundEvents))
+	if isProtocol23Plus {
```
tamirms, line 192: I wonder if the isProtocol23Plus check is necessary because it seems that feeRefundEvents will always be empty if the ledger version is less than 23. If that is the case, I think you can unconditionally set the order to:

```go
		allEvents = append(allEvents, feeEvents...)
		allEvents = append(allEvents, operationEvents...)
		allEvents = append(allEvents, feeRefundEvents...)
```

xdr/scval.go
```diff
@@ -35,7 +35,8 @@ func (address ScAddress) String() (string, error) {
 		poolID := address.MustLiquidityPoolId()
 		result, err = strkey.Encode(strkey.VersionByteLiquidityPool, poolID[:])
 	case ScAddressTypeScAddressTypeClaimableBalance:
-		result, err = strkey.Encode(strkey.VersionByteClaimableBalance, address.MustClaimableBalanceId().V0[:])
```
tamirms, line 38: thanks for fixing this bug. it looks like we also need to fix it in support/contractevents/generate.go , specifically the `makeAddress()` function

processors/token_transfer/token_transfer_processor.go
```diff
@@ -220,9 +319,7 @@ func (p *EventsProcessor) EventsFromOperation(tx ingest.LedgerTransaction, opInd
 	case xdr.OperationTypeInflation:
 		events, err = p.inflationEvents(tx, opIndex, op, opResult)
 	case xdr.OperationTypeInvokeHostFunction:
-		if !p.disableContractEvents {
```
tamirms, line 223: it looks like with this change disableContractEvents is no longer referenced anywhere in TTP. I think this flag was only introduced so we could test whether rpc could handle the additional ingestion load of classic events. Now that we have a core build with classic events, I think we can get rid of this flag entirely

processors/token_transfer/token_transfer_processor.go
```diff
@@ -42,12 +42,30 @@ var (
 type EventsProcessor struct {
 	networkPassphrase     string
 	disableContractEvents bool
+
+	// This flag is set to false by default.
+	// Instantiate TTP with this flag only if you know that your ledgerCloseMeta has unified events in TxMeta
+	// i.e, the ledgers were generated by stellar-core when the EMIT_CLASSIC_EVENTS flag was set to true
```
tamirms: the `BACKFILL_STELLAR_ASSET_EVENTS` flag also needs to be set to true if we're ingesting ledgers older than protocol 23

ingest/ledger_transaction.go
Shaptic: > I dont think xdr/ledger_close_meta.go is the place to add code to get feeRefund informatin for a particular transaction

Why not? We [already do](https://github.com/stellar/go/blob/protocol-23/xdr/ledger_close_meta.go#L124-L150) have similar helpers that just take an index.

Shaptic (approved):

ingest/ledger_transaction.go
karthikiyer56: I hear you and I stand corrected.
I still have some implemendation details that will need to be sorted out for this.
As discussed, can we punt that to after this PR goes in?
i will rework that part in a separate PR

processors/token_transfer/token_transfer_processor.go
karthikiyer56: > because it seems that feeRefundEvents will always be empty if the ledger version is less than 23.

This is not true.
There are fee refund events out there for existing smart contract transactions and we need to account for those if someone decides to do a backfill and then run TTP with `WithUnifiedEventsStreamEnabled()` enabled.

If we dont have this check, then it will not calculate feeRefund correctly.
TL;Dr - this check is deliberate and required.

tamirms: could you describe the code path where `feeRefundEvents` is non-empty and the ledger version is less than 23?

The only place where I could find `feeRefundEvents` being populated is in:

 https://github.com/stellar/go/blob/6a2fce4d6a1de9df800397b8d1d4331c1c406490/processors/token_transfer/token_transfer_processor.go#L182-L185

PR #5722, opened 2025-06-13
services/horizon: protocol 23 state archival/eviction integration tests

urvisavla:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What
Integration test for testing state archival eviction changes in protocol 23.

### Why
#5613 

### Known limitations
N/A


services/horizon/internal/test/integration/integration.go
```diff
@@ -940,6 +943,68 @@ func (i *Test) WaitUntilLedgerEntryTTL(ledgerKey xdr.LedgerKey) {
 	assert.True(i.t, ttled)
 }
 
+func (i *Test) WaitUntilLedgerEntryIsEvicted(ledgerKey xdr.LedgerKey) {
+	lk, err := ledgerKey.MarshalBinary()
+	assert.NoError(i.t, err)
+
+	evicted := false
+	for attempt := 0; attempt < 50; attempt++ {
```
sreuland: for consideration could use `assert.Eventually` and eliminate the loop boilerplate.

services/horizon/internal/test/integration/integration.go
```diff
@@ -940,6 +943,68 @@ func (i *Test) WaitUntilLedgerEntryTTL(ledgerKey xdr.LedgerKey) {
...
+		}
+		i.t.Log("Waiting for ledger entry to be evicted")
+		time.Sleep(time.Second)
+	}
+	assert.True(i.t, evicted)
+}
+
+func (i *Test) getLatestLedgerSequenceRPC() uint32 {
+	ch := jhttp.NewChannel("http://localhost:"+strconv.Itoa(StellarRPCPort), nil)
+	client := jrpc2.NewClient(ch, nil)
```
sreuland: since the [rpc go client](https://github.com/stellar/stellar-rpc/blob/main/client/main.go) has been added recently, is it worthwhile to use that instead on these new tests? Not asking to refactor other existing usages.

services/horizon/internal/test/integration/integration.go
sreuland: also for thought, externalize the timeout as a parameter to make more visible and stipulated by the caller?

services/horizon/internal/test/integration/integration.go
```diff
@@ -940,6 +943,68 @@ func (i *Test) WaitUntilLedgerEntryTTL(ledgerKey xdr.LedgerKey) {
...
+	err := client.CallResult(context.Background(), "getLedgers", protocol.GetLedgersRequest{
+		StartLedger: sequence,
+		Pagination: &protocol.LedgerPaginationOptions{
+			Limit: 1,
+		},
+	}, &response)
+	assert.NoError(i.t, err)
+
+	var lcm xdr.LedgerCloseMeta
+	err = xdr.SafeUnmarshalBase64(response.Ledgers[0].LedgerMetadata, &lcm)
```
sreuland: should `response.Ledgers` be asserted for length of 1 first?

urvisavla: done

services/horizon/internal/test/integration/integration.go
urvisavla: done

services/horizon/internal/test/integration/integration.go
urvisavla: good point. done.

services/horizon/internal/test/integration/integration.go
urvisavla: good suggestion. done.

services/horizon/internal/test/integration/integration.go
```diff
@@ -940,6 +944,53 @@ func (i *Test) WaitUntilLedgerEntryTTL(ledgerKey xdr.LedgerKey) {
...
+		assert.NoError(i.t, err)
+
+		for _, key := range keys {
+			evictedKey, err := key.MarshalBinary()
+			assert.NoError(i.t, err)
+
+			if bytes.Equal(lk, evictedKey) {
+				i.t.Log("Ledger entry found in evicted ledger keys in ledger", sequence)
+				return true
+			}
```
tamirms: there is an `Equals()` function defined on xdr.LedgerKey which can be used here instead of marshaling the xdr value to binary

services/horizon/internal/test/integration/integration.go
```diff
@@ -940,6 +944,53 @@ func (i *Test) WaitUntilLedgerEntryTTL(ledgerKey xdr.LedgerKey) {
 	assert.True(i.t, ttled)
 }
 
+func (i *Test) WaitUntilLedgerEntryIsEvicted(ledgerKey xdr.LedgerKey, waitTime time.Duration) {
+	lk, err := ledgerKey.MarshalBinary()
+	assert.NoError(i.t, err)
+	sequence := i.getLatestLedgerSequenceRPC()
+
+	assert.Eventually(i.t, func() bool {
+		lcm := i.getLedgerRPC(sequence)
```
tamirms, line 948: what happens if rpc has not yet ingested the ledger, will this call fail / crash?

tamirms: @sreuland implemented a ledgerbackend using rpc that should handle the case where the ledger is not yet ingested. If you would like to use that ledger backend the code will look like:


```go
func (i *Test) WaitUntilLedgerEntryEvicted(ledgerKey xdr.LedgerKey, activeLedgerSequence uint32) {
	backend := ledgerbackend.NewRPCLedgerBackend(ledgerbackend.RPCLedgerBackendOptions{
		RPCServerURL: "http://localhost:" + strconv.Itoa(StellarRPCPort),
	})

	ctx := context.Background()
	err := backend.PrepareRange(ctx, ledgerbackend.UnboundedRange(activeLedgerSequence))
	if err != nil {
		i.t.Fatalf("failed to prepare range: %v", err)
	}

	evicted := false
	for ledgerSeq := activeLedgerSequence; !evicted && ledgerSeq < activeLedgerSequence+1000; ledgerSeq++ {
		ledger, err := backend.GetLedger(ctx, ledgerSeq)
		if err != nil {
			require.NoError(i.t, err)
		}
		keys, err := ledger.EvictedLedgerKeys()
		require.NoError(i.t, err)
		for _, key := range keys {
			if key.Equals(ledgerKey) {
				evicted = true
				for attempt := 0; attempt < 10; attempt++ {
					root, err := i.horizonClient.Root()
					require.NoError(i.t, err)
					if uint32(root.HorizonSequence) >= ledgerSeq {
						break
					}
					time.Sleep(time.Second)
				}
			}
		}
	}

	require.NoError(i.t, backend.Close())
	require.True(i.t, evicted)
}
```

services/horizon/internal/integration/sac_test.go
```diff
@@ -494,15 +465,140 @@ func TestExpirationAndRestoration(t *testing.T) {
...
+		Asset: txnbuild.CreditAsset{
+			Code:   code,
+			Issuer: issuer,
+		},
+		SourceAccount: itest.Master().Address(),
+	})
+	assert.NoError(t, err)
+	// set the contract id to storeContractID because we are restoring a fake asset balance
+	restoreFootprint.Ext.SorobanData.Resources.Footprint.ReadWrite[0].ContractData.Contract.ContractId = &storeContractID
+	itest.MustSubmitOperations(itest.MasterAccount(), itest.Master(), &restoreFootprint)
```
tamirms: there is a restoration helper function which I think would be easier to use here:

```golang
	// restore evicted balance
	sourceAccount, restoreOp := itest.RestoreFootprint(itest.Master().Address(), balanceToExpireLedgerKey)
	itest.MustSubmitOperations(&sourceAccount, itest.Master(), &restoreOp)
```

services/horizon/internal/integration/sac_test.go
```diff
@@ -494,15 +465,140 @@ func TestExpirationAndRestoration(t *testing.T) {
...
+		code:             code,
+		issuer:           issuer,
+		numAccounts:      0,
+		balanceAccounts:  0,
+		numContracts:     1,
+		balanceContracts: big.NewInt(37),
+		contractID:       storeContractID,
+	})
+
+	balanceToExpireLedgerKey := xdr.LedgerKey{
```
tamirms: nit: balanceToExpireLedgerKey -> balanceToEvictLedgerKey

urvisavla: updated.

services/horizon/internal/integration/sac_test.go
urvisavla: done

services/horizon/internal/test/integration/integration.go
urvisavla: done.

services/horizon/internal/test/integration/integration.go
urvisavla: Updated the code to use rpcbackend since it has builtin retries. also added a loop to wait for horizon to sync up to the ledger where the eviction occurred. However, choosing to create a new instance of the backend to fetch each ledger as it keeps the code cleaner and since this is for a single test so efficiency is not a concern. Also,  keeping the wait time instead of using ledger sequence because I think it's better for the test to be time bound.

services/horizon/internal/test/integration/integration.go
```diff
@@ -940,6 +942,51 @@ func (i *Test) WaitUntilLedgerEntryTTL(ledgerKey xdr.LedgerKey) {
...
+		sequence += 1
+		keys, err := lcm.EvictedLedgerKeys()
+		require.NoError(i.t, err)
+		for _, key := range keys {
+			if key.Equals(ledgerKey) {
+				i.t.Log("Ledger entry found in evicted ledger keys in ledger", sequence)
+
+				// wait for horizon to catch up
+				for attempt := 0; attempt < 10; attempt++ {
```
sreuland: could use `require.eventually` with same waitTime limit on this inner loop just for consistency on how test manages waiting?

sreuland (approved):
lgtm, nice work in capturing the eviction as new test.

tamirms (approved):

services/horizon/internal/test/integration/integration.go
urvisavla: good suggestion. done.

PR #5723, opened 2025-06-13
Add opt-in for unified events when Core supports >= Protocol 23

Shaptic:
### What
Ensures that `EMIT_CLASSIC_EVENTS` and `BACKFILL_STELLAR_ASSET_EVENTS` are enabled when you opt in to unified events via `CaptiveCoreTomlParams`.

### Why
We need to use this in RPC/Galexie.

### Known limitations
n/a


ingest/ledgerbackend/toml.go
```diff
@@ -578,6 +581,11 @@ func (c *CaptiveCoreToml) setDefaults(params CaptiveCoreTomlParams) {
 	if params.EnforceSorobanTransactionMetaExtV1 {
 		enforceOption(&c.EnableEmitSorobanTransactionMetaExtV1)
 	}
+	if params.EmitUnifiedEvents &&
+		c.checkCoreVersion(params).greaterThanOrEqual(minVersionForUnifiedEvents) {
```
urvisavla: You might want to check the core protocol version instead of the build version to make it more robust. I'm trying to do the same for the backfill_restore_meta flag here https://github.com/stellar/go/pull/5730.

Shaptic: Discussed offline - can wait on this :+1:

urvisavla (approved):

Issue #5553, opened 2024-12-11
Enable ingestion via CDP in horizon verify-range docker image


tamirms:
There is a [verify-range](https://github.com/stellar/go/tree/master/services/horizon/docker/verify-range) docker image which is used to detect ingestion regressions in horizon release candidates. It works by ingesting a range of ledgers and then [dumping](https://github.com/stellar/go/blob/master/services/horizon/docker/verify-range/start#L47-L73) the contents of the horizon history tables to disk. Then it repeats the same workflow but this time using the horizon release candidate. Finally, the output from both runs are [compared](https://github.com/stellar/go/blob/master/services/horizon/docker/verify-range/start#L161-L169). The job succeeds if the output from both runs are identical.

The verify-range job is effective in detecting regressions but it takes very long to run on full history. Also, the verify-range job often fails on tasks because it consumes too much memory. Both of these issues can be solved if we update the verify-range docker image to ingest from a GCS datastore of ledgers instead of using captive core.

mollykarcher:
Alternative that was discussed at team meeting on 3/19 was to retire verify-range in it's current form (ex. running in separate AWS account via AWS batch), but to instead recreate the same verification behavior using Stellar ETL, as is being done to verify TTP.

PR #5732, opened 2025-06-18
services/horizon/cmd:  add datastore config option to ingest verify-range cmd

sreuland:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

added `--ledgerbackend` and `--datastore-config` options to `ingest verify-range` command

working example invocation:

```
./horizon ingest verify-range --from 57609343 --to 57609353 \
--ledgerbackend datastore \
--datastore-config <my gcs datastore config file for pubnet> \
--db-url <my empty horizon db url> \
--network-passphrase "Public Global Stellar Network ; September 2015"\
 --history-archive-urls https://history.stellar.org/prd/core-live/core_live_001
```

### Why

Enable the verify ledger range procedure to optionally use pre-computed ledger metadata from a datastore instead of captive core for performance.

This is the first of two overall changes needed to accomplish #5553 , next up after this is to change some verify range bench deployments to use these new CDP arguments instead of captive core settings.

### DIFF
--- a/services/horizon/cmd/ingest.go
+++ b/services/horizon/cmd/ingest.go
@@ -14,20 +14,20 @@ import (
 	horizon "github.com/stellar/go/services/horizon/internal"
 	"github.com/stellar/go/services/horizon/internal/db2/history"
 	"github.com/stellar/go/services/horizon/internal/ingest"
+	"github.com/stellar/go/support/config"
 	support "github.com/stellar/go/support/config"
 	"github.com/stellar/go/support/db"
 	"github.com/stellar/go/support/log"
 )
 
-var ingestCmd = &cobra.Command{
-	Use:   "ingest",
-	Short: "ingestion related commands",
-}
-
 var ingestBuildStateSequence uint32
 var ingestBuildStateSkipChecks bool
 var ingestVerifyFrom, ingestVerifyTo, ingestVerifyDebugServerPort uint32
 var ingestVerifyState bool
+var ingestVerifyLedgerBackendStr string
+var ingestVerifyStorageBackendConfigPath string
+var ingestVerifyLedgerBackendType ingest.LedgerBackendType
+var processVerifyRangeFn = processVerifyRange
 
 var ingestBuildStateCmdOpts = []*support.ConfigOption{
 	{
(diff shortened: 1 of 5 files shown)


services/horizon/cmd/ingest.go
```diff
@@ -81,79 +81,13 @@ var ingestVerifyRangeCmdOpts = []*support.ConfigOption{
...
-		)
-		if err != nil {
-			return err
-		}
-
-		log.Info("Range run successfully!")
-		return nil
+	generateLedgerBackendOpt(&ingestVerifyLedgerBackendStr, &ingestVerifyLedgerBackendType),
+	{
+		Name:      "datastore-config",
```
Shaptic, line 86: Can the presence of this imply `--ledgerbackend datastore`? Less parameters that way

services/horizon/cmd/ingest.go
```diff
@@ -178,156 +112,220 @@ var stressTestCmdOpts = []*support.ConfigOption{
...
 }
 
-var ingestStressTestCmd = &cobra.Command{
-	Use:   "stress-test",
-	Short: "runs ingestion pipeline on a ledger with many changes. warning! requires clean DB.",
-	Long:  "runs ingestion pipeline on a ledger with many changes. warning! requires clean DB.",
-	RunE: func(cmd *cobra.Command, args []string) error {
-		for _, co := range stressTestCmdOpts {
-			if err := co.RequireE(); err != nil {
+func DefineIngestCommands(rootCmd *cobra.Command, horizonConfig *horizon.Config, horizonFlags config.ConfigOptions) {
```
sreuland, line 115: all of the red is related to refactoring the sub commands to be scoped as closures rather than package singleton and then added to root cmd, allows re-create root cmd repeatedly and binding to same sub commands(different instances each time) in same process, iniitally for test's sake.

services/horizon/cmd/ingest.go
sreuland: I think it's a valid suggestion. These same flags are used together on the `db reingest` command and was following convention for consistency. I will create a ticket to open this up for design decision, whether to deprecate the `--ledgerbackend` in favor of presence of new(er) flags related to a particular ledger backend(rpc as another source..) driving the backend type and defaulting to cc if none present.

sreuland: https://github.com/stellar/go/issues/5734

Shaptic: makes sense, thanks!

PR #5733, opened 2025-06-20
services/horizon/internal/ingest/processors: Relax ttl invariant in AssetStatsProcessor

tamirms:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

In the AssetStatsProcessor we have an invariant which asserts that a TTL entry cannot be updated (without first being restored) if it was archived.

However, that invariant does not apply to temporary contract data ledger entries. A TTL ledger entry does not include any information about whether its corresponding contract data ledger entry is kept within persistent or temporary storage. So, we have no way of asserting the invariant only on TTLs for persistent contract data ledger entries.

Therefore, this PR removes the invariant assertion because, otherwise, horizon ingestion will halt in the case that a temporary contract data ledger entry is re-created before eviction. 

I also updated the code in ingest/sac/contract_data.go to explicitly check that contract balances and asset metadata are stored in persistent storage. This allows the asset stats processor to ignore changes corresponding to temporary ledger entries.

### Known limitations

[N/A]


ingest/sac/contract_data.go
```diff
@@ -79,6 +79,9 @@ func AssetFromContractData(ledgerEntry xdr.LedgerEntry, passphrase string) (xdr.
 	if contractData.Key.Type != xdr.ScValTypeScvLedgerKeyContractInstance {
 		return xdr.Asset{}, false
 	}
+	if contractData.Durability != xdr.ContractDataDurabilityPersistent {
```
sreuland, line 82: This check may not be necessary as it looks like [asset info is only stored as durable](https://github.com/stellar/rs-soroban-env/blob/v0.0.16/soroban-env-host/src/native_contract/token/asset_info.rs#L6)? If the entry is not asset info the function produces same false response below on ln 114.

tamirms: the asset info is stored as part of the contract instance and instance storage implies persistent durability:

https://developers.stellar.org/docs/build/guides/storage/use-instance

sreuland: Yes, which is why I was wondering why this additional check on durability is needed? since it's implied for asset info entries from sac's, which are the only entries this function interacts with.

tamirms: I agree the check is not necessary. Perhaps it could be replaced with a comment explaining that instance storage implies persistence and that is why we don't need to explicitly check for it like we do in  ContractBalanceFromContractData()

sreuland: ok, thanks for clarification.

PR #5736, opened 2025-06-24
services/horizon/docker: cdp enabled verify range image

### LINKED ISSUE #5553: Enable ingestion via CDP in horizon verify-range docker image

sreuland:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

* Updated the verify-range image to use optional GCS Datastore settings if present from environment to use instead of captive core for accessing ledgers. Refer to [README.md](https://github.com/stellar/go/pull/5736/files#diff-689188d34f82a03e4e61112951c05f1870b2373356c1ea4e9da5e1a608423f03R29)
* Refactored the verify-range container to run the old and new horizon processes in parallel if CDP is enabled for additional performance optimization. Does not perform parallelization if captive core, as would require more config changes for unique ports on the captive instances per new/old.
  *  some early performance numbers on these changes as seen from batch execution of latest network Ledger ranges:
     * 3.75 hours processing time elapsed for 57609343 to  57794662 which is 185319 ledgers, equating to an hourly rate of about 50k/hr is observed. 
     * for comparison, roughly 2 hours processing time elapses on average for 20k ledgers when running the original verify-range image in batch executions with smaller recent network ranges of about 20k ledgers, equating to a rate of 10k/hr. 
     * It looks like the combination of pre-computed datastore ledgers and parallelized 'new' and 'old' range processors is yielding roughly 5x boost in throughput.
* this leverages the updated verify-range command which supports datastore config from [part #1](https://github.com/stellar/go/pull/5732)

closes #5553 

### Why

can save time/complexity of runtime when using pre-computed meta from cdp rather than live captive core.

to build and run verify-range locally and use a datastore, checkout the go repo, from top folder:
```
docker build --platform linux/amd64 \
--build-arg="GO_VERSION=1.23.4" \
--build-arg="STELLAR_CORE_VERSION=22.3.0-2485.e643061a4.jammy" \
-f services/horizon/docker/verify-range/Dockerfile \
-t verify-range:latest services/horizon/docker/verify-range/
```
```
docker run -e FROM=63 -e TO=127 -e BRANCH=<tag, branch, pr ref, commit> \
-e BASE_BRANCH=master \
-v /path/to/gcp_credentials.json:/tmp/gcp.json \
-e GOOGLE_APPLICATION_CREDENTIALS=/tmp/gcp.json \
-v /path/to/pubnet_datastore_config.toml:/tmp/datastore-config.toml \
-e DATASTORE_CONFIG=/tmp/datastore-config.toml \
verify-range:latest
```

### DIFF
--- a/.github/workflows/horizon.yml
+++ b/.github/workflows/horizon.yml
@@ -124,8 +124,8 @@ jobs:
     runs-on: ubuntu-22.04
     env:
       GO_VERSION: 1.23.4
-      STELLAR_CORE_VERSION: 22.3.0-2485.e643061a4.focal
-      CAPTIVE_CORE_STORAGE_PATH: /tmp
+      STELLAR_CORE_VERSION: 22.3.0-2485.e643061a4.jammy
+      CAPTIVE_CORE_STORAGE_PATH: /data/working
     steps:
       - uses: actions/checkout@v3
         with:
@@ -134,7 +134,7 @@ jobs:
 
       - name: Build and test the Verify Range Docker image
         run: |
-          docker build --build-arg="GO_VERSION=$GO_VERSION" -f services/horizon/docker/verify-range/Dockerfile -t stellar/horizon-verify-range services/horizon/docker/verify-range/
+          docker build --build-arg="GO_VERSION=$GO_VERSION" --build-arg="STELLAR_CORE_VERSION=$STELLAR_CORE_VERSION" -f services/horizon/docker/verify-range/Dockerfile -t stellar/horizon-verify-range services/horizon/docker/verify-range/
           # Any range should do for basic testing, this range was chosen pretty early in history so that it only takes a few mins to run
           docker run -e BRANCH=$(git rev-parse HEAD) -e FROM=10000063 -e TO=10000127 stellar/horizon-verify-range
 
--- a/.github/workflows/horizon.yml
+++ b/.github/workflows/horizon.yml
@@ -125,7 +125,6 @@ jobs:
     env:
       GO_VERSION: 1.23.4
       STELLAR_CORE_VERSION: 22.3.0-2485.e643061a4.jammy
-      CAPTIVE_CORE_STORAGE_PATH: /data/working
     steps:
       - uses: actions/checkout@v3
         with:
--- a/services/horizon/docker/verify-range/README.md
+++ b/services/horizon/docker/verify-range/README.md
@@ -45,7 +45,7 @@ This image supports connecting to GCS buckets for ledger data instead of captive
       ```
 
 #### GCS Datastore settings
-- Purpose: Defines the GCS bucket name and ledger partioning used on the buckets.
+- Purpose: Defines the GCS bucket name and ledger partioning used on the buckets. These settings are referenced as a single toml file at runtime. Here is an example [datastore_config.toml](../../../galexie/config.example.toml)
 - Two options are available to provide this to container:
   - As an environment variable:
     - Pass the datastore TOML config as a string(including line breaks, tabs) in the `DATASTORE_CONFIG_PLAIN` environment variable:
--- a/services/horizon/docker/verify-range/README.md
+++ b/services/horizon/docker/verify-range/README.md
@@ -70,3 +70,25 @@ it will run the following ranges:
 | 1                           | 127    | 191  |
 | 2                           | 191    | 255  |
 | 3                           | 255    | 319  |
+
+Running the image locally with `docker run`
+* example to run verify range with captive, same as before:
+  ```
+  docker run -e FROM=63 \
+                      -e TO=127 \
+                      -e BRANCH=<target version> \
+                      -e BASE_BRANCH=master \
+                      verify-range:latest
+  ```  
+* example to run verify range with gcs datastore:
+  ```
+  docker run -e FROM=63 \
+                      -e TO=127 \
+                      -e BRANCH=<target version> \
+                      -e BASE_BRANCH=master \
+                      -v /host/path/to/gcreds.json:/tmp/gcp.json \
+                      -e GOOGLE_APPLICATION_CREDENTIALS=/tmp/gcp.json \
+                      -v /host/path/to/datastore-config.tom:/tmp/datastore-config.toml \
+                      -e DATASTORE_CONFIG=/tmp/datastore-config.toml \
+                      verify-range:latest
+  ```  
--- a/services/horizon/docker/verify-range/README.md
+++ b/services/horizon/docker/verify-range/README.md
@@ -28,33 +28,32 @@ to provide an external volume mount to the container for `/data` of at least 300
 
 ### Datastore and GCP Credentials Usage
 
-This image supports connecting to GCS buckets for ledger data instead of captive core. To use this feature, set three environment variables:
+This image supports connecting to GCS buckets for ledger data instead of captive core. To use this feature configure the container with these additional settings:
 
-#### 1. `GCP_CREDS`
-- **Purpose:** Provide GCP credentials to the container.
-- **Two ways to provide:**
-  - **As an environment variable:**
-    - Pass the JSON credentials as a string in the `GCP_CREDS` environment variable:
+#### GCP Credentials
+- Purpose: To access GCS buckets the image needs GCP credentials. 
+- Two options are available to provide this to container:
+  - As an environment variable:
+    - Pass the GCP JSON credentials as a string in a `GCP_CREDS` environment variable:
       ```sh
       docker run -e GCP_CREDS='{...}' ...
       ```
-  - **As a mounted file:**
-    - Mount a host file containing the credentials to the container, e.g.:
+  - As a volume mount:
+    - Mount the GCP json credentials file on host to the container, e.g.:
       ```sh
-      docker run -v /host/path/gcp.json:/tmp/gcp.json -e GOOGLE_APPLICATION_CREDENTIALS=/tmp/gcp.json ...
+      docker run -v /host/path/credentials.json:/tmp/credentials.json -e GOOGLE_APPLICATION_CREDENTIALS=/tmp/credentials.json ...
       ```
-    - The application will use the credentials from the mounted file.
 
-#### 2. `DATASTORE_CONFIG_PLAIN`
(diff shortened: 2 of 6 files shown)


.github/workflows/horizon.yml
```diff
@@ -124,8 +124,7 @@ jobs:
     runs-on: ubuntu-22.04
     env:
       GO_VERSION: 1.23.4
-      STELLAR_CORE_VERSION: 22.3.0-2485.e643061a4.focal
-      CAPTIVE_CORE_STORAGE_PATH: /tmp
```
karthikiyer56, line 128: why did you have to remove `CAPTIVE_CORE_STORAGE_PATH`?
Was this a unused flag that you have cleaned up here?

karthikiyer56: I think it is coz of this?
https://github.com/stellar/go/pull/5736/files#diff-689188d34f82a03e4e61112951c05f1870b2373356c1ea4e9da5e1a608423f03R5

services/horizon/docker/verify-range/README.md
```diff
@@ -20,7 +25,41 @@ This docker image allows running multiple instances of `horizon ingest verify-co
 | `BATCH_START_LEDGER` | First ledger of the AWS Batch Job, must be a checkpoint ledger or 1. |
 | `BATCH_SIZE`         | Size of the batch, must be multiple of 64.                           |
 
-#### Example
+
+### Datastore and GCP Credentials Usage
+
+This image supports connecting to GCS buckets for ledger data instead of captive core. To use this feature, set three environment variables:
```
karthikiyer56: A few comments about the README.md
It is a bit confusing, atleast to me.

1. The text here says `set 3 env vars`
and then you have `#### 1` and `#### 2`, but no `#### 3`
If I am reading this right, the 3 env variables are `GCP_CREDS`, `DATASTORE_CONFIG_PLAIN` and `DATASTORE_CONFIG` ?

2. And its an either/or for the last 2 env variables, right?
As in you either specify creds in `DATASTORE_CONFIG_PLAIN` OR specify file location via `DATASTORE_CONFIG`

Can we clarify this/reword it in the README pl

services/horizon/docker/verify-range/start
```diff
@@ -1,16 +1,125 @@
...
+CONTAINER_WORKING_DIR="/data/job_${AWS_BATCH_JOB_ARRAY_INDEX:-0}"
+echo "Job working directory path on data volume: $CONTAINER_WORKING_DIR"
+
+# Ensure CONTAINER_WORKING_DIR exists and is empty
+if [ -d "$CONTAINER_WORKING_DIR" ]; then
+    rm -rf "$CONTAINER_WORKING_DIR"
+fi	
+mkdir -p "$CONTAINER_WORKING_DIR"
+
+dump_horizon_db() {
```
karthikiyer56, line 15: is there a logical change/addition in this `dump_horizon_db` function from what it was earlier?
I am trying to eyeball it and it looks the same but there seem to be 3-4 lines added to this new one that I cant figure out by looking.

Can you help highlight the changes from the existing function, if any, in this `dump_horizon_db`?

There is a lot to check/review here and I want to skip the parts that are same

services/horizon/docker/verify-range/start
```diff
@@ -1,16 +1,125 @@
...
+	cat /tmp/datastore-config.toml
+fi	
+
 # configure postgres
 if [ -z "${PGDATA}" ]; then
-    export PGDATA="/data"
+    export PGDATA="$CONTAINER_WORKING_DIR/postgres"
+fi
+
+if [ ! -d "$PGDATA" ]; then
```
karthikiyer56: This  if check is not required.
it can simply be 
```
rm -rf $PGDATA
mkdir -p $PGDATA
```

tamirms: the same can be applied to `CONTAINER_WORKING_DIR` a few lines above

.github/workflows/horizon.yml
sreuland: yes, that is set dynamically within the image to a sub dir of the CONTAINER_WORKING_DIR which can be mounted to external volume of container.

services/horizon/docker/verify-range/README.md
sreuland: cleaned that up - https://github.com/stellar/go/pull/5736/commits/20fdcd6a6f91f2cc5a453b9d0c38c4ed3e5a1b1a, thanks!

services/horizon/docker/verify-range/README.md
```diff
@@ -20,7 +25,40 @@ This docker image allows running multiple instances of `horizon ingest verify-co
...
+- Two options are available to provide this to container:
+  - As an environment variable:
+    - Pass the datastore TOML config as a string(including line breaks, tabs) in the `DATASTORE_CONFIG_PLAIN` environment variable:
+      ```sh
+      docker run -e DATASTORE_CONFIG_PLAIN='[buffered_storage_backend_config]\nbuffer_size = 5\n ...' 
+      ```
+  - As a volume mount:
+    - Mount the datastore toml config file from host to the container, e.g.:
+      ```sh
+      docker run -v /host/path/datastore-config.toml:/tmp/datastore-config.toml -e DATASTORE_CONFIG=/tmp/datastore-config.toml
```
urvisavla, line 58: Should we provide an example of the data store config or maybe include a sample config file?

services/horizon/docker/verify-range/start
sreuland: nothing functional changed, I refactored it to use function params for db urls instead of hardcoding.

services/horizon/docker/verify-range/README.md
sreuland: yes, I was thinking the same and needed this nudge, [done](https://github.com/stellar/go/pull/5736/commits/3209828bc1902d33fe534e9cfee6cd52e310a862)

services/horizon/docker/verify-range/start
```diff
@@ -41,132 +150,90 @@ if [ ! -z "$AWS_BATCH_JOB_ARRAY_INDEX" ]; then
...
+rm -rf "$CONTAINER_WORKING_DIR/compare"
+mkdir -p "$CONTAINER_WORKING_DIR/compare"
+
+run_new_horizon() {
+    mkdir -p "$CONTAINER_WORKING_DIR"/new_runtime
+    cd "$CONTAINER_WORKING_DIR"/new_runtime
+    echo "Starting dump_horizon_db new_history with LEDGERBACKEND=$LEDGERBACKEND and DATASTORE_CONFIG=$DATASTORE_CONFIG"
+    "$CONTAINER_WORKING_DIR/horizon_new" --db-url "$DATABASE_URL_NEW" db migrate up 
+    alter_tables_unlogged "new"
+    "$CONTAINER_WORKING_DIR/horizon_new" --db-url "$DATABASE_URL_NEW" ingest verify-range --from $FROM --to $TO --verify-state
```
urvisavla, line 197: I know this is the existing behavior but do you know why --verify-state is required for running "new" horizon verify-range command and not for the "old"?

services/horizon/docker/verify-range/start
```diff
@@ -41,132 +150,90 @@ if [ ! -z "$AWS_BATCH_JOB_ARRAY_INDEX" ]; then
...
-	compare history_effects
-	compare history_ledgers
-	compare history_operations
-	compare history_operation_claimable_balances
-	compare history_operation_participants
-	compare history_trades
-	compare history_transactions
-	compare history_transaction_claimable_balances
-	compare history_transaction_participants
+if [ ! -z "$DATASTORE_CONFIG" ]; then
```
urvisavla, line 214: great optimization!

services/horizon/docker/verify-range/start
```diff
@@ -41,132 +150,90 @@ if [ ! -z "$AWS_BATCH_JOB_ARRAY_INDEX" ]; then
...
+    (run_new_horizon) &
+    (run_old_horizon) &
+    wait
+else
+    # don't attempt to run parallel with captive cores
+    run_new_horizon
+    run_old_horizon
 fi
 
+# Now run the compare functions
```
urvisavla, line 224: I know this is existing behavior but noticed that history_transaction_liquidity_pools and history_operation_liquidity_pools dumped but not compared.

services/horizon/docker/verify-range/start
sreuland: done - https://github.com/stellar/go/pull/5736/commits/1b03425e3f325cc17aab6c634accdd26a239ab17

services/horizon/docker/verify-range/start
sreuland: good question, I wondered along the same about why it's using `verify-range` for new branch but `reingest range` for base branch. It appears both run the same ingestion processes that populate the history tables and it would seem like we could use either command for both and then `--verify-state` in there as well. I didn't want to change this at this point to avoid unrelated potential regression, but we could investigate and tweak this as another pass later?

services/horizon/docker/verify-range/start
sreuland: yes, it was [intentional fix to avoid false negatives on compare due to non-deterministic db ids](https://github.com/stellar/go/pull/5147) , I'm not sure why we decided to export to file still though, that might be just some tech debt.

tamirms (approved):

PR #5739, opened 2025-06-29
Add support for muxed information from contractEvents to show up in o…

karthikiyer56:
…perations endpoint

<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

### Why

### Known limitations

[TODO or N/A]


services/horizon/internal/ingest/processors/operations_processor.go
```diff
@@ -820,19 +810,47 @@ func (operation *transactionOperationWrapper) parseAssetBalanceChangesFromContra
...
+		// Leave empty - no muxed fields will be added
+	default:
+		panic(fmt.Errorf("invalid memo type: %v", memo.Type))
+	}
+
+	if memoType != "" {
+		balanceChange["destination_muxed_id_type"] = memoType
+		balanceChange["destination_muxed_id"] = memoVal
 	}
+	fmt.Println("AM HEERE")
```
karthikiyer56: I had to add these prints here since this happens in the background and is not breakpointable when debugging integration test

karthikiyer56: removed now

services/horizon/internal/integration/sac_test.go
```diff
@@ -116,6 +116,67 @@ func TestContractMintToAccount(t *testing.T) {
 	})
 }
 
+func TestSacTransfertWithMuxedInfo(t *testing.T) {
+	if integration.GetCoreMaxSupportedProtocol() < 23 {
+		t.Skip("This test run does not support less than Protocol 23")
+	}
+
+	itest := integration.NewTest(t, integration.Config{
+		HorizonEnvironment: map[string]string{"INGEST_DISABLE_STATE_VERIFICATION": "true", "CONNECTION_TIMEOUT": "360000"},
```
tamirms: I don't think CONNECTION_TIMEOUT is necessary

services/horizon/internal/integration/sac_test.go
```diff
@@ -116,6 +116,67 @@ func TestContractMintToAccount(t *testing.T) {
...
+		EnableStellarRPC:   true,
+		QuickExpiration:    true,
```
tamirms: since this test does not require contract data to expire I think we can remove `QuickExpiration: true`

services/horizon/internal/ingest/contractevents/events.go
```diff
@@ -8,21 +8,22 @@ import (
...
 // EventType represents the type of Stellar asset Contract event
-type EventType int
+type EventType string
 
 const (
-	EventTypeTransfer EventType = iota
-	EventTypeMint
-	EventTypeClawback
-	EventTypeBurn
+	EventTypeInvalid  EventType = ""
```
sreuland: this might be a bit confusing to dedicate a type to 'Invalid', it is used primarily as zero value correct? Rather than dedicate a type value can it be removed and references to EventType handle this condition as an idiomatic zero value instead?

karthikiyer56: done.
removed `EventTypeInvalid`

services/horizon/internal/integration/sac_test.go
karthikiyer56: done

services/horizon/internal/integration/sac_test.go
karthikiyer56: done

services/horizon/internal/ingest/processors/operations_processor.go
```diff
@@ -820,19 +805,44 @@ func (operation *transactionOperationWrapper) parseAssetBalanceChangesFromContra
...
+	case xdr.MemoTypeMemoId:
+		memoType, memoVal = "uint64", fmt.Sprintf("%d", memo.MustId())
+	case xdr.MemoTypeMemoHash:
+		memoType, memoVal = "bytes", memo.MustHash().HexString()
+	case xdr.MemoTypeMemoReturn:
+		memoType, memoVal = "bytes", memo.MustRetHash().HexString()
+	case xdr.MemoTypeMemoNone:
+		// Leave empty - no muxed fields will be added
+	default:
+		panic(fmt.Errorf("invalid memo type: %v", memo.Type))
```
sreuland: can this error be propagated outwards instead of panic inline? Let upper layers of ingestion get this and determine how to panic?

services/horizon/internal/integration/sac_test.go
karthikiyer56: I had merely copied over these settings from the previous test and hadnt given them much thought.
That being said, removing connection timeout is fine, but when I remove the `QuickExpiration` setting, it triggers the logic in `func upgradeLimits()` to be run. (Setting `QuickExpiration=true` allows it to skip the upgradeLimits() function) 

And then horizon fails when trying to submit the Tx to the stellar core client in `upgradeLimits()`
```
response, err := i.Client().SubmitTransactionXDR(b64)
```
```
time="2025-06-30T14:46:46.929-07:00" level=warning msg="Cleared submission due to timeout" hash=df97719376d1ffa3822f86d792c48510f30e205cfe30d7db85384eb2f2a9eea3 listeners=1 pid=67328 service=txsub.submissionList
time="2025-06-30T14:46:46.929-07:00" level=info msg="Finished request" app_name=undefined app_version=undefined bytes=1246 client_name=go-stellar-sdk client_version=2.1.0 duration=30.999319916 host="localhost:8000" ip="[::1]" ip_port="[::1]:54145" method=POST path=/transactions pid=67328 referer=undefined req=karthik-LNQJHFVHJD-mbp/svZIkmIn37-000046 route=/transactions status=504 streaming=false x_forwarder_for=
    integration.go:780: 
        	Error Trace:	/Users/karthik/WS/stellar-go/services/horizon/internal/test/integration/integration.go:780
        	            				/Users/karthik/WS/stellar-go/services/horizon/internal/test/integration/integration.go:236
        	            				/Users/karthik/WS/stellar-go/services/horizon/internal/integration/sac_test.go:124
        	Error:      	Received unexpected error:
        	            	horizon error: "Transaction Submission Timeout" - check horizon.Error.Problem for more information
```

not sure why this is.

services/horizon/internal/ingest/processors/operations_processor_test.go
```diff
@@ -396,7 +396,7 @@ func (s *OperationsProcessorTestSuiteLedger) TestOperationTypeInvokeHostFunction
 				found++
 			}
 		}
-		s.Assert().Equal(found, 4, "should have one balance changed record for each of mint, burn, clawback, transfer")
+		s.Assert().Equal(4, found, "should have one balance changed record for each of mint, burn, clawback, transfer")
```
sreuland, line 474: A new test case with muxed info to assert `assetBalanceChanged["destination_muxed_id_type"]` and `assetBalanceChanged["destination_muxed_id"]`  are emitted may be good here also.

services/horizon/internal/ingest/processors/operations_processor.go
karthikiyer56: done

services/horizon/internal/integration/sac_test.go
```diff
@@ -116,6 +116,67 @@ func TestContractMintToAccount(t *testing.T) {
 	})
 }
 
+func TestSacTransfertWithMuxedInfo(t *testing.T) {
+	if integration.GetCoreMaxSupportedProtocol() < 23 {
+		t.Skip("This test run does not support less than Protocol 23")
+	}
+
+	itest := integration.NewTest(t, integration.Config{
+		HorizonEnvironment: map[string]string{"INGEST_DISABLE_STATE_VERIFICATION": "true"},
```
tamirms: I see what you mean when you mentioned that this configuration was copied over from TestContractMintToAccount. I think that you can omit this configuration entirely. Can you update this test and HorizonEnvironment to remove the HorizonEnvironment configuration?

services/horizon/internal/integration/sac_test.go
```diff
@@ -116,6 +116,67 @@ func TestContractMintToAccount(t *testing.T) {
 	})
 }
 
+func TestSacTransfertWithMuxedInfo(t *testing.T) {
```
tamirms: I suggest renaming the test to: TestContractTransferWithMuxedInfo

I think this is more consistent with the names of the other tests and it's not necessary to include "Sac" in the test name since that is implied by the fact it lives in sac_test.go

services/horizon/internal/ingest/processors/operations_processor_test.go
karthikiyer56: done

karthikiyer56: there is on more `destination_muxed_id_type`
`destination_muxed_id` can only ever be a uint64, and it will be reprensentef as a string in the restapi response

services/horizon/internal/ingest/contractevents/events_test.go
```diff
@@ -554,60 +553,3 @@ func mustGetContractID(asset xdr.Asset) *xdr.ContractId {
 	contractId := xdr.ContractId(id)
 	return &contractId
 }
-
-func makeV4MapData(amount *big.Int, memo xdr.Memo) xdr.ScVal {
```
karthikiyer56, line 558: just moved from here to `generate.go` and made it public, since this helper is now used outside this package

services/horizon/internal/integration/sac_test.go
karthikiyer56: done

protocols/horizon/operations/main.go
```diff
@@ -401,15 +401,14 @@ type RestoreFootprint struct {
...
-	From                   string `json:"from,omitempty"`
-	To                     string `json:"to,omitempty"`
-	Amount                 string `json:"amount"`
-	DestinationMuxedIdType string `json:"destination_muxed_id_type,omitempty"`
-	DestinationMuxedId     string `json:"destination_muxed_id,omitempty"`
+	Type               string `json:"type"`
+	From               string `json:"from,omitempty"`
+	To                 string `json:"to,omitempty"`
+	Amount             string `json:"amount"`
+	DestinationMuxedId string `json:"destination_muxed_id,omitempty"`
```
Shaptic, line 411: Should this be
```suggestion
	DestinationMuxedId uint64 `json:"destination_muxed_id,string,omitempty"`
```
or does that cause problems in the "memo set to zero" case due to the `omitempty`?

karthikiyer56: I actually didnt know that `DestinationMuxedId uint64` `json:.....string` was something you could do

Shaptic: Yeah it's a clever trick since it lets you work with the native value on the Go side, but in this case I think if you were to receive a memo with a value of zero, you'd end up omitting it. So you'd actually have to make it a pointer (`*uint64`) which makes the code way more unnecessarily complicated, so I think keeping it as a string internally (since there's no semantic value Go-side in having it be a uint64 [no math or comparisons being done, etc.]) is better here.

karthikiyer56: > or does that cause problems in the "memo set to zero" case due to the omitempty?
I think it will.
There is no way to distinguish b/w the `DestinationMuxedId uint64` not being set VS it being set to 0, like you said.
But then an argument can be made - what does it mean to have a 0 valued MuxedId to begin with? Best to ignore it in that case?

I do see it in other places too as uint64, string/omitEmpty. In the same file too, as a matter or fact
https://github.com/stellar/go/blob/08098be92cd0a361ea9399db00eb4eb0e68415b1/protocols/horizon/operations/main.go#L303

Ima make it to 
`DestinationMuxedId uint64 `json:"destination_muxed_id,string,omitempty"``

Shaptic: >  But then an argument can be made - what does it mean to have a 0 valued MuxedId to begin with? Best to ignore it in that case?

I disagree, e.g. you could use zero to refer to the "base account" or something to keep the same code path. The presence of a memo is the differentiator we care about here rather than the value itself.

Shaptic: That's an interesting observation re: existing code, though... I wonder if it's different in the case of `MuxedAccount` vs. `Memo` usage. Thoughts @tamirms?

karthikiyer56: the only recourse then is to make it a *string then

Shaptic: If it's a `string` on the Go side, it'll be `""` in the absence of a memo (and thus omitted), and be set to a `"%d"` in the ID case (and thus included), right? So you don't need a pointer. Basically I'm saying I think it's good as-is but wanted to bring up the discussion point for visibility & understanding.

karthikiyer56: Actually, it still makes sense to have it just as "DestinationMuxedId string `json:"destination_muxed_id,omitempty"`"

When saving it as a jsonblob in DB, I save it as a string.
https://github.com/stellar/go/pull/5739/files/08098be92cd0a361ea9399db00eb4eb0e68415b1#diff-5f37fbbec67583bbaa94cbe645d8a4aa046505cee2c37be660818ef340645ab1R820

So, even if the actualMemo set in the event is uint64(0), it will still be captured correctly/stored in DB as a string,  and show up in the restapi response.

Tl;dr - IMO, it makes sense to retain the existing logic and not change it

Shaptic: Yep, exactly :+1: agreed!

sreuland (approved):
looks good, thanks for responding to suggestions.

Issue #5662, opened 2025-04-22
Update events parsing code in horizon ingestion to support cap-0067


tamirms:
Horizon parses SAC events in the operations and effects processors. That code needs to be updated because the event schema will change once [cap-0067](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0067.md) is implemented.

[Asset Balance Changes](https://github.com/stellar/go/blob/master/protocols/horizon/operations/main.go#L388-L410) derived from SAC events are also exposed in the horizon operations endpoint response:

```go
// Type   - refers to the source SAC Event
//
//	it can only be one of 'transfer', 'mint', 'clawback' or 'burn'
//
// From   - this is classic account that asset balance was changed,
//
//	or absent if not applicable for function
//
// To     - this is the classic account that asset balance was changed,
//
//	         or absent if not applicable for function
//
//		for asset contract event type, it can be absent such as 'burn'
//
// Amount - expressed as a signed decimal to 7 digits precision.
// Asset  - the classic asset expressed as issuer and code.
type AssetContractBalanceChange struct {
	base.Asset
	Type   string `json:"type"`
	From   string `json:"from,omitempty"`
	To     string `json:"to,omitempty"`
	Amount string `json:"amount"`
}
```

We might need to update the response schema to include muxed information which was added to [cap-0067](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0067.md).

PR #5744, opened 2025-07-07
Add logic for muxed info in effects processor

karthikiyer56:
This was a miss on my part to add muxedInfo to the effects endpoint.
The main issue that this fixes is #5662


karthikiyer56:
There is no need to update the Changelog, since no new field is added in any REST endpoint as a part of this PR.
The existing `account_muxed` and `account_muxed_id` fields in the effects response will correspond to the appropriate muxed information parsed from the SAC event

services/horizon/internal/ingest/processors/effects_processor.go
```diff
@@ -1491,12 +1502,34 @@ func (e *effectsWrapper) addInvokeHostFunctionEffects(events []xdr.ContractEvent
 			}
 
 			if strkey.IsValidEd25519PublicKey(evt.To) {
```
Shaptic, line 1504: Won’t this check fail if it’s an M address? Or is it still a G in that case and only the id is present.

services/horizon/internal/ingest/processors/effects_processor.go
```diff
@@ -1491,12 +1502,34 @@ func (e *effectsWrapper) addInvokeHostFunctionEffects(events []xdr.ContractEvent
...
+					if err != nil {
+						return errors.Wrapf(err, "invokeHostFunction asset details error in generating destination muxed address")
+					}
+					destMuxedAddress = muxedAcc.Address()
+				}
+				var err error
+				if destMuxedAddress == "" {
+					err = e.add(
+						evt.To,
+						null.String{},
```
Shaptic, line 1520: I know this is the same as the previous code but for my own edification why is this the null string?

services/horizon/internal/ingest/processors/effects_processor.go
```diff
@@ -1491,12 +1502,34 @@ func (e *effectsWrapper) addInvokeHostFunctionEffects(events []xdr.ContractEvent
...
+				} else {
+					err = e.add(
+						evt.To,
+						null.StringFrom(destMuxedAddress),
+						history.EffectAccountCredited,
+						toDetails)
+				}
+
+				if err != nil {
 					return errors.Wrapf(err, "invokeHostFunction asset details from contract xfr-to had an error")
```
Shaptic: Suggestion: better error message? At the very least xfr is a typo

services/horizon/internal/ingest/processors/effects_processor.go
```diff
@@ -1509,13 +1542,34 @@ func (e *effectsWrapper) addInvokeHostFunctionEffects(events []xdr.ContractEvent
...
+						null.String{},
+						history.EffectAccountCredited,
+						details)
+				} else {
+					err = e.add(
+						evt.To,
+						null.StringFrom(destMuxedAddress),
+						history.EffectAccountCredited,
+						details)
+				}
```
Shaptic, line 1569: Should this be a function given the repeated code? Or is it different enough arg-wise to warrant separate implementation.

Shaptic (approved):
Gonna trust the tests are correct, LGTM implementation wise

services/horizon/internal/ingest/processors/effects_processor.go
karthikiyer56: Its supposed to fail in the case of an M-address.
We dont expect M-srkey to ever show up in the to/from in the contractEvent

services/horizon/internal/ingest/processors/effects_processor.go
karthikiyer56: This is the signature of the `e.add` function. The code expects to pass in a null if there is no muxedAddress to be had
`func (e *effectsWrapper) add(address string, addressMuxed null.String, effectType history.EffectType, details map[string]interface{}) error `

services/horizon/internal/ingest/processors/effects_processor.go
karthikiyer56: I agree its lame.
I cant think of anything else.
I hate to play the "its existing code" card, but in this case, I just might 🙈

karthikiyer56: alright.
reworded it to be better (?)

services/horizon/internal/ingest/processors/effects_processor.go
karthikiyer56: Its different enough and the effort to make it a function is not worth it

PR #5748, opened 2025-07-10
services/galexie: add implementation for S3 and S3-compatible data storage.

overcat:
<details>
  <summary>PR Checklist</summary>
  
### PR Structure

* [x] This PR has reasonably narrow scope (if not, break it down into smaller PRs).
* [x] This PR avoids mixing refactoring changes with feature changes (split into two PRs
  otherwise).
* [x] This PR's title starts with name of package that is most changed in the PR, ex.
  `services/friendbot`, or `all` or `doc` if the changes are broad or impact many
  packages.

### Thoroughness

* [x] This PR adds tests for the most critical parts of the new functionality or fixes.
* [x] I've updated any docs ([developer docs](https://developers.stellar.org/api/), `.md`
  files, etc... affected by this change). Take a look in the `docs` folder for a given service,
  like [this one](https://github.com/stellar/go/tree/master/services/horizon/internal/docs).

### Release planning

* [x] I've reviewed the changes in this PR and if I consider them worthwhile for being mentioned on release notes then I have updated the relevant `CHANGELOG.md` within the  component folder structure. For example, if I changed horizon, then I updated ([services/horizon/CHANGELOG.md](services/horizon/CHANGELOG.md). I add a new line item describing the change and reference to this PR. If I don't update a CHANGELOG, I acknowledge this PR's change may not be mentioned in future release notes.  
* [x] I've decided if this PR requires a new major/minor version according to
  [semver](https://semver.org/), or if it's mainly a patch change. The PR is targeted at the next
  release branch if it's not a patch change.
</details>

### What

Closes #5746

### Why

See #5746 

### Known limitations

Integration tests have not been added yet.

PR #5749, opened 2025-07-12
services/galexie: add integration tests for S3 storage.

overcat:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

### Release planning

</details>

### What

This is the follow-up PR for #5748. Let's merge #5748 first, and then we'll handle this PR.

### Why

See #5748

### Known limitations
N/A


urvisavla:
@overcat we've been a bit busy with the Protocol 23 release but we'll review this soon. Thanks!

.github/workflows/galexie.yml
```diff
@@ -36,10 +40,16 @@ jobs:
         run: |
           docker pull "$GALEXIE_INTEGRATION_TESTS_QUICKSTART_IMAGE"
 
+      - name: Pull LocalStack image (for S3)
+        if: ${{ matrix.storage_type == 'S3' }}
+        shell: bash
+        run: docker pull localstack/localstack:latest
```
tamirms: I think it would be better to use a fixed version. the `latest` tag is mutable and it's possible that there could be some updates which break our tests.

overcat: Good idea, I fixed the tag. e8aa374a9bac1082c3478b2da0f611258edd77ad

overcat:
> @overcat this looks great! could you add one more test case to the integration tests? I don't think we have any coverage which exercises the code path where we receive a precondition failed error upon trying to insert a file that already exists. I think the following test case should exercise that path:
> 
> 1. initialize empty bucket
> 2. scan-and-fill --start 4 --end 10
> 3. scan-and-fill --start 8 --end 15
> 
> In step (3) galexie should try to insert ledger files 8, 9, and 10. But, it should receive a precondition failed error from AWS and handle that error by skipping over to the next files

Hi @tamirms, I would like the existing [TestAppend](https://github.com/stellar/go/blob/e8aa374a9bac1082c3478b2da0f611258edd77ad/services/galexie/internal/integration_test.go#L110-L115) test to include this scenario. Theoretically, it is best if we can ensure that the object has not changed, but the current datastore interface does not provide this capability. If necessary, we can add a function to the datastore to return the object's identifier.

tamirms:
> I would like the existing [TestAppend](https://github.com/stellar/go/blob/e8aa374a9bac1082c3478b2da0f611258edd77ad/services/galexie/internal/integration_test.go#L110-L115) test to include this scenario.

sure, that sounds good!

>Theoretically, it is best if we can ensure that the object has not changed, but the current datastore interface does not provide this capability. If necessary, we can add a function to the datastore to return the object's identifier.

I'm not sure what would be the equivalent of that for GCS. Alternatively, you could return the last modified timestamp of the object. I believe that is available for both GCS and S3

overcat:
Hey @tamirms, I think 6804884f11d9b6a8abc0ed0b3fa7f5c06c63816a might be what you're looking for. Let me know if I'm on the right track so I can get those unit tests for `GetFileLastModified` added! cc @urvisavla

tamirms:
>  Let me know if I'm on the right track so I can get those unit tests for GetFileLastModified added!

@overcat that looks good to me!

tamirms:
hi @overcat , I just want to check if you will be updating TestAppend as you mentioned in this [comment](https://github.com/stellar/go/pull/5749#issuecomment-3091464407) or if you plan to do that in a separate pr?

overcat:
Hi @tamirms, I have already added it in 6804884f11d9b6a8abc0ed0b3fa7f5c06c63816a

https://github.com/stellar/go/blob/6804884f11d9b6a8abc0ed0b3fa7f5c06c63816a/services/galexie/internal/integration_test.go#L134-L137

services/galexie/internal/integration_test.go
```diff
@@ -92,6 +111,12 @@ func (s *GalexieTestSuite) TestAppend() {
 	err := rootCmd.ExecuteContext(s.ctx)
```
tamirms, line 134: the comment on line 108, "first populate ledgers 4-5", is incorrect. It should be:

// first populate ledgers 6-7

overcat: Fixed 4e87b60f88688bc55fa31768a0e93f7e43dab609

